PR #6949, opened 2024-04-25
Fast subgroup checks for BLS precompile

Marchhill:
## Changes

- Implement fast subgroup checks for EIP-2537

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


src/Nethermind/Nethermind.Evm/Precompiles/Bls/SubgroupChecks.cs
```diff
@@ -0,0 +1,46 @@
...
+{
+    public static readonly UInt256 Beta = new((byte[])[0x5F,0x19,0x67,0x2F,0xDF,0x76,0xCE,0x51,0xBA,0x69,0xC6,0x07,0x6A,0x0F,0x77,0xEA,0xDD,0xB3,0xA9,0x3B,0xE6,0xF8,0x96,0x88,0xDE,0x17,0xD8,0x13,0x62,0x0A,0x00,0x02,0x2E,0x01,0xFF,0xFF,0xFF,0xFE,0xFF,0xFE], true);
+    public static bool G1IsInSubGroup(ReadOnlySpan<byte> p)
+    {
+        UInt256 x = new(p[..64], true);
+
+        Span<byte> phi = stackalloc byte[128];
+        UInt256 xNew = Beta * x;
+        xNew.ToBigEndian().CopyTo(phi);
+        p[64..].CopyTo(phi);
```
LukaszRozmej, line 24: second copy is overriding first copy? Probably you want to CopyTo a slice, or with an offset.

LukaszRozmej: ToBigEndian can receive target, so no 2nd copy is needed.

src/Nethermind/Nethermind.Evm/Precompiles/Bls/SubgroupChecks.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        Span<byte> mulArgs = stackalloc byte[128 + 32];
+        Span<byte> p2 = stackalloc byte[128];
+        p.CopyTo(mulArgs);
+        (x * x).ToBigEndian().CopyTo(mulArgs);
```
LukaszRozmej, line 29: same problem

src/Nethermind/Nethermind.Evm/Precompiles/Bls/SubgroupChecks.cs
```diff
@@ -0,0 +1,46 @@
...
+        Pairings.BlsG1Mul(mulArgs, p2);
+
+        Span<byte> addArgs = stackalloc byte[2 * 128];
+        Span<byte> res = stackalloc byte[128];
+
+        phi.CopyTo(addArgs);
+        p2.CopyTo(addArgs[128..]);
```
LukaszRozmej, line 36: I think all those copies are not needed.
You can stackalloc one bigger buffer and then use its slices in more optimal way, so you don't need to copy back and forth.

src/Nethermind/Nethermind.Evm/Precompiles/Bls/SubgroupChecks.cs
```diff
@@ -0,0 +1,46 @@
...
+        Pairings.BlsG1Add(addArgs, res);
+
+        return res.ToArray().All(x => x == 0);
```
LukaszRozmej, line 39: that to Array is not needed, you can use `IndexOfAnyExcept`

LukaszRozmej (changes requested):

src/Nethermind/Nethermind.Evm/Precompiles/Bls/SubgroupChecks.cs
```diff
@@ -0,0 +1,46 @@
...
+    }
```
LukaszRozmej, line 40: Stack optimized version
```suggestion
    [SkipLocalsInit]
    public static bool G1IsInSubGroup(ReadOnlySpan<byte> p)
    {
        UInt256 x = new(p[..64], true);
        UInt256.Multiply(x, Beta, out x);

        Span<byte> addArgs = stackalloc byte[2 * 128];
        x.ToBigEndian(addArgs[..64]);
        p[64..].CopyTo(addArgs[64..128]);

        Span<byte> mulArgs = stackalloc byte[128 + 32];
        p.CopyTo(mulArgs);
        x = new(p[..64], true);
        UInt256.Multiply(x, x, out x);
        x.ToBigEndian(mulArgs[128..]);
        Pairings.BlsG1Mul(mulArgs, addArgs[128..]);

        Span<byte> res = mulArgs[..128];
        Pairings.BlsG1Add(addArgs, res);

        return res.IndexOfAnyExcept((byte)0) == -1;
    }
```

Marchhill: Thanks, the initial state of the PR was quite rough I should have written it wasn't ready for review, my bad! Cut down on the copies now. I'm not sure if `[SkipLocalsInit]` is safe to do here as the buffers should all be initialised to 0

PR #6951, opened 2024-04-26
Perf/read rpc message to end

LukaszRozmej:
## Changes

- Reads JSON RPC message to end before trying to deserialize it.
- Helps with performance on very long requests (which are fairly rare and less realistic scenarios)

- [x] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

Would be worth comparing performance on very many batched requests (as many as you can under 30MB request limit) as it might be better or worse than original.

@marcindsobczak Would be worth comparing NewPayload performance with big but reasonable size blocks.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
+++ b/src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
@@ -137,9 +137,7 @@ public class JsonRpcProcessor : IJsonRpcProcessor
         try
         {
             // Asynchronously reads data from the PipeReader.
-            ReadResult readResult = _jsonRpcConfig.MaxRequestBodySize is not null
-                ? await reader.ReadAtLeastAsync((int)_jsonRpcConfig.MaxRequestBodySize.Value)
-                : await reader.ReadToEndAsync();
+            ReadResult readResult = await reader.ReadToEndAsync();
 
             buffer = readResult.Buffer;
             // Placeholder for a result in case of deserialization failure.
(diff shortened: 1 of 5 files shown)


src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
```diff
@@ -137,9 +137,7 @@ JsonRpcResult GetParsingError(string error, Exception? exception = null)
         try
         {
             // Asynchronously reads data from the PipeReader.
-            ReadResult readResult = _jsonRpcConfig.MaxRequestBodySize is not null
-                ? await reader.ReadAtLeastAsync((int)_jsonRpcConfig.MaxRequestBodySize.Value)
-                : await reader.ReadToEndAsync();
```
LukaszRozmej: I'm torn about this one, @benaadams ?

benaadams: Am wondering about adding a timeout also?

benaadams: vs a very slow request; like if the full request takes more than 1sec to receive chop it?

LukaszRozmej: We have some global timeouts on json rpc

benaadams (approved):

src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
LukaszRozmej: Will think about it a bit more.

PR #6952, opened 2024-04-26
Skip hash validations when block comes from NewPayload 

LukaszRozmej:
## Changes

- If block comes from NewPayload, we calculated those hashes in `TryGetBlock`, no need to calculate them again.

- [x] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Validators/BlockValidator.cs
+++ b/src/Nethermind/Nethermind.Consensus/Validators/BlockValidator.cs
@@ -88,15 +88,17 @@ public class BlockValidator : IBlockValidator
     {
         return ValidateSuggestedBlock(block, out _);
     }
+
     /// <summary>
     /// Suggested block validation runs basic checks that can be executed before going through the expensive EVM processing.
     /// </summary>
     /// <param name="block">A block to validate</param>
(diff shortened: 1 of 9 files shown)


benaadams (approved):

src/Nethermind/Nethermind.Consensus/Validators/BlockValidator.cs
```diff
@@ -134,15 +136,20 @@ public bool ValidateSuggestedBlock(Block block, out string? errorMessage)
...
+                errorMessage = BlockErrorMessages.InvalidTxRoot(block.Header.TxRoot!, txRoot);
+                return false;
+            }
 
-        if (!ValidateWithdrawals(block, spec, out errorMessage))
-            return false;
+            if (!ValidateWithdrawals(block, spec, out errorMessage))
+            {
+                return false;
+            }
```
smartprogrammer93, line 151: Why is this inside the if statement of `validatehashes`?

PR #6957, opened 2024-04-27
Optimize Rlp Encoding

benaadams:
## Changes

- Use FrozenDictionary for encoder lookup
- Use struct wrapper for `Type` for dicitionary devirtualization
- Use `Instance` decoder for `BlockInfo`, `Account`, `LogInfo` rather than type lookup
- Use `Instance` decoder for `Transaction` rather than instatiating new one each time.
- Use 256bit assignment rather than Span copy for Hash256 encoding
- Seal `TxDecoder` and `TrieNode`
- Split Rlp and Non-Rlp key generation into seperate methods rather than checking on each loop iteration.

- [x] Optimization
- [x] Refactoring

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
+++ b/src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
@@ -77,7 +77,7 @@ namespace Nethermind.Serialization.Rlp
 
         private static Dictionary<TypeAsKey, IRlpDecoder> _decoderBuilder = new();
         private static FrozenDictionary<TypeAsKey, IRlpDecoder>? _decoders;
-        public static FrozenDictionary<TypeAsKey, IRlpDecoder> Decoders = _decoders ??= _decoderBuilder.ToFrozenDictionary();
+        public static FrozenDictionary<TypeAsKey, IRlpDecoder> Decoders => _decoders ??= _decoderBuilder.ToFrozenDictionary();
 
         public struct TypeAsKey(Type key) : IEquatable<TypeAsKey>
         {
--- a/src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
+++ b/src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
@@ -75,7 +75,9 @@ namespace Nethermind.Serialization.Rlp
 
         public int Length => Bytes.Length;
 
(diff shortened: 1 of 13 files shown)


src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
```diff
@@ -73,10 +75,32 @@ public Rlp(byte[] bytes)
...
+            public override int GetHashCode() => _key.GetHashCode();
+        }
+
+        public static void RegisterDecoder(Type type, IRlpDecoder decoder)
+        {
+            Dictionary<TypeAsKey, IRlpDecoder> dict = new(Decoders)
+            {
+                [type] = decoder
+            };
+            Decoders = dict.ToFrozenDictionary();
```
LukaszRozmej, line 98: should we eagerly create frozen dictionary or better create it lazy?

benaadams: Something like? https://github.com/NethermindEth/nethermind/pull/6957/commits/052d8b0982dceff711f8a2533e08447cd66ab9e4

LukaszRozmej (approved):

src/Nethermind/Nethermind.Trie/TrieNode.cs
```diff
@@ -507,7 +509,10 @@ public void ResolveKey(ITrieNodeResolver tree, ref TreePath path, bool isRoot, I
             if (rlp is null || IsDirty)
             {
                 ref readonly CappedArray<byte> oldRlp = ref rlp is not null ? ref rlp.Data : ref CappedArray<byte>.Empty;
-                CappedArray<byte> fullRlp = RlpEncode(tree, ref path, bufferPool);
+                CappedArray<byte> fullRlp = NodeType == NodeType.Branch ?
```
flcl42, line 512: RlpEncode contains this check already, does it optimize something?

benaadams: Removes about 11 stack frames since Branch is the most common and RlpEncode doesn't inline or tailcall because the return is larger than a pointer size

benaadams: Will try to get a trace that shows what it looks like now

benaadams: After PR

image

Before it has 3 more methods on each recursion

Issue #6690, opened 2024-02-11
Add post state values for block execution to the execution witness


tanishqjasoria:
Post verkle, we include ExecutionWitness in the ExecutionPayload
Draft PR: https://github.com/ethereum/consensus-specs/pull/3230/files
In ExecutionWitness, aside from VerkleProof, we also include Keys that were modifed and the corresponding values before(pre-state) and after(post-state) modification.

In the current implemenation, we already include the kpre-state values, now we just need to collect the changed values and add it to the ExecutionWitness.

shubham-sonthalia:
Hi @tanishqjasoria is this issue still relevant? Can I give it a shot?

tanishqjasoria:
yes, this is still relevent.

PR #6966, opened 2024-05-01
Add post state values for ExecutionWitness

yerke26:
## Changes

Completed generation of ExecutionWitness by adding post state NewValue
https://github.com/NethermindEth/nethermind/issues/6690

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Remarks

### DIFF
--- a/src/Nethermind/Nethermind.Verkle.Tree/VerkleTree.Prover.cs
+++ b/src/Nethermind/Nethermind.Verkle.Tree/VerkleTree.Prover.cs
@@ -39,28 +39,54 @@ public partial class VerkleTree
 
         VerkleProof proof = CreateVerkleProof(keys, out rootPoint);
 
-        SpanDictionary<byte, List<SuffixStateDiff>> stemStateDiff = new(Bytes.SpanEqualityComparer);
+        SpanDictionary<byte, List<SuffixStateDiff>> stemStateDiffWithCurrentValue = new(Bytes.SpanEqualityComparer);
         foreach (var key in keys)
         {
             SuffixStateDiff suffixData = new() { Suffix = key[31], CurrentValue = Get(key) };
             Span<byte> keyStem = key.AsSpan()[..31];
-            if (!stemStateDiff.TryGetValue(keyStem, out List<SuffixStateDiff> suffixStateDiffList))
+            if (!stemStateDiffWithCurrentValue.TryGetValue(keyStem, out List<SuffixStateDiff> suffixStateDiffList))
             {
                 suffixStateDiffList = [];
-                stemStateDiff.TryAdd(keyStem, suffixStateDiffList);
+                stemStateDiffWithCurrentValue.TryAdd(keyStem, suffixStateDiffList);
             }
             suffixStateDiffList.Add(suffixData);
         }
 
-        var stemStateDiffList = new StemStateDiff[stemStateDiff.Count];
+        var stemStateDiffList = new StemStateDiff[stemStateDiffWithCurrentValue.Count];
         int i = 0;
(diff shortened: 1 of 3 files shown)


tanishqjasoria (changes requested):
do not commit inside a function randomly, we like to commit everything at once at the end of block execution - you can see that in the blockProcessor

better idea would be to get the execution witness and then after the commit is done then get the post-state values  and update the execution witness accordingly.

or we store stateDiff for each block - for the values that were changed - maybe you can use that to get the both pre and post state values after the commit.

src/Nethermind/Nethermind.Verkle.Tree/VerkleTree.Prover.cs
```diff
@@ -63,6 +63,32 @@ public ExecutionWitness GenerateExecutionWitness(byte[][] keys, out Banderwagon
         return new ExecutionWitness(stemStateDiffList, proof);
     }
 
+    public void UpdateExecutionWitness(ExecutionWitness executionWitness, byte[][] keys)
```
tanishqjasoria, line 66: rename to a more appropiate name - like UpdateWithPostStateValues

tanishqjasoria: also no need to pass keys here - you already have that information in executionWitness

src/Nethermind/Nethermind.Verkle.Tree/VerkleTree.Prover.cs
```diff
@@ -63,6 +63,32 @@ public ExecutionWitness GenerateExecutionWitness(byte[][] keys, out Banderwagon
...
+        {
+            List<SuffixStateDiff> newValueSuffixDiffs = stemStateDiffWithNewValue[stateDiff.Stem.Bytes];
+            int i = 0;
+            stateDiff.SuffixDiffs.ForEach(suffixDiffs =>
+            {
+                suffixDiffs.NewValue = newValueSuffixDiffs[i].NewValue;
+                i++;
+            });
+        });
+    }
```
tanishqjasoria, line 90: you can just directly iterate over executionWitness and then query value and assign it directly there. 
no need to allocate a new dictionary.

tanishqjasoria (changes requested):

src/Nethermind/Nethermind.Verkle.Tree/VerkleTree.Prover.cs
```diff
@@ -63,6 +63,18 @@ public ExecutionWitness GenerateExecutionWitness(byte[][] keys, out Banderwagon
         return new ExecutionWitness(stemStateDiffList, proof);
     }
 
+    public void UpdateWithPostStateValues(ExecutionWitness executionWitness)
+    {
+        executionWitness.StateDiff.ForEach(stateDiff =>
+        {
+            stateDiff.SuffixDiffs.ForEach(suffixDiffs =>
+            {
+                byte[] key = stateDiff.Stem.Bytes.Append(suffixDiffs.Suffix).ToArray();
```
tanishqjasoria, line 72: this creates a lot of allocations 
what you can do here is something like this

use stackalloc to allocate a 32 byte array 
in the first loop. then just copy the stem to the first 31 bytes and then in the second loop update the last byte to query the required key

tanishqjasoria (changes requested):
just one changes, everything else looks fine

tanishqjasoria (approved):

PR #6967, opened 2024-05-02
Less allocations for node processing

benaadams:
## Changes

- Introduce struct enumerators rather than using Linq

Addresses the following allocations
image

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Core/Collections/ArrayPoolList.cs
+++ b/src/Nethermind/Nethermind.Core/Collections/ArrayPoolList.cs
@@ -5,7 +5,6 @@ using System;
 using System.Buffers;
 using System.Collections;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Runtime.CompilerServices;
 using System.Threading;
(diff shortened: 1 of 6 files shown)


src/Nethermind/Nethermind.Network.Discovery/RoutingTable/NodeTable.cs
```diff
@@ -73,42 +76,140 @@ public void RefreshNode(Node node)
...
+
+    public ClosestNodesFromNodeEnumerator GetClosestNodes(byte[] nodeId, int bucketSize)
     {
         CheckInitialization();
+        return new ClosestNodesFromNodeEnumerator(Buckets, nodeId, _nodeDistanceCalculator, Math.Min(bucketSize, _discoveryConfig.BucketSize));
+    }
+
+    public struct ClosestNodesFromNodeEnumerator : IEnumerator<Node>, IEnumerable<Node>
+    {
+        private readonly List<Node> _sortedNodes;
```
LukaszRozmej, line 163: ArrayPoolList?

src/Nethermind/Nethermind.Network.Discovery/RoutingTable/NodeBucket.cs
```diff
@@ -28,26 +29,71 @@ public NodeBucket(int distance, int bucketSize, float dropFullBucketProbability
...
-                    node = node.Previous;
+                    return true;
                 }
             }
+
+            Current = null!;
+            return false;
+        }
+
+        public void Reset() => throw new NotSupportedException();
```
LukaszRozmej, line 80: Should we have Reset's throw not supported exceptions? Shouldn't be hard to reset all of them?

benaadams: Not if never use it, docs say don't bother https://learn.microsoft.com/en-us/dotnet/api/system.collections.ienumerator.reset?view=net-8.0#remarks

image


Hid behind explicit implementation https://github.com/NethermindEth/nethermind/pull/6967/commits/e2586909ec940bb9271620645abedd6ddfe058a9

src/Nethermind/Nethermind.Network.Discovery/RoutingTable/NodeTable.cs
benaadams: Changed to use ArrayPoolList https://github.com/NethermindEth/nethermind/pull/6967/commits/a144110a4b712760e32dabac8f9c60e2a758b9d2

src/Nethermind/Nethermind.Core/Collections/ArrayPoolList.cs
```diff
@@ -121,6 +121,49 @@ void ICollection.CopyTo(Array array, int index)
...
+
+        Count = count;
+        if (count < _capacity / 2)
+        {
+            // Reduced to less than half of the capacity, resize the array.
+            T[] newArray = _arrayPool.Rent(count);
+            _array.AsSpan(0, count).CopyTo(newArray);
+            T[] oldArray = Interlocked.Exchange(ref _array, newArray);
+            _capacity = newArray.Length;
+            _arrayPool.Return(oldArray);
```
LukaszRozmej, line 143: Is it worth it, to potentially copy over this data to the new buffer? 
I expect this data to be consumed fairly quickly and the whole buffer would return.

IMO probably not worth it at

LukaszRozmej (approved):

src/Nethermind/Nethermind.Core/Collections/ArrayPoolList.cs
benaadams: The consumers only want 12 or 16 elements by default; so copy isn't that big, but can be hundreds or thousands of nodes

PR #6970, opened 2024-05-02
Reduce TxPool allocations

benaadams:
## Changes

- Don't allocate Linq for .ToArray snapshot
- Invert the flow to pass method into the iterator, rather than passing enumerator out

Removes the following allocations 
image

- [x] Optimization
- [x] Refactoring

#### Requires testing

- [x] No


src/Nethermind/Nethermind.TxPool/Collections/TxDistinctSortedPool.cs
```diff
@@ -80,46 +85,49 @@ public void UpdatePool(IAccountStateProvider accounts, Func<Address, AccountStru
...
                 {
-                    tx.GasBottleneck = changedGasBottleneck;
+                    _worstSortedValues.Add(tx, tx.Hash!);
                 }
-            }
 
-            ReadOnlySpan<Transaction> txs = CollectionsMarshal.AsSpan(_transactionsToRemove);
-            for (int i = 0; i < txs.Length; i++)
+                UpdateWorstValue();
+                lastElement = bucket.Max;
```
benaadams, line 122: LastElement was previously never changed for comparision even though the worse value was changed

src/Nethermind/Nethermind.TxPool/Collections/TxDistinctSortedPool.cs
```diff
@@ -80,46 +85,49 @@ public void UpdatePool(IAccountStateProvider accounts, Func<Address, AccountStru
                 Debug.Assert(bucket.Count > 0);
 
                 accounts.TryGetAccount(address, out AccountStruct account);
-                UpdateGroupNonLocked(address, account, bucket, changingElements);
+                UpdateGroupNonLocked(address, account, bucket, updateElements);
             }
         }
 
-        private void UpdateGroupNonLocked(Address groupKey, AccountStruct groupValue, EnhancedSortedSet<Transaction> bucket, Func<Address, AccountStruct, EnhancedSortedSet<Transaction>, IEnumerable<(Transaction Tx, UInt256? changedGasBottleneck)>> changingElements)
+        private void UpdateGroupNonLocked(Address groupKey, AccountStruct groupValue, EnhancedSortedSet<Transaction> bucket, UpdateGroupDelegate updateElements)
```
LukaszRozmej, line 92: `groupkey` is not used?

src/Nethermind/Nethermind.TxPool/Collections/TxDistinctSortedPool.cs
```diff
@@ -17,6 +17,10 @@ namespace Nethermind.TxPool.Collections
 {
     public class TxDistinctSortedPool : DistinctValueSortedPool<ValueHash256, Transaction, AddressAsKey>
     {
+        public delegate void UpdateGroupDelegate(in AccountStruct account, EnhancedSortedSet<Transaction> transactions, ref Transaction? lastElement, UpdateTransactionDelegate updateTx);
+        public delegate void UpdateTransactionDelegate(EnhancedSortedSet<Transaction> bucket, Transaction tx, in UInt256? changedGasBottleneck, ref Transaction? lastElement);
+
+        private readonly UpdateTransactionDelegate _updateTx;
         private readonly List<Transaction> _transactionsToRemove = new();
         protected int _poolCapacity;
         private readonly ILogger _logger;
```
LukaszRozmej, line 26: logger not used?

src/Nethermind/Nethermind.TxPool/Collections/TxDistinctSortedPool.cs
LukaszRozmej: Transactions in buckets are sorted by nonces, so changing gas bottleneck cannot change order

LukaszRozmej (commented):
Overall looks good, but I want @marcindsobczak to review this as well

src/Nethermind/Nethermind.TxPool/Collections/TxDistinctSortedPool.cs
benaadams: Nope

src/Nethermind/Nethermind.TxPool/Collections/TxDistinctSortedPool.cs
benaadams: To use `GetEnumerator` and have structs is fiendisly complicated as there are 3 types of Enumerators two at same level and one which needs to be embedded in the others; having them as IEnumeratable causes multiple allocations on every tx add to the pool (which is the whole tx pool every block + others)

benaadams:
> How much allocation does it save? All the marked ones?

Yep

src/Nethermind/Nethermind.TxPool/Collections/TxDistinctSortedPool.cs
benaadams: Removed

benaadams:
> How much allocation does it save? All the marked ones?

Top allocations on master + PR now looks like this

image

LukaszRozmej (approved):
Approve from me, but please don't merge until @marcindsobczak reviews!

benaadams:
> Top allocations on master + PR now looks like this

With `string` changes

image

src/Nethermind/Nethermind.TxPool/Collections/TxDistinctSortedPool.cs
LukaszRozmej: remove?

benaadams: Haha, thought I had. Have now

marcindsobczak (approved):
Looks good. Took me a while to understand what is going on here, but actually it is just delegating low level update logic to be triggered from `TxPool` instead of  `TxDistinctSortedPool`. I don't see any inconsistencies

PR #6971, opened 2024-05-02
Reduce string allocs by networking

benaadams:
## Changes

- Use `IPAddress` comparision rather than string comparision
- Use ReadOnlySpan<byte> Utf8 comparision rather than materializing string
- Lazy create `Host`, `Port` strings in `Node`
- Pre-calucate common `Port` strings

Bunch of unnessary strings being created

image

image

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Network.Enr/NodeRecordSigner.cs
+++ b/src/Nethermind/Nethermind.Network.Enr/NodeRecordSigner.cs
@@ -54,53 +54,43 @@ public class NodeRecordSigner : INodeRecordSigner
         while (rlpStream.Position < startPosition + recordRlpLength)
         {
             ReadOnlySpan<byte> key = rlpStream.DecodeByteArraySpan();
-            if (key.Length == 2)
+            switch (key.Length)
             {
-                if (key.SequenceEqual(EnrContentKey.IdU8))
-                {
+                case 2 when key.SequenceEqual(EnrContentKey.IdU8):
                     rlpStream.SkipItem();
                     nodeRecord.SetEntry(IdEntry.Instance);
-                }
-                else if (key.SequenceEqual(EnrContentKey.IpU8))
-                {
+                    break;
+                case 2 when key.SequenceEqual(EnrContentKey.IpU8):
                     ReadOnlySpan<byte> ipBytes = rlpStream.DecodeByteArraySpan();
                     IPAddress address = new(ipBytes);
                     nodeRecord.SetEntry(new IpEntry(address));
-                }
-            }
-            else if (key.Length == 3)
-            {
-                if (key.SequenceEqual(EnrContentKey.EthU8))
-                {
+                    break;
+                case 3 when key.SequenceEqual(EnrContentKey.EthU8):
                     _ = rlpStream.ReadSequenceLength();
                     _ = rlpStream.ReadSequenceLength();
                     byte[] forkHash = rlpStream.DecodeByteArray();
                     long nextBlock = rlpStream.DecodeLong();
                     nodeRecord.SetEntry(new EthEntry(forkHash, nextBlock));
-                }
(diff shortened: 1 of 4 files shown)


src/Nethermind/Nethermind.Network.Enr/NodeRecordSigner.cs
```diff
@@ -52,44 +53,54 @@ public NodeRecord Deserialize(RlpStream rlpStream)
...
+            {
                 // snap
-                default:
-                    canVerify = false;
-                    rlpStream.SkipItem();
-                    nodeRecord.Snap = true;
-                    break;
+                canVerify = false;
+                rlpStream.SkipItem();
+                nodeRecord.Snap = true;
```
LukaszRozmej, line 103: Maybe switch?

```
            switch (key.Length)
            {
                case 2 when key.SequenceEqual(EnrContentKey.IdU8):
                    rlpStream.SkipItem();
                    nodeRecord.SetEntry(IdEntry.Instance);
                    break;
                case 2:
                    if (key.SequenceEqual(EnrContentKey.IpU8))
                    {
                        ReadOnlySpan<byte> ipBytes = rlpStream.DecodeByteArraySpan();
                        IPAddress address = new(ipBytes);
                        nodeRecord.SetEntry(new IpEntry(address));
                    }

                    break;
                case 3 when key.SequenceEqual(EnrContentKey.EthU8):
                    _ = rlpStream.ReadSequenceLength();
                    _ = rlpStream.ReadSequenceLength();
                    byte[] forkHash = rlpStream.DecodeByteArray();
                    long nextBlock = rlpStream.DecodeLong();
                    nodeRecord.SetEntry(new EthEntry(forkHash, nextBlock));
                    break;
                case 3 when key.SequenceEqual(EnrContentKey.TcpU8):
                    int tcpPort = rlpStream.DecodeInt();
                    nodeRecord.SetEntry(new TcpEntry(tcpPort));
                    break;
                case 3:
                    if (key.SequenceEqual(EnrContentKey.UdpU8))
                    {
                        int udpPort = rlpStream.DecodeInt();
                        nodeRecord.SetEntry(new UdpEntry(udpPort));
                    }

                    break;
                default:
                    if (key.SequenceEqual(EnrContentKey.Secp256K1U8))
                    {
                        ReadOnlySpan<byte> keyBytes = rlpStream.DecodeByteArraySpan();
                        CompressedPublicKey reportedKey = new(keyBytes);
                        nodeRecord.SetEntry(new Secp256K1Entry(reportedKey));
                    }
                    else
                    {
                        // snap
                        canVerify = false;
                        rlpStream.SkipItem();
                        nodeRecord.Snap = true;
                    }

                    break;
            }
```

LukaszRozmej (approved):

src/Nethermind/Nethermind.Network.Enr/NodeRecordSigner.cs
```diff
@@ -54,53 +54,43 @@ public NodeRecord Deserialize(RlpStream rlpStream)
...
+                    ReadOnlySpan<byte> keyBytes = rlpStream.DecodeByteArraySpan();
+                    CompressedPublicKey reportedKey = new(keyBytes);
+                    nodeRecord.SetEntry(new Secp256K1Entry(reportedKey));
+                    break;
+                default:
+                    // snap
+                    canVerify = false;
+                    rlpStream.SkipItem();
+                    nodeRecord.Snap = true;
+                    break;
```
LukaszRozmej, line 93: It is not the same as it was though.

Before if length was 2 or 3 this would never been invoked. Now this will be invoked (but maybe it should?)

benaadams: Was mistake; was invoked in original switch (before the `if` changes)

PR #6977, opened 2024-05-03
Pyspec new dir and state test support

smartprogrammer93:
## Changes

- Support latest release of pyspec tests including state tests.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No


LukaszRozmej (approved):

LukaszRozmej (commented):
There are a lot more directories in the package - I think we should add more test suits to cover them also?

smartprogrammer93:
> There are a lot more directories in the package - I think we should add more test suits to cover them also?

@LukaszRozmej added issue to track here #6978

PR #6981, opened 2024-05-05
Traverse trie branches in same stack frame

benaadams:
## Changes

- Shrink TraverseBranch(es) method (for smaller loop body) and change to a loop as branches can go quite deep, so branches are all traversed in one method call
  - Less stack spaces setup, returned through, etc
- seal `HealingTrieStore` and `ScopedTrieStore`
- Also makes perf traces easier to evaluate as is less branching at lots of different points

Before

image

After

image

Also Before (showing how deep it goes)

image

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Trie/PatriciaTree.cs
+++ b/src/Nethermind/Nethermind.Trie/PatriciaTree.cs
@@ -9,6 +9,7 @@ using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Runtime.CompilerServices;
+using System.Runtime.ExceptionServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Nethermind.Core;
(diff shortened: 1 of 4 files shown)


src/Nethermind/Nethermind.Trie/PatriciaTree.cs
```diff
@@ -646,6 +662,28 @@ void Trace(TrieNode node, in TraverseContext traverseContext)
...
+                    default:
+                        return ref ThrowNotSupported(node);
+                }
+            }
+
+            [DoesNotReturn]
+            [StackTraceHidden]
+            static ref readonly CappedArray<byte> TraverseBranch(TrieNode node)
+            {
+                throw new InvalidOperationException($"Branch node {node.Keccak} should already be handled");
```
asdacap, line 684: I think its possible that it traverse from branch to extension then back to branch.

benaadams: Have tenary inside `TraverseExtension` that calls `TraverseBranches` direcly so should be ok.

benaadams: For Extension 

Before 

image


After 

image

asdacap (approved):
LGTM

PR #6983, opened 2024-05-06
Added CHUNK_FILL_COST

### LINKED ISSUE #6961: implement the CHUNK_FILL_COST for verkle witness

yerke26:
Closes #6961 

## Changes

- CHUNK_FILL_COST in VerkleExecutionWitness

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Remarks

### DIFF
--- a/src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
+++ b/src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
@@ -136,7 +136,8 @@ namespace Nethermind.Evm.TransactionProcessing
             if (commit) WorldState.Commit(spec, tracer.IsTracingState ? tracer : NullTxTracer.Instance);
 
             // declare the execution witness to collect witness and also charge gas
-            IExecutionWitness executionWitness = spec.IsVerkleTreeEipEnabled ? new VerkleExecWitness(LogManager) : new NoExecWitness();
+            var verkleWorldState = WorldState as VerkleWorldState;
+            IExecutionWitness executionWitness = spec.IsVerkleTreeEipEnabled ? new VerkleExecWitness(LogManager, verkleWorldState!) : new NoExecWitness();
             executionWitness.AccessForTransaction(tx.SenderAddress!, tx.To!, !tx.Value.IsZero);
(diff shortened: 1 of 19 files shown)


src/Nethermind/Nethermind.State/VerkleWorldState.cs
```diff
@@ -52,6 +52,11 @@ public class VerkleWorldState : IWorldState
     // // False positives would be problematic as the code _must_ be persisted
     private readonly LruKeyCache<Hash256> _codeInsertFilter = new(2048, "Code Insert Filter");
 
+    public byte[]? GetValue(Hash256 key)
```
tanishqjasoria, line 55: here we only need this to check if the value is present or not - so we dont need to actually return the value
you can just return true and false and have a funciton like `ValuePresentInTree`
Also, you can add similar functions in the tree also, to just check presence - this will be much faster then finding and returning the value

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -136,7 +136,8 @@ protected virtual TransactionResult Execute(Transaction tx, in BlockExecutionCon
             if (commit) WorldState.Commit(spec, tracer.IsTracingState ? tracer : NullTxTracer.Instance);
 
             // declare the execution witness to collect witness and also charge gas
-            IExecutionWitness executionWitness = spec.IsVerkleTreeEipEnabled ? new VerkleExecWitness(LogManager) : new NoExecWitness();
+            var verkleWorldState = WorldState as VerkleWorldState;
```
tanishqjasoria, line 139: we need some kind of error handling here or inside `VerkleExecWitness` class.
make the input nullable and then check in the constructor, if the value is null - throw an error

tanishqjasoria (changes requested):

PR #6929, opened 2024-04-18
Javascript Tracer resource leak

Marchhill:
## Changes

- Ensure that tracer resources (ArrayPoolList) are disposed

#### What types of changes does your code introduce?

- [x] Bugfix (a non-breaking change that fixes an issue)
- [ ] New feature (a non-breaking change that adds functionality)
- [ ] Breaking change (a change that causes existing functionality not to work as expected)
- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

#### If yes, did you write tests?

- [ ] Yes
- [x] No

#### Requires documentation update

- [ ] Yes
- [x] No

#### Requires explanation in Release Notes

- [ ] Yes
- [x] No

PR #6987, opened 2024-05-07
Fix too eager dispose in js block tracing

LukaszRozmej:
## Changes

- Makes engine dispose lazy on serialization, fixing the issue it was disposed too eagerly introduced in #6929
- Disposes engines on exception
- Removes ArrayPool as it is hard to dispose it just in the right time

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

@natebeauregard can you test it?

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Tracing/GethStyleTracer.cs
+++ b/src/Nethermind/Nethermind.Consensus/Tracing/GethStyleTracer.cs
@@ -126,9 +126,10 @@ public class GethStyleTracer : IGethStyleTracer
             _processor.Process(block, ProcessingOptions.Trace, blockTracer.WithCancellation(cancellationToken));
             return blockTracer.BuildResult().SingleOrDefault();
         }
-        finally
+        catch
         {
             blockTracer.TryDispose();
+            throw;
         }
     }
 
@@ -177,9 +178,10 @@ public class GethStyleTracer : IGethStyleTracer
             _processor.Process(block, ProcessingOptions.Trace, tracer.WithCancellation(cancellationToken));
             return tracer.BuildResult().SingleOrDefault();
         }
-        finally
+        catch
         {
             tracer.TryDispose();
+            throw;
         }
     }
 
@@ -212,9 +214,10 @@ public class GethStyleTracer : IGethStyleTracer
             _processor.Process(block, ProcessingOptions.Trace, tracer.WithCancellation(cancellationToken));
             return tracer.BuildResult();
         }
-        finally
+        catch
         {
             tracer.TryDispose();
+            throw;
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/Nethermind/Nethermind.Consensus/Tracing/GethStyleTracer.cs
```diff
@@ -126,9 +126,10 @@ public GethLikeTxTrace Trace(Hash256 blockHash, int txIndex, GethTraceOptions op
             _processor.Process(block, ProcessingOptions.Trace, blockTracer.WithCancellation(cancellationToken));
             return blockTracer.BuildResult().SingleOrDefault();
         }
-        finally
+        catch
```
Marchhill, line 129: Shouldn't we always dispose the tracer, not only when there is an exception?

src/Nethermind/Nethermind.Evm/Tracing/GethStyle/Custom/JavaScript/GethLikeBlockJavaScriptTracer.cs
```diff
@@ -33,7 +33,7 @@ public GethLikeBlockJavaScriptTracer(IWorldState worldState, IReleaseSpec spec,
 
     public override void StartNewBlockTrace(Block block)
     {
-        _engines = new ArrayPoolList<IDisposable>(block.Transactions.Length + 1);
+        _engines = new List<IDisposable>(block.Transactions.Length + 1);
```
Marchhill, line 36: What was the problem with using ArrayPoolList as long as we dispose it? I guess we will have worse performance if we switch to a normal list

src/Nethermind/Nethermind.Consensus/Tracing/GethStyleTracer.cs
LukaszRozmej: So we actually need to dispose the js engines in the tracer and potentially arraypoollist. 
1. I've removed arryapoollist as js tracers are already so slow and so much overhead that it won't make any difference to have it here.
2. we don't want to dispose engine's here as the engine's can have result objects attached to them, we need to serialize those results objects and only then we will be able to dispose the engines. This should be done by our serialization logic.
3. Only on failure we want to make sure we disposed already created engines.

src/Nethermind/Nethermind.Evm/Tracing/GethStyle/Custom/JavaScript/GethLikeBlockJavaScriptTracer.cs
LukaszRozmej: It was hard to properly dispose it in this case.

natebeauregard:
@LukaszRozmej Changes look good, I'll test this later today while gathering the performance numbers to ensure that everything is working ok and will approve then

natebeauregard (approved):
Testing looked good, all of the JS tracers can be executed from the JSON-RPC endpoint. LGTM

PR #6988, opened 2024-05-07
Don't use tracing for non-traced rpc calls

benaadams:
## Changes

- Tracing is used to allow for cancellation, however to also introduces overheads from actually capturing all the trace data, even if it just throws it away.
- Instead introduce a IsCancelled property that can be checked once per loop

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Evm/VirtualMachine.cs
+++ b/src/Nethermind/Nethermind.Evm/VirtualMachine.cs
@@ -818,6 +818,11 @@ internal sealed class VirtualMachine<TLogger> : IVirtualMachine where TLogger : 
 #endif
             Instruction instruction = (Instruction)code[programCounter];
 
+            if (_txTracer.IsCancelled)
+            {
+                ThrowOperationCanceledException();
+            }
+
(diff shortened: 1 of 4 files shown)


LukaszRozmej (approved):

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -818,6 +819,11 @@ private CallResult ExecuteCall<TTracingInstructions>(EvmState vmState, ReadOnlyM
 #endif
             Instruction instruction = (Instruction)code[programCounter];
 
+            if (isCancelable && _txTracer.IsCancelled)
+            {
+                ThrowOperationCanceledException();
+            }
```
LukaszRozmej, line 825: How is that much faster than IsTracingInstructions?

benaadams: IsTracingInstructions == true comes with the additional load of copying every stack data change, memory copy etc

PR #6991, opened 2024-05-07
fix: Use dpkg instead uname to get unified arch in postinst debian script

kayano:
It seems `uname` changed what it returns doing `uname -m`. 
I'm getting `aarch64` running `uname -m` which results in wrong binary being installed when using nethermind ppa on arm64 machine.

## Changes

- use dpkg instead uname in ppa postinst debian script

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [x] Other: Ubuntu packaging

#### Requires testing

- [ ] Yes
- [x] No

### DIFF
--- a/scripts/deployment/ppa/debian/postinst
+++ b/scripts/deployment/ppa/debian/postinst
@@ -18,7 +18,7 @@ set -e
 
 case "$1" in
   configure)
-    arch=$(uname -m)
+    arch=$(dpkg --print-architecture)
 
     if [[ "$arch" == arm* ]]; then
       url=to_be_replaced


rubo:
What OS does return `aarch` instead of `arm`?

kayano:
> What OS does return `aarch` instead of `arm`?

```

lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.4 LTS
Release:	22.04
Codename:	jammy

uname -a
Linux  #18~22.04.1-Ubuntu SMP Fri Apr  5 17:56:39 UTC 2024 aarch64 aarch64 aarch64 GNU/Linux
```

rubo:
Good catch. We'll improve the script further to check the architectures explicitly and exit if not supported instead of falling back to the wrong one. Thank you for your contribution.

rubo (approved):

PR #6999, opened 2024-05-08
Keccak perf

benaadams:
## Changes

- Avoid the copy in and copy out step by operating over same data. Since there are so many fields they spill to stack anyway (and is stackalloc'd array anyway to start), so operate with stack and remove two 200 byte copies, or 50 ulongs of memory shuffling

KeccakF asm

Before
```
; Total bytes of code 2606
````
After
```
; Total bytes of code 1942
```

- [x] Optimization

#### Requires testing

- [x] No

### CODE REVIEW

**Review by LukaszRozmej — COMMENTED** [review]
Do you have any benchmarks?

File: src/Nethermind/Nethermind.Core/Crypto/KeccakHash.cs [hunk 1]
```diff
@@ -104,130 +100,98 @@ private static void KeccakF(Span<ulong> st)
...
+                bCu ^= st.amu();
+
+                bCa ^= st.asa();
+                bCe ^= st.ase();
+                bCi ^= st.asi();
                 de = bCa ^ RotateLeft(bCi, 1);
+                bCo ^= st.aso();
                 di = bCe ^ RotateLeft(bCo, 1);
-                @do = bCi ^ RotateLeft(bCu, 1);
+                bCu ^= st.asu();
```
LukaszRozmej, line 136: in many steps we have bCx and dX steps mixed, maybe it would be more readable if we unmix them. I think they are not dependent on each other.

src/Nethermind/Nethermind.Core/Crypto/KeccakHash.cs
```diff
@@ -104,130 +100,98 @@ private static void KeccakF(Span<ulong> st)
...
-                bCu = RotateLeft(asu, 14);
-                eba = bCa ^ ((~bCe) & bCi);
-                eba ^= RoundConstants[round];
+                bCa = da ^ st.aba();
+                bCe = RotateLeft(st.age() ^ de, 44);
+                bCi = RotateLeft(st.aki() ^ di, 43);
+                eba = (bCa ^ ((~bCe) & bCi)) ^ RoundConstants[round];
+                bCo = RotateLeft(st.amo() ^ @do, 21);
                 ebe = bCe ^ ((~bCi) & bCo);
+                bCu = RotateLeft(st.asu() ^ du, 14);
```
LukaszRozmej, line 149: same thing here

src/Nethermind/Nethermind.Core/Crypto/KeccakHash.cs
```diff
@@ -236,135 +200,80 @@ private static void KeccakF(Span<ulong> st)
...
+                st.ako() = emu ^ ((~esa) & ebe);
+                st.aku() = esa ^ ((~ebe) & egi);
+
+                ebu = RotateLeft(ebu ^ du, 27);
+                ega = RotateLeft(ega ^ da, 36);
+                eke = RotateLeft(eke ^ de, 10);
+                st.ama() = ebu ^ ((~ega) & eke);
+                emi = RotateLeft(emi ^ di, 15);
+                st.ame() = ega ^ ((~eke) & emi);
+                eso = RotateLeft(eso ^ @do, 56);
```
LukaszRozmej, line 252: similara eXX with aXX mixing

src/Nethermind/Nethermind.Core/Crypto/KeccakHash.cs
```diff
@@ -104,130 +100,98 @@ private static void KeccakF(Span<ulong> st)
...
-                bCa = RotateLeft(abe, 1);
-                agi ^= di;
-                bCe = RotateLeft(agi, 6);
-                ako ^= @do;
-                bCi = RotateLeft(ako, 25);
-                amu ^= du;
-                bCo = RotateLeft(amu, 8);
-                asa ^= da;
-                bCu = RotateLeft(asa, 18);
+                st.abe();
```
LukaszRozmej, line 165: this is not needed?

LukaszRozmej (commented):
Do you have any - even simple benchmark?

src/Nethermind/Nethermind.Core/Crypto/KeccakHash.cs
benaadams: Keen eye

benaadams:
> Do you have any - even simple benchmark?

I benchmarked it with Paprika's benchamarks and it was pretty much the same 😢

There are two many varibles and too few registers so it spills to stack quite a lot and is very very sensitive to instruction order and data dependencies formed

PR #7001, opened 2024-05-08
Use existing array for precompile result

benaadams:
## Changes

- Use existing array for precompile result, when Memory array backed and size same

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Evm/Precompiles/IdentityPrecompile.cs
+++ b/src/Nethermind/Nethermind.Evm/Precompiles/IdentityPrecompile.cs
@@ -29,7 +29,7 @@ namespace Nethermind.Evm.Precompiles
 
         public (ReadOnlyMemory<byte>, bool) Run(in ReadOnlyMemory<byte> inputData, IReleaseSpec releaseSpec)
         {
-            return (inputData, true);
+            return (inputData.ToArray(), true);
         }
     }
 }
--- a/src/Nethermind/Nethermind.Evm/VirtualMachine.cs
+++ b/src/Nethermind/Nethermind.Evm/VirtualMachine.cs
@@ -140,7 +140,7 @@ public class VirtualMachine : IVirtualMachine
         public static CallResult StackOverflowException => new(EvmExceptionType.StackOverflow); // TODO: use these to avoid CALL POP attacks
         public static CallResult StackUnderflowException => new(EvmExceptionType.StackUnderflow); // TODO: use these to avoid CALL POP attacks
         public static CallResult InvalidCodeException => new(EvmExceptionType.InvalidCode);
-        public static CallResult Empty => new(Array.Empty<byte>(), null);
+        public static CallResult Empty => new(default, null);
 
         public CallResult(EvmState stateToExecute)
         {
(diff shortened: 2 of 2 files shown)


src/Nethermind/Nethermind.Evm/Precompiles/IdentityPrecompile.cs
```diff
@@ -29,7 +29,7 @@ public long DataGasCost(in ReadOnlyMemory<byte> inputData, IReleaseSpec releaseS
 
         public (ReadOnlyMemory<byte>, bool) Run(in ReadOnlyMemory<byte> inputData, IReleaseSpec releaseSpec)
         {
-            return (inputData, true);
+            return (inputData.ToArray(), true);
```
benaadams, line 32: Not enturely sure why, but IdentityPrecompile isn't really used so shouldn't really matter

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -690,7 +691,13 @@ private CallResult ExecutePrecompile(EvmState state, IReleaseSpec spec)
...
         {
             (ReadOnlyMemory<byte> output, bool success) = precompile.Run(callData, spec);
-            CallResult callResult = new(output.ToArray(), success, !success);
+            byte[] result = (MemoryMarshal.TryGetArray(output, out ArraySegment<byte> outputArray) &&
+                outputArray.Offset == 0 && outputArray.Count == output.Length) ?
+                outputArray.Array :
+                // Wasn't array backed or Memory wasn't the whole array
+                output.ToArray();
+
+            CallResult callResult = new(result, success, !success);
```
LukaszRozmej, line 700: We can just change `CallResult` to accept `ReadOnlyMemory<byte>` as it is already like this in property:

```
public ReadOnlyMemory<byte> Output { get; }
```

LukaszRozmej (commented):
Also would it be better to be `bool TryRun(in ReadOnlyMemory<byte> inputData, IReleaseSpec releaseSpec, out ReadOnlyMemory<byte> result)` ?

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
benaadams: Changed

benaadams:
> Also would it be better to be `bool TryRun(in ReadOnlyMemory<byte> inputData, IReleaseSpec releaseSpec, out ReadOnlyMemory<byte> result)` ?

More of a break if another chain has their own precompiles? (Don't know if they do)

LukaszRozmej:
> > Also would it be better to be `bool TryRun(in ReadOnlyMemory<byte> inputData, IReleaseSpec releaseSpec, out ReadOnlyMemory<byte> result)` ?
> 
> More of a break if another chain has their own precompiles? (Don't know if they do)

huh why? I am askung about just changing interface, `bool` can go via register, or the tuple goes via registers?

LukaszRozmej (approved):

benaadams:
> > > Also would it be better to be `bool TryRun(in ReadOnlyMemory<byte> inputData, IReleaseSpec releaseSpec, out ReadOnlyMemory<byte> result)` ?
> > 
> > 
> > More of a break if another chain has their own precompiles? (Don't know if they do)
> 
> huh why? I am askung about just changing interface, `bool` can go via register, or the whole tuple goes via registers?

I mean if any existing extenal precompiles/plugin as they would need to change their signiture to upgrade?

benaadams:
Would be a better signiture though :)

LukaszRozmej:
> > > > Also would it be better to be `bool TryRun(in ReadOnlyMemory<byte> inputData, IReleaseSpec releaseSpec, out ReadOnlyMemory<byte> result)` ?
> > > 
> > > 
> > > More of a break if another chain has their own precompiles? (Don't know if they do)
> > 
> > 
> > huh why? I am askung about just changing interface, `bool` can go via register, or the whole tuple goes via registers?
> 
> I mean if any existing extenal precompiles/plugin as they would need to change their signiture to upgrade?

I don't think there are any, and if yes this is trivial change

PR #7008, opened 2024-05-10
Wait for background JumpDestinationAnalysis if in progress

benaadams:
## Changes

- If its already started processing it will likely finish before starting the processing from start will
- Also replaced queued/received section in logs which are always 0, with from cache + analysed which is more interesting data

image

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Processing/ProcessingStats.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/ProcessingStats.cs
@@ -39,6 +39,8 @@ namespace Nethermind.Consensus.Processing
         private long _processingMicroseconds;
         private long _lastTotalCreates;
         private long _lastReportMs;
+        private long _lastContractsAnalysed;
+        private long _lastCachedContractsUsed;
 
         public ProcessingStats(ILogger logger)
         {
@@ -51,7 +53,7 @@ namespace Nethermind.Consensus.Processing
 #endif
         }
 
-        public void UpdateStats(Block? block, IBlockTree blockTreeCtx, int recoveryQueueSize, int blockQueueSize, long blockProcessingTimeInMicros)
+        public void UpdateStats(Block? block, IBlockTree blockTreeCtx, long blockProcessingTimeInMicros)
         {
             const string resetColor = "\u001b[37m";
             const string whiteText = "\u001b[97m";
@@ -85,8 +87,6 @@ namespace Nethermind.Consensus.Processing
             Metrics.LastDifficulty = block.Difficulty;
             Metrics.GasUsed = block.GasUsed;
             Metrics.GasLimit = block.GasLimit;
-            Metrics.RecoveryQueueSize = recoveryQueueSize;
-            Metrics.ProcessingQueueSize = blockQueueSize;
 
             Metrics.BlockchainHeight = block.Header.Number;
             Metrics.BestKnownBlockNumber = blockTreeCtx.BestKnownNumber;
@@ -117,6 +117,8 @@ namespace Nethermind.Consensus.Processing
                     long chunkCreates = Evm.Metrics.Creates - _lastTotalCreates;
                     long chunkSload = Evm.Metrics.SloadOpcode - _lastTotalSLoad;
                     long chunkSstore = Evm.Metrics.SstoreOpcode - _lastTotalSStore;
+                    long contractsAnalysed = Evm.Metrics.ContractsAnalysed - _lastContractsAnalysed;
+                    long cachedContractsUsed = Db.Metrics.CodeDbCache - _lastCachedContractsUsed;
                     double chunkMGas = Metrics.Mgas - _lastTotalMGas;
                     double mgasPerSecond = chunkMicroseconds == 0 ? -1 : chunkMGas / chunkMicroseconds * 1_000_000.0;
                     double totalMgasPerSecond = totalMicroseconds == 0 ? -1 : Metrics.Mgas / totalMicroseconds * 1_000_000.0;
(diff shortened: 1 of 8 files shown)


src/Nethermind/Nethermind.Evm/CodeAnalysis/JumpDestinationAnalyzer.cs
```diff
@@ -19,17 +20,56 @@ public sealed class JumpDestinationAnalyzer(ReadOnlyMemory<byte> code)
...
+                Thread thread = Thread.CurrentThread;
+                ThreadPriority priority = thread.Priority;
+                // We are waiting, so drop priority to normal (BlockProcessing runs at higher priority).
+                thread.Priority = ThreadPriority.Normal;
+
+                // Already in progress, wait for completion.
+                resetEvent.Wait();
+
+                // Restore the priority of the thread.
+                thread.Priority = priority;
```
LukaszRozmej, line 67: should this be in finally?

src/Nethermind/Nethermind.Evm/CodeAnalysis/JumpDestinationAnalyzer.cs
```diff
@@ -164,8 +207,28 @@ private static void MarkJumpDestinations(long[] jumpDestinationBitmap, int pos,
...
+                    thread.Priority = ThreadPriority.AboveNormal;
+
+                    _jumpDestinationBitmap ??= CreateJumpDestinationBitmap();
+
+                    // Release the MRES to be GC'd
+                    _analysisComplete = _jumpDestinationBitmap;
+                    // Signal complete.
+                    analysisComplete.Set();
+                    // Restore the priority of the thread.
+                    thread.Priority = priority;
```
LukaszRozmej, line 227: should this be in finally?

LukaszRozmej (commented):
The queues in logs are very useful in full sync or catching up to head.

src/Nethermind/Nethermind.Evm/CodeAnalysis/JumpDestinationAnalyzer.cs
benaadams: This on definately; other not so important as threadpool resets it anyway; however better if we do as that's implementation detail

benaadams:
> The queues in logs are very useful in full sync or catching up to head.

Switch modes when the queues are empty?

image

benaadams:
Though not happy with `scontracts` as a term 😅

benaadams:
`exec code  from cache     982 | new     11` instead

image

src/Nethermind/Nethermind.Consensus/Processing/ProcessingStats.cs
```diff
@@ -200,8 +202,22 @@ public void UpdateStats(Block? block, IBlockTree blockTreeCtx, int recoveryQueue
...
+
+                    string extraStats;
+                    if (recoveryQueue > 0 || processingQueue > 0)
+                    {
+                        extraStats = $"recover {recoveryQueue,5:N0} | process {processingQueue,5:N0}";
+                    }
+                    else
+                    {
+                        extraStats = $"exec code {resetColor} from cache {cachedContractsUsed,7:N0} |{resetColor} new {contractsAnalysed,6:N0}";
+                    }
```
LukaszRozmej, line 217: Hm... previously we created one string in `UpdateStats`, now we are creating 3.
I would in-line this into `_logger.Info()` call.

src/Nethermind/Nethermind.Evm/CodeAnalysis/JumpDestinationAnalyzer.cs
```diff
@@ -19,17 +20,62 @@ public sealed class JumpDestinationAnalyzer(ReadOnlyMemory<byte> code)
...
+            if (previous is null)
+            {
+                ManualResetEventSlim analysisComplete = new(initialState: false);
+                previous = Interlocked.CompareExchange(ref _analysisComplete, analysisComplete, null);
+                if (previous is null)
+                {
+                    // Not already in progress, so start it.
+                    var bitmap = CreateJumpDestinationBitmap();
+                    _jumpDestinationBitmap = bitmap;
+                    // Release the MRES to be GC'd
```
LukaszRozmej, line 48: Would be good to Dispose MRES too.

src/Nethermind/Nethermind.Evm/CodeAnalysis/JumpDestinationAnalyzer.cs
```diff
@@ -19,17 +20,62 @@ public sealed class JumpDestinationAnalyzer(ReadOnlyMemory<byte> code)
...
+                    _analysisComplete = bitmap;
+                    // Signal complete.
+                    analysisComplete.Set();
+                    return bitmap;
+                }
```
LukaszRozmej, line 53: Dispose it too when set fails

src/Nethermind/Nethermind.Evm/CodeAnalysis/JumpDestinationAnalyzer.cs
```diff
@@ -164,8 +213,34 @@ private static void MarkJumpDestinations(long[] jumpDestinationBitmap, int pos,
...
+                    ThreadPriority priority = thread.Priority;
+                    try
+                    {
+                        // Boost the priority of the thread as block processing may be waiting on this.
+                        thread.Priority = ThreadPriority.AboveNormal;
+
+                        _jumpDestinationBitmap ??= CreateJumpDestinationBitmap();
+
+                        // Release the MRES to be GC'd
+                        _analysisComplete = _jumpDestinationBitmap;
```
LukaszRozmej, line 231: Dispose MRES?

src/Nethermind/Nethermind.Consensus/Processing/ProcessingStats.cs
LukaszRozmej: Should we update stats outside the main thread?

benaadams: Moved to background thread. Doesn't use StringBuilder? Is 3 string per second isn't much in the scheme of things?

src/Nethermind/Nethermind.Evm/CodeAnalysis/JumpDestinationAnalyzer.cs
benaadams: MRES doesn't have a finalizer and is only when you access the WaitHandle property (as we do with `WaitHandleExtensions`) that it creates a finalizable object that needs to be disposed. 

Regular MRE needs to be disposed; but MRES doesn't use any OS sync primitives unless you access the WaitHandle and cause it to allocate one.

src/Nethermind/Nethermind.Consensus/Processing/ProcessingStats.cs
benaadams: Generating on background thread looks a good call

LukaszRozmej: But each string interpolation creates a StringBuilder?

benaadams: Post C#10 .NET6.0 is complicated https://devblogs.microsoft.com/dotnet/string-interpolation-in-c-10-and-net-6/
image

And it caches via ArrayPool withthe main allocations happening post Log.Info in NLogger

image

benaadams:
Added odd/even for peers report as its frequent enough that can see both and too frequent to have both at same time (two lines)

image

LukaszRozmej (approved):

PR #7012, opened 2024-05-13
Refactor/separate block production and running

asdacap:
- Separate the building of block and the triggering of the building of the block.
- Make it easy to identify if a plugin is replacing only the building of the block or signalling of the block production, which does not change a lot actually.
- Remove the need of IManualBlockProductionTrigger and just call IBlockProducer directly, except of `evm_mine` which may want to trigger event which is used to put the block in main chain.

## Changes

- Created IBlockProducerRunner which have the interface of the current IBlockProducer.
- The interface of IBlockProducer is now the same as IManualBlockProductionTrigger.
- Move any use of IManualBlockProductionTrigger to just calling IBlockProducer directly.

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No


src/Nethermind/Nethermind.Consensus/IBlockProducer.cs
```diff
@@ -1,16 +1,19 @@
...
 
 public interface IBlockProducer
 {
-    Task Start();
-    Task StopAsync();
-    bool IsProducingBlocks(ulong? maxProducingInterval);
-    event EventHandler<BlockEventArgs> BlockProduced;
+    Task<Block?> BuildBlock(
+        BlockHeader? parentHeader = null,
+        CancellationToken? cancellationToken = null,
```
LukaszRozmej, line 16: minor: convention puts cancellationToken often as last parameter to separate it from more logic-oriented parameters.

src/Nethermind/Nethermind.Consensus/IBlockProducerRunner.cs
```diff
@@ -0,0 +1,16 @@
...
+using System;
+using System.Threading.Tasks;
+using Nethermind.Core;
+
+namespace Nethermind.Consensus;
+
+public interface IBlockProducerRunner
+{
+    Task Start();
+    Task StopAsync();
```
LukaszRozmej, line 13: We don't have truly async implementations of this?

src/Nethermind/Nethermind.Consensus/Producers/IBlockProducerInfo.cs
```diff
@@ -1,14 +1,16 @@
...
+using Nethermind.Core;
 using Nethermind.Evm.Tracing;
 
 namespace Nethermind.Consensus.Producers
 {
     public interface IBlockProducerInfo
     {
         IBlockProducer BlockProducer { get; }
-        IManualBlockProductionTrigger BlockProductionTrigger { get; }
+        Func<BlockHeader, bool> Condition { get; }
```
LukaszRozmej, line 13: Cryptic?

LukaszRozmej (commented):
Mixed feelings: It doesn't seem to make things simpler? Also am I right that we are loosing easily triggering block production from a simple plugin as we are not exposing trigger?

asdacap:
Yea, it does not look like it help much unfortunately. It does expose `IConsensusPlugin.CreateBlockProducerRunner` for the purpose of the plugin orchestrating when to produce block. IMO it make it easier for the plugin to arbitrarily create block without creating side effect by calling `IBlockProducer.BuildBlock`.

src/Nethermind/Nethermind.Consensus/IBlockProducerRunner.cs
asdacap: Only clique for stop

src/Nethermind/Nethermind.Consensus/Producers/IBlockProducerInfo.cs
asdacap: Switched to custom interface.

src/Nethermind/Nethermind.Consensus/IBlockProducerRunner.cs
```diff
@@ -9,7 +9,7 @@ namespace Nethermind.Consensus;
 
 public interface IBlockProducerRunner
 {
-    Task Start();
+    void Start();
```
LukaszRozmej, line 12: or we could make it a ValueTask if we wanted flexibility

LukaszRozmej (approved):

PR #7013, opened 2024-05-13
Fix commit in ClientVersionV1 to have only 4 bytes of info following the spec

smartprogrammer93:
Following spec here: https://github.com/ethereum/execution-apis/blob/main/src/engine/identification.md

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No


MarekM25 (approved):

MarekM25 (commented):
I don't  think we should add version and commit to client info

Issue #2324, opened 2020-09-21
Only recalculate storage roots on block commit


tkstanczak:
// TODO: it seems that we are unnecessarily recalculating root hashes all the time in storage?
            foreach (Address address in toUpdateRoots)
            {
                // since the accounts could be empty accounts that are removing (EIP-158)
                if (_stateProvider.AccountExists(address))
                {
                    Keccak root = RecalculateRootHash(address);
                    _stateProvider.UpdateStorageRoot(address, root);
                }
            }

PR #7021, opened 2024-05-14
Only recalculate storage roots once per block

benaadams:
Resolves https://github.com/NethermindEth/nethermind/issues/2324

## Changes

- Only recalculate root hashes for storages once per block than for each tx; if multiple tx use the same contract it is wasteful to recalcuate them multiple times.
- Also means can recalucate all the storage roots in parallel at the end.
- Addresses some long standing TODOs and issue https://github.com/NethermindEth/nethermind/issues/2324

Now 1.5% of block processing when running in Parallel

image

- [x] Optimization

#### Requires testing

- [x] No


benaadams:
Hive tests pass https://github.com/NethermindEth/nethermind/actions/runs/9148502578

LukaszRozmej (approved):

Issue #2157, opened 2020-07-20
Excessive peering


tkstanczak:
When a node has a big max active peers number then we connect to same peers over and over again.
Some smart delays should be introduced based on whether the nodes are on the same chain  or have same capabilities.

Generally we should try more often (every 5 minutes) nodes that respond with:
 * TooManyPeers

And less often to:
 * unreachable nodes (every 15 minutes?)

And even less often (once per hour or less frequently) to nodes that respond with
 * a wrong ChainId
 * no capabilities matched

Current behaviour leads to excessive memory allocations and CPU drain on handshake processing.

kamilchodola:
@benaadams any update on that?
@asdacap wasn't that improved at some point?

PR #7026, opened 2024-05-15
Improve connected peer distribution

benaadams:
Resolves https://github.com/NethermindEth/nethermind/issues/2157

## Changes

- Disfavour nodes with same ip addresses of nodes tried recently
- Reject peers with differing udp and tcp ports as we only store 1 port per peer so need to be the same

image

- [x] Optimization

#### Requires testing

- [x] Yes

- [x] Yes


src/Nethermind/Nethermind.Network.Discovery/Lifecycle/NodeLifecycleManager.cs
```diff
@@ -190,14 +190,24 @@ private void ProcessNodes(NeighborsMsg msg)
...
+                // ports differed between tcp and udp
+                return;
+            }
             if (node.Address.Address == _localhost)
             {
                 if (_logger.IsTrace)
                     _logger.Trace($"Received localhost as node address from: {msg.FarPublicKey}, node: {node}");
                 continue;
             }
+            else if (!NetworkStorage.NodesFilter.Set(node.Address.Address))
```
asdacap, line 206: I believe this should be better placed in `DiscoveryManager`.

src/Nethermind/Nethermind.Network/NetworkStorage.cs
```diff
@@ -17,10 +20,13 @@ namespace Nethermind.Network
 {
     public class NetworkStorage : INetworkStorage
     {
+        public static LruKeyCache<IPAddress> NodesFilter { get; } = new(2048, "Ip Filter");
```
asdacap, line 23: Please refrain from static. It breaks encapsulation.

flcl42:
Might be better to store both ports. There is no agreement about having same ports; same ports may be easier to detect by ISP/bad actors

src/Nethermind/Nethermind.Network.Discovery/Lifecycle/NodeLifecycleManager.cs
benaadams: Moved to DiscoveryManager

src/Nethermind/Nethermind.Network/NetworkStorage.cs
benaadams: Readonly instance

benaadams:
> Might be better to store both ports. There is no agreement about having same ports; same ports may be easier to detect by ISP/bad actors

Captured in issue https://github.com/NethermindEth/nethermind/issues/7247

benaadams:
image

src/Nethermind/Nethermind.Network.Discovery.Test/DiscoveryMessageSerializerTests.cs
```diff
@@ -192,10 +192,20 @@ public void NeighborsMessageTest()
...
+                    Assert.That(messageNode, Is.Null);
+                    continue;
+                }
+                Assert.That(messageNode, Is.Not.Null);
+
+
+                Assert.That(deserializedNode.Host, Is.EqualTo(message.Nodes[i]!.Host));
+                Assert.That(deserializedNode.Port, Is.EqualTo(message.Nodes[i]!.Port));
+                Assert.That(deserializedNode.IdHash, Is.EqualTo(message.Nodes[i]!.IdHash));
+                Assert.That(deserializedNode, Is.EqualTo(message.Nodes[i]));
```
LukaszRozmej, line 208: Can be one assert (with fluent asserts)

src/Nethermind/Nethermind.Network.Discovery.Test/EIP8DiscoveryTests.cs
```diff
@@ -87,10 +87,10 @@ public void NeighborsFormatTest()
...
             Assert.That(msg.ExpirationTime, Is.EqualTo(1136239445));
-            Assert.That(msg.Nodes[0].IdHash.ToString(true), Is.EqualTo("0xca25217b2fbd0ae6d435be1f0f99282a930dac3ca4c3358900da74e7caa7eee4"));
-            Assert.That(msg.Nodes[1].IdHash.ToString(true), Is.EqualTo("0x6b1e430439aaf971bf408b99c0097bc560238d10d5642e44cdd89d8192bb0554"));
-            Assert.That(msg.Nodes[2].IdHash.ToString(true), Is.EqualTo("0x7d38ab8183dce7e53e5d5536956a8de7ab45aaeb12948848a30efae067f2b2e9"));
-            Assert.That(msg.Nodes[3].IdHash.ToString(true), Is.EqualTo("0xaaa7e418c80aa0a3e9a29af1ecf831fdcc1093510d0c73febaea674e69e7e58b"));
+            Assert.That(msg.Nodes[0]?.IdHash?.ToString(true), Is.Null);
+            Assert.That(msg.Nodes[1]?.IdHash?.ToString(true), Is.EqualTo("0x6b1e430439aaf971bf408b99c0097bc560238d10d5642e44cdd89d8192bb0554"));
+            Assert.That(msg.Nodes[2]?.IdHash?.ToString(true), Is.EqualTo("0x7d38ab8183dce7e53e5d5536956a8de7ab45aaeb12948848a30efae067f2b2e9"));
+            Assert.That(msg.Nodes[3]?.IdHash?.ToString(true), Is.Null);
```
LukaszRozmej, line 93: Can be one assert

LukaszRozmej (changes requested):
I think we shouldn't disallow different tcp and udp ports. We need to support this scenario.

src/Nethermind/Nethermind.Network.Discovery.Test/DiscoveryMessageSerializerTests.cs
benaadams: That make it harder to work out what went wrong in error message? (e.g. line number)

LukaszRozmej: Not really as fluent asserts have way better messages

benaadams:
Closing in favour of https://github.com/NethermindEth/nethermind/pull/8426

PR #7039, opened 2024-05-17
Intra-block cache

### LINKED ISSUE #179: Keep caches between transactions
currently we clear caches on commit and commit happens after every tx
very inefficient

benaadams:
Resolves #179

I'd solved this a long time ago; however always had issues doing the more ambitious step of sharing the cache between blocks; and parallel pre-caching.

Recognising the perfect can be the enemy of the good; and that we can make a journey with many steps; this introduces just the part of sharing state and storage between tx in the same block; rather than always starting afresh for each tx.

## Changes

- Maintain state and storage cache between the txs in a block

Even though it clears every block in this iteration

*  \> 50% of state reads are cache hits
*  ~ 21% of storage reads are cache hits

image

- [x] Optimization

#### Requires testing

- [x] No


benaadams:
Hive test pass https://github.com/NethermindEth/nethermind/actions/runs/9122197603

LukaszRozmej (approved):

PR #7046, opened 2024-05-18
Txpool state cache

benaadams:
## Changes

- Adding a state cache to txPool and building on https://github.com/NethermindEth/nethermind/pull/7039 to push the changed accounts to the pool with the block so those accounts can be invalidated from the cache. (As TxPool was causing a signifcant dominance in the remaining state access)
   - When I updated the accounts rather than just invalidating them in cache that had a higher read rate; perhaps due to bringing them to the top of LRU; when they wouldn't be used in a following block, and pushing out older accounts that would. (e.g. if not included in block, are more likely to be still in tx pool in next block vs an account that was just included in a block vacating the tx pool)

- [x] Optimization

#### Requires testing

- [x] Yes

- [x] Yes (ish) adjusted existing test to be cache aware and tested results against mainnet

#### StateDb Reads

Before ave 51.7k ops/m, max 140k ops/m

image

After ave 13.5k ops/m, max 35.5k ops/m

image

#### State Tree

Before (validation) reads 4.82k op/m; (txpool) reader reads 10.9k/s

image

After (validation) reads 2.2k op/m; (txpool) reader reads 1.1k/s;  cache hits 11.9k op/m

image

#### Trie Operations

Before RLP.Encode 103k ops/m

image

After RLP.Encode 20.5k ops/m

image

#### DarkPool uneffected (i.e. not over dropping txs)

Before ave 18%

image

After ave 16%

image

### DIFF
--- a/src/Nethermind/Nethermind.State/StateProvider.cs
+++ b/src/Nethermind/Nethermind.State/StateProvider.cs
@@ -21,14 +21,15 @@ namespace Nethermind.State
     internal class StateProvider
     {
         private const int StartCapacity = Resettable.StartCapacity;
-        private readonly ResettableDictionary<AddressAsKey, Stack<int>> _intraBlockCache = new();
+        private readonly ResettableDictionary<AddressAsKey, Stack<int>> _intraTxCache = new();
         private readonly ResettableHashSet<AddressAsKey> _committedThisRound = new();
         private readonly HashSet<AddressAsKey> _nullAccountReads = new();
         // Only guarding against hot duplicates so filter doesn't need to be too big
         // Note:
         // False negatives are fine as they will just result in a overwrite set
         // False positives would be problematic as the code _must_ be persisted
-        private readonly LruKeyCache<Hash256AsKey> _codeInsertFilter = new(2048, "Code Insert Filter");
+        private readonly LruKeyCacheNonConcurrent<Hash256AsKey> _codeInsertFilter = new(2048, "Code Insert Filter");
+        private readonly LruCacheNonConcurrent<AddressAsKey, Account> _blockCache = new(8192, "Account Cache");
 
         private readonly List<Change> _keptInCache = new();
         private readonly ILogger _logger;
@@ -90,7 +91,7 @@ namespace Nethermind.State
 
         public bool AccountExists(Address address)
         {
-            if (_intraBlockCache.TryGetValue(address, out Stack<int> value))
+            if (_intraTxCache.TryGetValue(address, out Stack<int> value))
             {
                 return _changes[value.Peek()]!.ChangeType != ChangeType.Delete;
             }
@@ -361,7 +362,7 @@ namespace Nethermind.State
             for (int i = 0; i < _currentPosition - snapshot; i++)
             {
                 Change change = _changes[_currentPosition - i];
-                Stack<int> stack = _intraBlockCache[change!.Address];
+                Stack<int> stack = _intraTxCache[change!.Address];
                 if (stack.Count == 1)
                 {
                     if (change.ChangeType == ChangeType.JustCache)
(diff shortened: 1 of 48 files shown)


src/Nethermind/Nethermind.State/StateProvider.cs
```diff
@@ -798,6 +799,11 @@ public Change(ChangeType type, Address address, Account? account)
             public Account? Account { get; }
         }
 
+        public AddressAsKey[]? ChangedAddresses()
+        {
+            return _blockCache.Count == 0 ? Array.Empty<AddressAsKey>() : _blockCache.Keys.ToArray();
```
LukaszRozmej, line 804: Does it makes sense to allocate it as array?

src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -206,6 +210,20 @@ private void ProcessNewHeads()
...
+                                // Not sequential block, reset cache
+                                _accountCache.Reset();
+                            }
+                            else
+                            {
+                                // Sequential block, just remove changed accounts from cache
+                                _accountCache.RemoveAccounts(accountChanges);
+                            }
+                            _lastBlockNumber = args.Block.Number;
+                            _lastBlockHash = args.Block.Hash;
```
LukaszRozmej, line 225: IMO: we should extract also changes for those accounts, not only the accounts themselves.

src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -206,6 +210,20 @@ private void ProcessNewHeads()
                     {
                         try
                         {
+                            AddressAsKey[]? accountChanges = args.Block.AccountChanges;
+                            if (accountChanges is null || accountChanges.Length == 0 || args.Block.ParentHash != _lastBlockHash || _lastBlockNumber + 1 != args.Block.Number)
```
LukaszRozmej, line 214: Ca we hide it under some `CanUseCache` method?

src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -737,6 +755,72 @@ private void TimerOnElapsed(object? sender, EventArgs e)
...
 
+        internal void ResetAddress(Address address)
+        {
+            _accountCache.RemoveAccounts([address]);
+        }
+
+        private sealed class AccountCache : IAccountStateProvider
+        {
+            private readonly IAccountStateProvider _provider;
+            private readonly LruCache<AddressAsKey, AccountStruct>[] _caches;
```
LukaszRozmej, line 766: why are there multiple cachess here?

benaadams: To reduce lock contention so it choose which bucket to go in (and therefore lock to use) based on last nibble of address so 16 caches

src/Nethermind/Nethermind.State/StateProvider.cs
benaadams: What would you use?

Detaches it from block processing; isn't a concurrent Dictionary so can't be accessed safely and the processing it just about to clear it for the next block.

Could pass along the full dictionary and allocate a new one; but using the updated data caused more cache misses. Would probably need to add `UpdateValueButNotRecency` method for the Lru; as if an account was used in last block, most of them aren't used in the very next block; so want older address that weren't included and still in the txpool to survive.

src/Nethermind/Nethermind.TxPool/TxPool.cs
benaadams: And apply the changes to cache? See above

src/Nethermind/Nethermind.State/StateProvider.cs
LukaszRozmej: If there is a good time to release it then it can be ArrayPoolList

LukaszRozmej (approved):

src/Nethermind/Nethermind.State/StateProvider.cs
benaadams: Ohhh, good point

PR #7048, opened 2024-05-19
Parallel Trie Branch KeyGeneration

benaadams:
## Changes

- For the root of the Trie calculate all the branches keys in parallel (which occurs in the GetLength path)

Before 7+% of block processing
image

After 1.5% of block processing

image

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
+++ b/src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
@@ -422,8 +422,8 @@ namespace Nethermind.Serialization.Rlp
                 ThrowArgumentOutOfRangeException();
             }
 
-            Unsafe.Add(ref MemoryMarshal.GetReference(buffer), position) = 160;
-            Unsafe.As<byte, ValueHash256>(ref Unsafe.Add(ref MemoryMarshal.GetReference(buffer), position + 1)) = hash.ValueHash256;
+            Unsafe.Add(ref MemoryMarshal.GetReference(buffer), (nuint)position) = 160;
+            Unsafe.As<byte, ValueHash256>(ref Unsafe.Add(ref MemoryMarshal.GetReference(buffer), (nuint)position + 1)) = hash.ValueHash256;
             return newPosition;
 
             [DoesNotReturn]
--- a/src/Nethermind/Nethermind.Trie/TrieNode.Decoder.cs
+++ b/src/Nethermind/Nethermind.Trie/TrieNode.Decoder.cs
@@ -238,7 +238,7 @@ namespace Nethermind.Trie
                     }
                 }
                 return totalLength;
-            }            
+            }
 
             private static int GetChildrenRlpLengthForBranchRlpParallel(ITrieNodeResolver tree, TreePath rootPath, TrieNode item, ICappedArrayPool? bufferPool)
             {
@@ -277,7 +277,7 @@ namespace Nethermind.Trie
                             local += childNode.Keccak is null ? childNode.FullRlp.Length : Rlp.LengthOfKeccakRlp;
                         }
                     }
-                    
+
                     return local;
                 },
                 local =>
--- a/src/Nethermind/Nethermind.Trie/TrieNode.Decoder.cs
+++ b/src/Nethermind/Nethermind.Trie/TrieNode.Decoder.cs
@@ -155,62 +155,52 @@ namespace Nethermind.Trie
             {
                 item.EnsureInitialized();
                 // Tail call optimized.
-                if (item.HasRlp)
-                {
-                    return GetChildrenRlpLengthForBranchRlp(tree, ref path, item, bufferPool);
-                }
-                else
-                {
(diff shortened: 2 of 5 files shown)


benaadams:
Hive test pass https://github.com/NethermindEth/nethermind/actions/runs/9149515737

src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
```diff
@@ -422,8 +422,8 @@ public static int Encode(Span<byte> buffer, int position, Hash256 hash)
                 ThrowArgumentOutOfRangeException();
             }
 
-            Unsafe.Add(ref MemoryMarshal.GetReference(buffer), position) = 160;
-            Unsafe.As<byte, ValueHash256>(ref Unsafe.Add(ref MemoryMarshal.GetReference(buffer), position + 1)) = hash.ValueHash256;
+            Unsafe.Add(ref MemoryMarshal.GetReference(buffer), (nuint)position) = 160;
+            Unsafe.As<byte, ValueHash256>(ref Unsafe.Add(ref MemoryMarshal.GetReference(buffer), (nuint)position + 1)) = hash.ValueHash256;
```
LukaszRozmej, line 426: why `nuint`? We only support 64bit systems already.

src/Nethermind/Nethermind.Trie/TrieNode.Decoder.cs
```diff
@@ -163,6 +165,55 @@ private static int GetChildrenRlpLengthForBranch(ITrieNodeResolver tree, ref Tre
...
+                item.EnsureInitialized();
+                // Tail call optimized.
+                if (item.HasRlp)
+                {
+                    return GetChildrenRlpLengthForBranchRlpParallel(tree, path, item, bufferPool);
+                }
+                else
+                {
+                    return GetChildrenRlpLengthForBranchNonRlpParallel(tree, path, item, bufferPool);
+                }
```
LukaszRozmej, line 179: `? :` 🙏

src/Nethermind/Nethermind.Trie/TrieNode.Decoder.cs
```diff
@@ -163,6 +165,55 @@ private static int GetChildrenRlpLengthForBranch(ITrieNodeResolver tree, ref Tre
...
+            }
+
+            private static int GetChildrenRlpLengthForBranchNonRlpParallel(ITrieNodeResolver tree, TreePath rootPath, TrieNode item, ICappedArrayPool bufferPool)
+            {
+                int totalLength = 0;
+                Parallel.For(0, BranchesCount,
+                () => 0,
+                (i, state, local) =>
```
LukaszRozmej, line 187: lacks indentation? A bit harder to read.

src/Nethermind/Nethermind.Trie/TrieNode.Decoder.cs
```diff
@@ -163,6 +165,55 @@ private static int GetChildrenRlpLengthForBranch(ITrieNodeResolver tree, ref Tre
...
+                {
+                    object data = item._data[i];
```
LukaszRozmej, line 189: `item._data?[i]` for nullability?

src/Nethermind/Nethermind.Trie/TrieNode.Decoder.cs
LukaszRozmej, line 187: `state` not used, can be `_`

src/Nethermind/Nethermind.Trie/TrieNode.Decoder.cs
```diff
@@ -189,6 +240,53 @@ private static int GetChildrenRlpLengthForBranchNonRlp(ITrieNodeResolver tree, r
...
+                    if (data is null)
+                    {
+                        int length = rlpStream.PeekNextRlpLength();
+                        local += length;
+                        rlpStream.SkipBytes(length);
+                    }
+                    else
+                    {
+                        if (ReferenceEquals(data, _nullNode) || data is null)
```
LukaszRozmej, line 261: data cannot be null here

LukaszRozmej (commented):
Why this can only be supported for the root key?

src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
benaadams: Intent; the cast is to avoid some additional movsx that occur with `int`; could cast to `ulong` but the reason it to be unsigned and machine word size so `nuint` makes more sense?

benaadams:
> Why this can only be supported for the root key?

🤔 idea was only single level of Parallism, not sure how Parallel.For performs if recursed at each level. Will test

Scooletz:
> 🤔 idea was only single level of Parallism, not sure how Parallel.For performs if recursed at each level. Will test

Doesn't it fall into poor parallelism category? Also: how many cores one can expect?

benaadams:
95.1% is in ProcessTransaction; not sure it much gained by further parallising it?

image

LukaszRozmej (approved):

PR #7050, opened 2024-05-20
Feature DB File warmer

asdacap:
- As the state is being read, the database will be cached in the OS cache, therefore the block processing improve over time. It takes about 30 minutes of constant block processing (assuming catching up) or about 2 weeks in real time for this effect to continue making consistent testing hard to achieve. Plus if I just want it to run as fast as it could to test for memory issue for example, it take some time.
- This PR introduce a flag that will read the files from DB on start so that the OS cache is filled up early on.
- On a system with 192GB of RAM (max RAM possible on a consumer level hardware at the moment) this can cache the whole state db resulting in a close to 0 iops at ssd level.
- On a system with lower available memory than the size of the db, it will attempt to priority lower level and newer file first, skipping other file entirely. I did not test the effect on such sysstem, but I don't have much hope due to the random nature of the workload.
- Graph is (before, after, before, after). `sudo sysctl -w vm.drop_caches=3` was run between runs.
Screenshot from 2024-05-20 13-42-48

- [X] Optimization

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Tested with forward syncing node from an old backup.

### DIFF
--- a/src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
+++ b/src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
@@ -7,6 +7,7 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.IO;
 using System.IO.Abstractions;
+using System.Linq;
 using System.Reflection;
 using System.Threading;
 using ConcurrentCollections;
@@ -183,6 +184,11 @@ public class DbOnTheRocks : IDb, ITunableDb
                 }
             }
 
+            if (_perTableDbConfig.EnableFileWarmer)
+            {
+                WarmupFile(_fullPath, db);
+            }
+
             return db;
         }
         catch (DllNotFoundException e) when (e.Message.Contains("libdl"))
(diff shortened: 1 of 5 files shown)


src/Nethermind/Nethermind.Db.Test/DbOnTheRocksTests.cs
```diff
@@ -129,6 +129,22 @@ public void Dispose_wont_cause_ObjectDisposedException_when_batch_is_still_open(
...
+            config.EnableFileWarmer = true;
+            DbOnTheRocks db = new("testFileWarmer", GetRocksDbSettings("testFileWarmer", "FileWarmerTest"), config, LimboLogs.Instance);
+            for (int i = 0; i < 1000; i++)
+            {
+                db[i.ToBigEndianByteArray()] = i.ToBigEndianByteArray();
+            }
+            db.Dispose();
+
+            db = new("testFileWarmer", GetRocksDbSettings("testFileWarmer", "FileWarmerTest"), config, LimboLogs.Instance);
+            db.Dispose();
```
LukaszRozmej, line 145: prefer `using`?

src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
```diff
@@ -202,6 +210,90 @@ private RocksDb Open(string path, (DbOptions Options, ColumnFamilies? Families)
...
+                    }
+                    catch (IOException e)
+                    {
+                        // Something unusual, but nothing noteworthy.
+                        _logger.Warn($"Exception warming up {fullPath} {e}");
+                    }
+                }
+            });
+        }).ToArray();
+        Task.WaitAll(tasks);
```
LukaszRozmej, line 294: Wouldn't Parallel.For be better?

src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
```diff
@@ -202,6 +210,90 @@ private RocksDb Open(string path, (DbOptions Options, ColumnFamilies? Families)
...
+            if (take)
+            {
+                totalSize += (long)metadata.metadata.FileSize;
+            }
+            return take;
+        })
+            // We reverse them again so that lower level goes last so that it is the freshest.
+            // Not all of the available memory is actually available so we are probably over reading things.
+            .Reverse()
+            .ToList();
```
LukaszRozmej, line 257: minor: as fileMetadatas are a List to begin with, instead of TakeWhile, you could just reduce its Count with: https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.collectionsmarshal.setcount?view=net-8.0#system-runtime-interopservices-collectionsmarshal-setcount-1(system-collections-generic-list((-0))-system-int32)

and just reverse, without making another list.

src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
```diff
@@ -183,6 +186,11 @@ private RocksDb Open(string path, (DbOptions Options, ColumnFamilies? Families)
                 }
             }
 
+            if (_perTableDbConfig.EnableFileWarmer)
```
LukaszRozmej, line 189: Minor - Can we remove this general `EnableFileWarmer` flag and already use the one related to actual db? It would be better for user to set just 1 flag instead of 2.

LukaszRozmej (commented):
Interesting. @benaadams something for you to check out too!

benaadams:
> Interesting. @benaadams something for you to check out too!

I can confirm that rebooting tanks performance 😉

src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
asdacap: The general one is to set the default for other db. Only the state flag need to be specified.

benaadams:
> > Interesting. @benaadams something for you to check out too!
> 
> I can confirm that rebooting tanks performance 😉
> 
> Then needs a while to brew


Can confirm this is straight back into good perf after a reboot

image

src/Nethermind/Nethermind.Db.Test/DbOnTheRocksTests.cs
asdacap: Done

src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
asdacap: No idea. I've switched to `Parallel.ForEach`.

LukaszRozmej (approved):

benaadams:
Perhaps default this to `true` if you have enough memory for the given statedb directory?

asdacap:
Not sure on that. I'm still erring towards `false`. Its not known if the user is on VPS for example, or there are other software running.

PR #7052, opened 2024-05-20
Support for eth/69

### LINKED ISSUE #6926: Analyze and Implement eth/69 Drop pre-merge fields
https://github.com/ethereum/EIPs/pull/8271

alexb5dh:
Resolves #6926

## Changes

- Implements [eth/69](https://eips.ethereum.org/EIPS/eip-7642)
- Makes peer TD nullable and updates related strategies

- [x] Optimization (bandwidth used, see https://github.com/ethereum/EIPs/pull/8271#issuecomment-1983279804)

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

Unit tests are present, but additional integration testing is needed.


src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/V69/Eth69ProtocolHandler.cs
```diff
@@ -0,0 +1,110 @@
...
+                break;
+            default:
+                base.HandleMessage(message);
+                break;
+        }
+    }
+
+    private void Handle(StatusMessage69 status)
+    {
+        status.TotalDifficulty = 0; // TODO handle properly for eth/69
```
alexb5dh, line 80: This value is later assigned to current `ISyncPeer` `TotalDifficulty`, which has a lot of usages. While it looks like none of them should actually apply in the post-merge chains with code branching elsewhere, to fully confirm, one would need to do additional manual/integration testing, for example do a full sync from one node to another using the new protocol version.

Another potential problem, is someone using eth/69 with a non-PoS chain, for which `TotalDifficulty` is actually being needed for syncing.

src/Nethermind/Nethermind.Network.Test/P2P/Subprotocols/Eth/V69/Eth69ProtocolHandlerTests.cs
```diff
@@ -0,0 +1,230 @@
...
+    private IPooledTxsRequestor _pooledTxsRequestor = null!;
+    private IGossipPolicy _gossipPolicy = null!;
+    private ISpecProvider _specProvider = null!;
+    private Block _genesisBlock = null!;
+    private Eth68ProtocolHandler _handler = null!;
+    private ITxGossipPolicy _txGossipPolicy = null!;
+    private ITimerFactory _timerFactory = null!;
+
+    [SetUp]
+    public void Setup()
```
LukaszRozmej, line 60: The only downside here is that tests aren't parallelizable. But if they are fast it is fine.

LukaszRozmej (commented):
The one thing I don't like the is extending RlpBehaviors, rest is mostly minor stuff.

src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/V69/Eth69ProtocolHandler.cs
alexb5dh: @LukaszRozmej where do you prefer to use `TerminalTotalDifficulty`? `Status` message of `ProtocolHandlerBase` don't have such property.

This assignment is here just because handler doesn't support nullable `TotalDifficulty` and changing it will affect a lot of files.

src/Nethermind/Nethermind.Network.Test/P2P/Subprotocols/Eth/V69/StatusMessageSerializerTests.cs
```diff
@@ -0,0 +1,67 @@
...
+            "f8444501a0c89efdaa54c0f20c7adf612882df0950f5a951637e0307cdcb4c672f298b8bc6a0044852b2a670ade5407e78fb2863c51de9fcb96542a07186fe3aeda6bb8a116d"
+        },
+        new object[]
+        {
+            new StatusMessage69(),
+            "c480808080"
+        },
+    ];
+
+    [Theory]
```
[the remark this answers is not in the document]
alexb5dh: I thought `Test` in NUnit is an alternative to the `Fact` from xUnit, where `Theory` can represent only parameterized tests. But I haven't touched NUnit for years, so not sure about its conventions.

src/Nethermind/Nethermind.Network.Test/P2P/Subprotocols/Eth/V69/Eth69ProtocolHandlerTests.cs
alexb5dh: Take around 500ms locally, but that depends on the machine of course.
All handler tests are using `Setup`s like this, so I don't think this one is worth rethinking.

src/Nethermind/Nethermind.Serialization.Rlp/RlpBehaviors.cs
```diff
@@ -31,5 +31,6 @@ public enum RlpBehaviors
     /// </summary>
     InMempoolForm = 64,
     ExcludeHashes = 128,
-    All = AllowExtraBytes | ForSealing | Storage | Eip658Receipts | AllowUnsigned | SkipTypedWrapping | InMempoolForm | ExcludeHashes,
+    Eip7642Messages = 256,
```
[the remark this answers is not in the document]
alexb5dh: Will do without adding new `RlpBehaviors`, but this will need new decoder for `TxReceipt`, since bloom (de-)serialization happens there, not in `ReceiptsMessageSerializer`.

  Just a note, as I don't see any other protocol implementations adding new `IRlpDecoder`s.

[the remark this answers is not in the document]
alexb5dh: Done in https://github.com/NethermindEth/nethermind/pull/7052/commits/6f3240186249520f2677411733d00664bd21928e

src/Nethermind/Nethermind.Network.Test/P2P/Subprotocols/Eth/V69/StatusMessageSerializerTests.cs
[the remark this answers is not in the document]
LukaszRozmej: you can have `TestCases` and `Values` to get parametrized tests too.

src/Nethermind/Nethermind.Network.Test/P2P/Subprotocols/Eth/V69/Eth69ProtocolHandlerTests.cs
LukaszRozmej: would be better to re-do it without using class fields, then could be run in parallel, but fine

src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/V62/Messages/StatusMessageSerializer.cs
```diff
@@ -10,6 +10,15 @@ public class StatusMessageSerializer : IZeroInnerMessageSerializer<StatusMessage
     {
         private const int ForkHashLength = 5;
 
+        private readonly bool _includeTd;
+
+        public StatusMessageSerializer() : this(true) { }
+
+        protected StatusMessageSerializer(bool includeTd)
```
LukaszRozmej, line 17: can we not use abbreviations too much, `includeTotalDifficulty` is fine

LukaszRozmej (approved):

LukaszRozmej:
should we merge it?

alexb5dh:
> should we merge it?

We still don't have tests for this from Ethereum team. I would prefer not to merge it before checking them.
@MarekM25?

src/Nethermind/Nethermind.Synchronization/SnapSync/SnapSyncAllocationStrategyFactory.cs
```diff
@@ -12,7 +12,8 @@ public class SnapSyncAllocationStrategyFactory : StaticPeerAllocationStrategyFac
     {
 
         private static readonly IPeerAllocationStrategy DefaultStrategy =
-            new SatelliteProtocolPeerAllocationStrategy<ISnapSyncPeer>(new TotalDiffStrategy(new BySpeedStrategy(TransferSpeedType.SnapRanges, true), TotalDiffStrategy.TotalDiffSelectionType.CanBeSlightlyWorse), "snap");
+            // TODO: use TotalDiffStrategy in non-merge chains?
```
asdacap, line 15: Does not matter for snap sync. In fact, why not just remove the `TotalDiffStrategy` wrapper. It just cause mantainance issue like what you are having now.

alexb5dh: Wouldn't we need it for any PoW chains we support?

asdacap: Not really, it will just sleep the peer if it fail to return correct snap.

src/Nethermind/Nethermind.Synchronization/Peers/SyncPeerPool.cs
```diff
@@ -408,7 +408,8 @@ private async Task RunRefreshPeerLoop()
                     {
                         UpgradeAllocations();
                         // cases when we want other nodes to resolve the impasse (check Goerli discussion on 5 out of 9 validators)
-                        if (syncPeer.TotalDifficulty == _blockTree.BestSuggestedHeader?.TotalDifficulty && syncPeer.HeadHash != _blockTree.BestSuggestedHeader?.Hash)
+                        if (syncPeer.TotalDifficulty is { } syncPeerTD && syncPeerTD == _blockTree.BestSuggestedHeader?.TotalDifficulty &&
```
alexb5dh, line 411: Peers without TD won't participate in this. Is this ok?

src/Nethermind/Nethermind.Synchronization/ParallelSync/MultiSyncModeSelector.cs
```diff
@@ -147,15 +148,20 @@ public void Update()
                 bool inBeaconControl = _beaconSyncStrategy.ShouldBeInBeaconModeControl();
                 (UInt256? peerDifficulty, long? peerBlock) = ReloadDataFromPeers();
                 // if there are no peers that we could use then we cannot sync
-                if (peerDifficulty is null || peerBlock is null || peerBlock == 0)
+                if (peerBlock is null or 0)
```
alexb5dh, line 151: Changes in this file basically allow `Snapshot` to have peers with nullable `TD`.

smartprogrammer93 (approved):
LGTM, still no merge until we have some pytests for this. lets check when those will be available

src/Nethermind/Nethermind.Synchronization/Peers/AllocationStrategies/TransitioningPeerAllocationStrategy.cs
```diff
@@ -0,0 +1,40 @@
...
+
+    public TransitioningPeerAllocationStrategy(IPoSSwitcher poSSwitcher, IPeerAllocationStrategy innerStrategy)
+    {
+        if (poSSwitcher.TransitionFinished)
+        {
+            _currentStrategy = CreatePostMergeStrategy(innerStrategy);
+        }
+        else
+        {
+            poSSwitcher.Transitioned += (_, _) => _currentStrategy = CreatePostMergeStrategy(innerStrategy);
```
alexb5dh, line 27: We can save a few bytes of statically used memory by unsubscribing everything after `Transitioned` is fired, as it's expected to trigger only once. Should we?

benaadams: Can set to `null` if nothing else should be subscribed (removes all events)

src/Nethermind/Nethermind.Merge.Plugin/Synchronization/PosForwardHeaderProvider.cs
```diff
@@ -67,8 +67,9 @@ private void TryUpdateTerminalBlock(BlockHeader currentHeader)
     // Used only in get block header in pre merge forward header provider, this hook stops pre merge forward header provider.
     protected override bool ImprovementRequirementSatisfied(PeerInfo? bestPeer)
```
asdacap, line 68: You should not need to care about this class. Does this cause a problem somehow?

asdacap: Nevermind, I misread.

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
```diff
@@ -85,7 +89,7 @@ private static bool HasStateForBlock(IBlockchainBridge blockchainBridge, BlockHe
 
     public ResultWrapper<string> eth_protocolVersion()
     {
-        int highestVersion = P2PProtocolInfoProvider.GetHighestVersionOfEthProtocol();
+        int highestVersion = _protocolsManager.GetHighestProtocolVersion(Protocol.Eth);
```
asdacap, line 92: You know, you can extract this method out into its own class, let say `IEthP2PRpcModule`, register it as another module and skip a whole lot of change.

src/Nethermind/Nethermind.Consensus/AlwaysPoS.cs
```diff
@@ -23,9 +23,8 @@ private AlwaysPoS() { }
 
     public long? ConfiguredTerminalBlockNumber => 0;
 
-#pragma warning disable CS0067
-    public event EventHandler TerminalBlockReached;
-#pragma warning restore CS0067
+    public event EventHandler TerminalBlockReached { add { } remove { } }
+    public event EventHandler Transitioned { add { } remove { } }
```
asdacap, line 27: Is this required? Why not just `TerminalBlockReached`?

src/Nethermind/Nethermind.Synchronization/Peers/AllocationStrategies/TransitioningPeerAllocationStrategy.cs
```diff
@@ -0,0 +1,40 @@
...
+using Nethermind.Consensus;
+using Nethermind.Stats;
+
+namespace Nethermind.Synchronization.Peers.AllocationStrategies;
+
+/// <inheritdoc />
+/// <summary>
+/// Uses <see cref="TotalDiffStrategy"/> before the merge and <see cref="LastBlockStrategy"/> after.
+/// </summary>
+public class TransitioningPeerAllocationStrategy : IPeerAllocationStrategy
```
asdacap, line 15: Just skip this altogether. Don't add more code when you can remove code.

asdacap (changes requested):

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
alexb5dh: Which method exactly and how would that help?

src/Nethermind/Nethermind.Consensus/AlwaysPoS.cs
alexb5dh: Wanted to have transition finalized to minimize chance of reorg before enabling eth/69.

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
asdacap: Just `eth_protocolVersion` since that is the only one that use `_protocolsManager`. Eg: Make an `IEthCommonRpcModule`, put `eth_protocolVersion` into it, moving it out of `IEthRpcModule`. If no other plugin is using it, then you don't need to change other plugin.

benaadams:
Hive test should be available now

src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/V62/Eth62ProtocolHandler.cs
```diff
@@ -209,18 +197,7 @@ bool CanAcceptBlockGossip()
             }
         }
 
-        private bool CanGossip => EnsureGossipPolicy();
-
-        private bool EnsureGossipPolicy()
-        {
-            if (!_gossipPolicy.CanGossipBlocks)
-            {
-                SyncServer.StopNotifyingPeersAboutNewBlocks();
```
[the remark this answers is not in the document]
alexb5dh: Added new notification and reverted this change

src/Nethermind/Nethermind.Synchronization/SyncServer.cs
```diff
@@ -424,7 +427,7 @@ private void OnNewHeadBlock(object? sender, BlockEventArgs blockEventArgs)
 
         private void NotifyOfNewBlock(PeerInfo? peerInfo, ISyncPeer syncPeer, Block broadcastedBlock, SendBlockMode mode)
         {
-            if (!_gossipPolicy.CanGossipBlocks) return;
+            if (!_gossipPolicy.CanGossipBlocks && !syncPeer.AlwaysNotifyOfNewBlock) return;
```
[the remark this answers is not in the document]
alexb5dh: Added new notification and reverted this change

src/Nethermind/Nethermind.Synchronization/SyncServer.cs
```diff
@@ -54,6 +55,9 @@ public class SyncServer : ISyncServer
         private readonly Hash256 _pivotHash;
         private BlockHeader? _pivotHeader;
 
+        private const int BlockRangeUpdateFrequency = 32;
```
alexb5dh, line 57: @LukaszRozmej, changes in this file are the "separate notification" approach we've discussed.

src/Nethermind/Nethermind.JsonRpc/Modules/Parity/EthProtocolInfo.cs
```diff
@@ -12,8 +12,8 @@ public class EthProtocolInfo
         [JsonPropertyName("version")]
         public byte Version { get; set; }
 
-        [JsonPropertyName("difficulty")]
-        public UInt256 Difficulty { get; set; }
+        [JsonPropertyName("difficulty"), JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+        public UInt256? Difficulty { get; set; } = 0;
```
LukaszRozmej:
```suggestion
        public UInt256? Difficulty { get; set; } = UInt256.Zero;
```

src/Nethermind/Nethermind.Network/ProtocolsManager.cs
```diff
@@ -430,5 +432,17 @@ public void SendNewCapability(Capability capability)
...
             }
         }
+
+        public int GetHighestProtocolVersion(string protocol)
+        {
+            int highestVersion = 0;
+            foreach (Capability capability in _capabilities)
+            {
+                if (capability.ProtocolCode == protocol && highestVersion < capability.Version)
+                    highestVersion = capability.Version;
```
LukaszRozmej:
```suggestion
                if (capability.ProtocolCode == protocol)
                    highestVersion = Math.Max(highestVersion, capability.Version);
```

src/Nethermind/Nethermind.Synchronization/SyncServer.cs
```diff
@@ -422,6 +427,41 @@ private void OnNewHeadBlock(object? sender, BlockEventArgs blockEventArgs)
...
+
+            if (!_pool.AllPeers.Any())
+                return;
+
+            Block latestBlock = latestBlockEventArgs.Block;
+            BlockHeader? latestUpdate = _lastBlockRangeUpdate?.latest;
+
+            // Notify at most once per 32 blocks
+            if (latestUpdate is not null && latestBlock.Number - latestUpdate.Number < BlockRangeUpdateFrequency)
+                return;
```
LukaszRozmej: Maybe simpler would be:
```suggestion
            if (latestBlock.Number % BlockRangeUpdateFrequency != 0)
                return;
```

src/Nethermind/Nethermind.Synchronization/SyncServer.cs
```diff
@@ -422,6 +427,41 @@ private void OnNewHeadBlock(object? sender, BlockEventArgs blockEventArgs)
             }
         }
 
+        private void OnNewRange(object? sender, BlockEventArgs latestBlockEventArgs)
+        {
+            if (Genesis is null)
+                return;
+
+            if (!_pool.AllPeers.Any())
```
LukaszRozmej:
```suggestion
            if (_pool.PeerCount == 0)
```

src/Nethermind/Nethermind.Synchronization/SyncServer.cs
LukaszRozmej: You don't need to track `latestUpdate` at all.

src/Nethermind/Nethermind.Synchronization/SyncServer.cs
alexb5dh: It's `IEnumerable`

src/Nethermind/Nethermind.Synchronization/SyncServer.cs
alexb5dh: Used that approach initially, but decided to change - in case if callback gets non-sequential blocks (i.e. gaps) or the same block twice. Is any of this possible?

LukaszRozmej: Yes it is possible on re-orgs, but I think it is a non-issue to send similar/same announcements.

PR #7058, opened 2024-05-21
Fix for #6960 and new unit tests for testing dependent batches

### LINKED ISSUE #6960: [Archive] Gnosis archive sync hangs on BeaconHeaders after "minimumLength ('-192') must be a non-negative value" error
Started a Gnosis archive node sync.
All was going smooth on stage before actual processing (beaconHeaders) but after a while I got an exception:
```
System.ArgumentOutOfRangeException: minimumLength ('-192') must be a non-negative value. (Parameter 'minimumLength')
Actual value was -192.
   at System.ArgumentOutOfRangeException.ThrowNegative[T](T value, String paramName)
   at System.ArgumentOutOfRangeException.ThrowIfNegative[T](T value, String paramName)
   at System.Buffers.SharedArrayPool`1.Rent(Int32 minimumLength)
   at Nethermind.Synchronization.FastBlocks.HeadersSyncFeed.BuildDependentBatch(HeadersSyncBatch batch, Int64 addedLast, Int64 addedEarliest) in /_/src/Nethermind/Nethermind.Synchronization/FastBlocks/FastHeadersSyncFeed.cs:line 459
   at Nethermind.Synchronization.FastBlocks.HeadersSyncFeed.InsertHeaders(HeadersSyncBatch batch) in /_/src/Nethermind/Nethermind.Synchronization/FastBlocks/FastHeadersSyncFeed.cs:line 593
   at Nethermind.Synchronization.FastBlocks.HeadersSyncFeed.HandleResponse(HeadersSyncBatch batch, PeerInfo peer) in /_/src/Nethermind/Nethermind.Synchronization/FastBlocks/FastHeadersSyncFeed.cs:line 389
   at Nethermind.Synchronization.ParallelSync.SyncDispatcher`1.DoHandleResponse(T request, PeerInfo allocatedPeer) in /_/src/Nethermind/Nethermind.Synchronization/ParallelSync/SyncDispatcher.cs:line 186
```

It appeared 5 times and my progress freezed at:
`Beacon Headers from block 1 to block 33647907 | 26,693,985 / 33,647,907 | queue         0 | current       11,718 Blk/s | total        9,047 Blk/s`
and was not moving any forward.

As this was on an automated node can't tell if restart helps.

ak88:
Fixes #6960

`InsertHeaders` now properly handles null when a dependent batch arrives. 

The added unit test also showcases different header batches that is allowed and not. 
Among them are a bit of strange behavior, since we allow a batch to start with null, but not empty.

This is not allowed
**batch 1-3**
header2, header3 

This is allowed
**batch 1-3**
null,header2, header3 

Also this will fix several instances of `ArrayPoolList` not being disposed in `HeaderSyncFeed`.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Synchronization.Test/FastBlocks/FastHeadersSyncTests.cs
+++ b/src/Nethermind/Nethermind.Synchronization.Test/FastBlocks/FastHeadersSyncTests.cs
@@ -439,7 +439,7 @@ namespace Nethermind.Synchronization.Test.FastBlocks
                         c--;
                     }
                 if (!useNulls)
-                    list = list.Where(h => h != null).ToList();
+                    list = list.Where(h => h is not null).ToList();
                 batch.Response = list.ToPooledList();
             }
 
--- a/src/Nethermind/Nethermind.Synchronization.Test/FastBlocks/FastHeadersSyncTests.cs
+++ b/src/Nethermind/Nethermind.Synchronization.Test/FastBlocks/FastHeadersSyncTests.cs
@@ -403,9 +403,9 @@ namespace Nethermind.Synchronization.Test.FastBlocks
         [TestCase(80, 1, 1, true, false)]
         [TestCase(80, 1, 1, true, true)]
         //All empty reponse
-        [TestCase(0, 192, 1, false,false)]
+        [TestCase(0, 192, 1, false, false)]
         //All null reponse
-        [TestCase(0, 192, 1, false,true)]
+        [TestCase(0, 192, 1, false, true)]
         public async Task Can_insert_all_good_headers_from_dependent_batch_with_missing_or_null_headers(int nullIndex, int count, int increment, bool shouldReport, bool useNulls)
         {
             var peerChain = Build.A.BlockTree().OfChainLength(1000).TestObject;
@@ -439,7 +439,7 @@ namespace Nethermind.Synchronization.Test.FastBlocks
                         c--;
                     }
                 if (!useNulls)
-                    list = list.Where(h=> h != null).ToList();
+                    list = list.Where(h => h != null).ToList();
                 batch.Response = list.ToPooledList();
             }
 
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Synchronization/FastBlocks/FastHeadersSyncFeed.cs
```diff
@@ -270,14 +275,21 @@ private void HandleDependentBatches(CancellationToken cancellationToken)
...
+                    lowest = LowestInsertedBlockHeader?.Number;
+                    cancellationToken.ThrowIfCancellationRequested();
 
-                processedBatchCount++;
+                    processedBatchCount++;
+                }
+                finally
+                {
+                    dependentBatch.Dispose();  
+                }
```
LukaszRozmej, line 292:
```
using (dependentBatch)
{
}
```

LukaszRozmej (commented):
why missing headers is not allowed?

src/Nethermind/Nethermind.Synchronization.Test/FastBlocks/FastHeadersSyncTests.cs
```diff
@@ -439,7 +439,7 @@ void FillBatch(HeadersSyncBatch batch, long start, bool applyNulls)
                         c--;
                     }
                 if (!useNulls)
-                    list = list.Where(h=> h != null).ToList();
+                    list = list.Where(h => h != null).ToList();
```
LukaszRozmej, line 442: is not nul

LukaszRozmej (approved):

ak88:
> why missing headers is not allowed?
> 
> EDIT: Ah the response should have same number of items as request - it is protocol rules. But maybe we can work around it?

I was pointing it out because it seems a little odd to allow empty or null in the end, but only null in the start. So if the rule is: The response has to be the same length as reques size, we should not allow empty in the end either.

src/Nethermind/Nethermind.Synchronization/FastBlocks/FastHeadersSyncFeed.cs
ak88: fixed

asdacap (approved):

PR #7067, opened 2024-05-24
Fix for not marking peer as low quality too soon if there is only one peer connected

### LINKED ISSUE #6803: Do not mark peer as low quality too soon if there is only one peer connected
In a situation where there is only one peer connected we have a logic which can very quickly mark this peer as low quality one:
https://github.com/NethermindEth/nethermind/blob/d18e4f695d979060804237aae16f92488c8f8d1f/src/Nethermind/Nethermind.Synchronization/SnapSync/SnapSyncFeed.cs#L172

It is problematic especially on testing scenario where we set only 1 selected peer to sync from but can be problematic on small chains as well. Would be really nice to improve that part of code to make sure that maybe first node should update pivot and then recheck this peer once more.

ManiBAJPAI22:
Fixes  #6803 

## Changes

- _List the changes_

- [x] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

#### Notes on testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks

_If the number of consecutive failures from the current peer (peerLastFailures) exceeds the AllowedInvalidResponses threshold, and all the recent failures are from this peer ```(allLastFailures == peerLastFailures)```, the peer is marked as LesserQuality.

In a scenario where there is only one peer connected, this condition will be met quickly, leading to the peer being marked as ```LesserQuality``` even though there are no other peers to compare with.

To address this issue and improve the behavior in situations with a single peer, we can modify the logic to update the pivot and retry with the same peer before marking it as ```LesserQuality```, when the current peer is the only one with failures, instead of immediately marking it as ```LesserQuality```, the code will update the pivot, clear the result log, and log a message indicating that the pivot has been updated and the peer will be retried.

The ```SyncResponseHandlingResult.OK``` is returned, allowing the sync process to retry with the same peer after updating the pivot.
If there are other peers with failures ```(allLastFailures != peerLastFailures)```, then the original logic of marking the current peer as LesserQuality is preserved.

This change should improve the behavior in scenarios with a single peer by giving the peer another chance after updating the pivot, before marking it as ```LesserQuality```. _


src/Nethermind/Nethermind.Synchronization/SnapSync/SnapSyncFeed.cs
```diff
@@ -162,15 +164,29 @@ public SyncResponseHandlingResult AnalyzeResponsePerPeer(AddRangeResult result,
...
-                                if (peerLastFailures > AllowedInvalidResponses)
-                                {
-                                    if (allLastFailures == peerLastFailures)
-                                    {
-                                        _logger.Trace($"SNAP - peer to be punished:{peer}");
-                                        return SyncResponseHandlingResult.LesserQuality;
-                                    }
+                                      peerLastFailures++;
+
+                                               if (peerLastFailures > AllowedInvalidResponses)
```
asdacap, line 169: Please cleanup indentation

ManiBAJPAI22: Sure !!

ManiBAJPAI22:
Indentations managed 
@asdacap kindly review it.

src/Nethermind/Nethermind.Synchronization/SnapSync/SnapSyncFeed.cs
```diff
@@ -168,18 +168,29 @@ public SyncResponseHandlingResult AnalyzeResponsePerPeer(AddRangeResult result,
...
+                                        
+                                        _snapProvider.UpdatePivot();
+                                        _resultLog.Clear();
+
+                                        _logger.Trace($"SNAP - updating pivot and retrying with peer: {peer}");
+
+                                        
+                                        return SyncResponseHandlingResult.OK;
+                                    }
+                                    else
```
benaadams, line 180: Won't this mark peer as lesser quality unless all failures were that peer? (i.e. if its less than all, it will also mark as Lesser)

ManiBAJPAI22:
@benaadams, @asdacap,

I've updated the code to address the concern regarding the peer quality evaluation logic. I nearly missed that detail, so thank you for pointing it out.

ManiBAJPAI22:
@benaadams @asdacap 
Any other thing that I need to do ?

src/Nethermind/Nethermind.Synchronization/SnapSync/SnapSyncFeed.cs
```diff
@@ -110,90 +110,93 @@ public override SyncResponseHandlingResult HandleResponse(SnapSyncBatch? batch,
         }
 
         public SyncResponseHandlingResult AnalyzeResponsePerPeer(AddRangeResult result, PeerInfo peer)
+{
```
asdacap, line 113: Indentation is still wrong.

ManiBAJPAI22: Sorry for the delayed response. 
I guess now Indentations are managed.

asdacap (approved):
LGTM.

asdacap:
Should put a unit tests for this case though.

ManiBAJPAI22:
Yeah checking on it !

stdevMac:
Closing — the unit test asdacap requested in Sept 2024 was never added, and the branch is ~600 days stale. If #6803 still reproduces, please reopen with a focused fix and tests.

PR #7070, opened 2024-05-24
Use tighter locking for Node list enumerator

### LINKED ISSUE #6995: NullReference during discovery process when syncing
During a SnapSync I noticed this Exception appears on mainnet

```
Error during discovery process: System.NullReferenceException: Object reference not set to an instance of an object.
   at Nethermind.Network.Discovery.RoutingTable.NodeTable.ClosestNodesEnumerator.MoveNext() in /_/src/Nethermind/Nethermind.Network.Discovery/RoutingTable/NodeTable.cs:line 112
   at Nethermind.Network.Discovery.NodesLocator.LocateNodesAsync(Byte[] searchedNodeId, CancellationToken cancellationToken) in /_/src/Nethermind/Nethermind.Network.Discovery/NodesLocator.cs:line 81
   at Nethermind.Network.Discovery.DiscoveryApp.RunDiscoveryProcess() in /_/src/Nethermind/Nethermind.Network.Discovery/DiscoveryApp.cs:line 459
```

Seems like it does not affect sync but not sure how it would affect peers pool.

benaadams:
Resolves #6995

## Changes

- Use more explict locking for the Node list enumerators

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Network.Discovery/RoutingTable/NodeTable.cs
+++ b/src/Nethermind/Nethermind.Network.Discovery/RoutingTable/NodeTable.cs
@@ -128,6 +128,10 @@ public class NodeTable : INodeTable
                     return true;
                 }
             }
+            catch (NullReferenceException)
+            {
+                // TODO: investigate why this happens
+            }
             finally
             {
                 _itemEnumerator.Dispose();
--- a/src/Nethermind/Nethermind.Network.Discovery/RoutingTable/NodeTable.cs
+++ b/src/Nethermind/Nethermind.Network.Discovery/RoutingTable/NodeTable.cs
@@ -128,13 +128,10 @@ public class NodeTable : INodeTable
                     return true;
                 }
             }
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Network.Discovery/RoutingTable/NodeTable.cs
```diff
@@ -128,6 +128,10 @@ public bool MoveNext()
                     return true;
                 }
             }
+            catch (NullReferenceException)
+            {
+                // TODO: investigate why this happens
```
LukaszRozmej, line 133: maybe log the internals of the enumerator here?

benaadams: Changed it to be tigher on the locking and not have floating locks

LukaszRozmej (approved):

PR #7074, opened 2024-05-25
Fast Call to Non contract addresses

benaadams:
## Changes

- If a call is to a non-contract address; just transfer the value rather than constructing a full call frame 
    - between 1% to 10% of calls per block are to non-contract (i.e. effectively only ETH transfers)

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Evm/CodeAnalysis/CodeInfo.cs
+++ b/src/Nethermind/Nethermind.Evm/CodeAnalysis/CodeInfo.cs
@@ -29,6 +29,7 @@ namespace Nethermind.Evm.CodeAnalysis
         }
 
         public bool IsPrecompile => Precompile is not null;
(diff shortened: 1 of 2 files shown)


benaadams:
Hive tests pass https://github.com/NethermindEth/nethermind/actions/runs/9232345307

LukaszRozmej (commented):
Would be good to run cosnesus hive tests

benaadams:
> Would be good to run cosnesus hive tests

https://github.com/NethermindEth/nethermind/actions/runs/9237001249

Same 256 failures as the prewarm branch

https://github.com/NethermindEth/nethermind/actions/runs/9238390120

Checking master

benaadams:
Consensus hive failures are recent EIP7610 tests https://github.com/ethereum/EIPs/pull/8161

LukaszRozmej (approved):

Issue #6843, opened 2024-03-19
[TxPool] TxPool limits break - it should keep 2048 tx at maximum


kamilchodola:
**Description**
Recently on one of the nodes I noticed that TxPool for some reason is breaking through 2048 limit and is as high as 10k or somethimes even more
<img width="798" alt="image" src="https://github.com/NethermindEth/nethermind/assets/43241881/09e3dc7c-a2bd-4e45-b955-d4bb5cdda6a6">

We should make sure we never go across this limit on both TxPool and blobPool and seems like recent improvements around that area didn't helped on that.

marcindsobczak:
There is a problem with removing worst tx from the pool
`Failed to remove the last item from the pool, the current state is 8366/2048. System.Func1[System.String]`
It is becoming from here:
https://github.com/NethermindEth/nethermind/blob/master/src/Nethermind/Nethermind.TxPool/Collections/SortedPool.cs#L485

Unfortunately logging is broken and we are not logging details of worst tx. Working on it.

aimxhaisse:
We also experienced this in one of mainnet instances (we run about ~12 Nethermind with identical configurations) ; restarting it will cap it again below 2K and from there it will be fine.

When this happens, there is a snowballing scenario:

- memory pressure increases
- the GC has to do more work
- as a result, execution time of some blocks can be > 15 secondes at tail
- if used for validation, you miss attestations on connected validators

This can take days to manifest before it impacts validation. On the logging side, you'll see this log from time to time:

```
22 Apr 23:32:36 | TxPool exceeds the config size 17165/2048 
```

Adding some more graphs (similar to @kamilchodola 's ones):

TX count:

![2024-04-24-103619_2171x391_scrot](https://github.com/NethermindEth/nethermind/assets/149659/1e139bd2-94e7-4f0d-9cc6-d27d22cbeaf5)

CPU usage of the node: the spikes here are when the GC has to do a lot of work:

![2024-04-24-103210_2160x405_scrot](https://github.com/NethermindEth/nethermind/assets/149659/56672637-f895-46c6-8573-9d1ccb2144f1)

Impact on block execution time:

![2024-04-24-103653_2173x422_scrot](https://github.com/NethermindEth/nethermind/assets/149659/c7e69ce2-4753-41a1-8f19-991e7ef8769b)

After restart, TX count goes down to 2K like other instances of the fleet:

[...]

PR #7079, opened 2024-05-26
Remove tx also from worst sorted values

benaadams:
Contributes to https://github.com/NethermindEth/nethermind/issues/6843

## Changes

- When removing from tx pool; also remove from WorstSortedValues; so non-existant values aren't in the WorstSorted txs

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.TxPool/Collections/SortedPool.cs
+++ b/src/Nethermind/Nethermind.TxPool/Collections/SortedPool.cs
@@ -443,6 +443,7 @@ namespace Nethermind.TxPool.Collections
         {
             if (_cacheMap.Remove(key))
             {
+                _worstSortedValues.Remove(value);
                 UpdateIsFull();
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.TxPool/Collections/SortedPool.cs
```diff
@@ -441,6 +441,9 @@ private void UpdateSortedValues(EnhancedSortedSet<TValue> bucket, TValue? previo
         /// </summary>
         protected virtual bool Remove(TKey key, TValue value)
         {
+            // Always remove from worst values; as may have been where it came from and dangling item
+            _worstSortedValues.Remove(value);
```
LukaszRozmej, line 445: you can then probably remove removing from here in `TryRemoveNonLocked`

marcindsobczak (approved):

PR #7080, opened 2024-05-26
Remove removed past PathHashes

benaadams:
## Changes

- When memory pruning an item; remove its entry from the LruCache (as its just been pruned)
- Shrink the size of the LruCache by default from 2.1M items to 1M items

Otherwise after a couple prunes the cache quickly grows to the maximum 2.1M items; some of which will never be used again.

image

- [x] Optimization

#### Requires testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Remarks

### DIFF
--- a/src/Nethermind/Nethermind.Init/InitializeStateDb.cs
+++ b/src/Nethermind/Nethermind.Init/InitializeStateDb.cs
@@ -104,11 +104,15 @@ public class InitializeStateDb : IStep
                 if (_logger.IsWarn) _logger.Warn($"Detected {pruningConfig.CacheMb}MB of pruning cache config. Pruning cache more than 2000MB is not recommended as it may cause long memory pruning time which affect attestation.");
             }
 
+            // 8 bytes Node pointer, (LinkedListNode 8 pointer + 8 pointer) + (HashAndTinyPath: 32 bytes for value + 8 bytes for path) + 32 bytes ValueHash256
+            int KeyEntryMemorySize = 8 + (8 + 8) + (32 + 8) + 32; // 96 bytes
+
             pruningStrategy = Prune
                 .WhenCacheReaches(pruningConfig.CacheMb.MB())
                 // Use of ratio, as the effectiveness highly correlate with the amount of keys per snapshot save which
                 // depends on CacheMb. 0.05 is the minimum where it can keep track the whole snapshot.. most of the time.
-                .TrackingPastKeys((int)(pruningConfig.CacheMb.MB() * pruningConfig.TrackedPastKeyCountMemoryRatio / 48))
+                // With the default 2000MB cache and 0.1 ratio, it can keep track of just over 1M keys.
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Init/InitializeStateDb.cs
```diff
@@ -104,11 +104,15 @@ public Task Execute(CancellationToken cancellationToken)
                 if (_logger.IsWarn) _logger.Warn($"Detected {pruningConfig.CacheMb}MB of pruning cache config. Pruning cache more than 2000MB is not recommended as it may cause long memory pruning time which affect attestation.");
             }
 
+            // 8 bytes Node pointer, (LinkedListNode 8 pointer + 8 pointer) + (HashAndTinyPath: 32 bytes for value + 8 bytes for path) + 32 bytes ValueHash256
+            int KeyEntryMemorySize = 8 + (8 + 8) + (32 + 8) + 32; // 96 bytes
```
asdacap, line 108: Can you double the default config also.

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -792,13 +792,14 @@ bool CanRemove(Hash256? address, TinyTreePath path, in TreePath fullPath, ValueH
...
                 HashAndTinyPath key = keyValuePair.Key;
-                if (_pastPathHash.TryGet(new(key.addr, in key.path), out ValueHash256 prevHash))
+                if (_pastPathHash.TryGet(key, out ValueHash256 prevHash))
                 {
                     TreePath fullPath = key.path.ToTreePath(); // Micro op to reduce double convert
                     if (CanRemove(key.addr, key.path, fullPath, prevHash, keyValuePair.Value))
                     {
                         Metrics.RemovedNodeCount++;
                         writeBatch.Remove(key.addr, fullPath, prevHash);
+                        _pastPathHash.Delete(key);
```
asdacap, line 802: Will this make mem pruning slower, (since it does more work), or faster? (since it become smaller).

src/Nethermind/Nethermind.Init/InitializeStateDb.cs
benaadams: Does it need 2.1M hashes?

benaadams: GC time is a function of number of objects (rather than size) I can redesign an Lru so it uses a giant array and indexes rather than pointers, if you want to go this high

asdacap: Yes, I've specifically, tune this amount.

benaadams: PR to eradicate the objects https://github.com/NethermindEth/nethermind/pull/7100

PR #7082, opened 2024-05-27
Optimization/pre warm intra block cache withdrawals

LukaszRozmej:
## Changes

- Pre warm withdrawals too

- [x] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Withdrawals/WithdrawalProcessor.cs
+++ b/src/Nethermind/Nethermind.Consensus/Withdrawals/WithdrawalProcessor.cs
@@ -31,9 +31,10 @@ public class WithdrawalProcessor : IWithdrawalProcessor
 
         if (block.Withdrawals is not null)
         {
-            for (int i = 0; i < block.Withdrawals.Length; i++)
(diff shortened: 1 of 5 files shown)


src/Nethermind/Nethermind.Consensus/Withdrawals/WithdrawalProcessor.cs
```diff
@@ -31,8 +31,9 @@ public void ProcessWithdrawals(Block block, IReleaseSpec spec)
 
         if (block.Withdrawals is not null)
         {
-            foreach (var withdrawal in block.Withdrawals)
+            for (int i = 0; i < block.Withdrawals.Length; i++)
```
benaadams, line 34: Take array to local if using for with length; or Jit won't eliminate the bounds checks as the array length could change
```csharp
var withdrawals = block.Withdrawals;
for (int i = 0; i < withdrawals.Length; i++)
    Withdrawal withdrawal = withdrawals[i];
```

benaadams (approved):

smartprogrammer93 (approved):

PR #7087, opened 2024-05-27
Use System.Collections.Concurrent

benaadams:
## Changes

- Use larger System.Collections.Concurrent.ConcurrentDictionary

- [x] Optimization


LukaszRozmej (approved):
maybe more empirical test on the cache sizes?

benaadams:
> maybe more empirical test on the cache sizes?

Can tweak in main PR

PR #7092, opened 2024-05-28
Reset state per block during archive sync

### LINKED ISSUE #7083: Archive sync OOM very quickly on Holesky
Some recent master change breaks archive sync and it OOMs super quickly (in 2 minutes fill 25 GB of RAM usage).

It worked well on: **Nethermind/v1.27.0-unstable+927f1d4a/linux-x64/dotnet8.0.5**

Does not happen on gnosis as very first blocks are small but happens within first 10k blocks on holesky which are bigger right after genesis.

image

benaadams:
Resolves #7083 

## Changes

- Reset world state per block when processing a long chain

#### Requires testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Remarks

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.cs
@@ -143,6 +143,7 @@ public partial class BlockProcessor : IBlockProcessor
                 }
 
                 preBlockStateRoot = processedBlock.StateRoot;
+                _stateProvider.Reset(resizeCollections: true);
             }
 
             if (options.ContainsFlag(ProcessingOptions.DoNotUpdateHead))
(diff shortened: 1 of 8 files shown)


src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.cs
```diff
@@ -143,6 +143,7 @@ public Block[] Process(Hash256 newBranchStateRoot, List<Block> suggestedBlocks,
                 }
 
                 preBlockStateRoot = processedBlock.StateRoot;
+                _stateProvider.Reset(resizeCollections: true);
```
LukaszRozmej, line 146: this would clear those caches per block? Maybe would be better to clear it per batch?

LukaszRozmej (commented):
@kamilchodola would be good to test the alternative I suggest.

src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.cs
benaadams: Blocks are fast anyway; builds up a lot of new junk that wasn't there before.

Can look at optimizing for a cross block cache as a different exercise?

LukaszRozmej (approved):

PR #7094, opened 2024-05-28
Fix block metrics + reduce metric contention

benaadams:
## Changes

- Use `ThreadLocal<long>` for contended metrics (cpu cache contention)
- Report using the processing thread's local for block metric in logs (so is only block, and not all threads work)

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Db/Metrics.cs
+++ b/src/Nethermind/Nethermind.Db/Metrics.cs
@@ -25,6 +25,7 @@ namespace Nethermind.Db
             }
         }
         private static ThreadLocal<long> _codeDbCache = new(trackAllValues: true);
+        [Description("Number of Code DB cache reads on thread.")]
         public static long ThreadLocalCodeDbCache => _codeDbCache.Value;
         public static void IncrementCodeDbCache() => _codeDbCache.Value++;
 
@@ -61,6 +62,7 @@ namespace Nethermind.Db
             }
         }
         private static ThreadLocal<long> _stateTreeReads = new(trackAllValues: true);
+        [Description("Number of State Trie reads on thread.")]
         public static long ThreadLocalStateTreeReads => _stateTreeReads.Value;
         public static void IncrementStateTreeReads() => _stateTreeReads.Value++;
 
@@ -121,6 +123,7 @@ namespace Nethermind.Db
             }
         }
         private static ThreadLocal<long> _storageTreeReads = new(trackAllValues: true);
+        [Description("Number of storage trie reads on thread.")]
         public static long ThreadLocalStorageTreeReads => _storageTreeReads.Value;
         public static void IncrementStorageTreeReads() => _storageTreeReads.Value++;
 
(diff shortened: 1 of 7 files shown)


src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -2659,7 +2659,7 @@ private EvmExceptionType InstructionLog<TTracing>(EvmState vmState, ref EvmStack
         where TTracingRefunds : struct, IIsTracing
         where TTracingStorage : struct, IIsTracing
     {
-        Metrics.SstoreOpcode++;
+        Metrics.IncrementSStoreOpcode();
```
LukaszRozmej, line 2662: why do we need those Increment methods?

benaadams: To not expose the thread local outside metrics as its an implementation detail

benaadams: To follow up on this the property getter needs to be a vanilla `long` for graphana as it doesn't know what a `ThreadLocal<long>` is or how to process it; so when it's queried it sums all the thread local values to get a total value.

Can't `++` that value as its the sum of all threads; don't really want to expose the ThreadLocal directly as is implementation detail as it could be `Interlocked.Increment` or some other approach, etc

LukaszRozmej: Ok why not just add a setter that would set the ThreadLocal value - so the external code can just `++` on it?

src/Nethermind/Nethermind.Db/Metrics.cs
```diff
@@ -12,23 +12,76 @@ public static class Metrics
...
+                foreach (var value in _stateTreeCacheHits.Values)
+                {
+                    total += value;
+                }
+                return total;
+            }
+        }
+
+        private static ThreadLocal<long> _stateTreeCacheHits = new(trackAllValues: true);
+        public static void IncrementStateTreeCacheHits() => _stateTreeCacheHits.Value++;
```
LukaszRozmej, line 48:
```suggestion
        public static long StateTreeCache
        {
            get
            {
                long total = 0;
                foreach (var value in _stateTreeCacheHits.Values)
                {
                    total += value;
                }
                return total;
            }
            set
            {
                _stateTreeCacheHits.Value = value;
            }
        }

        private static ThreadLocal<long> _stateTreeCacheHits = new(trackAllValues: true);
```

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
LukaszRozmej: Ah it would read all threads and write to 1 thread. Would have to add `SstoreOpcodeLocal` that reads from threadlocal and increment that.

src/Nethermind/Nethermind.Db/Metrics.cs
benaadams: getter would return the sum of all thread values; would add one to it; then set the current thread's value to the total+1

So it would start exponentially increasing

LukaszRozmej (approved):

src/Nethermind/Nethermind.Db/Metrics.cs
LukaszRozmej: Yeah would need to add `StateTreeCacheLocal` with a proper getter too

PR #7096, opened 2024-05-28
Add custom Eth module that exposes additional OP related fields

flcl42:
## Changes

- Add OP specific fields to the receipts. They are different for deposit tx and the rest

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

- [x] Yes
- [x] No


LukaszRozmej (commented):
looks a bit overengeneered, couldn't we have `OptimismEthRpcModule : IEthRpcModule` and then return results with derived classes that can have different json serialization rules?

src/Nethermind/Nethermind.Serialization.Rlp/ReceiptArrayStorageDecoder.cs
```diff
@@ -19,12 +19,7 @@ public class ReceiptArrayStorageDecoder : IRlpStreamDecoder<TxReceipt[]>
     private static readonly IRlpValueDecoder<TxReceipt> CompactValueDecoder = Rlp.GetValueDecoder<TxReceipt>(RlpDecoderKey.Storage);
 
     public const int CompactEncoding = 127;
-    private readonly bool _useCompactEncoding = true;
-
-    public ReceiptArrayStorageDecoder(bool compactEncoding = true)
-    {
-        _useCompactEncoding = compactEncoding;
-    }
+    private readonly bool _useCompactEncoding = compactEncoding;
```
LukaszRozmej, line 22: this field is probably not needed now.

src/Nethermind/Nethermind.Optimism/Rpc/IOptimismEthRpcModule.cs
```diff
@@ -0,0 +1,337 @@
...
+using Nethermind.JsonRpc;
+using Nethermind.JsonRpc.Data;
+using Nethermind.JsonRpc.Modules;
+using Nethermind.JsonRpc.Modules.Eth;
+using Nethermind.State.Proofs;
+
+namespace Nethermind.Optimism;
+
+[RpcModule(ModuleType.Eth)]
+public interface IOptimismEthRpcModule : IRpcModule
```
LukaszRozmej, line 21: can this extend IEthRpcModule?

flcl42: Let me try to solve it via generics

src/Nethermind/Nethermind.Serialization.Rlp/ReceiptArrayStorageDecoder.cs
flcl42: If you mean we can drop non compact encoding, there a small concern: IReceiptConfig mentions that usage of compact encoding may effect RPC performance. Does it mean non compact receipts may still have users? If no - let's remove old encoding for sure

LukaszRozmej: I mean you can use default constructor parameter `compactEncoding` and don't need the field.

src/Nethermind/Nethermind.Optimism/Rpc/IOptimismEthRpcModule.cs
LukaszRozmej: why would they be needed? You can have a return type of base class and then return derived type - it should serialize correctly in json rpc.

flcl42:
@LukaszRozmej the latest commit a bit more complicated. But in addition to deduplication, I wanted to bring this general PoC of abstracted key data structures. Maybe we can do the same with transactions, blocks, etc later

src/Nethermind/Nethermind.Optimism/Rpc/IOptimismEthRpcModule.cs
flcl42: We can use IResultWrapper, but we will lose type information. Which may be useful for tests, documentation generation and less convenient for devs.

src/Nethermind/Nethermind.Serialization.Rlp/ReceiptArrayStorageDecoder.cs
flcl42: Oh indeed

src/Nethermind/Nethermind.Optimism/Rpc/OptimismEthRpcModule.cs
```diff
@@ -2,151 +2,109 @@
...
 using Nethermind.Serialization.Rlp;
-using Nethermind.State.Proofs;
+using Nethermind.State;
 using Nethermind.TxPool;
+using Nethermind.Wallet;
 
 namespace Nethermind.Optimism;
 
-public class OptimismEthRpcModule : IEthRpcModule
+public class OptimismEthRpcModule : EthRpcModule, IOptimismEthRpcModule
```
LukaszRozmej, line 31: Do we benefit from inheritance over composition?

deffrian (approved):

LukaszRozmej (approved):
I kind of would prefer composition for the module, but it is fine.

flcl42:
> I kind of would prefer composition for the module, but it is fine.

@LukaszRozmej I don't mind, but why?

LukaszRozmej:
> > I kind of would prefer composition for the module, but it is fine.
> 
> @LukaszRozmej I don't mind, but why?

Composition is always more flexible

PR #7097, opened 2024-05-28
If worsevalue removed but not found, reinsert new worst item

### LINKED ISSUE #6843: [TxPool] TxPool limits break - it should keep 2048 tx at maximum

benaadams:
Resolves #6843

## Changes

- If a worst value is removed for the group, but wasn't in the group, readd a new worst value for the group

#### Requires testing

- [x] No


LukaszRozmej (approved):
Let's wait for @marcindsobczak review.

marcindsobczak (approved):

PR #7100, opened 2024-05-28
Reduce objects used by LruCaching

benaadams:
## Changes

- The speed of GC is a function of the number of objects allocated (rather than size)
    - Reduce number of objects by 4M by using an array with offsets rather than linked list for Lru
    - Disadvantage is fixed size

- [x] Optimization

#### Requires testing

- [x] Yes

- [x] Yes


src/Nethermind/Nethermind.Core/Caching/LruCacheLowObject.cs
```diff
@@ -0,0 +1,249 @@
...
+        private Stack<int> _freeOffsets = new();
+        private readonly LruCacheItem[] _items;
+
+        public LruCacheLowObject(int maxCapacity, string name)
+        {
+            ArgumentOutOfRangeException.ThrowIfLessThan(maxCapacity, 1);
+
+            _name = name;
+            _maxCapacity = maxCapacity;
+            _cacheMap = new Dictionary<TKey, int>(maxCapacity); // do not initialize it at the full capacity
```
LukaszRozmej, line 30: comment irrelevant?

src/Nethermind/Nethermind.Core/Caching/LruCacheLowObject.cs
```diff
@@ -0,0 +1,249 @@
...
+            newNode.Prev = offset;
+            _leastRecentlyUsed = offset;
+        }
+
+        private struct LruCacheItem(TKey k, TValue v)
+        {
+            public readonly TKey Key = k;
+            public TValue Value = v;
+            public int Next;
+            public int Prev;
```
LukaszRozmej, line 246: We could using `Deque` (there is one is solution from a package using array under) removing need for those 2 fields.
Would make tracking indexes in the dictionary a bit harder and we would probably need to expose native indexes of Dequeue (instead of collection indexes - as they would change constantly).

But it is potential huge win on size - 8 bytes less for each LruCacheItem.

LukaszRozmej (approved):

src/Nethermind/Nethermind.Core/Caching/LruCacheLowObject.cs
LukaszRozmej: Although it constantly compacts the array if something is removed from the middle, so probably a bad idea in the end.

src/Nethermind/Nethermind.Core/Caching/LruCacheLowObject.cs
benaadams: A little ;)

PR #7101, opened 2024-05-29
Fix forward sync regression, block processing slows down over time

asdacap:
- It still regress by about 20% when prewarm is disabled. When enable net increase is about 20%. This is with db file warmer on a 196GB ram machine.
- Before, After
Screenshot from 2024-05-29 10-31-05

## Changes

- Fix

- [X] Optimization

### DIFF
--- a/src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
+++ b/src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
@@ -11,6 +11,7 @@ using Nethermind.Core.Collections;
 using Nethermind.Core.Crypto;
 using Nethermind.Core.Extensions;
 using Nethermind.Core.Resettables;
+using Nethermind.Int256;
 using Nethermind.Logging;
 using Nethermind.State.Tracing;
 using Nethermind.Trie.Pruning;
@@ -28,7 +29,7 @@ namespace Nethermind.State
         private readonly ILogManager? _logManager;
         internal readonly IStorageTreeFactory _storageTreeFactory;
         private readonly ResettableDictionary<AddressAsKey, StorageTree> _storages = new();
-        private readonly HashSet<Address> _toUpdateRoots = new();
+        private readonly HashSet<AddressAsKey> _toUpdateRoots = new();
 
         /// <summary>
         /// EIP-1283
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
```diff
@@ -36,7 +36,7 @@ internal sealed class PersistentStorageProvider : PartialStorageProviderBase
         private readonly ResettableDictionary<StorageCell, byte[]> _originalValues = new();
 
         private readonly ResettableHashSet<StorageCell> _committedThisRound = new();
-        private readonly Dictionary<StorageCell, byte[]> _blockCache = new(4_096);
+        private readonly Dictionary<Address, Dictionary<StorageCell, byte[]>> _blockCache = new(4_096);
```
benaadams, line 39:
```suggestion
        private readonly Dictionary<AddressAsKey, Dictionary<StorageCell, byte[]>> _blockCache = new(4_096);
```

Does the second dictionary need to be full `StorageCell` as key or can it be just the `StorageCell.Index`?

benaadams (approved):

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
asdacap: No idea.

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
```diff
@@ -36,7 +36,7 @@ internal sealed class PersistentStorageProvider : PartialStorageProviderBase
         private readonly ResettableDictionary<StorageCell, byte[]> _originalValues = new();
 
         private readonly ResettableHashSet<StorageCell> _committedThisRound = new();
-        private readonly Dictionary<StorageCell, byte[]> _blockCache = new(4_096);
+        private readonly Dictionary<AddressAsKey, Dictionary<StorageCell, byte[]>> _blockCache = new(4_096);
```
benaadams, line 39: As what is saved to tree is only the index `tree.Set(change.StorageCell.Index, change.Value)`

Looks like it can be:

```suggestion
        private readonly Dictionary<AddressAsKey, Dictionary<UInt256, byte[]>> _blockCache = new(4_096);
```

Rather than using the much larger `Address|UInt256|bool` tuple that StorageKey is

benaadams (approved):

PR #7102, opened 2024-05-29
Fix cannot disable prewarm

asdacap:
- Fix migration config throw error when setting a value because it check the default value via `ToString()` which does not match the string value of the `DefaultValue`.


benaadams (approved):

LukaszRozmej (approved):

LukaszRozmej:
Should we use comparison that ignores character casing?

asdacap:
Ahhh good point.

PR #7104, opened 2024-05-29
Fix pruning regression

asdacap:
- Don't know why, can't figure it out, but it is reproducable.
- Normally the remove ratio (`rate(nethermind_removed_node_count[$__rate_interval])/rate(nethermind_persisted_node_count[$__rate_interval])`) is around 87%. For some reason, it drops to 20% after around 4000 blocks. This would increase database growth by about 6x.
- Before, after.
Screenshot_2024-05-29_14-45-51

### DIFF
--- a/src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
+++ b/src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
@@ -276,7 +276,7 @@ namespace Nethermind.Trie.Pruning
 
         // Track some of the persisted path hash. Used to be able to remove keys when it is replaced.
         // If null, disable removing key.
-        private LruCacheLowObject<HashAndTinyPath, ValueHash256>? _pastPathHash;
+        private LruCache<HashAndTinyPath, ValueHash256>? _pastPathHash;
 
         // Track ALL of the recently re-committed persisted nodes. This is so that we don't accidentally remove
         // recommitted persisted nodes (which will not get re-persisted).


src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -276,7 +276,7 @@ public override bool Equals(object? obj)
 
         // Track some of the persisted path hash. Used to be able to remove keys when it is replaced.
         // If null, disable removing key.
-        private LruCacheLowObject<HashAndTinyPath, ValueHash256>? _pastPathHash;
+        private LruCache<HashAndTinyPath, ValueHash256>? _pastPathHash;
```
benaadams, line 279: Regular LruCache drops block validation throughput by 2/3rds after second prune 😢

asdacap: Can you try `--Db.StateDbWriteBufferSize 256000000` and see if it take about 8 to 10 prune to happen again?

benaadams: Will tell you the result in many hours 😅 The prunes aren't that frequent

LukaszRozmej: this would mean something is wrong with LruCacheLowObject

kamilchodola: @benaadams can you reduce pruning cache to small number so will be more frequent?

asdacap: Well, the memory pruning time increase, so thats unfortunate.

benaadams: > Can you try --Db.StateDbWriteBufferSize 256000000 and see if it take about 8 to 10 prune to happen again?

Seems to have not really slowed down yet; the is a drift down and a drift up in memory; but nothing as dramatic as before
image

> That is likely the reason you observe slowdown instead of this variable itself.

What's the premise here? The change caused less items to be deleted, so more new files to be written? Which means it didn't get slower? 🤔

asdacap: Less new files. Delete takes up space until it is compacted to bottom.

benaadams: > see if it take about 8 to 10 prune to happen again?

Small drop at 5; large drop at 15
image

asdacap: Then the drop is not because of the LRU, but likely about creation of new uncached files.

benaadams: Any way to resolve?

benaadams: The weird bit is restarting (without the file preload) and performance jumps back up

asdacap: No idea. Maybe try on linux to see if it has the same effect.

benaadams: Did a manual db load; as done on startup, didn't change anything.

Would `WriteFlags.DisableWAL` for the millions of removes help?

benaadams:
Think know what's wrong with the cache

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
benaadams: Or flush the db cache if its now filled with lots of stuff that will no longer be accessed?

benaadams: Bypassing WAL for the delete batch seems to have fixed it; testing a bit more

benaadams (approved):

benaadams:
Not 100% have fixed the issue with slowdown, but it fixes the issue with the  pruning which caused the improvement in slow down. So will monitor the slow down and try to fix it better if it happens again

benaadams:
Am seeing the exact same removed and nodes cached with `LruCache` (left) and `LruCacheLowObject` (right) middle graph; but more performance collapse with `LruCache`

image

benaadams:
Were some issues though with the key types

benaadams:
Retesting as a change wasn't in; not sure if git or me

asdacap:
Can you put the other change on another PR? I just want to revert that one line in this one.

asdacap:
Keep in mind, it take about 4000 blocks for the remove ratio to drop. I recommend having and old backup and replaying from that or you'll have to wait 10 hour.

benaadams:
Done

PR #7105, opened 2024-05-29
Increase Docker creation timeout for oneClick

kamilchodola:
As Docker image creation takes longer and longer (now very close to 10 minutes) timeout is happening from time to time so to prevent that increasing it to 20 minutes.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [X] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [X] No

- [ ] Yes
- [X] No


brbrr (approved):
lgtm!

PR #7109, opened 2024-05-30
Refactor/make pre warmer worldstate agnostic

asdacap:
- Make pre block warmer world state type agnostic.

## Types of changes

- [X] Refactoring


LukaszRozmej (commented):
I have mixed feelings, we are bloating `IWorldState` and kind of hide a bit the intention of the hashes.

asdacap:
We should hide it. A different state db might not even have rlp encoding at all.

src/Nethermind/Nethermind.State/IWorldState.cs
```diff
@@ -108,4 +108,11 @@ public interface IWorldState : IJournal<Snapshot>, IReadOnlyStateProvider
 
     void CommitTree(long blockNumber);
     ArrayPoolList<AddressAsKey>? GetAccountChanges();
+
+    /// <summary>
+    /// PreBlockCache related
+    /// </summary>
+    bool SupportPreBlockCache => false;
+    bool IsCacheDirty => false;
+    void ClearCache() { }
```
LukaszRozmej, line 117: Can we at least hide it under some `PreBlockCache` property?

asdacap: Are you sure? `PreBlockCaches` is clearly implementation specific. I'd even err towards `PrepareForPreBlockWarmer()` instead of the `IsCacheDirty` and `ClearCache` here.

LukaszRozmej: We can expose something that has only `IsCacheDirty` and `ClearCache`. I just want to minimize elements in `IWorldState`

src/Nethermind/Nethermind.State/ReadOnlyWorldStateManager.cs
```diff
@@ -36,13 +36,23 @@ ILogManager logManager
 
     public IStateReader GlobalStateReader { get; }
 
-    public IWorldState CreateResettableWorldState(PreBlockCaches? sharedHashes = null) =>
-        new WorldState(sharedHashes is not null
-                ? new PreCachedTrieStore(_readOnlyTrieStore, sharedHashes.RlpCache)
-                : _readOnlyTrieStore,
+    public IWorldState CreateResettableWorldState(IWorldState? forWarmup = null)
+    {
+        if (forWarmup != null)
```
LukaszRozmej, line 41: is not null

src/Nethermind/Nethermind.State/ReadOnlyWorldStateManager.cs
```diff
@@ -36,13 +36,23 @@ ILogManager logManager
...
+        {
+            PreBlockCaches preBlockCaches = (forWarmup as IPreBlockCaches)!.Caches;
```
LukaszRozmej, line 43: either do nullability and `?.` or do proper hard cast that throws exception.

src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
```diff
@@ -17,20 +17,20 @@
...
     private readonly ObjectPool<SystemTransaction> _systemTransactionPool = new DefaultObjectPool<SystemTransaction>(new DefaultPooledObjectPolicy<SystemTransaction>(), Environment.ProcessorCount);
     private readonly ILogger _logger = logManager.GetClassLogger<BlockCachePreWarmer>();
 
     public Task PreWarmCaches(Block suggestedBlock, Hash256? parentStateRoot, CancellationToken cancellationToken = default)
     {
-        if (preBlockCaches is not null)
+        if (targetWorldState is not null)
         {
-            if (preBlockCaches.IsDirty)
+            if (targetWorldState.IsCacheDirty)
```
LukaszRozmej, line 30: Let's remove `IsCacheDirty` for simplicity and have `ClearCache` return `true` if cache was dirty and `false` if it was empty.

LukaszRozmej (approved):

PR #7116, opened 2024-05-31
Use string for additional rocksdb config

asdacap:
- Turns out, you can use string to access options that is not available from the C (and therefore c#) binding.
- Notable options are `compaction_pri=kByCompensatedSize` which prioritize removing items.
- No significant impact on state db though. 
- Still, easy flag for experimentation.

#### Requires testing

- [ ] Yes
- [X] No

- [X] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
+++ b/src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
@@ -707,8 +707,14 @@ public class DbOnTheRocks : IDb, ITunableDb
         if (dbConfig.AdditionalRocksDbOptions is not null)
         {
             IntPtr optsPtr = Marshal.StringToHGlobalAnsi(dbConfig.AdditionalRocksDbOptions);
-            _rocksDbNative.rocksdb_get_options_from_string(options.Handle, optsPtr, options.Handle);
-            Marshal.FreeHGlobal(optsPtr);
+            try
+            {
+                _rocksDbNative.rocksdb_get_options_from_string(options.Handle, optsPtr, options.Handle);
+            }
+            finally
+            {
+                Marshal.FreeHGlobal(optsPtr);
+            }
         }
 
         #endregion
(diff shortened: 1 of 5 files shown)


src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
```diff
@@ -708,6 +703,14 @@ private long GetMemtableSize()
             options.EnableStatistics();
         }
         options.SetStatsDumpPeriodSec(dbConfig.StatsDumpPeriodSec);
+
+        if (dbConfig.AdditionalRocksDbOptions != null)
```
LukaszRozmej, line 707: is not null

src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
```diff
@@ -708,6 +703,14 @@ private long GetMemtableSize()
...
+        {
+            IntPtr optsPtr = Marshal.StringToHGlobalAnsi(dbConfig.AdditionalRocksDbOptions);
+            _rocksDbNative.rocksdb_get_options_from_string(options.Handle, optsPtr, options.Handle);
+            Marshal.FreeHGlobal(optsPtr);
```
LukaszRozmej, line 711: finally?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
rubo:
```suggestion
        if (dbConfig.AdditionalRocksDbOptions is not null)
````

src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
rubo: You mean `try/finally` for freeing?

asdacap: done

PR #7125, opened 2024-06-03
Keep full tx pool in worst sorted set

benaadams:
Resolves https://github.com/NethermindEth/nethermind/issues/6843

## Changes

- Rather than more complex juggling of keeping last tx from each group; just keep full tx pool; then values can be kept in lockset with the pool and logic simplified

#### Requires testing

- [x] No


LukaszRozmej:
I prefer to fix the issue.

PR #7128, opened 2024-06-03
use AccountChanges only in main processing thread

LukaszRozmej:
## Changes

- use AccountChanges only in main processing thread

- [x] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.TxPool/TxPool.cs
+++ b/src/Nethermind/Nethermind.TxPool/TxPool.cs
@@ -211,18 +211,22 @@ namespace Nethermind.TxPool
                         try
                         {
                             ArrayPoolList<AddressAsKey>? accountChanges = args.Block.AccountChanges;
-                            if (!CanUseCache(args.Block, accountChanges))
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -211,18 +211,22 @@ private void ProcessNewHeads()
...
                             }
-                            else
-                            {
-                                // Sequential block, just remove changed accounts from cache
-                                _accountCache.RemoveAccounts(accountChanges);
-                            }
-                            args.Block.AccountChanges = null;
-                            accountChanges?.Dispose();
 
                             _lastBlockNumber = args.Block.Number;
```
benaadams, line 231: I think these need to move inside the null check?

benaadams (approved):

PR #7130, opened 2024-06-03
Persistently remove worst item until actually removed

### LINKED ISSUE #6843: [TxPool] TxPool limits break - it should keep 2048 tx at maximum

LukaszRozmej:
Resolves #6843?

## Changes

- Change access to `value` on removal.
- RemoveLast will try until it succeeds.

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.TxPool/Collections/SortedPool.cs
+++ b/src/Nethermind/Nethermind.TxPool/Collections/SortedPool.cs
@@ -12,6 +12,7 @@ using System.Threading;
 using Nethermind.Core.Collections;
 using Nethermind.Core.Threading;
 using Nethermind.Logging;
+using Org.BouncyCastle.Asn1.Cms.Ecc;
 
 namespace Nethermind.TxPool.Collections
 {
@@ -227,33 +228,10 @@ namespace Nethermind.TxPool.Collections
         {
             if (Remove(key, out value) && value is not null)
             {
-                TGroupKey groupMapping = MapToGroup(value);
-                if (_buckets.TryGetValue(groupMapping, out EnhancedSortedSet<TValue>? bucketSet))
+                if (RemoveFromBucket(value, out EnhancedSortedSet<TValue>? bucketSet))
                 {
                     bucket = bucketSet;
-                    TValue? last = bucketSet.Max;
-                    if (bucketSet.Remove(value))
-                    {
-                        if (bucket.Count == 0)
-                        {
-                            _buckets.Remove(groupMapping);
-                            if (last is not null)
-                            {
-                                _worstSortedValues.Remove(last);
-                                UpdateWorstValue();
-                            }
-                        }
-                        else
-                        {
-                            UpdateSortedValues(bucketSet, last);
-                        }
-
-                        _snapshot = null;
-
-                        return true;
-                    }
-
-                    Removed?.Invoke(this, new SortedPoolRemovedEventArgs(key, value, groupMapping, evicted));
+                    Removed?.Invoke(this, new SortedPoolRemovedEventArgs(key, value, evicted));
                 }
                 else // just for safety
                 {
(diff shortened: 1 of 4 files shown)


src/Nethermind/Nethermind.TxPool/Collections/SortedPool.cs
```diff
@@ -439,26 +452,17 @@ private void UpdateSortedValues(EnhancedSortedSet<TValue> bucket, TValue? previo
         /// <summary>
         /// Actual removal mechanism.
         /// </summary>
-        protected virtual bool Remove(TKey key, TValue value)
+        protected virtual bool Remove(TKey key, out TValue? value)
         {
-            // Always remove from worst values; as may have been where it came from and dangling item
-            bool removed = _worstSortedValues.Remove(value);
```
LukaszRozmej, line 445: this couldn't work because value comes from `_cacheMap.TryGetValue` in the first place

src/Nethermind/Nethermind.TxPool/Collections/SortedPool.cs
```diff
@@ -371,16 +369,31 @@ public virtual bool TryInsert(TKey key, TValue value, out TValue? removed)
...
+                if (_worstSortedValues.Remove(value))
+                {
+                    TGroupKey groupMapping = MapToGroup(value);
+                    if (_buckets.TryGetValue(groupMapping, out EnhancedSortedSet<TValue>? bucket))
+                    {
+                        UpdateSortedValues(bucket, previousLast: default);
+                    }
+                }
+
+                goto TryAgain;
```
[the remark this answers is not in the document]
benaadams: Is value removed from `_worstSortedValues` in all paths and `UpdateWorstValue()` called?

e.g. if `_buckets.TryGetValue` it will loop with the same value in `_worstValue.GetValueOrDefault()` and end up in infinite loop?

src/Nethermind/Nethermind.TxPool/Collections/SortedPool.cs
```diff
@@ -225,38 +225,41 @@ private bool TryRemove(TKey key, out TValue? value, [NotNullWhen(true)] out ICol
...
+                        return true;
                     }
 
                     Removed?.Invoke(this, new SortedPoolRemovedEventArgs(key, value, groupMapping, evicted));
                 }
+                else // just for safety
+                {
+                    _worstSortedValues.Remove(value);
+                    UpdateWorstValue();
+                }
```
LukaszRozmej, line 262: Added it here

marcindsobczak (commented):
What if somehow tx is not present in `_cacheMap`, but present in bucket and `_worstValue`? `Remove` in `TryRemoveNonLocked` will return `false`, in `RemoveLast` will then remove it from `_worstSortedValues`, but it will remain in bucket. `UpdateSortedValues` will readd it to  `_worstSortedValues` and finally as `_worstValue`. I think we need to make an attempt to remove it from everywhere (so buckets and cacheMap) when in `RemoveLast`, in `if (_buckets.TryGetValue ...)`, just before `UpdateSortedValues`

LukaszRozmej:
> What if somehow tx is not present in `_cacheMap`, but present in bucket and `_worstValue`? `Remove` in `TryRemoveNonLocked` will return `false`, in `RemoveLast` will then remove it from `_worstSortedValues`, but it will remain in bucket. `UpdateSortedValues` will readd it to `_worstSortedValues` and finally as `_worstValue`. I think we need to make an attempt to remove it from everywhere (so buckets and cacheMap) when in `RemoveLast`, in `if (_buckets.TryGetValue ...)`, just before `UpdateSortedValues`

What replicated on my machine was that `_buckets` and `_cacheMap` were consistent, and `_worstSortedValues` weren't - it had additional items. So this is the scenario I am trying to address.

marcindsobczak:
It's risky, if scenario with tx in bucket (and worstValue), but not in cacheMap will happen, we will end in endless loop. I prefer to avoid it, even if it is not current issue. And solution is simple, just remove from bucket and cacheMap before triggering UpdateSortedValues

marcindsobczak (approved):

src/Nethermind/Nethermind.TxPool/Collections/SortedPool.cs
```diff
@@ -371,16 +374,32 @@ public virtual bool TryInsert(TKey key, TValue value, out TValue? removed)
...
+                    return true;
+                }
+
+                TValue value = worstValue.Key;
+                if (_worstSortedValues.Remove(value))
+                {
+                    TGroupKey groupMapping = MapToGroup(value);
+                    if (_buckets.TryGetValue(groupMapping, out EnhancedSortedSet<TValue>? bucket))
+                    {
+                        bucket.Remove(value);
```
marcindsobczak, line 393: We can potentially end up with an empty bucket
```suggestion
                        if (bucket.Remove(value) && bucket.Count == 0)
                        {
                             _buckets.Remove(groupMapping);
                        }
```

benaadams (approved):
Could you trim the log also so it doesn't print the full tx at Warn level?

benaadams:
AccountAbstraction and Mev not happy with it though

marcindsobczak (approved):

benaadams (approved):

PR #7132, opened 2024-06-04
Limit contended paths to physical cores not logical cores

benaadams:
## Changes

- Use physical cores rather than logical cores as the degree of parallisation to reduce contention

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
@@ -9,6 +9,7 @@ using Nethermind.Core;
 using Nethermind.Core.Crypto;
 using Nethermind.Core.Specs;
 using Nethermind.Core.Threading;
+using Nethermind.Core.Cpu;
 using Nethermind.Evm;
 using Nethermind.Evm.Tracing;
 using Nethermind.Evm.TransactionProcessing;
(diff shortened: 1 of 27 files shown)


src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
```diff
@@ -54,7 +55,7 @@ private void PreWarmCachesParallel(Block suggestedBlock, Hash256 parentStateRoot
 
         try
         {
-            ParallelOptions parallelOptions = new() { MaxDegreeOfParallelism = Math.Max(1, Environment.ProcessorCount - 2), CancellationToken = cancellationToken };
+            ParallelOptions parallelOptions = new() { MaxDegreeOfParallelism = Math.Max(1, RuntimeInformation.PhysicalCoreCount - 2), CancellationToken = cancellationToken };
```
asdacap, line 58: I'd make this one specifically configurable.

asdacap (approved):

LukaszRozmej:
Is it helpful in any way? Logical Cores are most useful when thread is waiting for I/O, which should happen here.

benaadams:
> Is it helpful in any way? Logical Cores are most useful when thread is waiting for I/O, which should happen here.

Not if your I/O has lock contention; then are just increasing the wall clock time with more contention

image

LukaszRozmej (approved):

PR #7143, opened 2024-06-06
Skip prewarming tx if main processing has caught up

benaadams:
## Changes

- Skip transactions for prewarming where main processing has processed or is being processed

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Core/Block.cs
+++ b/src/Nethermind/Nethermind.Core/Block.cs
@@ -113,6 +113,8 @@ public class Block
 
     [JsonIgnore]
     public ArrayPoolList<AddressAsKey>? AccountChanges { get; set; }
+    [JsonIgnore]
+    internal volatile int TransactionProcessed;
 
     public override string ToString() => ToString(Format.Short);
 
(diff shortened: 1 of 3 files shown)


src/Nethermind/Nethermind.Core/Block.cs
```diff
@@ -113,6 +113,8 @@ public Transaction[] Transactions
 
     [JsonIgnore]
     public ArrayPoolList<AddressAsKey>? AccountChanges { get; set; }
+    [JsonIgnore]
+    internal volatile int TransactionProcessed;
```
LukaszRozmej, line 117: Can we pass it in different way? Block should be more of domain object.
Similar problem is with AccountChanges.

benaadams: Is a bit messy because of where the compents surface where currently the only shared comms is the block.

Will think about if it can be improved though

LukaszRozmej: lets introduce some sharable block processing context

PR #7160, opened 2024-06-10
Fix low object LruCaches

benaadams:
## Changes

- Replace was overriting `.Next` and `.Prev` after item had been moved. Carry those values through to new node

#### Requires testing

- [x] Yes

- [x] Yes

### DIFF
--- a/src/Nethermind/Nethermind.Core.Test/Caching/LruKeyCacheLowObjectTests.cs
+++ b/src/Nethermind/Nethermind.Core.Test/Caching/LruKeyCacheLowObjectTests.cs
@@ -79,6 +79,20 @@ namespace Nethermind.Core.Test.Caching
         }
 
         [Test]
+        public void Beyond_capacity_lru()
+        {
+            LruKeyCacheLowObject<AddressAsKey> cache = new(Capacity, "test");
+            for (int i = 0; i < Capacity * 2; i++)
+            {
+                for (int ii = 0; ii < Capacity / 2; ii++)
+                {
+                    cache.Set(_addresses[i]);
+                }
+                cache.Set(_addresses[i]);
+            }
+        }
+
+        [Test]
         public void Can_delete()
         {
             LruKeyCacheLowObject<AddressAsKey> cache = new(Capacity, "test");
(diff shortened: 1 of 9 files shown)


src/Nethermind/Nethermind.Core.Test/Caching/LruKeyCacheLowObjectTests.cs
```diff
@@ -78,6 +78,20 @@ public void Beyond_capacity()
...
+            for (int i = 0; i < Capacity * 2; i++)
+            {
+                for (int ii = 0; ii < Capacity / 2; ii++)
+                {
+                    cache.Set(_addresses[i]);
+                }
+                cache.Set(_addresses[i]);
+            }
+        }
+
```
LukaszRozmej, line 94: no asserts in this tests?

src/Nethermind/Nethermind.Core.Test/Caching/LruKeyCacheTests.cs
```diff
@@ -72,17 +72,31 @@ public void Beyond_capacity()
...
+            for (int i = 0; i < Capacity * 2; i++)
+            {
+                for (int ii = 0; ii < Capacity / 2; ii++)
+                {
+                    cache.Set(_addresses[i]);
+                }
+                cache.Set(_addresses[i]);
+            }
+        }
+
```
LukaszRozmej, line 94: no asserts in this tests?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Core.Test/Caching/LruKeyCacheLowObjectTests.cs
benaadams: throws exception if its wrong

src/Nethermind/Nethermind.Core.Test/Caching/LruKeyCacheTests.cs
benaadams: throws exception if its wrong

PR #7164, opened 2024-06-10
Keep spinning if next in queue for McsLock

benaadams:
## Changes

- If thread is next in queue keep spinning rather than waiting for signal
- Singal unlock in background as next in line will still be spinning so additional latency only effects threads waiting for singal; meaning lockholder returns faster

Faster!
image

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Core/Threading/McsLock.cs
+++ b/src/Nethermind/Nethermind.Core/Threading/McsLock.cs
@@ -27,8 +27,6 @@ public class McsLock
     /// </summary>
     private PaddedTail _tail;
 
-    internal volatile ThreadNode? _currentLockHolder = null;
-
     /// <summary>
     /// Acquires the lock. If the lock is already held, the calling thread is placed into a queue and
     /// enters a busy-wait state until the lock becomes available.
@@ -38,7 +36,7 @@ public class McsLock
         ThreadNode node = _node.Value!;
 
         // Check for reentrancy.
-        if (ReferenceEquals(node, _currentLockHolder))
+        if (node.State != (nuint)LockState.Unlocked)
             ThrowInvalidOperationException();
 
         node.State = (nuint)LockState.Waiting;
@@ -54,8 +52,6 @@ public class McsLock
         }
 
         Interlocked.MemoryBarrier();
-        // Set current lock holder.
-        _currentLockHolder = node;
 
         return new Disposable(this);
 
@@ -149,7 +145,7 @@ public class McsLock
                 if (Interlocked.CompareExchange(ref _lock._tail.Value, null, node) == node)
                 {
                     // Clear current lock holder.
-                    _lock._currentLockHolder = null;
+                    node.State = (nuint)LockState.Unlocked;
                     return;
                 }
 
@@ -161,7 +157,7 @@ public class McsLock
 
             ThreadNode next = node.Next!;
             // Clear current lock holder.
-            _lock._currentLockHolder = null;
+            node.State = (nuint)LockState.Unlocked;
             // Pass the lock to the next thread by setting its 'Locked' flag to false.
             next.State = (nuint)LockState.ReadyToAcquire;
 
@@ -212,7 +208,8 @@ public class McsLock
     {
         ReadyToAcquire = 0,
         Waiting = 1,
-        Acquired = 2
+        Acquired = 2,
+        Unlocked = 3
     }
 
     [StructLayout(LayoutKind.Explicit, Size = 128)]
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Core/Threading/McsLock.cs
```diff
@@ -175,6 +190,14 @@ private static void SpinTillNextNotNull(ThreadNode node)
         }
 
         private static void SignalUnlock(ThreadNode nextNode)
+        {
+            // Signal the next node to wake up via ThreadPool to return faster
+            // If it was next in line it should be still spinning
+            // rather sleeping so not need to wake up
+            ThreadPool.UnsafeQueueUserWorkItem((node) => BackgroundSignal(node), nextNode, preferLocal: true);
```
Scooletz, line 197: Is this really faster to offload the signal to the `ThreadPool`?

src/Nethermind/Nethermind.Core/Threading/McsLock.cs
```diff
@@ -76,16 +81,26 @@ private static void WaitForUnlock(ThreadNode node, ThreadNode predecessor)
         // the next thread in line may be sleeping when lock is released.
         while (node.State != (nuint)LockState.ReadyToAcquire)
         {
-            if (sw.NextSpinWillYield)
+            if (predecessor.State == (nuint)LockState.Waiting &&
```
Scooletz, line 84: So this weakens the condition to enter the `WaitForSignal` based on `Monitor` from _next will yield_ to _next will yield AND the predecessor is still waiting_.

src/Nethermind/Nethermind.Core/Threading/McsLock.cs
```diff
@@ -76,16 +81,26 @@ private static void WaitForUnlock(ThreadNode node, ThreadNode predecessor)
...
+                // Otherwise reset spinwait
+                sw.Reset();
             }
             else
             {
+                // Otherwise spin
                 sw.SpinOnce();
             }
+
+            Interlocked.MemoryBarrier();
```
Scooletz, line 103: What is synchronized here, what accesses?

src/Nethermind/Nethermind.Core/Threading/McsLock.cs
```diff
@@ -47,7 +47,12 @@ public Disposable Acquire()
         {
             WaitForUnlock(node, predecessor);
         }
+        else
+        {
+            node.State = (nuint)LockState.Acquired;
+        }
 
+        Interlocked.MemoryBarrier();
```
Scooletz, line 55: Could you comment what requires fencing here?

Scooletz (commented):
It would be great to see a test/bench that shows the gains. Beside that, some comments provided.

Scooletz:
Additional 2cc @benaadams 

A separate comment about padding and usage for `_currentLockHolder` and `_tail` that currently may land in the same cache line. 

1. `_currentLockHolder` as far as I see is used only for reentrancy check. If this is the case, the entrance could be stored in the node itself so that writes to the holder do not collide with `_tail` writes. Yes, it would require each node to store one bit of information.
2. if the one above does not hold, we should pad these two. `_currentLockHolder` is written by the thread that acquires and releases the lock while the `_tail` is accessed by any thread that wants to lock. Padding would help.

src/Nethermind/Nethermind.Core/Threading/McsLock.cs
benaadams: Yes, because to signal requires reaquiring the lock etc; its more expensive than queing to the ThreadPool.

Signalling via the ThreadPool is more lagged; however with this change the next in queue should be spining so not waiting for a signal; is only already sleeping threads that would expericent that latency, but they are already asleep so 🤷‍♂️.

Means a faster return for releasing the lock.

src/Nethermind/Nethermind.Core/Threading/McsLock.cs
benaadams: Yes; so if prior has the lock it will keep spinning but will only be one thread burning cpu (next in line)

benaadams:
>  Padding would help.

Added padding to _tail

> if we assumed one McsLock hold at the time why not make ThreadNode static?

Many different threads could have a different lock; so they couldn't all share the same static field? Or do you mean `ThreadStatic` rather than `ThreadLocal`?

stdevMac:
Closing — drafted and untouched since June 2024. Locking primitives have moved on; reopen if there's a reason to revive.

PR #7174, opened 2024-06-12
Perf/allow multiple sync allocation per peer

asdacap:
- At the moment, during sync, we have exactly one request per peer at any moment (per request type), for sync, which is behind an allocation system.
- This PR allows it to be more than one which is useful to account for the latency and processing overhead by the server.
- Add a config `--Sync.AllocationSlots` default to 2 to configure how allocation per peer.
- Disabled for headers as there seems to be a bug with it. 
- Useful for when you have one peer that you know is fast (because it ran locally), but can be faster with concurrent requests.
- Graph is 8, 4, 2, 1 allocation slots, syncing from another nethermind nodes, with background task parallelism set to 64 (unthrottled). 
Screenshot_2024-06-12_20-19-20
- When the background thread is set back to 1 on the server, 2 allocation slots only improve throughput slightly
- Graph is 4, 2, 1.
- Screenshot_2024-06-13_16-27-55

- [X] Optimization

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Synchronization/Peers/PeerInfo.cs
+++ b/src/Nethermind/Nethermind.Synchronization/Peers/PeerInfo.cs
@@ -249,7 +249,7 @@ namespace Nethermind.Synchronization.Peers
                 return cachedContext;
             }
 
-            cachedContext = Enum.GetValues<AllocationContexts>()
+            cachedContext = FastEnum.GetValues<AllocationContexts>()
                 .Where(aCtx => IsOnlyOneContext(aCtx) && (contexts & aCtx) != 0)
                 .ToArray();
 
--- a/src/Nethermind/Nethermind.Synchronization/Peers/PeerInfo.cs
+++ b/src/Nethermind/Nethermind.Synchronization/Peers/PeerInfo.cs
@@ -20,7 +20,7 @@ namespace Nethermind.Synchronization.Peers
 {
     public class PeerInfo
     {
-        private static readonly Dictionary<AllocationContexts, int> DefaultAllowances = new Dictionary<AllocationContexts, int>()
+        public static readonly Dictionary<AllocationContexts, int> DefaultAllowances = new Dictionary<AllocationContexts, int>()
         {
             {AllocationContexts.Headers, 1},
             {AllocationContexts.Bodies, 1},
@@ -250,7 +250,7 @@ namespace Nethermind.Synchronization.Peers
             }
 
             cachedContext = Enum.GetValues<AllocationContexts>()
-                .Where(aCtx => IsNotCompositeContexts(aCtx) && (contexts & aCtx) != 0)
+                .Where(aCtx => IsOnlyOneContext(aCtx) && (contexts & aCtx) != 0)
                 .ToArray();
 
             SeparatedContextsCache.TryAdd(contexts, cachedContext);
@@ -259,7 +259,7 @@ namespace Nethermind.Synchronization.Peers
 
         private static ConcurrentDictionary<AllocationContexts, AllocationContexts[]> SeparatedContextsCache = new ConcurrentDictionary<AllocationContexts, AllocationContexts[]>();
 
-        bool IsNotCompositeContexts(AllocationContexts x)
+        public static bool IsOnlyOneContext(AllocationContexts x)
         {
             return (x & (x - 1)) == 0;
         }
--- a/src/Nethermind/Nethermind.Synchronization/Peers/PeerInfo.cs
+++ b/src/Nethermind/Nethermind.Synchronization/Peers/PeerInfo.cs
@@ -6,6 +6,7 @@ using NonBlocking;
 using System.Collections.Generic;
 using System.Linq;
 using System.Runtime.CompilerServices;
+using System.Text;
 using System.Threading;
 using FastEnumUtility;
 using Nethermind.Blockchain.Synchronization;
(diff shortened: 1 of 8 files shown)


src/Nethermind/Nethermind.Synchronization/Peers/PeerInfo.cs
```diff
@@ -19,14 +20,33 @@ namespace Nethermind.Synchronization.Peers
...
         }
 
         public NodeClientType PeerClientType => SyncPeer?.ClientType ?? NodeClientType.Unknown;
 
-        public AllocationContexts AllocatedContexts { get; private set; }
+        public ConcurrentDictionary<AllocationContexts, int> AllocationSlots { get; init; }
+
+        public Dictionary<AllocationContexts, int> AvailableAllocationSlots => AllocationSlots
+            .Select(kv => (kv.Key, _allocationAllowances[kv.Key] - kv.Value))
+            .ToDictionary();
```
LukaszRozmej, line 49: simple array would be better? Or in-line array even.
You could ref and Interlock.Increment it.

src/Nethermind/Nethermind.Synchronization/Peers/PeerInfo.cs
```diff
@@ -158,6 +202,66 @@ private static string BuildContextString(AllocationContexts contexts)
...
+            SeparatedContextsCache.TryAdd(contexts, cachedContext);
+            return cachedContext;
+        }
+
+        private static ConcurrentDictionary<AllocationContexts, AllocationContexts[]> SeparatedContextsCache = new ConcurrentDictionary<AllocationContexts, AllocationContexts[]>();
+
+        public static bool IsOnlyOneContext(AllocationContexts x)
+        {
+            return (x & (x - 1)) == 0;
+        }
```
LukaszRozmej, line 265: Do we need this and the cache, can't we just yield return from the method?

src/Nethermind/Nethermind.Synchronization/Peers/PeerInfo.cs
```diff
@@ -158,6 +202,66 @@ private static string BuildContextString(AllocationContexts contexts)
...
+        public override string ToString() => $"[{BuildSlotContextString()} ][{BuildContextString(SleepingContexts)} ]{SyncPeer}";
+
+        private AllocationContexts[] SeparateAllocationContexts(AllocationContexts contexts)
+        {
+            if (SeparatedContextsCache.TryGetValue(contexts, out AllocationContexts[] cachedContext))
+            {
+                return cachedContext;
+            }
+
+            cachedContext = Enum.GetValues<AllocationContexts>()
```
LukaszRozmej, line 252: FastEnum?

LukaszRozmej (commented):
So why not 4 or 3 as default?

src/Nethermind/Nethermind.Synchronization/Peers/PeerInfo.cs
asdacap: Swtiched to enumerable

src/Nethermind/Nethermind.Synchronization/Peers/PeerInfo.cs
asdacap: But why? Its probably faster this way anyway. No alloc after cache is full.

src/Nethermind/Nethermind.Synchronization/Peers/PeerInfo.cs
asdacap: Done

asdacap:
2 as default because we don't wanna queue too much and cause timeout. The default background task parallelism is 1 so any speedup is by default is due to the latency between the node. We kinda probably should set large packets like body to 1, and small packet like headers to 4 maybe. But we already autoadjust their size based on latency, so it should be fine.

benaadams:
> So why not 4 or 3 as default?

Probably above 2 is heavy for peer on other side as you are doubling its load?

src/Nethermind/Nethermind.Synchronization/Peers/PeerInfo.cs
```diff
@@ -158,6 +201,66 @@ private static string BuildContextString(AllocationContexts contexts)
...
+
+            cachedContext = FastEnum.GetValues<AllocationContexts>()
+                .Where(aCtx => IsOnlyOneContext(aCtx) && (contexts & aCtx) != 0)
+                .ToArray();
+
+            SeparatedContextsCache.TryAdd(contexts, cachedContext);
+            return cachedContext;
+        }
+
+        private static ConcurrentDictionary<AllocationContexts, AllocationContexts[]> SeparatedContextsCache = new ConcurrentDictionary<AllocationContexts, AllocationContexts[]>();
```
LukaszRozmej, line 259: If this is a cache - initialize once and make it FrozenDictionary.

LukaszRozmej (changes requested):
Moved to FrozenDictionary, but this sounds very wasteful for me.

We should have a struct:
```
    public struct AllocationAllowences
    {
        public byte Header { get; }
        public byte Bodies { get; }
        public byte Receipts { get; }
        public byte State { get; }
        public byte Snap { get; }
    }
```
that will take 5 bytes in memory instead of multiple dictionaries that are even dynamically constructed!

benaadams:
Switched to this branch during OldBodies; seem to have neither OldBodies syncing nor peer logs about connected peers anymore

image


Inbound traffic very high though

kamilchodola:
So seems like OldHeaders being included seems like crucial thing as this bottlenecks the early stage of sync on small chains (like joc with low peer count which is usually max 2/3 for loong time)

PR #7177, opened 2024-06-12
Replace `JwtSecurityToken` with `JsonWebToken`

rubo:
## Changes

- Replaced `JwtSecurityToken` with the new `JsonWebToken` that's up to 30% faster, as Microsoft [claims](https://github.com/dotnet/aspnetcore/issues/49469)
- Made the `JwtAuthentication.Authenticate()` method async because of the underlying handler async API, which in turn caused chained async changes in several places

- [x] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Sockets/IWebSocketsModule.cs
+++ b/src/Nethermind/Nethermind.Sockets/IWebSocketsModule.cs
@@ -10,7 +10,7 @@ namespace Nethermind.Sockets
     public interface IWebSocketsModule
     {
         string Name { get; }
-        Task<ISocketsClient> CreateClient(WebSocket webSocket, string client, HttpContext context);
+        ValueTask<ISocketsClient> CreateClient(WebSocket webSocket, string client, HttpContext context);
         void RemoveClient(string clientId);
         Task SendAsync(SocketsMessage message);
     }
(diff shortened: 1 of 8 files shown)


smartprogrammer93 (approved):

src/Nethermind/Nethermind.Sockets/IWebSocketsModule.cs
```diff
@@ -10,7 +10,7 @@ namespace Nethermind.Sockets
     public interface IWebSocketsModule
     {
         string Name { get; }
-        ISocketsClient CreateClient(WebSocket webSocket, string client, HttpContext context);
+        Task<ISocketsClient> CreateClient(WebSocket webSocket, string client, HttpContext context);
```
smartprogrammer93, line 13: I would use valuetask as we won't need the task later

smartprogrammer93 (approved):

src/Nethermind/Nethermind.Sockets/IWebSocketsModule.cs
LukaszRozmej: Or no task at all?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Sockets/IWebSocketsModule.cs
smartprogrammer93: One of the interface implementations added an await call. Thats why it was added

LukaszRozmej: ah and it is the main one

benaadams (approved):

PR #7188, opened 2024-06-16
Prewarm only on main thread

benaadams:
## Changes

- Prewarm caches only on main thread, not durig eth_simulate (not sure we need to to this though)

#### Requires testing

- [x] No


LukaszRozmej:
We don't need it as it should be controlled by existance of pre-warmer in block processor. We will want to pre-warm on multiple different occasions - for example block producing.

PR #7194, opened 2024-06-19
Throw on non-empty storage

flcl42:
Fixes [some hive tests](https://hivetests.ethdevops.io//suite.html?suiteid=1718646748-5944c27fc6e3800e29bc8bcb6203fca6.json&suitename=consensus#)

## Clarifications
The check is applied in two places: on top level - when we have a create contract transacation; in a CREATE/CREATE call. 
On top level we should not expect storqage to be non zero because selfdestructed contract - so we don't need to clean up
When CREATE* call is executed we return due to collision and do not reach post-selfdestruct clean up
We revert in any case

## Changes

- Throw on non-empty storage according to [EIP-7610](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7610.md)
- Save from metrics failures
- Update tests

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes, will add a bit later
- [ ] No


src/Nethermind/Ethereum.Blockchain.Block.Test/Pyspecs/EipTests.cs
```diff
@@ -1,24 +0,0 @@
...
-using System.Collections.Generic;
-using System.Threading.Tasks;
-using Ethereum.Test.Base;
-using NUnit.Framework;
-
-namespace Ethereum.Blockchain.Block.Test.Pyspecs;
-
-[TestFixture]
-[Parallelizable(ParallelScope.All)]
-public class EipTests : BlockchainTestBase
```
LukaszRozmej, line 13: why remove those 3 classes?

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -2410,7 +2404,7 @@ private EvmExceptionType InstructionSelfDestruct<TTracing>(EvmState vmState, ref
 
         bool accountExists = _state.AccountExists(contractAddress);
         if (accountExists && (_codeInfoRepository.GetCachedCodeInfo(_worldState, contractAddress, spec).MachineCode.Length != 0 ||
-                              _state.GetNonce(contractAddress) != 0))
+                              _state.GetNonce(contractAddress) != 0 || _worldState.GetStorageRoot(contractAddress) != Keccak.EmptyTreeHash))
```
LukaszRozmej, line 2407: looks like it is the same conditions here and in `PrepareAccountForContractDeployment`, can you make it into a method?

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -187,20 +187,29 @@ private static void UpdateMetrics(ExecutionOptions opts, UInt256 effectiveGasPri
...
-                float gasPrice = (float)((double)effectiveGasPrice / 1_000_000_000.0);
-                Metrics.MinGasPrice = Math.Min(gasPrice, Metrics.MinGasPrice);
-                Metrics.MaxGasPrice = Math.Max(gasPrice, Metrics.MaxGasPrice);
+                try
+                {
+                    float gasPrice = (float)((double)effectiveGasPrice / 1_000_000_000.0);
+                    if (float.IsInfinity(gasPrice))
+                    {
+                        return;
+                    }
```
LukaszRozmej, line 196: how can it be infinity?

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -187,20 +187,29 @@ private static void UpdateMetrics(ExecutionOptions opts, UInt256 effectiveGasPri
...
-                Metrics.BlockTransactions++;
+                    Metrics.AveGasPrice = (Metrics.AveGasPrice * Metrics.Transactions + gasPrice) / (Metrics.Transactions + 1);
+                    Metrics.EstMedianGasPrice += Metrics.AveGasPrice * 0.01f * float.Sign(gasPrice - Metrics.EstMedianGasPrice);
+                    Metrics.Transactions++;
+
+                    Metrics.BlockAveGasPrice = (Metrics.BlockAveGasPrice * Metrics.BlockTransactions + gasPrice) / (Metrics.BlockTransactions + 1);
+                    Metrics.BlockEstMedianGasPrice += Metrics.BlockAveGasPrice * 0.01f * float.Sign(gasPrice - Metrics.BlockEstMedianGasPrice);
+                    Metrics.BlockTransactions++;
+                }
+                catch { }
```
LukaszRozmej, line 212: Do we still expect other exceptions?

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
flcl42: Yep

flcl42:
I'm still not sure about following things: shall we have None [here](https://github.com/NethermindEth/nethermind/pull/7194/files#diff-9742ac85df13f5104064b9cf775ea804cc783bc16b9655d45c55774b13be19e6R2411), how shall we consider SELFDESTRUCT (we were cleaning storage for after it, might require cleaning more things, seems like)

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
LukaszRozmej: Can we just keep them as double then?

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
LukaszRozmej: can we avoid them in code?

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
benaadams: Cap effectiveGasPrice at `float.MaxValue`?

Is `3 x 10^38` or `300_000_000_000_000_000_000 Ether` which is way higher than max supply; so isn't providing additional info at higher values

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -187,7 +187,9 @@ private static void UpdateMetrics(ExecutionOptions opts, UInt256 effectiveGasPri
         {
             if (opts is ExecutionOptions.Commit or ExecutionOptions.None)
             {
-                float gasPrice = (float)((double)effectiveGasPrice / 1_000_000_000.0);
+                // log2(3e38) ~ 127.8, if effectiveGasPrice is more than 128bit long it may lead to an overflow
+                float gasPrice = effectiveGasPrice[2] != 0 || effectiveGasPrice[3] != 0 ? 3e29f : (float)((double)effectiveGasPrice / 1_000_000_000.0);
```
benaadams, line 191:
```suggestion
                float gasPrice = effectiveGasPrice[2] != 0 || effectiveGasPrice[3] != 0 ? float.MaxValue : (float)((double)effectiveGasPrice / 1_000_000_000.0);
```

flcl42: I've just divided `effectiveGasPrice` by 1_000_000_000 as if it'd be `float.MaxValue` / 1e9. Code below potentially has this number multiplied by tx count, may be good to have some room. Or maybe I'm overthinking

benaadams: Makes sense. Could even go with a lower value or early exit if price is that high.

Is just for logging range boundaries; and an outlier gas price larger than anyone could pay for a single unit of not sure is helpful info? 🤔

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -580,26 +582,14 @@ protected virtual void PayFees(Transaction tx, BlockHeader header, IReleaseSpec
...
+            if (WorldState.AccountExists(contractAddress) && contractAddress.IsNonZeroAccount(spec, _codeInfoRepository, WorldState))
             {
-                CodeInfo codeInfo = _codeInfoRepository.GetCachedCodeInfo(WorldState, contractAddress, spec);
-                bool codeIsNotEmpty = codeInfo.MachineCode.Length != 0;
-                bool accountNonceIsNotZero = WorldState.GetNonce(contractAddress) != 0;
-
-                // TODO: verify what should happen if code info is a precompile
-                // (but this would generally be a hash collision)
-                if (codeIsNotEmpty || accountNonceIsNotZero)
+                if (Logger.IsTrace)
```
LukaszRozmej, line 587: Collapse to single line?

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
LukaszRozmej, line 191: `(effectiveGasPrice[2] | effectiveGasPrice[3]) != 0` ?

LukaszRozmej (approved):

src/Nethermind/Ethereum.Test.Base/GeneralTestBase.cs
```diff
@@ -109,7 +109,7 @@ protected EthereumTestResult RunTest(GeneralStateTest test, ITxTracer txTracer)
             Stopwatch stopwatch = Stopwatch.StartNew();
             IReleaseSpec? spec = specProvider.GetSpec((ForkActivation)test.CurrentNumber);
 
-            if (spec is Cancun) KzgPolynomialCommitments.InitializeAsync();
+            if (spec is Cancun) KzgPolynomialCommitments.InitializeAsync().Wait();
```
LukaszRozmej, line 112: make the whole method async and update usages?

benaadams (approved):
Does it need to be gated behind an eip7610 chianspec flag?

Also has test submodule changes in PR, not sure if you want those

flcl42:
> Does it need to be gated behind an eip7610 chianspec flag?

rare case of retroactive application, starting from block 0

> Also has test submodule changes in PR, not sure if you want those

Yup, it's an Ethereum tests update.

benaadams:
> > Does it need to be gated behind an eip7610 chianspec flag?
> 
> Rare case of retroactive application, starting from block 0

Yeah but does it apply to every chain?

benaadams:
> > Does it need to be gated behind an eip7610 chianspec flag?
> 
> Rare case of retroactive application, starting from block 0

Yeah but does it apply to every chain?

Could it cause consensus issue if we include it in next release but other EL's haven't included it yet?

flcl42:
> > > Does it need to be gated behind an eip7610 chianspec flag?
> > 
> > 
> > Rare case of retroactive application, starting from block 0

> Yeah but does it apply to every chain?

Yes
 
> Could it cause consensus issue if we include it in next release but other EL's haven't included it yet?

Should not, the case covers theoretically possible collision, that are not expected to happen in practice. And some clients already added it the same way

And if some custom chain would try to make use of just storage accounts, we would be aware of that.

PR #7197, opened 2024-06-19
Fix by setting the cache to empty items

LukaszRozmej:
## Changes

- Fix by setting the cache to empty items

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
+++ b/src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
@@ -432,8 +432,7 @@ namespace Nethermind.State
             {
                 foreach (UInt256 storageSlot in values.Keys)
                 {
-                    ref var value = ref CollectionsMarshal.GetValueRefOrNullRef(values, storageSlot);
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
```diff
@@ -428,7 +428,14 @@ public override void ClearStorage(Address address)
             base.ClearStorage(address);
 
             // Bit heavy-handed, but we need to clear all the cache for that address
-            _blockCache.Remove(address);
+            if (_blockCache.TryGetValue(address, out Dictionary<UInt256, byte[]> values))
+            {
+                foreach (UInt256 storageSlot in values.Keys)
+                {
+                    ref var value = ref CollectionsMarshal.GetValueRefOrNullRef(values, storageSlot);
```
benaadams, line 435: Why `CollectionsMarshal` and not just `values[storageSlot] = StorageTree.EmptyBytes;`?

benaadams (approved):

PR #7203, opened 2024-06-20
Skip updating prewarm caches on main thread

benaadams:
## Changes

- Main processing thread should only read from prewarm cache, not spend extra work updating them

- [x] Optimization

#### Requires testing

- [x] No


LukaszRozmej:
Does it improve anything? Potentially very small chance of waiting on dictionary? Otherwise, it complicates the code and I don't like it.

benaadams:
> Does it improve anything? Potentially very small chance of waiting on dictionary?

TryGet doesn't use any locks; GetOrAdd does TryGet, then if that fails TryAdd which aquires lock (potentially contending with prewarm) and allocates memory.

However the MainProcessor will never read that value from the prewam cache ever again, so why do the extra work?

src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.cs
```diff
@@ -106,6 +106,9 @@ public Block[] Process(Hash256 newBranchStateRoot, List<Block> suggestedBlocks,
         bool notReadOnly = !options.ContainsFlag(ProcessingOptions.ReadOnlyChain);
         int blocksCount = suggestedBlocks.Count;
         Block[] processedBlocks = new Block[blocksCount];
+
+        // Main thread should only read from prewarm caches, not spend extra time updating them.
+        _stateProvider.ReadOnlyPreWarmCaches = BlockchainProcessor.IsMainProcessingThread;
```
LukaszRozmej, line 111: pass it with `_preBlockCache` in constructor. Would call it `populatePreBlockCache` instead.

LukaszRozmej (changes requested):

src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.cs
benaadams: Changed

src/Nethermind/Nethermind.State/WorldState.cs
```diff
@@ -54,17 +54,18 @@ public WorldState(ITrieStore trieStore, IKeyValueStore? codeDb, ILogManager? log
...
-            _stateProvider = new StateProvider(trieStore.GetTrieStore(null), codeDb, logManager, stateTree, PreBlockCaches?.StateCache);
-            _persistentStorageProvider = new PersistentStorageProvider(trieStore, _stateProvider, logManager, storageTreeFactory, PreBlockCaches?.StorageCache);
+            _stateProvider = new StateProvider(trieStore.GetTrieStore(null), codeDb, logManager, stateTree, PreBlockCaches?.StateCache, populatePreBlockCache);
+            _persistentStorageProvider = new PersistentStorageProvider(trieStore, _stateProvider, logManager, storageTreeFactory, PreBlockCaches?.StorageCache, populatePreBlockCache);
             _transientStorageProvider = new TransientStorageProvider(logManager);
         }
 
-        public WorldState(ITrieStore trieStore, IKeyValueStore? codeDb, ILogManager? logManager, PreBlockCaches? preBlockCaches)
-            : this(trieStore, codeDb, logManager, null, preBlockCaches: preBlockCaches)
+        public WorldState(ITrieStore trieStore, IKeyValueStore? codeDb, ILogManager? logManager, PreBlockCaches? preBlockCaches, bool populatePreBlockCache = true)
```
LukaszRozmej, line 67: wonder if it would be better to make it `false` as default and explicitly set true, but not sure

LukaszRozmej (approved):

src/Nethermind/Nethermind.State/WorldState.cs
benaadams: Too many locations; unless invert the meaning of the param.

Is only really really true if it also has the `preBlockCaches` not being null

PR #7206, opened 2024-06-20
Reduce allocations in metrics backed by Zero Contention Counter

benaadams:
While the ThreadLocal removes all contention; it does allocate a new List<T> every time the values are totalled.

image

## Changes

- Use reflection to iterate the values to total in same why that runtime constructs the `List<T>`; so no allocations to total
- Introduce new class `ZeroContentionCounter` to hide this complexity
- Runtime issue raised https://github.com/dotnet/runtime/issues/103684

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Core/Threading/ZeroContentionCounter.cs
+++ b/src/Nethermind/Nethermind.Core/Threading/ZeroContentionCounter.cs
@@ -1,0 +1,41 @@
+// SPDX-FileCopyrightText: 2024 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System.Reflection;
+using System.Threading;
+
+namespace Nethermind.Core.Threading;
+public class ZeroContentionCounter
+{
+    private ThreadLocal<BoxedLong> _threadLocal = new(() => new BoxedLong(), trackAllValues: true);
+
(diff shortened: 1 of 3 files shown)


LukaszRozmej (commented):
Isn't reflection slow? Can we do Reflection.Emit or something to make it fast?

Or use compiled delegates: https://stackoverflow.com/a/65140648

benaadams:
Changed to  typed compiled expression

src/Nethermind/Nethermind.Core/Threading/ZeroContentionCounter.cs
```diff
@@ -0,0 +1,94 @@
...
+            assignNextSlot,
+            loop,
+            Expression.Label(breakLabel, totalVar)
+        );
+
+        // Lambda
+        Expression<Func<ThreadLocal<BoxedLong>, long>> lambda =
+            Expression.Lambda<Func<ThreadLocal<BoxedLong>, long>>(block, threadLocalParam);
+
+        return lambda.Compile();
```
LukaszRozmej, line 82: Can you add equivalent pseudo-code for that in comment?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Core/Threading/ZeroContentionCounter.cs
benaadams: Added

PR #7212, opened 2024-06-22
Add consistent progress bars for syncing

benaadams:
## Changes

- Rather than just Snap having a progress bar; also have progress bar for Downloads, Old headers, Old Bodies and Old Receipts
- Split each char of the progress bar into 8 elements for higher precision in same chars
```'⡀', '⡄', '⡆', '⡇', '⣇', '⣧', '⣷', '⣿'```
- Output multi-line branch sync logs a max of every 10 sec
- If log to be output is same as previous log, skip it
- Only output one line for sync change (Move "Sync mode changed" to Debug)
- Bring block etc to first element on sync change log

Before
image

After
image

- [x] Other: Logging

#### Requires testing

- [x] No

### CODE REVIEW

File: src/Nethermind/Nethermind.Logging/Progress.cs [hunk 1]
```diff
@@ -0,0 +1,20 @@
+// SPDX-FileCopyrightText: 2024 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System;
+
+namespace Nethermind.Logging;
+public class Progress
```
LukaszRozmej, line 7: static

src/Nethermind/Nethermind.Logging/Progress.cs
```diff
@@ -0,0 +1,20 @@
...
+    private static char[] _progressChars = { ' ', '⡀', '⡄', '⡆', '⡇', '⣇', '⣧', '⣷', '⣿' };
+
+    public static string GetMeter(float value, int max, int width = 40)
+    {
+        width = Math.Max(4, width - 3);
+        float progressF = value / max * width;
+        int progress = (int)Math.Floor(progressF);
+        int progressChar = (int)((progressF - progress) * _progressChars.Length);
+
+        return $"[{new string('⣿', progress)}{_progressChars[progressChar]}{new string(' ', width - progress - 1)}]";
```
LukaszRozmej, line 18: wouldn't StringBuilder be cheaper?

src/Nethermind/Nethermind.Synchronization/Reporting/SyncReport.cs
```diff
@@ -290,37 +282,64 @@ private void WriteFullSyncReport()
...
             {
-                _logger.Info($"Old Headers  {Pad(FastBlocksHeaders.CurrentValue, _blockPaddingLength)} / {_paddedPivot} ({FastBlocksHeaders.CurrentValue / (float)(_fastBlocksPivotNumber + 1),8:P2}) | queue {Pad(HeadersInQueue.CurrentValue, SpeedPaddingLength)} | current {Pad(FastBlocksHeaders.CurrentPerSecond, SpeedPaddingLength)} Blk/s | total {Pad(FastBlocksHeaders.TotalPerSecond, SpeedPaddingLength)} Blk/s");
+                float percentage = Math.Clamp(FastBlocksHeaders.CurrentValue / (float)(_fastBlocksPivotNumber + 1), 0, 1);
+                string headersReport = $"Old Headers  {FastBlocksHeaders.CurrentValue,BlockPaddingLength:N0} / {_paddedPivot} ({percentage,8:P2}) {Progress.GetMeter(percentage, 1)} queue {HeadersInQueue.CurrentValue,QueuePaddingLength:N0} | current {FastBlocksHeaders.CurrentPerSecond,SpeedPaddingLength:N0} Blk/s";
+                // Compare ignoring speed change
+                if (!_lastHeadersReport.AsSpan().StartsWith(headersReport.AsSpan(0, headersReport.LastIndexOf("| current"))))
+                {
+                    _lastHeadersReport = headersReport;
+                    _logger.Info(headersReport);
+                }
```
LukaszRozmej, line 307: make this local method? Used 3x

src/Nethermind/Nethermind.Synchronization/Reporting/SyncReport.cs
```diff
@@ -290,37 +282,64 @@ private void WriteFullSyncReport()
...
         }
 
         private void WriteFastBlocksReport(SyncMode currentSyncMode)
         {
             if ((currentSyncMode & SyncMode.FastHeaders) == SyncMode.FastHeaders)
             {
-                _logger.Info($"Old Headers  {Pad(FastBlocksHeaders.CurrentValue, _blockPaddingLength)} / {_paddedPivot} ({FastBlocksHeaders.CurrentValue / (float)(_fastBlocksPivotNumber + 1),8:P2}) | queue {Pad(HeadersInQueue.CurrentValue, SpeedPaddingLength)} | current {Pad(FastBlocksHeaders.CurrentPerSecond, SpeedPaddingLength)} Blk/s | total {Pad(FastBlocksHeaders.TotalPerSecond, SpeedPaddingLength)} Blk/s");
+                float percentage = Math.Clamp(FastBlocksHeaders.CurrentValue / (float)(_fastBlocksPivotNumber + 1), 0, 1);
+                string headersReport = $"Old Headers  {FastBlocksHeaders.CurrentValue,BlockPaddingLength:N0} / {_paddedPivot} ({percentage,8:P2}) {Progress.GetMeter(percentage, 1)} queue {HeadersInQueue.CurrentValue,QueuePaddingLength:N0} | current {FastBlocksHeaders.CurrentPerSecond,SpeedPaddingLength:N0} Blk/s";
+                // Compare ignoring speed change
```
LukaszRozmej, line 302: This repeats too

benaadams:
> I wonder if we started using some Pool of StringBuilders instead of string interpolation, would we reduce memory allocations significantly?

C#10 made string interpolation good https://devblogs.microsoft.com/dotnet/string-interpolation-in-c-10-and-net-6/

src/Nethermind/Nethermind.Synchronization/Reporting/SyncReport.cs
benaadams: Changed to method

src/Nethermind/Nethermind.Synchronization/Reporting/SyncReport.cs
benaadams: Changed to method

src/Nethermind/Nethermind.Logging/Progress.cs
benaadams: Made it string.Concat with some pre-made strings

src/Nethermind/Nethermind.Logging/Progress.cs
benaadams: Done

LukaszRozmej (approved):

PR #7213, opened 2024-06-23
Option to run Discovery V4 and V5 on the same port

alexb5dh:
- Versions to use (v4, v5, both) are now configurable, but same port is used when multiple are enabled. In such case same UDP connection (Netty `IDatagramChannel`) is used.
- Channel is injected into Lantern.Discv5 implementation via adapter for its `IUdpConnection`.
- Connection-management code moved from `DiscoveryApp` to a new `DiscoveryConnectionsPool`.
- Discovery initialization code moved from `InitializeNetwork` to a new `CompositeDiscoveryApp`.
- When both versions are enabled, all incoming messages are received by v4 handler first, which than forwards it to v5 if it doesn't manage to recognize the format. Otherwise, only enabled handler is being used.

~~Depends on https://github.com/Pier-Two/Lantern.Discv5/pull/105 being published on Nuget~~


src/Nethermind/Nethermind.Init/Steps/InitializeNetwork.cs
```diff
@@ -307,20 +339,25 @@ private void InitDiscovery()
         IDiscoveryConfig discoveryConfig = _api.Config<IDiscoveryConfig>();
 
         SameKeyGenerator privateKeyProvider = new(_api.NodeKey.Unprotect());
-        NodeIdResolver nodeIdResolver = new(_api.EthereumEcdsa);
 
-        if (discoveryConfig.Discv5Enabled)
+        _api.DiscoveryConnections = new DiscoveryConnectionsPool(_logger, _networkConfig, discoveryConfig);
+
+        // TODO should port checks or separate bool values be used here?
```
alexb5dh, line 345: This needs to be discussed

src/Nethermind/Nethermind.Network.Discovery/Discv5/DiscoveryV5App.cs
```diff
@@ -90,18 +92,20 @@ .. _discoveryDb.GetAllValues().Select(enr => enrFactory.CreateFromBytes(enr, ide
...
             })
             .WithSessionOptions(sessionOptions)
             .WithTableOptions(new TableOptions(bootstrapEnrs.Select(enr => enr.ToString()).ToArray()))
             .WithEnrBuilder(enrBuilder)
-            .WithLoggerFactory(new NethermindLoggerFactory(logManager, true))
-            .Build();
-
+            .WithLoggerFactory(new NethermindLoggerFactory(logManager, true));
 
+        discv5Builder.Build(); // Force-add default services
```
alexb5dh, line 105: This flow can be simplified once Lantern.Discv5 allows to override internal services.
I'll add a simple PR to their repository, once this implementation is agreed upon.

src/Nethermind/Nethermind.Init/Steps/InitializeNetwork.cs
alexb5dh: https://github.com/NethermindEth/nethermind/pull/7213/files#diff-bd19098570ab009a94a742e85abdcecd40471f86ef75fda621acbef5408b1bc0R353

alexb5dh: Question is - how should we specify these options in config:
- V4 enabled/disabled
- V4 UDP port
- V5 enabled/disabled
- V5 UDP port

For now I'm using `0` or negative port values as a "protocol disabled" configuration.

alexb5dh: V4 uses `DiscoveryEnabled` from `IInitConfig`. Should we move v5 there also?

flcl42: `DiscoveryEnabled` is whether discovery runs at all, and IDiscoveryConfig may tell which version is used. Hm maybe just add `Version` = "discv4/discv5/both" or something like that.

And yeah it might be better instead of Init.DiscoveryEnabled to have Discovery.Enabled, but it is how it is. We can make any flag obsolete but no sure if it's worth it

src/Nethermind/Nethermind.Network.Discovery/DiscoveryConfig.cs
```diff
@@ -43,5 +43,5 @@ public class DiscoveryConfig : IDiscoveryConfig
 
     public string Bootnodes { get; set; } = string.Empty;
 
-    public bool Discv5Enabled { get; set; } = false;
+    public DiscoveryVersion DiscoveryVersion { get; set; } = DiscoveryVersion.All;
```
flcl42, line 46: Should be V4 by default for now

src/Nethermind/Nethermind.Network.Discovery/CompositeDiscoveryApp.cs
```diff
@@ -0,0 +1,219 @@
...
+    public Task StopAsync() => Task.WhenAll(
+        _connections.StopAsync(),
+        _v4?.StopAsync() ?? Task.CompletedTask,
+        _v5?.StopAsync() ?? Task.CompletedTask
+    );
+
+    public void AddNodeToDiscovery(Node node)
+    {
+        // Add to v4 only
+        _v4?.AddNodeToDiscovery(node);
```
alexb5dh, line 122: Don't think v5 need to participate in this as of now, method implementation is actually empty there. Used by:
- `ProtocolsManager.AddNodeToDiscovery`
- `NodeSourceToDiscV4Feeder.AddToDiscoveryApp`

alexb5dh: Decided to add node to both v4 and v5. Will implement `AddNodeToDiscovery` in `DiscoveryV5App`.

src/Nethermind/Nethermind.Network.Discovery/Discv5/DiscoveryV5App.cs
```diff
@@ -57,50 +60,44 @@ public DiscoveryV5App(SameKeyGenerator privateKeyProvider, IApiWithNetwork api,
...
+           .AddSingleton(_sessionOptions.Signer);
 
         EnrFactory enrFactory = new(new EnrEntryRegistry());
 
         Lantern.Discv5.Enr.Enr[] bootstrapEnrs = [
             .. bootstrapNodes.Where(e => e.StartsWith("enode:"))
                 .Select(e => new Enode(e))
-                .Select(e => new EnrBuilder()
-                    .WithIdentityScheme(sessionOptions.Verifier, sessionOptions.Signer)
-                    .WithEntry(EnrEntryKey.Id, new EntryId("v4"))
```
alexb5dh, line 70: @flcl42, why have you decided to use `"v4"` here?
- If it's a node id - shouldn't it be unique for each one?
- If it's a protocol-id - I think it should be `"discv5"`, according to the [doc](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-wire.md).

flcl42: It's identity scheme for enr, not for discv5 packets. And for discv5 it's [v4](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-wire.md#handshake-message-packet-flag--2), which also mentioned [here](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-wire.md#handshake-message-packet-flag--2) (=enr has v4 scheme)

src/Nethermind/Nethermind.Network.Discovery.Test/NettyDiscoveryHandlerTests.cs
```diff
@@ -171,6 +174,33 @@ public async Task NeighborsSentReceivedTest()
             AssertMetrics(210);
         }
 
+        [Test]
+        [TestCase("010203")]
```
flcl42, line 178: Can be just TestCase instead of two attrs. Could you confirm it handles empty, and big packets well, please?

flcl42 (approved):

src/Nethermind/Nethermind.Network.Discovery.Test/NettyDiscoveryHandlerTests.cs
alexb5dh: Regarding other cases. This test should actually just test that v4 handler forwards messages to v5 handler if one doesn't match the v4 format. It contained a placeholder code I forgot to replace. Fixed it, but I don't see a need for this test specifically to cover different messages sizes, as internal forwarding logic doesn't depend on it.

flcl42: Seemed to me it might reveal potential issues on receiver(v4) side

alexb5dh: Added validation on packet sizes for both handlers

src/Nethermind/Nethermind.Network.Discovery/Discv5/NettyDiscoveryV5Handler.cs
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    public IAsyncEnumerable<UdpReceiveResult> ReadMessagesAsync(CancellationToken token = default) =>
+        _inboundQueue.Reader.ReadAllAsync(token);
+
+    public Task ListenAsync(CancellationToken token = default) => Task.CompletedTask;
+    public void Close() => _inboundQueue.Writer.Complete();
+
+    public static void Register(IServiceCollection services) => services
+        .AddSingleton<NettyDiscoveryV5Handler>()
```
asdacap, line 65: NP: Why not just a single `.AddSingleton<IUdpConnection, NettyDiscoveryV5Handler>`?

src/Nethermind/Nethermind.Network.Discovery/CompositeDiscoveryApp.cs
```diff
@@ -0,0 +1,222 @@
...
+    private readonly ILogger _logger;
+    private readonly SameKeyGenerator _privateKeyProvider;
+    private readonly INetworkConfig _networkConfig;
+    private readonly IDiscoveryConfig _discoveryConfig;
+    private IConnectionsPool _connections { get; set; }
+
+    private IDiscoveryApp? _v4;
+    private IDiscoveryApp? _v5;
+
+    public CompositeDiscoveryApp(IApiWithNetwork api, ILogger logger, SameKeyGenerator privateKeyProvider)
```
asdacap, line 41: NP: Pass individual component, not IApiWithNetwork.

src/Nethermind/Nethermind.Network.Discovery/Discv5/DiscoveryV5App.cs
```diff
@@ -33,11 +34,13 @@ public class DiscoveryV5App : IDiscoveryApp
     private readonly IApiWithNetwork _api;
     private readonly Logging.ILogger _logger;
     private readonly IDiscoveryConfig _discoveryConfig;
-    private readonly SimpleFilePublicKeyDb _discoveryDb;
+    private readonly IFullDb _discoveryDb;
```
asdacap, line 37: NP: Why not IKeyValueStore? missing some method?

src/Nethermind/Nethermind.Network.Discovery/Discv5/NettyDiscoveryV5Handler.cs
```diff
@@ -0,0 +1,67 @@
...
+{
+    private readonly ILogger _logger;
+    private readonly Channel<UdpReceiveResult> _inboundQueue;
+
+    private IChannel? _nettyChannel;
+
+    public NettyDiscoveryV5Handler(ILogManager loggerManager) : base(loggerManager)
+    {
+        _logger = loggerManager.GetClassLogger<NettyDiscoveryV5Handler>();
+        _inboundQueue = Channel.CreateUnbounded<UdpReceiveResult>();
```
asdacap, line 30: NP: Might cause memory leak in case reader is too slow or hang or crash. Probably set a somewhat high bound. The writer already use TryWrite, so thats fine.

asdacap (approved):

src/Nethermind/Nethermind.Network.Discovery/Discv5/NettyDiscoveryV5Handler.cs
alexb5dh: `DiscoveryV5App.InitializeChannel` uses it to resolve `NettyDiscoveryV5Handler` from services directly:
```csharp
var handler = _serviceProvider.GetRequiredService<NettyDiscoveryV5Handler>();
```

Initial idea was that other internal services may be fetched from `_serviceProvider` like this, but looks like this won't be needed. Can simplify this registration and either just save `NettyDiscoveryV5Handler` instance or replace code above with:
```csharp
var handler = (NettyDiscoveryV5Handler)_serviceProvider.GetRequiredService<IUdpConnection>();
```

src/Nethermind/Nethermind.Network.Discovery/Discv5/DiscoveryV5App.cs
alexb5dh: Some of the methods are missing from `IKeyValueStore`, `GetAllValues` being an example. Replaced it with `IDb` instead.

PR #7216, opened 2024-06-23
Skip eth_getLogs error for NonValidatorNodes

benaadams:
## Changes

- Don't output `Error when handling Id:N, eth_getLogs` for NonValidatorNode as is likely is because they aren't syncing receipts

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
+++ b/src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
@@ -14,6 +14,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Connections;
 using Microsoft.AspNetCore.Http;
+using Nethermind.Blockchain.Synchronization;
 using Nethermind.Core.Collections;
 using Nethermind.Core.Extensions;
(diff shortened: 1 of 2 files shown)


LukaszRozmej (changes requested):
This is not the place to do it. `ResultWrapper` has `IsTemporary` (Might be renamed?) which would translate into `SuppressWarning` thus we could use it directly from `eth_getLogs`

benaadams:
> This is not the place to do it. `ResultWrapper` has `IsTemporary` (Might be renamed?) which would translate into `SuppressWarning` thus we could use it directly from `eth_getLogs`

Not sure I understand; pass `_syncConfig.NonValidatorNode` into `eth_getLogs` ?

LukaszRozmej:
> > This is not the place to do it. `ResultWrapper` has `IsTemporary` (Might be renamed?) which would translate into `SuppressWarning` thus we could use it directly from `eth_getLogs`
> 
> Not sure I understand; pass `_syncConfig.NonValidatorNode` into `eth_getLogs` ?

I would pass it into `EthRpcModule`

benaadams:
This will be resolved post Pectra; don't think is needed anymore

Issue #7171, opened 2024-06-12
1.27rc slow down after half a day


xanatos:
Purely anecdotal. I've been trying the 1.27rc on an old machine I have together with the new Lighthouse 5.2.0 . I've rebuilt both dbs from scratch using the 1.26 . Then I switched to the 1.27rc. There are no running validators or other programs (but I use the machine for websurfing).

Note that the machine is "old" (Intel i5-8500, 32gb ram, Crucial P3 2TB m.2), AND I'm using Windows 11 23H2.

The 1.27rc is FAST. The MGas/s are very high (compared to my main validator). But tonight at around 3:28 in the morning it slowed down and it remained slow. I had to relaunch it this morning to have it return to its "best" values. In the EventViewer there was nothing "strange".

I've exctracted from the log files the Mgas/s and plotted them. Yellow part is the slow down. Green part is after relaunching.

![Screenshot 2024-06-12 095348](https://github.com/NethermindEth/nethermind/assets/821344/29b66791-c0ea-4bc9-b284-9ec339709166)

I'm adding some screenshots made with the ProcessExplorer (one yesterday evening, one this morning) to compare the use of resources.

Yesterday:

![Screenshot 2024-06-11 195441](https://github.com/NethermindEth/nethermind/assets/821344/639f4930-7224-4690-8e44-b0b914140be2)
![Screenshot 2024-06-11 195648](https://github.com/NethermindEth/nethermind/assets/821344/54e8addd-cde0-4f7f-9801-bb9d0cdfd7a6)

This morning:

![Screenshot 2024-06-12 082954](https://github.com/NethermindEth/nethermind/assets/821344/72a3a2f1-83e3-4f72-9c2c-80d8ba2c08aa)
![Screenshot 2024-06-12 083002](https://github.com/NethermindEth/nethermind/assets/821344/19fee36b-d7b9-4170-81a2-24ea88fd84bc)

And I'm including a zip file with the logs.

[mainnet.logs.zip](https://github.com/user-attachments/files/15800704/mainnet.logs.zip)

As I've said, this could be totally anecdotal. If in the next 24h it will repeat then I'll add a comment.

xanatos:
While not in a so evident way, the moving average of the MGas/s seems to be still going down.

LukaszRozmej:
[...]

PR #7218, opened 2024-06-24
Mcslock reduce threadlocals

benaadams:
Contributes to  #7171

## Changes

- Reduce the number of ThreadLocals in use (seems to be issue on Windows over time and lots end up on finalization queue)
- Also add cache padding to the nodes

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Core.Test/MCSLockTests.cs
+++ b/src/Nethermind/Nethermind.Core.Test/MCSLockTests.cs
@@ -88,27 +88,4 @@ public class MCSLockTests
         var expectedOrder = Enumerable.Range(0, numberOfThreads).ToList();
         CollectionAssert.AreEqual(expectedOrder, executionOrder, "Threads did not acquire lock in the order they were started.");
     }
-
-    [Test]
-    public void NonReentrantTest()
-    {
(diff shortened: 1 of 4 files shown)


src/Nethermind/Nethermind.Core.Test/MCSLockTests.cs
```diff
@@ -88,27 +88,4 @@ public void LockFairnessTest()
         var expectedOrder = Enumerable.Range(0, numberOfThreads).ToList();
         CollectionAssert.AreEqual(expectedOrder, executionOrder, "Threads did not acquire lock in the order they were started.");
     }
-
-    [Test]
-    public void NonReentrantTest()
```
LukaszRozmej, line 93: why remove?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Core.Test/MCSLockTests.cs
benaadams: Hard to test in simlar way. Current uses aren't reentrant, will keep eye out on future uses I guess 🤔

PR #7241, opened 2024-07-03
feat: add CLI tool for transition to Verkle trees

### LINKED ISSUE #6695: add a cli tool to convert a synced merkle tree to verkle tree
We need to implement a tool that can convert the merkle tree of the synced state to a verkle tree. We need this to be able to test and get estimates for post-verkle mainnet.

marcuspang:
Fixes #6695

## Changes

- Add a new project to instantiate a Verkle tree with the current database, which is specified in the `.cfg` file
  - This is done by calling `dotnet run` in `src/Nethermind/Nethermind.VerkleMigration.Cli`
  - Unit test for migration logic can be found in `src/Nethermind/Nethermind.VerkleMigration.Cli.Test`
- As the preimage DB has not been implemented yet, the `VerkleTreeMigrator` will always return the first 20 bytes when retrieving the preimage of keccak(address) bytes (this is derived from the current path of the tree visitor)

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

Currently only unit tests are written. There are no benchmark tests written as of yet, let me know if this is necessary in this PR.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks
Should the output be dumped in a text file? Or should we simply make sure it's stored in the config file's verkle DB?


src/Nethermind/Nethermind.State/VerkleStateTree.cs
```diff
@@ -70,6 +71,36 @@ public void SetCode(Address address, byte[] code)
...
+        // Update code size
+        AccountStruct? account = Get(address);
+        if (account.HasValue)
+        {
+            var newAccount = new Account(account.Value.Nonce, account.Value.Balance, (UInt256)code.Length, account.Value.Version, new Hash256(account.Value.StorageRoot), Keccak.Compute(code));
+            Set(address, newAccount);
+        }
+    }
+
+    public byte[] GetCode(Address address, Hash256? stateRoot = null)
```
tanishqjasoria, line 84: mark this as internal - as we want to use this just for test

src/Nethermind/Nethermind.State/VerkleStateTree.cs
```diff
@@ -70,6 +71,36 @@ public void SetCode(Address address, byte[] code)
...
             chunkId += 1;
         }
+
+        // Update code size
+        AccountStruct? account = Get(address);
+        if (account.HasValue)
+        {
+            var newAccount = new Account(account.Value.Nonce, account.Value.Balance, (UInt256)code.Length, account.Value.Version, new Hash256(account.Value.StorageRoot), Keccak.Compute(code));
+            Set(address, newAccount);
+        }
```
tanishqjasoria, line 81: dont add this here - the way this is used in client - we update the code size and then call SetCode()
so you should move it outside

src/Nethermind/Nethermind.State/VerkleTransition/AccountTreeMigrator.cs
```diff
@@ -0,0 +1,229 @@
...
+            {
+                addressHash = ConstructFullHash(node.Key, _currentPath);
+            }
+            byte[]? addressBytes = _preImageDb.Get(addressHash);
+            if (addressBytes is not null)
+            {
+                var address = new Address(addressBytes);
+                MigrateAccount(address, account);
+
+                if (account.IsContract)
```
tanishqjasoria, line 162: use account.HasCode instead
right now both means the same, but in future EOA can also have code

src/Nethermind/Nethermind.State/VerkleTransition/AccountTreeMigrator.cs
```diff
@@ -0,0 +1,229 @@
...
+            }
+            else
+            {
+                addressHash = ConstructFullHash(node.Key, _currentPath);
+            }
+            byte[]? addressBytes = _preImageDb.Get(addressHash);
+            if (addressBytes is not null)
+            {
+                var address = new Address(addressBytes);
+                MigrateAccount(address, account);
```
tanishqjasoria, line 160: here just check if the account has code, then get the codeSize and update the account with the new codeSize

src/Nethermind/Nethermind.State/VerkleTransition/AccountTreeMigrator.cs
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    public void VisitLeaf(TrieNode node, TrieVisitContext trieVisitContext, ReadOnlySpan<byte> value)
+    {
+        List<byte[]> currentPath = trieVisitContext.IsStorage ? _currentPathStorage : _currentPath;
+
+        while (currentPath.Count >= trieVisitContext.Level && trieVisitContext.Level > 0)
+        {
+            currentPath.RemoveAt(currentPath.Count - 1);
+        }
```
tanishqjasoria, line 140: i cant see this being used anywhere

tanishqjasoria (changes requested):

src/Nethermind/Nethermind.State/VerkleTransition/AccountTreeMigrator.cs
marcuspang: Currently still facing an edge case for large trees, where some of the leaves do not generate the full keccak hash based on my path tracking logic. Will need to spend more time debugging this

src/Nethermind/Nethermind.State/VerkleStateTree.cs
```diff
@@ -72,6 +73,28 @@ public void SetCode(Address address, byte[] code)
         }
     }
 
+    public byte[] GetCode(Address address, Hash256? stateRoot = null)
```
marcuspang, line 76: Made this `public` so that I could move the migration logic to a separate project (`Nethermind.VerkleMigration.Cli`)

PR #7244, opened 2024-07-05
Inform test repo on push to master/release branches

kamilchodola:
Whenever docker image is created for master or release inform test repo, that there is new image to be used in tests (will be used to control if master tests should be triggered or more detailed release tests should be in place)

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [X] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [X] No

- [ ] Yes
- [X] No

### DIFF
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -59,3 +59,16 @@ jobs:
             --build-arg CI=$CI \
             --build-arg COMMIT_HASH=$GITHUB_SHA \
             . --push
+            
+  upgrade_node:
+    name: Update 24/7 node
+    runs-on: [sync-agent-1280gb]
+    needs: [publish-docker]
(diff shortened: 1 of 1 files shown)


rubo (approved):

rubo:
If this will be end up in the main branch, maybe it's better to have this as a separate workflow that depends on docker publisher.

kamilchodola:
@rubo Actually was thinking about it - just the thing is that as you see this is supposed to be very minimal job - stop, pull and run - that is why I thought about it in this place BUT will move it to our testing repo and trigger it from there and here I'd just call other repo on success.

So please don;t merge - to be adjusted.

kamilchodola:
@rubo @brbrr please re-review once more - change approach and node related stuff will be done in different repo.

PR #7252, opened 2024-07-09
Feature/geth like system tx

LukaszRozmej:
## Changes

- Decouples system transaction from normal transaction processing
- Makes system transaction support Aura and non-Aura versions
- Moves `BeaconBlockRootHandler` to use system transactions

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

Sync some Aura chains in archive.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -456,10 +480,7 @@ protected void ExecuteEvmCall(
 
             Snapshot snapshot = WorldState.TakeSnapshot();
 
-            // Fixes eth_estimateGas.
-            // If sender is SystemUser subtracting value will cause InsufficientBalanceException
-            if (validate || !tx.IsSystem())
-                WorldState.SubtractFromBalance(tx.SenderAddress, tx.Value, spec);
```
LukaszRozmej, line 462: `OriginalValidate` was introduced due to this condition. For every other condition validation off is equivalent with transaction being system transaction, but for this one condition we need to decouple this logic, and even if it is system transaction validation can be on/off independently, this is used in `eth_estimateGas`

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -82,31 +91,41 @@ protected enum ExecutionOptions
...
 
         public TransactionResult Trace(Transaction transaction, in BlockExecutionContext blCtx, ITxTracer txTracer) =>
-            Execute(transaction, in blCtx, txTracer, ExecutionOptions.NoValidation);
+            ExecuteCore(transaction, in blCtx, txTracer, ExecutionOptions.NoValidation);
+
+        private TransactionResult ExecuteCore(Transaction tx, in BlockExecutionContext blCtx, ITxTracer tracer, ExecutionOptions opts)
+        {
+            if (tx.IsSystem())
+            {
+                _systemTransactionProcessor ??= new SystemTransactionProcessor(SpecProvider, WorldState, VirtualMachine, _codeInfoRepository, _logManager);
```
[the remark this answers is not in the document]
LukaszRozmej: It is `??=` so it will be created lazy, but only once.

src/Nethermind/Nethermind.Evm/TransactionProcessing/SystemTransactionProcessor.cs
```diff
@@ -0,0 +1,64 @@
...
+
+public sealed class SystemTransactionProcessor : TransactionProcessorBase
+{
+    private readonly bool _isAura;
+
+    /// <summary>
+    /// Hacky flag to execution options, to pass information how original validate should behave.
+    /// Needed to decide if we need to subtract transaction value.
+    /// </summary>
+    private const int OriginalValidate = 2 << 30;
```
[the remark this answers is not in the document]
LukaszRozmej: estimateGas

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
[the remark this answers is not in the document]
LukaszRozmej: So creation or not is outside of SystemTransactionProcessor. If it is not created it won't be - as system address is never touched due to forced NoValidation mode. If it is created than in AuRa the `SystemTransactionReleaseSpec` will guarantee that EIP-158 is disabled so it won't be deleted.

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -454,10 +478,7 @@ protected virtual TransactionResult IncrementNonce(Transaction tx, BlockHeader h
 
             Snapshot snapshot = WorldState.TakeSnapshot();
 
-            // Fixes eth_estimateGas.
-            // If sender is SystemUser subtracting value will cause InsufficientBalanceException
-            if (validate || !tx.IsSystem())
-                WorldState.SubtractFromBalance(tx.SenderAddress, tx.Value, spec);
+            PayValue(tx, spec, opts);
```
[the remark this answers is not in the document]
LukaszRozmej: No, as we force NoValidation

src/Nethermind/Nethermind.Evm/TransactionProcessing/SystemTransactionProcessor.cs
```diff
@@ -0,0 +1,64 @@
...
+        ICodeInfoRepository? codeInfoRepository,
+        ILogManager? logManager)
+        : base(specProvider, worldState, virtualMachine, codeInfoRepository, logManager)
+    {
+        _isAura = SpecProvider.SealEngine == SealEngineType.AuRa;
+    }
+
+    protected override TransactionResult Execute(Transaction tx, in BlockExecutionContext blCtx, ITxTracer tracer, ExecutionOptions opts) =>
+        base.Execute(tx, in blCtx, tracer, !opts.HasFlag(ExecutionOptions.NoValidation)
+            ? opts | (ExecutionOptions)OriginalValidate | ExecutionOptions.NoValidation
```
LukaszRozmej, line 36: We force NoValidation mode here, and save original value under `OriginalValidate` flag for estimate gas edge case.

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
MarekM25: Ok, thanks. I understand the logic. On the other hand, this logic assumes that there are sysCalls that pay values. Based on my quick check, that’s not the case. In sysCalls, we set Value = 0. This means we could simplify the original check and remove that hacky edge case flag.

So original logic could be:
`if (validate && !tx.IsSystem())`

MarekM25:
@kamilchodola this PR will need a special attention from QA team (pls don't start now, wait for green-light). We will have to test AuRa sys calls.

Recommended things:
- resync our gnosis archive
- sync Energyweb/volta archive/non-archive
- sync Chiado
- perhaps, update validators (but could be done with normal cycle)

If we have any working Posdao tests than we could use it, but I guess it's not possible

benaadams:
Aura failure looks legit (after a number of reruns)

>  Failed priority_should_return_correctly [77 ms]
  Error Message:
   Expected root to be a collection with 3 item(s), but {0x942921b14f1b1c385cd7e0cc2ef7abe5598c8358.00000000=4@Contract.2, 0x942921b14f1b1c385cd7e0cc2ef7abe5598c8358.00010203=3@Contract.2}
contains 1 item(s) less than
{0x942921b14f1b1c385cd7e0cc2ef7abe5598c8358.00010203=3@Contract.2, 0xb7705ae4c6f81b66cdb323c65f4e8133690fc099.00000000=1@Contract.1, 0x942921b14f1b1c385cd7e0cc2ef7abe5598c8358.00000000=4@Contract.1}.

MarekM25 (approved):

src/Nethermind/Nethermind.Consensus/Processing/TransactionProcessorAdapterExtensions.cs
```diff
@@ -18,7 +18,7 @@ public static TransactionResult ProcessTransaction(this ITransactionProcessorAda
         ProcessingOptions processingOptions,
         IWorldState stateProvider)
     {
-        if (processingOptions.ContainsFlag(ProcessingOptions.DoNotVerifyNonce))
+        if (processingOptions.ContainsFlag(ProcessingOptions.DoNotVerifyNonce) && currentTx.SenderAddress != Address.SystemUser)
```
tanishqjasoria, line 21: should we add a flag in the tx to check if it is a systemTx?

tanishqjasoria (commented):
I still do not understand the use of OriginalValidate, is it supposed to be for a case in future where we transfer value in SystemTransaction? because AFAIK we dont do that right now?

src/Nethermind/Nethermind.Blockchain/BeaconBlockRoot/BeaconBlockRootHandler.cs
```diff
@@ -1,39 +1,48 @@
...
 {
-    public void ApplyContractStateChanges(Block block, IReleaseSpec spec, IWorldState stateProvider)
-    {
-        if (!spec.IsBeaconBlockRootAvailable ||
-            block.IsGenesis ||
-            block.Header.ParentBeaconBlockRoot is null)
-            return;
-
-        Address eip4788Account = spec.Eip4788ContractAddress ?? Eip4788Constants.BeaconRootsAddress;
+    private const long GasLimit = 30_000_000L;
```
benaadams, line 15: Should this be from chainspec, why this value?

benaadams (approved):

src/Nethermind/Nethermind.Blockchain/BeaconBlockRoot/BeaconBlockRootHandler.cs
LukaszRozmej: directly from EIP: https://eips.ethereum.org/EIPS/eip-4788

src/Nethermind/Nethermind.Blockchain/BeaconBlockRoot/BeaconBlockRootHandler.cs
```diff
@@ -1,39 +1,48 @@
...
+                SenderAddress = Address.SystemUser,
+                GasLimit = GasLimit,
+                GasPrice = UInt256.Zero,
+            };
+
+            transaction.Hash = transaction.CalculateHash();
+
+            try
+            {
+                processor.Execute(transaction, header, NullTxTracer.Instance);
```
flcl42, line 40: Do we add Eip4788ContractAddress to access list like geth does? If the call fails should the address stay in the list if so?
Do we need to trace system calls in some additional way?
Is the result the same in case of failure like for other implementations?

src/Nethermind/Nethermind.Specs/SystemTransactionReleaseSpec.cs
```diff
@@ -7,133 +7,8 @@
...
-        public bool IsEip6780Enabled => _spec.IsEip6780Enabled;
-        public bool IsEip4788Enabled => _spec.IsEip4788Enabled;
-        public Address Eip4788ContractAddress => _spec.Eip4788ContractAddress;
-        public bool IsEip2935Enabled => _spec.IsEip2935Enabled;
-        public bool IsEip7709Enabled => _spec.IsEip7709Enabled;
-        public Address Eip2935ContractAddress => _spec.Eip2935ContractAddress;
-        public UInt256 ForkBaseFee => _spec.ForkBaseFee;
-        public UInt256 BaseFeeMaxChangeDenominator => _spec.BaseFeeMaxChangeDenominator;
-        public long ElasticityMultiplier => _spec.ElasticityMultiplier;
+        public override bool IsEip158Enabled => !isAura;
```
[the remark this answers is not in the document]
LukaszRozmej: ah good question, so it should be `!isAura && spec.IsEip158Enabled`?

src/Nethermind/Nethermind.Blockchain/BeaconBlockRoot/BeaconBlockRootHandler.cs
LukaszRozmej: 1. Not sure, but it works on mainnet, so I guess it is the same?
2. Tracing is outside of the scope of this PR. We might want to add special tracing method that will trace system transactions too.
3. Good question, failure should definitely fail the block, not sure about the reason.

flcl42: 1. Very theoretical, but it may cost less gas like 29_999_900 and pass but we will charge 30_000_100 and fail. And traces may differ I guess. Traces are important for L2 provers I guess.
2. Yep, I'm not sure at all what to expect for system tx in this regard
3. It should fail silently and proceed with regular txs

LukaszRozmej: Geth code: https://github.com/ethereum/go-ethereum/blob/master/core/state_processor.go#L180

LukaszRozmej: applied, we need to add tracing in separate PR

flcl42 (approved):

PR #7259, opened 2024-07-11
Heuristic tx censorship detection

Arindam2407:
## Changes

Censorship Detection Using Heuristics

- We maintain a cache [CensorshipDetector](https://github.com/Arindam2407/nethermind/blob/heuristic_tx_censorship_detection/src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector.cs) of size 4 which indicates whether newly cached blocks are potentially censoring transactions.
- On caching of a block, we check if the maximum gas price among transactions included in the block is lower than the maximum gas price in the tx pool.
- If so, we mark PotentialCensorship true and check if all the 4 latest cached blocks indicate PotentialCensorship to be true. 
- If so, ShouldOverrideBuilder becomes true.

## Requires testing
 - [x] Yes
 - [ ] No

## If yes, did you write tests?
- [x] Yes
- [ ] No


src/Nethermind/Nethermind.Blockchain/BlockTree.cs
```diff
@@ -47,6 +47,8 @@ public partial class BlockTree : IBlockTree
         private readonly IDb _metadataDb;
         private readonly IBlockStore _badBlockStore;
 
+        private readonly PotentialCensorshipCache _potentialCensorshipCache;
```
LukaszRozmej, line 50: I would prefer if this lived as a separate component, outside of BlockTree. It can react to `BlockAddedToMain` event.

src/Nethermind/Nethermind.Blockchain/Blocks/PotentialCensorshipCache.cs
```diff
@@ -0,0 +1,79 @@
...
+
+public class PotentialCensorshipCache
+{
+    private static PotentialCensorshipCache? _instance;
+
+    public static PotentialCensorshipCache Instance()
+    {
+        _instance ??= new PotentialCensorshipCache();
+        return _instance;
+    }
```
LukaszRozmej, line 20: Why singleton?

src/Nethermind/Nethermind.Blockchain/Blocks/PotentialCensorshipCache.cs
```diff
@@ -0,0 +1,79 @@
...
+    private const int CacheSize = 4;
+
+    private readonly LruCache<Hash256, TxCensorshipDetector>
+        _potentialCensorshipCache = new(CacheSize, CacheSize, "potentialCensorshipCache");
+
+    public struct TxCensorshipDetector
```
LukaszRozmej, line 26: `TxCensorshipInfo`?

src/Nethermind/Nethermind.Blockchain/Blocks/PotentialCensorshipCache.cs
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    public PotentialCensorshipCache() { }
+
+    public void Delete(Hash256 blockHash)
+    {
+        _potentialCensorshipCache.Delete(blockHash);
+    }
+
+    public void Cache(ref Block block)
```
LukaszRozmej, line 45: ref not needed

src/Nethermind/Nethermind.Blockchain/Blocks/PotentialCensorshipCache.cs
```diff
@@ -0,0 +1,79 @@
...
+    {
+        UInt256 maxGasPriceInBlock = 0;
+
+        for (int i = 0; i < block.Transactions.Length; i++)
+        {
+            maxGasPriceInBlock = maxGasPriceInBlock > block.Transactions[i].GasPrice ? maxGasPriceInBlock : block.Transactions[i].GasPrice;
```
LukaszRozmej, line 51:
```suggestion
            maxGasPriceInBlock = UInt256.Max(maxGasPriceInBlock, block.Transactions[i].GasPrice);
```

src/Nethermind/Nethermind.Blockchain/Blocks/PotentialCensorshipCache.cs
```diff
@@ -0,0 +1,79 @@
...
+        TxGasPriceSortedCollection.TxHashGasPricePair pair = TxGasPriceSortedCollection.Instance().GetFirstPair();
+
+        if (pair.GasPrice > maxGasPriceInBlock)
+        {
+            _potentialCensorshipCache.Set(block.Hash, new TxCensorshipDetector(true, pair.Hash));
+        }
+        else
+        {
+            _potentialCensorshipCache.Set(block.Hash, new TxCensorshipDetector(false, pair.Hash));
+        }
```
LukaszRozmej, line 63:
```suggestion
     _potentialCensorshipCache.Set(block.Hash, new TxCensorshipDetector(pair.GasPrice > maxGasPriceInBlock, pair.Hash));
```

src/Nethermind/Nethermind.Blockchain/Blocks/PotentialCensorshipCache.cs
```diff
@@ -0,0 +1,79 @@
...
+
+using Nethermind.Core;
+using Nethermind.Core.Caching;
+using Nethermind.Core.Crypto;
+using Nethermind.Int256;
+using Nethermind.TxPool.Collections;
+
+namespace Nethermind.Blockchain.Blocks;
+
+public class PotentialCensorshipCache
```
LukaszRozmej, line 12: `CensorshipDetector`?

src/Nethermind/Nethermind.Merge.Plugin/Data/GetPayloadV3Result.cs
```diff
@@ -18,7 +19,13 @@ public GetPayloadV3Result(Block block, UInt256 blockFees, BlobsBundleV1 blobsBun
 
     public override ExecutionPayloadV3 ExecutionPayload { get; }
 
-    public bool ShouldOverrideBuilder { get; }
+    public bool ShouldOverrideBuilder
+    {
+        get
+        {
+            return PotentialCensorshipCache.Instance().CensorshipDetected();
```
LukaszRozmej, line 26: avoid singletons

src/Nethermind/Nethermind.TxPool/Collections/TxGasPriceSortedCollectionEvents.cs
```diff
@@ -0,0 +1,40 @@
+// SPDX-FileCopyrightText: 2022 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System;
+using Nethermind.Core.Crypto;
+using Nethermind.Int256;
+
+namespace Nethermind.TxPool.Collections
+{
+    public partial class TxGasPriceSortedCollection
```
LukaszRozmej, line 10: partial classes are fine, suggest naming file: `TxGasPriceSortedCollection.Events.cs`

src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -69,6 +69,7 @@ public class TxPool : ITxPool, IDisposable
         private Transaction[]? _blobTransactionSnapshot;
         private long _lastBlockNumber = -1;
         private Hash256? _lastBlockHash;
+        public readonly TxGasPriceSortedCollection _txGasPriceSortedCollection;
```
LukaszRozmej, line 72: Would be good if it is its own component, independent of TxPool

src/Nethermind/Nethermind.Blockchain/Blocks/PotentialCensorshipCache.cs
```diff
@@ -0,0 +1,79 @@
...
+        for (int i = 0; i < block.Transactions.Length; i++)
+        {
+            maxGasPriceInBlock = maxGasPriceInBlock > block.Transactions[i].GasPrice ? maxGasPriceInBlock : block.Transactions[i].GasPrice;
+        }
+
+        TxGasPriceSortedCollection.TxHashGasPricePair pair = TxGasPriceSortedCollection.Instance().GetFirstPair();
+
+        if (pair.GasPrice > maxGasPriceInBlock)
+        {
+            _potentialCensorshipCache.Set(block.Hash, new TxCensorshipDetector(true, pair.Hash));
```
LukaszRozmej, line 58: We can have only 1 transaction censored per block? For the future where inclusion lists are a thing it would be good to gather all censored tx hashes.

src/Nethermind/Nethermind.Blockchain/Blocks/PotentialCensorshipCache.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+        else
+        {
+            _potentialCensorshipCache.Set(block.Hash, new TxCensorshipDetector(false, pair.Hash));
```
LukaszRozmej, line 62: does it make sense to set something when it is not censored?

src/Nethermind/Nethermind.Blockchain/Blocks/PotentialCensorshipCache.cs
```diff
@@ -0,0 +1,79 @@
...
+    public bool CensorshipDetected()
+    {
+        ValueHash256 toCompare = _potentialCensorshipCache.GetValues()[0].TxHash;
+        foreach (TxCensorshipDetector suspect in _potentialCensorshipCache.GetValues())
+        {
+            if (!suspect.PotentialCensorship || suspect.TxHash != toCompare)
+            {
+                return false;
+            }
+        }
```
LukaszRozmej, line 75: I don't get this logic, can you explain?

src/Nethermind/Nethermind.TxPool/Collections/TxGasPriceSortedCollection.cs
```diff
@@ -0,0 +1,191 @@
...
+    public TxGasPriceSortedCollection(int capacity = 16)
+    {
+        _capacity = capacity;
+        _cacheMap = [];
+        _sortedSet = new EnhancedSortedSet<TxHashGasPricePair>(new GasPriceComparer());
+    }
+
+    public int Count => _cacheMap.Count;
+
+    public bool TryRemove(ref Transaction tx, [NotNullWhen(true)] out TxHashGasPricePair? removed)
```
LukaszRozmej, line 64: ref not needed

src/Nethermind/Nethermind.TxPool/Collections/TxGasPriceSortedCollection.cs
```diff
@@ -0,0 +1,191 @@
...
+    {
+        using McsLock.Disposable lockRelease = Lock.Acquire();
+
+        if (tx is null)
+        {
+            removed = default;
+            return false;
+        }
+
+        ValueHash256 hash = tx.CalculateHash();
```
LukaszRozmej, line 74: why do we need to calculate hash here?

LukaszRozmej (commented):
Good first try and place to polish it a bit!

src/Nethermind/Nethermind.Blockchain/Blocks/PotentialCensorshipCache.cs
Arindam2407: We need singleton because we need to get the status of CensorshipDetected when sending the Payload, which is static.

src/Nethermind/Nethermind.Blockchain/Blocks/PotentialCensorshipCache.cs
Arindam2407: I found it convenient to detect censorship. Can be changed

src/Nethermind/Nethermind.Blockchain/Blocks/PotentialCensorshipCache.cs
Arindam2407: For the cached blocks, if potential censorship is false for any one, there is no censorship. On the other hand, if the highest tx in the TxGasPriceSortedCollection is not the same for all cached blocks, there is no censorship. Reason being, there might be a new tx added to the collection with a higher gas price than the block's maximum after the block was committed. It can happen continuously over a few blocks.

src/Nethermind/Nethermind.Blockchain/Blocks/PotentialCensorshipCache.cs
Arindam2407: It is a very simplistic check. If the highest tx (in terms of GasPrice) in the block is still lower than the highest tx in the collection, then there might be censorship. It is only being used to mark ShouldOverrideBuilder true or false. Since we are maintaining an ordered collection, we could expand it in the future to include all censored tx hashes.

src/Nethermind/Nethermind.Merge.Plugin.Test/EngineModuleTests.Setup.cs
```diff
@@ -46,9 +46,12 @@ namespace Nethermind.Merge.Plugin.Test;
 
 public partial class EngineModuleTests
 {
+    private INethermindApi? _api;
+
     [SetUp]
     public Task Setup()
     {
+        _api = Substitute.For<INethermindApi>();
```
LukaszRozmej, line 54: you don't need that

src/Nethermind/Nethermind.Merge.Plugin.Test/EngineModuleTests.Setup.cs
```diff
@@ -93,7 +96,8 @@ private IEngineRpcModule CreateEngineModule(MergeTestBlockchain chain, ISyncConf
             new GetPayloadV3Handler(
                 chain.PayloadPreparationService!,
                 chain.SpecProvider!,
-                chain.LogManager),
+                chain.LogManager,
+                _api!.CensorshipDetector),
```
LukaszRozmej, line 100: You don't even need that if censorship detector is made optional

src/Nethermind/Nethermind.Merge.Plugin/Handlers/GetPayloadV3Handler.cs
```diff
@@ -14,11 +15,17 @@ namespace Nethermind.Merge.Plugin.Handlers;
 /// </summary>
 public class GetPayloadV3Handler : GetPayloadHandlerBase<GetPayloadV3Result>
 {
-    public GetPayloadV3Handler(IPayloadPreparationService payloadPreparationService, ISpecProvider specProvider, ILogManager logManager) : base(
+    private readonly CensorshipDetector _censorshipDetector;
+    public GetPayloadV3Handler(
+        IPayloadPreparationService payloadPreparationService,
+        ISpecProvider specProvider,
+        ILogManager logManager,
+        CensorshipDetector censorshipDetector) : base(
```
LukaszRozmej, line 23: Make CensorhipDetector optional here, and ShouldOverrideBuilder should be false if it isn't provided.

src/Nethermind/Nethermind.Merge.Plugin/Data/GetPayloadV3Result.cs
```diff
@@ -3,23 +3,26 @@
...
+        _censorshipDetector = censorshipDetector;
     }
 
     public BlobsBundleV1 BlobsBundle { get; }
 
     public override ExecutionPayloadV3 ExecutionPayload { get; }
 
-    public bool ShouldOverrideBuilder { get; }
-
+    public bool ShouldOverrideBuilder => _censorshipDetector.CensorshipDetected;
```
LukaszRozmej, line 25: Make this `{get; init}` and just set it when censorship is detected.
This type is directly serialized and should be considered DTO.

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector.cs
```diff
@@ -0,0 +1,91 @@
...
+
+    private void OnBlockProcessing(object? sender, BlockEventArgs e)
+    {
+        Task.Run(() => Cache(e.Block));
+    }
+
+    private void OnNewHead(object? sender, BlockEventArgs e)
+    {
+        Task.Run(() => Cache(e.Block));
+    }
```
LukaszRozmej, line 49: Now you can easily cache same block twice, is this desired?

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector.cs
```diff
@@ -0,0 +1,91 @@
...
+        Transaction[] blockTransactions = block.Transactions;
+
+        UInt256 maxGasPriceInBlock = 0;
+        for (int i = 0; i < blockTransactions.Length; i++)
+        {
+            maxGasPriceInBlock = UInt256.Max(maxGasPriceInBlock, blockTransactions[i].GasPrice);
+        }
+
+        UInt256 maxGasPriceInPool = 0;
+        foreach (Transaction tx in from KeyValuePair<AddressAsKey, Transaction[]> txsBySender in txs
```
LukaszRozmej, line 63: We don't use LINQ syntax in our code, can you convert it to Select method?

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector.cs
```diff
@@ -0,0 +1,91 @@
...
+                                   let tx = txsBySender.Value[0]
+                                   select tx)
+        {
+            _accounts.TryGetAccount(tx.SenderAddress, out AccountStruct senderAccount);
+            if (senderAccount.Nonce == tx.Nonce)
+            {
+                // check transaction.GasBottleneck, ask Marcin Sobczak
+                maxGasPriceInPool = UInt256.Max(tx.GasPrice, maxGasPriceInPool);
+            }
```
LukaszRozmej, line 72: @marcindsobczak could have check the nonce using GasBottleneck? This info should already be in TxPool.

LukaszRozmej (commented):
Love the simplifications!
Few minor things to polish.

Next steps:
- Write tests for CensorshipDetector class - either unit or proper integration using BlockchainTests or descendant class.
- Test in real world.
- Consider implementing different censorship detection strategies.

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector.cs
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        _censorshipDetector.Set(block.Number, maxGasPriceInPool > maxGasPriceInBlock);
```
LukaszRozmej, line 75: Consider adding some logs when we detect potential censorship

marcindsobczak (commented):
In the way you wrote it - it will be working, but you overcomplicated it. You are taking snapshot of the whole pool, then selecting lowest nonce from each bucket, then reading accounts and looking for next-nonce-to-be-executed and then iterating for the biggest gas price (which post-1559 is equal to `maxPriorityFeePerGas`).

Instead, you can simplify it and do a snapshot of lowest-nonce-per-sender only and then just look for the tx with biggest gasBottleneck.

marcindsobczak:
One more potential issue - there are 2 ways of paying tx fee. New way, post London fork, you are paying `baseFee` + `priority fee`. Priority fee is just `tx.GasPrice`.  But there are pre-London tx types as well - `legacy` and `access list`.  These types have only GasPrice, which all is consumed when tx is executed. For these types, under `tx.GasPrice`, you have baseFee + priority fee. Post 1559 `tx.GasPrice` means priority fee only. So you probably need take it into consideration when looking for potentially censored txs.
One example: Let's say that we have in the pool legacy tx with GasPrice equal 10 Gwei. In the last block, the most expensive tx is 1559-type and has tx.GasPrice (priorityFee) equal 1 Gwei. BaseFee is 20Gwei.
In the current form, you will consider it as censorship. But in reality, our legacy tx is too cheap to be included (as baseFee is 20 Gwei and tx is offering only 10 Gwei).

So what I will recommend here is to prepare `maxGasPriceInBlock` as tx.GasPrice + block base fee if best tx was supporting 1559 or just tx.GasPrice if was legacy or access list.

Then you only need to find tx in mempool with biggest gasBottleneck - gas bottleneck has inside current base fee + priority fee

marcindsobczak:
And one more thing - blob transactions. There is a separate market with blob date fee and a hard limit of max 6 blob txs per block. We probably don't need to check censorship in the context of blob transactions. But we can miss some censorship attempts because of it.

Example: Our most expensive tx in the pool is 1559-type tx with `tx.GasPrice` equal 2 Gwei.
In the block you found the most expensive tx -  a blob tx with `tx.GasPrice` equal 3 Gwei. You don't consider it as censorship - our best tx is cheper than best tx in the block.
But if we exclude all blob transactions, let's say next most expensive tx in the block pays only 1 Gwei. And it was actually censorship on the tx which we have in the pool.

My recommendation - exclude blob-type txs when setting `maxGasPriceInBlock`

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector.cs
Arindam2407: No, this is a mistake. We have to distinguish between what we do when block is processed and when block is added. Potential solution is to maintain an additional slightly bigger cache for processed blocks. As the blocks get confirmed, they are simply added to the main cache.

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector.cs
```diff
@@ -0,0 +1,108 @@
...
+                }
+                else
+                {
+                    maxGasPriceInBlock = UInt256.Max(maxGasPriceInBlock, tx.GasPrice + block.BaseFeePerGas);
+                }
+            }
+        }
+
+        Transaction bestTx = _txPool.GetBestTxOfEachSender().First();
+        UInt256 maxGasPriceInPool = bestTx.Supports1559 ? bestTx.GasPrice + block.BaseFeePerGas : bestTx.GasPrice;
```
marcindsobczak, line 81: This tx can potentially not be ready (has nonce gap). All required info already is in `gasBottleneck`.
`UInt256 maxGasPriceInPool = bestTx.GasBottleneck;` would be enough, because it is already `gasPremium + baseFee` for 1559+ and `gasPrice` for others. And `gasBottleneck` would be 0 if there is a nonce gap

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector.cs
```diff
@@ -0,0 +1,108 @@
...
+            if (!tx.SupportsBlobs)
+            {
+                if (tx.Supports1559)
+                {
+                    maxGasPriceInBlock = UInt256.Max(maxGasPriceInBlock, tx.GasPrice);
+                }
+                else
+                {
+                    maxGasPriceInBlock = UInt256.Max(maxGasPriceInBlock, tx.GasPrice + block.BaseFeePerGas);
+                }
```
marcindsobczak, line 76: it should be in reverse: gas price + base fee for 1559 and just gas price for not supporting 1559

marcindsobczak (commented):
Good changes, but still a bit to improve.
Also it would be great to add some unit tests - tests will help you to catch potential mistakes/inconsistences in implementation and we will be protected from broking it by accident in the future

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector.cs
Arindam2407:
```suggestion
        UInt256 maxGasPriceInPool = 0;
        foreach (Transaction tx in _txPool.GetBestTxOfEachSender())
        {
            if (tx.GasBottleneck is not null || tx.GasBottleneck > 0)
            {
                maxGasPriceInPool = (UInt256)tx.GasBottleneck;
                break;
            }
            else
            {
                continue;
            }
        }
```
How about this? We check if the GasBottleneck is calculated and the transaction is ready. If so, we simply set maxGasPriceInPool and break, instead of iterating throughout all best transactions.

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector.cs
Arindam2407:
```suggestion
                if (tx.Supports1559)
                {
                    UInt256 txGasPrice = UInt256.Min(tx.MaxFeePerGas, tx.MaxPriorityFeePerGas + block.BaseFeePerGas);
                    maxGasPriceInBlock = UInt256.Max(maxGasPriceInBlock, txGasPrice);
                }
                else
                {
                    maxGasPriceInBlock = UInt256.Max(maxGasPriceInBlock, tx.GasPrice);
                }
```
I am considering changing it according to the calculation given in the [GasPriceTxComparerHelper.cs](https://github.com/NethermindEth/nethermind/blob/master/src/Nethermind/Nethermind.Consensus/Comparers/GasPriceTxComparerHelper.cs) file

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector.cs
marcindsobczak: Good idea, perfectly logical. We don't even need to iterate further if best tx doesn't have `tx.GasBottleneck > 0` - it will mean that we will not find any tx with positive gas bottleneck. So we can focus only on first tx from this collection.

Going further - maybe we can get from `SortedPool` only one, best tx? We would need to create e.g. `GetBest()` method, which will call `GetFirsts()` and return only first element. Then we can replace `GetBestTxOfEachSender()` with e.g. `GetBestTx()`, which will return `GetBest()` from SortedPool

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector.cs
marcindsobczak: Good idea, both would work.
More context: You don't need to check `MaxFeePerGas` - if tx was included, it means it was high enough. At this point only matters `MaxPriorityFeePerGas` and `BaseFee`. But if you use comparer, it will be actually even more simple - you can collect `bestTxInBlock` instead of `maxGasPriceInBlock` and then just use comparer to compare `bestTxInBlock` with `bestTxInPool`, all comparing-related logic would be reused, so you will be sure there is no mistake.

But in production code we use [GasPriceTxComparer](https://github.com/NethermindEth/nethermind/blob/master/src/Nethermind/Nethermind.Consensus/Comparers/GasPriceTxComparer.cs), `GasPriceTxComparerHelper` is used only if one of txs doesn't have calculated gasBottleneck and is called from `GasPriceTxComparer`

src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -731,7 +731,7 @@ public UInt256 GetLatestPendingNonce(Address address)
             return maxPendingNonce;
         }
 
-        public IEnumerable<Transaction> GetBestTxOfEachSender() => _transactions.GetFirsts();
+        public Transaction GetBestTx() => _transactions.GetBest() ?? new();
```
marcindsobczak, line 734: maybe nullable `Transaction? GetBestTx()`?

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector.cs
```diff
@@ -18,77 +16,76 @@ namespace Nethermind.Consensus.Processing;
...
-                }
+                bestTxInBlock = tx;
             }
         }
 
-        Transaction bestTx = _txPool.GetBestTxOfEachSender().First();
-        UInt256 maxGasPriceInPool = bestTx.Supports1559 ? bestTx.GasPrice + block.BaseFeePerGas : bestTx.GasPrice;
+        Transaction bestTxInPool = _txPool.GetBestTx();
+
+        bool unconfirmedBlockPotentialCensorship = _comparer.Compare(bestTxInBlock, bestTxInPool) > 0;
```
marcindsobczak, line 72: and then do it only if `bestTxInPool` is not `null`. But not sure if worth refactoring

marcindsobczak (approved):
LGTM

src/Nethermind/Nethermind.Blockchain.Test/BlockTreeTests.cs
```diff
@@ -1800,8 +1800,6 @@ public void BlockAddedToMain_should_have_updated_Head()
             Block block0 = Build.A.Block.WithNumber(0).WithDifficulty(1).TestObject;
             Block block1 = Build.A.Block.WithNumber(1).WithDifficulty(2).WithParent(block0).TestObject;
             AddToMain(blockTree, block0);
-            blockTree.SuggestBlock(block0);
-            blockTree.UpdateMainChain(new[] { block0 }, true);
```
LukaszRozmej, line 1804: Is this not needed?

src/Nethermind/Nethermind.TxPool/TxPool.cs
LukaszRozmej: +1 don't create artificial transaction, return null if not present

src/Nethermind/Nethermind.TxPool/NullTxPool.cs
```diff
@@ -40,6 +40,8 @@ public class NullTxPool : ITxPool
 
         public bool RemoveTransaction(Hash256? hash) => false;
 
+        public Transaction GetBestTx() => new();
```
LukaszRozmej, line 43: should be null.

src/Nethermind/Nethermind.Merge.Plugin/Data/GetPayloadV3Result.cs
```diff
@@ -3,23 +3,26 @@
...
 using Nethermind.Core;
 using Nethermind.Int256;
+using Nethermind.Consensus.Processing;
 
 namespace Nethermind.Merge.Plugin.Data;
 
 public class GetPayloadV3Result : GetPayloadV2Result
 {
-    public GetPayloadV3Result(Block block, UInt256 blockFees, BlobsBundleV1 blobsBundle) : base(block, blockFees)
+    private readonly CensorshipDetector? _censorshipDetector;
```
LukaszRozmej, line 12: Is it needed here? I think not.

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector.cs
```diff
@@ -0,0 +1,113 @@
...
+    {
+        _txPool = txPool;
+        _comparer = comparer;
+        _blockProcessor = blockProcessor;
+        _blockTree = blockTree;
+        _logger = logManager?.GetClassLogger() ?? throw new ArgumentNullException(nameof(logManager));
+        _temporaryCensorshipDetector = new(
+            _temporaryCacheSize, _temporaryCacheSize, "temporaryCensorshipDetector"
+            );
+        _censorshipDetector = new(_cacheSize, _cacheSize, "censorshipDetector");
```
LukaszRozmej, line 43: prefer in-line field initialization where possible

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector.cs
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    private void OnBlockAddedToMain(object? sender, BlockReplacementEventArgs e)
+    {
+        Task.Run(() => Cache(e.Block));
+    }
+
+    private void PreCachingOperation(Block block)
+    {
+        Transaction bestTxInBlock = new Transaction();
```
LukaszRozmej, line 60: just take first transaction instead of creating artificial one, you can use `for` to easier manage the loop

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector.cs
```diff
@@ -0,0 +1,113 @@
...
+            {
+                bestTxInBlock = tx;
+            }
+        }
+
+        Transaction bestTxInPool = _txPool.GetBestTx();
+
+        bool unconfirmedBlockPotentialCensorship = _comparer.Compare(bestTxInBlock, bestTxInPool) > 0;
+
+        _temporaryCensorshipDetector.Set(block.Number, unconfirmedBlockPotentialCensorship);
```
LukaszRozmej, line 74: maybe temporaryCensorshipDetector keys should be based on Hash or both Hash and BlockNumber? In case of reorgs we can have multiple blocks on same height (same blocknumber)?

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector.cs
```diff
@@ -0,0 +1,113 @@
...
+        }
+    }
+
+    public bool GetTemporaryCensorshipStatus(long blockNumber) => _temporaryCensorshipDetector.Get(blockNumber);
+
+    public bool GetCensorshipStatus(long blockNumber) => _censorshipDetector.Get(blockNumber);
+
+    public bool TemporaryCacheContainsBlock(long blockNumber) => _temporaryCensorshipDetector.Contains(blockNumber);
+
+    public bool CacheContainsBlock(long blockNumber) => _censorshipDetector.Contains(blockNumber);
```
LukaszRozmej, line 106: Not the biggest fan of exposing those just for testing purposes.

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector.cs
```diff
@@ -0,0 +1,113 @@
...
+
+public class CensorshipDetector : IDisposable
+{
+    private readonly ITxPool _txPool;
+    private readonly IComparer<Transaction> _comparer;
+    private readonly IBlockProcessor _blockProcessor;
+    private readonly IBlockTree _blockTree;
+    private readonly ILogger _logger;
+    private readonly LruCache<long, bool> _temporaryCensorshipDetector;
+    private readonly LruCache<long, bool> _censorshipDetector;
```
LukaszRozmej, line 24: Naming: maybe call it _censoredBlocks, not _censorhipDetector

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector.cs
LukaszRozmej: Better solution is to just have methods like `IEnumerable<long> GetCensoredBlocks()` that would return block numbers we think are censored and base test on that. Makes more sense from API standpoint.

LukaszRozmej (changes requested):
Overall very good, but I think few more details need to be polished.

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector.cs
```diff
@@ -0,0 +1,113 @@
...
+        }
+    }
+
+    public bool CensorshipDetected
+    {
+        get
+        {
+            return _censorshipDetector.GetValues().All(s => s);
+        }
+    }
```
LukaszRozmej, line 98: Maybe we should change it to a method, taking a parameter of returning block number, and then when checking values in cache - check that we are checking parent block numbers? (or even hashes?)

marcindsobczak:
Maybe worth to add a metric, e.g. `NumberOfCensoredTransactionsFound`? Or `NumberOfCensoredBlocks`

src/Nethermind/Nethermind.Blockchain.Test/BlockTreeTests.cs
Arindam2407: No, because AddToMain(blockTree, block0) does the exact same thing. Its code duplication.

Arindam2407:
> Maybe worth to add a metric, e.g. `NumberOfCensoredTransactionsFound`? Or `NumberOfCensoredBlocks`

We could have a counter of number of blocks for which ShouldOverrideBuilder is true but that is not exactly the number of censored blocks

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector.cs
Arindam2407: I have added the GetCensoredBlocks method. I have removed the above two methods but kept the bottom two. They should not be necessary but my tests are failing without them. Would need some help on that.

Arindam2407:
> Maybe worth to add a metric, e.g. `NumberOfCensoredTransactionsFound`? Or `NumberOfCensoredBlocks`

Added GetCensoredBlocksCount() Method

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector/CensorshipDetector.cs
```diff
@@ -0,0 +1,187 @@
...
+        _comparer = comparer;
+        _blockProcessor = blockProcessor;
+        _logger = logManager?.GetClassLogger() ?? throw new ArgumentNullException(nameof(logManager));
+        _censorshipDetectorConfig = censorshipDetectorConfig;
+
+        if (_censorshipDetectorConfig.AddressesForCensorshipDetection is not null)
+        {
+            foreach (string hexString in _censorshipDetectorConfig.AddressesForCensorshipDetection)
+            {
+                if (Address.IsValidAddress(hexString, true) && Address.TryParse(hexString, out Address address))
```
LukaszRozmej, line 47: not sure if you need to call both, `TryParse` should suffice

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector/CensorshipDetector.cs
```diff
@@ -0,0 +1,187 @@
...
+    public Hash256? ParentHash { get; }
+
+    public BlockCensorshipInfo(bool isCensored, Hash256? parentHash)
+    {
+        IsCensored = isCensored;
+        ParentHash = parentHash;
+    }
+}
+
+public readonly struct BlockNumberHash
```
LukaszRozmej, line 171: Don't we need `: IEquatable<BlockNumberHash>` ?

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector/CensorshipDetector.cs
```diff
@@ -0,0 +1,187 @@
...
+        Transaction bestTxInBlock = block.Transactions[0];
+        SortedDictionary<Address, bool> _blockAddressCensorshipDetectorHelper = [];
+        // Number of unique addresses specified by the user for censorship detection, to which txs are sent in the block
+        long _blockCensorshipDetectionUniqueAddressCount = 0;
+
+        foreach (Transaction tx in block.Transactions)
+        {
+            if (!tx.SupportsBlobs)
+            {
+                if (_poolAddressCensorshipDetectorHelper.TryGetValue(tx.To!, out _) && !_blockAddressCensorshipDetectorHelper.TryGetValue(tx.To!, out _))
```
LukaszRozmej, line 106: if you are not interested in value, use `ContainsKey`

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector/CensorshipDetector.cs
```diff
@@ -0,0 +1,187 @@
...
+namespace Nethermind.Consensus.Processing.CensorshipDetector;
+
+public class CensorshipDetector : IDisposable
+{
+    private readonly ITxPool _txPool;
+    private readonly IComparer<Transaction> _comparer;
+    private readonly IBlockProcessor _blockProcessor;
+    private readonly ILogger _logger;
+    private readonly ICensorshipDetectorConfig _censorshipDetectorConfig;
+    private readonly SortedDictionary<Address, long> _poolAddressCensorshipDetectorHelper = [];
```
LukaszRozmej, line 25: Any reason for using `SortedDcitionary` over `Dictionary`? The 2nd one should be faster.
Use `AddressAsKey` for key - should be bit faster.

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector/CensorshipDetector.cs
```diff
@@ -0,0 +1,187 @@
...
+        Task.Run(() => Cache(e.Block));
+    }
+
+    private void Cache(Block block)
+    {
+        Transaction bestTxInPool = _txPool.GetBestTx();
+        long _poolCensorshipDetectionUniqueAddressCount = Metrics.PoolCensorshipDetectionUniqueAddressesCount;
+
+        Transaction bestTxInBlock = block.Transactions[0];
+        SortedDictionary<Address, bool> _blockAddressCensorshipDetectorHelper = [];
```
LukaszRozmej, line 98: What is the point of this dictionary? If only to check that we count an address once, then `HashSet` should work better

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector/CensorshipDetector.cs
```diff
@@ -0,0 +1,187 @@
...
+
+                if (_comparer.Compare(bestTxInBlock, tx) > 0)
+                {
+                    bestTxInBlock = tx;
+                }
+            }
+        }
+
+        bool _highPayingTxCensorship = _comparer.Compare(bestTxInBlock, bestTxInPool) > 0;
+        bool _addressCensorship = _blockCensorshipDetectionUniqueAddressCount * 2 < _poolCensorshipDetectionUniqueAddressCount;
```
LukaszRozmej, line 120: naming convention: only (private) fields should be starting with `_`, local variables should use `camelCase`

LukaszRozmej, line 120: I kind of dislike this logic:
1. `_poolCensorshipDetectionUniqueAddressCount` doesn't take into account nonces, maybe those transactions aren't ready to be included.
2. `_poolCensorshipDetectionUniqueAddressCount` doesn't take into account basefee or gas premium, maybe other transactions just pay better?

This leads that this can easily identify false positives.

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector/CensorshipDetector.cs
```diff
@@ -0,0 +1,187 @@
...
+    {
+        IsCensored = isCensored;
+        ParentHash = parentHash;
+    }
+}
+
+public readonly struct BlockNumberHash
+{
+    public long Number { get; }
+    public Hash256 Hash { get; }
```
LukaszRozmej, line 174: If it is a struct, maybe use ValueHash? @benaadams

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector/CensorshipDetector.cs
```diff
@@ -0,0 +1,187 @@
...
+public class CensorshipDetector : IDisposable
+{
+    private readonly ITxPool _txPool;
+    private readonly IComparer<Transaction> _comparer;
+    private readonly IBlockProcessor _blockProcessor;
+    private readonly ILogger _logger;
+    private readonly ICensorshipDetectorConfig _censorshipDetectorConfig;
+    private readonly SortedDictionary<Address, long> _poolAddressCensorshipDetectorHelper = [];
+    private readonly LruCache<BlockNumberHash, BlockCensorshipInfo> _potentiallyCensoredBlocks;
+    private IEnumerable<BlockNumberHash> _censoredBlocks = [];
```
LukaszRozmej, line 27: Please use `List<BlockNumberHash>` for the field itself

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector/CensorshipDetector.cs
```diff
@@ -0,0 +1,187 @@
...
+                Metrics.LastCensoredBlockNumber = block.Number;
+                if (_logger.IsInfo)
+                {
+                    _logger.Info($"Censorship detected for block no. {block.Number} with hash {block.Hash!}");
+                }
+            }
+        }
+    }
+
+    public IEnumerable<BlockNumberHash> GetCensoredBlocks() => _censoredBlocks;
```
LukaszRozmej, line 146: `_censoredBlocks` will only grow, it is never cleaned up.

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector/CensorshipDetector.cs
```diff
@@ -0,0 +1,187 @@
...
+        // Checking last 3 blocks for potential censorship
+        if (blockCensorshipInfo.IsCensored && block.Number > 2)
+        {
+            BlockCensorshipInfo b1 = _potentiallyCensoredBlocks.Get(new BlockNumberHash(block.Number - 1, blockCensorshipInfo.ParentHash!));
+            BlockCensorshipInfo b2 = _potentiallyCensoredBlocks.Get(new BlockNumberHash(block.Number - 2, b1.ParentHash!));
+            BlockCensorshipInfo b3 = _potentiallyCensoredBlocks.Get(new BlockNumberHash(block.Number - 3, b2.ParentHash!));
+
+            if (!_censoredBlocks.Contains(blockNumberHash) && b1.IsCensored && b2.IsCensored && b3.IsCensored)
+            {
+                _censoredBlocks = _censoredBlocks.Append(blockNumberHash);
```
LukaszRozmej, line 135: append on IEnumerable is bad performance, List.Add is better

LukaszRozmej (changes requested):

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector/CensorshipDetector.cs
```diff
@@ -0,0 +1,228 @@
...
+        foreach (Transaction? bestTx in _poolAddressCensorshipBestTx.Values)
+        {
+            if (bestTx is not null && _comparer.Compare(bestTx, worstTxInBlock) <= 0)
+            {
+                poolCensorshipDetectionUniqueAddressCount++;
+            }
+        }
+
+        bool isCensored = _comparer.Compare(bestTxInBlock, _txPool.GetBestTx()) > 0
+        || blockCensorshipDetectionUniqueAddressCount * 2 < poolCensorshipDetectionUniqueAddressCount;
```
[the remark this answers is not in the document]
marcindsobczak: What we have:
1. We are iterating TxPool txs (one - best one- per sender) and building `_poolAddressCensorshipBestTx`, which will have best tx which is sending to every receiver (so just best tx per every `To` pointed in config)
2. We are iterating all block txs and looking for receiver address (`tx.To`) in previously built `_poolAddressCensorshipBestTx`. If found, incrementing `blockCensorshipDetectionUniqueAddressCount`. So `blockCensorshipDetectionUniqueAddressCount` means a number of txs in the block which were sending txs to receiver `To` from the configured list. `blockAddressCensorshipDetectorHelper` is ensuring that we are counting every receiver only once.
3. We are iterating `_poolAddressCensorshipBestTx` and comparing with worst tx in the block. If tx in collection is better, incrementing `poolCensorshipDetectionUniqueAddressCount`. So `poolCensorshipDetectionUniqueAddressCount` means a number of configured receiver addresses with tx in our TxPool, which were better than worst tx in the block, but not included in the block
4. Then we are setting `isCensored` if at least one of 2 parts is true.
- comparing best tx in the block with best tx in our pool and `true` if tx from the pool is better. But we are checking just best tx, no matter which `To` it has. So we can have censorship detected even if we don't have in the pool any tx with `To` from the configured list. Is it expected?
- checking if 2 * number of receivers which have tx sending to them in the block is lower than number of receivers which have tx in our pool which is better than worst tx in the block.

In current form we will have censorship detected always, when we have in the pool ANY tx better than best tx in the block. Is it expected behavior? Maybe we could change `bool isCensored = _comparer.Compare(bestTxInBlock, _txPool.GetBestTx()) > 0` to check best tx from `_poolAddressCensorshipBestTx.Values` instead of `_txPool.GetBestTx()`? But I'm not sure what is expected behavior here

We are also considering that there is a censorship, if we found 2*X receivers with txs in our pool better than worst tx in the block, and in the block were included less than X receivers (ofc counting only receivers from the configured list)

marcindsobczak (commented):
How I understand current logic
- potential censorship detected if the best tx from TxPool is better than the best tx in the block
- or if number of receivers (provided in config) with tx in TxPool which were better than worst tx in the block (so were able to be included) is bigger than twice a number of receivers from the provided list with txs actualy included in the block
- censorship detected if potential censorship were detected in 4 consecutive blocks

Looks fine, I think just one small issue to fix - more in comment

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector/CensorshipDetector.cs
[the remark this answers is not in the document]
marcindsobczak: Probably it is fine as it is

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector/CensorshipDetector.cs
```diff
@@ -0,0 +1,228 @@
...
+        _potentiallyCensoredBlocks = new(_cacheSize, _cacheSize, "potentiallyCensoredBlocks");
+        _censoredBlocks = new(_cacheSize);
+        _blockProcessor.BlockProcessing += OnBlockProcessing;
+    }
+
+    private void OnBlockProcessing(object? sender, BlockEventArgs e)
+    {
+        UInt256 baseFee = e.Block.BaseFeePerGas;
+
+        IEnumerable<Transaction> poolBestTransactions = _txPool.GetBestTxOfEachSender();
```
LukaszRozmej, line 63: One thing, maybe instead of getting all from pool, we should loop and get best only for tracked addresses, we will pull only that number of transactions from pool instead of pulling up to 2k transactions here.

So instead of having `GetBestTxOfEachSender()` we could have `GetBestTransaction(Address sender)`.

Also we need to check if this transaction is includable?

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector/CensorshipDetector.cs
```diff
@@ -0,0 +1,228 @@
...
+        Transaction worstTxInBlock = block.Transactions[0];
+        HashSet<AddressAsKey> blockAddressCensorshipDetectorHelper = [];
+        // Number of unique addresses specified by the user for censorship detection, to which txs are sent in the block
+        long blockCensorshipDetectionUniqueAddressCount = 0;
+
+        foreach (Transaction tx in block.Transactions)
+        {
+            if (!tx.SupportsBlobs)
+            {
+                if (_poolAddressCensorshipBestTx.ContainsKey(tx.To!)
```
LukaszRozmej, line 94: I think we could compare with transaction that is not includable?

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector/CensorshipDetector.cs
```diff
@@ -0,0 +1,228 @@
...
+
+    private void Cache(Block block)
+    {
+        Transaction bestTxInBlock = block.Transactions[0];
+        Transaction worstTxInBlock = block.Transactions[0];
+        HashSet<AddressAsKey> blockAddressCensorshipDetectorHelper = [];
+        // Number of unique addresses specified by the user for censorship detection, to which txs are sent in the block
+        long blockCensorshipDetectionUniqueAddressCount = 0;
+
+        foreach (Transaction tx in block.Transactions)
```
LukaszRozmej, line 90: Can you put a detailed comment about the logic of this method?

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector/CensorshipDetector.cs
marcindsobczak: Our tracked addresses are in all TxPool txs as `To`, not senders. So we will still need to iterate

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector/CensorshipDetector.cs
marcindsobczak: in `_poolAddressCensorshipBestTx` there are only includable txs, it was checked when building this collection, in line 66 (`tx.GasBottleneck > baseFee`)

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector/CensorshipDetector.cs
```diff
@@ -0,0 +1,228 @@
...
+                    blockCensorshipDetectionUniqueAddressCount++;
+                }
+
+                if (_comparer.Compare(bestTxInBlock, tx) > 0)
+                {
+                    bestTxInBlock = tx;
+                }
+                else
+                {
+                    worstTxInBlock = tx;
```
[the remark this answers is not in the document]
marcindsobczak: good catch, I think we need to check it. Otherwise we can potentially end with second-best tx in the block as `worstTxInBlock`

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector/CensorshipDetector.cs
marcindsobczak, line 63: Maybe let's add a check here `if(_poolAddressCensorshipBestTx.Length != 0)`, we will skip all snapshots and most of the logic if there is no list provided, so I think in most cases. And if list is empty (`_poolAddressCensorshipBestTx`), also skip most of the logic in `Cache` and just compare best tx in the pool with best tx in the block (so enf `if` before `isCensored`)?

LukaszRozmej: ok ignore then, bad idea

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector/CensorshipDetector.cs
marcindsobczak: for non-ready txs gas bottleneck is 0, so yes, tx readiness is guaranteed by `tx.GasBottleneck > baseFee`

LukaszRozmej: Right, but @Arindam2407 this should land in comments as it is a bit cryptic

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector/CensorshipDetector.cs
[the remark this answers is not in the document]
Arindam2407: I don't think we need to change the isCensored method since we are dealing with two separate kinds of censorship detection here. I think it seems a bit convoluted because of a lack of comments.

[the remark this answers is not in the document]
LukaszRozmej: Ok this is the last line I have doubts about:
1. First part we are checking: `_comparer.Compare(bestTxInBlock, _txPool.GetBestTx()) > 0`, this means we found transaction in txPool that is better than best transaction in block - right? Should we quantify it in any matter? Like have X% threshold of being better?
2. Second part checking that we included transactions from at least half of tracked addresses with better transactions than ones in block?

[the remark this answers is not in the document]
Arindam2407: That's right. I think an X% threshold is easy to implement if that is what we want. Although, if the best tx is even only slightly better than the best tx in block for 4 blocks straight, that definitely should raise suspicion.

LukaszRozmej (approved):

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector/CensorshipDetectorConfig.cs
```diff
@@ -0,0 +1,9 @@
+// SPDX-FileCopyrightText: 2024 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+namespace Nethermind.Consensus.Processing.CensorshipDetector;
+
+public class CensorshipDetectorConfig : ICensorshipDetectorConfig
+{
+    public string[]? AddressesForCensorshipDetection { get; set; } = null;
```
MarekM25, line 8: I would consider adding a boolean flag, Enabled. If Enabled is true, we will not listen to events, etc. For new heuristics, we might use something other than addresses in the config.

If Enabled is false, I would expect that we would have a NoCensorshipDetection class and do nothing. The default setting should be false

LukaszRozmej: If disabled, don't create `CensorhipDetector` at all

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector/CensorshipDetector.cs
```diff
@@ -61,8 +67,18 @@ public CensorshipDetector(
         _blockProcessor.BlockProcessing += OnBlockProcessing;
     }
 
+    private bool IsSyncing()
+    {
+        long bestSuggestedNumber = _blockTree.FindBestSuggestedHeader()?.Number ?? 0;
+        long headNumberOrZero = _blockTree.Head?.Number ?? 0;
+        return bestSuggestedNumber > headNumberOrZero + 8;
```
LukaszRozmej, line 78: we probably shouldn't even do `8` threshold, as here we should only check tx pool if we are processing current head block, right?

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector/ICensorshipDetectorConfig.cs
```diff
@@ -7,6 +7,14 @@ namespace Nethermind.Consensus.Processing.CensorshipDetector;
 
 public interface ICensorshipDetectorConfig : IConfig
 {
+    [ConfigItem(DefaultValue = "false",
+        Description = "Enabling censorship detection feature")]
+    bool Enabled { get; set; }
```
LukaszRozmej, line 12: Shouldn't we enable it by default? Or do we plan to start slowly?

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector/CensorshipDetector.cs
```diff
@@ -43,7 +43,11 @@ public CensorshipDetector(
         _txPool = txPool;
         _betterTxComparer = betterTxComparer;
         _blockProcessor = blockProcessor;
-        _blockCensorshipThreshold = censorshipDetectorConfig.BlockCensorshipThreshold;
+        int configBlockCensorshipThreshold = censorshipDetectorConfig.BlockCensorshipThreshold;
+        if (configBlockCensorshipThreshold > 1 && configBlockCensorshipThreshold <= CacheSize)
+        {
+            _blockCensorshipThreshold = configBlockCensorshipThreshold;
+        }
```
LukaszRozmej, line 50: Maybe make it `uint` and set `CasheSize` to the size of threshlod*4 or 8?

src/Nethermind/Nethermind.Consensus/Processing/CensorshipDetector/CensorshipDetector.cs
marcindsobczak: Right, my bad

LukaszRozmej (approved):

marcindsobczak (approved):
Looks good, let's just add a metric for potentially censored blocks

marcindsobczak:
Actually more useful will be not block number, but number of blocks. Maybe let's have both?

marcindsobczak (approved):

LukaszRozmej (approved):

LukaszRozmej (approved):

PR #7264, opened 2024-07-16
Improve Sync chains to sync more and faster with dynamic VMs

kamilchodola:
With that approach each master commit will be tested on each supported chain (excluding OP stack and Energyweb chains but just for now to shorten the time) so it will increase the coverage and reduce the costs for VMs.

Sample workflow which worked well
https://github.com/NethermindEth/nethermind/actions/runs/9956915219

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [X] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No


.github/workflows/sync-supported-chains.yml
```diff
@@ -7,37 +7,73 @@ on:
...
+      - name: Set Matrix
+        id: set-matrix
+        run: echo "matrix=$(jq -c . nethermind/scripts/workflow_config/sync_testnets_matrix.json)" >> $GITHUB_OUTPUT
+
+  create_a_runner:
+    needs: [setup-matrix]
+    strategy:
+      fail-fast: false
+      matrix:
+        client-network: ${{fromJson(needs.setup-matrix.outputs.matrix)}}
```
brbrr, line 39: IMO, `client-network` is a bit misleading, since it's actually a node configuration. I'd suggest renaming it into something like `config` or `node`

brbrr (approved):
Left few minor comments, but otherwise, it looks good!

.github/workflows/sync-supported-chains.yml
```diff
@@ -7,37 +7,73 @@ on:
     types:
       - completed
   workflow_dispatch:
+    inputs:
+      nethermind_image:
+        description: "Docker image to be used by action"
+        default: "nethermindeth/nethermind:master"
+        required: true
```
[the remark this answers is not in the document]
kamilchodola: Changed

.github/workflows/sync-supported-chains.yml
kamilchodola: Chnaged

PR #7269, opened 2024-07-17
Switch to Microsoft.IdentityModel.JsonWebTokens package

rubo:
## Changes

Replaced the `System.IdentityModel.Tokens.Jwt` package with `Microsoft.IdentityModel.JsonWebTokens` as recommended by Microsoft:

> As of IdentityModel 7x, this is a legacy tool that should be replaced with Microsoft.IdentityModel.JsonWebTokens.

#### What types of changes does your code introduce?

- [ ] Bugfix (a non-breaking change that fixes an issue)
- [ ] New feature (a non-breaking change that adds functionality)
- [ ] Breaking change (a change that causes existing functionality not to work as expected)
- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

#### If yes, did you write tests?

- [ ] Yes
- [x] No

PR #7273, opened 2024-07-18
Replace missing method in Kute

rubo:
## Changes

Replaced the missing method, causing a compile error introduced in #7269.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

### DIFF
--- a/tools/Nethermind.Tools.Kute/JsonRpcMethodFilter/ComposedJsonRpcMethodFilter.cs
+++ b/tools/Nethermind.Tools.Kute/JsonRpcMethodFilter/ComposedJsonRpcMethodFilter.cs
@@ -1,20 +1,13 @@
 // SPDX-FileCopyrightText: 2023 Demerzel Solutions Limited
 // SPDX-License-Identifier: LGPL-3.0-only
 
-using Microsoft.IdentityModel.Tokens;
-
 namespace Nethermind.Tools.Kute.JsonRpcMethodFilter;
 
-class ComposedJsonRpcMethodFilter : IJsonRpcMethodFilter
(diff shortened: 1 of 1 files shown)


emlautarom1 (approved):
Not a fan of `!_filters?.Any() ?? true` (very hard to understand) but it gets the job done

tools/Nethermind.Tools.Kute/JsonRpcMethodFilter/ComposedJsonRpcMethodFilter.cs
```diff
@@ -15,6 +15,6 @@ public ComposedJsonRpcMethodFilter(IEnumerable<IJsonRpcMethodFilter> filters)
     }
 
     public bool ShouldSubmit(string methodName) =>
-        _filters.IsNullOrEmpty() ||
-        _filters.Any(f => f.ShouldSubmit(methodName));
+        (!_filters?.Any() ?? true) ||
```
rubo, line 18: > Not a fan of !_filters?.Any() ?? true (very hard to understand) but it gets the job done

@emlautarom1 Since I don't maintain this tool, I can apply another approach you like more. As `_filters` is `IEnumerable`, using `Any` is faster than `Count`. Maybe you like more `_filter is null || !_filter.Any()`?

benaadams: Cache to `private readonly` in consructor? HasFilters

rubo: Alright, refactored and applied @benaadams' suggestion.

PR #7275, opened 2024-07-22
Log finder

srj31:
WIP LogIndex for optimized eth_getLogs call


src/Nethermind/Nethermind.Facade/Filters/AddressFilter.cs
```diff
@@ -13,6 +17,42 @@ public class AddressFilter
...
         private Bloom.BloomExtract[]? _addressesBloomIndexes;
         private Bloom.BloomExtract? _addressBloomExtract;
+        public static readonly IEnumerable<int> Any = [-1];
+
+        public IEnumerable<int> GetBlockNumbersFrom(LogIndexStorage logIndexStorage)
+        {
+            if (Addresses is not null)
+            {
+                var blocks = Addresses.Select(a => logIndexStorage.GetBlocksForAddress(a));
+                IEnumerator<int>[] enumerators = blocks.Select(b => b.GetEnumerator()).ToArray();
```
LukaszRozmej, line 27: `ToPooledList` instead `ToArray`? You need to dispose it later.

src/Nethermind/Nethermind.Db/LogIndexStorage.cs
```diff
@@ -0,0 +1,79 @@
...
+
+namespace Nethermind.Db
+{
+
+    public class LogIndexStorage
+    {
+        private Dictionary<AddressAsKey, int[]> _addressToBlocks = new Dictionary<AddressAsKey, int[]>();
+        private Dictionary<Hash256, int[]> _topicsToBlocks = new Dictionary<Hash256, int[]>();
+
+        public void LoadFile(Address address, string path)
```
LukaszRozmej, line 19: I Expect this is temporary?
We load the whole address index at once - we should load them page by page lazy.

src/Nethermind/Nethermind.Db/LogIndexStorage.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            const int maxLineLength = 1024;
+
+            if (!File.Exists(path))
+            {
+                return;
+            }
+            using SafeFileHandle fileHandle = File.OpenHandle(path, FileMode.OpenOrCreate);
+
+            byte[] rentedBuffer = ArrayPool<byte>.Shared.Rent(maxLineLength);
```
LukaszRozmej, line 29: You can use ArrayPoolList, it will grow itself

src/Nethermind/Nethermind.Db/LogIndexStorage.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            int read = RandomAccess.Read(fileHandle, rentedBuffer, offset);
+            Span<byte> bytes = default;
+            while (read > 0)
+            {
+                offset += read;
+                bytes = rentedBuffer.AsSpan(0, read);
+                int[] intData = new int[read / 4];
+
+                TurboPFor.p4nddec128v32(bytes.ToArray(), read / 4, intData);
```
LukaszRozmej, line 40: Would be good to avoid ToArray, Span should be passable

src/jwt.hex
```diff
@@ -0,0 +1 @@
+863524c99105fb2d71f8713a28e30d5090e2fe8348d4333ae607c8c310142031
```
LukaszRozmej, line 1: shouldn't be commited

src/Nethermind/Nethermind.Facade/Filters/LogOperators.cs
```diff
@@ -0,0 +1,93 @@
...
+                }
+            }
+
+
+            while (transactions.Count == enumerators.Length)
+            {
+                (Element blockNumber, IEnumerator<Element> enumerator) = transactions.Min;
+                (Element blockNumber2, IEnumerator<Element> enumerator2) = transactions.Max;
+
+                bool isIntersection = EqualityComparer<Element>.Default.Equals(blockNumber, blockNumber2);
```
LukaszRozmej, line 30: You can do just Equals, or do `where Element : IEquatable<Element>`

src/Nethermind/Nethermind.Facade/Filters/LogOperators.cs
```diff
@@ -0,0 +1,93 @@
...
+
+                transactions.Remove(blockNumber);
+
+                if (enumerator.MoveNext())
+                {
+                    transactions.Add(enumerator.Current!, enumerator);
+                }
```
LukaszRozmej, line 37: Don't you want to advance ALL values that are smaller than Max?

src/Nethermind/Nethermind.Facade/Filters/LogOperators.cs
```diff
@@ -0,0 +1,93 @@
+using System.Collections.Generic;
+using Nethermind.Core.Collections;
+
+namespace Nethermind.Blockchain.Filters.Topics
+{
+
+    public class LogOperators<Element>
+    {
+
+        public static IEnumerable<Element> Intersect(IEnumerator<Element>[] enumerators)
```
LukaszRozmej, line 10: Methods in this class would benefit from documentation as they are a bit confusing what they want to achieve.

src/Nethermind/Nethermind.Facade/Filters/LogFinder.cs
```diff
@@ -46,6 +50,64 @@ public class LogFinder : ILogFinder
...
+            _receiptStorage = receiptStorage ?? throw new ArgumentNullException(nameof(receiptStorage)); ;
+            _bloomStorage = bloomStorage ?? throw new ArgumentNullException(nameof(bloomStorage));
+            _receiptsRecovery = receiptsRecovery ?? throw new ArgumentNullException(nameof(receiptsRecovery));
+            _logger = logManager?.GetClassLogger<LogFinder>() ?? throw new ArgumentNullException(nameof(logManager));
+            _maxBlockDepth = maxBlockDepth;
+            _rpcConfigGetLogsThreads = Math.Max(1, Environment.ProcessorCount / 4);
+            _logIndexStorage = new LogIndexStorage();
+        }
+
+        public IEnumerable<FilterLog> FindLogsTest(LogFilter filter, CancellationToken cancellationToken = default)
```
LukaszRozmej, line 75: this is temp?

src/Nethermind/Nethermind.Crypto/TurboPFor.cs
```diff
@@ -0,0 +1,243 @@
...
+
+public static class TurboPFor
+{
+    private const string LibraryName = "ic";
+    private static string? _libraryFallbackPath;
+
+    static TurboPFor() => AssemblyLoadContext.Default.ResolvingUnmanagedDll += OnResolvingUnmanagedDll;
+
+    [DllImport(LibraryName)]
+    public static extern unsafe int vbenc32(int[] @in, int n, byte[] @out);
```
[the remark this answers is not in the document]
srj31: trying Span previously gave `Cannot marshal 'parameter #1': Non-blittable generic types cannot be marshaled`

src/Nethermind/Nethermind.Db/LogIndexStorage.cs
srj31: yup we were hoping we could test it out once we had the encoded data, but encoding it was taking too long

src/Nethermind/Nethermind.Facade/Filters/LogOperators.cs
srj31: I believe the result would be the same since there wont be any intersection until the values are greater than the Max but yea it will prevent the extra need to getting the .Max everytime

src/Nethermind/Nethermind.Facade/Filters/LogFinder.cs
srj31: yes, to be able to create an interface similar to the existing FindLogs for easier unit testing

PR #7278, opened 2024-07-22
Implement `debug_standardTraceBadBlockToFile` method

rubo:
## Changes

Added `debug_standardTraceBadBlockToFile` JSON-RPC method.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

Will be updated automatically.

- [ ] Yes
- [x] No


src/Nethermind/Nethermind.Consensus/Tracing/GethStyleTracer.cs
```diff
@@ -167,6 +171,24 @@ public IEnumerable<string> TraceBlockToFile(Hash256 blockHash, GethTraceOptions
...
+    public IEnumerable<string> TraceBadBlockToFile(Hash256 blockHash, GethTraceOptions options, CancellationToken cancellationToken)
+    {
+        ArgumentNullException.ThrowIfNull(blockHash);
+        ArgumentNullException.ThrowIfNull(options);
+
+        var block = _badBlockStore
+            .GetAll()
+            .Where(b => b.Hash == blockHash)
+            .FirstOrDefault()
+            ?? throw new InvalidOperationException("Only historical blocks");
```
rubo, line 183: @LukaszRozmej Is this error message relevant for bad blocks, or do we need a better one?

LukaszRozmej: Not sure, but generally it would mean we didn't find bad block

src/Nethermind/Nethermind.Consensus/Tracing/GethStyleTracer.cs
```diff
@@ -167,6 +171,24 @@ public IEnumerable<string> TraceBlockToFile(Hash256 blockHash, GethTraceOptions
...
 
+    public IEnumerable<string> TraceBadBlockToFile(Hash256 blockHash, GethTraceOptions options, CancellationToken cancellationToken)
+    {
+        ArgumentNullException.ThrowIfNull(blockHash);
+        ArgumentNullException.ThrowIfNull(options);
+
+        var block = _badBlockStore
+            .GetAll()
+            .Where(b => b.Hash == blockHash)
+            .FirstOrDefault()
```
LukaszRozmej, line 182: Maybe also require block number and don't load all bad blocks into memory?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Consensus/Tracing/GethStyleTracer.cs
rubo: I feel you, but we can't, as this RPC method requires block hash only. Even if it didn't, it would be strange to require the user to provide both number and hash. RPC methods typically require either a number or a hash.

PR #6366, opened 2023-12-13
Upgrade Kute

emlautarom1:
## Changes

- Add support tracing of responses to a file
- Validate JSON responses
- Fix issue where some metrics where being dropped
- Some refactoring

#### What types of changes does your code introduce?

- [ ] Bugfix (a non-breaking change that fixes an issue)
- [x] New feature (a non-breaking change that adds functionality)
- [ ] Breaking change (a change that causes existing functionality not to work as expected)
- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

#### If yes, did you write tests?

- [ ] Yes
- [x] No

#### Notes on testing

These changes do not affect the main Nethermind client.

#### Requires documentation update

- [x] Yes
- [ ] No

#### Requires explanation in Release Notes

- [ ] Yes
- [x] No

## Remarks

We've been using this branch for a while but it hasn't been merged to master yet.

PR #7279, opened 2024-07-22
Incorrect batch unwrapping in JSON-RPC replay tool

emlautarom1:
## Changes

- Fixes a bug where Kute would send a batch of requests individually but also as a batch, effectively sending all requests twice.
- Fixes a bug where Kute would crash when using the `-d | --dry` mode due to incorrect `null` handling.
- Optimize progress reporting when no batch unwrapping is requested.
- Properly cache certain class properties.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

As mentioned in https://github.com/NethermindEth/nethermind/pull/6366#issuecomment-1853918736 we need to introduce a proper test suite before adding any new features to Kute.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


tools/Nethermind.Tools.Kute/MessageProvider/JsonRpcMessageProvider.cs
```diff
@@ -26,26 +22,12 @@ await foreach (var msg in _provider.Messages)
...
-                    break;
-                case JsonValueKind.Array:
-                    if (_unwrapBatches)
-                    {
-                        foreach (JsonElement single in jsonDoc.RootElement.EnumerateArray())
-                        {
-                            yield return new JsonRpc.SingleJsonRpc(JsonDocument.Parse(single.ToString()));
-                        }
-                    }
-                    yield return new JsonRpc.BatchJsonRpc(jsonDoc);
```
emlautarom1, line 42: A double enumeration happens here: if we unwrap batches we yield the batch elements, but we also yield the batch request at the end (most likely a missing `else` branch)

tools/Nethermind.Tools.Kute/Application.cs
```diff
@@ -133,12 +132,14 @@ public async Task AnalyzeRequest((Task<HttpResponseMessage>, JsonRpc) task)
...
             case JsonRpc.BatchJsonRpc batch:
                 {
-                    HttpResponseMessage content;
+                    HttpResponseMessage? content;
                     using (_metrics.TimeBatch())
                     {
                         content = await task.Item1;
                     }
-                    var deserialized = JsonSerializer.Deserialize<JsonDocument>(content.Content.ReadAsStream());
+                    var deserialized = content != null
```
LukaszRozmej, line 140: is not null

tools/Nethermind.Tools.Kute/MessageProvider/JsonRpcMessageProvider.cs
LukaszRozmej: @kamilchodola ?

LukaszRozmej (approved):

kamilchodola (approved):

PR #7291, opened 2024-07-26
update known chain sizes

LukaszRozmej:

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [x] Other: chain size update

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks

@kamilchodola could you update pruned estimates for those chains?
Added some estimations for mainnet and sepolia, but I don't trust them.


benaadams:
Failed Update_known_chain_sizes [12 ms]
  Error Message:
   Expected value to be less than 240000000000L, but found <null>.

benaadams (approved):
Still think is a bit of a weird test 🙂

kamilchodola (approved):

kamilchodola:
@LukaszRozmej It looks ok as it is now probably.

PR #7308, opened 2024-08-05
Eth_simulate blob related bugfixes

OlegJakushkin:
Here we continue development of Eth_simulate feature, last time updated at [5530](https://github.com/NethermindEth/nethermind/pull/5530) 

In this PR we are implementing https://github.com/ethereum/execution-apis/pull/484 proposal. Fixing minor issues related to protocol updates such as:

 - [x] nethermind returns `blobBaseFee` field for a block, this should not be returned as its not part of the block header.
 - [x] nethermind does not return correct blob gas used
 - [x] gas used of call and block do not match
 - [x] nethermind is missing return fields
 - [x] nethermind returns `author` field that should not be there
 
detected by our testing team.

### DIFF
--- a/src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
+++ b/src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
@@ -59,7 +59,12 @@ namespace Nethermind.Evm.TransactionProcessing
             /// <summary>
             /// Commit and later restore state also skip validation, use for CallAndRestore
             /// </summary>
-            CommitAndRestore = Commit | Restore | NoValidation
+            CommitAndRestore = Commit | Restore | NoValidation,
+
+            /// <summary>
+            /// Execute in simulation mode, without making actual changes
+            /// </summary>
+            Simulation = 8
         }
 
         public TransactionProcessor(
@@ -444,7 +449,7 @@ namespace Nethermind.Evm.TransactionProcessing
             out long spentGas,
             out byte statusCode)
         {
-            bool validate = !opts.HasFlag(ExecutionOptions.NoValidation);
+            bool validate = !opts.HasFlag(ExecutionOptions.NoValidation) || opts.HasFlag(ExecutionOptions.Simulation);
 
             substate = null;
             spentGas = tx.GasLimit;
--- a/src/Nethermind/Nethermind.Facade/Proxy/Models/Simulate/SimulateBlockResult.cs
+++ b/src/Nethermind/Nethermind.Facade/Proxy/Models/Simulate/SimulateBlockResult.cs
@@ -15,5 +15,5 @@ public class SimulateBlockResult(Block source, bool includeFullTransactionData, 
     : BlockForRpc(source, includeFullTransactionData, specProvider)
 {
     public List<SimulateCallResult> Calls { get; set; } = new();
-    public UInt256 BlobBaseFee { get; set; }
+    //public UInt256 BlobBaseFee { get; set; }
 }
--- a/src/Nethermind/Nethermind.Facade/Simulate/SimulateBridgeHelper.cs
+++ b/src/Nethermind/Nethermind.Facade/Simulate/SimulateBridgeHelper.cs
@@ -3,9 +3,11 @@
 
 using System;
 using System.Collections.Generic;
+using System.Data.Common;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Runtime.InteropServices;
+using System.Transactions;
 using Nethermind.Blockchain;
 using Nethermind.Config;
 using Nethermind.Consensus.Processing;
@@ -14,10 +16,12 @@ using Nethermind.Core.Collections;
 using Nethermind.Core.Crypto;
 using Nethermind.Core.Specs;
 using Nethermind.Crypto;
+using Nethermind.Evm;
 using Nethermind.Evm.Tracing;
 using Nethermind.Evm.TransactionProcessing;
 using Nethermind.Facade.Proxy.Models.Simulate;
 using Nethermind.Int256;
+using Nethermind.Specs;
 using Nethermind.State;
 using Transaction = Nethermind.Core.Transaction;
 
@@ -58,14 +62,16 @@ public class SimulateBridgeHelper(SimulateReadOnlyBlocksProcessingEnvFactory sim
     public bool TrySimulate(
         BlockHeader parent,
         SimulatePayload<TransactionWithSourceDetails> payload,
+        SimulateBlockTracer simulateOutputTracer,
         IBlockTracer tracer,
         [NotNullWhen(false)] out string? error) =>
-        TrySimulate(parent, payload, tracer, simulateProcessingEnvFactory.Create(payload.Validation), out error);
+        TrySimulate(parent, payload, simulateOutputTracer, tracer, simulateProcessingEnvFactory.Create(payload.Validation), out error);
 
 
     private bool TrySimulate(
         BlockHeader parent,
         SimulatePayload<TransactionWithSourceDetails> payload,
+        SimulateBlockTracer simulateOutputTracer,
         IBlockTracer tracer,
         SimulateReadOnlyBlocksProcessingEnv env,
         [NotNullWhen(false)] out string? error)
@@ -84,6 +90,8 @@ public class SimulateBridgeHelper(SimulateReadOnlyBlocksProcessingEnvFactory sim
             {
                 nonceCache.Clear();
                 BlockHeader callHeader = GetCallHeader(blockCall, parent, payload.Validation, spec); //currentSpec is still parent spec
+
+
                 spec = env.SpecProvider.GetSpec(callHeader);
                 PrepareState(callHeader, parent, blockCall, env.WorldState, env.CodeInfoRepository, spec);
 
@@ -95,9 +103,24 @@ public class SimulateBridgeHelper(SimulateReadOnlyBlocksProcessingEnvFactory sim
                 {
                     callHeader.BaseFeePerGas = 0;
                 }
+
+                env.SetBlockBlobBaseFee(blockCall.BlockOverrides?.BlobBaseFee);
+
                 callHeader.Hash = callHeader.CalculateHash();
 
                 Transaction[] transactions = CreateTransactions(payload, blockCall, callHeader, stateProvider, nonceCache);
+
+
+                if (spec.IsEip4844Enabled)
+                {
+                    foreach (Transaction transaction in transactions)
+                    {
+                        callHeader.BlobGasUsed += BlobGasCalculator.CalculateBlobGas(transaction);
+                    }
+
+                    callHeader.ExcessBlobGas = BlobGasCalculator.CalculateExcessBlobGas(parent, spec);
+                }
+
                 if (!TryGetBlock(payload, env, callHeader, transactions, out Block currentBlock, out error))
                 {
                     return false;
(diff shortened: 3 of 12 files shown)


src/Nethermind/Nethermind.Facade/Proxy/Models/Simulate/SimulateBlockResult.cs
```diff
@@ -15,5 +15,5 @@ public class SimulateBlockResult(Block source, bool includeFullTransactionData,
     : BlockForRpc(source, includeFullTransactionData, specProvider)
 {
     public List<SimulateCallResult> Calls { get; set; } = new();
-    public UInt256 BlobBaseFee { get; set; }
+    //public UInt256 BlobBaseFee { get; set; }
```
LukaszRozmej, line 18: just remove?

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/SimulateTxExecutor.cs
```diff
@@ -47,6 +47,11 @@ protected override SimulatePayload<TransactionWithSourceDetails> Prepare(Simulat
                             callTransactionModel.Type = TxType.EIP1559;
                         }
 
+                        if (callTransactionModel.BlobVersionedHashes is not null)
+                        {
+                            callTransactionModel.Type = TxType.Blob;
+                        }
+
```
LukaszRozmej, line 54: Can we encapsulate all code related to tx type to `GetTxType` method?

src/Nethermind/Nethermind.Facade/Simulate/SimulateTxMutatorTracer.cs
```diff
@@ -17,7 +17,7 @@ namespace Nethermind.Facade.Simulate;
 
 internal sealed class SimulateTxMutatorTracer : TxTracer, ITxLogsMutator
 {
-    public const int ExecutionError = -32015;
+    public const int ExecutionError = -32000;
```
LukaszRozmej, line 20: all error codes are in `Nethermind.JsonRpc.ErrorCodes` class?

src/Nethermind/Nethermind.Facade/Simulate/SimulateBridgeHelper.cs
```diff
@@ -223,7 +262,7 @@ private static BlockHeader GetParent(BlockHeader parent, SimulatePayload<Transac
     {
         Transaction? transaction = transactionDetails.Transaction;
         transaction.SenderAddress ??= Address.Zero;
-        transaction.To ??= Address.Zero;
+        //transaction.To ??= Address.Zero;
```
LukaszRozmej, line 265: remove?

src/Nethermind/Nethermind.Facade/Simulate/SimulateTransactionProcessor.cs
```diff
@@ -20,13 +21,24 @@ public class SimulateTransactionProcessor(
     bool validate)
     : TransactionProcessor(specProvider, worldState, virtualMachine, codeInfoRepository, logManager), ITransactionProcessor
 {
+    public UInt256? BlobBaseFee { get; set; }
```
LukaszRozmej, line 24: Why can't it be passed in BlockExecutionContext already here?

src/Nethermind/Nethermind.Facade/Simulate/SimulateBridgeHelper.cs
```diff
@@ -95,9 +103,24 @@ public class SimulateBridgeHelper(SimulateReadOnlyBlocksProcessingEnvFactory sim
                 {
                     callHeader.BaseFeePerGas = 0;
                 }
+
+                env.SetBlockBlobBaseFee(blockCall.BlockOverrides?.BlobBaseFee);
```
LukaszRozmej, line 107: if (spec.IsEip4844Enabled)?

src/Nethermind/Nethermind.Facade/Simulate/SimulateBridgeHelper.cs
```diff
@@ -117,6 +140,22 @@ public class SimulateBridgeHelper(SimulateReadOnlyBlocksProcessingEnvFactory sim
                         processor.Process(stateProvider.StateRoot, suggestedBlocks, processingFlags, tracer)[0];
 
                 FinalizeStateAndBlock(stateProvider, processedBlock, spec, currentBlock, blockTree);
+                var current = simulateOutputTracer.Results.Last();
+
+                current.StateRoot = processedBlock.StateRoot ?? new Hash256("0x0000000000000000000000000000000000000000000000000000000000000000"); ;
```
LukaszRozmej, line 145: static: `Hash256.Zero`

src/Nethermind/Nethermind.Facade/Simulate/SimulateBridgeHelper.cs
```diff
@@ -117,6 +140,22 @@ public class SimulateBridgeHelper(SimulateReadOnlyBlocksProcessingEnvFactory sim
...
+                                                new Hash256("0x0000000000000000000000000000000000000000000000000000000000000000");
+
+                current.WithdrawalsRoot = processedBlock.WithdrawalsRoot ??
+                                          new Hash256("0x0000000000000000000000000000000000000000000000000000000000000000");
+
+                current.ExcessBlobGas = processedBlock.ExcessBlobGas ?? 0;
+
+                current.Withdrawals = processedBlock.Withdrawals ?? [];
+
+                current.Author = null;
```
LukaszRozmej, line 157: Can you make it into a method?

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -444,7 +449,7 @@ protected virtual TransactionResult IncrementNonce(Transaction tx, BlockHeader h
             out long spentGas,
             out byte statusCode)
         {
-            bool validate = !opts.HasFlag(ExecutionOptions.NoValidation);
+            bool validate = !opts.HasFlag(ExecutionOptions.NoValidation) || opts.HasFlag(ExecutionOptions.Simulation);
```
LukaszRozmej, line 452: instead of adding `ExecutionOptions.Simulation`, can we use some override in SimulateTransactionProcessor? I would like to avoid bloating that enum.

LukaszRozmej (changes requested):
Few minor things

src/Nethermind/Nethermind.Facade/Simulate/SimulateTransactionProcessor.cs
OlegJakushkin: @LukaszRozmej It seems to me the only place where we could control BEC would be in `SimulateBlockValidationTransactionsExecutor`. To implement this change we would need:
 - only use `SimulateBlockValidationTransactionsExecutor` in simulate (see `GetProcessor` in `SimulateBlockValidationTransactionsExecutor`) and bring "`bool validate`" into it
 - `SimulateBlockValidationTransactionsExecutor` will need to overload `BlockValidationTransactionsExecutor.ProcessTransactions` at line 39 

which is rather complicated in comparison to the current approach. Yet we can do it if it is better. Is it?

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
OlegJakushkin: Refactored, please take a look.

src/Nethermind/Nethermind.Facade/Simulate/SimulateBridgeHelper.cs
OlegJakushkin: Done

src/Nethermind/Nethermind.Facade/Simulate/SimulateBridgeHelper.cs
OlegJakushkin: Implemented, please take a look.

src/Nethermind/Nethermind.Facade/Simulate/SimulateBridgeHelper.cs
OlegJakushkin: Thank you added.

src/Nethermind/Nethermind.Facade/Simulate/SimulateBridgeHelper.cs
OlegJakushkin: done

src/Nethermind/Nethermind.Facade/Simulate/SimulateTxMutatorTracer.cs
OlegJakushkin: Done

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/SimulateTxExecutor.cs
OlegJakushkin: Done

src/Nethermind/Nethermind.Facade/Proxy/Models/Simulate/SimulateBlockResult.cs
OlegJakushkin: Done

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -59,7 +59,12 @@ protected enum ExecutionOptions
             /// <summary>
             /// Commit and later restore state also skip validation, use for CallAndRestore
             /// </summary>
-            CommitAndRestore = Commit | Restore | NoValidation
+            CommitAndRestore = Commit | Restore | NoValidation,
+
+            /// <summary>
+            /// Execute in simulation mode, without making actual changes
+            /// </summary>
+            Simulation = 8
```
LukaszRozmej, line 67: Can we drop this, why is it needed?

src/Nethermind/Nethermind.Facade/Simulate/SimulateTransactionProcessor.cs
LukaszRozmej: Yep pushed it myself

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
LukaszRozmej: pushed it

src/Nethermind/Nethermind.Evm/BlobGasCalculator.cs
```diff
@@ -31,12 +31,11 @@ public static ulong CalculateBlobGas(Transaction[] transactions)
 
     public static bool TryCalculateBlobGasPrice(BlockHeader header, Transaction transaction, out UInt256 blobGasPrice)
     {
-        if (!TryCalculateBlobGasPricePerUnit(header.ExcessBlobGas.Value, out UInt256 blobGasPricePerUnit))
-        {
-            blobGasPrice = UInt256.MaxValue;
-            return false;
-        }
-        return !UInt256.MultiplyOverflow(CalculateBlobGas(transaction), blobGasPricePerUnit, out blobGasPrice);
+        if (header.ExcessBlobGas != null &&
```
LukaszRozmej, line 34: I think this shouldn't be called if it is null in the first place

LukaszRozmej: So if any of transactions in block support blobs, you need to set ExcessBlobGas for the BlockHeader

LukaszRozmej: Or better - add it if spec.IsEip4844Enabled

OlegJakushkin: Yep - restored, fixed

PR #7309, opened 2024-08-05
Fix trigger of tests

kamilchodola:
I've used invalid event to distinguish if job should be triggered or not. Fixed

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [X] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

### DIFF
--- a/.github/workflows/sync-supported-chains.yml
+++ b/.github/workflows/sync-supported-chains.yml
@@ -227,7 +227,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Trigger notification action on test repo
-        if: github.event.workflow_run.head_branch == 'master' || contains(github.event.workflow_run.head_branch, 'release/')
+        if: github.event.workflow_run.head_branch == 'master' || startsWith(github.event.workflow_run.head_branch, 'release/')
         continue-on-error: true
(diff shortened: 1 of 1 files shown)


.github/workflows/sync-supported-chains.yml
```diff
@@ -227,7 +227,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Trigger notification action on test repo
-        if: github.event_name == 'push' && (startsWith(github.ref, 'refs/heads/master') || startsWith(github.ref, 'refs/heads/release/'))
+        if: github.event.workflow_run.head_branch == 'master' || contains(github.event.workflow_run.head_branch, 'release/')
```
LukaszRozmej, line 230: not sure about that `contains`, shouldn't be `startswith`?

kamilchodola: Valid point - but hard to verify how branch is provided with workflow_run so will verify shortly after merge if all is ok.

"push" prfixes with "refs/heads" where workflow_run should not but GH never stops surprising me. That's whY i chosen contains at first place but may hit some edge case with weird branches like "test/release/1.28.0" (which can happen)

PR #7319, opened 2024-08-09
Feature t8n gnosis

stdevMac:
## Changes

- Added support for running t8n ethereum/tests over gnosis

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No


src/Nethermind/Ethereum.Test.Base/GeneralTestBase.cs
```diff
@@ -60,25 +60,55 @@ protected static void Setup(ILogManager logManager)
...
 
         protected EthereumTestResult RunTest(GeneralStateTest test, ITxTracer txTracer)
+        {
+            return RunTest(test, false, txTracer);
+        }
+
+        protected EthereumTestResult RunTest(GeneralStateTest test, bool isGnosis, ITxTracer txTracer)
         {
             TestContext.Write($"Running {test.Name} at {DateTime.UtcNow:HH:mm:ss.ffffff}");
+            Console.WriteLine("Passing over here");
```
kamilchodola, line 84: Debug log?

stdevMac: Fixed [here](https://github.com/NethermindEth/nethermind/pull/7319/commits/2a284919c0696a44db6de586226314dcad174e76)

tools/evm/Program.cs
```diff
@@ -36,6 +36,7 @@ static void ConfigureT8NCommand(ref RootCommand rootCmd)
             var traceOpt = new Option<bool>("--trace", description: "Configures the use of the JSON opcode tracer. This tracer emits traces to files as trace-<txIndex>-<txHash>.jsonl", getDefaultValue: () => false);
             var traceNoStackOpt = new Option<bool>("--trace.noStack", description: "Trace no stack", getDefaultValue: () => false);
             var traceReturnDataOpt = new Option<bool>("--trace.returnData", description: "Trace return data", getDefaultValue: () => false);
+            var isGnosisTests = new Option<bool>("--run.gnosis", description: "If set to true, will run under Gnosis spec", getDefaultValue: () => false);
```
dmitriy-b, line 39: Just an idea for an improvement: instead of using a new `--run.gnosis` option maybe we can reuse  `--state.chainId` and choose `ISpecProvider` based on chain / [networkid](https://github.com/NethermindEth/nethermind/blob/master/src/Nethermind/Chains/gnosis.json#L44)? With such a way we will not need to add a new flag every time we have to add a new network to t8n.

yerke26 (approved):

tools/evm/Program.cs
yerke26: Good idea, removed --run.gnosis flag

PR #7322, opened 2024-08-12
Add `engine_getBlobsV1`

marcindsobczak:
## Changes

- Add `engine_getBlobsV1` ([spec](https://github.com/ethereum/execution-apis/pull/559))

~~Used `string` as a key in `BlobIndex`, because more intuitive `byte[]` (as we store `blobVersionedHashes`) were comparing keys by reference and we need to compare by value~~

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

docs will be automatically updated based on the added metadata

- [x] Yes
- [ ] No


src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -182,6 +183,8 @@ public class TxPool : ITxPool, IDisposable
 
         public int GetPendingBlobTransactionsCount() => _blobTransactions.Count;
 
+        public ConcurrentDictionary<string, List<Hash256>> GetBlobIndex() => _blobTransactions.GetBlobIndex;
```
LukaszRozmej, line 186: This breaks encapsulation, IMO tx pool should have a method that given blobhash, will return blob data

src/Nethermind/Nethermind.Merge.Plugin/Data/BlobAndProofV1.cs
```diff
@@ -0,0 +1,24 @@
...
+    public BlobAndProofV1(Transaction blobTx, int index)
+    {
+        if (blobTx is not { NetworkWrapper: ShardBlobNetworkWrapper wrapper })
+        {
+            throw new ArgumentException("Shard blob transaction should contain network wrapper data");
+        }
+
+        Blob = wrapper.Blobs[index];
+        Proof = wrapper.Proofs[index];
+    }
```
LukaszRozmej, line 21: IMO transaction passed here and validation here is bad design, it should happen elsewhere - probably in the tx pool, and maybe we need to load blobs from storage then.

src/Nethermind/Nethermind.TxPool/Collections/BlobTxDistinctSortedPool.cs
```diff
@@ -1,20 +1,94 @@
...
 namespace Nethermind.TxPool.Collections;
 
 public class BlobTxDistinctSortedPool(int capacity, IComparer<Transaction> comparer, ILogManager logManager)
     : TxDistinctSortedPool(capacity, comparer, logManager)
 {
+    protected override string ShortPoolName => "BlobPool";
+
+    public ConcurrentDictionary<string, List<Hash256>> GetBlobIndex => _blobIndex;
+
+    private readonly ConcurrentDictionary<string, List<Hash256>> _blobIndex = new();
```
LukaszRozmej, line 21: se byte[] as key with Bytes.EqualityComparer as supplied comparer for the dictionary.

src/Nethermind/Nethermind.TxPool/Collections/BlobTxDistinctSortedPool.cs
```diff
@@ -1,20 +1,94 @@
...
 using Nethermind.Logging;
 
 namespace Nethermind.TxPool.Collections;
 
 public class BlobTxDistinctSortedPool(int capacity, IComparer<Transaction> comparer, ILogManager logManager)
     : TxDistinctSortedPool(capacity, comparer, logManager)
 {
+    protected override string ShortPoolName => "BlobPool";
+
+    public ConcurrentDictionary<string, List<Hash256>> GetBlobIndex => _blobIndex;
```
LukaszRozmej, line 19: This shouldn't be public

LukaszRozmej (changes requested):

src/Nethermind/Nethermind.Merge.Plugin.Test/EngineModuleTests.V3.cs
```diff
@@ -594,10 +595,11 @@ public async Task GetBlobsV1_should_return_nulls_when_blobs_not_found([Values(1,
         ResultWrapper<GetBlobsV1Result> result = await rpcModule.engine_getBlobsV1(blobTx.BlobVersionedHashes!);
 
         result.Data.Should().BeEquivalentTo(expected);
-        result.Data.BlobsAndProofs.Length.Should().Be(numberOfRequestedBlobs);
+        BlobAndProofV1?[] resultBlobsAndProofs = result.Data.BlobsAndProofs.ToArray();
+        resultBlobsAndProofs.Length.Should().Be(numberOfRequestedBlobs);
         for (int i = 0; i < numberOfRequestedBlobs; i++)
         {
-            result.Data.BlobsAndProofs[i]!.Should().BeNull();
+            resultBlobsAndProofs[i]!.Should().BeNull();
```
LukaszRozmej, line 618: Same here, like there for the whole collection `Should().AllBeEquivalentTo(null)`

src/Nethermind/Nethermind.TxPool/BlobCatcher.cs
```diff
@@ -0,0 +1,39 @@
...
+
+using System.Collections.Generic;
+using System.Linq;
+using Nethermind.Core;
+using Nethermind.Core.Crypto;
+using Nethermind.TxPool.Collections;
+
+namespace Nethermind.TxPool;
+
+public class BlobCatcher(BlobTxDistinctSortedPool blobPool)
```
LukaszRozmej, line 11: I don't get the naming, what are we catching? Maybe `BlobFinder`?

src/Nethermind/Nethermind.TxPool/ITxPool.cs
```diff
@@ -16,7 +15,7 @@ public interface ITxPool
         int GetPendingTransactionsCount();
         int GetPendingBlobTransactionsCount();
         Transaction[] GetPendingTransactions();
-        ConcurrentDictionary<byte[], List<Hash256>> GetBlobIndex();
+        IEnumerable<BlobAndProofV1?> GetBlobsAndProofs(byte[][] blobVersionedHashes);
```
LukaszRozmej, line 18: I think we are messing abstraction layers here:
1. TxPool doesn't need to handle multiple searches directly - would only do one, and handle collections outside of it. 
2. `BlobAndProofV1` is typical RPC type and should not be used in TxPool directly

So it can look like:  `bool TryGetBlob([NotNullWhen(true)] out byte[]? blob, [NotNullWhen(true)] out byte[]? proof)`

src/Nethermind/Nethermind.Merge.Plugin.Test/EngineModuleTests.V3.cs
```diff
@@ -508,6 +510,166 @@ static void MarkAsUnprocessed(MergeTestBlockchain chain, int blockNumber)
...
+
+        if (requestSize > 128)
+        {
+            result.Result.Should().BeEquivalentTo(Result.Fail($"The number of requested blobs must not exceed 128"));
+            result.ErrorCode.Should().Be(MergeErrorCodes.TooLargeRequest);
+        }
+        else
+        {
+            result.Result.Should().Be(Result.Success);
+            result.Data.BlobsAndProofs.ToArray().Length.Should().Be(requestSize);
```
LukaszRozmej, line 535:
```suggestion
            result.Data.BlobsAndProofs.Should().HaveCount(requestSize);
```

src/Nethermind/Nethermind.Merge.Plugin.Test/EngineModuleTests.V3.cs
```diff
@@ -508,6 +510,166 @@ static void MarkAsUnprocessed(MergeTestBlockchain chain, int blockNumber)
...
+        ResultWrapper<GetBlobsV1Result> result = await rpcModule.engine_getBlobsV1(blobTx.BlobVersionedHashes!);
+
+        result.Data.Should().BeEquivalentTo(expected);
+        BlobAndProofV1?[] resultBlobsAndProofs = result.Data.BlobsAndProofs.ToArray();
+        resultBlobsAndProofs.Length.Should().Be(numberOfBlobs);
+        for (int i = 0; i < numberOfBlobs; i++)
+        {
+            resultBlobsAndProofs[i]!.Blob.Should().BeEquivalentTo(((ShardBlobNetworkWrapper)blobTx.NetworkWrapper!).Blobs[i]);
+            resultBlobsAndProofs[i]!.Proof.Should().BeEquivalentTo(((ShardBlobNetworkWrapper)blobTx.NetworkWrapper!).Proofs[i]);
+        }
```
LukaszRozmej, line 586:
```suggestion
        ShardBlobNetworkWrapper wrapper = ((ShardBlobNetworkWrapper)blobTx.NetworkWrapper!);
        resultBlobsAndProofs.Select(b => b.Blob).Should().BeEquivalentTo(wrapper.Blobs);
        resultBlobsAndProofs.Select(b => b.Proof).Should().BeEquivalentTo(wrapper.Proofs);
```

src/Nethermind/Nethermind.TxPool/BlobCatcher.cs
```diff
@@ -0,0 +1,40 @@
...
+    private BlobAndProofV1? GetBlobAndProofV1(byte[] requestedBlobVersionedHash)
+    {
+        if (blobPool.GetBlobIndex.TryGetValue(requestedBlobVersionedHash, out List<Hash256>? txHashes)
+            && txHashes[0] is not null
+            && blobPool.TryGetValue(txHashes[0], out Transaction? blobTx)
+            && blobTx.BlobVersionedHashes?.Length > 0)
+        {
+            for (int indexOfBlob = 0; indexOfBlob < blobTx.BlobVersionedHashes.Length; indexOfBlob++)
+            {
+                if (blobTx.BlobVersionedHashes[indexOfBlob] == requestedBlobVersionedHash
```
LukaszRozmej, line 30: This is reference equals, you probably want `Bytes.AreEqual()`

LukaszRozmej (changes requested):

src/Nethermind/Nethermind.TxPool/BlobCatcher.cs
marcindsobczak: `BlobFinder` was my first thought, but then `BlobCatcher` came to my mind, and I liked it - it sounded cool and a bit badass. I chose it, sacrificing accuracy for style. But after giving it more thought, it’s not as great as in first impression. Agree, Finder is more precise. Changed.

LukaszRozmej (approved):

src/Nethermind/Nethermind.TxPool/Collections/BlobTxDistinctSortedPool.cs
```diff
@@ -1,20 +1,94 @@
...
 using Nethermind.Logging;
 
 namespace Nethermind.TxPool.Collections;
 
 public class BlobTxDistinctSortedPool(int capacity, IComparer<Transaction> comparer, ILogManager logManager)
     : TxDistinctSortedPool(capacity, comparer, logManager)
 {
+    protected override string ShortPoolName => "BlobPool";
+
+    internal ConcurrentDictionary<byte[], List<Hash256>> GetBlobIndex => _blobIndex;
```
LukaszRozmej, line 19: If it is property should be `BlobIndex`

src/Nethermind/Nethermind.TxPool/BlobFinder.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using Nethermind.Core;
+using Nethermind.Core.Crypto;
+using Nethermind.Core.Extensions;
+using Nethermind.TxPool.Collections;
+
+namespace Nethermind.TxPool;
+
+public class BlobFinder(BlobTxDistinctSortedPool blobPool)
```
LukaszRozmej, line 13: Do we even need `BlobFinder` as it uses private parts of `BlobPool`, maybe it should be just under there - method in `BlobTxDistinctSortedPool`?

marcindsobczak: There is one problem. We have all blob index logic in `BlobTxDistinctSortedPool`, but there is also `PersistentBlobTxDistinctSortedPool`, which has a logic of saving/loading blob txs on disc. In `BlobFinder` we are using `TryGetValue`, which has override in `PersistentBlobTxDistinctSortedPool`. If we move a logic from `BlobFinder` to `BlobTxDistinctSortedPool`, we will skip override from `PersistentBlobTxDistinctSortedPool`. By using external class `BlobFinder` and blob pool as an argument, we are always using highest override of `TryGetValue`

src/Nethermind/Nethermind.TxPool/Collections/BlobTxDistinctSortedPool.cs
marcindsobczak: In current version `_blobIndex` is private and we use `GetBlobIndex` for external calls. So I think it should be `_blobIndex` then. But we can also drop `GetBlobIndex` and just use `_blobIndex` as property - and then rename to `BlobIndex`. I will change it

src/Nethermind/Nethermind.TxPool/BlobFinder.cs
marcindsobczak: I had some issues with overrides when implementing blob pool, but maybe it was another case. I'm not fully convinced that it will work like that, but I believe you and let's try. I will remove `BlobFinder` at all

marcindsobczak: It is working perfectly fine, I had completely wrong assumptions. Thanks for caching it!

LukaszRozmej (approved):

src/Nethermind/Nethermind.Merge.Plugin/Data/GetBlobsV1Result.cs
```diff
@@ -0,0 +1,12 @@
...
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System.Collections.Generic;
+using Nethermind.TxPool;
+
+namespace Nethermind.Merge.Plugin.Data;
+
+public class GetBlobsV1Result(IEnumerable<BlobAndProofV1?> blobsAndProofs)
+{
+    public readonly IEnumerable<BlobAndProofV1?> BlobsAndProofs = blobsAndProofs;
```
rubo, line 11: Nitpick: If no specific reason (I don't see one), this member should be a get-only property instead of a read-only field, according to Microsoft's guidelines.

```suggestion
    public IEnumerable<BlobAndProofV1?> BlobsAndProofs { get; } = blobsAndProofs;
```

rubo (approved):

tanishqjasoria (approved):

src/Nethermind/Nethermind.TxPool/Collections/BlobTxDistinctSortedPool.cs
```diff
@@ -1,20 +1,119 @@
...
+
+    protected void AddToBlobIndex(Transaction blobTx)
+    {
+        if (blobTx.BlobVersionedHashes?.Length > 0)
+        {
+            foreach (var blobVersionedHash in blobTx.BlobVersionedHashes)
+            {
+                if (blobVersionedHash?.Length == KzgPolynomialCommitments.BytesPerBlobVersionedHash)
+                {
+                    BlobIndex.AddOrUpdate(blobVersionedHash,
```
karalabe, line 70: This can be used my a malicious transaction to corrupt the mapping by sending the same blobs as a benign tx. Then send a replacement to nuke out all blobs, and the original correct mapping disappears, making the blob non-retrievable any more.

MarekM25: Thanks @karalabe🙏 , we’ll discuss it internally.

karalabe: @LukaszRozmej Said it's fine because the BlobIndex maps to a list of tx hashes, so should be ok.

PR #7327, opened 2024-08-13
Prewarm txns in order (of processing)

benaadams:
Contributes to https://github.com/NethermindEth/nethermind/issues/7171

## Changes

- Process transactions in order (the order that they will be required by main thread), rather than the partitioning scheme from Parallel.For

Partially resolves slow down on Windows

image

Before

image

After

image

- [x] Optimization

#### Requires testing

- [x] No

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
@@ -134,7 +134,7 @@ public class BlockCachePreWarmer(ReadOnlyTxProcessingEnvFactory envFactory, ISpe
                         scope.WorldState.WarmUp(tx.AccessList); // eip-2930
                     }
                     TransactionResult result = scope.TransactionProcessor.Trace(systemTransaction, new BlockExecutionContext(block.Header.Clone()), NullTxTracer.Instance);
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
```diff
@@ -98,29 +109,36 @@ void WarmupWithdrawals(ParallelOptions parallelOptions, IReleaseSpec spec, Block
...
         {
             if (parallelOptions.CancellationToken.IsCancellationRequested) return;
-            Parallel.For(0, block.Transactions.Length, parallelOptions, i =>
-            {
-                // If the transaction has already been processed or being processed, exit early
-                if (block.TransactionProcessed >= i) return;
 
+            int progress = 0;
+            // We want to start at 1 which is the second transaction, giving total count one less than the length
+            Parallel.For(1, block.Transactions.Length, parallelOptions, _ =>
```
LukaszRozmej, line 115: Let's not complicate things and keep starting at 0.
Will be simpler for more changes in the future too.

LukaszRozmej (approved):

smartprogrammer93 (approved):

Issue #6942, opened 2024-04-23
[Logs] No log info abotu invalid parameter applied


kamilchodola:
When executing command `./nethermind --JsonRpc.WebsocketsPort 8546`
Only thing I'm getting is this output:
```
2024-04-23 10-38-04.6152|Nethermind starting initialization.
2024-04-23 10-38-04.6402|Client version: Nethermind/v1.26.0-exp.4+62849a41/linux-x64/dotnet8.0.4
2024-04-23 10-38-04.6459|Loading embedded plugins
2024-04-23 10-38-04.6459|  Found plugin type Nethermind.Consensus.AuRa.AuRaPlugin
2024-04-23 10-38-04.6459|  Found plugin type Nethermind.Consensus.Clique.CliquePlugin
2024-04-23 10-38-04.6459|  Found plugin type Nethermind.Consensus.Ethash.EthashPlugin
2024-04-23 10-38-04.6460|  Found plugin type Nethermind.Consensus.Ethash.NethDevPlugin
2024-04-23 10-38-04.6460|  Found plugin type Nethermind.Hive.HivePlugin
2024-04-23 10-38-04.6460|  Found plugin type Nethermind.UPnP.Plugin.UPnPPlugin
Resolved executing directory as /root.
2024-04-23 10-38-04.6610|Loading 12 assemblies from /root/plugins
2024-04-23 10-38-04.6611|Loading assembly Nethermind.Optimism
2024-04-23 10-38-04.6649|  Found plugin type Nethermind.Optimism
2024-04-23 10-38-04.6649|Loading assembly Nethermind.Api
2024-04-23 10-38-04.6656|Loading assembly Nethermind.EthStats
2024-04-23 10-38-04.6660|  Found plugin type Nethermind.EthStats
2024-04-23 10-38-04.6661|Loading assembly Nethermind.Merge.AuRa
2024-04-23 10-38-04.6665|  Found plugin type Nethermind.Merge.AuRa
2024-04-23 10-38-04.6665|Loading assembly Nethermind.Mev
2024-04-23 10-38-04.6680|  Found plugin type Nethermind.Mev
2024-04-23 10-38-04.6681|Loading assembly Nethermind.Merge.Plugin
2024-04-23 10-38-04.6702|  Found plugin type Nethermind.Merge.Plugin
2024-04-23 10-38-04.6702|Loading assembly Nethermind.Consensus.AuRa
[...]

PR #7331, opened 2024-08-15
Validate and log when an unrecognized parameter/flag is used

wugalde19:
Resolves https://github.com/NethermindEth/nethermind/issues/6942

## Changes
Changes made into `src/Nethermind/Nethermind.Runner/Program.cs`

- Add a function to load all possible flags/params/arguments to be used and put them on a hash
  - The hash is initialized with the default flags allowed in the app and not related to flags in config files
- Add a function to validate current args against the hash generated
- Call the validation function into `Run` so that we initialize the app and validate before start running most of the logic.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

#### Notes on testing

I did not see any test file for Program.cs files

- [ ] Yes
- [X] No

- [ ] Yes
- [ ] No

## Remarks

I recorded a video showing how it works and the error log is displayed when the argument/option/flag used is not valid and when it is valid.
Also, the cases where no flag is provided and when a default app flag (like `--help`) is provided, and when we try using duplicate arguments

https://www.loom.com/share/0ac1011bac1247b992a9fbadba9e5b26?sid=eb48d13b-961e-490e-a1ff-5fe51a2580b0

List of commands used:
```
# Should work
./publish/nethermind --help

# Should work
./publish/nethermind -v

# Should work
./publish/nethermind

# Should work
./publish/nethermind --JsonRpc.WebSocketsPort 8546

# Duplicate argument error
./publish/nethermind --JsonRpc.WebSocketsPort 8546 --JsonRpc.WebSocketsPort 8546

# Unrecognized argument error
./publish/nethermind --JsonRpc.WebsocketsPort 8546
```

### DIFF
--- a/src/Nethermind/Nethermind.Runner/Program.cs
+++ b/src/Nethermind/Nethermind.Runner/Program.cs
@@ -587,26 +587,26 @@ public static class Program
         return info.ToString();
     }
 
-    private static HashSet<string> GetValidArguments(CommandLineApplication app)
-    {
-        var validArguments = new HashSet<string>();
-        foreach (var option in app.GetOptions())
-        {
-            if (!string.IsNullOrEmpty(option.LongName))
-            {
-                validArguments.Add(option.LongName);
-            }
-            if (!string.IsNullOrEmpty(option.ShortName))
-            {
-                validArguments.Add(option.ShortName);
-            }
-        }
-
-        return validArguments;
-    }
-
     private static bool ValidateArguments(string[] args, CommandLineApplication app)
     {
+        static HashSet<string> GetValidArguments(CommandLineApplication app)
+        {
+            var validArguments = new HashSet<string>();
+            foreach (var option in app.GetOptions())
+            {
+                if (!string.IsNullOrEmpty(option.LongName))
+                {
+                    validArguments.Add(option.LongName);
+                }
+                if (!string.IsNullOrEmpty(option.ShortName))
+                {
+                    validArguments.Add(option.ShortName);
+                }
+            }
+
+            return validArguments;
+        }
+
         // Get all valid options from the configuration files
         var validArguments = GetValidArguments(app);
 
--- a/src/Nethermind/Nethermind.Runner/Program.cs
+++ b/src/Nethermind/Nethermind.Runner/Program.cs
@@ -590,7 +590,9 @@ public static class Program
     private static bool ValidateArguments(string[] args, CommandLineApplication app)
     {
         // Get all valid options from the configuration files
-        var validArguments = app.GetOptions().SelectMany(o => new[] { o.LongName, o.ShortName }).ToArray();
+        var validArguments = new HashSet<string>(
+            app.GetOptions().SelectMany(o => new[] { o.LongName, o.ShortName }).Where(name => !string.IsNullOrEmpty(name))
+        );
 
         var argumentsNamesProvided = GetArgumentNames(args).ToList();
         foreach (var argumentName in argumentsNamesProvided)
--- a/src/Nethermind/Nethermind.Runner/Program.cs
+++ b/src/Nethermind/Nethermind.Runner/Program.cs
@@ -599,12 +599,10 @@ public static class Program
             }
         }
 
-        static IEnumerable<ReadOnlyMemory<char>> GetDuplicateArguments(IEnumerable<ReadOnlyMemory<char>> argumentsNames)
-        {
-            return argumentsNames.GroupBy(n => n, new MemoryContentsComparer<char>())
+        static IEnumerable<ReadOnlyMemory<char>> GetDuplicateArguments(IEnumerable<ReadOnlyMemory<char>> argumentsNames) =>
+            argumentsNames.GroupBy(n => n, new MemoryContentsComparer<char>())
                 .Where(g => g.Count() > 1)
                 .Select(g => g.Key);
-        }
 
         // Get all valid options from the configuration files
         HashSet<ReadOnlyMemory<char>> validArguments = GetValidArguments(app);
--- a/src/Nethermind/Nethermind.Runner/Program.cs
+++ b/src/Nethermind/Nethermind.Runner/Program.cs
@@ -114,6 +114,8 @@ public static class Program
         _logger.Info("Nethermind starting initialization.");
         _logger.Info($"Client version: {ProductInfo.ClientId}");
 
+        ValidateCommandLineFlags(args);
+
         string duplicateArgumentsList = string.Join(", ", GetDuplicateArguments(args));
         if (!string.IsNullOrEmpty(duplicateArgumentsList))
         {
@@ -590,5 +592,81 @@ public static class Program
             .Append("Runtime: ").AppendLine(ProductInfo.Runtime);
 
         return info.ToString();
+    }
+
+    private static HashSet<string> generateValidConfigOptionsHash()
+    {
+        // Initialize with the default options
+        var options = new HashSet<string>
+        {
+            "-?",
+            "-h",
+            "--help",
+            "-v",
+            "--version",
+            "-dd",
+            "--datadir",
+            "-c",
+            "--config",
+            "-d",
+            "--baseDbPath",
+            "-l",
+            "--log",
+            "-cd",
+            "--configsDirectory",
+            "-lcs",
+            "--loggerConfigSource",
+            "-pd",
+            "--pluginsDirectory",
+        };
+
+        // Add all the options from the configuration files
+        Type configurationType = typeof(IConfig);
+        IEnumerable<Type> configTypes = TypeDiscovery.FindNethermindBasedTypes(configurationType)
+            .Where(ct => ct.IsInterface);
+
+        foreach (Type configType in configTypes.Where(ct => !ct.IsAssignableTo(typeof(INoCategoryConfig))).OrderBy(c => c.Name))
+        {
+            if (configType is null)
+            {
+                continue;
+            }
+
+            ConfigCategoryAttribute? typeLevel = configType.GetCustomAttribute<ConfigCategoryAttribute>();
+
+            if (typeLevel is not null && (typeLevel?.DisabledForCli ?? true))
+            {
+                continue;
+            }
+
+            foreach (PropertyInfo propertyInfo in configType
+                            .GetProperties(BindingFlags.Public | BindingFlags.Instance)
+                            .OrderBy(p => p.Name))
+            {
+                ConfigItemAttribute? configItemAttribute = propertyInfo.GetCustomAttribute<ConfigItemAttribute>();
+                if (!(configItemAttribute?.DisabledForCli ?? false))
+                {
+                    options.Add($"--{configType.Name[1..].Replace("Config", string.Empty)}.{propertyInfo.Name}");
+                }
+            }
+        }
+
+        return options;
+    }
+
+    private static void ValidateCommandLineFlags(string[] args)
+    {
+        // Get all valid options from the configuration files
+        var validParameters = generateValidConfigOptionsHash();
+
+        foreach (var arg in args)
+        {
+            // Check if the argument is a parameter (starts with --)
+            if (arg.StartsWith("--") && !validParameters.Contains(arg))
+            {
+                string message = $"Unrecognized parameter: {arg}.\nRun --help for a list of available options and commands.";
+                throw new ArgumentException(message);
+            }
+        }
     }
 }
--- a/src/Nethermind/Nethermind.Runner/Program.cs
+++ b/src/Nethermind/Nethermind.Runner/Program.cs
@@ -587,12 +587,28 @@ public static class Program
         return info.ToString();
     }
 
+    private static HashSet<string> GetValidArguments(CommandLineApplication app)
+    {
+        var validArguments = new HashSet<string>();
+        foreach (var option in app.GetOptions())
+        {
+            if (!string.IsNullOrEmpty(option.LongName))
+            {
+                validArguments.Add(option.LongName);
+            }
+            if (!string.IsNullOrEmpty(option.ShortName))
+            {
+                validArguments.Add(option.ShortName);
+            }
+        }
+
+        return validArguments;
+    }
+
     private static bool ValidateArguments(string[] args, CommandLineApplication app)
     {
         // Get all valid options from the configuration files
-        var validArguments = new HashSet<string>(
-            app.GetOptions().SelectMany(o => new[] { o.LongName, o.ShortName }).Where(name => !string.IsNullOrEmpty(name))
-        );
+        var validArguments = GetValidArguments(app);
 
         var argumentsNamesProvided = GetArgumentNames(args).ToList();
         foreach (var argumentName in argumentsNamesProvided)
--- a/src/Nethermind/Nethermind.Runner/Program.cs
+++ b/src/Nethermind/Nethermind.Runner/Program.cs
@@ -114,9 +114,6 @@ public static class Program
         _logger.Info("Nethermind starting initialization.");
         _logger.Info($"Client version: {ProductInfo.ClientId}");
 
-        bool validArgs = ValidateArguments(args);
-        if (!validArgs) return;
-
         AppDomain.CurrentDomain.ProcessExit += CurrentDomainOnProcessExit;
         AssemblyLoadContext.Default.ResolvingUnmanagedDll += OnResolvingUnmanagedDll;
 
@@ -154,6 +151,9 @@ public static class Program
         TypeDiscovery.Initialize(typeof(INethermindPlugin));
 
         BuildOptionsFromConfigFiles(app);
+
+        bool validArgs = ValidateArguments(args, app);
+        if (!validArgs) return;
 
         app.OnExecute(async () =>
         {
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Runner/Program.cs
```diff
@@ -591,4 +593,57 @@ private static string GetProductInfo()
...
+                {
+                    options.Add($"--{configType.Name[1..].Replace("Config", string.Empty)}.{propertyInfo.Name}");
+                }
+            }
+        }
+
+        return options;
+    }
+
+    private static void ValidateCommandLineFlags(string[] args)
```
LukaszRozmej, line 634: If we are validating here I would also move `GetDuplicateArguments` and duplication validation under this one method.

src/Nethermind/Nethermind.Runner/Program.cs
```diff
@@ -591,4 +593,57 @@ private static string GetProductInfo()
...
+    {
+        // Get all valid options from the configuration files
+        var validParameters = generateValidConfigOptionsHash();
+
+        foreach (var arg in args)
+        {
+            // Check if the argument is a parameter (starts with --)
+            if (arg.StartsWith("--") && !validParameters.Contains(arg))
```
LukaszRozmej, line 642: In `GetDuplicateArguments` we actually already have local method `GetArgumentNames` that should be reused here.

src/Nethermind/Nethermind.Runner/Program.cs
```diff
@@ -591,4 +593,57 @@ private static string GetProductInfo()
 
         return info.ToString();
     }
+
+    private static HashSet<string> generateValidConfigOptionsHash()
```
LukaszRozmej, line 597: We use PascalCase in C# for method names.

src/Nethermind/Nethermind.Runner/Program.cs
```diff
@@ -591,4 +593,57 @@ private static string GetProductInfo()
...
+                ConfigItemAttribute? configItemAttribute = propertyInfo.GetCustomAttribute<ConfigItemAttribute>();
+                if (!(configItemAttribute?.DisabledForCli ?? false))
+                {
+                    options.Add($"--{configType.Name[1..].Replace("Config", string.Empty)}.{propertyInfo.Name}");
+                }
+            }
+        }
+
+        return options;
+    }
```
LukaszRozmej, line 632: Wouldn't it be better to use `CommandLineApplication.GetOptions()` to get all options and check on those instead of parsing them again here from config? And not all options are from configs, some are added explicitly!

LukaszRozmej (changes requested):

src/Nethermind/Nethermind.Runner/Program.cs
wugalde19: Hey @LukaszRozmej 
I just run a diff against what I had pushed and seems like I forgot to push the most updated version of the latest commit
I did actually noticed that some options are added explicitly and for that reason I added this
image

But not that you mention the `.GetOptions()` methods seems like a better way to get them
Gonna give it a try 
Thanks

wugalde19: You were right @LukaszRozmej 
I didn't need a new method to get all the options, the way you proposed did it for me
Here's a sample of the options I printed in the console to test
image

src/Nethermind/Nethermind.Runner/Program.cs
wugalde19: @LukaszRozmej I used it and it seems to work just fine 👍🏻

wugalde19:
@LukaszRozmej I've already addressed all your comments and updated the video to validate new changes
Feel free to review it again and let me know if you have more feedback
Thanks

src/Nethermind/Nethermind.Runner/Program.cs
```diff
@@ -591,4 +586,32 @@ private static string GetProductInfo()
 
         return info.ToString();
     }
+
+    private static bool ValidateArguments(string[] args, CommandLineApplication app)
+    {
+        // Get all valid options from the configuration files
+        var validArguments = app.GetOptions().SelectMany(o => new[] { o.LongName, o.ShortName }).ToArray();
+
+        var argumentsNamesProvided = GetArgumentNames(args).Select(a => a.ToString()).ToArray();
```
LukaszRozmej, line 595: You don't need to allocate strings, `MemoryContentsComparer` can compare `Memory<char>` directly without additional allocations

src/Nethermind/Nethermind.Runner/Program.cs
```diff
@@ -591,4 +586,32 @@ private static string GetProductInfo()
...
+
+        foreach (var argumentName in argumentsNamesProvided)
+        {
+            // Check if the argument provided is a valid option/argument
+            if (!validArguments.Contains(argumentName))
```
LukaszRozmej, line 600: `validArguments` should be `HashSet` for quick lookup. Right now your code is O(N^2), with `HashSet` it will be O(N)

src/Nethermind/Nethermind.Runner/Program.cs
```diff
@@ -591,4 +586,32 @@ private static string GetProductInfo()
...
+            {
+                string message = $"Unrecognized argument: {argumentName}.\nRun --help for a list of available options and commands.";
+                throw new ArgumentException(message);
+            }
+        }
+
+        string duplicateArgumentsList = string.Join(", ", GetDuplicateArguments(args));
```
LukaszRozmej, line 607: Don't pass args here again, pass `argumentsNamesProvided`

src/Nethermind/Nethermind.Runner/Program.cs
LukaszRozmej: Avoid `ToArray`, either keep it `IEnumerable` or use `ToPooledList` and use pooled memory (need to be disposed with `using` statement).

src/Nethermind/Nethermind.Runner/Program.cs
```diff
@@ -591,4 +586,32 @@ private static string GetProductInfo()
 
         return info.ToString();
     }
+
+    private static bool ValidateArguments(string[] args, CommandLineApplication app)
+    {
+        // Get all valid options from the configuration files
+        var validArguments = app.GetOptions().SelectMany(o => new[] { o.LongName, o.ShortName }).ToArray();
```
LukaszRozmej, line 593: avoid creating intermittent arrays of `new[] { o.LongName, o.ShortName }` with a simple loop.

src/Nethermind/Nethermind.Runner/Program.cs
```diff
@@ -256,29 +251,29 @@ private static void Run(string[] args)
         }
     }
 
-    private static IEnumerable<ReadOnlyMemory<char>> GetDuplicateArguments(string[] args)
+    private static ReadOnlyMemory<char> GetArgumentName(string arg) => arg.StartsWith("--") ? arg.AsMemory(2) : arg.StartsWith('-') ? arg.AsMemory(1) : ReadOnlyMemory<char>.Empty;
+    private static IEnumerable<ReadOnlyMemory<char>> GetArgumentNames(IEnumerable<string> args)
```
LukaszRozmej, line 255: Make those methods local in `ValidateArguments`, they don't make sense outside of it.

LukaszRozmej (changes requested):
Better design, but way to much allocations.

src/Nethermind/Nethermind.Runner/Program.cs
LukaszRozmej: Yep, didn't know it was an option until I looked for it either. But it makes sense it will be there.

src/Nethermind/Nethermind.Runner/Program.cs
wugalde19: Good call
No need to process/get the names again if we already did it

src/Nethermind/Nethermind.Runner/Program.cs
wugalde19: My main concern when I extracted those was `GetDuplicateArguments` not being able to reach that code.
I should have tested it before.
Thanks
Done

wugalde19:
@LukaszRozmej Thanks for all the feedback provided
Definitely learning a lot and it is forcing me to be more thoughtful with the code
Ready for another round of reviews

src/Nethermind/Nethermind.Runner/Program.cs
```diff
@@ -160,6 +152,9 @@ private static void Run(string[] args)
 
         BuildOptionsFromConfigFiles(app);
 
+        bool validArgs = ValidateArguments(args, app);
+        if (!validArgs) return;
```
LukaszRozmej, line 156: minor: would drop `validArgs` variable it doesn't add anything.
```suggestion
        if (!ValidateArguments(args, app)) return;
```

src/Nethermind/Nethermind.Runner/Program.cs
```diff
@@ -591,4 +565,69 @@ private static string GetProductInfo()
...
+                {
+                    lastWasArgument = false;
+                }
+            }
+        }
+
+        // Get all valid options from the configuration files
+        var validArguments = GetValidArguments(app);
+
+        var argumentsNamesProvided = GetArgumentNames(args).ToList();
```
LukaszRozmej, line 613: Try not to use `var` unless the types are anonymous or very long and obvious, makes harder to read and understand what is going on in GitHub

src/Nethermind/Nethermind.Runner/Program.cs
```diff
@@ -591,4 +565,69 @@ private static string GetProductInfo()
...
+        foreach (var argumentName in argumentsNamesProvided)
+        {
+            // Check if the argument provided is a valid option/argument
+            if (!validArguments.Contains(argumentName.ToString()))
```
LukaszRozmej, line 617: Ah we need that `ToString` conversion here...
But maybe we can made `HashSet<ReadOnlyMemory<char>>` to avoid it?

src/Nethermind/Nethermind.Runner/Program.cs
```diff
@@ -591,4 +565,69 @@ private static string GetProductInfo()
...
+                    validArguments.Add(option.LongName);
+                }
+                if (!string.IsNullOrEmpty(option.ShortName))
+                {
+                    validArguments.Add(option.ShortName);
+                }
+            }
+
+            return validArguments;
+        }
```
LukaszRozmej, line 587:
```suggestion
        static HashSet<ReadOnlyMemory<char>> GetValidArguments(CommandLineApplication app)
        {
            HashSet<ReadOnlyMemory<char>> validArguments = new(new MemoryContentsComparer<char>());
            foreach (var option in app.GetOptions())
            {
                if (!string.IsNullOrEmpty(option.LongName))
                {
                    validArguments.Add(option.LongName.AsMemory());
                }
                if (!string.IsNullOrEmpty(option.ShortName))
                {
                    validArguments.Add(option.ShortName.AsMemory());
                }
            }

            return validArguments;
        }
```

src/Nethermind/Nethermind.Runner/Program.cs
LukaszRozmej, line 613: I would either:
1. Leave it as `IEnumerable` and drop `ToList` - it can complain about re-enumeerating it, but as it is all in-memory there is no big downside
2. As it is local variable, use pooled memory - to avoid creating temp object:
```suggestion
        using ArrayPoolList<ReadOnlyMemory<char>> argumentsNamesProvided = GetArgumentNames(args).ToPooledList();
```

LukaszRozmej (commented):
Cool, IMO code now is quite good, would apply few final touches to make it perfect ;)

src/Nethermind/Nethermind.Runner/Program.cs
```diff
@@ -256,30 +251,9 @@ private static void Run(string[] args)
         }
     }
 
-    private static IEnumerable<ReadOnlyMemory<char>> GetDuplicateArguments(string[] args)
+    private static IEnumerable<ReadOnlyMemory<char>> GetDuplicateArguments(IEnumerable<ReadOnlyMemory<char>> argumentsNames)
```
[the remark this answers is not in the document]
wugalde19: You're right.
Done

src/Nethermind/Nethermind.Runner/Program.cs
wugalde19: I did try but it didn't work the way I did it yesterday, however, today I realized that I had to tweak the implementation of `GetValidArguments` a little bit to make it work, and it works now

src/Nethermind/Nethermind.Runner/Program.cs
wugalde19: Interesing that solving previous comment I did apply those same changes you are suggesting
Except they way you are initializing the hash

Your suggestion
```
HashSet<ReadOnlyMemory<char>> validArguments = new(new MemoryContentsComparer<char>());
```

My code 
```
HashSet<ReadOnlyMemory<char>> validArguments = new HashSet<ReadOnlyMemory<char>>(new MemoryContentsComparer<char>());
```


I changed it to use your way to initialize it (simpler)

src/Nethermind/Nethermind.Runner/Program.cs
wugalde19: Not quite sure I understand
Do you mean avoid using var and try to set an explicit type when possible?

The last part `makes harder to read and understand what is going on in GitHub` it's the part that confuses me

wugalde19: When I drop the `ToList` and the reason why I'm using it is to avoid this error
image

wugalde19: Using option 2 `using ArrayPoolList<ReadOnlyMemory<char>> argumentsNamesProvided = GetArgumentNames(args).ToPooledList();` gives me this other error
image

wugalde19: I decided to go with option 1 and replaced that line with the following
```
IEnumerable<ReadOnlyMemory<char>> argumentsNamesProvided = GetArgumentNames(args);
```

wugalde19:
@LukaszRozmej one step closer to a better implementation
You are great
Thanks for your patience

LukaszRozmej (approved):

src/Nethermind/Nethermind.Runner/Program.cs
LukaszRozmej: `ToPooledList` is an extension method in `Nethermind.Core.Extensions.EnumerableExtensions`

LukaszRozmej:
Congrats for persistence ;)

src/Nethermind/Nethermind.Runner/Program.cs
```diff
@@ -591,4 +557,76 @@ private static string GetProductInfo()
...
+        // Get all valid options from the configuration files
+        HashSet<ReadOnlyMemory<char>> validArguments = GetValidArguments(app);
+
+        IEnumerable<ReadOnlyMemory<char>> argumentsNamesProvided = GetArgumentNames(args);
+        foreach (ReadOnlyMemory<char> argumentName in argumentsNamesProvided)
+        {
+            // Check if the argument provided is a valid option/argument
+            if (!validArguments.Contains(argumentName))
+            {
+                throw new ArgumentException($"Unrecognized argument: {argumentName}.\nRun --help for a list of available options and commands.");
```
LukaszRozmej, line 618: One more thing, maybe better than throwing here would be doing something similar as in duplicate arguments:
1. logging an error.
2. setting ExitCode to specialized (new) exit code
3. returning false.

wugalde19: Actually, that's a great suggestion and makes a ton of sense
Let me add that real quick

wugalde19:
@LukaszRozmej Just applied changes suggested on last two comments

LukaszRozmej (approved):

src/Nethermind/Nethermind.Runner/Program.cs
stdevMac: @wugalde19 on Github, you can't infer the type or jump into the function definition to see what it is. 

It makes sense to user var if it is evident, like GetString(), that it should return some string. 

For the example in question, what does return `GetValidArguments`? The context of the changes does not make it clear.  

Makes more sense? I think it's what @LukaszRozmej was referring to.

wugalde19: Ohhhh I see
Of course, the idea is to provide more context of what a variable will contain unless it's easy to infer.
Yeah yeah makes sense
And that part about Github it's like that you cannot jump on Github between definitions easily as you can do it in some IDEs which makes things harder in terms of knowing/checking what a method returns and all of that

Perfect
Thanks for clarifying this to me @stdevMac

LukaszRozmej:
@wugalde19 thank you for your contribution 🫡

benaadams:
Fails for?
```
 --JsonRpc.JwtSecretFile=C:/ethereum/jwt.hex
```

🤔

wugalde19:
My pleasure @LukaszRozmej

benaadams:
Failed due to unrecognized argument - [JsonRpc.JwtSecretFile=C:/ethereum/jwt.hex].
Run --help for a list of available options and commands.

wugalde19:
> JsonRpc.JwtSecretFile=C:/ethereum/jwt.hex

@benaadams Please try again
@LukaszRozmej came out with a much better solution ([here](https://github.com/NethermindEth/nethermind/commit/ebdf27ac23faa6381f37c75f3b716c85e8e1c26b))
Those changes seemed to have fixed the issue you were having

PR #7328, opened 2024-08-14
Keccak pipelined calculator

Scooletz:
This PR proposes a new way of handling `Instruction.KECCAK256`, when it's followed by some instructions that are not result dependent (no immediate pop of the value). It does it by moving the computation to a separate component with a separate thread to compute and placing a special marker in `EvmStack`, so that when the value needs to be retrieved, it's retrieved from the `OffHeapStack`. You can think of it as a **CPU pipeline** enabler, that does something behind the scenes and materialized value on the stack when ready.

The preliminary benchmarks, that compare actual execution with the offloading, show the following difference

| Method           | Mean      | Error    | StdDev   | Ratio |
|----------------- |----------:|---------:|---------:|------:|
| ValueKeccak      | 209.94 ns | 2.212 ns | 2.069 ns |  1.00 |
| KeccakCalculator |  22.80 ns | 0.469 ns | 0.439 ns |  0.11 |

This, of course, does not take into consideration the overhead of checking in `EvmStack` whether or not the value is kept `OffHeap` and detecting what to do. This is just a rough estimation what could be done.

To reflect on a real example, `WrappedEth` was used: https://app.dedaub.com/ethereum/address/0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2/disassembled Here are some of the sequences that followed SHA3

```
0x5fa: SHA3      
0x5fb: DUP2      
0x5fc: SWAP1     
0x5fd: SSTORE    
```

```
0x5bd: SHA3      
0x5be: PUSH1     0x0
0x5c0: DUP6      
0x5c1: PUSH20    0xffffffffffffffffffffffffffffffffffffffff
0x5d6: AND       
0x5d7: PUSH20    0xffffffffffffffffffffffffffffffffffffffff
0x5ec: AND       
0x5ed: DUP2      
0x5ee: MSTORE 
```

A terrible case:

```
0x6ce: SHA3      
0x6cf: SLOAD
```

```
  0x773: SHA3      
  0x774: PUSH1     0x0
  0x776: CALLER    
  0x777: PUSH20    0xffffffffffffffffffffffffffffffffffffffff
  0x78c: AND       
  0x78d: PUSH20    0xffffffffffffffffffffffffffffffffffffffff
  0x7a2: AND       
  0x7a3: DUP2      
  0x7a4: MSTORE
```

## Changes

- _List the changes_

#### What types of changes does your code introduce?

- [ ] Bugfix (a non-breaking change that fixes an issue)
[...]

PR #7333, opened 2024-08-15
Memoize Keccaks

benaadams:
Follow up to https://github.com/NethermindEth/nethermind/pull/7328

## Changes

- Memoize Keccaks (.Length <= 32)
    - 16 way partitioned ClockCache (LRU); partitioned to reduce contention (adds are frequent)
    - Though very random same Keccak can be used multiple times, whether from
      - Parallel tx exectuions (same code run twice)
      - Multiple use of same contract keccak(address)
      - Same slot across many contracts in solidity mapping i.e. keccak(slot)
- Reduce ClockCache contention
   - Calling ConcurrentDictionary.Count takes locks and causes contention on read side
   - Use local count that is only changed when have write lock anyway
- More variance in Bytes GetHashCode

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Core/Caching/ClockCache.cs
+++ b/src/Nethermind/Nethermind.Core/Caching/ClockCache.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 
 using Nethermind.Core.Threading;
@@ -74,7 +75,8 @@ public sealed class ClockCache<TKey, TValue>(int maxCapacity) : ClockCacheBase<T
             return false;
         }
 
-        int offset = _cacheMap.Count;
+        int offset = _count;
+        Debug.Assert(_cacheMap.Count == _count);
         if (FreeOffsets.Count > 0)
         {
             offset = FreeOffsets.Dequeue();
@@ -86,6 +88,8 @@ public sealed class ClockCache<TKey, TValue>(int maxCapacity) : ClockCacheBase<T
 
         _cacheMap[key] = new LruCacheItem(offset, val);
         KeyToOffset[offset] = key;
+        _count++;
+        Debug.Assert(_cacheMap.Count == _count);
 
         return true;
     }
@@ -93,7 +97,7 @@ public sealed class ClockCache<TKey, TValue>(int maxCapacity) : ClockCacheBase<T
     private int Replace(TKey key)
     {
         int position = Clock;
-        int max = _cacheMap.Count;
+        int max = _count;
         while (true)
         {
             if (position >= max)
@@ -108,6 +112,7 @@ public sealed class ClockCache<TKey, TValue>(int maxCapacity) : ClockCacheBase<T
                 {
                     ThrowInvalidOperationException();
                 }
+                _count--;
                 break;
             }
 
@@ -132,6 +137,7 @@ public sealed class ClockCache<TKey, TValue>(int maxCapacity) : ClockCacheBase<T
 
         if (_cacheMap.Remove(key, out LruCacheItem? ov))
         {
+            _count--;
             KeyToOffset[ov.Offset] = default;
             ClearAccessed(ov.Offset);
             FreeOffsets.Enqueue(ov.Offset);
@@ -157,7 +163,7 @@ public sealed class ClockCache<TKey, TValue>(int maxCapacity) : ClockCacheBase<T
         return _cacheMap.ContainsKey(key);
     }
 
-    public int Count => _cacheMap.Count;
+    public int Count => _count;
 
     private class LruCacheItem(int offset, TValue v)
     {
--- a/src/Nethermind/Nethermind.Core/Crypto/Keccak.cs
+++ b/src/Nethermind/Nethermind.Core/Crypto/Keccak.cs
@@ -5,6 +5,8 @@ using System;
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
+using Nethermind.Core.Caching;
+using Nethermind.Core.Extensions;
 
 namespace Nethermind.Core.Crypto
 {
@@ -12,10 +14,11 @@ namespace Nethermind.Core.Crypto
     [DebuggerDisplay("{ToString()}")]
     public static class ValueKeccak
     {
+        private static readonly HashCache _cache = new();
         /// <returns>
         ///     <string>0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470</string>
         /// </returns>
-        public static readonly ValueHash256 OfAnEmptyString = InternalCompute(Array.Empty<byte>());
+        public static readonly ValueHash256 OfAnEmptyString = new("0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470");
 
         /// <returns>
         ///     <string>0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347</string>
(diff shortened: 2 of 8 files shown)


src/Nethermind/Nethermind.Core/Extensions/Bytes.cs
```diff
@@ -1124,28 +1126,38 @@ private static byte[] FromHexString(ReadOnlySpan<char> hexString)
...
+                {
+                    crc = BitOperations.Crc32C(crc, b);
+                }
+            }
+            else
+            {
+                foreach (byte b in span)
+                {
+                    crc = BitOperations.Crc32C(crc, b);
+                }
```
LukaszRozmej, line 1157: this two code paths could be unified?

src/Nethermind/Nethermind.Core/Extensions/Bytes.cs
```diff
@@ -23,23 +23,25 @@ namespace Nethermind.Core.Extensions
...
+
+        public static readonly BytesEqualityComparer EqualityComparer = new BytesEqualityComparer();
         public static readonly IEqualityComparer<byte[]?> NullableEqualityComparer = new NullableBytesEqualityComparer();
         public static readonly ISpanEqualityComparer<byte> SpanEqualityComparer = new SpanBytesEqualityComparer();
         public static readonly BytesComparer Comparer = new();
         public static readonly ReadOnlyMemory<byte> ZeroByte = new byte[] { 0 };
         public static readonly ReadOnlyMemory<byte> OneByte = new byte[] { 1 };
 
-        private class BytesEqualityComparer : EqualityComparer<byte[]>
+        public class BytesEqualityComparer : EqualityComparer<byte[]>
```
LukaszRozmej, line 35: Does it need to be public?

src/Nethermind/Nethermind.Core/Caching/ClockCache.cs
```diff
@@ -86,14 +88,16 @@ private bool SetSlow(TKey key, TValue val)
 
         _cacheMap[key] = new LruCacheItem(offset, val);
         KeyToOffset[offset] = key;
+        _count++;
```
LukaszRozmej, line 91: shouldn't all `count` interactions be interlocked?

src/Nethermind/Nethermind.Core/Crypto/Keccak.cs
```diff
@@ -56,16 +70,72 @@ public static ValueHash256 Compute(ReadOnlySpan<byte> input)
...
+            if (input.Length <= 32)
+            {
+                _cache.Set(input, hash);
+            }
+            return hash;
+        }
+
+        private sealed class HashCache
+        {
+            private const int CacheCount = 16;
```
LukaszRozmej, line 112: Why not 256 and smaller caches?

src/Nethermind/Nethermind.Core/Crypto/Keccak.cs
```diff
@@ -56,16 +70,72 @@ public static ValueHash256 Compute(ReadOnlySpan<byte> input)
                 return OfAnEmptyString;
             }
 
-            Unsafe.SkipInit(out ValueHash256 keccak);
-            KeccakHash.ComputeHashBytesToSpan(input, MemoryMarshal.AsBytes(MemoryMarshal.CreateSpan(ref keccak, 1)));
-            return keccak;
+            Unsafe.SkipInit(out ValueHash256 hash);
+            Unsafe.SkipInit(out BytesWrapper cacheKey);
+            if (input.Length <= 32)
```
LukaszRozmej, line 75: should we consider length > 32? Maybe put 32 to a const so it can be easily adjusted.

src/Nethermind/Nethermind.Core/Crypto/Keccak.cs
```diff
@@ -48,6 +51,17 @@ public static ValueHash256 Compute(string input)
             return InternalCompute(System.Text.Encoding.UTF8.GetBytes(input));
         }
 
+        private readonly struct BytesWrapper(byte[] bytes) : IEquatable<BytesWrapper>
+        {
+            internal readonly byte[] _bytes = bytes;
+            public static implicit operator BytesWrapper(byte[] bytes) => new(bytes);
+            public static implicit operator BytesWrapper(ReadOnlySpan<byte> bytes) => new(bytes.ToArray());
```
LukaszRozmej, line 58: Can we search by Span without allocating first, like in SpanDictionary?

src/Nethermind/Nethermind.Core/Caching/ClockCache.cs
benaadams: No they are all under lock; but using the field means the ConcurrentDictionary doesn't lock the reads to get a consistent count

src/Nethermind/Nethermind.Core/Crypto/Keccak.cs
benaadams: Done

src/Nethermind/Nethermind.Core/Crypto/Keccak.cs
benaadams: Sources of > 32 more likely to be single use? e.g. transactions, receipts, changed nodes rlps in trie
Also more expensive to store larger items as keys

Added const

src/Nethermind/Nethermind.Core/Crypto/Keccak.cs
benaadams: Not in the regular ConcurrentDictionary; though it should live an die pretty fast in Gen0.

Also another reason to limiting to 32 bytes

src/Nethermind/Nethermind.Core/Extensions/Bytes.cs
benaadams: To show the type the the Jit and go via concerete type rather than interface

src/Nethermind/Nethermind.Core/Extensions/Bytes.cs
benaadams: Done

src/Nethermind/Nethermind.Core/Crypto/Keccak.cs
LukaszRozmej: Well it is good I made SpanConcurrentDictionary ]:->

src/Nethermind/Nethermind.Core/Crypto/Keccak.cs
```diff
@@ -48,6 +52,17 @@ public static ValueHash256 Compute(string input)
             return InternalCompute(System.Text.Encoding.UTF8.GetBytes(input));
         }
 
+        private readonly struct BytesWrapper(byte[] bytes) : IEquatable<BytesWrapper>
+        {
+            internal readonly byte[] _bytes = bytes;
```
LukaszRozmej, line 57: Would it be better if it was in-line array of 32 (and maybe additional byte for length)?

benaadams: Is just a pointer copy/wrapper; if it was an inline array it would be an additional copy?

If it was 32 + 1 (33 bytes) it would also be multiple instructions to copy vs 8 bytes and also allocating for the `byte[]` path

LukaszRozmej (approved):
Would be great to use SpanDictionary in the cache (maybe have a derived type? And avoid allocating when searching in cache.

kamilchodola:
Will throw that for comaprision with master on same spec

Scooletz (commented):
Minor remarks while working on a different approach.

src/Nethermind/Nethermind.Core/Crypto/Keccak.cs
```diff
@@ -56,16 +71,72 @@ public static ValueHash256 Compute(ReadOnlySpan<byte> input)
...
-            return keccak;
+            Unsafe.SkipInit(out ValueHash256 hash);
+            Unsafe.SkipInit(out BytesWrapper cacheKey);
+            if (input.Length <= MaxCacheLength)
+            {
+                cacheKey = input;
+                if (_cache.TryGet(cacheKey, out hash)) return hash;
+            }
+
+            KeccakHash.ComputeHashBytesToSpan(input, MemoryMarshal.AsBytes(MemoryMarshal.CreateSpan(ref hash, 1)));
```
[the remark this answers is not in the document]
benaadams: WhyShouldntIKeepItItsMineGIF


Included that too now

benaadams:
Closing in favour of https://github.com/NethermindEth/nethermind/pull/7333

Issue #7313, opened 2024-08-06
Refactoring of Optimism Plugin


emlautarom1:
**Is your feature request related to a problem? Please describe.**

Currently, Nethermind supports Optimism through its plugin mechanism, though certain Optimism specific behaviors and constants are defined in Nethermind itself, resulting in severe coupling.

**Describe the solution you'd like**

Ideally, supporting for Optimism should not rely on having Optimism-specific classes/constants/behaviors. Note that this in turn requires Nethermind to adjust it's plugin mechanism to support injecting these components.

**Describe alternatives you've considered**

Do nothing and continue working with the current implementation. Continuing with the existing implementation implies no risk of introducing new bugs but will increase the long term maintenance cost when (not if) we introduce support for new chains (ex. Taiko).

**Additional context**

- Op-geth details: https://op-geth.optimism.io/
- Optimism Plugin preparation PR: https://github.com/NethermindEth/nethermind/issues/6221
- Optimism Plugin PR: https://github.com/NethermindEth/nethermind/pull/6146
- Deposit TX type bug: https://github.com/NethermindEth/nethermind/pull/6349/files

emlautarom1:
A first complication is the existence of Optimism-specific fields in the base `Transaction` type:

https://github.com/NethermindEth/nethermind/blob/6beafee976ebbcb685d1016ff331b3c0c0e7f2c9/src/Nethermind/Nethermind.Core/Transaction.cs#L31-L37

I think that we can just push these fields to a subclass, but I haven't had the time to look into this yet. The main challenge of the current design is in `TxDecoder`

https://github.com/NethermindEth/nethermind/blob/6beafee976ebbcb685d1016ff331b3c0c0e7f2c9/src/Nethermind/Nethermind.Serialization.Rlp/TxDecoder.cs#L36

`TxDecoder`, at a high level, decodes information about a `Transaction` from RLP, but in order to do so it needs to know about the `TxType`:

https://github.com/NethermindEth/nethermind/blob/6beafee976ebbcb685d1016ff331b3c0c0e7f2c9/src/Nethermind/Nethermind.Core/TxType.cs#L6

Depending on the `TxType` different execution paths are taken. For example:

https://github.com/NethermindEth/nethermind/blob/6beafee976ebbcb685d1016ff331b3c0c0e7f2c9/src/Nethermind/Nethermind.Serialization.Rlp/TxDecoder.cs#L76-L94

[...]

PR #7334, opened 2024-08-15
Refactor `TxDecoder`

emlautarom1:
Partially solves #7313

## Changes

- Refactor `TxDecoder` to dispatch (de/en)coding to appropriate sub (de/en)coders.

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

Tested against the existing test suite with a single breaking change

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks

The existing API for the `TxDecoder` remains untouched and only one (nonsensical) test was removed. The new design picks from a list of decoders the appropriate one based on the transaction type, throwing when the type is not recognized.

This is a first step in a list of PRs, and its intentionally kept "small" to avoid having a single large refactoring PR.


src/Nethermind/Nethermind.Blockchain.Test/Validators/TxValidatorTests.cs
```diff
@@ -171,7 +171,6 @@ public bool Before_eip_2930_has_to_be_legacy_tx(TxType txType, bool eip2930)
     [TestCase(TxType.AccessList, true, false, ExpectedResult = true)]
     [TestCase(TxType.EIP1559, true, false, ExpectedResult = false)]
     [TestCase(TxType.EIP1559, true, true, ExpectedResult = true)]
-    [TestCase((TxType)100, true, false, ExpectedResult = false)]
```
emlautarom1, line 174: The only test case removed since it's nonsensical: there is no know `TxType` with code `100`

src/Nethermind/Nethermind.Serialization.Rlp/MyTxDecoder/AccessListTxDecoder.cs
```diff
@@ -0,0 +1,243 @@
...
+        }
+
+        if ((rlpBehaviors & RlpBehaviors.AllowExtraBytes) == 0)
+        {
+            rlpStream.Check(lastCheck);
+        }
+
+        if ((rlpBehaviors & RlpBehaviors.ExcludeHashes) == 0)
+        {
+            if (_lazyHash && transactionSequence.Length <= ITxDecoder.MaxDelayedHashTxnSize)
```
emlautarom1, line 40: I really would like to refactor this to some self contained unit. This would allow us to remove the `lazyHash`, `ITxDecoder.MaxDelayedHashTxnSize` and `transactionSequence`.

The problem lies in `BlobTxDecoder`, where the logic is slightly different.

src/Nethermind/Nethermind.Serialization.Rlp/MyTxDecoder/BlobTxDecoder.cs
```diff
@@ -0,0 +1,357 @@
...
+                transaction.SetPreHashNoLock(transactionSequence);
+            }
+            else
+            {
+                // Just calculate the Hash as txn too large
+                transaction.Hash = Keccak.Compute(transactionSequence);
+            }
+        }
+
+        return transaction;
```
emlautarom1, line 75: I would appreciate a small note on this `ShardBlobNetworkWrapper` component since it makes this logic far more complicated than in the other cases.

src/Nethermind/Nethermind.Serialization.Rlp/MyTxDecoder/ITxDecoder.cs
```diff
@@ -0,0 +1,20 @@
...
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System;
+using Nethermind.Core;
+
+namespace Nethermind.Serialization.Rlp.MyTxDecoder;
+
+interface ITxDecoder
+{
+    public const int MaxDelayedHashTxnSize = 32768;
```
emlautarom1, line 11: What is the meaning/reasoning behind this constant?

src/Nethermind/Nethermind.Serialization.Rlp/MyTxDecoder/OptimismTxDecoder.cs
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    private static void DecodeSignature(Transaction transaction, ref Rlp.ValueDecoderContext decoderContext, RlpBehaviors rlpBehaviors)
+    {
+        ulong v = decoderContext.DecodeULong();
+        ReadOnlySpan<byte> rBytes = decoderContext.DecodeByteArraySpan();
+        ReadOnlySpan<byte> sBytes = decoderContext.DecodeByteArraySpan();
+        if (v == 0 && rBytes.IsEmpty && sBytes.IsEmpty) return;
+        transaction.Signature = SignatureBuilder.FromBytes(v + Signature.VOffset, rBytes, sBytes, rlpBehaviors) ?? transaction.Signature;
+    }
```
emlautarom1, line 166: From reading the code, it seems like Optimism transactions have no signature (we never encode it at least), yet `.DecodeX` methods move forward the Rlp stream/context forcing us to keep the code around.

src/Nethermind/Nethermind.Blockchain.Test/Validators/TxValidatorTests.cs
LukaszRozmej: Well the point of the this test case was that it failed?

src/Nethermind/Nethermind.Serialization.Rlp/TxDecoder.cs
```diff
@@ -2,887 +2,160 @@
...
-            _lazyHash = lazyHash;
-        }
+        _decoders = new() {
+            { TxType.Legacy, new LegacyTxDecoder(lazyHash, NewTx) },
+            { TxType.AccessList, new AccessListTxDecoder(lazyHash, NewTx) },
+            { TxType.EIP1559, new EIP1559TxDecoder(lazyHash, NewTx) },
+            { TxType.Blob, new BlobTxDecoder(lazyHash, NewTx) },
+            { TxType.DepositTx, new OptimismTxDecoder(lazyHash, NewTx) }
+        };
+    }
```
LukaszRozmej, line 38: Should this be static?

LukaszRozmej (commented):
I think we went from one extreme to the other. From sharing everything in one class to extreme code duplication in multiple classes. Can we share the code that is the same for different decoders either by base class or some static methods?

src/Nethermind/Nethermind.Serialization.Rlp/MyTxDecoder/AccessListTxDecoder.cs
LukaszRozmej: @benaadams

src/Nethermind/Nethermind.Blockchain.Test/Validators/TxValidatorTests.cs
emlautarom1: That case would result in the `TxValidator` rejecting the transaction as malformed. Now, we fail earlier in the decoding stage (required when computing the transaction hash).

The difference now is that we throw an exception instead of returning `false`.

src/Nethermind/Nethermind.Serialization.Rlp/TxDecoder.cs
emlautarom1: You mean the decoder list? If so, have this `lazyHash` property that we need to thread to all decoders. If we make them static then we would need to have a copy with a different `lazyHash` value, and it would be hard for the methods to decide which one to pick. I think a possible approach would be the following:

- Keep `_decoders` a property
- Make two static instances of `TxDecoder<T>` where one has `lazyHash = true`, and the other `lazyHash = false`.
- Make the constructor for `TxDecoder<T>` private/protected, thus making it esentially a singleton.

src/Nethermind/Nethermind.Serialization.Rlp/TxDecoder.cs
```diff
@@ -2,887 +2,164 @@
...
-        {
-            return TxObjectPool.Get();
-        }
-    }
-    public class SystemTxDecoder : TxDecoder<SystemTransaction> { }
-    public class GeneratedTxDecoder : TxDecoder<GeneratedTransaction> { }
+    // Rlp will try to find a public parameterless constructor during static initialization.
+    // The lambda cannot be removed due to static block initialization order.
+    [Obsolete("Use `TxDecoder.Instance` instead")]
+    public TxDecoder() : base(() => TxObjectPool.Get()) { }
```
emlautarom1, line 24: C# does not provide a good attribute to mark a method as "do not use" and the common solution for this case is to use `Obsolete` (see: https://stackoverflow.com/questions/27758/notify-developer-of-a-do-not-use-method)

src/Nethermind/Nethermind.Serialization.Rlp/ReceiptRecoveryBlock.cs
```diff
@@ -51,9 +51,12 @@ public Transaction GetNextTransaction()
...
 
-        Rlp.ValueDecoderContext decoderContext = new(_transactionData, true);
-        decoderContext.Position = _currentTransactionPosition;
-        TxDecoder.InstanceWithoutLazyHash.Decode(ref decoderContext, ref _txBuffer, RlpBehaviors.AllowUnsigned);
+        Rlp.ValueDecoderContext decoderContext = new(_transactionData, true)
+        {
+            Position = _currentTransactionPosition
+        };
+        TxDecoder.Instance.Decode(ref decoderContext, ref _txBuffer, RlpBehaviors.AllowUnsigned);
+        Hash256 _ = _txBuffer.Hash; // Force Hash evaluation
```
emlautarom1, line 59: This was the only instance where the hash was evaluated eagerly. Instead of distorting the API to conform to this single use case, we force the hash evaluation immediately.

emlautarom1:
> I think we went from one extreme to the other. From sharing everything in one class to extreme code duplication in multiple classes. Can we share the code that is the same for different decoders either by base class or some static methods?

I tried using some form of template pattern where different transaction types can override the "default" behavior. In my opinion, this resulted in heavily coupled code with a lot of implicit assumptions since the "base class" methods resulted in a mix of abstract and virtual that always throw (ex. `LegacyTxDecoder` needs a special `Encode` and `EncodeSignature`, forcing the template class to either have a virtual method that always throws, or an abstract method with additional parameters only for this case).

I think there are two main sources of duplication:
- The existence of `RlpStream` and `ref Rlp.ValueDecoderContext`
- The requirement to compute sequence lengths before encoding.

For the second case some higher order functions could help at the expense of performance. I will create a separate PR showing this approach and we can discuss if it is worth or not.

src/Nethermind/Nethermind.Serialization.Rlp/TxDecoder.cs
emlautarom1: Implemented the proposed approach and remove the `lazyHash` parameter.

src/Nethermind/Nethermind.Serialization.Rlp/MyTxDecoder/ITxDecoder.cs
LukaszRozmej: if the tx rlp is above it we calculate hash eagerly (probably in order not to keep that memory longer on rented array), if below we try to do it in the background.

LukaszRozmej:
`RlpStream` and `ValueDecoderContext` code might be able to be unified after: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/ref-struct-interfaces

src/Nethermind/Nethermind.Blockchain.Test/Validators/TxValidatorTests.cs
tkstanczak: do not assume the internal code when writing tests, tests are idnependent

tkstanczak: throwing exception instead of returning false is a potentially big change in many places in the code and may affect entire p2p and evm processing behaviour

significant amount of testing would have to be added

emlautarom1: What do you think @LukaszRozmej ? When we discussed this refactor we mentioned that on unknown transaction types we would throw. Should we move forward with that approach or do you want to go back to the drawing board?

src/Nethermind/Nethermind.Serialization.Rlp/MyTxDecoder/ITxDecoder.cs
emlautarom1: My question was more targeted to the reasoning behind that particular constant. Was it decided based on some benchmark/test, or does it have a particular meaning?

src/Nethermind/Nethermind.Blockchain.Test/Validators/TxValidatorTests.cs
emlautarom1: > Would be good to write a test for `NewPayload` with unknown tx type

Could you point me to where similar tests (if any) exists? I'm a bit lost on what we're trying to test.

LukaszRozmej (approved):

src/Nethermind/Nethermind.Blockchain.Test/Validators/TxValidatorTests.cs
LukaszRozmej: We need to guarantee that NewPayload returns correct error, we can add it in `EngineModuleTests` class.

LukaszRozmej: Added test and fixed exceptions, same error now on master and this branch

LukaszRozmej (approved):

emlautarom1:
During this refactor we explored an alternative design to the current RLP encoding/decoding API that is more declarative and results in less duplication:

https://github.com/NethermindEth/nethermind/blob/b9becf741a61fbeb9ce24bb72e9e9603dadbd9cb/src/Nethermind/Nethermind.Serialization.Rlp/MyTxDecoder/AccessListTxDecoder.cs

Essentially, we describe how to traverse a data structure (in this case a `Transaction`) to a `Reader/Writer` which is responsible for:
- During decoding, decodes sequences and the values it contains.
- During encoding, computes the length of the values and writes the appropriate sequences (including length prefixes)

Unfortunately I could not figure out a way to describe the traversal once and use it for both decoding and encoding, yet we are able to remove quite a bit of duplicated code. We might want to revisit this approach in the future.

PR #7335, opened 2024-08-16
feature: Add support for block validation (flashbots_validateBuilderSubmissionV3)

### LINKED ISSUE #7462: Implement flashbots endpoint (flashbots_validateBuilderSubmissionV3)
ref:  https://github.com/flashbots/builder/blob/df9c765067d57ab4b2d0ad39dbb156cbe4965778/eth/block-validation/api.go#L225-L246

rjnrohit:
Closes #7462 

This PR aims to implement flashbots endpoint (`flashbots_validateBuilderSubmissionV3`) ref: https://github.com/flashbots/builder/blob/df9c765067d57ab4b2d0ad39dbb156cbe4965778/eth/block-validation/api.go#L225-L246

1. It validates the sent payload's `block`, `blobs` and verifies if the `ParentHash`, `BlockHash`, `GasLimit`, and `GasUsed` in the params.Message match the sent payload and produce an error otherwise.
2. It tries to imitate block processing step and also process withdrawals
3. It tries to validate proposer payment using Txns & receipts generated in the block processing step

## Changes

It introduces a new `JsonRpc` plugin called `Nethermind.Flashbots` which implements a new endpoint called `flashbots_validateBuilderSubmissionV3`

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

I'll write the tests once I'm done with manual testing


src/Nethermind/Nethermind.BlockValidation/BlockValidation.cs
```diff
@@ -0,0 +1,53 @@
...
+        ValidateSubmissionHandler validateSubmissionHandler = new ValidateSubmissionHandler(
+            _api.BlockValidator ?? throw new ArgumentNullException(nameof(_api.BlockValidator)),
+            readOnlyTxProcessingEnv,
+            _api.GasLimitCalculator ?? throw new ArgumentNullException(nameof(_api.GasLimitCalculator)),
+            _blockValidationConfig
+        );
+        IFlashbotsRpcModule flashbotsRpcModule = new FlashbotsRpcModule(validateSubmissionHandler);
+
+        ArgumentNullException.ThrowIfNull(_api.RpcModuleProvider);
+        _api.RpcModuleProvider.RegisterSingle(flashbotsRpcModule);
```
LukaszRozmej, line 40: As the whole thing is based on state, we should register a BoundedModule with some concurrency limit and a factory for that. Check how `eth` module is used

src/Nethermind/Nethermind.BlockValidation/IFlashbotsRpcModule.cs
```diff
@@ -0,0 +1,19 @@
...
+using Nethermind.JsonRpc.Modules;
+
+namespace Nethermind.BlockValidation;
+
+[RpcModule(ModuleType.Flashbots)]
+public interface IFlashbotsRpcModule : IRpcModule
+{
+    [JsonRpcMethod(
+        Description = " validate the builder submissions as received by a relay",
+        IsSharable = true,
```
LukaszRozmej, line 16: definitely not sharable, it needs to be instance based as each instance will have it's own vision of what the state is.

src/Nethermind/Nethermind.Consensus/IGasLimitCalculator.cs
```diff
@@ -7,6 +7,6 @@ namespace Nethermind.Consensus
 {
     public interface IGasLimitCalculator
     {
-        long GetGasLimit(BlockHeader parentHeader);
+        long GetGasLimit(BlockHeader parentHeader, long? desiredGasLimit = null);
```
LukaszRozmej, line 10: Something to think about - can we avoid changing this interface, I kind of don't think we should?

src/Nethermind/Nethermind.BlockValidation/Handlers/ValidateBuilderSubmissionHandler.cs
```diff
@@ -0,0 +1,331 @@
...
+        {
+            return false;
+        }
+
+        _logger.Info($"Validated block Hash: {block.Header.Hash} Number: {block.Header.Number} ParentHash: {block.Header.ParentHash}");
+
+        return true;
+    }
+
+    public Task<ResultWrapper<BlockValidationResult>> ValidateSubmission(BuilderBlockValidationRequest request)
```
LukaszRozmej, line 301: minor: lets move the entry point to the top of the class so it is easy to understand where things start. Other methods should be layout in similar manner (logically close to each other)

src/Nethermind/Nethermind.BlockValidation/Handlers/ValidateBuilderSubmissionHandler.cs
```diff
@@ -0,0 +1,331 @@
...
+    {
+        stateProvider.StateRoot = processedBlock.StateRoot!;
+        stateProvider.Commit(currentSpec);
+        stateProvider.CommitTree(currentBlock.Number);
+        blockTree.SuggestBlock(processedBlock, BlockTreeSuggestOptions.ForceSetAsMain);
+        blockTree.UpdateHeadBlock(processedBlock.Hash!);
+    }
+
+
+    private bool ValidatePayload(Block block, Address feeRecipient, UInt256 expectedProfit, long registerGasLimit, bool useBalanceDiffProfit, bool excludeWithdrawals, out string? error)
```
LukaszRozmej, line 118: all those validation code can be nicely simplified with `ValidateResult` introduced in #7402

src/Nethermind/Nethermind.BlockValidation/Handlers/ValidateBuilderSubmissionHandler.cs
```diff
@@ -0,0 +1,331 @@
...
+        {
+            error = $"Invalid block header hash {block.Header.Hash}";
+            return false;
+        }
+
+        if (!_blockTree.IsBetterThanHead(block.Header))
+        {
+            error = $"Block {block.Header.Hash} is not better than head";
+            return false;
+        }
```
LukaszRozmej, line 130: We can validate only on future blocks?

src/Nethermind/Nethermind.BlockValidation/Handlers/ValidateBuilderSubmissionHandler.cs
```diff
@@ -0,0 +1,331 @@
...
+        IWorldState currentState = processingScope.WorldState;
+
+        UInt256 feeRecipientBalanceBefore = currentState.GetBalance(feeRecipient);
+
+        BlockProcessor blockProcessor = CreateBlockProcessor(currentState);
+
+        List<Block> suggestedBlocks = [block];
+        BlockReceiptsTracer blockReceiptsTracer = new();
+
+        ProcessingOptions processingOptions = new ProcessingOptions();
```
LukaszRozmej, line 158: This is not correct. We want: `ReadOnlyChain` and maybe `ForceProcessing`, `IgnoreParentNotOnMainChain`, `DoNotVerifyNonce` (if all that would be all under `Trace`)

src/Nethermind/Nethermind.BlockValidation/Handlers/ValidateBuilderSubmissionHandler.cs
LukaszRozmej, line 118: This method is too long, it consists of few actual different steps:
1. Initial Validation
2. Building environment
3. Actual processing
4. Calculating stuff after

And should be split - into local methods.

src/Nethermind/Nethermind.BlockValidation/Handlers/ValidateBuilderSubmissionHandler.cs
```diff
@@ -0,0 +1,331 @@
...
+        {
+            if (_logger.IsWarn) _logger.Warn($"Invalid block. Result of {payloadStr}. Error: {error}");
+            return BlockValidationResult.Invalid(error ?? "Block validation failed");
+        }
+
+        if (block is not null && !ValidateBlobsBundle(block.Transactions, blobsBundle, out string? blobsError))
+        {
+            if (_logger.IsWarn) _logger.Warn($"Invalid blobs bundle. Result of {payloadStr}. Error: {blobsError}");
+            return BlockValidationResult.Invalid(blobsError ?? "Blobs bundle validation failed");
+        }
```
LukaszRozmej, line 326: A but weird that we do validations after processing a block, shouldn't we do it first? (It would be less costly)

LukaszRozmej (changes requested):

src/Nethermind/Nethermind.BlockValidation/Handlers/ValidateBuilderSubmissionHandler.cs
rjnrohit: https://github.com/flashbots/builder/blob/df9c765067d57ab4b2d0ad39dbb156cbe4965778/eth/block-validation/api.go#L240

I followed the same order as they did, but yeah I agree doing this way would be costly

src/Nethermind/Nethermind.BlockValidation/Handlers/ValidateBuilderSubmissionHandler.cs
rjnrohit: Yeah it seems from this snippet: https://github.com/flashbots/builder/blob/df9c765067d57ab4b2d0ad39dbb156cbe4965778/core/blockchain.go#L2464

definition at: https://github.com/flashbots/builder/blob/df9c765067d57ab4b2d0ad39dbb156cbe4965778/core/forkchoice.go#L77

src/Nethermind/Nethermind.BlockValidation/Handlers/ValidateBuilderSubmissionHandler.cs
LukaszRozmej: W should consider changing it later, maybe @Ruteri knows why this is the order of checks?

src/Nethermind/Nethermind.Flashbots/Handlers/ValidateBuilderSubmissionHandler.cs
```diff
@@ -0,0 +1,432 @@
...
+    private readonly ReadOnlyTxProcessingEnvFactory _readOnlyTxProcessingEnvFactory;
+
+    private readonly IBlockTree _blockTree;
+    private readonly IHeaderValidator _headerValidator;
+    private readonly IBlockValidator _blockValidator;
+    private readonly ILogger _logger;
+
+    private readonly IFlashbotsConfig _flashbotsConfig;
+
+    private readonly IReceiptStorage _receiptStorage = new InMemoryReceiptStorage();
```
LukaszRozmej, line 49: Do we need receipt storage at all? Maybe `NullReceiptStorage`?

src/Nethermind/Nethermind.Flashbots/Handlers/ValidateBuilderSubmissionHandler.cs
```diff
@@ -0,0 +1,432 @@
...
+        }
+
+        payload.ParentBeaconBlockRoot = new Hash256(request.ParentBeaconBlockRoot);
+
+
+        BlobsBundleV1 blobsBundle = request.BlockRequest.BlobsBundle;
+
+        string payloadStr = $"BuilderBlock: {payload}";
+
+        _logger.Info($"blobs bundle blobs {blobsBundle.Blobs.Length} commits {blobsBundle.Commitments.Length} proofs {blobsBundle.Proofs.Length}");
```
LukaszRozmej, line 90: under `if (_logger.IsInfo)`

src/Nethermind/Nethermind.Flashbots/Handlers/ValidateBuilderSubmissionHandler.cs
```diff
@@ -0,0 +1,432 @@
...
+            }
+            _logger.Warn($"Proposer payment is not enough, trying last tx payment validation, expected: {expectedProfit}, actual: {feeRecipientBalanceDelta}");
+        }
+
+        return false;
+    }
+
+    private bool ValidateProcessedBlock(Block processedBlock, Address feeRecipient, UInt256 expectedProfit, TxReceipt[] receipts, out string? error)
+    {
+        // TxReceipt[] receipts = processedBlock.Hash != null ? _receiptStorage.Get(processedBlock.Hash) : [];
```
LukaszRozmej, line 348: Not needed?

src/Nethermind/Nethermind.Flashbots/Handlers/ValidateBuilderSubmissionHandler.cs
```diff
@@ -0,0 +1,432 @@
...
+        }
+
+        if (!_blockValidator.ValidateSuggestedBlock(block, out error))
+        {
+            return false;
+        }
+
+        if (ValidateProposerPayment(expectedProfit, useBalanceDiffProfit, feeRecipientBalanceAfter, amtBeforeOrWithdrawn)) return true;
+
+        if (!ValidateProcessedBlock(block, feeRecipient, expectedProfit, blockReceiptsTracer.TxReceipts.ToArray(), out error))
```
LukaszRozmej, line 260: We could avoid `ToArray`

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
```diff
@@ -404,9 +404,10 @@ public ResultWrapper<BlockForRpc> eth_getBlockByNumber(BlockParameter blockParam
             return ResultWrapper<string?>.Success(null);
         }
 
-        IByteBuffer buffer = PooledByteBufferAllocator.Default.Buffer(TxDecoder.Instance.GetLength(transaction, RlpBehaviors.None));
+        IByteBuffer buffer = PooledByteBufferAllocator.Default.Buffer(TxDecoder.Instance.GetLength(transaction, RlpBehaviors.SkipTypedWrapping));
         using NettyRlpStream stream = new(buffer);
-        TxDecoder.Instance.Encode(stream, transaction);
+        TxDecoder.Instance.Encode(stream, transaction, rlpBehaviors: RlpBehaviors.SkipTypedWrapping);
+
```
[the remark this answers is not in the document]
marcindsobczak: I think it's not needed, but it is not a problem as well, it's the same set of data, just in different forms. We are handling well both. Probably better to not change it if not needed, unless there is a strong argument for it

src/Nethermind/Nethermind.Flashbots/Flashbots.cs
```diff
@@ -0,0 +1,76 @@
...
+            _api.SpecProvider,
+            _api.LogManager
+        );
+
+        ReadOnlyTxProcessingEnv readOnlyTxProcessingEnv = new ReadOnlyTxProcessingEnv(
+            _api.WorldStateManager ?? throw new ArgumentNullException(nameof(_api.WorldStateManager)),
+            _api.BlockTree ?? throw new ArgumentNullException(nameof(_api.BlockTree)),
+            _api.SpecProvider,
+            _api.LogManager
+        );
```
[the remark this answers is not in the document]
rjnrohit: It's not needed anymore; I am removing it in one of the recent commits

src/Nethermind/Nethermind.Flashbots/Handlers/ValidateBuilderSubmissionHandler.cs
rjnrohit: No we don't need it anymore, I have removed it as well

src/Nethermind/Nethermind.Flashbots/Handlers/ValidateBuilderSubmissionHandler.cs
```diff
@@ -0,0 +1,427 @@
...
+        }
+
+        if (!_blockValidator.ValidateSuggestedBlock(block, out error))
+        {
+            return false;
+        }
+
+        if (ValidateProposerPayment(expectedProfit, useBalanceDiffProfit, feeRecipientBalanceAfter, amtBeforeOrWithdrawn)) return true;
+
+        if (!ValidateProcessedBlock(block, feeRecipient, expectedProfit, [.. blockReceiptsTracer.TxReceipts], out error))
```
LukaszRozmej, line 257: This still allocates array, I thought of changing parameter type to `IReadOnlyList` for example.

LukaszRozmej (approved):
Fix build

src/Nethermind/Nethermind.Flashbots/IFlashbotsConfig.cs
```diff
@@ -0,0 +1,32 @@
...
+using Nethermind.Config;
+
+namespace Nethermind.Flashbots;
+
+public interface IFlashbotsConfig : IConfig
+{
+    [ConfigItem(Description = "Whether to enable the Flashbots endpoints.", DefaultValue = "false")]
+    bool Enabled { get; set; }
+
+    [ConfigItem(Description = "If set to true, proposer payment is calculated as a balance difference of the fee recipient", DefaultValue = "false")]
```
rubo, line 13: We should keep the wording the same as this goes to the docs:

```suggestion
    [ConfigItem(Description = "Whether to calculate the proposer payment as a balance difference of the fee recipient.", DefaultValue = "false")]
```

src/Nethermind/Nethermind.Flashbots/IFlashbotsConfig.cs
```diff
@@ -0,0 +1,32 @@
...
+    public bool UseBalanceDiffProfit { get; set; }
+
+    [ConfigItem(Description = "If set to true, withdrawals to the fee recipient are excluded from the balance delta", DefaultValue = "false")]
```
rubo, line 16:
```suggestion
    [ConfigItem(Description = "Whether to exclude the withdrawals to the fee recipient from the balance difference.", DefaultValue = "false")]
```

src/Nethermind/Nethermind.Flashbots/IFlashbotsConfig.cs
```diff
@@ -0,0 +1,32 @@
...
+    public bool ExcludeWithdrawals { get; set; }
+
+    [ConfigItem(Description = "If set to true, the pre-warmer will be enabled", DefaultValue = "true")]
```
rubo, line 19:
```suggestion
    [ConfigItem(Description = "Whether to enable the pre-warmer.", DefaultValue = "true")]
```

src/Nethermind/Nethermind.Flashbots/IFlashbotsConfig.cs
```diff
@@ -0,0 +1,32 @@
...
+    public bool EnablePreWarmer { get; set; }
+
+    [ConfigItem(Description = "If set to true, the validation will be enabled", DefaultValue = "false")]
```
rubo, line 22:
```suggestion
    [ConfigItem(Description = "Whether to enable validation.", DefaultValue = "false")]
```

src/Nethermind/Nethermind.Flashbots/IFlashbotsConfig.cs
```diff
@@ -0,0 +1,32 @@
...
+    public bool EnableValidation { get; set; }
+
+    [ConfigItem(
+    Description = """
+        The number of concurrent instances for non-sharable calls:
+        - `flashbots_validateBuilderSubmissionV3`
```
rubo, line 28: It's unclear why this is formatted as a list. It renders as follows:

> The number of concurrent instances for non-sharable calls:
> - `flashbots_validateBuilderSubmissionV3`
> This limits the load on the CPU and I/O to reasonable levels. If the limit is exceeded, HTTP 503 is returned along with the JSON-RPC error. Defaults to the number of logical processors.

Given this option requires numbers, we expect numeric values as list items. The RPC endpoint name as a list item is confusing.

src/bench_precompiles
rubo, line 1: Do we need this change?

LukaszRozmej (approved):

mralj:
Hey team, pushed small bugfix commit, which fixes error:

```
System.NotSupportedException: Deserialization of types without a parameterless constructor, a singular parameterized constructor, or a parameterized constructor annotated with 'JsonConstructorAttribute' is not supported. Type 'RExecutionPa
yloadV3'.
```

I was getting when testing this using `rbuilder`

tanishqjasoria (approved):

LukaszRozmej (approved):

LukaszRozmej (approved):

src/Nethermind/Nethermind.Flashbots.Test/FlashbotsModuleTests.cs
```diff
@@ -59,7 +59,7 @@ public virtual async Task TestValidateBuilderSubmissionV3()
         ResultWrapper<FlashbotsResult> result = await rpc.flashbots_validateBuilderSubmissionV3(BlockRequest);
         result.Should().NotBeNull();
 
-        Assert.That(result.Result, Is.EqualTo(Result.Success));
+        // Assert.That(result.Result, Is.EqualTo(Result.Success));
```
LukaszRozmej, line 62: why those asserts are commented out?

rjnrohit: I'm working on this right now

LukaszRozmej (approved):

PR #7336, opened 2024-08-16
Fast Keccak cache (wait-free and lockless)

Scooletz:
This PR introduces Keccaks caching (see #7333 and related for other attempts) that is minimal in regards to locks, copying and memory overhead. This PR introduces a `KeccakCache` that 

1. is one-way set associative cache
2. allows to cache `128k` values
3. occupies a fixed `12MB` of unmanaged memory allocated once
4. uses a fast hash to bucketize
5. uses 1 `Interlocked.CompareExchange` to lock
6. uses 1 `Volatile.Write` to unlock

It does not augments the `ValueKeccak` itself, allowing the callsite owner to decide, whether or not the value should go through cache. Call sites amended:

1. `StateTree` address->Keccak calculations
7. `EVM` `Instruction.Keccak256` execution
8. `StorageTree` uint256 -> Keccak cell address calculations

There are several interesting methods used in this PR:

1. encoding a lot of state in a single `int` to that `volatile` and `Interlocked` can operate on it
2. wait-free, as if a caller fails to lock to check the value, they compute. If they fail to lock again, they don't put it in the cache
3. special cases for lengths of `20` and `32` that are compared without calling to the span
4. alignment for vectors, that should speed up the access and comparisons again
5. stack copying, as the entry is unmanaged, when a lock is taken for reading purposes, what is needed from `Entry` can be copied on the stack, so that the lock is released in a fast manner

## Changes

- introduce `KeccakCache` that allows to cache values of the Keccak.
- introduce a hardware accelerated Crc32c hash for Dictionary keys that is randomised per instance and use that for most of our dictionary key hashing as a common fast method (randomised so working out a hash collision attack on one node will not impact other nodes or work across restarts)

- [x] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Synthetic

Single threaded, hammering the same value:

| Method   | data     | Mean       | Error     | StdDev    |
|--------- |--------- |-----------:|----------:|----------:|
| Current  | Byte[20] | 210.804 ns | 2.4348 ns | 3.4132 ns |
| Memoized | Byte[20] |   8.127 ns | 0.0868 ns | 0.1299 ns |
| Current  | Byte[32] | 210.454 ns | 1.7798 ns | 2.6638 ns |
| Memoized | Byte[32] |   7.913 ns | 0.0780 ns | 0.1143 ns |

#### Profiling mainnet

Saves on approx 40% of keccaks when running mainnet (88780 to 53664 approx ~40%)

image

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks


Scooletz:
The run for comparing them https://github.com/NethermindEth/nethermind/actions/runs/10419906977

src/Nethermind/Nethermind.Core/Crypto/KeccakCache.cs
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    /// <summary>
+    /// Gets the bucket for tests.
+    /// </summary>
+    public static uint GetBucket(ReadOnlySpan<byte> input) => FastHash(input) & BucketMask;
+
+    [SkipLocalsInit]
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    private static uint FastHash(ReadOnlySpan<byte> input)
```
benaadams, line 124: Hash from end rather than start. The end of the bytes is higher entropy than the start

benaadams: nvm, you use the whole length

src/Nethermind/Nethermind.Core/Crypto/KeccakCache.cs
```diff
@@ -130,7 +130,7 @@ private static uint FastHash(ReadOnlySpan<byte> input)
 
         ref var b = ref MemoryMarshal.GetReference(input);
 
-        // Start with first
+        // Start with instance random, length and first as seed
```
Scooletz, line 133: @benaadams The length is not needed I believe as it provides not that much entropy. Also, the lenght later is encoded as a part of `HashAndLength` so can be compared directly in one comparison.

src/Nethermind/Nethermind.Core/Crypto/KeccakCache.cs
```diff
@@ -21,6 +21,7 @@ namespace Nethermind.Core.Crypto;
 /// </summary>
 public static unsafe class KeccakCache
 {
+    private static readonly uint s_instanceRandom = (uint)System.Security.Cryptography.RandomNumberGenerator.GetInt32(int.MinValue, int.MaxValue);
```
Scooletz, line 24: Ah yes, the initial seed so that nobody finds a way to attack all the nodes. 💪 @benaadams

src/Nethermind/Nethermind.Core/Crypto/KeccakCache.cs
benaadams: instance random will be same, [Benford's law](https://en.wikipedia.org/wiki/Benford%27s_law) says most of the leading bytes on big endian will skew the same (hence my other comment on high entropy at the end, rather than start; and more so if leading zeros aren't trucated), so in-order to not frequently start with same seed I'm incorporating length.

benaadams: i.e. including .Length gives more value in running the first Crc

src/Nethermind/Nethermind.Core/Crypto/KeccakCache.cs
```diff
@@ -0,0 +1,226 @@
...
+
+        ref Entry e = ref Unsafe.Add(ref Unsafe.AsRef<Entry>(Memory), index);
+
+        // Read aligned, volatile, won't be torn, check with computed
+        if (Volatile.Read(ref e.HashCode) == hashCode)
+        {
+            // There's a possibility of a hit, try lock.
+            int lockAndLength = Volatile.Read(ref e.LockAndLength);
+            // Lock by negating length if the length is the same size as input.
+            if (lockAndLength == input.Length && Interlocked.CompareExchange(ref e.LockAndLength, -lockAndLength, lockAndLength) == lockAndLength)
```
Scooletz, line 71: If you want to go this way, and make the lock less likely in case of a fail, we can squeeze all of the values together. Length is 1 byte long, hash is 2 bytes long, lock can be 1. I'm missing the reasoning in the comment why take this path. Have you found it faster? It's a bit less colliding.

src/Nethermind/Nethermind.Core/Crypto/KeccakCache.cs
```diff
@@ -0,0 +1,226 @@
...
+                // Local copy of 8+16+64 payload bytes.
+                Payload copy = e.Value;
+                // Copy Keccak256 directly to the local return variable, since we will overwrite if no match anyway.
+                keccak256 = e.Keccak256;
+
+                // Release the lock, by setting back to unnegated length.
+                Volatile.Write(ref e.LockAndLength, lockAndLength);
+
+                // Lengths are equal, the input length can be used without any additional operation.
+                if (input.Length == 32)
```
Scooletz, line 91: Is it faster this way? My benchmarks showed that the equality wasn't the big deal here. Especially as it falls almost right into the optimized path of vectorized?

Scooletz, line 91: Name constants. 32 is magic.

src/Nethermind/Nethermind.Core/Crypto/KeccakCache.cs
```diff
@@ -0,0 +1,226 @@
...
+        int length = Volatile.Read(ref e.LockAndLength);
+        // Negative value means that the entry is locked, set to int.MinValue to avoid confusion empty entry,
+        // since we are overwriting it anyway e.g. -0 would not be a reliable locked state.
+        if (length >= 0 && Interlocked.CompareExchange(ref e.LockAndLength, int.MinValue, length) == length)
+        {
+            e.HashCode = hashCode;
+            e.Keccak256 = keccak256;
+
+            // Fast copy for 2 common sizes
+            if (input.Length == 32)
```
Scooletz, line 137: name consts

src/Nethermind/Nethermind.Core/Crypto/KeccakCache.cs
```diff
@@ -0,0 +1,226 @@
...
+                    // Release the lock and compute.
+                    Volatile.Write(ref e.LockAndLength, lockAndLength);
+                    goto Compute;
+                }
+
+                // Take local copy of the payload and hash, to release the lock as soon as possible and make a key comparison without holding it.
+                // Local copy of 8+16+64 payload bytes.
+                Payload copy = e.Value;
+                // Copy Keccak256 directly to the local return variable, since we will overwrite if no match anyway.
+                keccak256 = e.Keccak256;
```
Scooletz, line 85: So this saves copying the whole, by storing the result directly in the return variable. This is good and much better than a blunt copy of 128

src/Nethermind/Nethermind.Core/Extensions/SpanExtensions.cs
```diff
@@ -166,5 +172,62 @@ public static ArrayPoolList<T> ToPooledList<T>(this in ReadOnlySpan<T> span)
             newList.AddRange(span);
             return newList;
         }
+
+        [SkipLocalsInit]
+        public static int FastHash(this ReadOnlySpan<byte> input)
```
Scooletz, line 177: Nice extraction.

src/Nethermind/Nethermind.Core/StorageCell.cs
```diff
@@ -17,17 +19,16 @@ namespace Nethermind.Core
...
         public ValueHash256 Hash => _isHash ? Unsafe.As<UInt256, ValueHash256>(ref Unsafe.AsRef(in _index)) : GetHash();
 
-        public bool IsHash => _isHash;
-
         private ValueHash256 GetHash()
         {
             Span<byte> key = stackalloc byte[32];
             Index.ToBigEndian(key);
-            return ValueKeccak.Compute(key);
+            return KeccakCache.Compute(key);
```
Scooletz, line 31: Ouch! Missed this

Scooletz (commented):
Some comments.

src/Nethermind/Nethermind.Core/Crypto/KeccakCache.cs
```diff
@@ -0,0 +1,226 @@
...
+        }
+
+    Compute:
+        keccak256 = ValueKeccak.Compute(input);
+
+        // Try lock and memoize
+        int length = Volatile.Read(ref e.LockAndLength);
+        // Negative value means that the entry is locked, set to int.MinValue to avoid confusion empty entry,
+        // since we are overwriting it anyway e.g. -0 would not be a reliable locked state.
+        if (length >= 0 && Interlocked.CompareExchange(ref e.LockAndLength, int.MinValue, length) == length)
```
Scooletz, line 131: This is different from the original, that never tested whether it's locked or not, and always try to claim it. The current version not only changes this, but also, will take a lock ONLY if a potential update between 128 and 131 didn't change the length. This could be a good thing, meaning, that if an entry is updated hard with different lengths, this might save one update over a frequently updated entry.

Scooletz:
Benchmarks updated in the PR description.

src/Nethermind/Nethermind.Core/Crypto/KeccakCache.cs
```diff
@@ -188,7 +186,7 @@ private struct Entry
         /// </summary>
         public const int Size = 128;
 
-        private const int PayloadStart = 8;
+        private const int PayloadStart = sizeof(uint);
```
Scooletz, line 189: The alignment @benaadams to leverage that with `Address`. I was thinking about it :-)

benaadams (approved):
Someone else should also review since I added a lot of changes

src/Nethermind/Nethermind.Core/Crypto/KeccakCache.cs
benaadams: Switched the orders of the compare and copy 🙂

Scooletz: This has been noticed 😅

Scooletz:
Requires mano a mano comparison with `master`. Same blocks etc.

Scooletz:
Let's have it run with https://github.com/NethermindEth/nethermind/pull/7346

src/Nethermind/Nethermind.Core/Extensions/SpanExtensions.cs
```diff
@@ -176,48 +176,49 @@ public static ArrayPoolList<T> ToPooledList<T>(this in ReadOnlySpan<T> span)
...
-            // }
-
-            uint bit = (uint)length & sizeof(byte);
-            b = ref Unsafe.Add(ref b, bit);
-            length -= (int)bit;
-
-            if ((length & sizeof(ushort)) != 0)
+            // Start with instance random, length and first ulong as seed
+            hash = BitOperations.Crc32C(hash, Unsafe.ReadUnaligned<ulong>(ref b));
+            if ((length & sizeof(ulong)) != 0)
```
Scooletz, line 192: @benaadams I believe this is not correct because if the length is bigger than 8, you never align to length & 7.

Right after reading the first long, 

```
                // Skip the bytes that don't align to ulong
                uint bits = (uint)length & 7;
                b = ref Unsafe.Add(ref b, bits);
                length -= (int)bits;
```

should follow in a branchless manner

benaadams (approved):

Scooletz:
Two more runs to compare logs:

- `master` : https://github.com/NethermindEth/nethermind/actions/runs/10940957136
- `keccak-cache`: https://github.com/NethermindEth/nethermind/actions/runs/10940953465

src/Nethermind/Nethermind.Core/Crypto/KeccakCache.cs
```diff
@@ -0,0 +1,217 @@
...
+    public static void ComputeTo(ReadOnlySpan<byte> input, out ValueHash256 keccak256)
+    {
+        // Special cases jump forward as unpredicted
+        if (input.Length == 0 || input.Length > Entry.MaxPayloadLength)
+        {
+            goto Uncommon;
+        }
+
+        int hashCode = input.FastHash();
+        uint index = (uint)hashCode & BucketMask;
```
LukaszRozmej, line 67: So it will hold only one hash with given prefix?

LukaszRozmej:
Can it give false positive? Find some Keccak for input that will be wrong (it will be keccak of different input, but have same length and same `FastHash`)?

src/Nethermind/Nethermind.Core/Crypto/KeccakCache.cs
Scooletz: It will hold one hash per bucket, where the bucket is determined by a `FastHash` over the `input` data. On collision over `FastHash` it tries to lock and exchange (using `Interlocked`) but if it fails it just moves on. This is to make it as cheap as possible for both contended and not contended buckets.

Scooletz:
No it can't give a false positive. The original input is preserved and properly compared. We use a fat struct that is copied on the stack and then compare, so no dirty reads/writes can happen. The copy is done under the lock.

PR #7338, opened 2024-08-16
Reduce contention in ClockCache

benaadams:
## Changes

- Reduce ClockCache contention
   - Calling ConcurrentDictionary.Count takes locks and causes contention on read side
   - Use local count that is only changed when have write lock anyway

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Core/Caching/ClockCacheBase.cs
+++ b/src/Nethermind/Nethermind.Core/Caching/ClockCacheBase.cs
@@ -21,6 +21,8 @@ public abstract class ClockCacheBase<TKey>
     protected Queue<int> FreeOffsets { get; } = new();
 
     protected int Clock { get; set; } = 0;
+    // Use local count to avoid lock contention with reads on ConcurrentDictionary.Count
+    protected int _count = 0;
 
     protected ClockCacheBase(int maxCapacity)
     {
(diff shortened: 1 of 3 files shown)


smartprogrammer93 (approved):

src/Nethermind/Nethermind.Core/Caching/ClockCacheBase.cs
```diff
@@ -21,6 +21,8 @@ public abstract class ClockCacheBase<TKey>
     protected Queue<int> FreeOffsets { get; } = new();
 
     protected int Clock { get; set; } = 0;
+    // Use local count to avoid lock contention with reads on ConcurrentDictionary.Count
+    protected int _count = 0;
```
Scooletz, line 25: Why introduce `_count` here while it's touched only by descendant classes?

src/Nethermind/Nethermind.Core/Caching/ClockKeyCache.cs
```diff
@@ -131,5 +137,5 @@ public bool Contains(TKey key)
         return _cacheMap.ContainsKey(key);
     }
 
-    public int Count => _cacheMap.Count;
+    public int Count => _count;
```
Scooletz, line 140:
```suggestion
    public int Count => Volatile.Read(ref _count);
```

Scooletz (commented):
Two comments + a general one, why `Count` is needed at all?

PR #7346, opened 2024-08-20
Enhanced Prometheur metrics' pusher

Scooletz:
This enhances `MonitoringService` with a new `MetricPusher`. The new pusher allows scheduling metrics push right after a block is processed. This should allow to measure multiple nodes with better comparability, by making snapshots of metrics at the same point of execution. What is expected is that once metrics are published at the end of each block processed, they should be aligned and allow better comparison between instances. 

## Changes

- _List the changes_

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Remarks


src/Nethermind/Nethermind.Monitoring/MetricPusher.cs
```diff
@@ -0,0 +1,177 @@
...
+
+                await PushOnce();
+            }
+
+            // Push the final state
+            await PushOnce();
+        });
+    }
+
+    public void Push() => _wait.Release();
```
benaadams, line 102: Is this called?

src/Nethermind/Nethermind.Monitoring/Config/MetricsConfig.cs
```diff
@@ -8,6 +8,8 @@ public class MetricsConfig : IMetricsConfig
         public string ExposeHost { get; set; } = "+";
         public int? ExposePort { get; set; } = null;
         public bool Enabled { get; set; } = false;
+
+        public bool PushAfterBlock { get; set; } = false;
```
kamilchodola, line 12: If idea here would work I'd probably prefer it to be enabled by default - this way we have much better view on metrics than now.

src/Nethermind/Nethermind.Monitoring/MetricPusher.cs
```diff
@@ -0,0 +1,177 @@
...
+
+    public MetricPusher(MetricPusherOptions options)
+    {
+        if (string.IsNullOrEmpty(options.Endpoint))
+            throw new ArgumentNullException(nameof(options.Endpoint));
+
+        if (string.IsNullOrEmpty(options.Job))
+            throw new ArgumentNullException(nameof(options.Job));
+
+        if (options.IntervalMilliseconds <= 0)
```
kamilchodola, line 38: If Interval set to 0 but PushAfterBlock=true then this exception does not make sense?

Scooletz:
Running the node: https://github.com/NethermindEth/nethermind/actions/runs/10664670485

Scooletz:
Run after fixes:

https://github.com/NethermindEth/nethermind/actions/runs/10682675711

PR #7353, opened 2024-08-22
Improve parallel signature recover

benaadams:
## Changes

- Originally started to address an observed performance stall with `Parallel.ForEach` resolved by switching to `Parallel.For` (per element dispatch vs integer range). However on comparing again the stall wasn't observed; so will call this more a refactoring 😅

- [x] Optimization
- [x] Refactoring

#### Requires testing

- [x] No


LukaszRozmej (approved):
In theory we could make sure first few transactions are recovered and then, try recover the rest in the background, then wait for that in TransactionProdcessor, would be more parallel.

PR #7360, opened 2024-08-25
Reduce collection resizing costs

benaadams:
## Changes

- Clear `ConcurrentDictionary` backing directly rather than using `.Clear` which resizes it to 31 elements meaning are many resizing costs to bring back up to size when repopulating
- Clear evm collections rather than resizing down to avoid paying the resizing back up costs
- When pruning use the array if created, rather than reenumerating the dictionary
- Clear precomplie cache (bugfix)

- [x] Optimization

#### Requires testing

- [x] No


src/Nethermind/Nethermind.Core/Collections/CollectionExtensions.cs
```diff
@@ -22,5 +26,17 @@ public static void AddRange<T>(this ICollection<T> list, params T[] items)
...
+
+        public static void NoResizeClear<TKey, TValue>(this ConcurrentDictionary<TKey, TValue> dict)
+                where TKey : notnull
+        {
+            // We iterate over the keys and remove them one by one because calling Clear() on
+            // the ConcurrentDictionary resets its capacity to 31 and then it has to constantly resize.
+            foreach (TKey key in dict.Keys)
+            {
+                dict.TryRemove(key, out _);
+            }
```
LukaszRozmej, line 38: Use `ConcurrentDictionaryLock` class to lock the dictionary

src/Nethermind/Nethermind.Core/Collections/CollectionExtensions.cs
```diff
@@ -22,5 +26,17 @@ public static void AddRange<T>(this ICollection<T> list, params T[] items)
...
         }
+
+        public static void NoResizeClear<TKey, TValue>(this ConcurrentDictionary<TKey, TValue> dict)
+                where TKey : notnull
+        {
+            // We iterate over the keys and remove them one by one because calling Clear() on
+            // the ConcurrentDictionary resets its capacity to 31 and then it has to constantly resize.
+            foreach (TKey key in dict.Keys)
+            {
+                dict.TryRemove(key, out _);
```
LukaszRozmej, line 37: Wouldn't it be better to grab the internals and clear the `_buckets` ourselves?

src/Nethermind/Nethermind.State/PreBlockCaches.cs
```diff
@@ -4,34 +4,124 @@
...
+        if (!cache.IsEmpty)
+        {
+            // Guard in case the cache is not empty after the loop
+            if (_logger.IsWarn)
+            {
+                _logger.Warn($"{paramName} didn't empty. Purging.");
+            }
+            // Recreate rather than clear to avoid resizing
+            cache = new(LockPartitions, InitialCapacity);
+        }
```
LukaszRozmej, line 122: If you do locking in `NoResizeClear` this is not needed.

LukaszRozmej (changes requested):

src/Nethermind/Nethermind.Core/Collections/CollectionExtensions.cs
benaadams: Done

src/Nethermind/Nethermind.State/PreBlockCaches.cs
benaadams: Done

src/Nethermind/Nethermind.Core/Collections/CollectionExtensions.cs
benaadams: Looks horrible to do because of the private types involved 😢

Really just need an option on .Clear to not resize back to `HashHelpers.GetPrime(31)` will raise issue in runtime

```csharp
Tables tables = _tables;
var newTables = new Tables(new VolatileNode[HashHelpers.GetPrime(DefaultCapacity)], tables._locks, new int[tables._countPerLock.Length], tables._comparer);
_tables = newTables;
_budget = Math.Max(1, newTables._buckets.Length / newTables._locks.Length);
```

benaadams: https://github.com/dotnet/runtime/issues/107016

benaadams: Actually...

benaadams: Done

src/Nethermind/Nethermind.Core/Caching/ClockCache.cs
```diff
@@ -154,7 +157,7 @@ public bool Delete(TKey key)
         using var lockRelease = _lock.Acquire();
 
         base.Clear();
-        _cacheMap.Clear();
+        _cacheMap.NoResizeClear();
```
LukaszRozmej, line 160: maybe have a `NoLockNoResizeClear` - as you don't need the `ConcurentDictionaryLock` here?

src/Nethermind/Nethermind.Core/Collections/CollectionExtensions.cs
```diff
@@ -22,5 +29,61 @@ public static void AddRange<T>(this ICollection<T> list, params T[] items)
                 list.Add(items[index]);
             }
         }
+
+        public static void NoResizeClear<TKey, TValue>(this ConcurrentDictionary<TKey, TValue> dictionary)
+                where TKey : notnull
+        {
+            if (dictionary is null || dictionary.IsEmpty)
```
LukaszRozmej, line 36: make it `ConcurrentDictionary<TKey, TValue>?` if nullable?

src/Nethermind/Nethermind.Core/Collections/CollectionExtensions.cs
```diff
@@ -22,5 +29,61 @@ public static void AddRange<T>(this ICollection<T> list, params T[] items)
...
+                    countPerLockAccess,
+                    Expression.Constant(0),
+                    Expression.ArrayLength(countPerLockAccess));
+
+                // Block to execute both clears
+                var block = Expression.Block(clearBuckets, clearCountPerLock);
+
+                // Compile the expression into a lambda
+                return Expression.Lambda<Action<ConcurrentDictionary<TKey, TValue>>>(block, dictionaryParam).Compile();
+            }
```
LukaszRozmej, line 85: Nice hack FTW!

Can we add a test for it? For example if we upgrade to .Net X and internals change the test should fail.

src/Nethermind/Nethermind.State/PreBlockCaches.cs
```diff
@@ -5,28 +5,101 @@
...
+        _clearStorageCache = () => _storageCache.NoResizeClear();
+        _clearStateCache = () => _stateCache.NoResizeClear();
+        _clearRlpCache = () => _rlpCache.NoResizeClear();
+        _clearPrecompileCache = () => _precompileCache.NoResizeClear();
+    }
+
+    public ConcurrentDictionary<StorageCell, byte[]> StorageCache => _storageCache;
+    public ConcurrentDictionary<AddressAsKey, Account> StateCache => _stateCache;
+    public ConcurrentDictionary<NodeKey, byte[]?> RlpCache { get; } = new(LockPartitions, InitialCapacity);
+    public ConcurrentDictionary<PrecompileCacheKey, (ReadOnlyMemory<byte>, bool)> PrecompileCache { get; } = new(LockPartitions, InitialCapacity);
```
LukaszRozmej, line 44: We have a property and a field initialized with `new`? Probably a mistake.

src/Nethermind/Nethermind.State/PreBlockCaches.cs
```diff
@@ -5,28 +5,101 @@
...
+        if (!_rlpCache.IsEmpty)
+        {
+            isDirty = true;
+            t2 = Task.Run(_clearRlpCache);
+        }
+        if (!_precompileCache.IsEmpty)
+        {
+            isDirty = true;
+            t3 = Task.Run(_clearPrecompileCache);
+        }
```
LukaszRozmej, line 73: put them in an array and do a loop? Will make code easier if we add more caches.

src/Nethermind/Nethermind.State/PreBlockCaches.cs
```diff
@@ -5,28 +5,101 @@
...
+        _clearPrecompileCache = () => _precompileCache.NoResizeClear();
+    }
+
+    public ConcurrentDictionary<StorageCell, byte[]> StorageCache => _storageCache;
+    public ConcurrentDictionary<AddressAsKey, Account> StateCache => _stateCache;
+    public ConcurrentDictionary<NodeKey, byte[]?> RlpCache { get; } = new(LockPartitions, InitialCapacity);
+    public ConcurrentDictionary<PrecompileCacheKey, (ReadOnlyMemory<byte>, bool)> PrecompileCache { get; } = new(LockPartitions, InitialCapacity);
 
-    public bool Clear()
+    public async Task ClearCachesInBackground()
```
LukaszRozmej, line 46: Why we benefit from background clearing?

src/Nethermind/Nethermind.State/PreBlockCaches.cs
```diff
@@ -5,28 +5,101 @@
...
+    private readonly Action _clearPrecompileCache;
+
+    private ConcurrentDictionary<StorageCell, byte[]> _storageCache = new(LockPartitions, InitialCapacity);
+    private ConcurrentDictionary<AddressAsKey, Account> _stateCache = new(LockPartitions, InitialCapacity);
+    private ConcurrentDictionary<NodeKey, byte[]?> _rlpCache = new(LockPartitions, InitialCapacity);
+    private ConcurrentDictionary<PrecompileCacheKey, (ReadOnlyMemory<byte>, bool)> _precompileCache = new(LockPartitions, InitialCapacity);
+
+    public PreBlockCaches(ILogManager logManager)
+    {
+        _logger = logManager?.GetClassLogger<WorldState>() ?? throw new ArgumentNullException(nameof(logManager));
```
LukaszRozmej, line 34: logManager and logger not used - remove

src/Nethermind/Nethermind.Core/Caching/ClockCache.cs
benaadams: Still do or it could have cleared one array but not the other and mess up a reader

src/Nethermind/Nethermind.State/PreBlockCaches.cs
benaadams: "The fix was a lie"

Yes mistake 😅

src/Nethermind/Nethermind.State/PreBlockCaches.cs
benaadams: Removed

src/Nethermind/Nethermind.State/PreBlockCaches.cs
benaadams: Can attach it to continuation of perwarmer (post cancellation) so its likely complete before its actually required to be cleared and not take up time in main thread

src/Nethermind/Nethermind.State/PreBlockCaches.cs
benaadams: Done

src/Nethermind/Nethermind.Core/Caching/ClockCache.cs
benaadams: The reader side isn't under lock

LukaszRozmej: ok you convinced me.

LukaszRozmej (approved):

src/Nethermind/Nethermind.Core/Collections/CollectionExtensions.cs
benaadams: Added test

PR #7362, opened 2024-08-26
Support for state override parameter in some RPC methods

### LINKED ISSUE #4021: Support for state overrides on eth_call and trace_call
It would be great to get the ability to do state overrides with nethermind.

State overrides are supported in geth for `eth_call` and `trace_call`.

Here's an example snippet (from geth):
`debug.traceCall({from: addr, to: zero, value: '0x1'}, 'latest', {'stateOverrides': {[addr]: {'balance': '0x0'}}})`

This blog post explains the use case for state overrides quite well: https://www.libevm.com/2022/01/12/advance-geth-pt-2-stateoverrides/

### LINKED ISSUE #6120: Support for blockOverride in eth_call, eth_estimateGas, eth_createAccessList
We need one more parameter: blockOverride to methods eth_call, eth_estimateGas, eth_createAccessList. https://github.com/NethermindEth/nethermind/blob/master/src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs#L337

This parameter should override values specified here: https://github.com/NethermindEth/nethermind/blob/master/src/Nethermind/Nethermind.Facade/BlockchainBridge.cs#L245 It should override parameters no matter if `treatBlockHeaderAsParentBlock` is true or false.

Geth has already implemented it, but only for eth_call and it could be treated as a reference. List of parameters that could be overridden can be found here: https://github.com/NethermindEth/nethermind/blob/master/src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs#L337

### LINKED ISSUE #7306: [Gnosis] eth_call - Invalid params
**Description**
We are getting an unexpected response when using `eth_call` against our Gnosis Mainnet nodes.

**Steps to Reproduce**

Example request:

`curl -H 'Content-Type: application/json' \
--data-raw '{"id": "1","method":"eth_call","params":[{"to":"0x74Cb5e4eE81b86e70f9045036a1C5477de69eE87","data":"0xc18f52260000000000000000000000000000000071727de22e5e9d8baf0edac6f37da03200000000000000000000000000000000000000000000000000000000000000400000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000009a00000000000000000000000000000000000000000000000000000000000000924263934db000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000020000000000000000000000000df4517e772e83eb4e8d082bcd6acbe6428b610e0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001200000000000000000000000000000000000000000000000000000000000000540000000000000000000000000009896800000000000000000000000000098968000000000000000000000000000000000000000000000000000000000000f424000000000000000000000000067f16f1f00000000000000000000000067f16f1f0000000000000000000000000000000000000000000000000000000000000620000000000000000000000000000000000000000000000000000000000000064000000000000000000000000000000000000000000000000000000000000003f84e1DCf7AD4e460CfD30791CCC4F9c8a4f820ec671688f0b900000000000000000000000029fcb43b46531bca003ddc8fcb67ffe91900c762000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000344b63e800d0000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000000100000000000000000000000038869bf66a61cf6bdb996a6ae40d5853fd43b526000000000000000000000000000000000000000000000000000000000000014000000000000000000000000075cf11467937ce3f2f357ce24ffc3dbf8fd5c2260000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001000000000000000000000000fd90fad33ee8b58f32c00aceead1358e4afc23f900000000000000000000000000000000000000000000000000000000000001c48d80ff0a00000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000172012dd68b007b46fbe91b9a7c3eda5a7a1063cb5b47000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000648d0dc49f0000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000100000000000000000000000075cf11467937ce3f2f357ce24ffc3dbf8fd5c22601fd90fad33ee8b58f32c00aceead1358e4afc23f9000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000640dd9692fe37347be72e261ac1f27f2cbd51b304ca058ee8c9b7efc36c414507fa75f1a5c461e134b6d59c17e25970052fcb7df3eabe4da7a8d614de481668a7d11433a92000000000000000000000000445a0683e494ea0c5af3e83c5159fbe47cf9e76500000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000a47bb37428000000000000000000000000f64da4efa19b42ef2f897a3d533294b892e6d99e0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000024d000000000000000000000000000000000000000000000000fd90fad33ee8b58f32c00aceead1358e4afc23f900000000000000000000000000000000000000000000000000000000000000410000000000000000000000000000000000000000000000000000000000000001e0000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000000000e0ececececececececececececececececececececececececececececececececd5ad5ad5ad5ad5ad5ad5ad5ad5ad5ad5ad5ad5ad5ad5ad5ad5ad5ad5ad5ad5af0000000000000000000000000000000000000000000000000000000000000025fefefefefefefefefefefefefefefefefefefefefefefefefefefefefefefefe04fefefefe00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000c1226f726967696e223a22687474703a2f2f736166652e676c6f62616c222c2270616464696e67223a225468697320706164732074686520636c69656e74446174614a534f4e20736f20746861742077652063616e206c6561766520726f6f6d20666f72206164646974696f6e616c20696d706c656d656e746174696f6e207370656369666963206669656c647320666f722061206d6f72652061636375726174652027707265566572696669636174696f6e4761732720657374696d6174652e220000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000ac4baa220440000000000000000000000000000000000000000000000000000000000000060000000000000000000000000000000000000000000000000000000000000096000000000000000000000000000000000000000000000000000000000000009a000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000020000000000000000000000000df4517e772e83eb4e8d082bcd6acbe6428b610e0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001200000000000000000000000000000000000000000000000000000000000000540000000000000000000000000009896800000000000000000000000000098968000000000000000000000000000000000000000000000000000000000000f424000000000000000000000000067f16f1f00000000000000000000000067f16f1f0000000000000000000000000000000000000000000000000000000000000620000000000000000000000000000000000000000000000000000000000000064000000000000000000000000000000000000000000000000000000000000003f84e1DCf7AD4e460CfD30791CCC4F9c8a4f820ec671688f0b900000000000000000000000029fcb43b46531bca003ddc8fcb67ffe91900c762000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000344b63e800d0000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000000100000000000000000000000038869bf66a61cf6bdb996a6ae40d5853fd43b526000000000000000000000000000000000000000000000000000000000000014000000000000000000000000075cf11467937ce3f2f357ce24ffc3dbf8fd5c2260000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001000000000000000000000000fd90fad33ee8b58f32c00aceead1358e4afc23f900000000000000000000000000000000000000000000000000000000000001c48d80ff0a00000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000172012dd68b007b46fbe91b9a7c3eda5a7a1063cb5b47000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000648d0dc49f0000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000100000000000000000000000075cf11467937ce3f2f357ce24ffc3dbf8fd5c22601fd90fad33ee8b58f32c00aceead1358e4afc23f9000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000640dd9692fe37347be72e261ac1f27f2cbd51b304ca058ee8c9b7efc36c414507fa75f1a5c461e134b6d59c17e25970052fcb7df3eabe4da7a8d614de481668a7d11433a92000000000000000000000000445a0683e494ea0c5af3e83c5159fbe47cf9e76500000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000a47bb37428000000000000000000000000f64da4efa19b42ef2f897a3d533294b892e6d99e0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000024d000000000000000000000000000000000000000000000000fd90fad33ee8b58f32c00aceead1358e4afc23f900000000000000000000000000000000000000000000000000000000000000410000000000000000000000000000000000000000000000000000000000000001e0000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000000000e0ececececececececececececececececececececececececececececececececd5ad5ad5ad5ad5ad5ad5ad5ad5ad5ad5ad5ad5ad5ad5ad5ad5ad5ad5ad5ad5af0000000000000000000000000000000000000000000000000000000000000025fefefefefefefefefefefefefefefefefefefefefefefefefefefefefefefefe04fefefefe00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000c1226f726967696e223a22687474703a2f2f736166652e676c6f62616c222c2270616464696e67223a225468697320706164732074686520636c69656e74446174614a534f4e20736f20746861742077652063616e206c6561766520726f6f6d20666f72206164646974696f6e616c20696d706c656d656e746174696f6e207370656369666963206669656c647320666f722061206d6f72652061636375726174652027707265566572696669636174696f6e4761732720657374696d6174652e2200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001000000000000000000000000df4517e772e83eb4e8d082bcd6acbe6428b610e00000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000a47bb37428000000000000000000000000f64da4efa19b42ef2f897a3d533294b892e6d99e00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000080000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"},"latest",{"0xdf4517e772e83eb4E8d082bCd6aCbe6428b610e0":{"balance":"0x152d02c7e14af6800000"}}]}' \
localhost:8545`

Response:

`{"jsonrpc":"2.0","error":{"code":-32602,"message":"Invalid params","data":"Incorrect parameters count, expected: 2, actual: 3"},"id":"1"}`

**Actual behavior**
We should get an actual response, as the call works on other binaries, such as Erigon.

**Expected behavior**

Expected response:

`{"jsonrpc":"2.0","id":"1","result":"0x000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000024000000000000000000000000000000000000000000000000000000000000001c49941055400000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000016000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000191e82000000000000000000000000000000000000000000000000001075eb54e26b7900000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000009300000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000000023078000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000104994105540000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000a000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000002b1c00000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000060000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"}`

**Environment**
 
- Operating System:  Ubuntu 22.04 LTS
 - Version:  Nethermind 1.27.1+d2d5d4ae
 - Installation Method: GitHub Release
 - Consensus Client: Lighthouse v5.2.1-9e12c21

**Logs**

image

alexb5dh:
Resolves #4021, resolves #6120, resolves #7306.

## Changes

- Support for state override set in the following RPC methods:
  -  `eth_call`, `eth_estimategas`
  - `trace_call`
  - `debug_traceCall`
- Makes following methods to run on top of the specified block instead of the previous one, similar to other clients:
  - `trace_call`, `trace_callMany`, `trace_rawTransaction`
  - `debug_traceCall`
- Use corresponding factory to create module for `TraceRpcModuleTests` and `DebugRpcModuleTests` - to make it more consistent with regular setup.
- ~~Fixes double-serialization of some of the parameters in the RPC tests.~~

#### Requires testing

- [x] Yes

- [x] Yes

- [x] Yes

- [x] Yes, RPC changes should probably be included.


src/Nethermind/Nethermind.JsonRpc/Modules/Trace/TraceRpcModule.cs
```diff
@@ -59,20 +56,20 @@ public static ParityTraceTypes GetParityTypes(string[] types) =>
         /// <summary>
         /// Traces one transaction. Doesn't charge fees.
         /// </summary>
-        public ResultWrapper<ParityTxTraceFromReplay> trace_call(TransactionForRpc call, string[] traceTypes, BlockParameter? blockParameter = null)
+        public ResultWrapper<ParityTxTraceFromReplay> trace_call(TransactionForRpc call, string[] traceTypes, BlockParameter? blockParameter = null, Dictionary<Address, AccountOverride>? stateOverride = null)
```
alexb5dh, line 59: `eth_call`, and `eth_estimategas` are actually the only methods where state override is supported in [Geth](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-eth#eth-call), and [some popular RPC providers](https://www.quicknode.com/docs/ethereum/eth_estimateGas).

While this PR extends it to other methods, there is a risk of a future signature mismatch, in case other clients update them differently. For example, add `options` as the next parameter, similar to `debug_traceCall`.

For these cases, it may be better to introduce new methods instead (like `nethermind_traceCall` or `trace_callv2`) with the same logic, but a new parameter added. and leave old ones intact.

src/Nethermind/Nethermind.Evm/StateOverridesExtensions.cs
```diff
@@ -43,9 +41,12 @@ public static void ApplyStateOverrides(
             }
         }
 
-        state.Commit(spec);
-        state.CommitTree(blockNumber);
-        state.RecalculateStateRoot();
+        if (commit)
```
alexb5dh, line 44: `eth_simulate` implementation does commit after an override, but looks like other methods do not need it. Still testing all the cases and will resolve this once sure.

src/Nethermind/Nethermind.JsonRpc/Modules/Trace/TraceRpcModule.cs
LukaszRozmej: Can you check what Reth does? If it doesn't add it than maybe we shouldn't either.
Other than that I wouldn't sweat about it, those methods are stable for years now

alexb5dh: Yeah, that did add it to some other methods apart from 2 Geth does:
- [eth](https://github.com/paradigmxyz/reth/blob/main/crates/rpc/rpc-eth-api/src/core.rs): `eth_call`, `eth_callMany`, `eth_estimateGas` (but not `eth_createAccessList` as requested in one of the linked issues)
- [trace](https://github.com/paradigmxyz/reth/blob/main/crates/rpc/rpc-api/src/trace.rs): just for `trace_call` (not for `trace_call_many` for some reason)
- [debug](https://github.com/paradigmxyz/reth/blob/main/crates/rpc/rpc-api/src/debug.rs): `debug_trace_call`, `debug_trace_call_many` (in all cases nested in 3rd `options` parameter)

LukaszRozmej: let's mimic Reth api then

src/Nethermind/Nethermind.Facade/Eth/TransactionForRpc.cs
```diff
@@ -176,7 +176,7 @@ public TransactionForRpc() { }
         return tx;
     }
 
-    public Transaction ToTransaction(ulong? chainId = null) => ToTransaction<Transaction>();
+    public Transaction ToTransaction(ulong? chainId = null) => ToTransaction<Transaction>(chainId);
```
alexb5dh, line 179: Not sure it not passing `chainId` was intended, but looks like a bug to me.

src/Nethermind/Nethermind.Trie/Pruning/OverlayTrieStore.cs
```diff
@@ -24,4 +23,7 @@ public override TrieNode FindCachedOrUnknown(Hash256? address, in TreePath path,
 
     public override byte[]? TryLoadRlp(Hash256? address, in TreePath path, Hash256 hash, ReadFlags flags = ReadFlags.None) =>
         base.TryLoadRlp(address, in path, hash, flags) ?? store.TryLoadRlp(address, in path, hash, flags);
+
+    // TODO clarify is ClearCache is reliable enough to use
+    public void ResetOverrides() => ClearCache();
```
alexb5dh, line 28: Comment under `ClearCache`  says that it's added "to support testing", but it seems like a good way to reset changes applied on top of `OverlayTrieStore` from state overrides.

src/Nethermind/Nethermind.State/OverridableWorldStateManager.cs
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    public IWorldState GlobalWorldState => _state;
+    public IStateReader GlobalStateReader => _reader;
+    public IReadOnlyTrieStore TrieStore => _overlayTrieStore.AsReadOnly();
+
+    public IWorldState CreateResettableWorldState(IWorldState? forWarmup = null)
+    {
+        // TODO add if needed?
+        if (forWarmup is not null)
```
alexb5dh, line 39: As of now `OverridableWorldStateManager.CreateResettableWorldState()` with non-null `forWarmup` is not used, but this `NotSupportedException` also doesn't look good.

Implementing missing part is a bit problematic due to `PreCachedTrieStore` not having an alternative for `TrieStore.ClearCache()`, which is used by `OverridableWorldState`.

src/Nethermind/Nethermind.Evm/StateOverridesExtensions.cs
```diff
@@ -71,13 +69,15 @@ void ApplyState(Dictionary<UInt256, Hash256> diff)
...
         this IWorldState stateProvider,
-        OverridableCodeInfoRepository overridableCodeInfoRepository,
+        IOverridableCodeInfoRepository overridableCodeInfoRepository,
         IReleaseSpec currentSpec,
         AccountOverride accountOverride,
         Address address)
     {
         if (accountOverride.Code is not null)
         {
+            stateProvider.InsertCode(address, accountOverride.Code, currentSpec);
```
alexb5dh, line 79: This is needed to prevent an "Account {address} is null when updating storage hash" error in `StateProvider` when overridden code tries to modify address storage.

src/Nethermind/Nethermind.Api/NethermindApi.cs
```diff
@@ -76,10 +76,11 @@ public NethermindApi(IConfigProvider configProvider, IJsonSerializer jsonSeriali
         public IBlockchainBridge CreateBlockchainBridge()
         {
             ReadOnlyBlockTree readOnlyTree = BlockTree!.AsReadOnly();
+            OverridableWorldStateManager overridableWorldStateManager = new(DbProvider!, WorldStateManager!.TrieStore, LogManager);
 
             // TODO: reuse the same trie cache here
             ReadOnlyTxProcessingEnv readOnlyTxProcessingEnv = new(
```
LukaszRozmej, line 82: Maybe let's create `OverridableTxProcessingEnv`?

src/Nethermind/Nethermind.Consensus/Processing/ReadOnlyTxProcessingScope.cs
```diff
@@ -2,12 +2,14 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 
 using Nethermind.Core.Crypto;
+using Nethermind.Evm;
 using Nethermind.Evm.TransactionProcessing;
 using Nethermind.State;
 
 namespace Nethermind.Consensus.Processing;
 
 public class ReadOnlyTxProcessingScope(
```
LukaszRozmej, line 11: maybe move to `OverridableTxProcessingScope`?

src/Nethermind/Nethermind.Evm/StateOverridesExtensions.cs
LukaszRozmej: @OlegJakushkin maybe this solves some issues on StateRoot in eth_simulate?

src/Nethermind/Nethermind.Facade/IBlockchainBridge.cs
```diff
@@ -28,9 +27,10 @@ public interface IBlockchainBridge : ILogFinder
         TxReceipt GetReceipt(Hash256 txHash);
         (TxReceipt? Receipt, TxGasInfo? GasInfo, int LogIndexStart) GetReceiptAndGasInfo(Hash256 txHash);
         (TxReceipt? Receipt, Transaction? Transaction, UInt256? baseFee) GetTransaction(Hash256 txHash, bool checkTxnPool = true);
-        CallOutput Call(BlockHeader header, Transaction tx, CancellationToken cancellationToken);
+        CallOutput Call(BlockHeader header, Transaction tx, Dictionary<Address, AccountOverride>? stateOverride, CancellationToken cancellationToken);
+        CallOutput Call(BlockHeader header, Transaction tx, CancellationToken cancellationToken) => Call(header, tx, null, cancellationToken);
```
LukaszRozmej, line 31:
```suggestion
        CallOutput Call(BlockHeader header, Transaction tx, Dictionary<Address, AccountOverride>? stateOverride = null, CancellationToken cancellationToken = default);
```

src/Nethermind/Nethermind.State/OverlayWorldStateManager.cs
```diff
@@ -29,17 +29,11 @@ public class OverlayWorldStateManager(
...
-                _codeDb,
-                logManager,
-                preBlockCaches)
-            : new WorldState(
-                overlayTrieStore,
-                _codeDb,
-                logManager);
+        ITrieStore trieStore = overlayTrieStore;
+        if ((forWarmup as IPreBlockCaches)?.Caches is { } preBlockCaches)
+            trieStore = new PreCachedTrieStore(trieStore, preBlockCaches.RlpCache);
```
LukaszRozmej, line 34:
```suggestion
        ITrieStore trieStore = (forWarmup as IPreBlockCaches)?.Caches is { } preBlockCaches ? new PreCachedTrieStore(trieStore, preBlockCaches.RlpCache) : overlayTrieStore;
```

LukaszRozmej (commented):
1. I think it would be good to separate ReadOnlyTxProcessingEnv with OverridableTxProcessingEnv instead of changing the first.

2. That double test serialization was also changed in different PR #7547

src/Nethermind/Nethermind.Facade/IBlockchainBridge.cs
alexb5dh: That was initially done to avoid changes in `BlockchainBridgeTests`, but it's probably a bit overdue, with 40+ changed files 😅

src/Nethermind/Nethermind.State/OverlayWorldStateManager.cs
alexb5dh: `trieStore` will be undefined in `new PreCachedTrieStore(trieStore, preBlockCaches.RlpCache)`

src/Nethermind/Nethermind.JsonRpc/Modules/Trace/TraceRpcModule.cs
```diff
@@ -144,15 +152,28 @@ private SearchResult<BlockHeader> SearchBlockHeaderForTraceCall(BlockParameter b
...
             {
                 return ResultWrapper<ParityTxTraceFromReplay>.Fail(headerSearch);
             }
 
-            Block block = new(headerSearch.Object!, new[] { tx }, Enumerable.Empty<BlockHeader>());
+            BlockHeader header = headerSearch.Object!.Clone();
+
+            // Fake next-block header to force BlockProcessor to start from <blockParameter> instead of <blockParameter> - 1
+            // TODO consider alternative approaches to achieve this
+            BlockHeader nextHeader = header.Clone();
```
alexb5dh, line 168: `TraceBlock` implementation called below will rollback to the previous block before starting processing, as needed for most of the `trace_*` methods. This code creates a fake "next" header, to actually enforce executing `tx` to use thestate from the same block as provided in `blockParameter` for `trace_call` and `trace_rawTransaction`.

This doesn't look ideal and may break with changes to `BlockProcessor`, but it gets the job done, without a lot of refactoring.

One of the alternative approaches can be to extend `processingOptions`, yet it may be a bit cumbersome.

src/Nethermind/Nethermind.JsonRpc.Test/RpcTest.cs
```diff
@@ -63,9 +59,19 @@ private static IJsonRpcService BuildRpcService<T>(T module) where T : class, IRp
         return service;
     }
 
-    public static JsonRpcRequest GetJsonRequest(string method, params string[]? parameters)
```
alexb5dh, line 66: Regarding similar RPC tests fixes in #7547 - I see it's still not in master and actually waiting in #7483.
- Reverting my fixes will break all or most of new tests in this PR
- I can try to cherry-pick #7547, but that will add around 10 unrelated files to changes.

src/Nethermind/Nethermind.JsonRpc.Benchmark/EthModuleBenchmarks.cs
```diff
@@ -66,7 +66,7 @@ public void GlobalSetup()
             stateProvider.Commit(spec);
             stateProvider.CommitTree(0);
 
-            WorldStateManager stateManager = new WorldStateManager(stateProvider, trieStore, dbProvider, LimboLogs.Instance);
+            OverridableWorldStateManager stateManager = new(dbProvider, trieStore.AsReadOnly(), LimboLogs.Instance);
```
alexb5dh, line 69: There's actually what seems like a bug, present even in `master` branch - balance set a few lines above is not returned in benchmarked `eth_getBalance` call, meaning setup is probably incorrect.

src/Nethermind/Nethermind.JsonRpc.Test/RpcTest.cs
emlautarom1: #7483 is now on master now

alexb5dh: I personally liked initial approach better, one with the conditional serialization and custom array construction. While not looking super nice, it accepted both objects and serialized JSON strings as parameters, and latter ones are much easier to pass into tests via `TestCase` attribute.

emlautarom1: Your point makes sense, specially when dealing with `TestCase` attributes. Feel free to create a PR if you want to revisit this topic

src/Nethermind/Nethermind.Consensus/Tracing/Tracer.cs
```diff
@@ -16,18 +16,20 @@ public class Tracer : ITracer
         private readonly IWorldState _stateProvider;
         private readonly IBlockchainProcessor _traceProcessor;
         private readonly IBlockchainProcessor _executeProcessor;
-        private readonly ProcessingOptions _processingOptions;
+        private readonly ProcessingOptions _executeOptions;
+        private readonly ProcessingOptions _traceOptions;
 
         public Tracer(IWorldState stateProvider, IBlockchainProcessor traceProcessor, IBlockchainProcessor executeProcessor,
-            ProcessingOptions processingOptions = ProcessingOptions.Trace)
+            ProcessingOptions executeOptions = ProcessingOptions.Trace, ProcessingOptions traceOptions = ProcessingOptions.Trace)
```
LukaszRozmej, line 23: Shouldn't `Execute` be always `Trace`
And `Trace` be always `TraceTransaction`
And we don't need the parameters?

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
```diff
@@ -232,11 +236,13 @@ private TransactionResult TryCallAndRestore(
...
             bool treatBlockHeaderAsParentBlock,
-            ITxTracer tracer)
+            ITxTracer tracer,
+            IOverridableTxProcessingScope? scope = null)
         {
             try
             {
-                return CallAndRestore(blockHeader, transaction, treatBlockHeaderAsParentBlock, tracer);
+                return CallAndRestore(blockHeader, transaction, treatBlockHeaderAsParentBlock, tracer,
+                    scope ?? _processingEnv.Build(blockHeader.StateRoot!));
```
LukaszRozmej, line 245: Maybe either always build scope here, or always build scope in the caller?

src/Nethermind/Nethermind.JsonRpc/Modules/Trace/TraceModuleFactory.cs
```diff
@@ -31,7 +30,7 @@ public class TraceModuleFactory(
     IPoSSwitcher poSSwitcher,
     ILogManager logManager) : ModuleFactoryBase<ITraceRpcModule>
 {
-    protected readonly IWorldStateManager _worldStateManager = worldStateManager;
+    protected readonly OverridableWorldStateManager _worldStateManager = new(dbProvider, worldStateManager.TrieStore, logManager);
```
LukaszRozmej, line 33: Shouldn't each TraceModule have their own `OverridableWorldStateManager` instead of all sharing one?
If they all share one overrides from one call, can affect the state of other call?

src/Nethermind/Nethermind.JsonRpc/Modules/Trace/TraceRpcModule.cs
```diff
@@ -110,46 +118,19 @@ public ResultWrapper<ParityTxTraceFromReplay> trace_rawTransaction(byte[] data,
...
-                    header.Timestamp + 1,
-                    header.ExtraData,
-                    header.BlobGasUsed,
-                    header.ExcessBlobGas);
-
-                header.TotalDifficulty = 2 * header.Difficulty;
-                header.BaseFeePerGas = baseFee;
-            }
-
-            return new SearchResult<BlockHeader>(header);
```
LukaszRozmej, line 141: Is this now not needed?

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/ExecutorBase.cs
```diff
@@ -23,7 +25,8 @@ protected ExecutorBase(IBlockchainBridge blockchainBridge, IBlockFinder blockFin
 
     public virtual ResultWrapper<TResult> Execute(
         TRequest call,
-        BlockParameter? blockParameter)
+        BlockParameter? blockParameter,
+        Dictionary<Address, AccountOverride>? stateOverride = null)
```
LukaszRozmej, line 29: ah that's is inconvenient, we have to pass it here?

src/Nethermind/Nethermind.JsonRpc/Modules/DebugModule/DebugModuleFactory.cs
```diff
@@ -75,11 +74,12 @@ public DebugModuleFactory(
         _badBlockStore = badBlockStore;
         _fileSystem = fileSystem ?? throw new ArgumentNullException(nameof(fileSystem));
         _logger = logManager.GetClassLogger();
+        _worldStateManager = new(dbProvider, worldStateManager.TrieStore, logManager);
```
LukaszRozmej, line 77: Same thing as with trace, if we have one world state manager for multiple instances, then we have one in-memory db, that can side effect between calls?

LukaszRozmej (changes requested):

src/Nethermind/Nethermind.JsonRpc/Modules/Trace/TraceRpcModule.cs
alexb5dh: As I understood, It was used to fix issues with rollback to the previous block being applied on the genesis - for `trace_call`, `trace_callMany` and `trace_rawTransaction`. Not needed as I removed this rollback for those methods.

src/Nethermind/Nethermind.JsonRpc/Modules/Trace/TraceModuleFactory.cs
alexb5dh: Yes, fixing it.

Used some basic code to reproduce it, but adding unit tests for this case that would be good enough to merge to master is a bit problematic with current implementation - it uses a single `TraceRpcModule` per context.

src/Nethermind/Nethermind.Trie/Pruning/OverlayTrieStore.cs
asdacap: I dont see a problem.

src/Nethermind/Nethermind.State/OverridableWorldStateManager.cs
```diff
@@ -0,0 +1,49 @@
...
+    private readonly IDb _codeDb;
+    private readonly StateReader _reader;
+    private readonly WorldState _state;
+
+    private readonly OverlayTrieStore _overlayTrieStore;
+    private readonly ILogManager? _logManager;
+
+    public OverridableWorldStateManager(IDbProvider dbProvider, IReadOnlyTrieStore trieStore, ILogManager? logManager)
+    {
+        dbProvider = new ReadOnlyDbProvider(dbProvider, true);
```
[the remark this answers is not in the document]
alexb5dh: Fixed

src/Nethermind/Nethermind.State/OverridableWorldStateManager.cs
alexb5dh: `PreCachedTrieStore` need to be wrapped into `OverlayTrieStore` to do a post-call cleanup, but latter requires `IReadOnlyTrieStore` interface.

It probably can be fixed in some way, just not sure it's worth the trouble at the moment.

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
```diff
@@ -248,12 +256,11 @@ private TransactionResult CallAndRestore(
             BlockHeader blockHeader,
             Transaction transaction,
             bool treatBlockHeaderAsParentBlock,
-            ITxTracer tracer)
+            ITxTracer tracer,
+            IOverridableTxProcessingScope scope)
```
LukaszRozmej, line 260: Minor: here and `TryCallAndRestore`. I think `scope` should be earlier in parameters, either 1st or 3rd. Would make it more readable and have most important parameters first.

LukaszRozmej (approved):

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
```diff
@@ -248,12 +256,11 @@ private TransactionResult CallAndRestore(
...
         {
             transaction.SenderAddress ??= Address.SystemUser;
-
-            Hash256 stateRoot = blockHeader.StateRoot!;
-            using IReadOnlyTxProcessingScope scope = _processingEnv.Build(stateRoot);
+            Hash256? stateRoot = blockHeader.StateRoot!;
```
emlautarom1, line 263: Is this nullable or not? The usage of both `?` and `!` makes it confusing

src/Nethermind/Nethermind.JsonRpc.Test/Modules/Eth/EthRpcModuleTests.EstimateGas.cs
```diff
@@ -257,4 +258,76 @@ public async Task should_estimate_transaction_with_deployed_code_when_eip3607_en
...
+
+        var resultOverrideBefore = await ctx.Test.TestEthRpc("eth_estimateGas", transaction, "latest", stateOverride);
+
+        var resultNoOverride = await ctx.Test.TestEthRpc("eth_estimateGas", transaction, "latest");
+
+        var resultOverrideAfter = await ctx.Test.TestEthRpc("eth_estimateGas", transaction, "latest", stateOverride);
+
+        Assert.Multiple(() =>
+        {
+            Assert.That(resultOverrideBefore, Is.EqualTo(resultOverrideAfter), resultOverrideBefore.Replace("\"", "\\\""));
```
emlautarom1, line 329: Prefer JSON over string comparison since it ignores differences like key ordering:

```suggestion
            using FluentAssertions.Json;

            JToken.Parse(resultOverrideBefore).Should().BeEquivalentTo(resultOverrideAfter);
```

emlautarom1 (approved):
LGTM

src/Nethermind/Nethermind.JsonRpc.Test/Modules/Eth/EthRpcModuleTests.EstimateGas.cs
alexb5dh: This looks nice, thanks

OlegJakushkin (approved):
Perfect! solves some issues on StateRoot in eth_simulate.

PR #7374, opened 2024-08-30
Add metric for new payload including all Json

benaadams:
## Changes

- Add metric for new payload including all Json

#### Requires testing

- [x] No


src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
```diff
@@ -267,10 +269,16 @@ JsonRpcResult GetParsingError(string error, Exception? exception = null)
...
-        }
 
-        // Completes the PipeReader's asynchronous reading operation.
-        await reader.CompleteAsync();
+            // Completes the PipeReader's asynchronous reading operation.
+            await reader.CompleteAsync();
+
+            if (isNewPayload)
+            {
+                Metrics.JsonRpcNewPayloadTime = stopwatch.ElapsedMilliseconds;
```
LukaszRozmej, line 278: If we want to add something lets add a dictionary and measure all the different types of RPC calls?

Probably would need to somehow aggregate multiple calls?

PR #7375, opened 2024-08-31
Only need single stack for stacks

benaadams:
## Changes

- Only need a single Stack for stacks as they are always done in pairs so use single one
- Using struct for Stack (that has both arrays) avoids calls to StelemRef to type check from having and Array as the generic param
- Halves the number of operations on Stack and depth book keeping

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Evm/EvmState.cs
+++ b/src/Nethermind/Nethermind.Evm/EvmState.cs
@@ -23,6 +23,11 @@ namespace Nethermind.Evm
         private class StackPool
         {
             private readonly int _maxCallStackDepth;
+            private readonly struct StackItem(byte[] dataStack, int[] returnStack)
+            {
+                public readonly byte[] DataStack = dataStack;
+                public readonly int[] ReturnStack = returnStack;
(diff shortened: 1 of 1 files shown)


Scooletz (approved):
Was looking at this exactly thing 😄

src/Nethermind/Nethermind.Evm/EvmState.cs
```diff
@@ -44,45 +47,27 @@ public StackPool(int maxCallStackDepth = VirtualMachine.MaxCallDepth * 2)
...
-            }
-
-            private int[] RentReturnStack()
+            public (byte[], int[]) RentStacks()
             {
-                if (_returnStackPool.TryPop(out int[] result))
+                if (_stackPool.TryPop(out StackItem result))
                 {
-                    return result;
+                    return (result.DataStack, result.ReturnStack);
```
LukaszRozmej, line 57: lets avoid returning new tuple here, return StackItem directly and add deconstructor to StackItem.

PR #7377, opened 2024-09-01
Add dual array concat overload

benaadams:
## Changes

- Add dual array overload vs `params byte[][]` where it uses two for loops. Is used by 17 call sites

image

- [x] Optimization

#### Requires testing

- [x] No


LukaszRozmej:
Can you look into the usages, maybe they can be optimized?

benaadams:
> Can you look into the usages, maybe they can be optimized?

https://github.com/NethermindEth/nethermind/pull/8007

PR #7381, opened 2024-09-03
Halve Hash256 creations

benaadams:
Before

image

After

image

## Changes

- _List the changes_

- [x] Optimization

#### Requires testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Remarks

### DIFF
--- a/src/Nethermind/Nethermind.Synchronization.Test/TrieScenarios.cs
+++ b/src/Nethermind/Nethermind.Synchronization.Test/TrieScenarios.cs
@@ -341,7 +341,7 @@ namespace Nethermind.Synchronization.Test
             return SetStorage(trieStore, account.ToAccountPath);
         }
 
-        private static StorageTree SetStorage(ITrieStore trieStore, Hash256 account)
+        private static StorageTree SetStorage(ITrieStore trieStore, in ValueHash256 account)
         {
             StorageTree remoteStorageTree = new(trieStore.GetTrieStore(account), Keccak.EmptyTreeHash, LimboLogs.Instance);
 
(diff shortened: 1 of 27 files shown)


src/Nethermind/Nethermind.Synchronization.Test/TrieScenarios.cs
```diff
@@ -363,7 +363,7 @@ private static void SetStorage(StateTree stateTree, ITrieStore trieStore, Addres
             SetStorage(stateTree, trieStore, address.ToAccountPath, account);
         }
 
-        private static void SetStorage(StateTree stateTree, ITrieStore trieStore, Hash256 address, Account account)
+        private static void SetStorage(StateTree stateTree, ITrieStore trieStore, ValueHash256 address, Account account)
```
LukaszRozmej, line 366: why not `in`?

src/Nethermind/Nethermind.Trie/NodeStorage.cs
```diff
@@ -119,11 +119,11 @@ private static Span<byte> GetHashBasedStoragePath(Span<byte> pathSpan, in ValueH
 
         if (Scheme == INodeStorage.KeyScheme.HalfPath && (readFlags & ReadFlags.HintReadAhead) != 0)
         {
-            if (address is null && path.Length > TopStateBoundary)
+            if (address == default && path.Length > TopStateBoundary)
```
LukaszRozmej, line 122: as comparing to default is now expensive, maybe lets save that to variable and avoid 2nd comparison in else

LukaszRozmej (commented):
We could change `HashesMessage` to take list of ValueHash256, not sure if it would be beneficial?

Not sure if we won't break any logic? Addres.Zero and Hash.Zero can have some special meaning where we used null's before?

PR #7386, opened 2024-09-03
Refactor `TxValidator`

emlautarom1:
Partially solves #7313

## Changes

- Refactor `TxValidator` to dispatch transaction validation to appropriate sub validators.

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

All unit tests are passing.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks

Please see the existing discussion.


src/Nethermind/Nethermind.Blockchain.Test/Validators/TestTransactionValidator.cs
```diff
@@ -1,40 +0,0 @@
...
-
-using System.Collections.Generic;
-using System.Diagnostics.CodeAnalysis;
-using Nethermind.Core;
-using Nethermind.Core.Specs;
-using Nethermind.TxPool;
-
-namespace Nethermind.Blockchain.Test.Validators
-{
-    public class TestTxValidator : ITxValidator
```
emlautarom1, line 12: Unused class

src/Nethermind/Nethermind.Optimism/OptimismTxValidator.cs
```diff
@@ -8,21 +8,11 @@
 
 namespace Nethermind.Optimism;
 
-public class OptimismTxValidator : ITxValidator
+public sealed class OptimismTxValidator : ITxValidator
```
emlautarom1, line 11: The existing `OptimismTxValidator` always validates transactions that are of `TxType.Deposit` without performing any validation at all. This does not look correct since there are validations that are performed on all existing `TxType`s like `IntrinsicGasTxValidator` but I might be missing some context.

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
ak88, line 1: Probably not worth it, but if all the `WellFormed`  methods was made static we could avoid a lot of virtual calls.

src/Nethermind/Nethermind.Optimism/OptimismTxValidator.cs
ak88: maybe @jmederosalvarado knows more about this?

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
```diff
@@ -2,294 +2,450 @@
...
+                new ExpectedChainIdTxValidator(chainId),
+                GasFieldsTxValidator.Instance,
+                ContractSizeTxValidator.Instance,
+                BlobFieldsTxValidator.Instance,
+                MempoolBlobTxValidator.Instance
+            ])
+        },
+    };
+
+    public TxValidator AddValidator(TxType type, ITxValidator validator)
```
ak88, line 65:
```suggestion
    public TxValidator AddOrReplaceValidator(TxType type, ITxValidator validator)
```

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
```diff
@@ -2,294 +2,450 @@
...
 using Nethermind.Int256;
-using Nethermind.TxPool;
 
-namespace Nethermind.Consensus.Validators
+namespace Nethermind.Consensus.Validators;
+
+public sealed class TxValidator(ulong chainId) : ITxValidator
 {
-    public class TxValidator : ITxValidator
+    private readonly Dictionary<TxType, ITxValidator> _validators = new()
```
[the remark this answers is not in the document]
LukaszRozmej: No we can't:
1. There is no guarantee it is continuous
2. There might be plugin TxType integer values, that are outside of pre-defined in TxType enum

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
LukaszRozmej: SetValidator?

emlautarom1: I initially went with `WithValidator` so the call site would look like:

```c#
_api.TxValidator
    .WithValidator(TxType.DepositTx, new OptimismTxValidator())
    .WithValidator(...)
```

As long as we're consistent with other classes with similar behavior any name is fine.

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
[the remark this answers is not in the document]
emlautarom1: We could use a fixed 256 item array since `TxType` is a `byte`, thus accessing the appropriate validator would be an array access.

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
emlautarom1: Could you elaborate a bit on how this would look? For example, if we make `IntrinsicGasTxValidator.IsWellFormed` static then we don't comply with the `ITxValidator` interface.

smartprogrammer93 (approved):

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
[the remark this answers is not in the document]
LukaszRozmej: 1. We can go with an array.
2. We need some centralized repository to register TxTypes. So you would register all required implementations (like TxDecoder, TxValidator, TxProcesor ect.) for tx type in one go

src/Nethermind/Nethermind.Optimism/OptimismTxValidator.cs
LukaszRozmej: Deposit transactions are special types, I think there should only be one per block on first step. Ping @jmederosalvarado , @deffrian or @flcl42 about it.

src/Nethermind/Nethermind.TxPool/ITxValidator.cs
```diff
@@ -9,6 +9,7 @@ namespace Nethermind.TxPool
 {
     public interface ITxValidator
     {
+        bool IsWellFormed(Transaction transaction, IReleaseSpec releaseSpec) => IsWellFormed(transaction, releaseSpec, out _);
```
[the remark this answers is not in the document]
LukaszRozmej: BTW it is unused - remove?

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
```diff
@@ -2,294 +2,450 @@
...
+
+    /// <summary>
+    /// Full and correct validation is only possible in the context of a specific block
+    /// as we cannot generalize correctness of the transaction without knowing the EIPs implemented
+    /// and the world state(account nonce in particular).
+    /// Even without protocol change, the tx can become invalid if another tx
+    /// from the same account with the same nonce got included on the chain.
+    /// As such, we can decide whether tx is well formed as long as we also validate nonce
+    /// just before the execution of the block / tx.
+    /// </summary>
```
LukaszRozmej, line 81: This is more of `<remarks>` than `<summary>`

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
```diff
@@ -2,294 +2,450 @@
...
+    public bool IsWellFormed(Transaction transaction, IReleaseSpec releaseSpec, out string? error)
+    {
+        if (!_validators.TryGetValue(transaction.Type, out ITxValidator? validator))
```
LukaszRozmej, line 84: remove nullability from `out ITxValidator? validator`

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
```diff
@@ -2,294 +2,450 @@
...
+            return false;
         }
 
-        private static bool Validate3860Rules(Transaction transaction, IReleaseSpec releaseSpec) =>
-            !transaction.IsAboveInitCode(releaseSpec);
+        return validator.IsWellFormed(transaction, releaseSpec, out error);
+    }
+}
+
+public sealed class AllTxValidator(List<ITxValidator> validators) : ITxValidator
```
LukaszRozmej, line 94: `CompositeTxValidator` (we use the term `Composite` in multiple places already for this pattern)

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
```diff
@@ -2,294 +2,450 @@
...
+        return true;
+    }
+}
+
+public sealed class IntrinsicGasTxValidator : ITxValidator
+{
+    public static readonly IntrinsicGasTxValidator Instance = new();
+    private IntrinsicGasTxValidator() { }
+
+    public bool IsWellFormed(Transaction transaction, IReleaseSpec releaseSpec, [NotNullWhen(false)] out string? error)
```
LukaszRozmej, line 117: if we made error a `ref` we could simplify the code a bit  - as it was originally

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
```diff
@@ -2,294 +2,450 @@
...
+            TxType.Legacy, new AllTxValidator([
+                IntrinsicGasTxValidator.Instance,
+                new LegacySignatureTxValidator(chainId),
+                ContractSizeTxValidator.Instance,
+                NonBlobFieldsTxValidator.Instance,
+            ])
+        },
+        {
+            TxType.AccessList, new AllTxValidator([
+                new ReleaseSpecTxValidator(spec => spec.IsEip2930Enabled),
```
LukaszRozmej, line 32:
```suggestion
                new ReleaseSpecTxValidator(static spec => spec.IsEip2930Enabled),
```

Add to all registries.

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
```diff
@@ -2,294 +2,450 @@
...
+        if (AdditionalCheck(transaction, releaseSpec))
+        {
             return true;
         }
+
+        if (releaseSpec.ValidateChainId)
+        {
+            error = TxErrorMessages.InvalidTxSignature;
+            return false;
+        }
```
LukaszRozmej, line 432: Instead of `AdditionalCheck` have a `virtual ValidateChainId` method, that you will override in `LegacySignatureTxValidator`

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
```diff
@@ -2,294 +2,450 @@
...
+public sealed class GasFieldsTxValidator : ITxValidator
+{
+    public static readonly GasFieldsTxValidator Instance = new();
+    private GasFieldsTxValidator() { }
+
+    public bool IsWellFormed(Transaction transaction, IReleaseSpec releaseSpec, [NotNullWhen(false)] out string? error)
+    {
+        error = null;
+
+        if (!releaseSpec.IsEip1559Enabled)
```
LukaszRozmej, line 174: I wonder if this is now needed if we guard that in `ReleaseSpecTxValidator`?

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
```diff
@@ -2,294 +2,450 @@
...
+                SignatureTxValidator.Instance,
+                new ExpectedChainIdTxValidator(chainId),
+                GasFieldsTxValidator.Instance,
+                ContractSizeTxValidator.Instance,
+                NonBlobFieldsTxValidator.Instance,
+            ])
+        },
+        {
+            TxType.Blob, new AllTxValidator([
+                new ReleaseSpecTxValidator(spec => spec.IsEip4844Enabled),
```
LukaszRozmej, line 53: Not sure but maybe:
```suggestion
                new ReleaseSpecTxValidator(spec => spec.IsEip4844Enabled && spec.IsEip1559Enabled),
```

LukaszRozmej (commented):
Overall I like it, just some details

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
LukaszRozmej: Lets go with either `WithValidator` or `SetValidator`

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
emlautarom1: I went with `AllTxValidator` since you could have a validator that checks that at least one validator returns `true`, that is, merge validations using `||` instead of `&&`, but I guess we don't have a use case for that.
Renamed to `CompositeTxValidator` for consistency.

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
emlautarom1: Renamed `AdditionalCheck` to `ValidateChainId`

src/Nethermind/Nethermind.TxPool/ITxValidator.cs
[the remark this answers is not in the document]
emlautarom1: Indeed, seems like the method was used allways in the concrete `TxDecoder`. Removed.

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
emlautarom1: This change would break existing behavior. I've added a test that passed on the original code and in this refactor but not with this change.

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
emlautarom1: As mentioned, we can't guard for this without changing the existing behavior.

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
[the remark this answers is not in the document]
emlautarom1: Replaced `Dictionary<TxType, ITxDecoder>` with `ITxDecoder[]`

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
emlautarom1: I don't see how we could apply this change without changing the `ITxValidator` interface.
Could you elaborate a bit on how the code would look?

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
```diff
@@ -2,294 +2,446 @@
...
+
+public sealed class TxValidator : ITxValidator
 {
-    public class TxValidator : ITxValidator
+    private readonly ITxValidator[] _validators;
+
+    public TxValidator(ulong chainId)
     {
-        private readonly ulong _chainIdValue;
+        _validators = new ITxValidator[byte.MaxValue + 1];
```
LukaszRozmej, line 24: why +1?

src/Nethermind/Nethermind.Optimism/OptimismTxValidator.cs
jmederosalvarado: DepositTxs are mostly validated on L1 smart contract, and later by the CL client. Whatever DepositTx the CL feeds to the EL client is already valid.

src/Nethermind/Nethermind.Api/NethermindApi.cs
```diff
@@ -208,7 +208,7 @@ public ISealEngine SealEngine
         public ITxPoolInfoProvider? TxPoolInfoProvider { get; set; }
         public IHealthHintService? HealthHintService { get; set; }
         public IRpcCapabilitiesProvider? RpcCapabilitiesProvider { get; set; }
-        public ITxValidator? TxValidator { get; set; }
+        public TxValidator? TxValidator { get; set; }
```
jmederosalvarado, line 211: I would like this to remain an interface. Any reason not to?

src/Nethermind/Nethermind.Api/IApiWithBlockchain.cs
```diff
@@ -74,7 +74,7 @@ public interface IApiWithBlockchain : IApiWithStores, IBlockchainBridgeFactory
         IHealthHintService? HealthHintService { get; set; }
         IRpcCapabilitiesProvider? RpcCapabilitiesProvider { get; set; }
         ITransactionComparerProvider? TransactionComparerProvider { get; set; }
-        ITxValidator? TxValidator { get; set; }
+        TxValidator? TxValidator { get; set; }
```
jmederosalvarado, line 77: I would like this to remain an interface. Any reason not to?

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
```diff
@@ -2,294 +2,446 @@
...
+            new ExpectedChainIdTxValidator(chainId),
+            GasFieldsTxValidator.Instance,
+            ContractSizeTxValidator.Instance,
+            BlobFieldsTxValidator.Instance,
+            MempoolBlobTxValidator.Instance
+        ]);
+    }
 
-        public TxValidator(ulong chainId)
+    public TxValidator WithValidator(TxType type, ITxValidator validator)
```
jmederosalvarado, line 60: Having something like `WithValidator` mutate the object itself looks weird. I'm not sure if this is how we do it in other parts of the code. I know we do it for object builders (e.g. `TransactionBuilder`) but not for the object itself

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
jmederosalvarado, line 24: Shouldn't we use the `Transaction.MaxTxType` introduced [here](https://github.com/NethermindEth/nethermind/pull/7334/files#diff-52dd146934935c1b5aa5650c8c5842ce450083c4f03ae076c0c1550745eeb750R22)

LukaszRozmej: Yeah we should

LukaszRozmej: Did it in my PR

src/Nethermind/Nethermind.Api/NethermindApi.cs
emlautarom1: I had to change it to a concrete class since during initialization we might want to register validators for new `TxType`s:

https://github.com/NethermindEth/nethermind/blob/9437715fd3590891aa718891686b8aa5178a3d10/src/Nethermind/Nethermind.Optimism/InitializeBlockchainOptimism.cs#L77

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
emlautarom1: I would implement `WithValidator` as a clone, that is, return a new `TxValidator` with the added validator, but I thought that it made little sense to pay for that due to our use case. Do you have any specific suggestion?

src/Nethermind/Nethermind.Api/IApiWithBlockchain.cs
emlautarom1: Required due to the change on `NethermindApi`

src/Nethermind/Nethermind.Specs/ReleaseSpec.cs
```diff
@@ -55,7 +55,12 @@ public class ReleaseSpec : IReleaseSpec
         // used only in testing
         public ReleaseSpec Clone() => (ReleaseSpec)MemberwiseClone();
 
-        public bool IsEip1559Enabled { get; set; }
+        public bool IsEip1559Enabled
+        {
+            get => _isEip1559Enabled || IsEip4844Enabled;
+            set => _isEip1559Enabled = value;
+        }
```
emlautarom1, line 62: NOTE: we might want to revisit this later since EIPs can require other EIPs.

This approach is kind of fragile, for example, EIP 4844 requires 2930 but we're not overriding `IsEip2930Enabled` like we're doing with `IsEip1559Enabled`.

LukaszRozmej: Yeah to be honest we should make sure that our release spec is following the spec, so have those cascades.

src/Nethermind/Nethermind.Core/ValidationResult.cs
```diff
@@ -0,0 +1,13 @@
...
+namespace Nethermind.Core;
+
+public record struct ValidationResult(string? Error)
+{
+    public static ValidationResult Success => new(null);
+    public static implicit operator bool(ValidationResult result) => result.AsBool();
+    public static implicit operator ValidationResult(string error) => new(error);
+    public override string ToString() => Error ?? "Success";
+    public bool AsBool() => Error is null;
+}
```
emlautarom1, line 13: NOTE: We have several `*Result` types that could be replaced with a generic `Result<E, T>`

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
emlautarom1: We ended up going with a `ValidatorResult` return value rather than the `TryX` approach of returning a `bool` with an associated `out` parameter.

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
emlautarom1: Resolved in the related comment.

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
emlautarom1: These tests showed that our `ReleaseSpec` was not setting up certain invariants. Changed so that now `EIP1559` is implied by `EIP4844`.

src/Nethermind/Nethermind.Core/ValidationResult.cs
LukaszRozmej: Maybe but not sure, here we are using syntactic sugar with implicit conversion to bool, which allows us to treat it as a boolean value. Not sure if we can leverage that with generics easily and if it would be worth it - would have to see it.

jmederosalvarado (approved):

PR #7406, opened 2024-09-09
Only write to tree net changes at end of block

benaadams:
## Changes

- Often the net changes for state and storage across all the tx in the block is nothing; so skip writing to db when this is the case
- Process storage trees in parallel for whole block

image

- [x] Optimization

#### Requires testing

- [x] No - covered by existing tests


src/Nethermind/Nethermind.State.Test/StorageProviderTests.cs
```diff
@@ -185,20 +185,17 @@ public void Commit_trees_clear_caches_get_previous_root()
...
-        ctx.StateProvider.Commit(Frontier.Instance);
-        ctx.StateProvider.CommitTree(0);
+        storageProvider.Commit(Frontier.Instance);
+        storageProvider.CommitTree(0);
 
         // block 2
         Hash256 stateRoot = ctx.StateProvider.StateRoot;
         storageProvider.Set(new StorageCell(ctx.Address1, 1), _values[2]);
         storageProvider.Commit(Frontier.Instance);
-        ctx.StateProvider.Commit(Frontier.Instance);
```
benaadams, line 196: `storageProvider` and `ctx.StateProvider` are the same, so using both is confusing

src/Nethermind/Nethermind.State.Test/StorageProviderTests.cs
```diff
@@ -185,20 +185,17 @@ public void Commit_trees_clear_caches_get_previous_root()
         WorldState storageProvider = BuildStorageProvider(ctx);
         storageProvider.Set(new StorageCell(ctx.Address1, 1), _values[1]);
         storageProvider.Commit(Frontier.Instance);
-
-        ctx.StateProvider.Commit(Frontier.Instance);
-        ctx.StateProvider.CommitTree(0);
+        storageProvider.Commit(Frontier.Instance);
```
benaadams, line 188: Although this is a double commit, so shouldn't do anything; it did help find a bug, so keeping

LukaszRozmej:
Does it increase throughput?

benaadams:
> Does it increase throughput?

Yes, a lot of mev front-run + tx + back-run end up with some net zero changes across the block but we merklize the changes when considering each tx individually. Also allows us to parallelize all the storage merklization for the block rather than doing it sequentially for each tx

benaadams:
> > Does it increase throughput?
> 
> Yes, a lot of mev front-run + tx + back-run end up with some net zero changes across the block but we merklize the changes when considering each tx individually.
> 
> Also allows us to parallelize all the storage merklization for the block rather than doing it sequentially for each tx

e.g. heavy SSTORE blocks have lower impact
image

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
```diff
@@ -240,7 +235,7 @@ protected override void CommitStorageRoots()
         }
 
         // Is overhead of parallel foreach worth it?
-        if (_toUpdateRoots.Count <= 4)
+        if (_toUpdateRoots.Count < 3)
```
Scooletz, line 238: This magic number is killing me 😢

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
```diff
@@ -213,6 +207,7 @@ protected override void CommitCore(IStorageTracer tracer)
             if (_stateProvider.AccountExists(address))
             {
                 _toUpdateRoots.Add(address);
+                GetOrCreateStorage(address, out _);
```
Scooletz, line 210: Why is it needed to create one? Comment please or fix.

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
```diff
@@ -260,8 +257,34 @@ void UpdateRootHashesSingleThread()
...
 
                 StorageTree storageTree = kvp.Value;
-                storageTree.UpdateRootHash(canBeParallel: true);
-                _stateProvider.UpdateStorageRoot(address: kvp.Key, storageTree.RootHash);
+                ref var dict = ref CollectionsMarshal.GetValueRefOrNullRef(_blockChanges, kvp.Key);
+                int writes = 0;
+                int skipped = 0;
+                foreach (var key in dict.Keys)
+                {
+                    ref var change = ref dict.GetValueRefOrNullRef(key);
```
Scooletz, line 265: Why iterate over keys only then to lookup the value `change`?

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
```diff
@@ -283,7 +306,35 @@ void UpdateRootHashesMultiThread()
                     return state;
                 }
                 StorageTree storageTree = kvp.Value;
-                storageTree.UpdateRootHash(canBeParallel: false);
+                ref var dict = ref CollectionsMarshal.GetValueRefOrNullRef(state.changes, kvp.Key);
```
Scooletz, line 309: This chunk is a direct copy of the single threaded. Could we extract a method maybe?

src/Nethermind/Nethermind.State/StateProvider.cs
```diff
@@ -579,6 +583,36 @@ public void Commit(IReleaseSpec releaseSpec, IWorldStateTracer stateTracer, bool
...
+            {
+                FlushToTree();
+            }
+        }
+
+        private void FlushToTree()
+        {
+            int writes = 0;
+            int skipped = 0;
+            foreach (var key in _blockChanges.Keys)
```
Scooletz, line 597: Similar as above, why loop over to then get the value manually? I don't believe it's cheaper, the lookup hashes the key while the iterator over kvp just scans, right?

src/Nethermind/Nethermind.State/StateProvider.cs
```diff
@@ -579,6 +583,36 @@ public void Commit(IReleaseSpec releaseSpec, IWorldStateTracer stateTracer, bool
...
+                    _tree.Set(key, change.After);
+                    writes++;
+                }
+                else
+                {
+                    skipped++;
+                }
+            }
+
+            if (writes > 0)
```
Scooletz, line 612: A repeated chunk. I'd extract `ReportWrites(writes, skipped)`

src/Nethermind/Nethermind.State/IWorldState.cs
```diff
@@ -108,9 +108,9 @@ public interface IWorldState : IJournal<Snapshot>, IReadOnlyStateProvider
 
     /* snapshots */
 
-    void Commit(IReleaseSpec releaseSpec, bool isGenesis = false, bool commitStorageRoots = true);
+    void Commit(IReleaseSpec releaseSpec, bool isGenesis = false, bool commitRoots = true);
 
-    void Commit(IReleaseSpec releaseSpec, IWorldStateTracer? tracer, bool isGenesis = false, bool commitStorageRoots = true);
+    void Commit(IReleaseSpec releaseSpec, IWorldStateTracer? tracer, bool isGenesis = false, bool commitRoots = true);
```
Scooletz, line 113: I think this is not the first change in the `IWorldState.Commit`. Should we consider flags maybe for the options, or do you think that `bool commitRoots` will stick for a while and are the only toggle that will be passed down?

LukaszRozmej (commented):
Looks a bit risky to be honest

src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.cs
```diff
@@ -341,13 +341,13 @@ protected virtual TxReceipt[] ProcessBlock(
...
         // Eip158Enabled=false, so we end up persisting empty accounts created while processing withdrawals.
-        _stateProvider.Commit(spec);
+        _stateProvider.Commit(spec, commitRoots: false);
 
         _executionRequestsProcessor.ProcessExecutionRequests(block, _stateProvider, receipts, spec);
 
         ReceiptsTracer.EndBlockTrace();
 
-        _stateProvider.Commit(spec, commitStorageRoots: true);
+        _stateProvider.Commit(spec, commitRoots: true);
```
Scooletz, line 350: Am I getting it right that after these, right now, when we access the root hash will be wrong until the final `commitRoots: true` is called, right? I don't see it as a problem in any execution path, but maybe it is?

Scooletz (changes requested):
Remarks provided.
The description of this PR states _Often the net changes for state and storage across all the tx in the block is nothing; so skip writing to db when this is the case_, but is it only the case? Isn't it amortizing writes in general so that if a given storage cell is updated 5 times, only the last one will be written back?

src/Nethermind/Nethermind.Db/Metrics.cs
```diff
@@ -2,9 +2,15 @@
...
 using System.ComponentModel;
+using System.Runtime.CompilerServices;
+using System.Threading;
 using Nethermind.Core.Attributes;
 using Nethermind.Core.Threading;
 
+[assembly: InternalsVisibleTo("Nethermind.Consensus")]
+[assembly: InternalsVisibleTo("Nethermind.State")]
+[assembly: InternalsVisibleTo("Nethermind.Evm")]
+[assembly: InternalsVisibleTo("Nethermind.TxPool")]
```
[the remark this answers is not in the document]
benaadams: Mostly is trying to shut them down from being sent to grafana

src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.cs
benaadams: Will be currently anyway unless pre-Eip658; doesn't change behaviour in this PR either; if Eip658 not enabled then it will commit the roots at end of tx

benaadams:
> Isn't it amortizing writes in general so that if a given storage cell is updated 5 times, only the last one will be written back?

This is sorta the case already at the tx level; the change here is moving that to the block level

benaadams:
> Isn't it amortizing writes in general so that if a given storage cell is updated 5 times, only the last one will be written back?

This is sorta the case already at the tx level; the change here is moving that to the block level.

Also fits with delayed root calculation if that becomes a thing (gathering all the root calculations to the end)

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
benaadams: < 3 == ❤️ 

Has to be written out in full 😉

benaadams:
> Looks a bit risky to be honest

Ignore the first commit; which is a bit noisy and refactoring confusing parameter names, dead methods etc

The look at the following two in isolation and isn't that big a change

asdacap:
Screenshot_2025-03-24_22-53-12

10% improvement on average mgas/sec. But tend to hangs randomly causing random drop in block/sec. Any idea why?

asdacap:
Screenshot_2025-03-24_23-28-50

Dip correlate with lock contention metric.

asdacap:
Screenshot_2025-03-24_23-31-44

In general it spawn much mork thread pool work item.

benaadams:
> In general it spawn much mork thread pool work item.

Latest change better?

src/Nethermind/Nethermind.State/StateProvider.cs
benaadams: Done

src/Nethermind/Nethermind.State/StateProvider.cs
benaadams: Changed

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
benaadams: Changed

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
benaadams: Chaned

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
benaadams: It can't create them on demand in the parallel loop as the storage dictionary is not concurrent

benaadams: Added comment

src/Nethermind/Nethermind.Core/Account.cs
```diff
@@ -82,6 +83,27 @@ private Account(Account account, in UInt256 nonce, in UInt256 balance)
...
+        public override bool Equals(object? obj) => Equals(obj as Account);
+        public static bool operator ==(Account? left, Account? right)
+        {
+            if (left is not null)
+            {
+                return left.Equals(right);
+            }
+            return right is null;
+        }
+        public override int GetHashCode() => throw new NotImplementedException();
```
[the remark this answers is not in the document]
benaadams: Added

benaadams:
> Seems to work, but lose about 50mgas/sec in the process.

🥲

So keep the parallel, but maybe limit to the min(physical, 16) rather than logical 🤔

src/Nethermind/Nethermind.Db/Metrics.cs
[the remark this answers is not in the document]
benaadams: Ones in the other projects are mostly incrementing cache hit metrics for a metric that has corresponding Db metric

src/Nethermind/Nethermind.Trie/Pruning/TinyTreePath.cs
```diff
@@ -43,7 +44,8 @@ public TreePath ToTreePath()
     public bool Equals(TinyTreePath other) => _data == other._data;
     public bool Equals(in TinyTreePath other) => _data == other._data;
     public override bool Equals(object? obj) => obj is TinyTreePath other && Equals(other);
-    public override int GetHashCode() => _data.GetHashCode();
+    // Need more variance than straightforward long.GetHashCode() as it determines lock contention in ConcurrentDictionary
+    public override int GetHashCode() => (int)BitOperations.Crc32C((uint)_data, ((ulong)_data & 0xffff_ffff_0000_0000) | ~(uint)_data);
```
benaadams, line 48: Reduced the contention in a couple ways, though this one is of probably most interesting to @asdacap

src/Nethermind/Nethermind.State.Test/StorageProviderTests.cs
Scooletz: The difference is in the order. CommitTree applies to the tree, so if you commit first, you push the changes in the tree to db and later apply it. Usually it's the other way around.

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
```diff
@@ -512,36 +576,41 @@ public StorageTree Create(Address address, IScopedTrieStore trieStore, Hash256 s
...
-            ref TValue value = ref CollectionsMarshal.GetValueRefOrAddDefault(_dictionary, storageCellIndex, out exists);
+            ref ChangeTrace value = ref CollectionsMarshal.GetValueRefOrAddDefault(_dictionary, storageCellIndex, out exists);
             if (!exists && _missingAreDefault)
             {
                 // Where we know the rest of the tree is empty
                 // we can say the value was found but is default
                 // rather than having to check the database
-                value = defaultValue;
+                value = ChangeTrace.ZeroBytes;
                 exists = true;
```
Scooletz, line 599: Why then return `exists` in any case? I do understand that in `CollectionsMarshal` it's needed to create the default but here, we always set everything and exists is `true`. I'd remove it even though it won't be aligned with the `CollectionsMarshal`

src/Nethermind/Nethermind.State/StateProvider.cs
```diff
@@ -579,6 +583,36 @@ public void Commit(IReleaseSpec releaseSpec, IWorldStateTracer stateTracer, bool
...
+        private void FlushToTree()
+        {
+            int writes = 0;
+            int skipped = 0;
+            foreach (var key in _blockChanges.Keys)
+            {
+                ref var change = ref CollectionsMarshal.GetValueRefOrNullRef(_blockChanges, key);
+                if (change.Before != change.After)
+                {
+                    change.Before = change.After;
```
Scooletz, line 602: So this is the mutability.

src/Nethermind/Nethermind.Trie/TrieNode.Decoder.cs
```diff
@@ -22,7 +22,7 @@ namespace Nethermind.Trie
 {
     public partial class TrieNode
     {
-        private const int StackallocByteThreshold = 256;
+        private const int StackallocByteThreshold = 384;
```
Scooletz, line 25: This magic number requires a comment.

Scooletz (approved):
Minor comments, one magic number to clarify if possible.

src/Nethermind/Nethermind.State.Test/StorageProviderTests.cs
benaadams: 🤔 Commit doesn't apply to db; moves the "checkpoint" so how far a restore can go, then CommitTree applies to db

src/Nethermind/Nethermind.State/IWorldState.cs
benaadams: Hopefully last change 😉

Mostly just renamed param, as also now effects stateroot

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
benaadams: Only adds on `_missingAreDefault` which is after selfdestruct or creation of a new storage trie at vanilla address (that we know is empty)

PR #7407, opened 2024-09-10
Reduce memory use during Archive Sync

benaadams:
## Changes

- TxTries and ReceiptTries don't do any pruning so don't need the ConcurrentDictionaries allocated just for Calculating roots

image

- Lazy allocate ConcurrenctDictionaries (+DirtyNodes in Trie)
- Change class => struct
- NodeKey class => struct
- Force GCs on archive sync

Seen when processing archive nodes (many blocks at once)

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
+++ b/src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
@@ -286,7 +286,7 @@ namespace Nethermind.Trie.Pruning
 
         // Track ALL of the recently re-committed persisted nodes. This is so that we don't accidentally remove
         // recommitted persisted nodes (which will not get re-persisted).
-        private readonly ConcurrentDictionary<HashAndTinyPathAndHash, long> _persistedLastSeen = new(CollectionExtensions.LockPartitions, 4 * 4096);
+        private readonly ConcurrentDictionary<HashAndTinyPathAndHash, long>? _persistedLastSeen;
 
         private bool _lastPersistedReachedReorgBoundary;
         private Task _pruningTask = Task.CompletedTask;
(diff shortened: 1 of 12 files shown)


MarekM25 (approved):
LGTM! LastSeen is not needed at all for receipts and traces. :)

Was it the reason for the memory leak?

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -286,7 +286,7 @@ public readonly void Dispose()
 
         // Track ALL of the recently re-committed persisted nodes. This is so that we don't accidentally remove
         // recommitted persisted nodes (which will not get re-persisted).
-        private readonly ConcurrentDictionary<HashAndTinyPathAndHash, long> _persistedLastSeen = new(CollectionExtensions.LockPartitions, 4 * 4096);
+        private readonly ConcurrentDictionary<HashAndTinyPathAndHash, long>? _persistedLastSeen;
```
LukaszRozmej, line 289: We don't have to touch any other usages?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
benaadams: Lazily allocated a few more (like the DirtyTrieCache isn't needed etc)

Changed Change and NodeKey to structs

benaadams:
> Was it the reason for the memory leak?

Yes is doing 400 blk/s so doing the block allocations at a far faster rate; plus since its not going via newPayload it doesn't trigger the GCKeeper so we don't trigger the memory release GCs (GC decides we need that memory due to the high allocation rate) or the MallocTrims

Issue #7392, opened 2024-09-04
Very big memory usage during sync of archival node (any chain)


kamilchodola:
Noticed on all supported chains that during archive sync we are consuming way too much memory,

Mainnet archive OOMed on 5mln blocks
Gnosis archive which runs very strong machine have usage of 110GB
Volta and Energyweb are having 20GB of usage after 2-3h of sync

benaadams:
Resolved by https://github.com/NethermindEth/nethermind/pull/7407

PR #7413, opened 2024-09-11
Use non-allocating Int64 key generation

benaadams:
Contributes to #7392

## Changes

- Create span over the Int64 rather than allocating byte[]

image

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Core/Extensions/Int64Extensions.cs
+++ b/src/Nethermind/Nethermind.Core/Extensions/Int64Extensions.cs
@@ -3,7 +3,9 @@
 
 using System;
 using System.Buffers.Binary;
+using System.Numerics;
 using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
 
 using Nethermind.Int256;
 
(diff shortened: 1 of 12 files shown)


src/Nethermind/Nethermind.Core/Extensions/Int64Extensions.cs
```diff
@@ -3,120 +3,32 @@
...
+using System.Runtime.InteropServices;
 
 using Nethermind.Int256;
 
 namespace Nethermind.Core.Extensions;
 
 public static class Int64Extensions
 {
-    public static byte[] ToBigEndianByteArrayWithoutLeadingZeros(this long value)
+    public static ReadOnlySpan<byte> MutateToBigEndianSpanWithoutLeadingZeros(ref this long value)
```
LukaszRozmej, line 16: WHy does it mutate the original value? Is that safe?

benaadams: Its needs to return the Span over valid stackspace so it takes the long by ref and creates the span over it; so its valid

The wrinkle is it also needs to change its byte order from little endian to big endian; trashing the local variable on the other side. 

However there are only two instances where that's important; one I moved this call down one so is last use; and other I take a copy to a new local so the rest of the method can keep using the original value

LukaszRozmej: wouldn't it be better to always stackalloc 8 bytes?

benaadams: Can't return the stackalloc'd to the caller as it will have been unwound so no longer valid. Caller needs to allocate the space

benaadams: I could through in an out var and then mutate that instead so
```csharp
ReadOnlySpan<byte> ToBigEndianSpanWithoutLeadingZeros(this long value, out long bigEndian)
```

Then in use it would be:
```
key.MutateToBigEndianSpanWithoutLeadingZeros(out _)
```
but that's pretty horrible too

LukaszRozmej: I like this one a lot better - not to cut yourself. Looks good enough to me.

LukaszRozmej (approved):

src/Nethermind/Nethermind.Core/Extensions/Int64Extensions.cs
benaadams: Isn't a T constraint for `ReverseEndianness`; which does seem a missed opportunity for the runtime

PR #7419, opened 2024-09-12
Use Reth's method of calculating MGas/s

benaadams:
## Changes

- Measure Evm transaction processing time, rather than block end to end
- Is now "Evm throughput" in logs

image

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Processing/IBlockProcessor.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/IBlockProcessor.cs
@@ -43,6 +43,16 @@ namespace Nethermind.Consensus.Processing
         event EventHandler<BlockProcessedEventArgs> BlockProcessed;
 
         /// <summary>
+        /// Fired when a block's transactions are been processed.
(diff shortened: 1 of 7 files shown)


LukaszRozmej:
I would keep in logs old way of calculating MGas/s, we can output new metrics for grafana with the Reth way.

src/Nethermind/Nethermind.Consensus/Processing/IBlockProcessor.cs
```diff
@@ -42,6 +42,16 @@ Block[] Process(
...
+        /// <summary>
+        /// Fired when a block's transactions are been processed.
+        /// </summary>
+        event EventHandler<Block> EvmProcessingStarted;
+
+        /// <summary>
+        /// Fired after a block's transactions have been processed.
+        /// </summary>
+        event EventHandler<Block> EvmProcessingComplete;
+
```
LukaszRozmej, line 54: Meh, hate it to be honest. Lets just directly set a metric.

PR #7423, opened 2024-09-12
Prewarm tx addresses in parallel

benaadams:
## Changes

- Prewam addresses only; in parallel to prewarming full txs (Sender+To)
  - If prewarming is slowed by actually processing the tx, the Accounts are still being preloaded
- Include beacon state root warming as part of this

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain/BeaconBlockRoot/BeaconBlockRootHandler.cs
+++ b/src/Nethermind/Nethermind.Blockchain/BeaconBlockRoot/BeaconBlockRootHandler.cs
@@ -15,6 +15,15 @@ public class BeaconBlockRootHandler(ITransactionProcessor processor) : IBeaconBl
 {
     private const long GasLimit = 30_000_000L;
 
+    public Address? BeaconRootsAddress(Block block, IReleaseSpec spec)
+    {
+        BlockHeader? header = block.Header;
+        var canInsertBeaconRoot = spec.IsBeaconBlockRootAvailable
+                                  && !header.IsGenesis
+                                  && header.ParentBeaconBlockRoot is not null;
+        return !canInsertBeaconRoot ? null : spec.Eip4788ContractAddress ?? Eip4788Constants.BeaconRootsAddress;
+    }
+
     public void StoreBeaconRoot(Block block, IReleaseSpec spec)
     {
         BlockHeader? header = block.Header;
--- a/src/Nethermind/Nethermind.Blockchain/BeaconBlockRoot/BeaconBlockRootHandler.cs
+++ b/src/Nethermind/Nethermind.Blockchain/BeaconBlockRoot/BeaconBlockRootHandler.cs
@@ -15,7 +15,7 @@ public class BeaconBlockRootHandler(ITransactionProcessor processor) : IBeaconBl
 {
     private const long GasLimit = 30_000_000L;
 
-    public (Address? toAddress, AccessList? accessList) BeaconRootsAccessList(Block block, IReleaseSpec spec)
+    public (Address? toAddress, AccessList? accessList) BeaconRootsAccessList(Block block, IReleaseSpec spec, bool includeStorageCells = true)
     {
         BlockHeader? header = block.Header;
         bool canInsertBeaconRoot = spec.IsBeaconBlockRootAvailable
@@ -26,18 +26,25 @@ public class BeaconBlockRootHandler(ITransactionProcessor processor) : IBeaconBl
             spec.Eip4788ContractAddress ?? Eip4788Constants.BeaconRootsAddress :
             null;
 
-        return (eip4788ContractAddress,
-            eip4788ContractAddress is null ?
-                null :
-                new AccessList.Builder()
-                    .AddAddress(eip4788ContractAddress)
-                    .AddStorage(block.Timestamp % 8191)
-                    .Build());
+        if (eip4788ContractAddress is null)
+        {
+            return (null, null);
+        }
+
+        var builder = new AccessList.Builder()
+            .AddAddress(eip4788ContractAddress);
+
+        if (includeStorageCells)
+        {
+            builder.AddStorage(block.Timestamp % 8191);
+        }
+
+        return (eip4788ContractAddress, builder.Build());
     }
 
     public void StoreBeaconRoot(Block block, IReleaseSpec spec)
     {
-        (Address? toAddress, AccessList? accessList) = BeaconRootsAccessList(block, spec);
+        (Address? toAddress, AccessList? accessList) = BeaconRootsAccessList(block, spec, includeStorageCells: false);
 
         if (toAddress is not null)
         {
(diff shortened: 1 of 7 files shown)


src/Nethermind/Nethermind.Blockchain/BeaconBlockRoot/BeaconBlockRootHandler.cs
```diff
@@ -15,6 +15,15 @@ public class BeaconBlockRootHandler(ITransactionProcessor processor) : IBeaconBl
 {
     private const long GasLimit = 30_000_000L;
 
+    public Address? BeaconRootsAddress(Block block, IReleaseSpec spec)
```
LukaszRozmej, line 18: Code duplication with `StoreBeaconRoot`, please unify.

src/Nethermind/Nethermind.Blockchain/BeaconBlockRoot/IBeaconBlockRootHandler.cs
```diff
@@ -7,5 +7,6 @@
 namespace Nethermind.Blockchain.BeaconBlockRoot;
 public interface IBeaconBlockRootHandler
 {
+    Address? BeaconRootsAddress(Block block, IReleaseSpec spec);
```
LukaszRozmej, line 10: Get prefix?

src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
```diff
@@ -24,7 +24,7 @@ public class BlockCachePreWarmer(ReadOnlyTxProcessingEnvFactory envFactory, ISpe
     private readonly ObjectPool<SystemTransaction> _systemTransactionPool = new DefaultObjectPool<SystemTransaction>(new DefaultPooledObjectPolicy<SystemTransaction>(), Environment.ProcessorCount);
     private readonly ILogger _logger = logManager.GetClassLogger<BlockCachePreWarmer>();
 
-    public Task PreWarmCaches(Block suggestedBlock, Hash256? parentStateRoot, CancellationToken cancellationToken = default)
+    public Task PreWarmCaches(Block suggestedBlock, Hash256? parentStateRoot, Address? beaconRootsAddress, CancellationToken cancellationToken = default)
```
LukaszRozmej, line 27: Would prefer to pass a list of AdditionalAddresses, not going into what those addresses are `params Span` would be most useful, but that is .net 9. This way it will be future proof.

src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
```diff
@@ -33,40 +33,38 @@ public Task PreWarmCaches(Block suggestedBlock, Hash256? parentStateRoot, Cancel
...
     // Parent state root is null for genesis block
-    private bool IsGenesisBlock(Hash256? parentStateRoot) => parentStateRoot is null;
+    private static bool IsGenesisBlock(Hash256? parentStateRoot) => parentStateRoot is null;
 
     public void ClearCaches() => targetWorldState?.ClearCache();
 
     public Task ClearCachesInBackground() => targetWorldState?.ClearCachesInBackground() ?? Task.CompletedTask;
 
-    private void PreWarmCachesParallel(Block suggestedBlock, Hash256 parentStateRoot, CancellationToken cancellationToken)
+    private void PreWarmCachesParallel(Block suggestedBlock, Hash256 parentStateRoot, Address? beaconRootsAddress, ParallelOptions parallelOptions, CancellationToken cancellationToken)
```
LukaszRozmej, line 59: `beaconRootsAddress` not used

src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
```diff
@@ -76,75 +74,144 @@ private void PreWarmCachesParallel(Block suggestedBlock, Hash256 parentStateRoot
...
+        private void WarmupAddresses(ParallelOptions parallelOptions, Block block, IReadOnlyTxProcessingScope scope)
         {
             if (parallelOptions.CancellationToken.IsCancellationRequested) return;
 
+            // Warm up BeaconRootsAddress first as is first used
+            if (BeaconRootsAddress is not null)
+            {
+                scope.WorldState.WarmUp(BeaconRootsAddress);
+                // Contract accesses this storage cell
+                scope.WorldState.WarmUp(new StorageCell(BeaconRootsAddress, block.Timestamp % 8191));
```
LukaszRozmej, line 187: StorageCell calculation should be done in `BeaconBlockRootHandler`

LukaszRozmej (commented):
Can we build general mechanism?

src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
benaadams: Could pass in a SystemTx AccessList maybe?

src/Nethermind/Nethermind.Blockchain/BeaconBlockRoot/BeaconBlockRootHandler.cs
benaadams: Resolved

src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
benaadams: Did AccesList and then can look at a .Combine for future SystemTxs

src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
benaadams: Removed

src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
benaadams: Moved

src/Nethermind/Nethermind.Blockchain/BeaconBlockRoot/BeaconBlockRootHandler.cs
```diff
@@ -15,25 +15,42 @@ public class BeaconBlockRootHandler(ITransactionProcessor processor) : IBeaconBl
...
             {
                 Value = UInt256.Zero,
                 Data = header.ParentBeaconBlockRoot.Bytes.ToArray(),
-                To = beaconRootsAddress,
+                To = toAddress,
                 SenderAddress = Address.SystemUser,
                 GasLimit = GasLimit,
                 GasPrice = UInt256.Zero,
-                AccessList = new AccessList.Builder().AddAddress(beaconRootsAddress).Build()
+                AccessList = accessList
```
LukaszRozmej, line 53: While it might be good idea to add `block.Timestamp % 8191` as storage to access list, Geth is not doing that. And if Geth is not doing that than an edge case can happen where geth transaction goes out of gas and ours don't (in general it shouldn't happen with current gas limit but still) - and that would result in consensus issue.

benaadams: Removed it for the tx access list

LukaszRozmej (approved):

PR #7427, opened 2024-09-13
Use ArrayPool for Discovery Deserialization

benaadams:
## Changes

- Alas need to use `ArraySegment<byte>` rather than `Span<byte>` as the DotNetty types don't accept Span

Mild allocation

image

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Core/Buffers/ArrayPoolDisposableReturn.cs
+++ b/src/Nethermind/Nethermind.Core/Buffers/ArrayPoolDisposableReturn.cs
@@ -1,0 +1,22 @@
+// SPDX-FileCopyrightText: 2024 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System;
+using System.Buffers;
+
+namespace Nethermind.Core.Buffers;
+
+public readonly struct ArrayPoolDisposableReturn : IDisposable
+{
+    private readonly byte[] _array;
+
+    private ArrayPoolDisposableReturn(byte[] array) => _array = array;
+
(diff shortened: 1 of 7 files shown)


src/Nethermind/Nethermind.Core/Buffers/ArrayPoolDisposableReturn.cs
```diff
@@ -0,0 +1,22 @@
+// SPDX-FileCopyrightText: 2024 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System;
+using System.Buffers;
+
+namespace Nethermind.Core.Buffers;
+
+public readonly struct ArrayPoolDisposableReturn : IDisposable
```
LukaszRozmej, line 9: why not use ArrayPoolList?

LukaszRozmej (commented):
Move to ArrayPoolList (add something to it if needed), we don't need more types.

src/Nethermind/Nethermind.Core/Buffers/ArrayPoolDisposableReturn.cs
benaadams: Just wanted type so I could do 

```
using v = ;
```
As single line without either adding a `ArrayPool<byte>.Shared.Return` before each return; or going full try/finally and changing the indentation 😅

LukaszRozmej: ArrayPoolList has that!

PR #7432, opened 2024-09-14
Min time between induced gcs

### LINKED ISSUE #7431: Recent memory optimizations made archive sync significantly slower
After optimizing memory usage on archive it seems like archive is much much slower than it used to be on 1.28.0.
Below screenshot for gnosis which is common use case.

image

benaadams:
Resolves #7431

## Changes

- Add min time between forced GCs
- Add periodic timer to GC during idle
- GC on background thread

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/GC/GCScheduler.cs
+++ b/src/Nethermind/Nethermind.Consensus/GC/GCScheduler.cs
@@ -28,7 +28,7 @@ public sealed class GCScheduler
     private long _countToGC = 0L;
     private Stopwatch _stopwatch = new();
 
-    public static GCScheduler Instance => new GCScheduler();
+    public static GCScheduler Instance { get; } = new GCScheduler();
 
     public GCScheduler()
     {
--- a/src/Nethermind/Nethermind.Consensus/GC/GCScheduler.cs
+++ b/src/Nethermind/Nethermind.Consensus/GC/GCScheduler.cs
@@ -11,13 +11,15 @@ namespace Nethermind.Consensus;
 
 public sealed class GCScheduler
 {
+    private const int CanPerformGC = 0;
+    private const int GCNotAllowed = 1;
     // Thresholds and limits for triggering garbage collection
     private const int BlocksBacklogTriggeringManualGC = 4;
     private const int MaxBlocksWithoutGC = 250;
     private const int MinSecondsBetweenForcedGC = 120;
 
-    // Flag indicating if a garbage collection is currently in progress
-    private static int _isPerformingGC = 0;
+    // Flag indicating if a garbage collection is currently in progress or disallowed
+    private static int _canPerformGC = CanPerformGC;
 
     // Timer for scheduling periodic garbage collections when idle
     private readonly Timer _gcTimer;
@@ -61,14 +63,12 @@ public sealed class GCScheduler
         }
 
         // Avoid setting the timer if there are items in the queue or timer is already set
-        if (queueCount > 0 || _gcTimerSet)
+        if (queueCount <= 0 && !_gcTimerSet)
         {
-            // Don't switch on if still processing blocks
-            return;
+            _gcTimerSet = true;
+            // Schedule GC every 2 minutes when idle
+            _gcTimer.Change(TimeSpan.FromMinutes(2), TimeSpan.FromMinutes(2));
         }
-        _gcTimerSet = true;
-        // Schedule GC every 2 minutes when idle
-        _gcTimer.Change(TimeSpan.FromMinutes(2), TimeSpan.FromMinutes(2));
     }
 
     /// <summary>
@@ -91,13 +91,12 @@ public sealed class GCScheduler
             _fireGC = false;
         }
 
-        if (!_gcTimerSet)
+        if (_gcTimerSet)
         {
-            return;
+            // Stop the GC timer as the system is no longer idle
+            _gcTimerSet = false;
+            _gcTimer.Change(Timeout.Infinite, Timeout.Infinite);
         }
-        // Stop the GC timer as the system is no longer idle
-        _gcTimerSet = false;
-        _gcTimer.Change(Timeout.Infinite, Timeout.Infinite);
     }
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


LukaszRozmej (changes requested):
Can we move this code to separate component like we have `GCKeeper`?

benaadams:
> Can we move this code to separate component like we have `GCKeeper`?

Moved to separate type; also added some coordination with the GCKeeper

benaadams:
Hmm broke something in the extraction; is calling GC all the time now 🤔

src/Nethermind/Nethermind.Merge.Plugin/GC/GCKeeper.cs
```diff
@@ -165,9 +167,7 @@ private async Task ScheduleGCInternal()
                     GCSettings.LargeObjectHeapCompactionMode = GCLargeObjectHeapCompactionMode.CompactOnce;
                 }
 
-                System.GC.Collect((int)generation, mode, blocking: true, compacting: compacting > 0);
-
-                MallocHelper.Instance.MallocTrim((uint)1.MiB());
+                GCScheduler.Instance.GCCollect((int)generation, mode, blocking: true, compacting: compacting > 0);
```
LukaszRozmej, line 170: should lines [165-168](https://github.com/NethermindEth/nethermind/pull/7432/files#diff-79d27af20a7736bf17e78b4578c54be208f0753194a14cfa53557ae2f9a6153fR165-R168) go into this method in GCScheduler?

src/Nethermind/Nethermind.Merge.Plugin/GC/GCKeeper.cs
```diff
@@ -29,7 +30,7 @@ public GCKeeper(IGCStrategy gcStrategy, ILogManager logManager)
     public IDisposable TryStartNoGCRegion(long? size = null)
     {
         size ??= _defaultSize;
-        if (_gcStrategy.CanStartNoGCRegion())
+        if (GCScheduler.MarkGCPaused() && _gcStrategy.CanStartNoGCRegion())
```
LukaszRozmej, line 33: So if there is a GC already we won't start NoGCRegion, right?

src/Nethermind/Nethermind.Consensus/GC/GCScheduler.cs
```diff
@@ -0,0 +1,192 @@
...
+        return Interlocked.CompareExchange(ref _isPerformingGC, 1, 0) == 0;
+    }
+
+    /// <summary>
+    /// Marks that garbage collection has finished.
+    /// </summary>
+    public static void MarkGCResumed()
+    {
+        Volatile.Write(ref _isPerformingGC, 0);
+    }
```
LukaszRozmej, line 129: So instead `_isPerformingGC` we should have `_canPerformGC` with `0` being `true`?

src/Nethermind/Nethermind.Consensus/GC/GCScheduler.cs
```diff
@@ -0,0 +1,192 @@
...
+namespace Nethermind.Consensus;
+
+public sealed class GCScheduler
+{
+    // Thresholds and limits for triggering garbage collection
+    private const int BlocksBacklogTriggeringManualGC = 4;
+    private const int MaxBlocksWithoutGC = 250;
+    private const int MinSecondsBetweenForcedGC = 120;
+
+    // Flag indicating if a garbage collection is currently in progress
```
LukaszRozmej, line 19: or is disallowed

src/Nethermind/Nethermind.Consensus/GC/GCScheduler.cs
```diff
@@ -0,0 +1,192 @@
...
+            _countToGC--;
+            // Trigger GC if block limit reached and minimum time elapsed
+            if (_countToGC <= 0 && _stopwatch.Elapsed.TotalSeconds > MinSecondsBetweenForcedGC)
+            {
+                _fireGC = false;
+                _stopwatch.Reset();
+                // Ensure only one GC task runs at a time
+                if (_lastGcTask.IsCompleted)
+                {
+                    _lastGcTask = PerformFullGCAsync();
```
LukaszRozmej, line 58: Should this be interlocked or something?

src/Nethermind/Nethermind.Consensus/GC/GCScheduler.cs
```diff
@@ -0,0 +1,192 @@
...
+
+        // Avoid setting the timer if there are items in the queue or timer is already set
+        if (queueCount > 0 || _gcTimerSet)
+        {
+            // Don't switch on if still processing blocks
+            return;
+        }
+        _gcTimerSet = true;
+        // Schedule GC every 2 minutes when idle
+        _gcTimer.Change(TimeSpan.FromMinutes(2), TimeSpan.FromMinutes(2));
```
LukaszRozmej, line 71: Minor/me: I dislike `if return` pattern unless on top of the method, would prefer nesting.

LukaszRozmej (approved):

src/Nethermind/Nethermind.Consensus/GC/GCScheduler.cs
benaadams: Currently is all called on same RunProcessingLoop thread

src/Nethermind/Nethermind.Consensus/GC/GCScheduler.cs
benaadams: .NET 9 allows enums to be used with `CompareExchange` so we can clean up some of these unhelpful 0, 1

Renamed and added some constants

src/Nethermind/Nethermind.Consensus/GC/GCScheduler.cs
benaadams: Added

src/Nethermind/Nethermind.Consensus/GC/GCScheduler.cs
benaadams: Inverted the ifs

src/Nethermind/Nethermind.Merge.Plugin/GC/GCKeeper.cs
benaadams: Wasn't my intention. Instead am now capturing the return and if it was true unpause in Dispose, but don't prevent NoGCRegion

PR #7436, opened 2024-09-15
Direct fail create tx rather than throwing exceptions

benaadams:
## Changes

- Direct fail create tx rather than throwing exceptions to immediately catch

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
+++ b/src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
@@ -526,9 +526,7 @@ namespace Nethermind.Evm.TransactionProcessing
 
                 if (substate.ShouldRevert || substate.IsError)
                 {
-                    if (Logger.IsTrace)
-                        Logger.Trace("Restoring state from before transaction");
-                    WorldState.Restore(snapshot);
+                    goto Fail;
                 }
                 else
                 {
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -526,8 +526,7 @@ protected virtual void ExecuteEvmCall(
 
                 if (substate.ShouldRevert || substate.IsError)
                 {
-                    if (Logger.IsTrace)
-                        Logger.Trace("Restoring state from before transaction");
+                    if (Logger.IsTrace) Logger.Trace("Restoring state from before transaction");
```
LukaszRozmej, line 529: `goto Fail;` here too?

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -572,13 +571,17 @@ protected virtual void ExecuteEvmCall(
                 }
 
                 spentGas = Refund(tx, header, spec, opts, substate, unspentGas, env.TxExecutionContext.GasPrice);
+                goto Complete;
             }
             catch (Exception ex) when (ex is EvmException or OverflowException) // TODO: OverflowException? still needed? hope not
             {
                 if (Logger.IsTrace) Logger.Trace($"EVM EXCEPTION: {ex.GetType().Name}:{ex.Message}");
-                WorldState.Restore(snapshot);
```
LukaszRozmej, line 579: `goto Fail;` here too? (probably not needed as it is next instruction?)

LukaszRozmej (approved):

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
benaadams: You get redund on revert would cause it to skip 

`spentGas = Refund...`

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
benaadams: Just falls through

PR #7438, opened 2024-09-15
Allow precache to clear in background

benaadams:
## Changes

- Allow prewarm cache to clear in background on prewarm continuation; only forcing at start of next block if not cleared

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
@@ -31,7 +31,7 @@ public class BlockCachePreWarmer(ReadOnlyTxProcessingEnvFactory envFactory, ISpe
         {
             if (targetWorldState.ClearCache())
             {
-                if (_logger.IsWarn) _logger.Warn("Caches are not empty. Clearing them.");
+                if (_logger.IsDebug) _logger.Debug("Caches are not empty. Clearing them.");
             }
 
             var physicalCoreCount = RuntimeInformation.PhysicalCoreCount;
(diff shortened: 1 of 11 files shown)


src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
```diff
@@ -31,7 +31,7 @@ public Task PreWarmCaches(Block suggestedBlock, Hash256? parentStateRoot, Access
         {
             if (targetWorldState.ClearCache())
             {
-                if (_logger.IsWarn) _logger.Warn("Caches are not empty. Clearing them.");
+                if (_logger.IsDebug) _logger.Debug("Caches are not empty. Clearing them.");
```
LukaszRozmej, line 34: This generally should not happen in normal execution, why lower to Debug?

src/Nethermind/Nethermind.Core/Collections/CollectionExtensions.cs
```diff
@@ -40,6 +40,12 @@ public static bool NoResizeClear<TKey, TValue>(this ConcurrentDictionary<TKey, T
 
             using var handle = dictionary.AcquireLock();
 
+            // Recheck under lock
+            if (dictionary is null || dictionary.IsEmpty)
```
LukaszRozmej, line 44: `dictionary?.IsEmpty == false`

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
```diff
@@ -330,7 +330,6 @@ public void CommitTrees(long blockNumber)
             _toUpdateRoots.Clear();
             // only needed here as there is no control over cached storage size otherwise
             _storages.Clear();
-            _preBlockCache?.NoResizeClear();
```
LukaszRozmej, line 333: Not needed now?

src/Nethermind/Nethermind.State/StateProvider.cs
```diff
@@ -857,7 +857,6 @@ public void CommitTree(long blockNumber)
             }
 
             _tree.Commit(blockNumber);
-            _preBlockCache?.NoResizeClear();
```
LukaszRozmej, line 860: Not needed now?

src/Nethermind/Nethermind.Core/Collections/CollectionExtensions.cs
benaadams: > dictionary?.IsEmpty == false

Crashes with that 

`dictionary?.IsEmpty ?? true` works though

benaadams:
Something not happy with this. Will make a simpler PR

benaadams:
Simpler that is kinda and optimiaztion bug fix https://github.com/NethermindEth/nethermind/pull/7442

PR #7447, opened 2024-09-17
5912 only static peers

### LINKED ISSUE #5912: OnlyStaticPeers should disallow connection from random peers. Move DicoveryEnabled to NetworkConfig. OnlyStaticPeers should implicitly disable Discovery.
**Description**
`OnlyStaticPeers` should disallow connection from random peers. Move `DiscoveryEnabled` to `NetworkConfig`. `OnlyStaticPeers` should implicitly disable Discovery.

**Steps to Reproduce**
1. Run nethermind with `OnlyStaticPeers` enabled
2. Notice how nethermind still connects to non static peers

**Actual behavior**
Connects to non static peers

**Expected behavior**
Should not connect to non static peers

 - Windows
 - 1.20
 - debugging
 - Prysm

ssonthal:
Fixes #5912 

## Changes

- Adding DiscoveryEnabled field in INetworkConfig and NetworkConfig classes. 
- Setting DiscoveryEnabled filed as !OnlyStaticPeers to ensure it gets set correctly at the beginning. 
- Adding an if condition based on DiscoveryEnabled value in initializeAllPeers method.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No


shubham-sonthalia:
@smartprogrammer93 please let me know if these changes look good.

smartprogrammer93:
@shubham-sonthalia nope, that is not what is needed here. There is already a DiscoveryEnabled config in InitConfigs, so you need to migrate that to Network. Then, set DiscoveryEnabled=false when OnlyStaticPeers=true. Also, you need to reject any incoming connections from peers that are not in the static list when OnlyStaticPeers=true.

shubham-sonthalia:
Hey @smartprogrammer93, I copied the DiscoveryEnabled field from InitConfig and set it to false if OnlyStaticPeer = true. 

For rejecting any incoming request from peer, I couldn't pinpoint the exact place. But through some analysis, I have added it in the AddNodes() method in the NodeTable class. Let me know if this looks good.

shubham-sonthalia:
I have moved the rejecting of request part of PoolManager class -> ProcessIncomingRequest method.

shubham-sonthalia:
Hey @smartprogrammer93 did you get a chance to go over the PR? Have I missed anything here?

smartprogrammer93:
@shubham-sonthalia yeah there is a couple. Changing the value of discovery based on static peers this way is incorrect. Also, migrating a sertting needs to not break existing setups. You need to deal with the old variable and migrate it to the correct place. And change all the usages.

shubham-sonthalia:
@smartprogrammer93 so should I remove the field DiscoveryEnabled from IInitConfig and then wherever this field was used - instead of IInitConfig, it should come from INetworkConfig?

ssonthal:
@smartprogrammer93 can you please comment on the above query?

asdacap:
I belive this is redundant with #7514

PR #7464, opened 2024-09-19
fix casting when tracing

LukaszRozmej:
## Changes

- Removes unsafe casts when tracing

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No


benaadams (approved):
Worried about api change for external consumers?

LukaszRozmej:
> Worried about api change for external consumers?

shouldn't be that bad, not many external consumers

PR #7469, opened 2024-09-21
Move txPool work from Processing loop event to txPool thread

benaadams:
## Changes

- Move clearing caches to txPool processing new head rather than main processing informing txPool of new head

2.9% of archive since since its processing blocks so fast 
image

- [x] Optimization

#### Requires testing

- [x] Yes

- [x] Yes (changed existing tests)

### DIFF
--- a/src/Nethermind/Nethermind.TxPool/TxPool.cs
+++ b/src/Nethermind/Nethermind.TxPool/TxPool.cs
@@ -191,10 +191,6 @@ namespace Nethermind.TxPool
         {
             try
             {
-                // Clear snapshot
-                _transactionSnapshot = null;
-                _blobTransactionSnapshot = null;
-                _hashCache.ClearCurrentBlockCache();
                 _headBlocksChannel.Writer.TryWrite(e);
             }
             catch (Exception exception)
@@ -211,8 +207,12 @@ namespace Nethermind.TxPool
             {
                 while (await _headBlocksChannel.Reader.WaitToReadAsync())
                 {
+                    _hashCache.ClearCurrentBlockCache();
                     while (_headBlocksChannel.Reader.TryRead(out BlockReplacementEventArgs? args))
                     {
+                        // Clear snapshot
+                        _transactionSnapshot = null;
+                        _blobTransactionSnapshot = null;
                         try
                         {
                             ArrayPoolList<AddressAsKey>? accountChanges = args.Block.AccountChanges;
--- a/src/Nethermind/Nethermind.TxPool/TxPool.cs
+++ b/src/Nethermind/Nethermind.TxPool/TxPool.cs
@@ -59,7 +59,7 @@ namespace Nethermind.TxPool
         private readonly UpdateGroupDelegate _updateBucket;
         private readonly UpdateGroupDelegate _updateBucketAdded;
 
-        public event EventHandler<BlockReplacementEventArgs>? TxPoolHeadChanged;
+        public event EventHandler<Block>? TxPoolHeadChanged;
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -205,14 +201,20 @@ private void OnHeadChange(object? sender, BlockReplacementEventArgs e)
             }
         }
 
+        public event EventHandler<BlockReplacementEventArgs>? HeadChanged;
```
LukaszRozmej, line 204: Having this event here feels super-weird. Can we at least make it somehow work in txpool context?

benaadams: Rename to `TxPoolHeadChanged`?

benaadams: Also used the event

src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -191,10 +194,6 @@ private void OnHeadChange(object? sender, BlockReplacementEventArgs e)
         {
             try
             {
-                // Clear snapshot
-                _transactionSnapshot = null;
-                _blobTransactionSnapshot = null;
-                _hashCache.ClearCurrentBlockCache();
```
LukaszRozmej, line 197: Could that lead to some inconsistencies in TxPool, so it could give us stale data in normal (not archive) processing mode?

benaadams: Shouldn't do as we do the bucket Update etc on the other half where this moves to

src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -191,10 +194,6 @@ private void OnHeadChange(object? sender, BlockReplacementEventArgs e)
         {
             try
             {
-                // Clear snapshot
-                _transactionSnapshot = null;
-                _blobTransactionSnapshot = null;
```
LukaszRozmej, line 196: Different question, can we make those two use fixed memory?
Basically we should be able to allocate something ~max tx pool count and only update it when needed, instead of allocating it all the time.

benaadams: Don't think so easily; txPool is quite a complicated set of data structures and don't know when the snapshot has finished being used (to change it)

However could use it in one more place

LukaszRozmej: Maybe at least pool them?

benaadams: They don't have a very good lifecycle as are multi-use

Would need to introduce a ref counted ArrayPoolList type thing.

Out of scope for this PR though?

LukaszRozmej: eh ok :(

benaadams: Something to look at with a bigger refactor of txpool

benaadams: Something to look at with a bigger refactor of txpool vs removing time impact on main processing loop

LukaszRozmej (approved):
Let's wait for @marcindsobczak approval before merging

benaadams:
Aside this also fix JsonRpc tests timing out

marcindsobczak (approved):
LGTM, worth to run on some node/nodes and observe for a few days

PR #7471, opened 2024-09-21
Move EnsureCanonical off main loop tx processing

benaadams:
## Changes

- Move EnsureCanonical to the background task and off main tx processing thread

2.8% of archive since since its processing blocks so fast
image

- [x] Optimization

#### Requires testing

- [x] No


LukaszRozmej (changes requested):
This is very much by design. If we move this to background thread we allow a situation where Head is already changed, but receipts cannot be found for given transactions through RPC. Data inconsistency for the user.

This is generally not acceptable in normal processing. Although we could design a separate batch mode for archive sync where this happens.

https://github.com/NethermindEth/nethermind/pull/6422

smartprogrammer93 (changes requested):
Yeah, maybe only on Archive or we can also do it under a certain flag in the configs. this way if someone wants the absolute speed, they can ignore receipts being consistent with head

benaadams:
> Yeah, maybe only on Archive or we can also do it under a certain flag in the configs. this way if someone wants the absolute speed, they can ignore receipts being consistent with head

Or add a cache for recent receipts?

benaadams:
> This is generally not acceptable in normal processing. Although we could design a separate batch mode for archive sync where this happens.

Added change work?

LukaszRozmej:
> > This is generally not acceptable in normal processing. Although we could design a separate batch mode for archive sync where this happens.
> 
> Added change work?

Not really, so one could subscribe to newhead subscription or just pull from standard rpc and then not find the receipts. Would be bad.

We need to delay the tree head to be completely safe.

benaadams:
Superceded by https://github.com/NethermindEth/nethermind/pull/12349

PR #7473, opened 2024-09-21
Calculate receipt blooms in parallel

benaadams:
## Changes

- Calculate the receipt blooms in parallel; saves ~1% on block validation

Before
image
After
image

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Evm/Tracing/BlockReceiptsTracer.cs
+++ b/src/Nethermind/Nethermind.Evm/Tracing/BlockReceiptsTracer.cs
@@ -78,7 +78,7 @@ public class BlockReceiptsTracer : IBlockTracer, ITxTracer, IJournal<int>, ITxTr
         {
             Logs = logEntries,
             TxType = transaction.Type,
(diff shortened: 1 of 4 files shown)


src/Nethermind/Nethermind.Evm/Tracing/BlockReceiptsTracer.cs
```diff
@@ -78,7 +78,7 @@ protected virtual TxReceipt BuildReceipt(Address recipient, long spentGas, byte
         {
             Logs = logEntries,
             TxType = transaction.Type,
-            Bloom = logEntries.Length == 0 ? Bloom.Empty : new Bloom(logEntries),
+            // Bloom calculated in parallel with other receipts
```
LukaszRozmej, line 81: Maybe better would be to start a task here, and then await it in BlockProcessor?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Evm/Tracing/BlockReceiptsTracer.cs
benaadams: 🤔

PR #7483, opened 2024-09-23
Refactor `TransactionForRpc` (Alternative)

emlautarom1:
Partially solves #7313

## Changes

- Make `TransactionForRpc` an abstract base class
- Introduce multiple subclasses of `TransactionForRpc` for specific Transaction types (`Legacy`, `AccessList`, `EIP1559`, `Blob` and `Optimism`)
- Support JSON (de)serializing
- Removed `TransactionForRpc -> Transaction` "default" conversions: conversions now use domain specific defaults (ex. `GasLimit` defaults to `90_000`) instead of C# specific.
- Introduced mechanism to register new subclasses.

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

- Added schema verification tests based on the Ethereum JSON-RPC spec (see https://ethereum.github.io/execution-apis/api-documentation/)

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks

This PR is an alternative version of #7446 that does not introduce a `RpcGenericTransaction` for deserializing purposes; instead we reuse the same type hierarchy for both serialization and deserialization. As a quick summary, this PR introduces the following changes:

## When deserializing

- We read JSON and construct a specific subclass of `TransactionForRpc` based on the `Type` field of the incoming JSON object (if this field is missing we default to `Legacy` transactions).
- Each subclass should implement a `ToTransaction` method that constructs a proper `Transaction` object while setting up the appropriate domain defaults when fields are missing (ex. if the transaction is `AccessList`, then the resulting `AccessList` field in `Transaction` is an empty access list, not `null`).

## When serializing
- We have multiple `Converter`s that know how to take a `Transaction` and build a `TransactionForRpc` (the above mentioned abstract base class). Each converter knows how to take a Transaction with a specific `TxType` and build a subclass (ex. there is a converter from `Transaction` to `BlobTransactionForRpc`). Each converter is based on the Ethereum JSON-RPC spec to ensure no fields are missing.
- To write JSON, we instruct the serializer to use the runtime type rather than the static type, that is, don't use `TransactionForRpc` but, for example, `EIP1559TransactionForRpc`.


src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/LegacyTransactionForRpc.cs
```diff
@@ -0,0 +1,115 @@
...
+    {
+        if (gasCap is null || gasCap == 0)
+            gasCap = long.MaxValue;
+
+        Gas = Gas is null || Gas == 0
+            ? gasCap
+            : Math.Min(gasCap.Value, Gas.Value);
+
+        From ??= Address.SystemUser;
+    }
```
emlautarom1, line 106: As the comment asks, can we remove this code?

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
```diff
@@ -0,0 +1,154 @@
...
+        // TODO: Refactoring transition code
+        public JsonConverter()
+        {
+            RegisterTransactionType(TxType.Legacy, typeof(LegacyTransactionForRpc));
+            RegisterTransactionType(TxType.AccessList, typeof(AccessListTransactionForRpc));
+            RegisterTransactionType(TxType.EIP1559, typeof(EIP1559TransactionForRpc));
+            RegisterTransactionType(TxType.Blob, typeof(BlobTransactionForRpc));
+            // TODO: Add Optimism:
+            // RegisterTransactionType(TxType.DepositTx, typeof(RpcOptimismTransaction))
+        }
```
emlautarom1, line 76: For ease of usage `TransactionForRpc` uses this converter which initializes itself with known transaction types. I think our options are

- Manually register an instance of this converter with manually registered known subclasses. This makes testing harder since we need to add this converter instance to each test that uses the default `EthereumJsonSerializer`.
- Assembly scan all subclasses of `TransactionForRpc` and register them in the converter default constructor, keeping the `[JsonConverter(typeof(JsonConverter))]` annotation. This approach is the one used on RLP (de)serialization.

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
```diff
@@ -0,0 +1,154 @@
...
+
+        public override TransactionForRpc? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        {
+            using var document = JsonDocument.ParseValue(ref reader);
+
+            // TODO: For some reason we might get an object wrapped in a String
+            using JsonDocument jsonObject = document.RootElement.ValueKind == JsonValueKind.String
+                ? JsonDocument.Parse(document.RootElement.GetString()!)
+                : document;
```
emlautarom1, line 96: @benaadams Do you know why on RPC parameter execution we might get JSON objects wrapped in strings? For example, instead of getting:

```json
{ "foo": 42 }
```

I get:

```json
"{\"foo\": 42 }"
```

This workaround deserializes the object again so we can actually treat as an object.

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
```diff
@@ -0,0 +1,154 @@
...
+
+            // NOTE: Should we default to a specific TxType?
+            TxType discriminator = default;
```
emlautarom1, line 99: We were (and we continue to) default to `Legacy` transactions.

src/Nethermind/Nethermind.JsonRpc.Test/Data/Eip2930Tests.cs
```diff
@@ -231,8 +238,11 @@ public void can_convert_fromTransaction_toTransactionForRpc_and_back(TxType txTy
         {
             Type = txType,
             AccessList = GetTestAccessList(),
+            ChainId = BlockchainIds.Mainnet,
+            SenderAddress = Address.SystemUser,
+            Data = Memory<byte>.Empty,
```
emlautarom1, line 243: Added only due to roundtripping comparison: converting to a `TransactionForRpc` instance will default these fields, and going back to `Transaction` will preserve those defaults.

src/Nethermind/Nethermind.JsonRpc.Test/Modules/Eth/EthRpcModuleTests.cs
```diff
@@ -78,15 +78,23 @@ public async Task Eth_get_transaction_by_block_hash_and_index()
...
+
+        var actualJson = JObject.Parse(serialized);
+        var expectedJson = JObject.Parse(
+            """{"jsonrpc":"2.0","result":{"hash":"0x7126cf20a0ad8bd51634837d9049615c34c1bff5e1a54e5663f7e23109bff48b","nonce":"0x2","blockHash":"0x29f141925d2d8e357ae5b6040c97aa12d7ac6dfcbe2b20e7b616d8907ac8e1f3","blockNumber":"0x3","transactionIndex":"0x1","from":"0xb7705ae4c6f81b66cdb323c65f4e8133690fc099","to":"0x942921b14f1b1c385cd7e0cc2ef7abe5598c8358","value":"0x1","gasPrice":"0x1","gas":"0x5208","input":"0x","type":"0x0","v":"0x25","s":"0x575361bb330bf38b9a89dd8279d42a20d34edeaeede9739a7c2bdcbe3242d7bb","r":"0xe7c5ff3cba254c4fe8f9f12c3f202150bb9a0aebeee349ff2f4acb23585f56bd"},"id":67}""");
+        actualJson.Should().BeEquivalentTo(expectedJson);
```
emlautarom1, line 85: Several tests that used to do string comparison now perform JSON equality. This is to avoid failing tests due to different key ordering in JSON objects.

src/Nethermind/Nethermind.JsonRpc.Test/Modules/Eth/EthRpcModuleTests.cs
```diff
@@ -1161,13 +1195,36 @@ public async Task Eth_create_access_list_sample(AccessListProvided accessListPro
...
+        long ensureDefaultsGasLimit;
+        {
+            var rpcTx = new LegacyTransactionForRpc();
+            rpcTx.EnsureDefaults(gasCap);
+            var tx = rpcTx.ToTransaction();
+            ensureDefaultsGasLimit = tx.GasLimit;
+        }
+
+        toTransactionWitDefaultsGasLimit.Should().Be(ensureDefaultsGasLimit);
+    }
```
emlautarom1, line 1226: It's not clear to me why we have the `EnsureDefault`s method. Again, can we delete it?

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.TransactionExecutor.cs
```diff
@@ -43,9 +44,28 @@ protected override ResultWrapper<TResult> Execute(BlockHeader header, Transactio
...
+                {
+                    positiveMaxFeePerGas = IsPositive(eip1559.MaxFeePerGas);
+                    positiveMaxPriorityFeePerGas = IsPositive(eip1559.MaxPriorityFeePerGas);
+                }
+
+                return positiveGasPrice || positiveMaxFeePerGas || positiveMaxPriorityFeePerGas;
+
+                // value?.IsZero == false <=> x > 0
+                static bool IsPositive(UInt256? value) => value?.IsZero == false;
+            }
```
emlautarom1, line 68: As the comment says, should we move this method to the base `TransactionForRpc` class directly?

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
```diff
@@ -280,8 +281,17 @@ public ResultWrapper<byte[]> eth_sign(Address addressData, byte[] message)
...
-        TxHandlingOptions options = rpcTx.Nonce is null ? TxHandlingOptions.ManagedNonce : TxHandlingOptions.None;
+        Transaction tx = rpcTx.ToTransaction();
+        tx.ChainId = _blockchainBridge.GetChainId();
+
+        // TODO: We might want to lift `Nonce` to `TransactionForRpc`
+        UInt256? nonce = null;
+        if (rpcTx is LegacyTransactionForRpc legacyTx)
+        {
+            nonce = legacyTx.Nonce;
+        }
```
emlautarom1, line 292: As the comment says, should we lift `Nonce` to the base `TransactionForRpc` class?

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/SimulateTxExecutor.cs
```diff
@@ -69,17 +72,28 @@ protected override SimulatePayload<TransactionWithSourceDetails> Prepare(Simulat
...
+                From = legacy.From,
+                Gas = legacy.Gas,
+                Value = legacy.Value,
+                Input = legacy.Input,
+                GasPrice = legacy.GasPrice,
+                ChainId = legacy.ChainId,
+                V = legacy.V,
+                R = legacy.R,
+                S = legacy.S,
+            };
```
emlautarom1, line 93: We're kind of "upgrading" the transaction type. Previous code did not take into consideration `AccessList` transaction (was this intentional)?

src/Nethermind/Nethermind.Optimism/Rpc/OptimismEthRpcModule.cs
```diff
@@ -106,16 +106,18 @@ public OptimismEthRpcModule(
...
+        Transaction tx = rpcTx.ToTransaction();
+        tx.ChainId = _blockchainBridge.GetChainId();
         tx.SenderAddress ??= _ecdsa.RecoverAddress(tx);
 
         if (tx.SenderAddress is null)
         {
             return ResultWrapper<Hash256>.Fail("Failed to recover sender");
         }
 
+        // TODO: Optimism transactions do not have a Nonce (it's always 0 when queried)
```
emlautarom1, line 120: Not sure what's the intention behind the orignal code. In the original `op-geth` codebase `Nonce` is implemented as method (not a field) that always returns `0` (see https://github.com/ethereum-optimism/op-geth/blob/8af19cf20261c0b62f98cc27da3a268f542822ee/core/types/deposit_tx.go#L79). Why are we setting a particular nonce here?

src/Nethermind/Nethermind.Runner/Ethereum/Api/ApiBuilder.cs
```diff
@@ -83,7 +81,7 @@ private ChainSpec LoadChainSpec(IJsonSerializer ethereumJsonSerializer)
 
             IChainSpecLoader loader = new ChainSpecLoader(ethereumJsonSerializer);
             ChainSpec chainSpec = loader.LoadEmbeddedOrFromFile(chainSpecFile, _logger);
-            TransactionForRpc.DefaultChainId = chainSpec.ChainId;
+            AccessListTransactionForRpc.DefaultChainId = chainSpec.ChainId; // TODO: Figure a proper fix instead of this hack
```
emlautarom1, line 84: Not a new hack, just a comment to remember the existing one.

src/Nethermind/Nethermind.Optimism/Rpc/OptimismTransactionForRpc.cs
```diff
@@ -3,27 +3,96 @@
...
+
+    #region Nethermind specific fields
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public UInt256? DepositReceiptVersion { get; set; }
+    #endregion
+
+#pragma warning disable CS8618
+    [JsonConstructor]
+    public OptimismTransactionForRpc() { }
+#pragma warning restore CS8618
```
emlautarom1, line 47: Several fields are required based on the original Optimism implementation (see: https://github.com/ethereum-optimism/op-geth/blob/60038121c7571a59875ff9ed7679c48c9f73405d/core/types/transaction_marshalling_test.go#L27-L68) but we we're accepting null/missing values.

Should we continue accepting these transactions or should we throw when required fields are missing?

emlautarom1:
Note that this PR **SHOULD NOT BE MERGED** right now. We want to wait for #7459. Also, if we merge right now we'll break Optimism support.

LukaszRozmej (commented):
1. Pushed some more changes - details in commit
2. Few comments added
3. Tests changes could be simplified with helper methods/class

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
LukaszRozmej: I changed it to register in one place

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/LegacyTransactionForRpc.cs
LukaszRozmej: we need it for defaults on eth_call and others

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
LukaszRozmej: Yes, no idea why it isn't there in the first place

emlautarom1: It was not on the base type since new Transaction types might not have a `Nonce`. For example, in Optimism we kind of don't have a Nonce (see https://github.com/ethereum-optimism/op-geth/blob/8af19cf20261c0b62f98cc27da3a268f542822ee/core/types/deposit_tx.go#L29-L46)

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/AccessListTransactionForRpc.cs
```diff
@@ -0,0 +1,56 @@
...
+using Nethermind.Core.Crypto;
+using Nethermind.Int256;
+
+namespace Nethermind.Facade.Eth.RpcTransaction;
+
+public class AccessListTransactionForRpc : LegacyTransactionForRpc
+{
+    // HACK: To ensure that serialized Txs always have a `ChainId` we keep the last loaded `ChainSpec`.
+    // See: https://github.com/NethermindEth/nethermind/pull/6061#discussion_r1321634914
+    public static ulong? DefaultChainId { get; set; }
```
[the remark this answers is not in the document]
emlautarom1: Agree. Worst case scenario we'll have to inject instances of `IBlockchainBridge` to get the current `ChainId` on each call site.

src/Nethermind/Nethermind.JsonRpc.Test/Modules/RpcTransaction/BlobTransactionForRpcTests.cs
```diff
@@ -0,0 +1,136 @@
...
+
+        Build.WithSignature(TestItem.RandomSignatureA).TestObject,
+        Build.WithSignature(TestItem.RandomSignatureB).TestObject,
+    ];
+
+    public static void ValidateSchema(JsonElement json)
+    {
+        // NOTE: This does not look correct since we are not checking for the exact value (`0x3`).
+        // Essentially we're accepting all hex values `0x00 ... 0xFF`, including `0x`
+        json.GetProperty("type").GetString().Should().MatchRegex("^0x([0-9a-fA-F]?){1,2}$");
```
emlautarom1, line 96: According to the spec (https://github.com/ethereum/execution-apis/blob/71260dde2d7fc5a12b9d171692a19a5e599820cf/src/schemas/transaction.yaml#L20) any hex byte should be acceptable.

I think this is a mistake and instead it should be defined to be a specific value like in, for example, EIP1559 (https://github.com/ethereum/execution-apis/blob/71260dde2d7fc5a12b9d171692a19a5e599820cf/src/schemas/transaction.yaml#L93-L96)

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/AccessListTransactionForRpc.cs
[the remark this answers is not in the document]
LukaszRozmej: Or just ISpecProvider

src/Nethermind/Nethermind.JsonRpc.Test/Modules/Eth/EthRpcModuleTests.cs
emlautarom1: Changed to `JToken.Parse(serialized).Should().BeEquivalentTo(/* some string */)`. I think this should be brief enough without compromising readability

src/Nethermind/Nethermind.Facade/Eth/IFromTransaction.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace Nethermind.Facade.Eth;
 
-public interface IFromTransaction<out T> : ITxTyped where T : TransactionForRpc
+public interface IFromTransaction<out T> where T : TransactionForRpc
```
LukaszRozmej, line 11: why seperate?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Facade/Eth/IFromTransaction.cs
emlautarom1: It was separated to allow the `TransactionForRpc.TransactionConverter` to also implement `IFromTransaction<T>` for consistency. Since that class will be hidden (`internal`) I don't think this change is needed. Reverted.

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/AccessListTransactionForRpc.cs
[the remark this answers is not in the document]
emlautarom1: Removed the hack. Now we pass a default `ChainId` on each usage of `ToTransaction`.

Nikki601 (approved):

Nikki601 (approved):

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.TransactionExecutor.cs
emlautarom1: Moved to the base `TransactionForRpc` class

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
emlautarom1: `JsonDocument` might not be the most efficient API to use. See https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/converters-how-to?pivots=dotnet-8-0#support-polymorphic-deserialization for more details.

src/Nethermind/Nethermind.Optimism/Rpc/OptimismTransactionForRpc.cs
```diff
@@ -3,27 +3,91 @@
...
+
+        tx.SourceHash = SourceHash ?? throw new ArgumentNullException(nameof(SourceHash));
+        tx.SenderAddress = From ?? throw new ArgumentNullException(nameof(From));
+        tx.To = To;
+        tx.Mint = Mint ?? 0;
+        tx.Value = Value ?? throw new ArgumentNullException(nameof(Value));
+        // TODO: Unsafe cast
+        tx.GasLimit = (long)(Gas ?? throw new ArgumentNullException(nameof(Gas)));
+        tx.IsOPSystemTransaction = IsSystemTx ?? false;
+        tx.Data = Input ?? throw new ArgumentNullException(nameof(Input));
```
emlautarom1, line 81: According to the official OP repo these fields are required and we should reject transactions that don't provide it:

https://github.com/ethereum-optimism/op-geth/blob/c283254e5447f127e3b9350860985911dab9cd2f/core/types/transaction_marshalling_test.go#L27-L67

src/Nethermind/Nethermind.Optimism/Rpc/OptimismTransactionForRpc.cs
emlautarom1: Changed to throwing when required fields are missing.

src/Nethermind/Nethermind.JsonRpc.Test/Modules/Eth/EthSimulateTestsPrecompilesWithRedirection.cs
```diff
@@ -144,17 +144,15 @@ function ecrecover(bytes32 hash, uint8 v, bytes32 r, bytes32 s) public  returns(
...
             To = contractAddress,
             SenderAddress = TestItem.AddressA,
             GasLimit = 3_500_000,
             GasPrice = 20.GWei()
-        })
-        {
-            Nonce = null
-        };
+        });
+        ((LegacyTransactionForRpc)transactionForRpc).Nonce = null;
```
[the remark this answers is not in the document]
emlautarom1: It was not on the base type since new Transaction types might not have a `Nonce`. For example, in Optimism we kind of don't have a Nonce (see https://github.com/ethereum-optimism/op-geth/blob/8af19cf20261c0b62f98cc27da3a268f542822ee/core/types/deposit_tx.go#L29-L46)

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
LukaszRozmej: Yeah not the biggest fan of that, can we try using type discriminators: https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/polymorphism?pivots=dotnet-8-0#customize-the-type-discriminator-name

```
[JsonPolymorphic(TypeDiscriminatorPropertyName = "TxType")]
[JsonDerivedType(typeof(LegacyTransactionForRpc), typeDiscriminator: "Legacy")] 
```

or `[JsonDerivedType(typeof(LegacyTransactionForRpc), typeDiscriminator: 0)]`

as we would have to make it pluginable maybe we should create custom resolver: https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/polymorphism?pivots=dotnet-8-0#configure-polymorphism-with-the-contract-model which would register all the tx types.

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
LukaszRozmej: Will deposit transaction interact with JSON RPC? Shouldn't it nonce be expected to be outputted there?

Otherwise: true, but `Nonce` sounds like such a basic transaction functionality that maybe it is better to make an exception, put it in base class and just ignore it or `DepositTransaction`? Or this shouldn't take in `TransactionForRpc` but `LegacyTransactionForRpc`? Or maybe create a class in between called like `UserTransactionForRpc`?

src/Nethermind/Nethermind.Optimism/Rpc/OptimismTransactionForRpc.cs
LukaszRozmej: Deposit transaction should never come from RPC? Maybe we should just fail here.

LukaszRozmej (changes requested):
Close but few things came up.

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/AccessListForRpc.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            Address = address;
+            StorageKeys = storageKeys;
+        }
+    }
+
+    public static AccessListForRpc FromAccessList(AccessList? accessList) =>
+        accessList is null
+        ? new AccessListForRpc([])
+        : new AccessListForRpc(accessList.Select(item => new Item(item.Address, [.. item.StorageKeys])).ToList());
```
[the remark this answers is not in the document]
emlautarom1: Yes it was: https://github.com/NethermindEth/nethermind/blob/f825abde15a1db44de82a9d8bda9d2ae7065c2d7/src/Nethermind/Nethermind.Facade/Eth/AccessListItemForRpc.cs#L32-L33

I thought that this laziness was not intentional but rather an artifact of using `IEnumerable`. Changed back to being lazy.

emlautarom1:
> Will deposit transaction interact with JSON RPC? Shouldn't it nonce be expected to be outputted there?

> Otherwise: true, but `Nonce` sounds like such a basic transaction functionality that maybe it is better to make an exception, put it in base class and just ignore it or `DepositTransaction`? Or this shouldn't take in `TransactionForRpc` but `LegacyTransactionForRpc`? Or maybe create a class in between called like `UserTransactionForRpc`?

@LukaszRozmej `Nonce` is not expected to come as part of deposit transactions and we'll get them in, for example, `eth_sendTransaction`. The `Nonce` field is only used during serialization for compatibility reasons and it uses the transaction receipt (see https://specs.optimism.io/protocol/deposits.html?search=#the-deposited-transaction-type). In all other scenarios is treated as `0`.

I've decided to change the input of `eth_sendTransaction` in the `OptimismRpcModule` module to `TransactionForRpc` to ensure we're overriding the base `EthRpcModule`. This override ignores the `Nonce` field (we let it default to `0` in the base `Transaction`) without changing the base Eth module behavior.

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
emlautarom1: Here is where I'm currently at when it comes to custom type resolvers:

I've created the following modifier function:

```C#
public static Action<JsonTypeInfo> JsonTypeInfoModifier(params (Type, TxType)[] txTypes)
{
    return jsonTypeInfo =>
    {
        Type baseValueObjectType = typeof(TransactionForRpc);
        if (jsonTypeInfo.Type == baseValueObjectType)
        {
            jsonTypeInfo.PolymorphismOptions = new JsonPolymorphismOptions
            {
                TypeDiscriminatorPropertyName = "type",
                IgnoreUnrecognizedTypeDiscriminators = true,
                UnknownDerivedTypeHandling = JsonUnknownDerivedTypeHandling.FailSerialization,
            };

            foreach (var (type, txType) in txTypes)
            {
                jsonTypeInfo.PolymorphismOptions.DerivedTypes.Add(new JsonDerivedType(type, (byte)txType));
            }
        }
    };
}
```

Our `EthereumJsonSerializer` needs to be modified to support injecting this modifier function, which we can do somewhat similar to how we're currently adding converters:

https://github.com/NethermindEth/nethermind/blob/f825abde15a1db44de82a9d8bda9d2ae7065c2d7/src/Nethermind/Nethermind.Serialization.Json/EthereumJsonSerializer.cs#L99-L106

We add

```C#
private static readonly List<Action<JsonTypeInfo>> _typeInfoResolverModifiers = new();
public static void AddTypeInfoResolverModifier(Action<JsonTypeInfo> modifier)
{
    _typeInfoResolverModifiers.Add(modifier);
        
    JsonOptions = CreateOptions(indented: false);
    JsonOptionsIndented = CreateOptions(indented: true);
}
```

And we modify

https://github.com/NethermindEth/nethermind/blob/f825abde15a1db44de82a9d8bda9d2ae7065c2d7/src/Nethermind/Nethermind.Serialization.Json/EthereumJsonSerializer.cs#L57-L61

to

```C#
private static JsonSerializerOptions CreateOptions(bool indented, IEnumerable<JsonConverter> converters = null, int maxDepth = 64)
{
    IJsonTypeInfoResolver typeInfoResolver = new DefaultJsonTypeInfoResolver();
    foreach (Action<JsonTypeInfo> resolverModifier in _typeInfoResolverModifiers)
    {
        typeInfoResolver = typeInfoResolver.WithAddedModifier(resolverModifier);
    }
        
    var options = new JsonSerializerOptions
    {
        TypeInfoResolver = typeInfoResolver,
        WriteIndented = indented,
```

There are some issues with this approach:

- It does not provide a mechanism to default to a specific type when the type discriminator is missing, instead it defaults to the base class of the hierarchy which in our case is abstract, thus throwing an exception. For our use case we want to default to `LegacyTransactionForRpc` when the discriminator is missing.
- My attempts at getting it to work don't trigger this error (which I get in a toy project that I'm using for testing), instead it fails with a "no available parameterless constructor" exception (note that we do have parameterless constructors). The full exception message is the following:

```Deserialization of types without a parameterless constructor, a singular parameterized constructor, or a parameterized constructor annotated with 'JsonConstructorAttribute' is not supported. Type 'Nethermind.Facade.Eth.RpcTransaction.TransactionForRpc'```

File: src/Nethermind/Nethermind.JsonRpc.Test/Modules/RpcTransaction/BlobTransactionForRpcTests.cs (hunk 15 above)
Comment (emlautarom1) [reply src/Nethermind/Nethermind.JsonRpc.Test/Modules/RpcTransaction/BlobTransactionForRpcTests.cs:96]: Replaced with `^0x3$` since it's most likely an error in the spec.

File: src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs [hunk 20]
```diff
@@ -80,21 +80,20 @@ internal static void RegisterTransactionType<T>() where T : TransactionForRpc, I
...
-            using JsonDocument jsonObject = document.RootElement.ValueKind == JsonValueKind.String
-                ? JsonDocument.Parse(document.RootElement.GetString()!)
-                : document;
-
-            TxType discriminator = DefaultTxType;
-            if (jsonObject.RootElement.TryGetProperty("type", out JsonElement typeProperty))
+            if (reader.ValueIsEscaped)
             {
-                discriminator = (TxType?)typeProperty.Deserialize(typeof(TxType), options) ?? DefaultTxType;
+                string escapedJson = JsonSerializer.Deserialize<string>(ref reader);
```
LukaszRozmej, line 85: Do we have to allocate the string?

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
```diff
@@ -80,21 +80,20 @@ internal static void RegisterTransactionType<T>() where T : TransactionForRpc, I
...
+                return JsonSerializer.Deserialize<TransactionForRpc>(escapedJson, options);
             }
 
+            // Copy reader so we can double parse to get type
+            Utf8JsonReader copyReader = reader;
+            TransactionType type = JsonSerializer.Deserialize<TransactionType>(ref copyReader, options);
+
+            TxType discriminator = (TxType)(type.Type ?? (ulong)DefaultTxType);
```
LukaszRozmej, line 93: Can we instead of deserializing and throwing away result,  just find the discriminator property?

LukaszRozmej: Ah it kind of does due to `TransactionType` being a simple class with one field only, maybe change it to struct?

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
emlautarom1: I think we do. I tried to do the following to avoid the allocation:

```c#
if (reader.ValueIsEscaped)
{
    reader.Skip();
    var jsonObject = new Utf8JsonReader(reader.ValueSpan);
    return JsonSerializer.Deserialize<Player>(ref jsonObject, options);
}
```

We "skip" the string but we try to deserialize the bytes that correspond to it. This fails in the object deserialization due to the quotes still being escaped as `\u0022` (see https://github.com/dotnet/runtime/issues/76130 for a similar issue).
It seems like the solution is either:
- Prevent the escaping at the origin
- Do a double parsing, first as a string, then as an object.

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
emlautarom1: Changed `TransactionType` to struct

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
LukaszRozmej: Does the escaping thing is only because of how we double serialize params in the test? I think we can fix that.

emlautarom1: That seems to be the case. I've made some changes to our testing code to avoid this situation, but the solution is a bit hacky in my opinion.

src/Nethermind/Nethermind.JsonRpc.Test/RpcTest.cs
```diff
@@ -63,10 +60,26 @@ private static IJsonRpcService BuildRpcService<T>(T module) where T : class, IRp
...
+                sb.Append(param);
+            }
+            else
+            {
+                sb.Append(JsonSerializer.Serialize(param));
+            }
+            sb.Append(',');
+        }
+        sb.Append(']');
+        var doc = JsonDocument.Parse(sb.ToString(), new JsonDocumentOptions { AllowTrailingCommas = true });
```
emlautarom1, line 80: To avoid a double serialization we skip JSON objects (strings starting with "{"). The solution is kind of hacky since we're constructing a JSON array manually by appending strings but it works.

A proper solution would be to accept `params object?[]?` as argument and perform the serialization here with a provided serializer (ie. `EthereumJsonSerializer`). This would avoid the double serialization and would free unit tests from needing to serialize parameters.

I did not implement the "proper" solution since there are +120 tests that use this code passing already serialized parameters and changing them would make this PR too big.

src/Nethermind/Nethermind.JsonRpc.Test/RpcTest.cs
```diff
@@ -63,10 +60,26 @@ private static IJsonRpcService BuildRpcService<T>(T module) where T : class, IRp
...
+            {
+                sb.Append(param);
+            }
+            else
+            {
+                sb.Append(JsonSerializer.Serialize(param));
+            }
+            sb.Append(',');
+        }
+        sb.Append(']');
```
LukaszRozmej, line 79: Alternatively maybe parameters, should be objects, that we would just set in `Params` property and they would get serialised?

src/Nethermind/Nethermind.JsonRpc.Test/RpcTest.cs
LukaszRozmej: Can we try proper version in a PR to this PR?

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
```diff
@@ -280,8 +281,12 @@ public ResultWrapper<byte[]> eth_sign(Address addressData, byte[] message)
 
     public virtual Task<ResultWrapper<Hash256>> eth_sendTransaction(TransactionForRpc rpcTx)
     {
-        Transaction tx = rpcTx.ToTransactionWithDefaults(_blockchainBridge.GetChainId());
-        TxHandlingOptions options = rpcTx.Nonce is null ? TxHandlingOptions.ManagedNonce : TxHandlingOptions.None;
+        Transaction tx = rpcTx.ToTransaction();
```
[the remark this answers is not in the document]
emlautarom1: If we do that then we cannot override in `OptimismEthRpcModule`. We could not override and make the Optimism module use `OptimismTransactionForRpc` but I'm not sure what method is going to be picked up. Need to double check

src/Nethermind/Nethermind.JsonRpc.Test/RpcTest.cs
emlautarom1: PR: https://github.com/NethermindEth/nethermind/pull/7547

LukaszRozmej (approved):

src/Nethermind/Nethermind.JsonRpc.Test/RpcTest.cs
emlautarom1: Solved by passing `object?[]?` as `parameters`.

Nikki601 (approved):

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/AuthorizationListForRpc.cs
```diff
@@ -0,0 +1,86 @@
...
+    private class Tuple
+    {
+        public UInt256 ChainId { get; set; }
+        public ulong Nonce { get; set; }
+        public Address Address { get; set; }
+
+        // TODO: Are these fields actually nullable
+        public UInt256? YParity { get; set; }
+        public UInt256? S { get; set; }
+        public UInt256? R { get; set; }
```
emlautarom1, line 37: @ak88 are these fields actually nullable or can we just default them to `0`?

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/AuthorizationListForRpc.cs
```diff
@@ -0,0 +1,86 @@
...
+                    new UInt256(tuple.AuthoritySignature.S),
+                    new UInt256(tuple.AuthoritySignature.R))));
+
+    public AuthorizationTuple[] ToAuthorizationList() => _tuples
+        .Select(tuple => new AuthorizationTuple(
+            (ulong)tuple.ChainId,
+            tuple.Address,
+            tuple.Nonce,
+            // TODO: This might be wrong
+            new Signature(tuple.R ?? 0, tuple.S ?? 0, (ulong)(tuple.YParity ?? 0)))
```
emlautarom1, line 70: @ak88 In the 7702 PR (https://github.com/NethermindEth/nethermind/pull/7459) there was no code that took a `TransactionForRpc` with a valid `AuthorizationList` and constructed a `Transaction` (we were never copying the `AuthorizationList` property).

This PR adds such behavior but I'm not sure if this exact line is correct.

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/AuthorizationListForRpc.cs
ak88: No they cannot be null

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/AuthorizationListForRpc.cs
ak88: We can drop the nullability

ak88 (approved):

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/SetCodeTransactionForRpc.cs
```diff
@@ -0,0 +1,56 @@
...
+
+public class SetCodeTransactionForRpc : EIP1559TransactionForRpc, IFromTransaction<SetCodeTransactionForRpc>
+{
+    public new static TxType TxType => TxType.SetCode;
+
+    public override TxType? Type => TxType;
+
+    public AuthorizationListForRpc AuthorizationList { get; set; }
+
+    #region Deprecated fields
```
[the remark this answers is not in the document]
ak88: Isnt that because there is an unsigned and signed version?

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/AuthorizationListForRpc.cs
emlautarom1: Removed nullability.

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/AuthorizationListForRpc.cs
emlautarom1: Nullability dropped, we'll revisit as needed

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/SetCodeTransactionForRpc.cs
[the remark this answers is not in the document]
emlautarom1: You're right, my mistake. Fixed.

src/Nethermind/Nethermind.JsonRpc.Test/JsonRpcSocketsClientTests.cs
```diff
@@ -80,7 +80,7 @@ static async Task<int> CountNumberOfMessages(Socket socket, CancellationToken to
                     byte[] buffer = new byte[10];
                     while (true)
                     {
-                        ReceiveResult? result = await stream.ReceiveAsync(buffer);
+                        ReceiveResult? result = await stream.ReceiveAsync(buffer).ConfigureAwait(false);
```
rubo, line 83: @emlautarom1 Why `ConfigureAwait(false)`?

emlautarom1: Change introduced by @benaadams

PR #7485, opened 2024-09-25
Refactor/Use DI for Synchronizer

asdacap:
- Add @smartprogrammer93 idea and apply it to Synchronizer.
- This uses .Net Core's DI and reflection to provide access to `INethermind*Api` property. This is by using a method that provide an `IServiceCollection` with all non-null property passed in as instance. No other change to the initialization code. Also means there are no lazy loading or can use the exact same initialization code in units test.
- Note, it passes in `IServiceCollection` instead of `IServiceProvider` because you can't extend/modify an existing `IServiceProvider`, and this interface is meant for the "module" so to speak, or the wiring codes. 
  - I guess you can manually forward the dependencies one by one, but then what if a plugin's custom component need an extra dependency to forward do? Its just a mess. 
  - This also means that you can make multiple `IServiceProvider` from the same `IServiceCollection` which creates a..... somewhat scope support.
  - This is also the solution for how to make the same "set" of type of different configuration. Clone/Modify IServiceCollection and make different `IServiceProvider`, or a tree of `IServiceProvider`.. manually.
- To put it in a different view, everything is the same, but this particular module wires itself up with automatic dependency injection.
- There are no concrete ideas on how to use this technique to the application as a whole, but hard to deal with components can start wiring things internally with DI. 

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Can sync.


src/Nethermind/Nethermind.Synchronization.Test/OldStyleFullSynchronizerTests.cs
```diff
@@ -61,31 +63,30 @@ public async Task Setup()
...
+                .AddSingleton<IPivot>(pivot)
+                .AddSingleton(Substitute.For<IProcessExitSource>())
+                .AddSingleton<IBetterPeerStrategy>(bestPeerStrategy)
+                .AddSingleton(new ChainSpec())
+                .AddSingleton(stateReader)
+                .AddSingleton<IBeaconSyncStrategy>(No.BeaconSync)
+                .AddSingleton<ILogManager>(LimboLogs.Instance);
+            dbProvider.ConfigureServiceCollection(serviceCollection);
+
+            _synchronizer = new Synchronizer(serviceCollection, syncConfig);
```
emlautarom1, line 88: Test code shows that now `Synchronizer` is completely opaque: there is no way to know that we have to build a `IServiceCollection` with those specific services without looking into the implementation.

emlautarom1 (commented):
Personally I'm a big fan of DI and I think several issues that we've faced (ex. recently after the `TxDecoder` refactor where we broke Optimism due to the usage of different instances rather than a single, global, coherent instance) could be completely eliminated.

Nevertheless, I don't think the proposed approach in this PR addresses any particular issue but rather it hides them. One of the observations is that with this approach "hard to deal with components can start wiring things internally with DI" but it could be argued that components should not wire themselves, rather they should ask for the required services and let the infrastructure provide them.

To be concrete, this approach relies on passing what is essentially a service locator to the constructor of the `Synchronizer` and let it acquire any dependency that it requires, which makes the class responsible for:
- Handling lifetimes: this was already the status quo so it's no better in that regard.
- Fetching required dependencies: dependencies are declared at runtime through `GetRequiredService` instead of statically as constructor parameters, deleting all information about what does this class require to function.

If there are any "hard" components like `Synchronizer` then I think we should ask ourselves if the component itself is the problem, that is, maybe is doing too much or making too many assumptions, rather than trying to hide its dependencies under the rug.

I would like to hear the opinion of the rest of the team but currently I'm not convinced that this change will have a positive effect on the codebase.

asdacap:
The problem with Synchronizer is that it does both the building and configuration of its component along with running. Its not exactly a component, its a module. Properly it should be `Synchronizer(ISyncFeed<BodiesSyncBatch>, **)` and the way to construct the sync feeds are declared somewhere else accounting for plugins. As for the test code, ideally there should be a common "ConfigureTestServiceCollection", or really "ConfigureApplicationServiceCollection" followed with "ConfigureTestEnvironment" somewhere. 

This PR does not attempt to do things properly. Its just way too much change and risk. I tried several times already, and we still don't have DI. IMO although not much, this PR does improve the codebase as it allow other PR to start using DI little by little and we can start doing things properly without introducing several large change. Also, it remove 2 files.

asdacap:
I'll close this in favour of #7500

PR #7486, opened 2024-09-25
Add github action to update the no-op-plugin branch

swapnilraj:
Creating a branch without the op plugin so that we can have a nice diff view of the OP plugin

example deployment of the website: https://nethermindeth.github.io/op-stack-diff/

The github action in https://github.com/NethermindEth/op-stack-diff/blob/main/.github/workflows/op-stack-diff.yml will update the fork html everyday


MarekM25 (approved):

MarekM25:
Some of the code is in the main codebase already, but overall, it’s a good idea

smartprogrammer93 (approved):

.github/workflows/update-no-op-plugin.yml
```diff
@@ -0,0 +1,38 @@
...
+    branches:
+      - master
+
+jobs:
+  remove-op-plugin:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
```
rubo, line 14: We need to use the latest version to avoid warnings:

```suggestion
      - name: Check out repository
        uses: actions/checkout@v4
```

swapnilraj: thanks

rubo (approved):

PR #7493, opened 2024-09-26
Add schema to configs

flcl42:
## Changes

- Add schema to chain configs, so it autocompletes

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [x] Other: Dev experience

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

####

Out of scope:
- schema for the chainspecs
- test configs for correctness according to the schema


LukaszRozmej (approved):
Should we regenerate schema automatically on some build actions? Maybe the same one that updates pivots?

flcl42:
> Should we regenerate schema automatically on some build actions? Maybe the same one that updates pivots?

Yeah, a good idea. And we can push it to a repo, change link to absolute github url, so it will become accessible everywhere outside the solution

rubo:
- There are build errors because of the PowerShell script
- Also, The requirement of the PowerShell fails the build for those who don't have it

PR #7496, opened 2024-09-26
Hotfix multiple instances of txdecoder usage

flcl42:
It's a hotfix that assumes another iteration of full clean up

## Changes

- Use the same TxDecoder everywhere

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Crypto/TransactionExtensions.cs
+++ b/src/Nethermind/Nethermind.Crypto/TransactionExtensions.cs
@@ -9,12 +9,10 @@ namespace Nethermind.Crypto
 {
     public static class TransactionExtensions
     {
-        private static readonly TxDecoder _txDecoder = TxDecoder.Instance;
-
         public static Hash256 CalculateHash(this Transaction transaction)
         {
             KeccakRlpStream stream = new();
-            _txDecoder.Encode(stream, transaction, RlpBehaviors.SkipTypedWrapping);
+            TxDecoder.Instance.Encode(stream, transaction, RlpBehaviors.SkipTypedWrapping);
             return stream.GetHash();
         }
     }
--- a/src/Nethermind/Nethermind.Crypto/TransactionExtensions.cs
+++ b/src/Nethermind/Nethermind.Crypto/TransactionExtensions.cs
@@ -4,15 +4,17 @@
 using Nethermind.Core;
 using Nethermind.Core.Crypto;
 using Nethermind.Serialization.Rlp;
+using System;
 
 namespace Nethermind.Crypto
 {
     public static class TransactionExtensions
     {
+        private static readonly Lazy<TxDecoder> _txDecoder = new(() => TxDecoder.Instance);
         public static Hash256 CalculateHash(this Transaction transaction)
         {
             KeccakRlpStream stream = new();
-            TxDecoder.Instance.Encode(stream, transaction, RlpBehaviors.SkipTypedWrapping);
+            _txDecoder.Value.Encode(stream, transaction, RlpBehaviors.SkipTypedWrapping);
             return stream.GetHash();
         }
     }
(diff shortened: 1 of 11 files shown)


src/Nethermind/Nethermind.JsonRpc/Modules/Trace/TraceRpcModule.cs
```diff
@@ -109,7 +108,7 @@ public ResultWrapper<IEnumerable<ParityTxTraceFromReplay>> trace_callMany(Transa
         /// </summary>
         public ResultWrapper<ParityTxTraceFromReplay> trace_rawTransaction(byte[] data, string[] traceTypes)
         {
-            Transaction tx = _txDecoder.Decode(new RlpStream(data), RlpBehaviors.SkipTypedWrapping);
+            Transaction tx = TxDecoder.Instance.Decode(new RlpStream(data), RlpBehaviors.SkipTypedWrapping);
```
emlautarom1, line 111: We can leave the field instead of inlining:

```suggestion
            private readonly TxDecoder _txDecoder => TxDecoder.Instance;
            // ...
            Transaction tx = _txDecoder.Decode(new RlpStream(data), RlpBehaviors.SkipTypedWrapping);
```

src/Nethermind/Nethermind.Crypto/TransactionExtensions.cs
```diff
@@ -9,12 +9,10 @@ namespace Nethermind.Crypto
 {
     public static class TransactionExtensions
     {
-        private static readonly TxDecoder _txDecoder = TxDecoder.Instance;
-
         public static Hash256 CalculateHash(this Transaction transaction)
         {
             KeccakRlpStream stream = new();
-            _txDecoder.Encode(stream, transaction, RlpBehaviors.SkipTypedWrapping);
+            TxDecoder.Instance.Encode(stream, transaction, RlpBehaviors.SkipTypedWrapping);
```
emlautarom1, line 15: Same as in `TraceRpcModule`

emlautarom1 (approved):
Good catch. Static initialization is a pain to deal with :disappointed:

src/Nethermind/Nethermind.JsonRpc/Modules/Trace/TraceRpcModule.cs
flcl42: Yes, and I switched all to lazy, looks better with it imo

src/Nethermind/Nethermind.Crypto/TransactionExtensions.cs
```diff
@@ -4,17 +4,17 @@
 using Nethermind.Core;
 using Nethermind.Core.Crypto;
 using Nethermind.Serialization.Rlp;
+using System;
 
 namespace Nethermind.Crypto
 {
     public static class TransactionExtensions
     {
-        private static readonly TxDecoder _txDecoder = TxDecoder.Instance;
```
benaadams, line 12: Couldn't it just be a property instead?
```
private static TxDecoder _txDecoder => TxDecoder.Instance;
```
Or even not exist and just use `TxDecoder.Instance` directly?

flcl42: I did it in such way, but it looked not very performant

benaadams (approved):
I think it would be done better than lots of Lazy<T>; but as a hotfix is ok

LukaszRozmej (commented):
Wouldn't it be better to change `Rlp.RegisterDecoders` to use `Instance` field/property first if exist instead of using constructor? Few lines of code.

Or use `[SkipGlobalRegistration]` on TxDecoder and register it from static constructor.

emlautarom1:
> Wouldn't it be better to change `Rlp.RegisterDecoders` to use `Instance` field/property first if exist instead of using constructor? Few lines of code.

This could work but we have several classes that expose this singleton instance, and each could have it's own required initialization order. That is, I don't think it's a solution that would "just work".

> Or use `[SkipGlobalRegistration]` on `TxDecoder` and register it from static constructor.

Implemented a fix using this approach in https://github.com/NethermindEth/nethermind/pull/7497. This also allows us to remove the public constructor and some workarounds due to static initialization order.

LukaszRozmej:
lets revert this ASAP

PR #7504, opened 2024-09-28
Switch to NUnit v4

rubo:
## Changes

Migrated to NUnit v4:

- Replaced `Assert.<Method>()` assertions with `Assert.That()` assertions
- Replaced `CollectionAssert` assertions with `Assert` assertions
- Replaced `TestContext.WriteLine` with`TestContext.Out.Write`
- Replaced obsolete `TimeoutAttribute` with `MaxTimeAttribute`

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

## Remarks

With this upgrade, the FluentAssertions package doesn't seem needed.


LukaszRozmej (approved):

LukaszRozmej:
Will `CancelAfter` fail the test?

rubo:
> Will `CancelAfter` fail the test?

@LukaszRozmej That's a good question. It won't. Actually, as the `Timeout` is gone, there are 2 other options—`MaxTime` and `CancelAfter`. `MaxTime` fails the test if the test exceeds the time limit. But it lets the test run as long as it takes and then checks the actual duration. `CancelAfter` is for canceling long-running tasks with a cancellation token, but most of our test APIs don't support that.
With this said, we better use `MaxTime` instead of `CancelAfter`.

LukaszRozmej:
Maybe we should add both?

rubo:
> Maybe we should add both?

In some cases, yes. But our API doesn't support it. Modifying it would bring many cascading changes, which I'd like to avoid in this PR. Another issue is overkill with timeouts. [Check this out](https://github.com/NethermindEth/nethermind/blob/master/src/Nethermind/Nethermind.Blockchain.Test/Consensus/CompositeTxSourceTests.cs#L19), for instance. Why do we need a timeout there? There are a bunch of such examples.

PR #7509, opened 2024-09-29
Reuse ClassLoggers

benaadams:
## Changes

- Reuse ClassLoggers via ConcurrentDictionary like Typed loggers
- Uses `[CallerFilePath]` for a compile time lookup to avoid using StackTrace to stack walk

- [x] Optimization

#### Requires testing

- [x] No


LukaszRozmej (approved):
So we don't want to have logger per instance?

benaadams:
> So we don't want to have logger per instance?

Doesn't really matter; is a logger per type that we want. The typed version `GetClassLogger<T>()` already has this cache on Type so is just adding it for string also

PR #7513, opened 2024-09-29
Disable txpool if SequencerUrl is set

benaadams:
## Changes

- Disable txpool if SequencerUrl is set or it does a bunch of busy work on NewHead/HeadChange, AddPeer

- [x] Optimization

#### Requires testing

- [x] No


src/Nethermind/Nethermind.Optimism/Rpc/RegisterOptimismRpcModules.cs
```diff
@@ -54,6 +54,11 @@ protected override void RegisterEthRpcModule(IRpcModuleProvider rpcModuleProvide
             _logger.Warn($"SequencerUrl is not set. Nethermind will behave as a Sequencer");
         }
 
+        if (_config.SequencerUrl is not null)
+        {
+            _api.TxPool.Disable();
+        }
+
```
LukaszRozmej, line 61: Maybe better not to create it in the first place or create some NullTxPool?

LukaszRozmej: There is already `InitializeBlockchainOptimism.cs` so it is easy to override there

LukaszRozmej: We do have a `NullTxPool`

benaadams:
Looks good

LukaszRozmej (approved):

src/Nethermind/Nethermind.Optimism/Rpc/RegisterOptimismRpcModules.cs
```diff
@@ -51,13 +51,12 @@ protected override void RegisterEthRpcModule(IRpcModuleProvider rpcModuleProvide
...
         if (_config.SequencerUrl is null && _logger.IsWarn)
         {
-            _logger.Warn($"SequencerUrl is not set. Nethermind will behave as a Sequencer");
+            _logger.Warn("SequencerUrl is not set. Nethermind will behave as a Sequencer");
         }
 
-        BasicJsonRpcClient? sequencerJsonRpcClient = _config.SequencerUrl is null
-            ? null
-            : new(new Uri(_config.SequencerUrl), _api.EthereumJsonSerializer, _api.LogManager);
-        ModuleFactoryBase<IEthRpcModule> ethModuleFactory = CreateEthModuleFactory();
```
jmederosalvarado, line 60: why delete the ModuleFactoryBase creation?

jmederosalvarado (commented):
Overall good. I left a question. Let's make sure we run a node and send it a tx before merging.

LukaszRozmej:
> Overall good. I left a question. Let's make sure we run a node and send it a tx before merging.

I think the tx will be ignored as the tx pool won't do anything with it

benaadams:
> > Overall good. I left a question. Let's make sure we run a node and send it a tx before merging.
> 
> I think the tx will be ignored as the tx pool won't do anything with it

Should forward it to the sequencerUrl

benaadams:
> > Overall good. I left a question. Let's make sure we run a node and send it a tx before merging.
> 
> I think the tx will be ignored as the tx pool won't do anything with it

Should forward it to the sequencerUrl; which is why no point in having the TxPool doing work on every new head

emlautarom1 (approved):

emlautarom1:
Tested in OP-Sepolia through `eth_sendRawTransaction` and I'm getting responses from the sequencer as expected.

Issue #6222, opened 2023-10-25
Optimism plugin: Refactoring chainspec


deffrian:
Refactoring of chain spec specific parts to be injectable by plugins.
Nethermind.Specs/ChainSpecStyle/OptimismParameters.cs

kamilchodola:
@deffrian To be closed?

PR #7540, opened 2024-10-02
Make chainspec extendable by plugins 

deffrian:
Fixes Closes Resolves #
#6222 
Right now our ChainSpec is a mess. This pr tries to fix some issues.

## Changes
Adds new interface `IChainSpecEngineParameters`. This works similar to `IConfig`. 
1. In runtime we search for all inheritors of this interface and create instances. We set corresponding fields from chainspec file.
2. During `ReleaseSpec` creation we call `IChainSpecEngineParameters.AddTransitions`. Plugin devs write implementation of this function if they want to add transitions.
3. After all `ReleaseSpec` created we call `IChainSpecEngineParameters.AdjustReleaseSpec`. Plugin devs write implementation of this function if they want to change any parameters in ReleaseSpec.

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


src/Nethermind/Nethermind.Optimism/OPConfigHelper.cs
```diff
@@ -6,16 +6,16 @@
 
 namespace Nethermind.Optimism;
 
-public class OptimismSpecHelper(OptimismParameters parameters) : IOptimismSpecHelper
+public class OptimismSpecHelper(OptimismChainSpecEngineParameters parameters) : IOptimismSpecHelper
 {
-    private readonly long _bedrockBlockNumber = parameters.BedrockBlockNumber;
-    private readonly ulong _regolithTimestamp = parameters.RegolithTimestamp;
+    private readonly long? _bedrockBlockNumber = parameters.BedrockBlockNumber;
+    private readonly ulong? _regolithTimestamp = parameters.RegolithTimestamp;
```
LukaszRozmej, line 12: Why do we have to make stuff nullable?

src/Nethermind/Nethermind.Optimism/OptimismChainSpecEngineParameters.cs
```diff
@@ -0,0 +1,62 @@
...
+        ArgumentNullException.ThrowIfNull(BedrockBlockNumber);
+        ArgumentNullException.ThrowIfNull(CanyonTimestamp);
+        ArgumentNullException.ThrowIfNull(EcotoneTimestamp);
+        ArgumentNullException.ThrowIfNull(FjordTimestamp);
+        ArgumentNullException.ThrowIfNull(GraniteTimestamp);
+        ArgumentNullException.ThrowIfNull(L1FeeRecipient);
+        ArgumentNullException.ThrowIfNull(L1BlockAddress);
+        ArgumentNullException.ThrowIfNull(CanyonBaseFeeChangeDenominator);
+        ArgumentNullException.ThrowIfNull(Create2DeployerAddress);
+        ArgumentNullException.ThrowIfNull(Create2DeployerCode);
```
LukaszRozmej, line 51: Should it have some sort of `Validate` method?

src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainSpecParametersProvider.cs
```diff
@@ -0,0 +1,99 @@
...
+    private void InitializeInstances()
+    {
+        Type type = typeof(IChainSpecEngineParameters);
+        IEnumerable<Type> types = TypeDiscovery.FindNethermindBasedTypes(type).Where(x => x.IsClass);
+        foreach (Type @class in types)
+        {
+            string engineName = @class.Name.Remove(@class.Name.Length - EngineParamsSuffix.Length);
+            if (!_chainSpecParameters.ContainsKey(engineName)) continue;
+
+            var instance = (IChainSpecEngineParameters)@class.GetConstructor(Type.EmptyTypes)!.Invoke(Array.Empty<object>());
```
LukaszRozmej, line 72: You can use `Activator` for simplicity

src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainSpecParametersProvider.cs
```diff
@@ -0,0 +1,99 @@
...
+                JsonProperty? jsonProperty = _chainSpecParameters[engineName].EnumerateObject().FirstOrDefault(p =>
+                    string.Compare(p.Name, property.Name, StringComparison.InvariantCultureIgnoreCase) == 0);
+
+                if (jsonProperty is null)
+                {
+                    throw new FormatException($"ChainSpec property {property.Name} is not set");
+                }
+
+                object value = ConfigSourceHelper.ParseValue(property.PropertyType, jsonProperty.Value.Value.ToString(), "chainspec", property.Name);
+                property.SetValue(instance, value);
```
LukaszRozmej, line 86: Can we just use json deserialization here?

src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainSpecParametersProvider.cs
```diff
@@ -0,0 +1,99 @@
...
+            throw new InvalidOperationException("No seal engine in chain spec");
+        }
+
+        return result;
+    }
+
+    private void InitializeInstances()
+    {
+        Type type = typeof(IChainSpecEngineParameters);
+        IEnumerable<Type> types = TypeDiscovery.FindNethermindBasedTypes(type).Where(x => x.IsClass);
```
LukaszRozmej, line 66: Do we want auto-discovery or explicit registration?

src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainSpecParametersProvider.cs
```diff
@@ -0,0 +1,99 @@
...
+        foreach (Type @class in types)
```
LukaszRozmej, line 67: Instances should be constructed lazily?

src/Nethermind/Nethermind.Specs/ChainSpecStyle/IChainSpecEngineParameters.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using System.Collections.Generic;
+
+public interface IChainSpecEngineParameters
+{
+    string? SealEngineType { get; }
+
+    void AddTransitions(SortedSet<long> blockNumbers, SortedSet<ulong> timestamps);
+
+    void AdjustReleaseSpec(ReleaseSpec spec, long startBlock, ulong? startTimestamp);
```
LukaszRozmej, line 14: maybe naming `ApplyTo`?

src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainSpecParametersProvider.cs
```diff
@@ -0,0 +1,99 @@
...
+            foreach (PropertyInfo property in @class.GetProperties(BindingFlags.Public | BindingFlags.Instance))
+            {
+                if (property.Name == "SealEngineType") continue;
+                JsonProperty? jsonProperty = _chainSpecParameters[engineName].EnumerateObject().FirstOrDefault(p =>
+                    string.Compare(p.Name, property.Name, StringComparison.InvariantCultureIgnoreCase) == 0);
+
+                if (jsonProperty is null)
+                {
+                    throw new FormatException($"ChainSpec property {property.Name} is not set");
+                }
```
LukaszRozmej, line 83: Can those be optional?

LukaszRozmej (commented):
I think the direction is good, but there is some details to iron out.

src/Nethermind/Nethermind.Optimism/OPConfigHelper.cs
deffrian: There might be no value in config, so better to make it nullable imo

src/Nethermind/Nethermind.Optimism/OptimismChainSpecEngineParameters.cs
deffrian: Forgot to remove. Was for debug purposes

src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainSpecParametersProvider.cs
emlautarom1: In the Transaction type refactoring we've been working with explicit registration. This should make transitioning to a DI approach easier but makes the immediate code harder since you need to register at the appropriate time, a lot of times directly into `static` fields.

src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainSpecParametersProvider.cs
```diff
@@ -0,0 +1,99 @@
...
+
+    private readonly ConcurrentDictionary<Type, IChainSpecEngineParameters> _instances = new();
+    public string SealEngineType { get; }
+
+    public ChainSpecParametersProvider(Dictionary<string, JsonElement> engineParameters)
+    {
+        foreach (var parameters in engineParameters)
+        {
+            _chainSpecParameters[parameters.Key] = parameters.Value;
+        }
```
emlautarom1, line 33: For shallow copies you can do:

```suggestion
        _chainSpecParameters = new Dictionary<string, JsonElement>(engineParameters, StringComparer.InvariantCultureIgnoreCase);
```

src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainSpecParametersProvider.cs
```diff
@@ -0,0 +1,99 @@
...
+
+public class ChainSpecParametersProvider : IChainSpecParametersProvider
+{
+    // TODO: test that all IChainSpecEngineParameters have this suffix
+    private const string EngineParamsSuffix = "ChainSpecEngineParameters";
+
+    private readonly Dictionary<string, JsonElement> _chainSpecParameters =
+        new(StringComparer.InvariantCultureIgnoreCase);
+
+    private readonly ConcurrentDictionary<Type, IChainSpecEngineParameters> _instances = new();
```
emlautarom1, line 25: Why does this have to be concurrent? We cannot add elements through `AllChainSpecParameters` and there is no other accessor.

src/Nethermind/Nethermind.Specs/ChainSpecStyle/IChainSpecParametersProvider.cs
```diff
@@ -0,0 +1,13 @@
...
+// SPDX-License-Identifier: LGPL-3.0-only
+
+namespace Nethermind.Specs.ChainSpecStyle;
+
+using System.Collections.Generic;
+
+public interface IChainSpecParametersProvider
+{
+    string SealEngineType { get; }
+    ICollection<IChainSpecEngineParameters> AllChainSpecParameters { get; }
```
emlautarom1, line 11: If we don't intend to allow modification (which I think we don't) prefer `IEnumerable` to `ICollection`

src/Nethermind/Nethermind.Optimism/OptimismChainSpecEngineParameters.cs
```diff
@@ -0,0 +1,62 @@
...
+
+    public Address? Create2DeployerAddress { get; set; }
+
+    public byte[]? Create2DeployerCode { get; set; }
+
+    public void AddTransitions(SortedSet<long> blockNumbers, SortedSet<ulong> timestamps)
+    {
+        ArgumentNullException.ThrowIfNull(RegolithTimestamp);
+        ArgumentNullException.ThrowIfNull(BedrockBlockNumber);
+        ArgumentNullException.ThrowIfNull(CanyonTimestamp);
```
emlautarom1, line 43: Some of these fields (like `CanyonTimestamp`) used to be nullable and still are here, but others have changed (like `RegolithTimestamp`).
Which fields are actually required and which are not?

src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainSpecParametersProvider.cs
deffrian: Same argument, do as in config. But I don't mind both tbh

src/Nethermind/Nethermind.Config/ConfigSourceHelper.cs
```diff
@@ -86,13 +103,76 @@ public static object GetDefault(Type type)
...
         {
+            if (Nullable.GetUnderlyingType(valueType) is var nullableType && nullableType is not null)
+            {
+                if (string.IsNullOrEmpty(itemValue) || itemValue.Equals("null", StringComparison.InvariantCultureIgnoreCase))
+                {
+                    return null;
+                }
+
+                return GetValue(nullableType, itemValue);
+            }
```
emlautarom1, line 155: This is pretty much the same check as in `ParseValue`. Any chance to refactor to a common utility?

emlautarom1 (commented):
Some questions regarding nullability and discovery, but overall LGTM.

src/Nethermind/Nethermind.Config/ConfigSourceHelper.cs
```diff
@@ -86,13 +106,76 @@ public static object GetDefault(Type type)
...
         }
 
+        public static bool TryFromHex(Type type, string itemValue, out object value)
+        {
+            if (!itemValue.StartsWith("0x"))
+            {
+                value = null;
+                return false;
+            }
+            switch (Type.GetTypeCode(type))
```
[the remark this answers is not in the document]
deffrian: Dont understand what you mean

src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainSpecParametersProvider.cs
deffrian: Construction is done on setup, so performance is not a concern(it takes small amount of time). I prefer to catch any miss-configuration errors at this point.

src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainSpecParametersProvider.cs
```diff
@@ -0,0 +1,99 @@
...
+        return result;
+    }
+
+    private void InitializeInstances()
+    {
+        Type type = typeof(IChainSpecEngineParameters);
+        IEnumerable<Type> types = TypeDiscovery.FindNethermindBasedTypes(type).Where(x => x.IsClass);
+        foreach (Type @class in types)
+        {
+            string engineName = @class.Name.Remove(@class.Name.Length - EngineParamsSuffix.Length);
```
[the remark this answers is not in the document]
deffrian: Not sure what you mean. It already looks for all inheritors of `IChainSpecEngineParameters`. I didn't make manual registration because ChainSpec is more similar to `IConfig`. So I thought we should behave the same way here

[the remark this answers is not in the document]
emlautarom1: My question was more in line with replacing the class name logic with something more "typesafe": right now it seems like I need to follow a specific naming convention but it's not enforced statically, and to figure it out I need to take a look into this implementation.

If I were to add another `IChainSpecEngineParameters` class, what would the requirements be and how would I figure them out?

src/Nethermind/Nethermind.Optimism/OPConfigHelper.cs
jmederosalvarado: @deffrian shouldn't we always have a value for these hardforks?

src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainSpecParametersProvider.cs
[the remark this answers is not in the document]
jmederosalvarado: I think what Lautaro means is that we should have a property `IChainSpecEngineParamerters.Name` or something similar, instead of using the name of the class. I understand @deffrian's point that we're being consistent with how `IConfig` works. But maybe we should start improving both.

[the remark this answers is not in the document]
emlautarom1: > I think what Lautaro means is that we should have a property `IChainSpecEngineParamerters.Name` or something similar, instead of using the name of the class.

Precisely, I just didn't want to push for any particular direction.

> we're being consistent with how `IConfig` works

Didn't know that we were using the class name approach there. If it is consistent with the existing style then let's move forward, but we might want to add a TODO to eventually improve (that is, move to the type system) these kind of requirements.

src/Nethermind/Nethermind.Consensus.AuRa/InitializationSteps/StartBlockProducerAuRa.cs
```diff
@@ -276,7 +277,8 @@ private ITxSource CreateTxSourceForProducer(ITxSource? additionalTxSource)
     {
         bool CheckAddPosdaoTransactions(IList<ITxSource> list, long auRaPosdaoTransition)
         {
-            if (auRaPosdaoTransition < AuRaParameters.TransitionDisabled && _validator is ITxSource validatorSource)
+            const long transitionDisabled = long.MaxValue;
+            if (auRaPosdaoTransition < transitionDisabled && _validator is ITxSource validatorSource)
```
emlautarom1, line 281:
```suggestion
            if (auRaPosdaoTransition != transitionDisabled && _validator is ITxSource validatorSource)
```

src/Nethermind/Nethermind.AuRa.Test/ChainSpecLoaderTest.cs
```diff
@@ -0,0 +1,94 @@
...
+            .Should().Be("0xb97036A26259B7147018913bD58a774cf91acf25");
+
+        chainSpec.ShanghaiTimestamp.Should().Be(ChiadoSpecProvider.ShanghaiTimestamp);
+        chainSpec.ShanghaiTimestamp.Should().Be(ChiadoSpecProvider.Instance.TimestampFork);
+    }
+
+    [Test]
+    public void Can_load_posdao_with_rewriteBytecode()
+    {
+        // TODO: modexp 2565
```
emlautarom1, line 76: Context on this TODO?

src/Nethermind/Nethermind.Consensus.AuRa/Validators/AuRaParameters.cs
```diff
@@ -0,0 +1,73 @@
...
+{
+    public enum ValidatorType
+    {
+        List,
+        Contract,
+        ReportingContract,
+        Multi
+    }
+
+    public class Validator
```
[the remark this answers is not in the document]
deffrian: I copy-pasted this code. This pr is already pretty big. Imo it's not good idea to refactor this also

src/Nethermind/Nethermind.Consensus.Ethash/EthashChainSpecEngineParameters.cs
```diff
@@ -0,0 +1,220 @@
+// SPDX-FileCopyrightText: 2024 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System;
+using System.Buffers;
+using System.Collections.Generic;
+using System.Globalization;
```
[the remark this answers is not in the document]
deffrian: For comment on top: This is chainSpec, so better to force people to set 0 explicitly so they don't missconfigure it

src/Nethermind/Nethermind.AuRa.Test/ChainSpecLoaderTest.cs
deffrian: Don't know. This is from the original test

src/Nethermind/Nethermind.Consensus.Ethash/EthashChainSpecEngineParameters.cs
```diff
@@ -0,0 +1,158 @@
...
+    public void ApplyToChainSpec(ChainSpec chainSpec)
+    {
+        chainSpec.MuirGlacierNumber = DifficultyBombDelays?.Keys.Skip(2).FirstOrDefault();
+        chainSpec.ArrowGlacierBlockNumber = DifficultyBombDelays?.Keys.Skip(4).FirstOrDefault();
+        chainSpec.GrayGlacierBlockNumber = DifficultyBombDelays?.Keys.Skip(5).FirstOrDefault();
+        chainSpec.HomesteadBlockNumber = HomesteadTransition;
+        chainSpec.DaoForkBlockNumber = DaoHardforkTransition;
+    }
+
+    private class DifficultyBombDelaysJsonConverter : JsonConverter<IDictionary<long, long>>
```
LukaszRozmej, line 106: We probably don't need it

src/Nethermind/Nethermind.Core/SealEngineType.cs
```diff
@@ -8,10 +8,8 @@ public static class SealEngineType
         public const string None = nameof(None);
         public const string AuRa = nameof(AuRa);
         public const string Clique = nameof(Clique);
-        public const string NethDev = nameof(NethDev);
         public const string Ethash = nameof(Ethash);
         public const string BeaconChain = nameof(BeaconChain);
-        public const string Optimism = nameof(Optimism);
```
LukaszRozmej, line 14: I think we could keep them there for simplicity.

src/Nethermind/Nethermind.Optimism/OptimismSynchronizerModule.cs
```diff
@@ -27,7 +27,7 @@ protected override void Load(ContainerBuilder builder)
             builder.AddSingleton<ITotalDifficultyStrategy>(
                 new FixedTotalDifficultyStrategy(
                     new CumulativeTotalDifficultyStrategy(),
-                    fixesBlockNumber: parameters.BedrockBlockNumber - 1,
+                    fixesBlockNumber: parameters.BedrockBlockNumber!.Value - 1,
```
LukaszRozmej, line 30: if it is required maybe it shouldn't be nullable?

src/Nethermind/Nethermind.Runner.Test/EthereumRunnerTests.cs
```diff
@@ -44,6 +48,10 @@ static EthereumRunnerTests()
 
     private static ICollection InitOnce()
     {
+        // we need this to discover ChainSpecEngineParameters
+        new CliqueConfig();
+        new OptimismConfig();
+        new TaikoPlugin();
```
LukaszRozmej, line 54: Can we do it in some more elegant way?

LukaszRozmej (commented):
Almost done, can you check if we need those custom dictionary converters?

src/Nethermind/Nethermind.Optimism/OptimismSynchronizerModule.cs
deffrian: I think it's good to make everything nullable. So it will not set zero by default and we will not forget to set it when it's needed

src/Nethermind/Nethermind.Consensus.AuRa/Config/AuRaChainSpecEngineParameters.cs
```diff
@@ -0,0 +1,191 @@
...
+                }
+            }
+            else
+            {
+                throw new ArgumentException("Cannot deserialize BlockReward.");
+            }
+
+            return value;
+        }
+    }
```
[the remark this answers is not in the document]
deffrian: This one is needed

src/Nethermind/Nethermind.Specs/ChainSpecStyle/Json/LongUInt256DictionaryConverter.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Buffers;
+using System.Collections.Generic;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Nethermind.Int256;
+using Nethermind.Serialization.Json;
+
+namespace Nethermind.Specs.ChainSpecStyle.Json;
+
+public class LongUInt256DictionaryConverter : JsonConverter<SortedDictionary<long, UInt256>>
```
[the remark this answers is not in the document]
deffrian: This one also

src/Nethermind/Nethermind.Runner.Test/EthereumRunnerTests.cs
deffrian: This is the only way I know

src/Nethermind/Nethermind.Optimism/OptimismSynchronizerModule.cs
LukaszRozmej: ok but handle `null` in elegant way then

src/Nethermind/Nethermind.Consensus.AuRa/Config/AuRaChainSpecEngineParameters.cs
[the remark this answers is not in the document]
LukaszRozmej: Why? We should be able to deserialize Dictionaries without special converters

src/Nethermind/Nethermind.Runner.Test/EthereumRunnerTests.cs
LukaszRozmej: what are we trying to achieve here?

src/Nethermind/Nethermind.Consensus.AuRa/Config/AuRaChainSpecEngineParameters.cs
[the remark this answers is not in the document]
deffrian: We actually need all of them because we cannot convert dictionary with hex encoded keys

src/Nethermind/Nethermind.Runner.Test/EthereumRunnerTests.cs
deffrian: My assumption is that .net is not loading plugins if we don't use them. If you remove this it will not discover `ChainSpecEngineParameters`

src/Nethermind/Nethermind.Consensus.AuRa/Config/AuRaChainSpecEngineParameters.cs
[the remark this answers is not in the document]
LukaszRozmej: sure we can, in `LongConverter` we need to override `ReadAsPropertyName` (and maybe `WriteAsPropertyName` too) similar as we have it in `UInt256Converter`

src/Nethermind/Nethermind.Optimism/OptimismSynchronizerModule.cs
LukaszRozmej: Can we use `TypeDiscovery` class for that?

src/Nethermind/Nethermind.Runner.Test/EthereumRunnerTests.cs
LukaszRozmej: Can we use `TypeDiscovery` class for that?

LukaszRozmej (approved):

PR #7542, opened 2024-10-03
Add and parallelise AccessList as part of address prewarming

benaadams:
## Changes

- Introduce a concurrent safe path for address and storage prewarming where no (non concurrent) storage caches are touched
- In addition prewarm access lists via this path
- Fan out and parallelise sub accesslist storages when there are a lot of them; so as not to cause head of line blocking and seralize prewarming by a large transaction

Resolves a lot of heavy access block issues on higher gas chains (e.g. mainnet) as they mostly use AccessLists; but doesn't completely solve it as on lower gas chains (e.g. Base) access lists aren't as heavily used.

- [x] Optimization

#### Requires testing

- [x] Yes

- [x] No - requires sync and head running testing (in progress)

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
@@ -91,6 +91,7 @@ public sealed class BlockCachePreWarmer(ReadOnlyTxProcessingEnvFactory envFactor
                 IReadOnlyTxProcessorSource env = _envPool.Get();
                 try
                 {
+                    // Only need one scope as we are taking a parallel safe path
                     using IReadOnlyTxProcessingScope scope = env.Build(stateRoot);
                     int progress = 0;
                     Parallel.For(0, block.Withdrawals.Length, parallelOptions,
--- a/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
@@ -88,25 +88,25 @@ public sealed class BlockCachePreWarmer(ReadOnlyTxProcessingEnvFactory envFactor
         {
             if (spec.WithdrawalsEnabled && block.Withdrawals is not null)
             {
-                int progress = 0;
-                Parallel.For(0, block.Withdrawals.Length, parallelOptions,
-                    _ =>
-                    {
-                        IReadOnlyTxProcessorSource env = _envPool.Get();
-                        int i = 0;
-                        try
+                IReadOnlyTxProcessorSource env = _envPool.Get();
+                try
+                {
+                    using IReadOnlyTxProcessingScope scope = env.Build(stateRoot);
+                    int progress = 0;
+                    Parallel.For(0, block.Withdrawals.Length, parallelOptions,
+                        _ =>
                         {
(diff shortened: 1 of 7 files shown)


src/Nethermind/Nethermind.State/WorldState.cs
```diff
@@ -110,22 +111,46 @@ public void Reset(bool resizeCollections = false)
...
+
+                    if (isParallelAccess && storages.Count > 10)
+                    {
+                        // We are already warming in parallel, but fan out if tx with
+                        // very many storages in access list so don't form a single
+                        // threaded queue with all warming waiting on one tx.
+                        Parallel.ForEach(storages, (storage) =>
+                        {
+                            _persistentStorageProvider.WarmUp(new StorageCell(address, storage), tree);
+                        });
```
LukaszRozmej, line 140: Should we do more custom partition here, like N storage cells per task?

src/Nethermind/Nethermind.State/StateProvider.cs
```diff
@@ -657,9 +658,11 @@ public StateProvider(IScopedTrieStore? trieStore,
             };
         }
 
-        public bool WarmUp(Address address)
+        public bool WarmUp(Address address, out Account? account)
         {
-            return GetState(address) is not null;
+            Debug.Assert(_populatePreBlockCache);
```
LukaszRozmej, line 663: shouldn't we check that flag for safety?

src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
```diff
@@ -207,20 +215,14 @@ private void WarmupAddresses(ParallelOptions parallelOptions, Block block)
         {
             if (parallelOptions.CancellationToken.IsCancellationRequested) return;
 
+            IReadOnlyTxProcessorSource env = PreWarmer._envPool.Get();
             try
             {
-                if (SystemTxAccessList is not null)
+                // Only need one scope as we are taking a parallel safe path
```
LukaszRozmej, line 221: Can you explain more, we are doing multi-threaded access to that instance?

LukaszRozmej (approved):

src/Nethermind/Nethermind.State/StateProvider.cs
benaadams: Is only called by prewarmer

LukaszRozmej: Can we make it more hidden?

src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
benaadams: Yes

However now `Warm(Address)` and `Warm(AccessList, , isParallelAccess: true)` only use the ConcurrentDictionary and Trie in readonly mode; so are safe for multi-thread access skipping all data structures in the scope (as they would have been reset anyway)

i.e.
1. is in ConcurrentDict return
2. Get from trie and put in ConcurrentDict

When `Warm(AccessList, , isParallelAccess: false)` like in transaction; it also fills out the data structures in the scope as they are likely needed by the tx

src/Nethermind/Nethermind.State/WorldState.cs
benaadams: Changed; will do some more testing with changes before merging

PR #7545, opened 2024-10-03
Shutter: P2P improvements

Marchhill:
## Changes

- Updated p2p with peer discovery and exchange
- Persistent p2p node id with external ip
- P2P refactoring
- Update configs

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


src/Nethermind/Nethermind.Shutter/Config/IShutterConfig.cs
```diff
@@ -11,14 +12,9 @@ namespace Nethermind.Shutter.Config;
...
-    // todo: replace with bootnodes when peer discovery added
     private const string DefaultP2PAddresses =
-@"/ip4/139.59.130.109/tcp/23003/p2p/12D3KooWRZoofMsnpsjkgvfPQUyGXZQnn7EVnb4tw4ghNfwMnnsj,
-/ip4/167.71.169.248/tcp/23003/p2p/12D3KooWGH3VxoSQXZ6wUuCmsv5caGQnhwfGejbkXH6uS2r7sehA,
-/ip4/139.59.130.109/tcp/23003/p2p/12D3KooWNxTiw7CvD1fuyye5P8qPhKTTrRBW6wwZwMdqdTxjYF2H,
-/ip4/178.128.192.239/tcp/23003/p2p/12D3KooWCdpkipTiuzVMfkV7yLLgqbFeAL8WmEP78hCoBGBYLugN,
-/ip4/45.55.192.248/tcp/23003/p2p/12D3KooWMPuubKqksfMxvLwEBDScaopTdvPLr5J5SMmBEo2zkcMz,
-/ip4/178.128.126.237/tcp/23003/p2p/12D3KooWAg1pGUDAfFWSZftpN3JjBfLUCGLQcZApJHv2VntdMS9U";
+@"/ip4/167.99.177.227/tcp/23003/p2p/12D3KooWD35AESYCttDEi3J5WnQdTFuM5JNtmuXEb1x4eQ28gb1s,
+/ip4/159.89.15.119/tcp/23003/p2p/12D3KooWRzAhgPA16DiBQhiuYoasYzJaQSAbtc5i5FvgTi9ZDQtS";
```
LukaszRozmej, line 17: Should this be in code? Rather than in config files?

src/Nethermind/Nethermind.Shutter/Config/IShutterConfig.cs
```diff
@@ -127,16 +123,13 @@ public void Validate()
...
-            }
-            catch (NotSupportedException)
-            {
-                throw new ArgumentException($"Could not decode Shutter keyper p2p address \"{addr}\".");
-            }
+            bootnodeP2PAddresses = BootnodeP2PAddresses.Select(addr => Multiaddress.Decode(addr)).ToArray();
+        }
+        catch (NotSupportedException)
+        {
+            throw new ArgumentException($"Could not decode Shutter bootnode p2p addresses.");
```
LukaszRozmej, line 132:
```suggestion
        catch (NotSupportedException e)
        {
            throw new ArgumentException($"Could not decode Shutter bootnode p2p addresses.", e);
```

src/Nethermind/Nethermind.Shutter/Config/ShutterConfig.cs
```diff
@@ -7,17 +7,17 @@ public class ShutterConfig : IShutterConfig
...
         public bool Enabled { get; set; }
         public bool Validator { get; set; } = true;
-        public string? SequencerContractAddress { get; set; }
-        public string? ValidatorRegistryContractAddress { get; set; }
-        public string? KeyBroadcastContractAddress { get; set; }
-        public string? KeyperSetManagerContractAddress { get; set; }
+        public string? SequencerContractAddress { get; set; } = "0xc5C4b277277A1A8401E0F039dfC49151bA64DC2E";
+        public string? ValidatorRegistryContractAddress { get; set; } = "0xefCC23E71f6bA9B22C4D28F7588141d44496A6D6";
+        public string? KeyBroadcastContractAddress { get; set; } = "0x626dB87f9a9aC47070016A50e802dd5974341301";
+        public string? KeyperSetManagerContractAddress { get; set; } = "0x7C2337f9bFce19d8970661DA50dE8DD7d3D34abb";
```
LukaszRozmej, line 13: similar, hardcoded vs config?

src/Nethermind/Nethermind.Shutter/ShutterP2P.cs
```diff
@@ -128,19 +135,26 @@ public async ValueTask DisposeAsync()
         await (_cts?.CancelAsync() ?? Task.CompletedTask);
     }
 
-    private class MyProto : IDiscoveryProtocol
+    private Identity GetPeerIdentity(IFileSystem fileSystem, IKeyStoreConfig cfg)
```
LukaszRozmej, line 138: Is this called only once?

Marchhill: yes

src/Nethermind/Nethermind.Shutter/Config/IShutterConfig.cs
ak88: These do not change depending on the network?

Marchhill: Yes, it defaults to gnosis but is overridden for chiado: https://github.com/NethermindEth/nethermind/blob/master/src/Nethermind/Nethermind.Runner/configs/chiado.cfg#L34

ak88: Why not handle it the same way as other config values and have it defined in the config file?

Marchhill: I don't mind either way, maybe it's a bit easier to make it default. I changed it because of this review comment: https://github.com/NethermindEth/nethermind/pull/7414#discussion_r1758950890

LukaszRozmej: Hm.. my comment was do we want it in default `gnosis.cfg` and not as defaults for all config :)

Marchhill: Ah I see what you mean, I think it makes sense to have in `gnosis.cfg` then, since it won't do anything unless the `Shutter.Enabled` flag is true

LukaszRozmej (approved):

flcl42 (approved):

src/Nethermind/Nethermind.Shutter/ShutterApi.cs
```diff
@@ -92,11 +103,11 @@ TimeSpan slotLength
 
         KeyValidator = new ShutterKeyValidator(_cfg, Eon, logManager);
 
-        InitP2P(_cfg, logManager);
+        InitP2P(ip);
```
ak88, line 106: Why not inject `IShutterP2P` instead ?

Marchhill: As in pass `IShutterP2P` to the `ShutterApi` constructor? I guess we could do this, and also for `IShutterEon` and `ShutterTime`. Instead I did all of the initialisation inside the API with virtual functions so the tests can override - this seems fine to me but do you think it's better to inject?

Marchhill: Ok I see, so you think it's ok to leave like this for now?

Marchhill: Going to resolve in separate PR

PR #7564, opened 2024-10-07
Fix null pointer in RecoverSignatures

ak88:
## Changes

Fix for change after review comment. 

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_


MarekM25 (approved):
Can you try to write a test?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Consensus/Processing/RecoverSignature.cs
```diff
@@ -87,7 +87,7 @@ public void RecoverData(Block block)
                     recoverFromEcdsa++;
                 }
 
-                if (tx.HasAuthorizationList)
+                if (poolTx is not null && tx.HasAuthorizationList)
```
LukaszRozmej, line 90: still increment recoverFromEcdsa++; ?

benaadams (approved):

src/Nethermind/Nethermind.Consensus/Processing/RecoverSignature.cs
ak88: We need to trigger the recovery if an authority is missing. We could add
`                        else if(tx.AuthorizationList[i].Authority is null)
`
maybe?

PR #7565, opened 2024-10-07
Remove snapshot; bypass downloading if available locally

flcl42:
## Changes

- Remove the need in snapshot for snap-synced OP mainnet
- Optimize snapshot loading in case it's loaded from local path

- [x] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

### DIFF
--- a/src/Nethermind/Nethermind.Init.Snapshot/InitDatabaseSnapshot.cs
+++ b/src/Nethermind/Nethermind.Init.Snapshot/InitDatabaseSnapshot.cs
@@ -71,44 +71,53 @@ public class InitDatabaseSnapshot : InitDatabase
 
         Directory.CreateDirectory(snapshotConfig.SnapshotDirectory);
 
-        if (GetCheckpoint(snapshotConfig) < Stage.Downloaded)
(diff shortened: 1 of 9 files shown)


benaadams (approved):

jmederosalvarado (approved):

src/Nethermind/Nethermind.Init.Snapshot/InitDatabaseSnapshot.cs
```diff
@@ -71,44 +71,53 @@ private async Task InitDbFromSnapshot(CancellationToken cancellationToken)
 
         Directory.CreateDirectory(snapshotConfig.SnapshotDirectory);
 
-        if (GetCheckpoint(snapshotConfig) < Stage.Downloaded)
+        if (!File.Exists(snapshotUrl))
```
LukaszRozmej, line 74: Url and file? Did you mean `snapshotFileName`? Would be good to abstract file manipulation to some methods instead of dign it directly.

PR #7568, opened 2024-10-07
fix blob index locking

LukaszRozmej:
## Changes

- Move blob removing under main lock.
- Move list to concurrent set so reading blob index can be thread safe without main lock

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

@marcindsobczak ca you write concurrency test that would fail under old version?


LukaszRozmej:
Alternatively we could put `TryGetBlobAndProof` under lock and change to `Dictionary<byte[], HashSet<Hash256>>`

LukaszRozmej:
went with v2

PR #7571, opened 2024-10-08
Perf/skip commit queue

asdacap:
- Turns out, when catching up, the time taking for commit patricia trie is about 20% 
- This PR optimize that a bit by changing the interface for committing and skipping commit queue within PatriciaTrie. 
- Also, because of the interface change, some checks are removed and it can use `ref TreePath` for the path.
- Also parallelize the commit.
- This improve overall block rate by about 5-10%.
- Graph is after, before, after, before.
- Commit storage time need a higher level change to parallelize. Maybe next week.
Screenshot from 2024-10-08 22-57-45

## Changes

- Remove CommitNode/FinishBlockCommit to a BeginCommit which returns a disposable ICommitter.
- PatriciaTrie call ICommitter.CommitNode instead of putting to a queue first

- [X] Optimization

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Still....


src/Nethermind/Nethermind.Trie/PatriciaTree.cs
```diff
@@ -138,64 +138,60 @@ public void Commit(long blockNumber, bool skipRoot = false, WriteFlags writeFlag
...
-                    if (_logger.IsTrace) Trace(blockNumber, node);
-                    TrieStore.CommitNode(blockNumber, node, writeFlags: writeFlags);
+                    maxLevelForConcurrentCommit++; // Another level
                 }
-
-                // reset objects
-                TreePath path = TreePath.Empty;
-                RootRef!.ResolveKey(TrieStore, ref path, true, bufferPool: _bufferPool);
-                SetRootHash(RootRef.Keccak!, true);
             }
```
LukaszRozmej: This is equivalent of:

```
            int maxLevelForConcurrentCommit = _writeBeforeCommit switch
            {
                > 64 * 16 => 1, // we separate at two top levels
                > 64 => 0, // we separate at top level
                _ => -1
            };
```

Sounds a bit too low for me?
Did you want more something like:

```
            int maxLevelForConcurrentCommit = _writeBeforeCommit switch
            {
                > 16 * 16 * 16 => 2, // we separate at three top levels
                > 16 * 16 => 1, // we separate at two top levels
                > 16 => 0, // we separate at top level
                _ => -1
            };
```

Again, regarding unbalanced tries and unbalanced updates, would it be better to have a more dynamic task/path split based on available threads instead of a static path split?

But this would make sense to change only if this is a potential bottleneck. Do you have some profiling sessions?

src/Nethermind/Nethermind.Trie/PatriciaTree.cs
```diff
@@ -138,64 +138,60 @@ public void Commit(long blockNumber, bool skipRoot = false, WriteFlags writeFlag
...
                 ThrowReadOnlyTrieException();
             }
 
             TrieNode node = nodeCommitInfo.Node;
-            TreePath path = nodeCommitInfo.Path;
             if (node!.IsBranch)
             {
-                // idea from EthereumJ - testing parallel branches
-                if (!_parallelBranches || !nodeCommitInfo.IsRoot)
+                if (path.Length > maxLevelForConcurrentCommit)
```
LukaszRozmej, line 182: So here instead of checking path.Length, wouldn't be better to check if we have a free thread to take (and maybe that path is at least 3-4)?

src/Nethermind/Nethermind.Trie/PatriciaTree.cs
```diff
@@ -208,13 +204,35 @@ private void Commit(NodeCommitInfo nodeCommitInfo, bool skipSelf = false)
...
+                    {
+                        return Task.Run(() =>
+                        {
+                            Commit(committer, ref childPath, new NodeCommitInfo(childNode!, node, idx), maxLevelForConcurrentCommit);
+                            committer.ReturnConcurrencyQuota();
+                        });
+                    }
+
+                    ArrayPoolList<Task>? childTasks = null;
+
```
LukaszRozmej, line 211: And here just have an `if` before and then do either for or Parallel.For

LukaszRozmej (approved):

benaadams (approved):

src/Nethermind/Nethermind.Trie/PatriciaTree.cs
asdacap: Its in the `committer.CanSpawnTask()` part that finally decide if there are threads available. This is limiting it to top level in case there arent enough node to justify multithreading at all.

src/Nethermind/Nethermind.Trie/PatriciaTree.cs
asdacap: It is slightly faster to call directly without creating intermediate TreePath to be put in a list. Thats why its done this way. Althought its probably does not matter.

src/Nethermind/Nethermind.Trie/PatriciaTree.cs
asdacap: This is per tree and does not go to storage, so its fairly balanced beacuse of address/slot->hash translation. Most storage tree are have little change though, something arount 50. If I'm not mistaken, multithreading does not help there. So this mainly help the last state tree. See yellow line in the graph. Parallizing between storage need higher level interface change.

LukaszRozmej: Yeah you are probably right parallelizing storage tries (and changing the interface) is way more important.

PR #7579, opened 2024-10-09
Fix/tool projects build errors

ak88:


tools/SendBlobs/BlobSender.cs
```diff
@@ -85,7 +85,7 @@ public async Task SendRandomBlobs(
             signers.Add(new(new Signer(chainId, privateKey, _logManager), nonce));
         }
 
-        TxDecoder txDecoder = new();
+        TxDecoder txDecoder = TxDecoder.Instance;
```
marcindsobczak, line 88: it's not used, can be removed

marcindsobczak (approved):

PR #7572, opened 2024-10-08
Fix Blob index locking v2

LukaszRozmej:
Alternative to #7568

## Changes

- Run TryGetBlobAndProof under lock, simplify to Dictionary

#### What types of changes does your code introduce?

- [x] Bugfix (a non-breaking change that fixes an issue)
- [ ] New feature (a non-breaking change that adds functionality)
- [ ] Breaking change (a change that causes existing functionality not to work as expected)
- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

#### If yes, did you write tests?

- [ ] Yes
- [x] No

PR #7584, opened 2024-10-10
Fix recreation of light blob tx collection

marcindsobczak:
## Changes

- Fix recreation of light blob tx collection

Fixes regression from https://github.com/NethermindEth/nethermind/pull/7572 and https://github.com/NethermindEth/nethermind/pull/7581

it was failing, because we are loading `LightTransactions` from db and trying to add to in-memory collection using `TryInsert`, which in `SortedPool` is calling `IncertCore`, which is overriden in `PersistentBlobTxDistinctSortedPool`. We should use directly `InsertCore` to skip the override which is adding txs to db

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No


kamilchodola (approved):

kamilchodola:
My node works again so can assume it is fixing the issue

PR #7585, opened 2024-10-10
Optimize Tasks collections in .Net 9

LukaszRozmej:
Builds upon  #7532

## Changes

- Move task collections to pooled ones and use spans when using `WaitAll`, `WhenAll` and `WhenAny`

- [x] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No


src/Nethermind/Nethermind.Consensus/Producers/MultipleBlockProducer.cs
```diff
@@ -32,23 +33,20 @@ protected MultipleBlockProducer(
...
+                    ? Task.FromResult<Block?>(null)
+                    : blockProducerInfo.BlockProducer.BuildBlock(parentHeader, blockProducerInfo.BlockTracer, cancellationToken: token));
             }
 
             IEnumerable<(Block? Block, T BlockProducer)> blocksWithProducers;
 
             try
             {
-                Block?[] blocks = await Task.WhenAll(produceTasks);
+                Block?[] blocks = await Task.WhenAll<Block?>(produceTasks.AsSpan());
```
rubo, line 49: Do we really need `.AsSpan()`?

rubo (approved):
Nitpick: In some places we add items to the pool in a `for` loop with `.Add()`, in other places, with LINQ's `.Select()` or `Enumerable.Range().Select()`. Isn't it worth unifying the approach?

benaadams (commented):
Why the .orig files?

LukaszRozmej:
> Why the .orig files?

github merge fail

benaadams (approved):

PR #7586, opened 2024-10-11
Add pivot updator for optimism

marcindsobczak:
## Changes

- adding `UnsafePivotUpdator` which is working similar as general `PivotUpdator` except using head block hash (minus 64 blocks) instead of finalized block hash.

In `PivotUpdator` we are using finalized block hash, because it can't be reorganized so is a safe starting point. Optimism is not providing finalized block hash until fully synced, so we need to use unsafe head block hash, which potentially can be reorganized and break the node. For more safety we are using head block minus 64. On Optimism reorganizations aren't common, so it can be a justified risk to significantly improve UX in most cases

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No


LukaszRozmej:
Wouldn't it be safer to get Head - N blocks? (Whatever that N would be)

marcindsobczak:
@LukaszRozmej it would be safer, but then we need to refactor and complicate the logic way more

Right now we are getting block hash and we are asking peers for a block by hash. Then we have block number and that's it.

If we want to have `head - N`, then we need to ask peers again about block `head - N` to get it's hash

marcindsobczak:
Ok after discussion we need to use `head - N`, so working on it

kamilchodola (approved):

flcl42 (approved):

src/Nethermind/Nethermind.Merge.Plugin/Synchronization/UnsafePivotUpdator.cs
```diff
@@ -0,0 +1,163 @@
...
+    ISyncConfig syncConfig,
+    IBlockCacheService blockCacheService,
+    IBeaconSyncStrategy beaconSyncStrategy,
+    IDb metadataDb,
+    ILogManager logManager)
+    : PivotUpdator(blockTree, syncModeSelector, syncPeerPool, syncConfig,
+        blockCacheService, beaconSyncStrategy, metadataDb, logManager)
+{
+
+    private const int NumberOfBlocksBehindHeadForSettingPivot = 64;
```
LukaszRozmej, line 34: `Reorganization.MaxDepth`?

src/Nethermind/Nethermind.Merge.Plugin/Synchronization/PivotUpdator.cs
```diff
@@ -126,38 +126,29 @@ private async void OnSyncModeChanged(object? sender, SyncModeChangedEventArgs sy
...
+    protected virtual async Task<(Hash256 Hash, long Number)?> TryGetPivotData(CancellationToken cancellationToken)
     {
+        // getting finalized block hash as it is safe, because can't be reorganized
         Hash256? finalizedBlockHash = _beaconSyncStrategy.GetFinalizedHash();
 
         if (finalizedBlockHash is null || finalizedBlockHash == Keccak.Zero)
         {
-            if (_logger.IsInfo && (_maxAttempts - _attemptsLeft) % 10 == 0) _logger.Info($"Waiting for Forkchoice message from Consensus Layer to set fresh pivot block [{_maxAttempts - _attemptsLeft}s]");
-
+            PrintWaitingForMessageFromCl();
```
LukaszRozmej, line 141: when you invert this `if` then the override looks quite similar in structure and probably can be mostly shared.

src/Nethermind/Nethermind.Merge.Plugin/Synchronization/UnsafePivotUpdator.cs
```diff
@@ -0,0 +1,163 @@
...
+                    return (potentialPivotBlockHash, potentialPivotBlockNumber);
+                }
+            }
+        }
+
+        PrintWaitingForMessageFromCl();
+        return null;
+    }
+
+    private long? TryGetHeadBlockNumberFromBlockCache(Hash256 headBlockHash)
```
LukaszRozmej, line 66: more or less same as `TryGetFinalizedBlockNumberFromBlockCache`

src/Nethermind/Nethermind.Merge.Plugin/Synchronization/UnsafePivotUpdator.cs
```diff
@@ -0,0 +1,163 @@
...
+                if (_logger.IsInfo) _logger.Info($"Loaded head block {headBlockHash} from block cache. Head block number: {headBlock.Header.Number}");
+                return headBlock.Header.Number;
+            }
+            if (_logger.IsDebug) _logger.Debug($"Hash of header found in block cache is {headBlock.Header.Hash} when expecting {headBlockHash}");
+        }
+
+        return null;
+    }
+
+    private async Task<long> TryGetHeadBlockNumberFromPeers(Hash256 headBlockHash, CancellationToken cancellationToken)
```
LukaszRozmej, line 82: more or less the same as `TryGetFinalizedBlockNumberFromPeers`

src/Nethermind/Nethermind.Merge.Plugin/Synchronization/UnsafePivotUpdator.cs
```diff
@@ -0,0 +1,163 @@
...
+                if (_logger.IsInfo) _logger.Info($"Loaded potential pivot block {potentialPivotBlockNumber} from block cache. Hash: {block.Hash}");
+                return block.Hash;
+            }
+        }
+
+        if (_logger.IsDebug) _logger.Debug("Header of pivot block not found in block cache");
+        return null;
+    }
+
+    private async Task<Hash256?> TryGetPotentialPivotBlockHashFromPeers(long potentialPivotBlockNumber, CancellationToken cancellationToken)
```
LukaszRozmej, line 131: very similar to other peer methods `TryGetPotentialPivotBlockHashFromPeers`

LukaszRozmej (commented):
I think the class hierarchy is too complicated duplicating a lot of stuff, details in comments.
This resulted of making a lot of stuff `protected` which is not great. 
Basically subpar code organization and reuse.

PR #7590, opened 2024-10-11
Change for 7702 for devnet-4

ak88:
New spec for 7702 here:
https://eips.ethereum.org/EIPS/eip-7702

## Changes

- PER_AUTH_BASE_COST now costs 15.000
- Fields are bounded as: 
assert auth.chain_id < 2 **64
assert auth.nonce < 2 **64
assert len(auth.address) == 20
assert auth.y_parity < 2 **8
assert auth.r < 2 **256
assert auth.s < 2 **256
- A zero address delegation will now reset the account to no code.
- EXTCODEHASH returns 0 on empty accounts as well.

Tuple fields have been bounded by their type during RLP serialization/deserialization, so any field exceeding the bounds will result in `RlpException` 

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_


src/Nethermind/Ethereum.Test.Base/AuthorizationListJson.cs
```diff
@@ -10,7 +11,7 @@ public class AuthorizationListJson
     public Address Address { get; set; }
     public ulong Nonce { get; set; }
     public ulong V { get; set; }
```
LukaszRozmej, line 13: Should it be `byte`?

ak88: Yes, this actually shouldn't be V but yParity/RecoveryId

LukaszRozmej (approved):

MarekM25 (approved):

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
```diff
@@ -328,15 +328,6 @@ public ValidationResult IsWellFormed(Transaction transaction, IReleaseSpec relea
...
             null or { Length: 0 } => TxErrorMessages.MissingAuthorizationList,
-            var authorizationList when authorizationList.Any(a => !ValidateAuthoritySignature(a.AuthoritySignature)) =>
-                TxErrorMessages.InvalidAuthoritySignature,
             _ => ValidationResult.Success
         };
-
-    private bool ValidateAuthoritySignature(Signature signature)
-    {
-        UInt256 sValue = new(signature.SAsSpan, isBigEndian: true);
-
```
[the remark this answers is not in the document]
ak88: It went from invalidating the tx to just invalidating the tuple instead, so the check is now in `TransactionProcessor`.

smartprogrammer93 (approved):
LGTM

PR #7596, opened 2024-10-13
Reduce branches in HexPrefix.FromBytes

benaadams:
## Changes

- Absorb odd outside loop
- Use 16bit LUT to do two nibbles at a time

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Trie/HexPrefix.cs
+++ b/src/Nethermind/Nethermind.Trie/HexPrefix.cs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 
 using System;
+using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Trie/HexPrefix.cs
```diff
@@ -39,23 +42,45 @@ public static byte[] ToBytes(byte[] path, bool isLeaf)
...
+
+            Span<ushort> nibbles = MemoryMarshal.CreateSpan(
+                ref Unsafe.As<byte, ushort>(ref MemoryMarshal.GetReference(span)),
+                span.Length / 2);
+
+            Debug.Assert(nibbles.Length == bytes.Length);
+
+            ref byte byteRef = ref MemoryMarshal.GetReference(bytes);
+            ref ushort lookup16 = ref MemoryMarshal.GetArrayDataReference(Lookup16);
+            for (int i = 0; i < nibbles.Length; i++)
```
LukaszRozmej, line 65: this looks SIMD'able?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Trie/HexPrefix.cs
benaadams: I think it needs a deeper look; like do we need to expand to nibbles and convert back at all? 🤔

PR #7598, opened 2024-10-13
Speed up docker builds

benaadams:
## Changes

- Don't publish ReadyToRun for faster deployment
- Nethermind is a very long running process so initial startup time isn't so important vs better instrumentation from tier0 to produce improved teir1 code via Jit PGO

- [x] Optimization
- [x] Build-related changes

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Runner/Dockerfile
+++ b/src/Nethermind/Nethermind.Runner/Dockerfile
@@ -23,4 +23,7 @@ RUN dotnet publish Nethermind.Runner.csproj -c release -o /app/publish -p:UseApp
 FROM base AS final
 WORKDIR /app
 COPY --from=publish /app/publish .
+
+ENV DOTNET_ReadyToRun 0
+
 ENTRYPOINT ["dotnet", "nethermind.dll"]
--- a/src/Nethermind/Nethermind.Runner/Dockerfile
+++ b/src/Nethermind/Nethermind.Runner/Dockerfile
@@ -23,5 +23,4 @@ RUN dotnet publish Nethermind.Runner.csproj -c release -o /app/publish -p:UseApp
 FROM base AS final
 WORKDIR /app
 COPY --from=publish /app/publish .
-
 ENTRYPOINT ["dotnet", "nethermind.dll"]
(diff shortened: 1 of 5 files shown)


rubo (changes requested):
We should not disable R2R for Docker only. Nethermind builds should not differ depending on distribution channels. If we consider R2R not important, we disable it in general.
The long build times for the Docker image are due to GitHub Actions runners. We can either switch to a more powerful one (~8 mins) or just wait a little more with the current one. It's not a big deal.

benaadams:
> We should not disable R2R for Docker only. Nethermind builds should not differ depending on distribution channels. If we consider R2R not important, we disable it in general. The long build times for the Docker image are due to GitHub Actions runners. We can either switch to a more powerful one (~8 mins) or just wait a little more with the current one. It's not a big deal.

I did disable it everywhere?

rubo:
> I did disable it everywhere?

Then why have it `false` or `0` explicitly?

benaadams:
> > I did disable it everywhere?
> 
> Then why have it `false` or `0` explicitly?

It only applies to publishing; so changed it to `false` in .csproj

The other one in docker files turns off ReadyToRun for the framework so it Jits everything but that can only be set via environment var not any project or build setting :(

rubo:
> The other one in docker files

Do we need it? Isn't it off by default?

src/Nethermind/Nethermind.Runner/Dockerfile
```diff
@@ -23,4 +23,5 @@ RUN dotnet publish Nethermind.Runner.csproj -c release -o /app/publish -p:UseApp
 FROM base AS final
 WORKDIR /app
 COPY --from=publish /app/publish .
+
```
rubo, line 26:
```suggestion
```

src/Nethermind/Nethermind.Runner/Nethermind.Runner.csproj
```diff
@@ -9,7 +9,7 @@
     <InvariantGlobalization>true</InvariantGlobalization>
     <ServerGarbageCollection>true</ServerGarbageCollection>
     <TieredPGO>true</TieredPGO>
-    <PublishReadyToRun>true</PublishReadyToRun>
+    <PublishReadyToRun>false</PublishReadyToRun>
```
rubo, line 12: Maybe it's better just to delete the line.

```suggestion
```

benaadams: sure

PR #7599, opened 2024-10-13
Set metrics consistently before request completion

benaadams:
## Changes

- Set metrics before request completion so Prometheus more reliably picks up the correct values at request end

#### Requires testing

- [x] No


src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
```diff
@@ -414,34 +415,30 @@ private void FireProcessingQueueEmpty()
...
-            Metrics.LastBlockProcessingTimeInMs = blockProcessingTimeInMs;
         }
 
-        if (!readonlyChain)
-        {
-            Metrics.RecoveryQueueSize = _recoveryQueue.Count;
-            Metrics.ProcessingQueueSize = _blockQueue.Count;
-            _stats.UpdateStats(lastProcessed, _blockTree, _stopwatch.ElapsedMicroseconds());
-        }
+        Metrics.BestKnownBlockNumber = _blockTree.BestKnownNumber;
```
LukaszRozmej, line 441: I think this should too be under the `if (!readonlyChain)`

benaadams: Fixed

LukaszRozmej (approved):

PR #7605, opened 2024-10-15
Perf/parallelize storage commit

asdacap:
- Parallelize storage commit by committing multiple storage at the same time.
- This need an interface change due to the locking of the dirty nodes.
  - The interface change also remove the need to know if its a state or storage trie.
- Also reduce the threshold at which it start parallel commit.
- This improve block processing time by 5%.
- Graph is before/after/before/after.
Screenshot from 2024-10-15 14-57-40
- Note that before, the storage time include the time it is waiting for a lock while memory pruning is happening. The time just for storage is the storage commit time minus pruning time. The times are all prometheus's rate.

- [X] Optimization

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Testing....


src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -190,7 +188,7 @@ private void CommitNode(long blockNumber, Hash256? address, ref TreePath path, i
 
                 if (!_pruningStrategy.PruningEnabled)
                 {
-                    PersistNode(address, path, node, blockNumber, writeFlags);
+                    PersistNode(address, path, node, blockNumber, writeBatch!, writeFlags);
```
LukaszRozmej, line 191: writeBatch can be null, but we are assuming here it is now null, probably because of `NonPruningTrieStoreCommitter`, but can we express this better? Maybe move this method to the Committer?

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -289,70 +287,78 @@ static void ThrowNodeIsNotSame(TrieNode node, TrieNode cachedNodeCopy) =>
...
+                // Note, archive node still use this path. This is because it need the block commit set to handle
+                // reorg announcement.
+                return _currentBlockCommitter.GetTrieCommitter(address, root, writeFlags);
+            }
+            else
+            {
+                // This happens when there are no block involved, such as during snap sync or just calculating
+                // patricia root.
+                return new NonPruningTrieStoreCommitter(this, 0, address, _nodeStorage.StartWriteBatch(), writeFlags);
+            }
```
LukaszRozmej, line 308:
```suggestion
            return _currentBlockCommitter is not null 
                // Note, archive node still use this path. This is because it needs the block commit set to handle reorg announcement.
                ? _currentBlockCommitter.GetTrieCommitter(address, root, writeFlags)  
                // This happens when there are no block involved, such as during snap sync or just calculating patricia root.
                : new NonPruningTrieStoreCommitter(this, 0, address, _nodeStorage.StartWriteBatch(), writeFlags);
```

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -289,70 +287,78 @@ static void ThrowNodeIsNotSame(TrieNode node, TrieNode cachedNodeCopy) =>
...
+            if (_logger.IsTrace) _logger.Trace($"Enqueued blocks {_commitSetQueue?.Count ?? 0}");
+            set.Seal(root);
+
+            bool shouldPersistSnapshot = _persistenceStrategy.ShouldPersist(set.BlockNumber);
+            if (shouldPersistSnapshot)
+            {
+                INodeStorage.WriteBatch currentBatch = _nodeStorage.StartWriteBatch();
+                try
+                {
+                    PersistBlockCommitSet(set, currentBatch);
```
LukaszRozmej, line 338: Why not `ParallelPersistBlockCommitSet`?

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -289,70 +287,78 @@ static void ThrowNodeIsNotSame(TrieNode node, TrieNode cachedNodeCopy) =>
...
+                    PruneCommitSet(set);
+                }
+                finally
+                {
+                    // For safety we prefer to commit half of the batch rather than not commit at all.
+                    // Generally hanging nodes are not a problem in the DB but anything missing from the DB is.
+                    currentBatch?.Dispose();
```
LukaszRozmej, line 345: I think this does not need nullable?

LukaszRozmej, line 345:
```suggestion
                // For safety we prefer to commit half of the batch rather than not commit at all.
                // Generally hanging nodes are not a problem in the DB but anything missing from the DB is.
                using INodeStorage.WriteBatch currentBatch = _nodeStorage.StartWriteBatch();
                PersistBlockCommitSet(set, currentBatch);
                PruneCommitSet(set);
```

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -602,7 +608,7 @@ private void PersistedNodeRecorder(TreePath treePath, Hash256 address, TrieNode
         /// <summary>
         /// Prunes persisted branches of the current commit set root.
         /// </summary>
-        private void PruneCurrentSet()
+        private void PruneCommitSet(BlockCommitSet set)
```
LukaszRozmej, line 611: should it be a method on `BlockCommitSet`?

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -705,22 +710,23 @@ private static ConcurrentQueue<BlockCommitSet> CreateQueueAtomic(ref ConcurrentQ
...
 
-        private void CreateCommitSet(long blockNumber)
+        private BlockCommitSet CreateCommitSet(long blockNumber)
         {
             if (_logger.IsDebug) _logger.Debug($"Beginning new {nameof(BlockCommitSet)} - {blockNumber}");
 
             // TODO: this throws on reorgs, does it not? let us recreate it in test
-            Debug.Assert(CurrentPackage is null || blockNumber == CurrentPackage.BlockNumber + 1, "Newly begun block is not a successor of the last one");
-            Debug.Assert(IsCurrentListSealed, "Not sealed when beginning new block");
+            Debug.Assert(!(_commitSetQueue ?? CreateQueueAtomic(ref _commitSetQueue)).TryDequeue(out BlockCommitSet lastSet) || blockNumber == lastSet.BlockNumber + 1, $"Newly begun block is not a successor of the last one.");
```
LukaszRozmej, line 718: why _commitSetQueue can't be:
a. Created at start
b. Created lazily through property

Would simplify the code a bit.

LukaszRozmej: Won't the Dequeue break the queue?

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -705,22 +710,23 @@ private static ConcurrentQueue<BlockCommitSet> CreateQueueAtomic(ref ConcurrentQ
...
+            Debug.Assert(!(_commitSetQueue ?? CreateQueueAtomic(ref _commitSetQueue)).TryDequeue(out lastSet) || lastSet.IsSealed, "Not sealed when beginning new block");
 
             BlockCommitSet commitSet = new(blockNumber);
             (_commitSetQueue ?? CreateQueueAtomic(ref _commitSetQueue)).Enqueue(commitSet);
             LatestCommittedBlockNumber = Math.Max(blockNumber, LatestCommittedBlockNumber);
+            // Why are we announcing **before** committing next block??
+            // Should it be after commit?
```
LukaszRozmej, line 725: Is this event now used outside of tests? Maybe it can be deleted?

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -1130,49 +1120,94 @@ public bool HasRoot(Hash256 stateRoot)
...
+            }
+
+            public ICommitter GetTrieCommitter(Hash256? address, TrieNode? root, WriteFlags writeFlags)
+            {
+                if (address == null) StateRoot = root;
+                if (trieStore._pruningStrategy.PruningEnabled)
+                    return new PruningTrieStoreCommitter(this, trieStore, commitSet.BlockNumber, address, root);
+                else
+                    return new NonPruningTrieStoreCommitter(trieStore, commitSet.BlockNumber, address,
+                        trieStore._nodeStorage.StartWriteBatch(), writeFlags);
```
LukaszRozmej, line 1143:
```suggestion
                if (address is null) StateRoot = root;
                return trieStore._pruningStrategy.PruningEnabled
                    ? new PruningTrieStoreCommitter(this, trieStore, commitSet.BlockNumber, address, root)
                    : new NonPruningTrieStoreCommitter(trieStore, commitSet.BlockNumber, address, trieStore._nodeStorage.StartWriteBatch(), writeFlags);

```

LukaszRozmej (commented):
Overall great, but due to large refactor I think code can be made more readable and there are quite a few possibilities to do that.

It is now confusing what is the difference, usage and purpose of `TrieStore.GetTrieStore` and the difference between `ITrieStore.BeginBlockCommit` and `IScopedTrieStore.BeginCommit`. We either need better naming, or some documenting comments or both.

Also adjust doc comments where method parameters changed.

src/Nethermind/Nethermind.Trie/Pruning/ITrieStore.cs
```diff
@@ -33,11 +33,19 @@ public interface ITrieStore : IDisposable
         byte[]? LoadRlp(Hash256? address, in TreePath path, Hash256 hash, ReadFlags flags = ReadFlags.None);
         byte[]? TryLoadRlp(Hash256? address, in TreePath path, Hash256 hash, ReadFlags flags = ReadFlags.None);
         INodeStorage.KeyScheme Scheme { get; }
-        ICommitter BeginCommit(TrieType trieType, long blockNumber, Hash256? address, TrieNode? root, WriteFlags writeFlags);
+        ICommitter BeginCommit(Hash256? address, TrieNode? root, WriteFlags writeFlags);
+        IBlockCommitter BeginBlockCommit(long blockNumber);
```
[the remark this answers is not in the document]
asdacap: Yes, this is exposed here because ScopedTrieStore calls it. Maybe it can be hidden somehow.

src/Nethermind/Nethermind.Trie/Pruning/ITrieStore.cs
```diff
@@ -33,11 +33,19 @@ public interface ITrieStore : IDisposable
...
     public interface IPruningTrieStore
     {
         public void PersistCache(CancellationToken cancellationToken);
     }
+
+    public interface IBlockCommitter : IDisposable
+    {
+        ICommitter GetTrieCommitter(Hash256? address, TrieNode? root, WriteFlags writeFlags = WriteFlags.None);
+        bool TryRequestConcurrencyQuota() => false;
+        void ReturnConcurrencyQuota() { }
```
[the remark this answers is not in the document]
asdacap: Don't know. What are the advantage of doing so? Like, if it return disposable, would the disposable take cpu cycle? And how would it know it need to parallelize? `out bool yes`?

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
```diff
@@ -318,17 +319,36 @@ private void SaveToTree(HashSet<AddressAsKey> toUpdateRoots, Change change)
...
+                        blockCommitter.ReturnConcurrencyQuota();
+                    }, storage.Value));
+                }
+                else
+                {
+                    storage.Value.Commit();
+                }
             }
 
+            Task.WaitAll(commitTask.ToArray());
```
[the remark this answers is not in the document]
asdacap: Ok. Did we enable .net 9 already?

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
asdacap: Yea, I originally split it out, but then merge it back in to reduce change.

src/Nethermind/Nethermind.State/WorldState.cs
```diff
@@ -174,8 +174,11 @@ public void DecrementNonce(Address address, UInt256 delta)
 
         public void CommitTree(long blockNumber)
         {
-            _persistentStorageProvider.CommitTrees(blockNumber);
-            _stateProvider.CommitTree(blockNumber);
+            using (IBlockCommitter committer = _trieStore.BeginBlockCommit(blockNumber))
+            {
+                _persistentStorageProvider.CommitTrees(committer);
+                _stateProvider.CommitTree();
```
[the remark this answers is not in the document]
asdacap: I don't think that is easy. The root of the storage tree need to be committed for the trees to connect.

src/Nethermind/Nethermind.Trie.Test/Pruning/TreeStoreTests.cs
```diff
@@ -137,13 +136,13 @@ public void Should_always_announce_block_number_when_pruning_disabled_and_persis
             using TrieStore fullTrieStore = CreateTrieStore(persistenceStrategy: Archive.Instance);
             IScopedTrieStore trieStore = fullTrieStore.GetTrieStore(null);
             fullTrieStore.ReorgBoundaryReached += (_, e) => reorgBoundaryCount += e.BlockNumber;
-            trieStore.BeginCommit(TrieType.State, 1, trieNode).Dispose();
+            fullTrieStore.BeginStateBlockCommit(1, trieNode).Dispose();
```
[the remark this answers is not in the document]
asdacap: I don't think so. Only PatriciaTree should call `BeginCommit`. Its just that this code here is for testing purpose.

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
asdacap: Ow shit, this is suppose to be peek.

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
asdacap: Move it back out?

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
asdacap: ok

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
asdacap: Used by TrieStoreBoundaryWatcher to update BestPersistedState.

src/Nethermind/Nethermind.Trie.Test/Pruning/TreeStoreTests.cs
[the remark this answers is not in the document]
LukaszRozmej: I mean for tests only, it is duplicated in many places.

src/Nethermind/Nethermind.Trie/Pruning/ITrieStore.cs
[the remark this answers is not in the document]
LukaszRozmej: Maybe move it to some other interface?

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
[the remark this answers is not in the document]
LukaszRozmej: not yet, only in PR, waiting for the release

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
LukaszRozmej: Ok then I think it can be done after commit

src/Nethermind/Nethermind.Trie/Pruning/ITrieStore.cs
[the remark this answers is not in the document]
asdacap: There is some weirdness going on. Gonna need a bigger PR to untangle. I'll just group them under `ITrieStoreInternal` for now.

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs

src/Nethermind/Nethermind.Trie.Test/Pruning/TreeStoreTests.cs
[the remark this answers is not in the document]
asdacap: Done

src/Nethermind/Nethermind.Trie.Test/TrieTests.cs
```diff
@@ -92,9 +92,9 @@ public void Single_leaf_update_next_blocks()
             using TrieStore trieStore = new(memDb, Prune.WhenCacheReaches(1.MB()), Persist.EveryBlock, _logManager);
             PatriciaTree patriciaTree = new(trieStore, _logManager);
             patriciaTree.Set(_keyA, _longLeaf1);
-            patriciaTree.Commit(0);
+            using (trieStore.BeginBlockCommit(0)) { patriciaTree.Commit(); }
```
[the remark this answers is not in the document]
asdacap: Donee

src/Nethermind/Nethermind.Trie/Pruning/ITrieStore.cs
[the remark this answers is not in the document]
LukaszRozmej: If the disposable is a struct (returned as struct) and doesn't have to be promoted on the heap then the impact is minimal.  The advantage is better locality on interface like with Commiters. Easy to see that it needs to be dealt with (someone can forget to call `ReturnConcurrencyQuota`.

Overall it is not something super important.

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -289,70 +287,78 @@ static void ThrowNodeIsNotSame(TrieNode node, TrieNode cachedNodeCopy) =>
...
         {
-            ArgumentOutOfRangeException.ThrowIfNegative(blockNumber);
-            EnsureCommitSetExistsForBlock(blockNumber);
-
-            int concurrency = _pruningStrategy.PruningEnabled
-                ? Environment.ProcessorCount
-                : 0; // The write batch when pruning is not enabled is not concurrent safe
+            if (_pruningStrategy.PruningEnabled)
+            {
+                if (_currentBlockCommitter is null) throw new InvalidOperationException($"With pruning triestore, {nameof(BeginBlockCommit)} must be called.");
```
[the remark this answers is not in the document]
LukaszRozmej: My thoughts about `BeginCommit` not being public, can we scope it under `BeginBlockCommit`?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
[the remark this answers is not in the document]
LukaszRozmej: Should this method be on `IBlockCommitter` instead of `TrieStore`?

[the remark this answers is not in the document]
asdacap: Ow, I don't think so. The `IScopedTrieStore`, (make more sense to think of it as THE trie store, while current ITrieStore is more like IBlockTrieStore), is used by the PatriciaTree even outside block processing/context. For example receipts root calculation. So the `ICommitter` is separate from `IBlockCommitter`.

PR #7607, opened 2024-10-15
Staging branch for devnet 4 

ak88:
changes for pecta devnet 4


src/Nethermind/Ethereum.Blockchain.Pyspec.Test/LoadPyspecTestsStrategy.cs
```diff
@@ -20,7 +20,7 @@ public class LoadPyspecTestsStrategy : ITestLoadStrategy
 
     public IEnumerable<IEthereumTest> Load(string testsDir, string wildcard = null)
     {
-        string testsDirectoryName = Path.Combine(AppContext.BaseDirectory, "PyTests", ArchiveVersion, ArchiveName.Split('.')[0]);
+        string testsDirectoryName = Path.Combine("E:/", "PyTests", ArchiveVersion, ArchiveName.Split('.')[0]);
```
MarekM25: revert @smartprogrammer93

ak88: already reverted

src/Nethermind/Nethermind.Core/ExecutionRequest/ExecutionRequest.cs
```diff
@@ -0,0 +1,153 @@
...
+                throw new InvalidOperationException("Buffer is not large enough to hold all data of requests");
+            }
+
+            // Copy the RequestData to the buffer at the current position
+            request.RequestData.CopyTo(buffer.Slice(currentPosition, request.RequestData.Length));
+            currentPosition += request.RequestData.Length;
+        }
+    }
+
+    public static ArrayPoolList<byte[]> GetFlatEncodedRequests(
```
ak88, line 69: Is this structure really needed in the block itself?

rjnrohit: do you mean, is there need to specify `ExecutionRequests` field in the block ? 

It is a workaround to to populate `ExecutionRequests` in `GetPayloadResultV4`, which is returned by `engine_getPaylaodV4`

src/Nethermind/Nethermind.Core/ExecutionRequest/ExecutionRequest.cs
```diff
@@ -0,0 +1,153 @@
...
+        IEnumerable<ExecutionRequest> depositRequests,
+        IEnumerable<ExecutionRequest> withdrawalRequests,
+        IEnumerable<ExecutionRequest> consolidationRequests
+    )
+    {
+        ArrayPoolList<byte[]> requests = new(3);
+        using ArrayPoolList<byte> depositBuffer = new(depositRequestsBytesSize);
+        using ArrayPoolList<byte> withdrawalBuffer = new(withdrawalRequestsBytesSize);
+        using ArrayPoolList<byte> consolidationBuffer = new(consolidationRequestsBytesSize);
```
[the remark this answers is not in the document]
rjnrohit: By doing so, it'll force expand the enumerables and later part of the code will try to do the same thing. L83, 88 and L93

src/Nethermind/Nethermind.Blockchain.Test/Validators/BlockValidatorTests.cs
```diff
@@ -199,51 +198,4 @@ public void ValidateSuggestedBlock_SuggestedBlockIsInvalid_CorrectErrorIsSet(Blo
 
         Assert.That(error, Does.StartWith(expectedError));
     }
-
-    [Test]
-    public void ValidateBodyAgainstHeader_BlockHasInvalidRequestRoot_ReturnsFalse()
```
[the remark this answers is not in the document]
rjnrohit: It is not possible to imitate this tests, as requests are generated and hash calculation is being done only during the block processing step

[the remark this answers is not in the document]
ak88: I was worried that we didnt check if requests_hash was null when the fork activated and vice versa, but i can see that we do in `ValidateParams`  https://github.com/NethermindEth/nethermind/blob/master/src/Nethermind/Nethermind.Merge.Plugin/Data/IExecutionPayloadParams.cs

But i think the opposite case is missing. The fork is activated and requests_hash is set.

src/Nethermind/Nethermind.Core/ExecutionRequest/ExecutionRequest.cs
ak88: Yes, i can see that. I was thinking maybe we could isolate it to being in `IBlockProductionContext`  in the cache or something like that.

rjnrohit: Yep, I was thinking to isolate it, but it would've required to create a wrapper to a block so that result of https://github.com/NethermindEth/nethermind/blob/84e114c11ac7902b53681f37ce81c4c92dbf2111/src/Nethermind/Nethermind.Consensus/Producers/BlockProducerBase.cs#L205 contains the requests. Moreover,  changes at many places would have to be done, all the way up to the `blockProcessor` step. 

To avoid all of these complexities, I chose to create a new field under `Block` class

ak88:
superceded by #7670

PR #7609, opened 2024-10-15
Cleanup/remove lookup by hash

asdacap:
- Remove lookup by hash. 
- Because we can't implement them with halfpath.
- And I think most network nowadays should have significant nethermind version that serve snap

- [ ] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No


asdacap:
Lets just wait some more then.

PR #7611, opened 2024-10-16
WIP: Handle single pivot update + single peer; allow long non-finality

flcl42:
## Changes

- Make use single pivot update in case of single peer: we update pivot block number in sync config once and if ; 
- Allow long non finality: sync does not progress if most recent finalized block is more than 128 blocks ago, which seems to be normal for L2s (got a case of 2000 blocks)

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

Need to sync as many nets as possible

- [x] Yes - maybe we can mention one new setting
- [ ] No

- [ ] Yes
- [x] No


marcindsobczak:
I'm trying with a different approach - to use head block minus 64 instead of finalized block. It will not be 100% safe, but in e.g. optimism there is no other option, as there is no finalized hash in FCU until fully synced
https://github.com/NethermindEth/nethermind/pull/7586

flcl42:
> I'm trying with a different approach - to use head block minus 64 instead of finalized block. It will not be 100% safe, but in e.g. optimism there is no other option, as there is no finalized hash in FCU until fully synced #7586

Sounds good! Let me test your approach with taiko. Wdyt about the second part of the pr that fixes the following?: when we have one peer and single fcu sent we can't start snapsync

marcindsobczak:
@flcl42 current aproach is not working with one peer? Why?

PR #7614, opened 2024-10-16
Block beacon header sync until global pivot is updated

### LINKED ISSUE #6304: Node stuck with sync when PivotUpdator changed block for second time based on FCU
At the beginning of the sync for long time there were messages like this:
`Pivot block from Consensus Layer too far from head. PivotBlockNumber: 7087258, TargetBlockNumber: 7087633, difference: 375 blocks. Max difference allowed: 128`

After that, I've noticed that new pivot block was found
```
23 Nov 11:34:20 | New pivot block has been set based on ForkChoiceUpdate from CL. Pivot block number: 7087603, hash: 0x473ca9b477646669691303e0d0649c0c2936872b50f4c75b3f84311720ae35ca
23 Nov 11:34:21 | Changing state UpdatingPivot to FastHeaders at processed: 0 | state: 0 | block: 0 | header: 0 | target block: 7087633 | peer block: 7087633
```

From the moment of finding new block I was getting those errors:
`23 Nov 11:34:22 | Unable to find beacon header at height 7087270. This is unexpected, forcing a new beacon sync.`

It seems like `PivotBlockNumber` is still set to 7087270 where new `PivotBlockNumber` after some time was set to: 7087603

marcindsobczak:
Fixes #6304

## Changes

- block a start of beacon header sync until global pivot is updated

Beacon header sync can be initialized only when global pivot is already set, otherwise it might result in conflicting pivots and a deadlock

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

### DIFF
--- a/src/Nethermind/Nethermind.Merge.Plugin.Test/EngineModuleTests.Setup.cs
+++ b/src/Nethermind/Nethermind.Merge.Plugin.Test/EngineModuleTests.Setup.cs
@@ -90,6 +90,7 @@ public partial class EngineModuleTests
             chain.LogManager);
         invalidChainTracker.SetupBlockchainProcessorInterceptor(chain.BlockchainProcessor);
         chain.BeaconSync = new BeaconSync(chain.BeaconPivot, chain.BlockTree, synchronizationConfig, blockCacheService, chain.PoSSwitcher, chain.LogManager);
+        chain.BeaconSync.AllowBeaconHeaderSync();
         EngineRpcCapabilitiesProvider capabilitiesProvider = new(chain.SpecProvider);
         return new EngineRpcModule(
             new GetPayloadV1Handler(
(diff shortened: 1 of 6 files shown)


LukaszRozmej (approved):

kamilchodola:
While fixing that did you managed to find a proper reproduction for that? For me it is quite random and did not noticed what exactly I need to do to make it fully reproducible.

marcindsobczak:
@kamilchodola I didn't reproduced an issue, but I reproduced the root cause of an issue - setting beacon pivot and starting beacon header sync before setting global pivot. It is still working fine most of the time, but rarely pivots are conflicting and it is breaking the node - I didn't reproduced that part. This PR is fixing the root cause, so we will not have this issue anymore

MarekM25:
@kamilchodola @marcindsobczak While fixing the block production issue, I encountered a few syncing issues with beacon headers. I think this can happen when your CL is not close to the head and needs to catch up to the tip of the chain.

MarekM25 (approved):
PR looks good, but @kamilchodola it will require some testing for sure. I guess a good idea is to restart CL (lighthouse)/EL a few times and trigger a new beacon sync in this way.

@marcindsobczak think about if it can be covered somehow by integreation tests

PR #7621, opened 2024-10-16
Log request on invalid RPC JSON

alexb5dh:
## Changes

Log RPC request data (sliced to 1000 chars) if it's not a valid JSON.

Example (first logged using ERROR level, then with request body at DEBUG): image

- [x] Other: Additional logs

#### Requires testing

- [x] No

- [x] No

- [x] No


src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
```diff
@@ -124,9 +126,18 @@ public async IAsyncEnumerable<JsonRpcResult> ProcessAsync(PipeReader reader, Jso
...
         // Sends an error response and stops the stopwatch.
-        JsonRpcResult GetParsingError(string error, Exception? exception = null)
+        JsonRpcResult GetParsingError(ReadOnlySequence<byte> buffer, string error, Exception? exception = null)
         {
             Metrics.JsonRpcRequestDeserializationFailures++;
+
+            if (Encoding.UTF8.TryGetString(buffer, out string data))
+            {
+                const int sliceSize = 1000;
+                error = data.Length > sliceSize
```
alexb5dh, line 136: Can potentially show request data at the DEBUG level instead, to avoid overflowing the logs with it.

src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
```diff
@@ -124,9 +126,18 @@ public async IAsyncEnumerable<JsonRpcResult> ProcessAsync(PipeReader reader, Jso
 
         // Handles general exceptions during parsing and validation.
         // Sends an error response and stops the stopwatch.
-        JsonRpcResult GetParsingError(string error, Exception? exception = null)
+        JsonRpcResult GetParsingError(ReadOnlySequence<byte> buffer, string error, Exception? exception = null)
```
LukaszRozmej, line 129: pass it by `ref`?

src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
LukaszRozmej: Parse whole string, to then slice it to 1000 is not great for performance, first slice it to some length, then decode.

LukaszRozmej: Yeah would also do it in debug only.

alexb5dh: > Parse whole string, to then slice it to 1000 is not great for performance, first slice it to some length, then decode.

Yeah, that's why [this](https://github.com/NethermindEth/nethermind/pull/7621/commits/f9c0237fde167c7e175e047e01fbb3035a70effb#diff-0bf00aa62ebcd0370f2cc4c8e650028e8905fd9e4d960a5f5829d878fd0feca8) was my initial implementation - linear, almost without redundant allocations.

Problem is, it only works with `ReadOnlySpan` and here we have `ReadOnlySequence`, which causes several problems:
- `Encoding` does not support parsing sequence to a pre-allocated array, as for span.
  - It's possible to replicate this via its `Decoder` and a custom cycle, but this will look complex and will probably have worse performance.
  - Another option would be to get span from the sequence, but that would mean additional data copying if latter has multiple segments.
- UTF8 character size varies. While for regular RPC JSON all would probably use a single byte, slicing bytes may break decoding if any irregular character is involved.

I'll try to find best compromise.

src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
alexb5dh: Is it needed? It's just 2 pointers + 2 ints: 
image

src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
LukaszRozmej: Maybe just use `Encoding.TryGetChars`?

src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
LukaszRozmej: A bit 8 bytes vs 24 bytes.

src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
alexb5dh: - It doesn't support `ReadOnlySequence`, so  copying may be needed if it has multiple segments.
- It just returns `false` if `chars` array is not enough.
image

LukaszRozmej: Ok, maybe instead of `1000` chars, let's just do `bytes.FirstSpan`, that should be cheap?

src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
alexb5dh: Made it pass via `in`/`readonly ref`  everywhere, but I do have some doubts this will provide any performance benefits.

src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
alexb5dh: Tried to solve this without copying a lot of framework code.
For now method will:
- do a simple call to `Decoder.Convert` when sequence has a single segment (probably most, if not all cases)
- call `GetChars` extension method and catch a specific kind of `ArgumentException` to detect chars overflow, if multiple segments are present

Both options are covered by unit tests.

If the second case still raises concerns, I'll do with a modified internal code.

LukaszRozmej (approved):

src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
LukaszRozmej: Would be good to explain that in comments in detail

benaadams (approved):

kamilchodola (approved):

PR #7624, opened 2024-10-17
Remove barriers from OP/BASE mainnet

kamilchodola:
As for now we have invalid Barriers for Base and some random number for op - I guess most healthy for networks and without any concerns on what to set here is to simply remove it and let sync entire history.

Will sync both chains with that and let you know about sync time and size.'

- [ ] Optimization
- [X] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No


flcl42 (approved):

benaadams (approved):

LukaszRozmej (approved):

kamilchodola:
Need to wait until op-mainnet will work properly so verify this.

PR #7641, opened 2024-10-21
Fix compiler warnings for tools

rubo:
## Changes

Fixed compiler warnings for tools projects, so we can have clean results on PR builds like before.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [x] Other: Tools

#### Requires testing

- [ ] Yes
- [x] No


tools/SendBlobs/SetupCli.cs
```diff
@@ -31,7 +31,7 @@ public static void SetupExecute(CommandLineApplication app)
 
         app.OnExecuteAsync(async cancellationToken =>
         {
-            string? rpcUrl = rpcUrlOption.Value();
+            string rpcUrl = rpcUrlOption.Value()!;
```
rubo, line 34: @flcl42 As I'm not familiar with the tool, I just added null-forgiving operators for CLI options to suppress warnings. A better solution is to either provide some default values or make those options mandatory with meaningful error messages. We can go this way for now until you decide what to do. Otherwise, feel free to commit changes to this branch.

flcl42: Could you add IsRequired to first several options also `CommandOption rpcUrlOption = app.Option("--rpcurl <rpcUrl>", "Url of the Json RPC.", CommandOptionType.SingleValue, (c) => c.IsRequired());`?

rubo: @flcl42 Done

LukaszRozmej (approved):

PR #7643, opened 2024-10-22
Shutter fix: check registration update message length

Marchhill:
## Changes

- Check length of update messages in validator registration contract

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Shutter/Contracts/ValidatorRegistryContract.cs
+++ b/src/Nethermind/Nethermind.Shutter/Contracts/ValidatorRegistryContract.cs
@@ -45,7 +45,14 @@ public class ValidatorRegistryContract(
         for (uint i = 0; i < updates; i++)
         {
             Update update = GetUpdate(header, updates - i - 1);
-            Message msg = new(update.Message.AsSpan()[..46]);
+
+            if (update.Message.Length != 46 || update.Signature.Length != 96)
+            {
+                if (_logger.IsDebug) _logger.Debug("Registration message was wrong length.");
+                continue;
+            }
+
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Shutter/Contracts/ValidatorRegistryContract.cs
```diff
@@ -45,7 +45,14 @@ public bool IsRegistered(BlockHeader header, in Dictionary<ulong, byte[]> valida
         for (uint i = 0; i < updates; i++)
         {
             Update update = GetUpdate(header, updates - i - 1);
-            Message msg = new(update.Message.AsSpan()[..46]);
+
+            if (update.Message.Length != 46 || update.Signature.Length != 96)
```
ak88, line 49: Which message is 96 bytes long ? 
I can only see the 46 byte message in https://github.com/gnosischain/specs/blob/master/shutter/low-level.md#validator-registry

Marchhill: It's a BLS signature (compressed G2 point) which must be 96 bytes, an exception will be thrown [here](https://github.com/NethermindEth/nethermind/blob/master/src/Nethermind/Nethermind.Crypto/BlsSigner.cs#L63) if it's not the correct length

ak88 (approved):

PR #7659, opened 2024-10-26
Refactor/Separate block store and Bad block store

asdacap:
- Extract `IBadBlockStore` from `IBlockStore`.
- Otherwise things will get confusing with a different/multiple format for block storage.

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Tested along with #7665 .

## Remarks

- Almost made a `KV<T>` but thought.. nah.


LukaszRozmej (approved):
To be honest I don't really see the need, but I don't see any harm either. So if it makes it simpler for you.

PR #7663, opened 2024-10-27
Vectorize Keccak state copy, remove temp array use

benaadams:
## Changes

- Vectorized Xor to state
- Removed additional temp stackalloc array + copy
- Simplified
- Fast copy to output for 256 and 512 bit sizes

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Core/Crypto/KeccakHash.cs
+++ b/src/Nethermind/Nethermind.Core/Crypto/KeccakHash.cs
@@ -315,13 +315,6 @@ namespace Nethermind.Core.Crypto
             st[0] = aba;
         }
 
-        public static Span<byte> ComputeHash(ReadOnlySpan<byte> input, int size = HASH_SIZE)
-        {
-            Span<byte> output = new byte[size];
-            ComputeHash(input, output);
-            return output;
-        }
-
         public static byte[] ComputeHashBytes(ReadOnlySpan<byte> input, int size = HASH_SIZE)
         {
             byte[] output = new byte[size];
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Core/Crypto/KeccakHash.cs
```diff
@@ -356,47 +356,102 @@ public static uint[] ComputeBytesToUint(ReadOnlySpan<byte> input, int size)
...
+                stateBytes[..output.Length].CopyTo(output);
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static unsafe void XorVectors(Span<byte> state, ReadOnlySpan<byte> input)
+        {
+            int length = input.Length;
+            int i = 0;
+            if (Vector256<byte>.IsSupported)
```
LukaszRozmej, line 420: Where `Vector512`?

src/Nethermind/Nethermind.Core/Crypto/KeccakHash.cs
```diff
@@ -356,47 +356,102 @@ public static uint[] ComputeBytesToUint(ReadOnlySpan<byte> input, int size)
...
+                // Apply padding within the current block
+                stateBytes[remainingInputLength] ^= 0x01; // Append bit '1' after the input
+                stateBytes[roundSize - 1] ^= 0x80;        // Set the last bit of the block to '1'
             }
 
+            // Process the final block
             KeccakF(state);
-            MemoryMarshal.AsBytes(state[..(size / sizeof(ulong))]).CopyTo(output);
+
+            if (output.Length == Vector256<byte>.Count)
```
LukaszRozmej, line 397: no `Vector128`?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Core/Crypto/KeccakHash.cs
benaadams: Not for copying out as hashes are only 256 and 512

src/Nethermind/Nethermind.Core/Crypto/KeccakHash.cs
benaadams: Added

PR #7665, opened 2024-10-28
Feature/Dont download existing block

asdacap:
- Based on #7659 
- With era import and portal, blocks may arbitrarily get downloaded during fast blocks. This mean not just from genesis, but any portion of the history and portal will randomly pick which block to store based on a configured radius.
- This is a problem as fast blocks does not need to download them anymore if it is already in database.
- This PR adds a check in old bodies and old receipts so that it does not re-download existing block.
- Unfortunately because of how rocksdb works checking if a block exist still require fetching it from database. So it is not fast and not parallelizing it does not speed it up much. The complicated code in this PR attempts to make it fast.
- With blocks imported via era from genesis to block 20900000, the whole old blocks+old receipts take about 15 minute with the checking for existing blocks portion taking about 10 minute. When attaching era directory directly, the checking phase take about 2 minute. Not instant, but at least it does not consume 2GBps of bandwith.
- OptimizeFilterForHits option is disabled for bodies and receipts to reduce overhead on checking if a block is available or not. This use up some disk space to store the last level bloom.

- [X] Optimization

#### Requires testing

- [X] Yes

- [X] Yes

#### Notes on testing

- [X] Era imported blocks.
- [ ] Mainet fast sync


src/Nethermind/Nethermind.Synchronization/FastBlocks/SyncStatusList.cs
```diff
@@ -77,6 +79,80 @@ public void GetInfosForBatch(BlockInfo?[] blockInfos)
...
+                // already exist. We switch to a bigger array to improve parallelization throughput
+                if (workingArray.Length < _parallelExistCheckSize)
+                {
+                    workingArray = new BlockInfo[_parallelExistCheckSize];
+                }
+            }
+
+            infos = workingArray;
+            return false;
+        }
```
LukaszRozmej, line 154: This could benefit from even more comments, or making code more readable in other ways - better naming, or named submethods or something.

src/Nethermind/Nethermind.Synchronization/FastBlocks/SyncStatusList.cs
```diff
@@ -77,6 +79,80 @@ public void GetInfosForBatch(BlockInfo?[] blockInfos)
...
+        /// Try get block infos of size `batchSize`.
+        /// </summary>
+        /// <param name="batchSize"></param>
+        /// <param name="blockExist"></param>
+        /// <param name="infos"></param>
+        /// <returns></returns>
+        public bool TryGetInfosForBatch(int batchSize, Func<BlockInfo, bool> blockExist, out BlockInfo?[] infos)
+        {
+            BlockInfo?[] outputArray = new BlockInfo?[batchSize];
+            BlockInfo?[] workingArray = new BlockInfo?[batchSize];
```
LukaszRozmej, line 92: Wonder if we could pool memory here. Probably yes ad `BodiesSyncBatch` and `ReceiptsSyncBatch` already do that for response, so why not for `Infos`?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Synchronization/FastBlocks/SyncStatusList.cs
asdacap: Separated to inner function

src/Nethermind/Nethermind.Synchronization/FastBlocks/SyncStatusList.cs
asdacap: Pooled workingArray.

PR #7675, opened 2024-10-28
Avoid expensive Stelem_Ref covariant checks in Trie

benaadams:
image

## Changes

- Perform own bounds checking and the Unsafe object assignment to avoid expensive Stelem_Ref checks from covariance on `object[]`
- Move some inline throws to ThrowHelpers

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Trie/TrieNode.cs
+++ b/src/Nethermind/Nethermind.Trie/TrieNode.cs
@@ -5,6 +5,7 @@ using System;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
 using System.Threading;
 using Nethermind.Core;
 using Nethermind.Core.Buffers;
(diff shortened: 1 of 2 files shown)


LukaszRozmej (approved):

src/Nethermind/Nethermind.Trie/TrieNode.cs
```diff
@@ -761,6 +760,26 @@ void ThrowAlreadySealed()
...
+        private void SetItem(int i, TrieNode node)
+        {
+            int index = IsExtension ? i + 1 : i;
+            var data = _data;
+            if ((uint)index > (uint)data.Length)
+            {
+                // Since we are accessing unsafely we need to do our own
+                // bounds check to ensure is safe; as the Jit won't help us.
+                ThrowIndexOutOfRangeException();
+            }
```
LukaszRozmej, line 777: I wonder if we could skip the checks in release mode as the index should come from previous logic and be correct (we never had out of range exception on this path)

PR #7677, opened 2024-10-29
Cache message to Debug logs

benaadams:
## Changes

- Move cache messages to Debug rather than Warm

#### Requires testing

- [x] No


LukaszRozmej (changes requested):
Identify and potentially fix the root cause

LukaszRozmej (approved):

src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.cs
```diff
@@ -133,9 +133,10 @@ the previous head state.*/
                 }
                 else
                 {
-                    if (preWarmer?.ClearCaches() ?? false)
+                    CacheType result = preWarmer?.ClearCaches() ?? default;
+                    if (result != default)
                     {
-                        if (_logger.IsWarn) _logger.Warn("Low txs, caches are not empty. Clearing them.");
+                        if (_logger.IsDebug) _logger.Debug($"Low txs, caches {result} are not empty. Clearing them.");
```
LukaszRozmej, line 139: keep it as Warn though?

LukaszRozmej (changes requested):
keep as warnings for now?

src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
```diff
@@ -30,9 +30,10 @@ public Task PreWarmCaches(Block suggestedBlock, Hash256? parentStateRoot, Access
     {
         if (preBlockCaches is not null)
         {
-            if (preBlockCaches.ClearCaches())
+            CacheType result = preBlockCaches.ClearCaches();
+            if (result != default)
             {
-                if (_logger.IsWarn) _logger.Warn("Caches are not empty. Clearing them.");
+                if (_logger.IsDebug) _logger.Debug($"Caches {result} are not empty. Clearing them.");
```
benaadams, line 36:
```suggestion
                if (_logger.IsDebug) _logger.Warn($"Caches {result} are not empty. Clearing them.");
```

src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.cs
benaadams, line 139:
```suggestion
                        if (_logger.IsDebug) _logger.Warn($"Low txs, caches {result} are not empty. Clearing them.");
```

benaadams: Changed to warn

LukaszRozmej (commented):
But you left `IsDebug`?

benaadams:
> But you left `IsDebug`?

Changed back to Warn

PR #7678, opened 2024-10-29
Bls Signature refactoring & performance

Marchhill:
## Changes

- Refactor BlsSigner to allow public key aggregation without allocating large buffer
- Change signature class to allow direct aggregation
- Use improved blst bindings
- Bugfix: not throwing exception when Shutter validator registry file has invalid public key

- [x] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Remarks

### DIFF
--- a/src/Nethermind/Nethermind.Crypto/BlsSigner.cs
+++ b/src/Nethermind/Nethermind.Crypto/BlsSigner.cs
@@ -123,6 +123,9 @@ public static class BlsSigner
             _point = new(buf);
         }
 
+        public void FromSk(Bls.SecretKey sk)
+            => _point.FromSk(sk);
+
         public bool TryDecode(ReadOnlySpan<byte> publicKeyBytes, out Bls.ERROR err)
             => _point.TryDecode(publicKeyBytes, out err);
 
--- a/src/Nethermind/Nethermind.Crypto/BlsSigner.cs
+++ b/src/Nethermind/Nethermind.Crypto/BlsSigner.cs
@@ -23,96 +23,121 @@ public static class BlsSigner
     private const int InputLength = 64;
 
     [SkipLocalsInit]
-    public static Signature Sign(Bls.SecretKey sk, ReadOnlySpan<byte> message)
+    // buf must be of size G2.Sz
+    public static Signature Sign(Span<long> buf, Bls.SecretKey sk, ReadOnlySpan<byte> message)
     {
-        G2 p = new(stackalloc long[G2.Sz]);
-        p.HashTo(message, Cryptosuite);
-        p.SignWith(sk);
-        return new(p.Compress());
-    }
-
-    [SkipLocalsInit]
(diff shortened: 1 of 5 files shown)


src/Nethermind/Nethermind.Crypto/BlsSigner.cs
```diff
@@ -23,96 +23,132 @@ public static class BlsSigner
...
+        }
+
+        G2 p = new(buf);
+        Signature s = new(p);
+        s.Sign(sk, message);
+        return s;
+    }
+
     public static Signature Sign(Bls.SecretKey sk, ReadOnlySpan<byte> message)
+        => Sign(new long[G2.Sz], sk, message);
```
LukaszRozmej, line 41: Can we stackalloc here?

src/Nethermind/Nethermind.Crypto/BlsSigner.cs
```diff
@@ -23,96 +23,132 @@ public static class BlsSigner
...
+
+    public static bool Verify(G1Affine publicKey, Signature signature, ReadOnlySpan<byte> message)
     {
-        G2 p = new(stackalloc long[G2.Sz]);
-        p.HashTo(message, Cryptosuite);
-        p.SignWith(sk);
-        return new(p.Compress());
+        int len = 2 * GT.Sz;
+        using ArrayPoolList<long> buf = new(len, len);
```
LukaszRozmej, line 46: would it be better to stackalloc?
Do you need stack memory cleared in this method?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Crypto/BlsSigner.cs
Marchhill: I made two versions of `Sign`, one takes a buffer as input so you can stackalloc, this one uses new

src/Nethermind/Nethermind.Crypto/BlsSigner.cs
Marchhill: GT points are quite large so don't think it's safe to allocate on the stack

src/Nethermind/Nethermind.Crypto/BlsSigner.cs
LukaszRozmej: but why can't it: `Sign(stackalloc long[G2.Sz], sk, message);` ?

src/Nethermind/Nethermind.Crypto/BlsSigner.cs
LukaszRozmej: How large?

PR #7679, opened 2024-10-29
Fix `+` serialization in JSON RPC

alexb5dh:
Fixes `+` being serialized as `\u002B` in JSON responses to make it consistent with other clients.

#### Requires testing

- [x] No

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Serialization.Json/EthereumJsonSerializer.cs
+++ b/src/Nethermind/Nethermind.Serialization.Json/EthereumJsonSerializer.cs
@@ -56,9 +56,6 @@ namespace Nethermind.Serialization.Json
 
         private static JsonSerializerOptions CreateOptions(bool indented, IEnumerable<JsonConverter> converters = null, int maxDepth = 64)
         {
-            var encoderSettings = new TextEncoderSettings(UnicodeRanges.BasicLatin);
-            encoderSettings.AllowCharacter('+');
-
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Serialization.Json/EthereumJsonSerializer.cs
```diff
@@ -63,6 +65,7 @@ private static JsonSerializerOptions CreateOptions(bool indented, IEnumerable<Js
                 DictionaryKeyPolicy = JsonNamingPolicy.CamelCase,
                 PropertyNameCaseInsensitive = true,
                 MaxDepth = maxDepth,
+                Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping,
```
alexb5dh, line 68: Unfortunately any configurations that allow to only add a `+` sign are internal, so not accessible without reflection or copying framework code.

Potential risks are described [here](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/character-encoding#serialize-all-characters). Should be fine as client isn't intended to work with any HTML/JS and doesn't need XSS protection.

benaadams: Also we are the generators of it and we aren't adding XSS stuff

LukaszRozmej (approved):

PR #7680, opened 2024-10-29
Receipts: Add option to not recover sender and write flags

asdacap:
- On Get() add option to not recover sender
  - This improve era export throughput by 4x.
- On Insert() add option to specify write flags.
  - On era export wal writes can reach 150 MBps.

- [X] Optimization

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain/Receipts/IReceiptFinder.cs
+++ b/src/Nethermind/Nethermind.Blockchain/Receipts/IReceiptFinder.cs
@@ -9,7 +9,7 @@ namespace Nethermind.Blockchain.Receipts
     public interface IReceiptFinder
     {
         Hash256? FindBlockHash(Hash256 txHash);
-        TxReceipt[] Get(Block block, bool recover = true);
+        TxReceipt[] Get(Block block, bool recover = true, bool recoverSender = true);
         TxReceipt[] Get(Hash256 blockHash, bool recover = true);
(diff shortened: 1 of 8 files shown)


src/Nethermind/Nethermind.Blockchain/Receipts/IReceiptFinder.cs
```diff
@@ -9,7 +9,7 @@ namespace Nethermind.Blockchain.Receipts
     public interface IReceiptFinder
     {
         Hash256? FindBlockHash(Hash256 txHash);
-        TxReceipt[] Get(Block block, bool recover = true);
+        TxReceipt[] Get(Block block, bool recover = true, bool recoverSender = true);
```
LukaszRozmej, line 12: Maybe would be better to collapse `recover` and `recoverSender` into an enum?

```
public enum ReceiptRecoveryOptions { None, TransactionAndBlock, TransactionAndBlockWithSender }
```

Hard to name

LukaszRozmej (approved):

src/Nethermind/Nethermind.Blockchain/Receipts/IReceiptFinder.cs
asdacap: I'd say, don't recover sender at all in receipt finder. Its a weird behaviour.

PR #7683, opened 2024-10-29
Allow bh sync instantly; use unsafe updater with taiko

flcl42:
## Changes

- Enables snap sync after single fcu

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Taiko/TaikoPlugin.cs
+++ b/src/Nethermind/Nethermind.Taiko/TaikoPlugin.cs
@@ -302,8 +302,7 @@ public class TaikoPlugin : IConsensusPlugin, ISynchronizationPlugin, IInitializa
 
         _peerRefresher = new PeerRefresher(_api.PeerDifficultyRefreshPool!, _api.TimerFactory, _api.LogManager);
         _api.DisposeStack.Push((PeerRefresher)_peerRefresher);
-        _ = new
-        PivotUpdator(
+        _ = new UnsafePivotUpdator(
             _api.BlockTree,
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Merge.Plugin/Synchronization/UnsafePivotUpdator.cs
```diff
@@ -19,18 +19,22 @@
...
+        ISyncModeSelector syncModeSelector,
+        ISyncPeerPool syncPeerPool,
+        ISyncConfig syncConfig,
+        IBlockCacheService blockCacheService,
+        IBeaconSyncStrategy beaconSyncStrategy,
+        IDb metadataDb,
+        ILogManager logManager) : base(blockTree, syncModeSelector, syncPeerPool, syncConfig,
+            blockCacheService, beaconSyncStrategy, metadataDb, logManager)
+    {
+        _beaconSyncStrategy.AllowBeaconHeaderSync();
```
marcindsobczak, line 35: `UnsafePivotUpdator` is used by optimism too and it is not expected to allow beacon headers sync from the very beginning there. Would be good to create second, non default constructor with additional flag or just add the flag. At the end - allow beacon heder sync from the beginning for taiko, don't allow for op

LukaszRozmej (approved):

PR #7691, opened 2024-10-31
Feature/verify trie on state finish

asdacap:
- Add option to verify trie on state finish.
- This should help to bisect if the issue is after sync or before sync.
- Note: By default halfpath remove past keys. If the verify trie did not finish before next prune cache, it will likely show removed keys. Either increase pruning cache, or disable key removal via `--Pruning.TrackedPastKeyCountMemoryRatio 0`.
- Also make default `--Init.DiagnosticMode VerifyTrie` to block.
- Also fix duplicated progress log at high speed.
- Also add existing code cache.

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

#### Notes on testing
 
- Tested manually on mainnet.

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain/Synchronization/ISyncConfig.cs
+++ b/src/Nethermind/Nethermind.Blockchain/Synchronization/ISyncConfig.cs
@@ -146,4 +146,7 @@ public interface ISyncConfig : IConfig
 
     [ConfigItem(Description = "_Technical._ MultiSyncModeSelector will wait for header to completely sync first.", DefaultValue = "false", HiddenFromDocs = true)]
     bool NeedToWaitForHeader { get; set; }
+
+    [ConfigItem(Description = "_Technical._ Run verify trie on state sync is finished.", DefaultValue = "false", HiddenFromDocs = true)]
+    bool VerifyTrieOnStateSyncFinished { get; }
 }
--- a/src/Nethermind/Nethermind.Blockchain/Synchronization/SyncConfig.cs
+++ b/src/Nethermind/Nethermind.Blockchain/Synchronization/SyncConfig.cs
@@ -66,6 +66,7 @@ namespace Nethermind.Blockchain.Synchronization
         public bool? SnapServingEnabled { get; set; } = null;
         public int MultiSyncModeSelectorLoopTimerMs { get; set; } = 1000;
         public bool NeedToWaitForHeader { get; set; }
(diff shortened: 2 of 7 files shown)


src/Nethermind/Nethermind.Synchronization/FastSync/ITreeSync.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace Nethermind.Synchronization.FastSync;
+
+public interface ITreeSync
+{
+    public event EventHandler<VerifyPostSyncCleanupEventArgs> OnVerifyPostSyncCleanup;
+
+    public class VerifyPostSyncCleanupEventArgs(Hash256 root) : EventArgs
+    {
+        public Hash256 Root => root;
+    }
```
LukaszRozmej, line 16: I think the names are bad, it is just invoked at the end of sync, `TreeSync` doesn't know what the subscriber will do. Also `On` prefix is for event handlers not events themselves.

```suggestion
    public event EventHandler<SyncCompleatedEventArgs> SyncCompleated;

    public class SyncCompleatedEventArgs(Hash256 root) : EventArgs
    {
        public Hash256 Root => root;
    }
```

src/Nethermind/Nethermind.Synchronization/Synchronizer.cs
```diff
@@ -380,12 +380,20 @@ private void ConfigureBodiesSyncComponent(ContainerBuilder serviceCollection)
...
 
         // Disable it by setting noop
         if (!syncConfig.FastSync) serviceCollection.AddSingleton<ISyncFeed<StateSyncBatch>, NoopSyncFeed<StateSyncBatch>>();
+
+        if (syncConfig.FastSync && syncConfig.VerifyTrieOnStateSyncFinished)
+        {
+            serviceCollection
+                .RegisterType<VerifyStateOnStateSyncFinished>()
+                .WithAttributeFiltering()
+                .As<IStartable>();
```
LukaszRozmej, line 395: Are there potentially multiple startables?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Synchronization/Synchronizer.cs
asdacap: Yes. Break lazyness though. Not a fan.

PR #7692, opened 2024-10-31
Perf/faster full tree visit

asdacap:
- Improved tree visitor core utilization.
- On R9 7950x Improved read throughput by 2.5x. With CPU utilization around 80-90%.
- When combined with #7691 (there is a code cache there), can finish mainnet verify trie in 4 minute 30 second.

Screenshot from 2024-10-31 12-28-45

- [X] Optimization

#### Requires testing

- [X] Yes

- [X] Yes

#### Notes on testing

- Tested manually on a mainnet backup.

### DIFF
--- a/src/Nethermind/Nethermind.Core/Threading/ThreadLimiter.cs
+++ b/src/Nethermind/Nethermind.Core/Threading/ThreadLimiter.cs
@@ -36,7 +36,7 @@ public class ThreadLimiter(int concurrency)
         Interlocked.Increment(ref _slots);
     }
 
-    public struct SlotReturner(ThreadLimiter limiter): IDisposable
+    public struct SlotReturner(ThreadLimiter limiter) : IDisposable
     {
         public void Dispose()
         {
--- a/src/Nethermind/Nethermind.Core/Threading/ThreadLimiter.cs
+++ b/src/Nethermind/Nethermind.Core/Threading/ThreadLimiter.cs
@@ -1,0 +1,46 @@
+// SPDX-FileCopyrightText: 2024 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System;
+using System.Threading;
+
+namespace Nethermind.Core.Threading;
+
+/// <summary>
+/// Encapsulate the pattern of checking if new task can be spawned based on a predefined limit.
+/// Used in multithreaded tree visit where we don't know if we can spawn task or not and spawning task itself
+/// is not a cheap operation.
+///
+/// Yes, I don't like the name. Give me a good one.
+/// </summary>
+/// <param name="concurrency">Desired concurrency which include the calling thread. So slot is slot-1.</param>
+public class ThreadLimiter(int concurrency)
+{
+    private int _slots = concurrency - 1;
+
+    public bool TryTakeSlot(out SlotReturner returner)
+    {
+        returner = new SlotReturner(this);
+        int newSlot = Interlocked.Decrement(ref _slots);
+        if (newSlot < 0)
+        {
+            Interlocked.Increment(ref _slots);
+            return false;
+        }
+
+        return true;
+    }
+
+    private void ReturnSlot()
+    {
+        Interlocked.Increment(ref _slots);
+    }
+
+    public struct SlotReturner(ThreadLimiter limiter): IDisposable
+    {
+        public void Dispose()
+        {
+            limiter.ReturnSlot();
+        }
+    }
+}
--- a/src/Nethermind/Nethermind.Core/Threading/ThreadLimiter.cs
+++ b/src/Nethermind/Nethermind.Core/Threading/ThreadLimiter.cs
@@ -16,13 +16,13 @@ namespace Nethermind.Core.Threading;
 /// <param name="concurrency">Desired concurrency which include the calling thread. So slot is slot-1.</param>
 public class ThreadLimiter(int concurrency)
 {
-    private int _slots = concurrency - 1;
+    private int _slots = concurrency;
 
     public bool TryTakeSlot(out SlotReturner returner)
     {
         returner = new SlotReturner(this);
         int newSlot = Interlocked.Decrement(ref _slots);
-        if (newSlot < 0)
+        if (newSlot < 1)
         {
             Interlocked.Increment(ref _slots);
             return false;
--- a/src/Nethermind/Nethermind.Core/Threading/ThreadLimiter.cs
+++ b/src/Nethermind/Nethermind.Core/Threading/ThreadLimiter.cs
@@ -21,12 +21,18 @@ public class ThreadLimiter(int concurrency)
     public bool TryTakeSlot(out SlotReturner returner)
     {
         returner = new SlotReturner(this);
-        int newSlot = Interlocked.Decrement(ref _slots);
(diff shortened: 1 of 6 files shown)


src/Nethermind/Nethermind.Core/Threading/ThreadLimiter.cs
```diff
@@ -0,0 +1,46 @@
...
+/// <param name="concurrency">Desired concurrency which include the calling thread. So slot is slot-1.</param>
+public class ThreadLimiter(int concurrency)
+{
+    private int _slots = concurrency - 1;
+
+    public bool TryTakeSlot(out SlotReturner returner)
+    {
+        returner = new SlotReturner(this);
+        int newSlot = Interlocked.Decrement(ref _slots);
+        if (newSlot < 0)
```
LukaszRozmej, line 25:
```suggestion
    private int _slots = concurrency;

    public bool TryTakeSlot(out SlotReturner returner)
    {
        returner = new SlotReturner(this);
        int newSlot = Interlocked.Decrement(ref _slots);
        if (newSlot < 1)
```

src/Nethermind/Nethermind.Core/Threading/ThreadLimiter.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            Interlocked.Increment(ref _slots);
```
LukaszRozmej, line 27: Can this be done better with `voliatale`? @Scooletz

src/Nethermind/Nethermind.Core/Threading/ThreadLimiter.cs
```diff
@@ -0,0 +1,46 @@
...
+/// Yes, I don't like the name. Give me a good one.
+/// </summary>
+/// <param name="concurrency">Desired concurrency which include the calling thread. So slot is slot-1.</param>
+public class ThreadLimiter(int concurrency)
+{
+    private int _slots = concurrency - 1;
+
+    public bool TryTakeSlot(out SlotReturner returner)
+    {
+        returner = new SlotReturner(this);
```
LukaszRozmej, line 23: you always return something that is Disposable, but expect it to be Disposed only if this was `true`, maybe make it nullable and return `null` if this is false?

src/Nethermind/Nethermind.Trie/VisitContext.cs
```diff
@@ -22,28 +25,16 @@ public class TrieVisitContext : IDisposable
...
+            internal init
             {
-                if (_semaphore is null)
-                {
-                    if (MaxDegreeOfParallelism == 1) throw new InvalidOperationException("Can not create semaphore for single threaded trie visitor.");
-                    _semaphore = new SemaphoreSlim(MaxDegreeOfParallelism, MaxDegreeOfParallelism);
-                }
-
-                return _semaphore;
+                _maxDegreeOfParallelism = VisitingOptions.AdjustMaxDegreeOfParallelism(value);
```
LukaszRozmej, line 30: for safety/complitness?

```suggestion
                _maxDegreeOfParallelism = VisitingOptions.AdjustMaxDegreeOfParallelism(value);
                _threadLimiter = null;
```

src/Nethermind/Nethermind.Trie/TrieNode.Visitor.cs
```diff
@@ -187,31 +190,43 @@ void VisitSingleThread(ref TreePath parentPath, ITreeVisitor<TNodeContext> treeV
...
-                                finally
+                                else
                                 {
-                                    visitContext.Semaphore.Release();
+                                    VisitChild(ref parentPath, i, GetChild(nodeResolver, ref parentPath, i), trieNodeResolver, treeVisitor, contextCopy, visitContext);
                                 }
-                            });
+                            }
+
+                            if (tasks != null && tasks.Count > 0) Task.WaitAll(tasks.ToArray());
```
LukaszRozmej, line 212:
```suggestion
                            if tasks?.Count > 0) Task.WaitAll(tasks.ToArray());
```

Also we can use do `AsSpan()` instead of `ToArray` in .net 9 (update #7585 when this is merged to master)

src/Nethermind/Nethermind.Core/Threading/ThreadLimiter.cs
asdacap: Well, problem is `SlotReturner` is a struct, making it a nullable struct is heavier, and making it reference type is probably even heavier as now its in heap. So it should not make a differece except for having a condition which would probably mess with cpu branch predictor anyway.

src/Nethermind/Nethermind.Core/Threading/ThreadLimiter.cs
Scooletz: It can't as you want to undo the previous `Decrement` in an atomic way.

src/Nethermind/Nethermind.Core/Threading/ThreadLimiter.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        return true;
+    }
+
+    private void ReturnSlot()
+    {
+        Interlocked.Increment(ref _slots);
+    }
+
+    public struct SlotReturner(ThreadLimiter limiter) : IDisposable
```
Scooletz, line 39:
```suggestion
    public readonly struct SlotReturner(ThreadLimiter limiter) : IDisposable
```

src/Nethermind/Nethermind.Core/Threading/ThreadLimiter.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.Threading;
+
+namespace Nethermind.Core.Threading;
+
+/// <summary>
+/// Encapsulate the pattern of checking if new task can be spawned based on a predefined limit.
+/// Used in multithreaded tree visit where we don't know if we can spawn task or not and spawning task itself
+/// is not a cheap operation.
+///
+/// Yes, I don't like the name. Give me a good one.
```
benaadams, line 14: ConcurrencyController and SlotReturner as just Slot?

src/Nethermind/Nethermind.Core/Threading/ThreadLimiter.cs
```diff
@@ -0,0 +1,46 @@
...
+
+    public bool TryTakeSlot(out SlotReturner returner)
+    {
+        returner = new SlotReturner(this);
+        int newSlot = Interlocked.Decrement(ref _slots);
+        if (newSlot < 1)
+        {
+            Interlocked.Increment(ref _slots);
+            return false;
+        }
```
LukaszRozmej, line 29: As this will fail and return false, way more often than it will return true, maybe this to avoid Decrement/Increment all the time?

```suggestion
        int newSlot = Volatile.Read(ref _slots)
        if (newSlot < 2)
        {
           return false;
        }
        else 
        {
                int newSlot = Interlocked.Decrement(ref _slots);
                if (newSlot < 1)
                {
                   Interlocked.Increment(ref _slots);
                   return false;
                }
        }        
```

asdacap: Yea, that will do

LukaszRozmej (approved):

PR #7712, opened 2024-11-04
Fix null ref in processing stats

benaadams:
## Changes

- `block.Header.GasBeneficiary` can be `null` handle this situation

#### Requires testing

- [x] No


smartprogrammer93 (approved):

benaadams:
> I do feel though that if gas beneficiary is null, then why this at all. its useless

Because 99.9999% of the time it will be set rather than the priority fees just disappearing to nowhere. Doesn't seem worth the additional handling?

PR #7714, opened 2024-11-04
Fix IPC data handling

rubo:
## Changes

Fixed the issue when Nethermind doesn't work with Ethers or web3.js because of line feed handling.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

Tested manually with Ethers, web3.js, and socat.


src/Nethermind/Nethermind.Sockets/IpcSocketMessageStream.cs
```diff
@@ -11,76 +11,69 @@ namespace Nethermind.Sockets;
...
-                EndOfMessage = endOfMessage,
-                CloseStatusDescription = null
-            };
+            endOfMessage = true;
+            buffer[(delimiter + 1)..read].CopyTo(_bufferedData);
+            read = delimiter + 1;
+        }
+        else
+        {
+            endOfMessage = delimiter != -1 || Socket.Available == 0;
```
rubo, line 59: The actual fix is ` || Socket.Available == 0`. The rest is a minor refactoring.

src/Nethermind/Nethermind.JsonRpc.Test/JsonRpcSocketsClientTests.cs
```diff
@@ -81,7 +81,7 @@ static async Task<int> CountNumberOfMessages(Socket socket, CancellationToken to
                     while (true)
                     {
                         ReceiveResult? result = await stream.ReceiveAsync(buffer).ConfigureAwait(false);
-                        if (result?.EndOfMessage == true)
+                        if (result is not null && result.EndOfMessage && (!result.Closed || result.Read != 0))
```
LukaszRozmej, line 84: Should we make `result.EndOfMessage && (!result.Closed || result.Read != 0)` into extension method I see it used 2 times

LukaszRozmej (approved):

src/Nethermind/Nethermind.JsonRpc.Test/JsonRpcSocketsClientTests.cs
rubo: Moved to a method. As it's used only these tests, I don't wanna make it an extension method so as not to clutter the `ReceiveResult` interface.

PR #7718, opened 2024-11-05
[Tests] Fix scenario where nethermind stops before capturing extra logs - convert script to Python

kamilchodola:
Fix scenario where nethermind stops before capturing extra logs

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [X] Other: _Tests_

#### Requires testing

- [ ] Yes
- [X] No

- [ ] Yes
- [X] No

### DIFF
--- a/scripts/waitForSync.py
+++ b/scripts/waitForSync.py
@@ -1,0 +1,64 @@
+import os
+import subprocess
+import sys
+
+network = os.getenv("NETWORK")
+
+bad_logs = {"Exception": 1, "Missing node found!": 1}
+good_logs = {"Processed": 0, "Stats after finishing state": 0}
+required_count = {"Processed": 20, "Stats after finishing state": 1}
+
+if network not in {"joc-mainnet", "joc-testnet", "linea-mainnet", "linea-sepolia"}:
+    good_logs["Synced Chain Head"] = 0
+    required_count["Synced Chain Head"] = 20
+
+container_mapping = {
+    "base-": "sedge-execution-op-l2-client",
+    "op-": "sedge-execution-op-l2-client",
+}
+default_container_name = "sedge-execution-client"
+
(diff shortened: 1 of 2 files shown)


LukaszRozmej (approved):

scripts/waitForSync.py
```diff
@@ -0,0 +1,63 @@
...
+            counter += 1
+            if counter >= 100:
+                print("Exiting after capturing extra logs due to error.")
+                sys.exit(1)
+            continue
+
+        if any(bad_log in line for bad_log in bad_logs):
+            print(f"Error: Found bad log in line: {line.strip()}")
+            found_bad_log = True
+            continue
```
brbrr, line 45: Not sure I understand why it's implemented this way. From my understanding:

When a bad log is found, we'll set the flag to true;
After that, we'll run the for loop 100 more times, and then exit.

Is it used to capture extra logs?

scripts/waitForSync.py
```diff
@@ -0,0 +1,63 @@
+import os
```
brbrr, line 1: Not sure, but was wondering if we could reuse the script from post-merge tests, what do you think? Probably not needed at this point, I guess?

brbrr (approved):
Looks good overal, but let few minor comments for you to decide.

scripts/waitForSync.py
kamilchodola: Yes exactly for this purpose

scripts/waitForSync.py
kamilchodola: I reached the limit of optimizing to various ise cases SH script and python gave me more flexibility and went quickly with rewrite.

I have no issue reusing existing one and keep them identical to some extent!

PR #7725, opened 2024-11-05
Fix RPC "No state available" when using latest block

alexb5dh:
Fixes issue introduced by #7362

### DIFF
--- a/src/Nethermind/Nethermind.Trie/Pruning/OverlayTrieStore.cs
+++ b/src/Nethermind/Nethermind.Trie/Pruning/OverlayTrieStore.cs
@@ -1,6 +1,7 @@
 // SPDX-FileCopyrightText: 2024 Demerzel Solutions Limited
 // SPDX-License-Identifier: LGPL-3.0-only
 
+using System;
 using Nethermind.Core;
 using Nethermind.Core.Crypto;
 using Nethermind.Logging;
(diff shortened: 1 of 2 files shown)


LukaszRozmej (approved):

src/Nethermind/Nethermind.Trie/Pruning/OverlayTrieStore.cs
```diff
@@ -12,10 +13,14 @@ public class OverlayTrieStore(IKeyValueStoreWithBatching? keyValueStore, IReadOn
...
-    public override TrieNode FindCachedOrUnknown(Hash256? address, in TreePath path, Hash256? hash)
+    internal override TrieNode FindCachedOrUnknown(Hash256? address, in TreePath path, Hash256? hash, bool isReadOnly)
     {
-        TrieNode node = base.FindCachedOrUnknown(address, in path, hash);
-        return node.NodeType == NodeType.Unknown ? store.FindCachedOrUnknown(address, in path, hash) : node;
+        ArgumentNullException.ThrowIfNull(hash);
+
+        TrieNode node = base.FindCachedOrUnknown(address, in path, hash, isReadOnly);
+        return node.NodeType == NodeType.Unknown
+            ? store.FindCachedOrUnknown(address, in path, hash) // no need to pass isReadOnly - IReadOnlyTrieStore overrides it as true
```
LukaszRozmej, line 22: Maybe we should pass it for completeness?

alexb5dh: Method with additional `isReadOnly` parameter is internal and not present in `IReadOnlyTrieStore` interface. This kind of makes sense, so would prefer to not add it.

PR #7688, opened 2024-10-30
Split storage ranges at the end of snap ranges sync to parallelize execution

damian-orzechowski:
## Changes

At the end of state ranges sync, when only large storage trie is left to be synced, snap batches will be processed sequentially. This change adds a mechanism to detect such scenario and partition the space left to sync in a similar way that account ranges are partition at the start of snap sync. Parallel execution of the batches improves last phase - usually manifesting itself as "stuck" at 100% progress.

#### What types of changes does your code introduce?

- [ ] Bugfix (a non-breaking change that fixes an issue)
- [x] New feature (a non-breaking change that adds functionality)
- [ ] Breaking change (a change that causes existing functionality not to work as expected)
- [x] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

#### If yes, did you write tests?

- [ ] Yes
- [ ] No

#### Notes on testing

_Optional. Remove if not applicable._

#### Requires documentation update

- [ ] Yes
- [x] No

_If yes, link the PR to the docs update or the issue with the details labeled `docs`. Remove if not applicable._

#### Requires explanation in Release Notes

- [ ] Yes
- [x] No

_If yes, fill in the details here. Remove if not applicable._

PR #7733, opened 2024-11-06
Split storage ranges to parallelize execution

damian-orzechowski:
## Changes

At the end of state ranges sync, when only large storage trie is left to be synced, snap batches will be processed sequentially. This change splits the range left to be requested for a given storage trie, if the processed data is less than 50% of requested range (maybe this shouldn't be 50%). The aim is to speed up processing of large storage tries, especially towards the end of snap ranges phase, when progress gets "stuck" at 100%. This is to aid the same problem as #7688, but with a slightly different approach.

Tested on OP-Mainnet - snap sync time:
- Master: ~1:42 minutes (1h "stuck" at 100%)
- After change: ~1:11 minutes (28min stuck at 93%)

| Chain  | Master | PR |
| ------------- | ------------- | ------------- |
| op-mainnet  1 | 102 min  | 71 min |
| op-mainnet  2 | 138 min  | 90 min |
| op-mainnet  3 | 180 min  | 60 min |
| mainnet  1 | 71 min | 66 min |
| mainnet  2 | 69 min | 65 min |

- [x] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


src/Nethermind/Nethermind.Synchronization/SnapSync/ProgressTracker.cs
```diff
@@ -342,6 +343,54 @@ public void EnqueueStorageRange(StorageRange storageRange)
...
+            if (lastProcessedHash > limitHash)
+                return;
+
+            limitHash ??= Keccak.MaxValue;
+
+            UInt256 limit = new UInt256(limitHash.Value.Bytes, true);
+            UInt256 lastProcessed = new UInt256(lastProcessedHash.Bytes, true);
+
+            var fullRange =
+                limit - (startingHash.HasValue ? new UInt256(startingHash.Value.Bytes, true) : new UInt256(0));
```
LukaszRozmej, line 357:
```suggestion
                limit - (startingHash.HasValue ? new UInt256(startingHash.Value.Bytes, true) : UInt256.Zero);
```

src/Nethermind/Nethermind.Synchronization/SnapSync/ProgressTracker.cs
```diff
@@ -29,6 +29,7 @@ public class ProgressTracker : IDisposable
         public const int HIGH_STORAGE_QUEUE_SIZE = STORAGE_BATCH_SIZE * 100;
         private const int CODES_BATCH_SIZE = 1_000;
         public const int HIGH_CODES_QUEUE_SIZE = CODES_BATCH_SIZE * 5;
+        private const uint StorageRangeSplitFactor = 2;
```
LukaszRozmej, line 32: have you experimented with more radical splitting factors? 4? 8? 16? 32?

LukaszRozmej (approved):
Please wait for @asdacap review too

src/Nethermind/Nethermind.Synchronization/SnapSync/ProgressTracker.cs
```diff
@@ -342,6 +343,54 @@ public void EnqueueStorageRange(StorageRange storageRange)
...
+
+            limitHash ??= Keccak.MaxValue;
+
+            UInt256 limit = new UInt256(limitHash.Value.Bytes, true);
+            UInt256 lastProcessed = new UInt256(lastProcessedHash.Bytes, true);
+
+            var fullRange =
+                limit - (startingHash.HasValue ? new UInt256(startingHash.Value.Bytes, true) : new UInt256(0));
+
+            if (lastProcessed < fullRange / StorageRangeSplitFactor)
```
asdacap, line 359: Not `lastProcessed < (fullRange / StorageRangeSplitFactor) + startingHash` ?

asdacap:
Can add unit test?

src/Nethermind/Nethermind.Synchronization/SnapSync/ProgressTracker.cs
damian-orzechowski: Changes

damian-orzechowski:
> Can add unit test?

Added

asdacap (approved):

LukaszRozmej (commented):
Would be also good to somehow indicate in the logs that this is in progress, maybe add a global counter of split request pending/completed and output it in the log? So the user has indication of progress.

src/Nethermind/Nethermind.Synchronization/SnapSync/ProgressTracker.cs
```diff
@@ -471,5 +542,55 @@ public void Dispose()
...
+            }
+
+            public void Decrement(ValueHash256 key, bool removeFromOwner)
+            {
+                lock (_lock)
+                {
+                    if (--Counter == 0 && removeFromOwner)
+                        owner.TryRemove(key, out _);
+                }
+            }
```
LukaszRozmej, line 574: Does counter need to be under same lock as safe execution?
If not better to do it on `Interlocked`

src/Nethermind/Nethermind.Synchronization/SnapSync/ProgressTracker.cs
```diff
@@ -455,6 +510,22 @@ private bool TryDequeNextSlotRange(out StorageRange item)
...
+
+        public void IncrementStorageRangeLock(ValueHash256 accountPath, uint number = 1)
+        {
+            _storageRangeLocks.AddOrUpdate(accountPath, k => new StorageRangeLock(number, _storageRangeLocks), (k, old) => old.Increment(number));
+        }
+
+        public void IncrementStorageRangeLockIfExists(ValueHash256 accountPath, uint number = 1)
+        {
+            if (!_storageRangeLocks.TryGetValue(accountPath, out IStorageRangeLock lockInfo)) return;
+            lockInfo.Increment(number);
```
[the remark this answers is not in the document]
Scooletz: What @LukaszRozmej said above. What if you call `IncrementStorageRangeLockIfExists` you get no `lockInfo` and immediately it's followed by `IncrementStorageRangeLock` that does the update? Does it mean that something is broken here?

src/Nethermind/Nethermind.Synchronization/SnapSync/ProgressTracker.cs
```diff
@@ -471,5 +542,55 @@ public void Dispose()
                 range?.Dispose();
             }
         }
+
+        public interface IStorageRangeLock
+        {
+            IStorageRangeLock Increment(uint value = 1);
+            void Decrement(ValueHash256 key, bool removeFromOwner);
```
Scooletz, line 549: Why `key` is needed here? Should it not be captured in the `ctor` of the lock, so that we just `Decrement(bool removeFromOwner)`. You get the lock by key so the key should not be needed here I think.

src/Nethermind/Nethermind.Synchronization/SnapSync/ProgressTracker.cs
```diff
@@ -471,5 +542,55 @@ public void Dispose()
                 range?.Dispose();
             }
         }
+
+        public interface IStorageRangeLock
+        {
+            IStorageRangeLock Increment(uint value = 1);
```
Scooletz, line 548: Why increment returns the lock object? My feeling is that it's to satisfy the `ConcurrentDictionary`. I'd keep it simple and make it void.

src/Nethermind/Nethermind.Synchronization/SnapSync/SnapProvider.cs
```diff
@@ -194,44 +194,42 @@ public AddRangeResult AddStorageRange(StorageRange request, SlotsAndProofs respo
...
 
-                    _progressTracker.EnqueueAccountRefresh(pathWithAccount, startingHash);
+                    _progressTracker.EnqueueAccountRefresh(pathWithAccount, startingHash, hashLimit);
                 }
 
                 return result;
             }
             finally
             {
+                rangeLock?.Decrement(pathWithAccount.Path, canRemoveFurtherLock);
```
Scooletz, line 232: Why do we need to take care of whether a lock should be removed or not? Should it not be the case that when the counter hits 0, we always remove?

Scooletz (commented):
Provided some comments. This seems similar to Paprika's ref counting.
I don't know how often the lock concurrent dictionary is used, but if it is not contended we could replace the concurrent with a lock (potentially, RWLSlim) and use some global atomicity. It might be the case that due to construction of the snap sync, we know that once the account is processed, it will never appear again. If this is the case it can be leveraged and used to know when to remove a thing from the dictionary (current condition)

src/Nethermind/Nethermind.Synchronization/SnapSync/ProgressTracker.cs
```diff
@@ -471,5 +542,55 @@ public void Dispose()
...
+            void Decrement(ValueHash256 key, bool removeFromOwner);
+            void ExecuteSafe(Action action);
+        }
+
+        public class StorageRangeLock(uint counter, ConcurrentDictionary<ValueHash256, IStorageRangeLock> owner)
```
asdacap, line 553: I'd just use a sharded lock to avoid such complications.
Also, you can just use `_pivot.UpdatedStorages.Add(pathWithAccount.Path);` to the splitted storage so that healing will fix it. But it limits the number of storage that can be splitted though.

src/Nethermind/Nethermind.Synchronization/SnapSync/SnapProviderHelper.cs
```diff
@@ -141,19 +139,21 @@ public static (AddRangeResult result, bool moreChildrenToRight, ProgressTracker.
...
-                return (AddRangeResult.DifferentRootHash, true, null);
+                return (AddRangeResult.DifferentRootHash, true);
             }
 
-            var storageLock = progressTracker.GetLockObjectForPath(accountHash);
-            storageLock.ExecuteSafe(() =>
+            // This will work if all StorageRange requests share the same AccountWithPath object which seems to be the case.
+            // If this is not true, StorageRange request should be extended with a lock object.
+            // That lock object should be shared between all other StorageRange requests for same account.
+            lock(account.Account)
```
LukaszRozmej, line 150: Main change here!

damian-orzechowski: The main point of having this complexity around locking was to lock only when necessary - that is when split is done and there us a chance of actually having parallel processing of storage ranges for same account. If the overhead is negligible, I'm all for it - it's the simplest.

damian-orzechowski: On having a conversation with @Scooletz yesterday, there could be a way to optimize it easily, if there is significant overhead. Stitching only happens when there are boundary proof nodes - so can lock only if that happens.

damian-orzechowski:
This test run shows this may bring missing nodes on storage - cannot merge yet
https://github.com/NethermindEth/nethermind/actions/runs/12183618666

src/Nethermind/Nethermind.Synchronization/SnapSync/SnapProviderHelper.cs
LukaszRozmej: lock is super cheap compared to I/O so locking even when not necessary shouldn't degrade performance

asdacap: Should be fine. That said, this code path is run quite a lot, might wannna double check if the lock is significant.

LukaszRozmej: If there is no contention then lock is very fast, if there is contention, well we still have to lock. Keep in mind in the previous solution we locked on potential contention while also accessed ConcurrentDictionary one or few times.

src/Nethermind/Nethermind.Synchronization/SnapSync/SnapProviderHelper.cs
```diff
@@ -138,14 +139,19 @@ public static (AddRangeResult result, bool moreChildrenToRight) AddStorageRange(
...
                 return (AddRangeResult.DifferentRootHash, true);
             }
 
-            StitchBoundaries(sortedBoundaryList, tree.TrieStore);
-
-            tree.Commit(writeFlags: WriteFlags.DisableWAL);
+            // This will work if all StorageRange requests share the same AccountWithPath object which seems to be the case.
+            // If this is not true, StorageRange request should be extended with a lock object.
+            // That lock object should be shared between all other StorageRange requests for same account.
+            lock (account.Account)
```
Scooletz, line 150: I don't like the `account` based locking and using the object that might be elsewhere locked or used for different purposes. Locking should have specific scoping and should be reasoned about locally.

Another take is that this lock is needed, correct me @damian-orzechowski if I'm wrong, only if we know that there's no pending requests. We could extract the check from `StitchBoundaries` and assert it explicitly before the lock.

Scooletz (changes requested):

src/Nethermind/Nethermind.Synchronization/SnapSync/SnapProviderHelper.cs
LukaszRozmej: we could create artificial object and pass it around, but I don't really see the point other than increasing the memory pressure and complicating the code. We added comments! :)

damian-orzechowski: Done some refactoring here #7930. The problem is that passing around the whole `StorageRange` request means also passing an index of the actual account processed (accounts are processed one at a time), so it doesn't look very nice.

Scooletz (approved):

damian-orzechowski:
More issues found - do not merge:
https://github.com/NethermindEth/nethermind/actions/runs/12595664462/job/35105410956

Scooletz:
Not sure if ready for a review. If yes, re-request pls.

PR #7735, opened 2024-11-06
Shutter: p2p logging

Marchhill:
## Changes

- Enable libp2p logging from config for easier debugging

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


flcl42 (approved):

flcl42:
It's done not in canonical way I guess, but will simplify a lot debugging related issues

flcl42:
We can log all as is and set filtering in config to display only what we need, but I doubt it's easy to describe to users

Marchhill:
> We can log all as is and set filtering in config to display only what we need, but I doubt it's easy to describe to users

I wouldn't expect any users to enable this so I hid in docs

PR #7744, opened 2024-11-11
t8n test

yerke26:
added t8n tests, fixed t8n code

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [x] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Remarks


yerke26:
@rubo should we add it to github actions?

src/Nethermind/Nethermind.JsonRpc/Data/ReceiptForRpc.cs
```diff
@@ -29,7 +29,7 @@ public ReceiptForRpc(Hash256 txHash, TxReceipt receipt, TxGasInfo gasInfo, int l
             BlobGasPrice = gasInfo.BlobGasPrice;
             From = receipt.Sender;
             To = receipt.Recipient;
-            ContractAddress = receipt.ContractAddress;
+            ContractAddress = receipt.ContractAddress ?? Address.Zero;
```
LukaszRozmej, line 32: Why would we serialize `Address.Zero` for transactions that don't create contracts?

This will also affect all the RPC method, are we sure we want to do it?

If we don't want to change RPC methods, but we want to change it here, you can inherit `ReceiptForRpc` and add this logic in the derived class.

tools/Evm/Evm.Test/T8nTests.cs
```diff
@@ -0,0 +1,260 @@
...
+    {
+        Execute(
+            new("testdata/5/", "alloc.json", "env.json", "txs.json", "Byzantium", "0x80"),
+            new OutputParams(alloc: "stdout", result: "stdout"),
+            expectedExitCode: 0,
+            expectedOutputFile: "testdata/5/exp.json");
+    }
+
+    [Test]
+    public void Test6()
```
LukaszRozmej, line 81: Can you name tests in some sensible way? I see it is based on `testdata` so `Test13_txs_London`?

tools/Evm/Evm/T8n/T8nInputProcessor.cs
```diff
@@ -95,6 +95,6 @@ private static (ISpecProvider, IReleaseSpec) GetSpec(T8nCommandArguments argumen
             specProvider.UpdateMergeTransitionInfo(env.CurrentNumber, 0);
         }
 
-        return (specProvider, overridableReleaseSpec);
+        return (specProvider, overridableReleaseSpec, spec);
```
LukaszRozmej, line 98: why do you need to differentiate between 2x release specs?

tools/Evm/Evm/T8n/T8nTest.cs
```diff
@@ -16,9 +16,10 @@
 
 namespace Evm.T8n;
 
-public class T8nTest(IReleaseSpec spec, ISpecProvider specProvider, Address currentCoinbase)
+public class T8nTest(IReleaseSpec spec, IReleaseSpec overridableReleaseSpec, ISpecProvider specProvider, Address currentCoinbase)
 {
     public IReleaseSpec Spec { get; set; } = spec;
+    public IReleaseSpec OverridableReleaseSpec { get; set; } = overridableReleaseSpec;
```
LukaszRozmej, line 22: confusing

LukaszRozmej (commented):
Would be better if we could have `testdata` as submodule, discussed options on slack. If not possible, it is acceptable to leave as is.

tools/Evm/Evm/Evm.csproj
```diff
@@ -5,6 +5,7 @@
         <TargetFramework>net9.0</TargetFramework>
         <ImplicitUsings>enable</ImplicitUsings>
         <Nullable>enable</Nullable>
+        <Version>1.0.0.0</Version>
```
LukaszRozmej, line 8: Are 4 parts needed? We normally go with semantic versioning which has major.minor.patch

tools/Evm/Evm/T8n/T8nInputReader.cs
```diff
@@ -76,7 +76,11 @@ private static InputData ReadStdInput()
         using var reader = new StreamReader(Console.OpenStandardInput());
         try
         {
-            return EthereumJsonSerializer.Deserialize<InputData>(reader.ReadToEnd());
+            var jsonText = reader.ReadToEnd();
+            TransactionMetaDataWrapper txsMetaDataWrapper = EthereumJsonSerializer.Deserialize<TransactionMetaDataWrapper>(jsonText);
```
LukaszRozmej, line 80: Do we have to `ReadToEnd`? Can we deserialize from `Stream` or `Reader` as we read it?

LukaszRozmej:
we should revisit this, fix and make it work
@claude review

LukaszRozmej:
Keeping this open as a draft. The value here is real — `tools/Evm/T8n` has no test coverage at all on master — but this needs a redo rather than a rebase.

**What has moved since this was opened:**

- `IBlockhashStore.ApplyBlockhashStateChanges` now takes `(BlockHeader, IReleaseSpec)`, not the `ITxTracer` this PR adds. There is also a `BlockAccessListManager` implementation of it now.
- `VirtualMachine` has been restructured, so the `InstructionSelfDestruct` `ReportAccountRead` hunk no longer applies where it sits.
- 92 of the 188 changed files are test fixtures, which is what makes the conflict resolution expensive.

**What is still worth salvaging:**

- The `tools/Evm/Evm.Test` project and `T8nTests.cs` harness — this is the part we actually want, and it is only ~300 lines.
- The two tracing-fidelity fixes the tests were written to expose (EIP-2935 blockhash store writes not surfacing through op-level storage tracing, and the missing account read on the SELFDESTRUCT inheritor). Both look like genuine gaps; they should be re-derived against current `master` and land with the tests that catch them.

**Still open from the original review:** whether `testdata` should be a submodule rather than committed fixtures. That decision should be made before the fixtures are re-added, not after.

Leaving in draft as the starting point for that work.

PR #7753, opened 2024-11-13
Increase timeout to try fix flaky test

LukaszRozmej:


damian-orzechowski (commented):
I was looking at this specific test failing: `HealBigSqueezedRandomTree` and to me it's not due to this particular timeout. If it failed due to this timeout, it would fail on assertion related to the tree - the healing would not be complete hence remote and local tries would be different.
It fails on this condition: `Assert.That(data.RequestedNodesCount, Is.LessThan(accounts.Count / 2));`, so tree is healed correctly. To me, we should change this timeout:
https://github.com/NethermindEth/nethermind/blob/bf5369448f2ee19db260c46d4833a808058163bd/src/Nethermind/Nethermind.Synchronization.Test/FastSync/StateSyncFeedTestsBase.cs#L121
As commented - too short timeout can cause the peer not to be allocated and all nodes re-added as pending. This increased the `RequestedNodesCount` and fails test.

PR #7764, opened 2024-11-16
Add Labels to ChainParameters to Enable Hardforks via Single Timestamp

### LINKED ISSUE #7738: Add Labels to chainspec to activate hardforks at once.
Right now to activate Dencun we need to add to chainspec:

```
    "eip4844TransitionTimestamp": "0x65687fd0",
    "eip4788TransitionTimestamp": "0x65687fd0",
    "eip1153TransitionTimestamp": "0x65687fd0",
    "eip5656TransitionTimestamp": "0x65687fd0",
    "eip6780TransitionTimestamp": "0x65687fd0"
```

We want to keep fine-grained eip transitions, but we would also like to specify it all at once with some label:

```
    "dencun": "0x65687fd0"
```

or 

```
    "cancun": "0x65687fd0"
```

It would be great if it worked both ways - so `cancun` flag could be checked, even if single eip timestamps would be set in config. If the activations conflict with each other some exception should be thrown.

It should work as that for all forks.

LuisDi98:
This PR introduces functionality to allow grouped EIP timestamps (e.g., dencun, cancun) in ChainParameters. This simplifies the process of activating multiple EIPs simultaneously through a single label, while preserving the ability to handle individual EIP timestamps.

Fixes #7738

### Changes
- Added DencunTransitionTimestamp and CancunTransitionTimestamp to ChainParameters class.
- Implemented ValidateNoTimestampConflicts to detect and validate timestamp conflicts between labels and individual EIPs.
- Updated ChainSpecParamsJson to support grouped label serialization/deserialization.
- Added tests in ChainParametersTests to verify label functionality and conflict detection.

### Types of Changes
What types of changes does your code introduce?

- New feature (a non-breaking change that adds functionality)

### Testing
Requires testing
- Yes

If yes, did you write tests?
- Yes

#### Notes on testing
Added unit tests in ChainParametersTests to verify:

- Setting and getting label timestamps.
- Conflict detection for mismatched individual EIP timestamps.
- Valid behavior when all timestamps match.

### Documentation
Requires documentation update
- No

Requires explanation in Release Notes
- Yes

### Release Notes:
- Added support for grouped EIP timestamps (dencun, cancun) in ChainParameters for simplifying hardfork activation.

### DIFF
--- a/src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainParameters.cs
+++ b/src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainParameters.cs
@@ -1,6 +1,8 @@
 // SPDX-FileCopyrightText: 2022 Demerzel Solutions Limited
 // SPDX-License-Identifier: LGPL-3.0-only
 
+using System;
+using System.Collections.Generic;
 using Nethermind.Core;
 using Nethermind.Core.Crypto;
 using Nethermind.Int256;
@@ -10,17 +12,28 @@ namespace Nethermind.Specs.ChainSpecStyle;
 public class ChainParameters
 {
 
-    private ulong? _dencunTransitionTimestamp;
-    private ulong? _cancunTransitionTimestamp;
+    private static readonly Dictionary<string, Func<ChainParameters, ulong?>> HardforkToEips = new()
+    {
+        { "Dencun", cp => cp.Eip4844TransitionTimestamp },
+        { "Cancun", cp => cp.Eip4788TransitionTimestamp },
+        { "EIP-1153", cp => cp.Eip1153TransitionTimestamp },
+        { "EIP-5656", cp => cp.Eip5656TransitionTimestamp },
+        { "EIP-6780", cp => cp.Eip6780TransitionTimestamp }
+    };
+
+
+    public static Func<ChainParameters, ulong?>? GetHardforkMapping(string hardfork)
+    {
+        return HardforkToEips.TryGetValue(hardfork, out var mapping) ? mapping : null;
+    }
 
     public ulong? DencunTransitionTimestamp
     {
         get
         {
-            if (Eip4844TransitionTimestamp == Eip4788TransitionTimestamp &&
-                Eip4788TransitionTimestamp == Eip1153TransitionTimestamp &&
-                Eip1153TransitionTimestamp == Eip5656TransitionTimestamp &&
-                Eip5656TransitionTimestamp == Eip6780TransitionTimestamp)
+            if (AreHardforkTimestampsEqual(
+                "Dencun", "Cancun", "EIP-1153",
+                "EIP-5656", "EIP-6780"))
             {
                 return Eip4844TransitionTimestamp;
             }
@@ -29,7 +42,6 @@ public class ChainParameters
         }
         set
         {
-            _dencunTransitionTimestamp = value;
             Eip4844TransitionTimestamp = value;
             Eip4788TransitionTimestamp = value;
             Eip1153TransitionTimestamp = value;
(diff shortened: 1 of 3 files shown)


src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainParameters.cs
```diff
@@ -9,6 +9,76 @@ namespace Nethermind.Specs.ChainSpecStyle;
...
+                Eip5656TransitionTimestamp == Eip6780TransitionTimestamp)
+            {
+                return Eip4844TransitionTimestamp;
+            }
+
+            return null;
+        }
+        set
+        {
+            _dencunTransitionTimestamp = value;
```
LukaszRozmej, line 32: Potentially we could have `_dencunTransitionTimestamp` and individual EIP's out of sync, so maybe always calculate it and don't have a separate field?

LukaszRozmej, line 32: Potentially we could have `_dencunTransitionTimestamp` and individual EIP's out of sync, so maybe always calculate it and don't have a separate field? Then you also don't need `Validate` methods

src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainParameters.cs
```diff
@@ -9,6 +9,76 @@ namespace Nethermind.Specs.ChainSpecStyle;
...
+    private ulong? _cancunTransitionTimestamp;
+
+    public ulong? DencunTransitionTimestamp
+    {
+        get
+        {
+            if (Eip4844TransitionTimestamp == Eip4788TransitionTimestamp &&
+                Eip4788TransitionTimestamp == Eip1153TransitionTimestamp &&
+                Eip1153TransitionTimestamp == Eip5656TransitionTimestamp &&
+                Eip5656TransitionTimestamp == Eip6780TransitionTimestamp)
```
LukaszRozmej, line 23: You could potentially use more declarative style of `Expression` Api to map individual properties to Hardfork property, that might be useful outside of this class too?

LukaszRozmej (changes requested):
We need to support all hardforks

src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainParameters.cs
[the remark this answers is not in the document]
LuisDi98: I created the helper method called "AreHardforkTimestampsEqual" in order to accomplish this one, which compares it instead of using the ifs inside each hardfork.

src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainParameters.cs
LuisDi98: You are so right... I made the change using the helper method previously suggested and if it's true, we assign one of those timestamps values so it calculates it dynamically. (code lines 32-42 and 56-65)

src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainParameters.cs
LuisDi98: Sure! In this refactor commit I created a dictionary at lines 15-22 so we can access only using the key tag, also a public static method GetHardforkMapping in case we need to get any value from this dictionary from any other classes outside

src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainParameters.cs
```diff
@@ -9,6 +11,84 @@ namespace Nethermind.Specs.ChainSpecStyle;
...
+        }
+        set
+        {
+            Eip4844TransitionTimestamp = value;
+            Eip4788TransitionTimestamp = value;
+            Eip1153TransitionTimestamp = value;
+            Eip5656TransitionTimestamp = value;
+            Eip6780TransitionTimestamp = value;
+        }
+    }
```
LukaszRozmej, line 74: Dencun and Cancun are same thing, confusing.

src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainParameters.cs
```diff
@@ -9,6 +11,84 @@ namespace Nethermind.Specs.ChainSpecStyle;
...
 {
+
+    private static readonly Dictionary<string, Func<ChainParameters, ulong?>> HardforkToEips = new()
+    {
+        { "Dencun", cp => cp.Eip4844TransitionTimestamp },
+        { "Cancun", cp => cp.Eip4788TransitionTimestamp },
+        { "EIP-1153", cp => cp.Eip1153TransitionTimestamp },
+        { "EIP-5656", cp => cp.Eip5656TransitionTimestamp },
+        { "EIP-6780", cp => cp.Eip6780TransitionTimestamp }
+    };
```
LukaszRozmej, line 22: I was thinking of something more like:

```
    public static readonly IReadOnlyList<Expression<Func<ChainParameters, ulong?>>> DencunEIPs =
    [
        cp => cp.Eip4844TransitionTimestamp,
        cp => cp.Eip4788TransitionTimestamp,
        cp => cp.Eip1153TransitionTimestamp,
        cp => cp.Eip5656TransitionTimestamp,
        cp => cp.Eip6780TransitionTimestamp
    ];
```

If we had same interface for `ChainParameters` and `ChainSpecParamsJson` that could be potentially reusable.

But that maybe an overkill.

LukaszRozmej (changes requested):
@LuisDi98  I think recent changes are step in wrong direction. Would prefer to avoid strings. Added complexity isn't bringing anything at the moment.

stdevMac:
Closing — changes-requested feedback was never addressed, and the branch (master-on-fork) has drifted for ~14 months. If grouped hardfork timestamps (#7738) are still desired, please reopen with a rebased branch.

PR #7766, opened 2024-11-19
Exclude empty proofs; 0x0 for 0 instead of 0x

### LINKED ISSUE #7759: Nethermind includes extraneous `0x` values in `eth_getProof` responses, breaking cryptographic consistency.
**Description**
`eth_getProof` includes erroneous `0x` values that don't make sense

**Steps to Reproduce**
Calling this `eth_getProof` request on Nethermind:
```
curl <Nethermind node URL>  \
  -X POST \
  --data '{"method":"eth_getProof","params":["0x3a10dc1a145da500d5fba38b9ec49c8ff11a981f",["0x0000000000000000000000000000000000000000000000000000000000000003","0x0000000000000000000000000000000000000000000000000000000000000004","0x0ddc93f0dfd4cec69030b20fc3c9037ae4a8a3151a4c66b05fa74272b3a48df8", "0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc","0x90375f0b936c9e37c8e89c250b2c970b724107b72ed642aa819ba7b257516d30","0xb53127684a568b3173ae13b9f8a6016e243e63b6e8ee1178d6a717850b5d6103","0xfd21a1ac9a14dff647460ce8ad2ccecb794a59a4cfbb8678b1f9900a6a99551f","0xfd21a1ac9a14dff647460ce8ad2ccecb794a59a4cfbb8678b1f9900a6a995520","0xfd21a1ac9a14dff647460ce8ad2ccecb794a59a4cfbb8678b1f9900a6a995521","0xfd21a1ac9a14dff647460ce8ad2ccecb794a59a4cfbb8678b1f9900a6a995523"],"0x143212F"],"id":1,"jsonrpc":"2.0"}' |jq
 ```

**Actual behavior**
The proof nodes array includes multiple `0x` entries, representing empty nodes. These extra empty nodes imply an unnecessary parent-child relationship between empty nodes, which breaks the cryptographic consistency expected in exclusion proofs.

**Expected behavior**
The proof nodes array should exclude redundant 0x entries in exclusion proofs, only including relevant nodes that provide a cryptographically verifiable path. Correct exclusion proofs from other clients do not include these extra empty nodes.

**Screenshots**
image

 - Operating System: Ubuntu 24.04
 - Version: `Nethermind/v1.29.1`

 - Installation Method:Github release
 - Consensus Client: Lighthouse

### PR DESCRIPTION
**Pull request by flcl42**
Resolves #7759 

Remove the difference in responses of proof API

## Changes

- Exclude empty proofs for not found from results
- No leading zeros for proof value
- 0 means 0x0 instead of 0x (null/empty array)

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### CODE REVIEW

**Review by MarekM25 — COMMENTED** [review]
https://github.com/NethermindEth/nethermind/issues/7759 @flcl42 qq: does it fix this issue or it is something different?

**Comment by flcl42** [comment]
> #7759 @flcl42 qq: does it fix this issue or it is something different?

Yep, seems like that. We are getting empty storage proof values and want not to return em anymore

File: src/Nethermind/Nethermind.Core/Extensions/Bytes.cs [hunk 1]
```diff
@@ -654,7 +658,7 @@ private static string ByteArrayToHexViaLookup32(byte[] bytes, bool withZeroX, bo
             int length = bytes.Length * 2 + (withZeroX ? 2 : 0) - leadingZerosFirstCheck;
             if (skipLeadingZeros && length == (withZeroX ? 2 : 0))
             {
-                return withZeroX ? "0x0" : "0";
+                return withZeroX ? ZeroHexValue : "0";
```
LukaszRozmej, line 661:
```suggestion
                return withZeroX ? ZeroHexValue : ZeroValue;
```

src/Nethermind/Nethermind.Core/JsonConverters/NullableByteReadOnlyMemoryConverter.cs
```diff
@@ -5,9 +5,9 @@
 using System.Text.Json;
 using System.Text.Json.Serialization;
 
-namespace Nethermind.Serialization.Json;
+namespace Nethermind.Core.JsonConverters;
```
LukaszRozmej, line 8: Why move projects and namespaces?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Core/JsonConverters/NullableByteReadOnlyMemoryConverter.cs
flcl42: Seems like those converters were moved to Core to solve circular dependency, so they'd got namespace not matching core project. I've just grouped them into a directory inside Core for convenience and changed namespace to match the structure. But I see the point, which is to have them all in one namespace. Is it critical? Seems like it creates unnecessary convention

flcl42: Anyway moved back so people will not lose converters bc of an additional ns

smartprogrammer93 (approved):

PR #7773, opened 2024-11-20
Support the deprecated `.cfg` extension in `-c` option

rubo:
## Changes

- Added backward compatibility for uncommon use cases specifying the `.cfg` extension in the `-c` option. For instance, `-c mainnet.cfg` instead of `-c mainnet`. In this case, Nethermind will handle it with a warning not to use it.
- Added warnings for deprecated command line options.
- Fixed log levels of plugin loader from "warn" to "info".

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

#### Notes on testing

Tested manually

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

To be added later.

### DIFF
--- a/src/Nethermind/Nethermind.Runner/Program.cs
+++ b/src/Nethermind/Nethermind.Runner/Program.cs
@@ -403,6 +403,15 @@ IConfigProvider CreateConfigProvider(ParseResult parseResult)
                 }
             }
         }
+        // For backward compatibility. To be removed in the future.
+        else if (Path.GetExtension(configFile).Equals(".cfg", StringComparison.Ordinal))
+        {
+            var name = Path.GetFileNameWithoutExtension(configFile)!;
+
+            configFile = $"{configFile[..^4]}.json";
+
+            logger.Warn($"'{name}.cfg' is deprecated. Use '{name}' instead.");
+        }
     }
     else
     {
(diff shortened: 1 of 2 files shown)


MarekM25 (approved):
I assume it was tested manually ;)

src/Nethermind/Nethermind.Runner/Program.cs
```diff
@@ -403,6 +426,15 @@ IConfigProvider CreateConfigProvider(ParseResult parseResult)
                 }
             }
         }
+        // For backward compatibility. To be removed in the future.
+        else if (Path.GetExtension(configFile).Equals(".cfg", StringComparison.Ordinal))
+        {
+            var name = Path.GetFileNameWithoutExtension(configFile)!;
+
+            configFile = $"{configFile[..^4]}.json";
```
LukaszRozmej, line 434:
```suggestion
            configFile = $"{name}.json";
```

LukaszRozmej (approved):

marcindsobczak (approved):

src/Nethermind/Nethermind.Runner/Program.cs
rubo: @LukaszRozmej At this point, `configFile` is the full path, while `name` is just the file name. So they're different.

PR #7780, opened 2024-11-21
Run verifytrie only once on state sync finished

asdacap:
- Not sure why network without CL run verify trie multiple time forever, but this should fix it.

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Synchronization/FastSync/VerifyStateOnStateSyncFinished.cs
+++ b/src/Nethermind/Nethermind.Synchronization/FastSync/VerifyStateOnStateSyncFinished.cs
@@ -25,8 +25,6 @@ public class VerifyStateOnStateSyncFinished(
 {
     private readonly ILogger _logger = logManager.GetClassLogger<VerifyStateOnStateSyncFinished>();
 
-    bool _wasRun = false;
-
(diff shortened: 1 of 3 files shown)


src/Nethermind/Nethermind.Synchronization/FastSync/VerifyStateOnStateSyncFinished.cs
```diff
@@ -25,13 +25,18 @@ public class VerifyStateOnStateSyncFinished(
...
+
     public void Start()
     {
         treeSync.SyncCompleted += TreeSyncOnOnVerifyPostSyncCleanup;
     }
 
     private void TreeSyncOnOnVerifyPostSyncCleanup(object? sender, ITreeSync.SyncCompletedEventArgs evt)
     {
+        if (_wasRun) return;
+        _wasRun = true;
```
LukaszRozmej, line 38: Maybe better just to deregister from handler?
```suggestion
        treeSync.SyncCompleted -= TreeSyncOnOnVerifyPostSyncCleanup;
```

benaadams (approved):

PR #7787, opened 2024-11-21
ParallelUnbalancedWork for efficient unbalanced parallel loops

benaadams:
## Changes

- Added the `ParallelUnbalancedWork` class to efficiently execute parallel loops, handling unbalanced workloads.
- Implemented static For methods to support parallel execution with and without thread-local data, including initialization and finalization functions.
- Utilized thread pooling and a shared counter `SharedCounter` to distribute iterations among threads dynamically.
- Aimed to optimize performance in scenarios where the workload per iteration is uneven, ensuring better resource utilization and reduced execution time; rather than the main thread `.Wait`ing for background threads to complete
- Less allocations than `Parallel.For`

| Method             | Mean     | Error    | StdDev   | Ratio | Allocated | Alloc Ratio |
|------------------- |---------:|---------:|---------:|------:|----------:|------------:|
| ParallelFor        | 543.9 ms | 12.13 ms | 18.15 ms |  1.00 |  19.21 KB |        1.00 |
| ParallelForEach    | 488.5 ms | 12.73 ms | 17.43 ms |  0.90 |  23.12 KB |        1.20 |
| UnbalancedParallel | 413.4 ms |  1.15 ms |  1.73 ms |  0.76 |    5.00 KB |        0.26 |

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
@@ -98,18 +98,13 @@ public sealed class BlockCachePreWarmer(ReadOnlyTxProcessingEnvFactory envFactor
         {
             if (spec.WithdrawalsEnabled && block.Withdrawals is not null)
             {
-                int progress = 0;
-                Parallel.For(0, block.Withdrawals.Length, parallelOptions,
-                    _ =>
+                ParallelUnbalancedWork.For(0, block.Withdrawals.Length, parallelOptions,
+                    i =>
                     {
                         IReadOnlyTxProcessorSource env = _envPool.Get();
-                        int i = 0;
                         try
                         {
                             using IReadOnlyTxProcessingScope scope = env.Build(stateRoot);
-                            // Process withdrawals in sequential order, rather than partitioning scheme from Parallel.For
-                            // Interlocked.Increment returns the incremented value, so subtract 1 to start at 0
-                            i = Interlocked.Increment(ref progress) - 1;
                             scope.WorldState.WarmUp(block.Withdrawals[i].Address);
                         }
                         finally
@@ -135,8 +130,7 @@ public sealed class BlockCachePreWarmer(ReadOnlyTxProcessingEnvFactory envFactor
 
         try
         {
-            int progress = 0;
-            Parallel.For(0, block.Transactions.Length, parallelOptions, _ =>
+            ParallelUnbalancedWork.For(0, block.Transactions.Length, parallelOptions, i =>
             {
                 using ThreadExtensions.Disposable handle = Thread.CurrentThread.BoostPriority();
                 IReadOnlyTxProcessorSource env = _envPool.Get();
@@ -144,9 +138,6 @@ public sealed class BlockCachePreWarmer(ReadOnlyTxProcessingEnvFactory envFactor
                 Transaction? tx = null;
                 try
                 {
-                    // Process transactions in sequential order, rather than partitioning scheme from Parallel.For
-                    // Interlocked.Increment returns the incremented value, so subtract 1 to start at 0
-                    int i = Interlocked.Increment(ref progress) - 1;
                     // If the transaction has already been processed or being processed, exit early
                     if (block.TransactionProcessed > i) return;
 
@@ -273,14 +264,9 @@ public sealed class BlockCachePreWarmer(ReadOnlyTxProcessingEnvFactory envFactor
                     }
                 }
 
-                int progress = 0;
-                Parallel.For(0, block.Transactions.Length, parallelOptions,
-                _ =>
+                ParallelUnbalancedWork.For(0, block.Transactions.Length, parallelOptions,
+                i =>
                 {
-                    int i = 0;
-                    // Process addresses in sequential order, rather than partitioning scheme from Parallel.For
-                    // Interlocked.Increment returns the incremented value, so subtract 1 to start at 0
-                    i = Interlocked.Increment(ref progress) - 1;
                     Transaction tx = block.Transactions[i];
                     Address? sender = tx.SenderAddress;
 
--- a/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
@@ -132,19 +132,19 @@ public sealed class BlockCachePreWarmer(ReadOnlyTxProcessingEnvFactory envFactor
 
         try
         {
-            ParallelUnbalancedWork.For(0, block.Transactions.Length, parallelOptions, (preWarmer: this, block, stateRoot, spec), static (i, state) =>
+            ParallelUnbalancedWork.For<BlockState>(0, block.Transactions.Length, parallelOptions, new(this, block, stateRoot, spec), static (i, state) =>
             {
-                IReadOnlyTxProcessorSource env = state.preWarmer._envPool.Get();
-                SystemTransaction systemTransaction = state.preWarmer._systemTransactionPool.Get();
+                IReadOnlyTxProcessorSource env = state.PreWarmer._envPool.Get();
+                SystemTransaction systemTransaction = state.PreWarmer._systemTransactionPool.Get();
                 Transaction? tx = null;
                 try
                 {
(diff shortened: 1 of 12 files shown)


src/Nethermind/Nethermind.Core/Threading/ParallelUnbalancedWork.cs
```diff
@@ -0,0 +1,331 @@
...
+        {
+            [FieldOffset(64)]
+            public int Value = value;
+        }
+    }
+
+    /// <summary>
+    /// Represents the base data shared among threads during parallel execution.
+    /// </summary>
+    private class BaseData(int threads, int fromInclusive, int toExclusive)
```
LukaszRozmej, line 176: Shouldn't `threads` be captured to a field?

src/Nethermind/Nethermind.Core/Threading/ParallelUnbalancedWork.cs
```diff
@@ -0,0 +1,331 @@
...
+    {
+        /// <summary>
+        /// Gets the shared counter for indices.
+        /// </summary>
+        public SharedCounter Index { get; } = new SharedCounter(fromInclusive);
+        public SemaphoreSlim Event { get; } = new(initialCount: 0);
```
LukaszRozmej, line 182: Why not `ManualResetEventSlim`? I don't think you need counting.

src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
```diff
@@ -126,24 +123,19 @@ private void WarmupTransactions(ParallelOptions parallelOptions, IReleaseSpec sp
 
         try
         {
-            int progress = 0;
-            Parallel.For(0, block.Transactions.Length, parallelOptions, _ =>
+            ParallelUnbalancedWork.For(0, block.Transactions.Length, parallelOptions, (preWarmer: this, block, stateRoot, spec), static (i, state) =>
```
LukaszRozmej, line 126: for readability create a private struct for `State` with optional `spec` and use it in this class.

src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.cs
```diff
@@ -334,11 +335,15 @@ protected virtual TxReceipt[] ProcessBlock(
...
     private static void CalculateBlooms(TxReceipt[] receipts)
     {
-        int index = 0;
-        Parallel.For(0, receipts.Length, _ =>
+        ParallelUnbalancedWork.For(
+            0,
+            receipts.Length,
+            ParallelUnbalancedWork.DefaultOptions,
+            receipts,
+            static (i, receipts) =>
```
LukaszRozmej, line 343: If you would make a `ForEach` version, then you don't need to pass any state to it.
Or maybe by default (or by overload) also pass the element to the delegate.

src/Nethermind/Nethermind.Consensus/Processing/RecoverSignature.cs
```diff
@@ -48,13 +49,20 @@ public void RecoverData(Block block)
                 // so we assume the rest of txs in the block are already recovered
                 return;
 
-            Parallel.For(0, txs.Length, i =>
+            ParallelUnbalancedWork.For(
+                0,
+                txs.Length,
+                ParallelUnbalancedWork.DefaultOptions,
+                txs,
+                static (i, txs) =>
```
LukaszRozmej, line 57: Same here, would benefit from `ForEach`

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
```diff
@@ -266,20 +268,29 @@ void UpdateRootHashesSingleThread()
         void UpdateRootHashesMultiThread()
         {
             // We can recalculate the roots in parallel as they are all independent tries
-            Parallel.ForEach(_storages, RuntimeInformation.ParallelOptionsLogicalCores, kvp =>
+            var storages = _storages.ToArray();
```
LukaszRozmej, line 271: `ToPooledList` ?

LukaszRozmej (commented):
Can you provide some kind of benchmark?

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
benaadams: Used Pooled List

src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.cs
benaadams: ForEach would need locking on the iterator so introduce contention?

src/Nethermind/Nethermind.Core/Threading/ParallelUnbalancedWork.cs
benaadams: SemaphoreSlim pulses first waiting thread, ManualResetEventSlim pulses all waiting threads (though is only 1 waiting, so maybe not much difference)

src/Nethermind/Nethermind.Core/Threading/ParallelUnbalancedWork.cs
benaadams: Does look a bit weird, captured to field instead

benaadams:
| Method             | Mean     | Error    | StdDev   | Ratio | Allocated | Alloc Ratio |
|------------------- |---------:|---------:|---------:|------:|----------:|------------:|
| ParallelFor        | 543.9 ms | 12.13 ms | 18.15 ms |  1.00 |  19.21 KB |        1.00 |
| ParallelForEach    | 488.5 ms | 12.73 ms | 17.43 ms |  0.90 |  23.12 KB |        1.20 |
| UnbalancedParallel | 413.4 ms |  1.15 ms |  1.73 ms |  0.76 |    5.00 KB |        0.26 |

src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
```diff
@@ -98,24 +98,21 @@ private void WarmupWithdrawals(ParallelOptions parallelOptions, IReleaseSpec spe
...
             {
-                int progress = 0;
-                Parallel.For(0, block.Withdrawals.Length, parallelOptions,
-                    _ =>
+                ParallelUnbalancedWork.For(0, block.Withdrawals.Length, parallelOptions, (preWarmer: this, block, stateRoot),
+                    static (i, state) =>
                     {
-                        IReadOnlyTxProcessorSource env = _envPool.Get();
-                        int i = 0;
+                        IReadOnlyTxProcessorSource env = state.preWarmer._envPool.Get();
```
LukaszRozmej, line 104: super minor: maybe better to pass the pool directly rather than whole prewarmer as it is not needed?
And maybe build custom struct for it as those are used in 2 places?

src/Nethermind/Nethermind.Core/Threading/ParallelUnbalancedWork.cs
LukaszRozmej: Rider was complaining about it.

src/Nethermind/Nethermind.Core/Threading/ParallelUnbalancedWork.cs
```diff
@@ -0,0 +1,332 @@
...
+        /// <summary>
+        /// Gets the number of active threads.
+        /// </summary>
+        public int ActiveThreads => Volatile.Read(ref _activeThreads);
+
+        /// <summary>
+        /// Marks a thread as completed.
+        /// </summary>
+        /// <returns>The number of remaining active threads.</returns>
+        public int MarkThreadCompleted()
```
LukaszRozmej, line 199: minor: in theory this could be misused by calling `MarkThreadCompleted` first, but as this is private class we can ignore it.

LukaszRozmej (approved):
I have a feeling `ParallelUnbalancedWork` deserves either to be a PR to dotnet or at least its own nuget package.

TeddyAlbina:
> I have a feeling `ParallelUnbalancedWork` deserves either to be a PR to dotnet or at least its own nuget package.

I agree this should be in dotnet itself

PR #7788, opened 2024-11-22
Only accept txs once synced

benaadams:
Is a lot of additional work on processing blocks and adding tx in the txpool that is completely irrelevant until synced. Skip this extra work.

Effects all types of sync (snap, full, archive etc)

## Changes

- Skip trying to add tx to txpool if not yet synced
- Don't process block events in txpool if not yet synced

- [x] Optimization

#### Requires testing

- [x] Yes

- [x] Yes


src/Nethermind/Nethermind.Blockchain/ChainHeadInfoProvider.cs
```diff
@@ -51,6 +58,24 @@ public ChainHeadInfoProvider(IChainHeadSpecProvider specProvider, IBlockTree blo
...
+            get
+            {
+                if (_hasSynced)
+                {
+                    return false;
+                }
+
+                long bestSuggestedNumber = _blockTree.FindBestSuggestedHeader()?.Number ?? 0;
+                long headNumberOrZero = _blockTree.Head?.Number ?? 0;
+                bool isSyncing = bestSuggestedNumber == 0 || bestSuggestedNumber > headNumberOrZero + 8;
```
LukaszRozmej, line 73: Can you make this extension method on block tree? I think I've seen this or similar logic in few other places before, can we unify?

src/Nethermind/Nethermind.Blockchain/ChainHeadInfoProvider.cs
```diff
@@ -27,14 +29,19 @@ public ChainHeadInfoProvider(ISpecProvider specProvider, IBlockTree blockTree, I
         {
         }
 
-        public ChainHeadInfoProvider(IChainHeadSpecProvider specProvider, IBlockTree blockTree, IReadOnlyStateProvider stateProvider, ICodeInfoRepository codeInfoRepository)
+        public ChainHeadInfoProvider(IChainHeadSpecProvider specProvider, IBlockTree blockTree, IReadOnlyStateProvider stateProvider, ICodeInfoRepository codeInfoRepository, bool forceSynced = false)
```
LukaszRozmej, line 32: as `forceSynced` is only used in test maybe either make it internal init property or add this logic in derived test-only type?

src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -408,6 +410,8 @@ public void RemovePeer(PublicKey nodeId)
 
         public AcceptTxResult SubmitTx(Transaction tx, TxHandlingOptions handlingOptions)
         {
+            if (_headInfo.IsSyncing) return AcceptTxResult.AlreadyKnown;
```
LukaszRozmej, line 413: Shouldn't we have a new result for this scenario?

benaadams:
Accept local tx

src/Nethermind/Nethermind.TxPool/TxPool.cs
marcindsobczak: +1 for creating `AcceptTxResult.Syncing` or something like that

marcindsobczak (commented):
It's a great change, but I would definitely not return `AcceptTxResult.AlreadyKnown` when syncing as it can be misleading

src/Nethermind/Nethermind.Blockchain/ChainHeadInfoProvider.cs
benaadams: Done

src/Nethermind/Nethermind.Blockchain/ChainHeadInfoProvider.cs
benaadams: Done

src/Nethermind/Nethermind.TxPool/TxPool.cs
benaadams: Added

src/Nethermind/Nethermind.Blockchain/BlockTreeExtensions.cs
```diff
@@ -10,5 +10,14 @@ public static class BlockTreeExtensions
         public static ReadOnlyBlockTree AsReadOnly(this IBlockTree blockTree) => new(blockTree);
 
         public static BlockHeader? GetProducedBlockParent(this IBlockTree blockTree, BlockHeader? parentHeader) => parentHeader ?? blockTree.Head?.Header;
+
+        public static (bool isSyncing, long headNumberOrZero, long bestSuggested) IsSyncing(this IBlockTree blockTree, int maxDistanceForSynced = 0)
```
LukaszRozmej, line 14: I think we can rename it for simplicity. Normally our head should be on Genesis by default, if it isn't there is something wrong. Or we can just make it nullable here.

```suggestion
        public static (bool isSyncing, long headNumber, long bestSuggested) IsSyncing(this IBlockTree blockTree, int maxDistanceForSynced = 0)
```

LukaszRozmej (approved):

PR #7789, opened 2024-11-22
Fix/broken storage after snap sync

asdacap:
- During snap sync, when a large storage was updated to a different storage root, a refresh operation would happen and the new storage root would get re-fetched, and the resulting storage tree would be incomplete and require healing. This would get detected during healing as the path to the leaf at state level would be missing and need to be updated. However, if after the refresh, the storage root is updated back to its original value, the path to the leaf in the state tree would be correct as it was already persisted in the beginning and the healing for the storage tree would not get triggered. This seems to be the case with some contract on some network. 
- This PR share the pivot class between snap and state sync in `StateSyncPivot`, originally was `Pivot` specific for snap sync. When a storage healing was scheduled, the address to the storage is added to `UpdatedStorages` in the `StateSyncPivot`. This usually happens less than 100 times. During healing, once the root was scheduled for persist, the new storage root for each of these contract is checked for existence. If any of the root is missing, the missing storage root is queued for download through standard healing code path and therefore the rest of the missing storage tree will be downloaded.
- Also fix `StateMin/MaxDistanceFromHead` did not account for `FastSyncLag` which cause sync issue on network without CL. 
- Also add a simple code path for #7605 storage fix. Limits the amount of storage that can be parallelized though.

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Added some log when such situation happen, confirmed it happens.

### DIFF
--- a/src/Nethermind/Nethermind.Core/ContainerBuilderExtensions.cs
+++ b/src/Nethermind/Nethermind.Core/ContainerBuilderExtensions.cs
@@ -34,7 +34,9 @@ public static class ContainerBuilderExtensions
             object? val = propertyInfo.GetValue(source);
             if (val != null)
             {
-                configuration.RegisterInstance(val).As(propertyInfo.PropertyType);
+                configuration.RegisterInstance(val)
+                    .As(propertyInfo.PropertyType)
+                    .ExternallyOwned();
             }
         }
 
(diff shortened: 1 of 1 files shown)


kamilchodola (approved):
Imho works much better now - much less TrieExceptions but tested a bit too early and still had some but later Ashraf added another fixes.

Have a 20 nodes running on that now but as it takes time to reproduce I'd rather prefer merging that and monitoring "Sync Supported Chains" action to see if it will happen ever again.

benaadams (approved):

PR #7792, opened 2024-11-22
Bootnodes for mekong

ak88:


src/Nethermind/Chains/mekong.json
```diff
@@ -74,6 +74,12 @@
     "extraData": "",
     "gasLimit": "0x17d7840"
   },
+  "nodes":
```
benaadams, line 77: Already has empty `nodes` item at bottom, so that would overwrite this one being later?

ak88: good catch

benaadams (approved):

src/Nethermind/Chains/mekong.json
kamilchodola: Actually started before noticing that and bootnodes were not loaded correctly :) So interesting thing for future.

PR #7805, opened 2024-11-26
Only initialize cache inside prewarmer

benaadams:
## Changes

- Getting BeaconBlockRootHandler AccessList initializes the rlp cache; move it inside the prewarmer rather than outside otherwise causes "Caches Rlp are not empty. Clearing them" warning.

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
@@ -17,6 +17,7 @@ using Nethermind.Int256;
 using Nethermind.Logging;
 using Nethermind.State;
 using Nethermind.Core.Eip2930;
+using Nethermind.Core.Collections;
 
 namespace Nethermind.Consensus.Processing;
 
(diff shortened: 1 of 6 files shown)


MarekM25 (approved):

src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
```diff
@@ -198,17 +206,31 @@ private void WarmupTransactions(ParallelOptions parallelOptions, IReleaseSpec sp
...
+
+            ArrayPoolList<AccessList> list = new(systemAccessLists.Length);
+
+            foreach (IHasAccessList systemAccessList in systemAccessLists)
+            {
+                list.Add(systemAccessList.GetAccessList(block, spec));
+            }
+
+            return list;
+        }
```
LukaszRozmej, line 232: Wouldn't merging AccessList be better?

benaadams: Is only ever one so far; can look at that when more

PR #7818, opened 2024-11-26
Stop accepting JsonRpc requests during shutdown

benaadams:
## Changes

image

- Reject incoming requests during shutdown; been bugging me for a while :)


MarekM25 (approved):
LGTM

However, I think it doesn't cause any issues other than logging.

PR #7819, opened 2024-11-26
Full flush on shutdown

benaadams:
## Changes

- Perform full flush on shutdown

#### Requires testing

- [x] No


rubo (approved):

LukaszRozmej:
@asdacap can you double check?

PR #7820, opened 2024-11-26
Don't double log invalid blocks

benaadams:
## Changes

- Only write one log line per invalid block, rather than two

Before

image

After

image

- [x] Refactoring

#### Requires testing

- [x] No


benaadams:
> How does log level change affect double logging? Do we have the same somewhere else with the level of info?

Yes, just further down log same thing at Info but with addition of block number so that log is more useful

PR #7825, opened 2024-11-27
Fix dodgy bracket precedent

benaadams:
## Changes

- Incorrect bracketing for fast path

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Evm/VirtualMachine.cs
+++ b/src/Nethermind/Nethermind.Evm/VirtualMachine.cs
@@ -2177,7 +2177,7 @@ internal sealed class VirtualMachine<TLogger> : IVirtualMachine where TLogger : 
             gasLimitUl += GasCostOf.CallStipend;
         }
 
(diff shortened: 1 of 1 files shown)


LukaszRozmej:
Hm... this was the last time it was different: https://github.com/NethermindEth/nethermind/blame/08c2bc38e7e65b96f1dd4cea7972700b62a9c707/src/Nethermind/Nethermind.Evm/VirtualMachine.cs#L1746

Then changed here: https://github.com/NethermindEth/nethermind/blame/816facec9f5e4fb7fed4bde36e866a43d0633c5e/src/Nethermind/Nethermind.Evm/VirtualMachine.cs#L2579

LukaszRozmej (approved):

benaadams:
> Hm... this was the last time it was different: [f77565f](https://github.com/NethermindEth/nethermind/commit/f77565f56aa7ba9b92625fc9b89c1e0801500596)

Is mostly optimization to not access state when value is zero

PR #7828, opened 2024-11-28
Use static SHA256.TryHashData rather than thread local instance

benaadams:
## Changes

- Use static SHA256`.TryHashData` rather than thread local instance and `.TryComputeHash`

- [x] Refactoring

#### Requires testing

- [x] No


LukaszRozmej (commented):
Also change in `Sha256Precompile` and `ShaMerkleTree`

LukaszRozmej (approved):

PR #7830, opened 2024-11-28
Move lowest inserted body number out of blocktree

asdacap:
- Mode LowestInsertedBodyNumber out of IBlockTree.
- Minor change, just make sense. Era and portal will just make it not very clear
- Delete some interface method from `IBlockTree` and `IBlockStore`. 
- LowestInsertedReceipt is not in IBlockTree at all. 

- [x] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Testing...

### DIFF
--- a/src/Nethermind/Nethermind.Api/NethermindApi.cs
+++ b/src/Nethermind/Nethermind.Api/NethermindApi.cs
@@ -226,6 +226,7 @@ namespace Nethermind.Api
         public CensorshipDetector CensorshipDetector { get; set; } = null!;
         public IWallet? Wallet { get; set; }
         public IBadBlockStore? BadBlocksStore { get; set; }
+        public BlockStore? BlockStore { get; set; }
         public ITransactionComparerProvider? TransactionComparerProvider { get; set; }
         public IWebSocketsManager WebSocketsManager { get; set; } = new WebSocketsManager();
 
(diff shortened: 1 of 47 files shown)


benaadams (approved):

src/Nethermind/Nethermind.Api/NethermindApi.cs
```diff
@@ -226,6 +226,7 @@ public ISealEngine SealEngine
         public CensorshipDetector CensorshipDetector { get; set; } = null!;
         public IWallet? Wallet { get; set; }
         public IBadBlockStore? BadBlocksStore { get; set; }
+        public BlockStore? BlockStore { get; set; }
```
LukaszRozmej, line 229: Shouldn't it be `IBlockStore` in the API?

src/Nethermind/Nethermind.Blockchain/Blocks/IBlockStore.cs
```diff
@@ -19,9 +19,6 @@ public interface IBlockStore
     byte[]? GetRlp(long blockNumber, Hash256 blockHash);
     ReceiptRecoveryBlock? GetReceiptRecoveryBlock(long blockNumber, Hash256 blockHash);
     void Cache(Block block);
-
-    // These two are used by blocktree. Try not to use them...
-    void SetMetadata(byte[] key, byte[] value);
-    byte[]? GetMetadata(byte[] key);
     bool HasBlock(long blockNumber, Hash256 blockHash);
+    long? LowestInsertedBodyNumber { get; set; }
```
LukaszRozmej, line 23: Does it need public set? Can it just update it internally on `Insert`?

asdacap: Can't as erafiles and portal would also use insert.

src/Nethermind/Nethermind.Api/NethermindApi.cs
asdacap: Moved to ISyncPointers.

LukaszRozmej (approved):

PR #7831, opened 2024-11-28
Faster sync tests

asdacap:
- Faster I say! _Faster_!
- Turns out most of the slowdown is in `SyncFeed.Finish` that calls GC. This PR changes that so that it is only called on state sync finished.
- Also remove global lock in peer allocation.
- Also create a cached block tree builder that caches blocktree at db level.

- [X] Optimization

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Mainnet sync ok. Confirmed gc called once.

### DIFF
--- a/src/Nethermind/Nethermind.Core.Test/CachedBlockTreeBuilder.cs
+++ b/src/Nethermind/Nethermind.Core.Test/CachedBlockTreeBuilder.cs
@@ -1,0 +1,55 @@
+// SPDX-FileCopyrightText: 2024 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System;
+using Nethermind.Blockchain;
+using Nethermind.Core.Test.Builders;
+using Nethermind.Db;
+using NonBlocking;
+
+namespace Nethermind.Core.Test;
+
+public class CachedBlockTreeBuilder
+{
+    private static ConcurrentDictionary<string, CachedDb> _cachedDbs = new();
+    private record CachedDb(
+        MemDb blocksDb,
+        MemDb metadataDb,
+        MemDb headersDb,
+        MemDb blockNumbersDb,
+        MemDb blockInfo
+    );
+
+    public static IBlockTree BuildCached(string key, Func<BlockTreeBuilder> blockTreeBuilderFactory)
+    {
+        if (_cachedDbs.TryGetValue(key, out CachedDb? db))
+        {
+            return Build.A.BlockTree()
+                .WithBlocksDb(MemDb.CopyFrom(db.blocksDb))
+                .WithMetadataDb(MemDb.CopyFrom(db.metadataDb))
+                .WithHeadersDb(MemDb.CopyFrom(db.headersDb))
+                .WithBlocksNumberDb(MemDb.CopyFrom(db.blockNumbersDb))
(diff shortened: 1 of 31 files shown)


src/Nethermind/Nethermind.Core.Test/CachedBlockTreeBuilder.cs
```diff
@@ -0,0 +1,55 @@
...
+            CachedDb cachedValue = new CachedDb(
+                MemDb.CopyFrom(builder.BlocksDb),
+                MemDb.CopyFrom(builder.MetadataDb),
+                MemDb.CopyFrom(builder.HeadersDb),
+                MemDb.CopyFrom(builder.BlockNumbersDb),
+                MemDb.CopyFrom(builder.BlockInfoDb)
+            );
+            _cachedDbs.TryAdd(key, cachedValue);
+
+            return builder.TestObject;
```
LukaszRozmej, line 47: Technically you should use `GetOrAdd` on `ConcurrentDictionary`

src/Nethermind/Nethermind.Synchronization/Peers/SyncPeerAllocation.cs
```diff
@@ -29,14 +29,15 @@ public class SyncPeerAllocation
         public bool HasPeer => Current is not null;
 
         public SyncPeerAllocation(PeerInfo peerInfo, AllocationContexts contexts)
-            : this(new StaticStrategy(peerInfo), contexts)
+            : this(new StaticStrategy(peerInfo), contexts, null)
         {
         }
 
-        public SyncPeerAllocation(IPeerAllocationStrategy peerAllocationStrategy, AllocationContexts contexts)
+        public SyncPeerAllocation(IPeerAllocationStrategy peerAllocationStrategy, AllocationContexts contexts, Lock? allocationLock)
```
LukaszRozmej, line 36: Make:
```suggestion
        public SyncPeerAllocation(IPeerAllocationStrategy peerAllocationStrategy, AllocationContexts contexts, Lock? allocationLock = null)
```

and don't pass `null` when not needed?

src/Nethermind/Nethermind.Synchronization/ParallelSync/SyncFeed.cs
```diff
@@ -44,7 +44,6 @@ private void ChangeState(SyncFeedState newState)
         public virtual void Finish()
         {
             ChangeState(SyncFeedState.Finished);
-            GC.Collect(2, GCCollectionMode.Aggressive, true, true);
```
LukaszRozmej, line 47: I think this GC was also fired on Headers, Blocks and Receipts!

LukaszRozmej (commented):
I think GC was not only for state?

src/Nethermind/Nethermind.Synchronization/ParallelSync/SyncFeed.cs
asdacap: Thats fine, the state take most of the allocation.

asdacap: ok

LukaszRozmej (approved):

PR #7833, opened 2024-11-28
Decrease message severity for non-existing account

benaadams:
## Changes

- Decrease message severity for non-existing account `Error` -> `Debug` (throws exception so if issue will be picked up elsewhere)
- Move if checks on update vs create to use combined method for this purpose

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.State/StateProvider.cs
+++ b/src/Nethermind/Nethermind.State/StateProvider.cs
@@ -177,7 +177,7 @@ namespace Nethermind.State
                 Account result = GetThroughCache(address);
                 if (result is null)
                 {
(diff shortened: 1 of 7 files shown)


src/Nethermind/Nethermind.State/StateProvider.cs
```diff
@@ -177,7 +177,7 @@ Account GetThroughCacheCheckExists()
                 Account result = GetThroughCache(address);
                 if (result is null)
                 {
-                    if (_logger.IsError) _logger.Error("Updating balance of a non-existing account");
```
benaadams, line 180: Main change

LukaszRozmej (approved):

flcl42 (approved):

PR #7838, opened 2024-11-29
Use new Json Pipeline overloads

benaadams:
## Changes

- Use the new async Json Pipeline overloads introduced in .NET 9.0 https://github.com/dotnet/runtime/pull/101461 to respond better to back pressure on the Json response.

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Serialization.Json/EthereumJsonSerializer.cs
+++ b/src/Nethermind/Nethermind.Serialization.Json/EthereumJsonSerializer.cs
@@ -143,18 +143,15 @@ namespace Nethermind.Serialization.Json
         public async ValueTask<long> SerializeAsync<T>(Stream stream, T value, bool indented = false, bool leaveOpen = true)
         {
             var writer = GetPipeWriter(stream, leaveOpen);
-            Serialize(writer, value, indented);
+            await JsonSerializer.SerializeAsync(writer, value, indented ? JsonOptionsIndented : _jsonOptions);
             await writer.CompleteAsync();
 
             long outputCount = writer.WrittenCount;
             return outputCount;
         }
 
-        public void Serialize<T>(IBufferWriter<byte> writer, T value, bool indented = false)
-        {
-            using var jsonWriter = new Utf8JsonWriter(writer, CreateWriterOptions(indented));
-            JsonSerializer.Serialize(jsonWriter, value, indented ? JsonOptionsIndented : _jsonOptions);
-        }
+        public Task SerializeAsync<T>(PipeWriter writer, T value, bool indented = false)
+            => JsonSerializer.SerializeAsync(writer, value, indented ? JsonOptionsIndented : _jsonOptions);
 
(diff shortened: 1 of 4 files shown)


src/Nethermind/Nethermind.Serialization.Json/StreamPipeWriter.cs
```diff
@@ -16,16 +16,14 @@ namespace Nethermind.Serialization.Json;
...
-public interface ICountingBufferWriter : IBufferWriter<byte>
+public abstract class CountingWriter : PipeWriter
 {
-    long WrittenCount { get; }
-    ValueTask CompleteAsync(Exception? exception = null);
+    public long WrittenCount { get; protected set; }
 }
 
-public sealed class CountingPipeWriter : ICountingBufferWriter
+public sealed class CountingPipeWriter : CountingWriter
```
LukaszRozmej, line 24: Why is it better to inherit PipeWriter, when it is clearly a decorator?

src/Nethermind/Nethermind.Serialization.Json/EthereumJsonSerializer.cs
```diff
@@ -143,18 +143,15 @@ private JsonWriterOptions CreateWriterOptions(bool indented)
         public async ValueTask<long> SerializeAsync<T>(Stream stream, T value, bool indented = false, bool leaveOpen = true)
         {
             var writer = GetPipeWriter(stream, leaveOpen);
-            Serialize(writer, value, indented);
+            await JsonSerializer.SerializeAsync(writer, value, indented ? JsonOptionsIndented : _jsonOptions);
```
LukaszRozmej, line 146: Does it make sense to be `ValueTask` method now? Won't that always allocate?

LukaszRozmej: `SerializeAsync` returns a `Task`, so it will always allocate?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Serialization.Json/EthereumJsonSerializer.cs
benaadams: Then this method will allocate an additional Task vs if it synchronously completes it won't?

LukaszRozmej: won't it always allocate as the inner Task is not synchronous?

benaadams: The inner task will normally be synchrounous; will only go async when the outbound http buffer is full and it suspends

LukaszRozmej: Why didn't they make it ValueTask then?

PR #7839, opened 2024-11-29
Don't fail when stats can't retrieve beneficiary

benaadams:
## Changes

- Stats shouldn't effect processing

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Processing/ProcessingStats.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/ProcessingStats.cs
@@ -113,7 +113,7 @@ namespace Nethermind.Consensus.Processing
             catch (Exception ex)
             {
                 // Don't allow exception to escape to ThreadPool
-                if (_logger.IsError) _logger.Error($"{nameof(ProcessingStats)} Error when generating processing statistics", ex);
+                 if (_logger.IsError) _logger.Error("Error when generating processing statistics", ex);
             }
         }
 
--- a/src/Nethermind/Nethermind.Consensus/Processing/ProcessingStats.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/ProcessingStats.cs
@@ -134,9 +134,17 @@ namespace Nethermind.Consensus.Processing
 
             if (_lastBranchRoot is null || !_stateReader.HasStateForRoot(_lastBranchRoot) || block.StateRoot is null || !_stateReader.HasStateForRoot(block.StateRoot))
                 return;
-            UInt256 beforeBalance = _stateReader.GetBalance(_lastBranchRoot, beneficiary);
-            UInt256 afterBalance = _stateReader.GetBalance(block.StateRoot, beneficiary);
-            UInt256 rewards = beforeBalance < afterBalance ? afterBalance - beforeBalance : default;
+
+            UInt256 rewards = default;
+            try
+            {
+                UInt256 beforeBalance = _stateReader.GetBalance(_lastBranchRoot, beneficiary);
+                UInt256 afterBalance = _stateReader.GetBalance(block.StateRoot, beneficiary);
+                rewards = beforeBalance < afterBalance ? afterBalance - beforeBalance : default;
+            }
+            catch
+            {
+            }
 
             long currentSelfDestructs = Evm.Metrics.SelfDestructs;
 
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Consensus/Processing/ProcessingStats.cs
```diff
@@ -134,9 +134,17 @@ void IThreadPoolWorkItem.Execute()
...
+            UInt256 rewards = default;
+            try
+            {
+                UInt256 beforeBalance = _stateReader.GetBalance(_lastBranchRoot, beneficiary);
+                UInt256 afterBalance = _stateReader.GetBalance(block.StateRoot, beneficiary);
+                rewards = beforeBalance < afterBalance ? afterBalance - beforeBalance : default;
+            }
+            catch
+            {
+            }
```
LukaszRozmej, line 147: I think this catch should be on whole `Execute` method. and lets add logs on exception

LukaszRozmej (changes requested):

src/Nethermind/Nethermind.Consensus/Processing/ProcessingStats.cs
benaadams: Added catch for whole Execute

LukaszRozmej (approved):

src/Nethermind/Nethermind.Consensus/Processing/ProcessingStats.cs
```diff
@@ -134,9 +147,17 @@ void IThreadPoolWorkItem.Execute()
...
```
LukaszRozmej, line 160: Remove this try

src/Nethermind/Nethermind.Consensus/Processing/ProcessingStats.cs
```diff
@@ -105,6 +105,19 @@ public void UpdateStats(Block? block, Hash256 branchRoot, long blockProcessingTi
...
         void IThreadPoolWorkItem.Execute()
+        {
+            try
+            {
+                Execute();
+            }
+            catch (Exception ex)
+            {
+                // Don't allow exception to escape to ThreadPool
+                if (_logger.IsError) _logger.Error($"{nameof(ProcessingStats)} Error", ex);
```
LukaszRozmej, line 116:
```suggestion
                if (_logger.IsError) _logger.Error($"{nameof(ProcessingStats)} Error when generating processing statistics", ex);
```

src/Nethermind/Nethermind.Consensus/Processing/ProcessingStats.cs
benaadams: Will mean it won't generate a Processing line when just this bit fails

LukaszRozmej, line 160:
```suggestion
            UInt256 beforeBalance = _stateReader.GetBalance(_lastBranchRoot, beneficiary);
            UInt256 afterBalance = _stateReader.GetBalance(block.StateRoot, beneficiary);
            UInt256 rewards = beforeBalance < afterBalance ? afterBalance - beforeBalance : default;
```

LukaszRozmej: It shouldn't fail. It failed because `To` was null

src/Nethermind/Nethermind.Consensus/Processing/ProcessingStats.cs
```diff
@@ -105,6 +105,19 @@ public void UpdateStats(Block? block, Hash256 branchRoot, long blockProcessingTi
...
         void IThreadPoolWorkItem.Execute()
+        {
+            try
+            {
+                Execute();
+            }
+            catch (Exception ex)
+            {
+                // Don't allow exception to escape to ThreadPool
+                if (_logger.IsError) _logger.Error($"{nameof(ProcessingStats)} Error when generating processing statistics", ex);
```
LukaszRozmej, line 116:
```suggestion
                 if (_logger.IsError) _logger.Error("Error when generating processing statistics", ex);
```

src/Nethermind/Nethermind.Consensus/Processing/ProcessingStats.cs
LukaszRozmej: at least add logging in catch

benaadams: Added

LukaszRozmej (approved):

PR #7843, opened 2024-12-01
Optimize EvmStack

benaadams:
## Changes

- Eliminate double bounds checks and improve vectors as most frequently called methods

- [x] Optimization

#### Requires testing

- [x] No


src/Nethermind/Nethermind.Core/Extensions/Bytes.cs
```diff
@@ -27,8 +27,10 @@ public static unsafe partial class Bytes
         public static readonly IEqualityComparer<byte[]> EqualityComparer = new BytesEqualityComparer();
         public static readonly IEqualityComparer<byte[]?> NullableEqualityComparer = new NullableBytesEqualityComparer();
         public static readonly BytesComparer Comparer = new();
-        public static readonly ReadOnlyMemory<byte> ZeroByte = new byte[] { 0 };
-        public static readonly ReadOnlyMemory<byte> OneByte = new byte[] { 1 };
+        public static ReadOnlyMemory<byte> ZeroByteMemory = new byte[] { 0 };
+        public static ReadOnlyMemory<byte> OneByteMemory = new byte[] { 1 };
+        public static ReadOnlySpan<byte> ZeroByteSpan() => new byte[] { 0 };
+        public static ReadOnlySpan<byte> OneByteSpan() => new byte[] { 1 };
```
LukaszRozmej, line 33:
```suggestion
        public static ReadOnlySpan<byte> ZeroByteSpan() => ZeroByteMemory.Span;
        public static ReadOnlySpan<byte> OneByteSpan() => OneByteMemory.Span;
```

LukaszRozmej: Can it be property instead of method?

src/Nethermind/Nethermind.Evm/EvmStack.cs
```diff
@@ -161,17 +151,14 @@ public void PushUInt32(in int value)
...
             Word shuffle = Vector256.Create(
-                (byte)
-                31, 30, 29, 28, 27, 26, 25, 24,
-                23, 22, 21, 20, 19, 18, 17, 16,
-                15, 14, 13, 12, 11, 10, 9, 8,
-                7, 6, 5, 4, 3, 2, 1, 0);
+                0x18191a1b1c1d1e1ful,
+                0x1011121314151617ul,
+                0x08090a0b0c0d0e0ful,
+                0x0001020304050607ul).AsByte();
```
LukaszRozmej, line 161: shouldn't this vector be a static or something?

src/Nethermind/Nethermind.Evm/EvmStack.cs
```diff
@@ -245,11 +227,10 @@ public bool PopUInt256(out UInt256 result)
...
```
LukaszRozmej, line 233: same here

LukaszRozmej (approved):

src/Nethermind/Nethermind.Evm/EvmStack.cs
benaadams: Jit automagically handles that for you for vectors 
image

LukaszRozmej (approved):

PR #7845, opened 2024-12-01
TrieStore to Channels

benaadams:
## Changes

- Wrapping a `ConcurrentQueue` in a `BlockingCollection` burns a ton of CPU spinning; whereas its a pattern ideally suited to `Channel`s; and using true `async`

- [x] Optimization
- [x] Refactoring

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
+++ b/src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
@@ -12,6 +12,7 @@ using System.Threading.Channels;
 using System.Threading.Tasks;
 using Nethermind.Core;
 using Nethermind.Core.Collections;
+using Nethermind.Core.Cpu;
 using Nethermind.Core.Crypto;
 using Nethermind.Core.Extensions;
 using Nethermind.Logging;
@@ -32,7 +33,7 @@ public class TrieStore : ITrieStore, IPruningTrieStore
     private readonly Task[] _dirtyNodesTasks = [];
     private readonly ConcurrentDictionary<HashAndTinyPath, Hash256?>[] _persistedHashes = [];
     private readonly Action<TreePath, Hash256?, TrieNode> _persistedNodeRecorder;
-    private readonly Task[] _disposeTasks = new Task[Environment.ProcessorCount];
+    private readonly Task[] _disposeTasks = new Task[RuntimeInformation.PhysicalCoreCount];
 
     // This seems to attempt prevent multiple block processing at the same time and along with pruning at the same time.
     private readonly object _dirtyNodesLock = new object();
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -790,26 +790,31 @@ void TopLevelPersist(TrieNode tn, Hash256? address2, TreePath path)
...
+            for (int index = 0; index < _disposeTasks.Length; index++)
             {
-                while (disposeQueue.TryTake(out INodeStorage.WriteBatch disposable, Timeout.Infinite))
+                _disposeTasks[index] = Task.Run(async () =>
                 {
-                    disposable.Dispose();
-                }
-            });
-        }
+                    await foreach (var disposable in disposeQueue.Reader.ReadAllAsync())
```
LukaszRozmej, line 800: will it correctly partition items from channel between 4 tasks/threads here?

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -824,28 +829,28 @@ void TopLevelPersist(TrieNode tn, Hash256? address2, TreePath path)
...
             {
-                disposeQueue.Add(writeBatch);
+                await disposeQueue.Writer.WriteAsync(writeBatch);
                 writeBatch = _nodeStorage.StartWriteBatch();
             }
         }
 
-        tn.CallRecursively(DoPersist, address2, ref path, GetTrieStore(address2), true, _logger);
-        disposeQueue.Add(writeBatch);
+        await tn.CallRecursivelyAsync(DoPersist, address2, path, GetTrieStore(address2), true, _logger);
```
LukaszRozmej, line 852: Not sure if this can break anything... probably not

src/Nethermind/Nethermind.Trie/TrieNode.cs
```diff
@@ -976,6 +977,72 @@ public void CallRecursively(
...
+                if (data is not null)
+                {
+                    for (int i = 0; i < data.Length; i++)
+                    {
+                        if (data[i] is TrieNode child)
+                        {
+                            if (logger.IsTrace) logger.Trace($"Persist recursively on child {i} {child} of {this}");
+                            int previousLength = AppendChildPath(ref currentPath, i);
+                            await child.CallRecursivelyAsync(action, storageAddress, currentPath, resolver, skipPersisted, logger, maxPathLength, resolveStorageRoot);
+                            currentPath.TruncateMut(previousLength);
```
LukaszRozmej, line 1014: No point to mutate the path as we have to do the copy anyways?

asdacap (approved):

benaadams:
Will wait on this as might be better way to do it with https://github.com/NethermindEth/nethermind/pull/7787 without introducing a second code path

PR #7852, opened 2024-12-02
GC: Clear references in ArrayPoolList before returning to pool

benaadams:
## Changes

- Clear object references in ArrayPoolList before returning to pool; otherwise can inadvertently extend object lifetime from reference from array in pool still pointing at large object

image
 

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Core/Collections/ArrayPoolList.cs
+++ b/src/Nethermind/Nethermind.Core/Collections/ArrayPoolList.cs
@@ -5,6 +5,7 @@ using System;
 using System.Buffers;
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Runtime.CompilerServices;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Core/Collections/ArrayPoolList.cs
```diff
@@ -335,13 +358,18 @@ public void Dispose()
...
+            _array = null!;
             if (array is not null)
             {
-                _arrayPool.Return(_array);
-                _array = null!;
+                if (RuntimeHelpers.IsReferenceOrContainsReferences<T>())
+                {
+                    // Release any references to the objects in the array so can be GC'd.
+                    Array.Clear(array, 0, Count);
+                }
```
LukaszRozmej, line 369: This code is repeated 3x, create helper method that takes array as parameter.

LukaszRozmej (approved):

src/Nethermind/Nethermind.Core/Collections/ArrayPoolList.cs
benaadams: Moved to method

PR #7855, opened 2024-12-03
Change/use persisted pointer for fast header

asdacap:
- Currently unlike oldbodies, orldreceipt or beacon header, the pointer for fast header is initialized via binary search from genesis to sync pivot. 
- This cause a problem with era/portal which may insert header in between arbitrarily.
- This change the pointer to be persisted like beacon header and the pointer is changed manually in fast header instead of on every block/header insert.
- Also don't download already persisted header. This only 2x faster than downloading though.

- [X] Optimization
- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Mainnet sync work.

### DIFF
--- a/src/Nethermind/Nethermind.Merge.Plugin/Synchronization/BeaconHeadersSyncFeed.cs
+++ b/src/Nethermind/Nethermind.Merge.Plugin/Synchronization/BeaconHeadersSyncFeed.cs
@@ -36,7 +36,12 @@ public sealed class BeaconHeadersSyncFeed : HeadersSyncFeed
     protected override bool AllHeadersDownloaded => (_blockTree.LowestInsertedBeaconHeader?.Number ?? long.MaxValue) <=
                                                     _pivot.PivotDestinationNumber || _chainMerged;
 
-    protected override BlockHeader? LowestInsertedBlockHeader => _blockTree.LowestInsertedBeaconHeader;
+    protected override BlockHeader? LowestInsertedBlockHeader
+    {
+        get => _blockTree.LowestInsertedBeaconHeader;
+        set => _blockTree.LowestInsertedBeaconHeader = value;
+    }
+
     protected override MeasuredProgress HeadersSyncProgressReport => _syncReport.BeaconHeaders;
 
     protected override MeasuredProgress HeadersSyncQueueReport => _syncReport.BeaconHeadersInQueue;
--- a/src/Nethermind/Nethermind.Merge.Plugin/Synchronization/BeaconHeadersSyncFeed.cs
+++ b/src/Nethermind/Nethermind.Merge.Plugin/Synchronization/BeaconHeadersSyncFeed.cs
@@ -43,7 +43,7 @@ public sealed class BeaconHeadersSyncFeed : HeadersSyncFeed
         {
             // LowestInsertedBeaconHeader is set in blocktree when BeaconHeaderInsert is set.
             // TODO: Probably should move that logic here so that `LowestInsertedBeaconHeader` is set only once per batch.
-v        }
+        }
     }
 
     protected override MeasuredProgress HeadersSyncProgressReport => _syncReport.BeaconHeaders;
--- a/src/Nethermind/Nethermind.Merge.Plugin/Synchronization/BeaconHeadersSyncFeed.cs
+++ b/src/Nethermind/Nethermind.Merge.Plugin/Synchronization/BeaconHeadersSyncFeed.cs
@@ -138,6 +138,10 @@ public sealed class BeaconHeadersSyncFeed : HeadersSyncFeed
         return base.InsertHeaders(batch);
     }
 
+    protected override void OnLowestInsertedHeaderInBatch(BlockHeader lowestInsertedHeader)
+    {
+    }
+
     private void ConnectHeaderChainInInvalidChainTracker(IReadOnlyList<BlockHeader?> batchResponse)
     {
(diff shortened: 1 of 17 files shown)


src/Nethermind/Nethermind.Synchronization/FastBlocks/FastBlocksBatch.cs
```diff
@@ -81,7 +81,9 @@ public double? WaitingTime
             => (_handlingStartTime ?? _stopwatch.ElapsedMilliseconds) - (_waitingStartTime ?? _handlingStartTime ?? _stopwatch.ElapsedMilliseconds);
         public double? HandlingTime
             => (_handlingEndTime ?? _stopwatch.ElapsedMilliseconds) - (_handlingStartTime ?? _stopwatch.ElapsedMilliseconds);
-        public long? MinNumber { get; set; }
+
+        /// Minimum header number for allocation strategy.
+        public virtual long? MinNumber { get; }
```
LukaszRozmej, line 86: should it be abstract?

src/Nethermind/Nethermind.Synchronization/FastBlocks/FastHeadersSyncFeed.cs
```diff
@@ -334,13 +339,20 @@ private bool HasDependencyToProcess
...
+            {
+                batch = new();
+                batch.StartNumber = Math.Max(HeadersDestinationNumber,
+                    _lowestRequestedHeaderNumber - _headersRequestSize);
+                batch.RequestSize = (int)Math.Min(_lowestRequestedHeaderNumber - HeadersDestinationNumber,
+                    _headersRequestSize);
+                _lowestRequestedHeaderNumber = batch.StartNumber;
+
+                batch = ProcessPersistedPortion(batch);
+            } while (batch is null && ShouldBuildANewBatch());
```
LukaszRozmej, line 355: This is sligthly confusing, shouldn't the loop be outside of this method?

src/Nethermind/Nethermind.Synchronization/FastBlocks/FastHeadersSyncFeed.cs
```diff
@@ -456,14 +466,69 @@ private static HeadersSyncBatch BuildDependentBatch(HeadersSyncBatch batch, long
...
+
+            using ArrayPoolList<BlockHeader> headers = new ArrayPoolList<BlockHeader>(1);
+            headers.Add(lastHeader);
+            for (long i = batch.EndNumber - 1; i >= batch.StartNumber; i--)
+            {
+                // Don't worry about fork, `InsertHeaders` will check for fork and retry if it is not on the right fork.
+                BlockHeader nextHeader = _blockTree.FindHeader(lastHeader.ParentHash!, i);
+                if (nextHeader is null) break;
+                lastHeader = nextHeader;
+                headers.Add(lastHeader);
```
LukaszRozmej, line 508: more readable:
```suggestion
                headers.Add(nextHeader);
                lastHeader = nextHeader;
```

src/Nethermind/Nethermind.Synchronization/FastBlocks/FastHeadersSyncFeed.cs
```diff
@@ -456,14 +466,69 @@ private static HeadersSyncBatch BuildDependentBatch(HeadersSyncBatch batch, long
...
+            }
+
+            ArrayPoolList<BlockHeader> reversedBatch = new ArrayPoolList<BlockHeader>(headers.Count);
+            foreach (BlockHeader blockHeader in headers.Reverse())
+            {
+                reversedBatch.Add(blockHeader);
+            }
```
LukaszRozmej, line 515: Why need a second `ArrayPoolList`? Add `Reverse` to the class and reverse in-place (same as `List<>`)

src/Nethermind/Nethermind.Merge.Plugin/Synchronization/BeaconHeadersSyncFeed.cs
```diff
@@ -36,7 +36,15 @@ public sealed class BeaconHeadersSyncFeed : HeadersSyncFeed
...
 
-    protected override BlockHeader? LowestInsertedBlockHeader => _blockTree.LowestInsertedBeaconHeader;
+    protected override BlockHeader? LowestInsertedBlockHeader
+    {
+        get => _blockTree.LowestInsertedBeaconHeader;
+        set
+        {
+            // Set in blocktree
+        }
+    }
```
LukaszRozmej, line 46: Looks fishy, can you elaborate in comment?

src/Nethermind/Nethermind.Synchronization/FastBlocks/FastBlocksBatch.cs
asdacap: Probably

src/Nethermind/Nethermind.Synchronization/FastBlocks/FastHeadersSyncFeed.cs
asdacap: Moved outside

src/Nethermind/Nethermind.Synchronization/FastBlocks/FastHeadersSyncFeed.cs
asdacap: Changed to `AsSpan().Reverse()`.

LukaszRozmej (approved):

PR #7863, opened 2024-12-04
Add DEBUG logs to output when sync fails

kamilchodola:
Adding a DEBUG logs to output of an action whenever it fails.
Helpful when investigating syncing problems, json parsing errors etc.

Additionally adjusted machines types as some of them were too big for their purpose (sync only to tip) - may make action a bit smaller but surely less costly.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [X] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

- [ ] Yes
- [X] No

- [ ] Yes
- [X] No

### DIFF
--- a/.github/workflows/sync-supported-chains.yml
+++ b/.github/workflows/sync-supported-chains.yml
@@ -125,7 +125,7 @@ jobs:
       - name: Install Sedge environment
         run: |
           echo "Downloading sedge sources..."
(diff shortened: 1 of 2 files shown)


brbrr (approved):
LGTM! One small note about repetition of the log config folder creation. Maybe we could use some variables to create the folder after all the branching?

kamilchodola:
@brbrr good suggestion - will apply it in follow up PR

PR #7871, opened 2024-12-05
Annotate debug errors

benaadams:
## Changes

- Add `DEBUG/ERROR` to errors output during Debug logging

- [x] Refactoring

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
@@ -81,7 +81,7 @@ public sealed class BlockCachePreWarmer(ReadOnlyTxProcessingEnvFactory envFactor
         }
         catch (Exception ex)
         {
-            if (_logger.IsDebug) _logger.Warn($"Error pre-warming {suggestedBlock.Number}. {ex}");
+            if (_logger.IsDebug) _logger.Warn($"DEBUG/ERROR Error pre-warming {suggestedBlock.Number}. {ex}");
         }
         finally
         {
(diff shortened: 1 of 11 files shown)


LukaszRozmej (approved):

benaadams:
> Not sure about this approach. Maybe it's better to display log levels for each message? Say, in 3 letter labels like "INF", "WRN", "ERR", "DBG".

You are trying to make @kamilchodola sad 😅

benaadams:
> Not sure about this approach. Maybe it's better to display log levels for each message? Say, in 3 letter labels like "INF", "WRN", "ERR", "DBG".

Is so can log debug to file and filter out these as noise
```
<logger name="*" minlevel="Info" writeTo="auto-colored-console-async">
    <filters>
        <when condition="contains('${message}', 'DEBUG/ERROR')" action="Ignore" />
            <when condition="true" action="Log" />
            </filters>
</logger>
```

Though question might be why do we do mismatched logging?

kamilchodola:
@rubo Case is - it is not actually DBG log - it is an Error which is thrown only if debug logs are enabled. 
So imho - something worth knowing about but not important enough to be a "standard error". 

But what it does to me personally is - i have some scripts which are parsing logs to make sure sync went well and nothing unexpected happened. When I enable debug logging to file only (keeping info to file) I still get some unexpected logs like this to info level and need any way to filter those out to not make my tests fail.

LukaszRozmej:
Yeah it is special diagnostic case, bit of a hack, but whatever.

rubo:
> it is not actually DBG log - it is an Error which is thrown only if debug logs are enabled

Then things are more confusing, if not misleading:

```cs
 if (_logger.IsDebug) _logger.Error()
```

It kinda defeats the purpose of having log levels. I don't have any suggestions at the moment but generally, we need to come up with a better approach. Otherwise, these hacky things tend to replicate themselves here and there.

kamilchodola:
Agree and already was thinking on how to solve that but didn't come with any solution for now.

Reason for this PR was actually that this happened again :D

PR #7873, opened 2024-12-06
Fix/error on shutdown

asdacap:
- Fix error on shutdown very early
- Fix error on shutdown on statesync

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Reproducable by adding `Thread.Sleep` in treesync.

### DIFF
--- a/src/Nethermind/Nethermind.Synchronization/ParallelSync/SyncDispatcher.cs
+++ b/src/Nethermind/Nethermind.Synchronization/ParallelSync/SyncDispatcher.cs
@@ -115,20 +115,22 @@ namespace Nethermind.Synchronization.ParallelSync
 
                             // Use Task.Run to make sure it queues it instead of running part of it synchronously.
                             _activeTasks.AddCount();
-                        try
-                        {
-                            Task task = Task.Run(
-                                () =>
-                                {
-                                    try
+
+                            Task task;
+                            try
+                            {
+                                task = Task.Run(
+                                    () =>
                                     {
-                                        return DoDispatch(cancellationToken, allocatedPeer, request, allocation);
-                                    }
-                                    finally
-                                    {
-                                        _activeTasks.Signal();
-                                    }
-                                });
+                                        try
+                                        {
+                                            return DoDispatch(cancellationToken, allocatedPeer, request, allocation);
+                                        }
+                                        finally
+                                        {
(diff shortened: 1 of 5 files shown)


src/Nethermind/Nethermind.Synchronization/ParallelSync/SyncDispatcher.cs
```diff
@@ -109,7 +114,19 @@ public async Task Start(CancellationToken cancellationToken)
...
+                                {
+                                    try
+                                    {
+                                        return DoDispatch(cancellationToken, allocatedPeer, request, allocation);
+                                    }
+                                    finally
+                                    {
+                                        _activeTasks.Signal();
+                                    }
+                                });
```
LukaszRozmej, line 129: Unexpected scenario, but for completion:
```suggestion
                        try
                        {
                            Task task = Task.Run(
                                () =>
                                {
                                    try
                                    {
                                        return DoDispatch(cancellationToken, allocatedPeer, request, allocation);
                                    }
                                    finally
                                    {
                                        _activeTasks.Signal();
                                    }
                                });
                            }
                            catch
                            {
                                _activeTasks.Signal();                            
                            }
```

src/Nethermind/Nethermind.Synchronization/ParallelSync/SyncDispatcher.cs
```diff
@@ -272,5 +289,22 @@ private void UpdateState(SyncFeedState state)
...
+
+        public ValueTask DisposeAsync()
+        {
+            if (_disposed)
+            {
+                return ValueTask.CompletedTask;
+            }
+            _disposed = true;
+
+            _activeTasks.Signal();
```
LukaszRozmej, line 301: Can the event start with count 0 and we can drop this signal?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Synchronization/ParallelSync/SyncDispatcher.cs
asdacap: No, if it start with 0 once something add to it, then back to 0, it will throw on new task.

LukaszRozmej (approved):

PR #7874, opened 2024-12-06
All projects target same build output

LukaszRozmej:
## Changes

- Point all projects to same build output
- Reduces build size 30GB->600MB for each release configuration (debug/release).
- fixes version of c#, so if someone in the future will come back to previous commits he will have guaranteed builds (latest can have backward compatibility issues!)

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [x] Other: build

#### Requires testing

- [x] Yes
- [ ] No

#### Notes on testing

Check if release still works fine.
If it is problematic, we can have a separate output for Runner project, which should be compatible with the current release mode.


LukaszRozmej:
We just need to do that for tests

PR #7879, opened 2024-12-09
Increase default gas limit from 30M to 36M

benaadams:
## Changes

- 36M is the current safe limit for all participants pre-Pectra (EIP-7623) and taking into account an active propagation size limit bug in CL clients.

See also: https://ethresear.ch/t/on-increasing-the-block-gas-limit-technical-considerations-path-forward/21225

Also being done by Erigon: https://github.com/erigontech/erigon/pull/13057

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Runner.Test/ConfigFilesTests.cs
+++ b/src/Nethermind/Nethermind.Runner.Test/ConfigFilesTests.cs
@@ -346,9 +346,9 @@ public class ConfigFilesTests : ConfigFileTestsBase
 
     [TestCase("chiado", 17_000_000L, 5UL, 3000)]
     [TestCase("gnosis", 17_000_000L, 5UL, 3000)]
-    [TestCase("mainnet", 30_000_000L)]
-    [TestCase("sepolia", 30_000_000L)]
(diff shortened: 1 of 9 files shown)


LukaszRozmej (approved):

kamilchodola (approved):

kamilchodola:
@benaadams few more tests to be adjusted like:
Test_StoreBeaconRoot_AccessListNotNull

benaadams:
> Test_StoreBeaconRoot_AccessListNotNull

Why is it not failing? 🤔

benaadams:
> Can search by a 30_000_000 phrase in entire solution

Ah, got yah

smartprogrammer93 (approved):
LFG

benaadams:
> @benaadams few more tests to be adjusted like: 
>
> ...
>
> Can search by a 30_000_000 phrase in entire solution

I don't think they should change as the block limit will still be 30M so we shouldn't bake in a higher value (though perhaps they should be reading from block header instead)

LukaszRozmej:
System calls should stay 30mln

PR #7881, opened 2024-12-10
Fix op-mainnet snapshot

deffrian:
Fixes Closes Resolves #
Pre-bedrock legacy transactions might not have signature

## Changes

- Adds op legacy tx decoder and signature verifier

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Optimism/OptimismLegacyTxDecoder.cs
+++ b/src/Nethermind/Nethermind.Optimism/OptimismLegacyTxDecoder.cs
@@ -1,0 +1,34 @@
+// SPDX-FileCopyrightText: 2024 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System;
+using Nethermind.Core;
+using Nethermind.Core.Crypto;
+using Nethermind.Core.Specs;
+using Nethermind.Serialization.Rlp;
+using Nethermind.Serialization.Rlp.TxDecoders;
+using Nethermind.TxPool;
+
+namespace Nethermind.Optimism;
+
+public class OptimismLegacyTxDecoder : LegacyTxDecoder<Transaction>
+{
+    protected override Signature? DecodeSignature(ulong v, ReadOnlySpan<byte> rBytes, ReadOnlySpan<byte> sBytes, Signature? fallbackSignature = null,
+        RlpBehaviors rlpBehaviors = RlpBehaviors.None)
+    {
+        return base.DecodeSignature(v, rBytes, sBytes, fallbackSignature, rlpBehaviors | RlpBehaviors.AllowUnsigned);
+    }
+}
+
+public class OptimismLegacyTxValidator(OptimismSpecHelper specHelper) : ITxValidator
+{
+    public ValidationResult IsWellFormed(Transaction transaction, Block? block, IReleaseSpec releaseSpec)
+    {
+        if (block is null || specHelper.IsBedrock(block.Header))
+        {
+            return transaction.Signature is null ? new ValidationResult("Empty signature") : ValidationResult.Success;
+        }
+
+        return ValidationResult.Success;
+    }
+}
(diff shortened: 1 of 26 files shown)


src/Nethermind/Nethermind.Optimism/OptimismLegacyTxDecoder.cs
```diff
@@ -0,0 +1,38 @@
...
+        }
+        return base.DecodeSignature(v, rBytes, sBytes, fallbackSignature, rlpBehaviors);
+    }
+}
+
+public class OptimismLegacyTxValidator(OptimismSpecHelper specHelper) : ITxValidator
+{
+    public ValidationResult IsWellFormed(Transaction transaction, IReleaseSpec releaseSpec, BlockHeader? header = null)
+    {
+        if (header is null || specHelper.IsBedrock(header))
```
LukaszRozmej, line 31: Could IsBedrock be in `IReleaseSpec`, then we don't have to pass header?

`IReleaseSpec` was created to have all fork information.

LukaszRozmej (changes requested):

src/Nethermind/Nethermind.Optimism/OptimismLegacyTxDecoder.cs
deffrian: we can check if eip1559 is enabled. Not sure it's good idea. But as I know ReleaseSpec is not aware of optimism forks

LukaszRozmej: Would be good to make IReleaseSpec extendable by plugins then

LukaszRozmej (approved):

src/Nethermind/Nethermind.Optimism/OptimismLegacyTxDecoder.cs
```diff
@@ -0,0 +1,39 @@
...
+}
+
+public class OptimismLegacyTxValidator : ITxValidator
+{
+    public ValidationResult IsWellFormed(Transaction transaction, IReleaseSpec releaseSpec)
+    {
+        // In op bedrock eip1559 activated with bedrock
+        if (releaseSpec.IsEip1559Enabled)
+        {
+            return transaction.Signature is null ? new ValidationResult("Empty signature") : ValidationResult.Success;
```
emlautarom1, line 34: Before we were not performing any validations on Deposit transactions, but now we require that post-Bedrock transactions contain a non-null signature. Is this correct? If so, shouldn't we also validate the signature (using `SignatureTxValidator`)?

deffrian: if signature is null here it means that while decoding signature in `DecodeSignature` we got inside  `if (v == 0 && rBytes.IsEmpty && sBytes.IsEmpty)`. Without this pr we would crash inside `DecodeSignature`.

emlautarom1: That makes sense, what I'm missing is why we don't validate the signature of a Legacy transaction post bedrock. More so, why are we not performing the standard Legacy validations:

https://github.com/NethermindEth/nethermind/blob/d00410fbba8dd79d287c428c9aa4b02f399f9737/src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs#L24-L30

With these changes, Legacy transactions that are processed in Optimism are not validated at all, when previously they went through several checks.

deffrian: Oh, that means I made a mistake. But why we don't check DepositTransaction?
`api.RegisterTxType<OptimismTransactionForRpc>(new OptimismTxDecoder<Transaction>(), Always.Valid);`

emlautarom1: It dates back to the original Optimism plugin PR where the `OptimismTxValidator` performed no checks on Deposit transactions

https://github.com/NethermindEth/nethermind/blob/6edd397acfea5f262796db5a0f15592cf5d6cf53/src/Nethermind/Nethermind.Optimism/OptimismTxValidator.cs#L20-L26

We didn't change the logic in the Transaction refactoring, so the current validator for Deposit transactions always succeeds.

emlautarom1: Addressed in https://github.com/NethermindEth/nethermind/pull/7887.

PR #7882, opened 2024-12-10
fix timestamp differences

rjnrohit:
THe PR fixes dummy block addition issues for `eth_mutlicall`


src/Nethermind/Nethermind.Config/BlocksConfig.cs
```diff
@@ -23,6 +23,8 @@ public class BlocksConfig : IBlocksConfig
 
         public ulong SecondsPerSlot { get; set; } = 12;
 
+        public ulong SimulateSecondsPerSlot { get; set; } = 1;
```
LukaszRozmej, line 26: Maybe it fits better in `JsonRcpConfig`?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Config/BlocksConfig.cs
rjnrohit: Yes, we can do it if we want to make it configurable.

MarekM25 (changes requested):
Have you discussed it with Geth Devs? I think 12 seconds is better. What is the reasoning for 1nd?

LukaszRozmej:
> Have you discussed it with Geth Devs? I think 12 seconds is better. What is the reasoning for 1nd?

Yes this is actually unifying eth_simulate space. Main reasoning is to have similar cross-chain behaviour.

rjnrohit:
> Have you discussed it with Geth Devs? I think 12 seconds is better. What is the reasoning for 1nd?

Yes, I had a call with the eth_simulate devs teams; geth has 1 second increment, you can check from here: https://github.com/s1na/go-ethereum/blob/f25fcd34e11061568c187c44dd02fb6e9c8546de/internal/ethapi/simulate.go#L48

MarekM25:
> > Have you discussed it with Geth Devs? I think 12 seconds is better. What is the reasoning for 1nd?
> 
> Yes this is actually unifying eth_simulate space. Main reasoning is to have similar cross-chain behaviour. User can provide overrides for it if he wants.

yes, so that's why our SecondsPerSlot can be used, providing overrides sounds worse

MarekM25:
> > Have you discussed it with Geth Devs? I think 12 seconds is better. What is the reasoning for 1nd?
> 
> Yes, I had a call with the eth_simulate devs teams; geth has 1 second increment, you can check from here: https://github.com/s1na/go-ethereum/blob/f25fcd34e11061568c187c44dd02fb6e9c8546de/internal/ethapi/simulate.go#L48

but why? I think the better approach would be to do SecondsPerSlot and this value is configurable

MarekM25 (commented):
I don't think we need a seperate config value for this, either hardcode 1 and people will override it through RPC -> like geth
or just do what we are doing now and discuss with the geth team

rjnrohit:
> I don't think we need a seperate config value for this, either hardcode 1 and people will override it through RPC -> like geth or just do what we are doing now and discuss with the geth team

for geth, it is completely hardcoded and not configurable either

MarekM25:
> > I don't think we need a seperate config value for this, either hardcode 1 and people will override it through RPC -> like geth or just do what we are doing now and discuss with the geth team
> 
> for geth, it is completely hardcoded and not configurable either

https://github.com/s1na/go-ethereum/pull/new/simulate/timestamp-default

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/SimulateTxExecutor.cs
```diff
@@ -120,91 +121,67 @@ public override ResultWrapper<IReadOnlyList<SimulateBlockResult>> Execute(
                 $"Too many blocks provided, node is configured to simulate up to {_blocksLimit} while {call.BlockStateCalls?.Count} were given",
                 ErrorCodes.InvalidParams);
 
-        secondsPerSlot ??= new BlocksConfig().SecondsPerSlot;
+        ulong secondsPerSlot = new BlocksConfig().SecondsPerSlot;
```
LukaszRozmej, line 124: you cannot rely on `new` here, as you will ignore configured value

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/SimulateTxExecutor.cs
```diff
@@ -120,91 +121,67 @@ public override ResultWrapper<IReadOnlyList<SimulateBlockResult>> Execute(
...
-                        StateOverrides = null,
-                        Calls = []
-                    });
-                }
-            }
-
-            call.BlockStateCalls.Sort((b1, b2) => b1.BlockOverrides!.Number!.Value.CompareTo(b2.BlockOverrides!.Number!.Value));
+            // sort element of the completeBlockStateCalls by block number
+            completeBlockStateCalls.Sort((a, b) => a.BlockOverrides.Number!.Value.CompareTo(b.BlockOverrides.Number!.Value));
+            call.BlockStateCalls = [.. completeBlockStateCalls];
```
LukaszRozmej, line 184: Why do we need to sort at the end? Shouldn't we just error out if it wasn't consecutive in the first place?

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/SimulateTxExecutor.cs
```diff
@@ -120,91 +121,67 @@ public override ResultWrapper<IReadOnlyList<SimulateBlockResult>> Execute(
...
                 ErrorCodes.InvalidParams);
 
-        secondsPerSlot ??= new BlocksConfig().SecondsPerSlot;
+        ulong secondsPerSlot = new BlocksConfig().SecondsPerSlot;
 
         if (call.BlockStateCalls is not null)
         {
-            long lastBlockNumber = -1;
-            ulong lastBlockTime = 0;
+            long lastBlockNumber = header.Number;
```
LukaszRozmej, line 128: Can we break down this setup to some meaningful helper methods? It is hard to follow the intent.

LukaszRozmej (changes requested):

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/SimulateTxExecutor.cs
rjnrohit: Okay, this is what we had before, let me change it.

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/SimulateTxExecutor.cs
rjnrohit: Sure

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/SimulateTxExecutor.cs
rjnrohit: Yes you are correct, there is no need for sorting. Anyway we don't have to check whether the blocks are in consecutive order, as the above the block will error out/

rjnrohit: Yes, you are correct; we don't need to sort. We don't need to add a check either for the blocks to be consecutive as  the for loop will take care of it

LukaszRozmej (approved):

MarekM25 (approved):

PR #7884, opened 2024-12-10
Bls pairing check optimistation: precompute lines

Marchhill:
## Changes

- Precompute lines for Miller loop

- [x] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

#### Notes on testing

Covered by state tests.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Evm/Precompiles/Bls/PairingCheckPrecompile.cs
+++ b/src/Nethermind/Nethermind.Evm/Precompiles/Bls/PairingCheckPrecompile.cs
@@ -19,6 +19,7 @@ namespace Nethermind.Evm.Precompiles.Bls;
 public class PairingCheckPrecompile : IPrecompile<PairingCheckPrecompile>
 {
     private const int PairSize = 384;
+    private const int PrecomputeLinesSize = 6 * 6 * 68;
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Evm/Precompiles/Bls/PairingCheckPrecompile.cs
```diff
@@ -42,9 +43,12 @@ private PairingCheckPrecompile() { }
         G1 x = new(stackalloc long[G1.Sz]);
         G2 y = new(stackalloc long[G2.Sz]);
 
-        using ArrayPoolList<long> buf = new(GT.Sz * 2, GT.Sz * 2);
+        int sz = GT.Sz * 2 + PrecomputeLinesSize;
+        using ArrayPoolList<long> buf = new(sz, sz);
```
LukaszRozmej, line 47: stack space wouldn't be better? It is > 2kb so not sure.

MarekM25 (commented):
it seems that it might be not needed, do you think it is worth adding?

PR #7892, opened 2024-12-11
Add additional block lag for better sync mode switching for chain with shorter block processing time

damian-orzechowski:
For linea (and probably other chains with short block time), there is a problem with switching between various phases of sync causing node to fall into a long running loop, though could already start processing blocks. The problem occurs for:
- switching from fast headers to state sync (snap)
- switching from state sync to block processing

In the case for linea (no beacon), best header is set based on latest downloaded header from peer. Target block is assumed to be best peer's head. When headers are downloaded, the `FastSyncLag` (32 blocks) is used as a distance from the target, so best header is always minimum 32 blocks behind target. Conditions for switching between fast sync, state sync and block processing also use the criteria of having block `header >= target - FastSyncLag`, which effectively means some transitions can only happen when header is exactly 32 blocks behind target. Similar case is with target state sync block which is set to best header upon setting pivot for sync. With 2s block processing time, chain can advance fast enough for this condition never to be fulfilled considering `MulitSyncModeSelected` check frequency (1s).

## Changes

The fix is to add additional lag to be used when checking sync transition to allow more distance between target and header / state, so that `MulitSyncModeSelected` can evaluate the change conditions even when chain is progressing fast.

Added new configuration for linea chains only to minimize the impact.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


LukaszRozmej:
Maybe just make `FastSyncLag` configurable?

damian-orzechowski:
> Maybe just make `FastSyncLag` configurable? We could similarly make check frequency configurable. Or we could make them dependent on `SecondsPerSlot`

Changing `FastSyncLag` alone won't have an impact, because it is used for both downloading header and checking sync switch transition. We'll end up with the same - only switch when `header == target - FastSyncLag`.

asdacap (approved):

src/Nethermind/Nethermind.Blockchain/Synchronization/SyncConfig.cs
```diff
@@ -74,6 +74,11 @@ public string? PivotHash
         public int StateMinDistanceFromHead { get; set; } = 32;
         public bool GCOnFeedFinished { get; set; } = true;
 
+        /// Additional delay in blocks between best suggested header and synced state to allow faster state switching for PoW chains
+        /// with higher block processing frequency. Effectively this is the max allowed difference between best header (used as sync
+        /// pivot) and synced state block, to assume that state is synced and node can start processing blocks
```
LukaszRozmej, line 79: Why not doc comment?

src/Nethermind/Nethermind.Synchronization/ParallelSync/MultiSyncModeSelector.cs
```diff
@@ -399,7 +399,7 @@ private bool ShouldBeInFastSyncMode(Snapshot best)
             // and we need to sync away from it.
             // Note: its ok if target block height is not accurate as long as long full sync downloader does not stop
             //  earlier than this condition below which would cause a hang.
-            bool notReachedFullSyncTransition = best.Header < best.TargetBlock - FastSyncLag;
+            bool notReachedFullSyncTransition = best.Header < best.TargetBlock - FastSyncLag - _syncConfig.HeaderStateDistance;
```
LukaszRozmej, line 402: Maybe add `FastSyncLag` and `HeaderStateDistance` once and use it in all 3 places?

LukaszRozmej (approved):

PR #7896, opened 2024-12-11
Prevent file logging before configuration initialization

rubo:
## Changes

This PR prevents file logging before configuration initialization. The reason is that now Nethermind uses the log rules at the very beginning to avoid having sudden log style changes after loading the configuration. However, on some user configurations, the Nethermind directory is inaccessible for the Nethermind process, and that causes an error.

Another approach might be writing that part of the logs in the user's temp directory. But till now, that logs were missing, and perhaps having them in the file somewhere is not that useful.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

Tested manually.

### DIFF
--- a/src/Nethermind/Nethermind.Logging.NLog/NLogManager.cs
+++ b/src/Nethermind/Nethermind.Logging.NLog/NLogManager.cs
@@ -20,6 +20,11 @@ namespace Nethermind.Logging.NLog
         private const string DefaultFileTargetName = "file-async_wrapped";
         private const string DefaultFolder = "logs";
 
+        /// <summary>
+        /// The constructor to use when the configuration is not yet initialized.
+        /// </summary>
+        public NLogManager() { /* Log in temp dir? */ }
+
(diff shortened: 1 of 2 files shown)


MarekM25:
Do you know which PR could add this bug? It would be easier to understand the fix

rubo:
> Do you know which PR could add this bug? It would be easier to understand the fix

I don't recall which one exactly, but one of mine. Logs have been changed from this

old

to this

new

so they have their colors and formatting right from the start.

Since the configuration is not initialized at that early stage, the logger instance logs in the default data directory, i.e., the Nethermind directory. And for some setups, it causes access issues. Thus, this PR disables file logging for that stage (like we had previously) while keeping the nice logs.

PR #7898, opened 2024-12-11
Use Ordinal rather than InvariantCulture

benaadams:
## Changes

- We aren't doing any culture sensitive tests, so just use Ordinal

- [x] Optimization

#### Requires testing

- [x] No


rubo (approved):
Finally :)

PR #7906, opened 2024-12-13
Feature/disable nodedata on halfpath

asdacap:
- Disable nodedata protocol on halfpath as it cannot serve it.
- This prevent other nethermind node from trying.

- [X] Optimization

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Tested by logging at hello message.

### DIFF
--- a/src/Nethermind/Nethermind.State/IWorldStateManager.cs
+++ b/src/Nethermind/Nethermind.State/IWorldStateManager.cs
@@ -11,6 +11,7 @@ public interface IWorldStateManager
     IWorldState GlobalWorldState { get; }
     IStateReader GlobalStateReader { get; }
     IReadOnlyTrieStore TrieStore { get; }
+    bool SupportHashLookup { get; }
 
     /// <summary>
     /// Used by read only tasks that need to execute blocks.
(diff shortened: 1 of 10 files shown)


src/Nethermind/Nethermind.State/IWorldStateManager.cs
```diff
@@ -11,6 +11,7 @@ public interface IWorldStateManager
     IWorldState GlobalWorldState { get; }
     IStateReader GlobalStateReader { get; }
     IReadOnlyTrieStore TrieStore { get; }
+    bool SupportHashLookup { get; }
```
LukaszRozmej, line 14: Not sure if needed as we already expose `TrieStore` and it already kind-of exposes `Scheme`, so maybe extension method would suffice? Or default implementation?

LukaszRozmej (approved):

src/Nethermind/Nethermind.State/IWorldStateManager.cs
asdacap: Ideally, we should not expose triestore at all. Its just a bit.. un-nice at the moment.

PR #7919, opened 2024-12-16
Slight db adjustment for snap sync perf

asdacap:
- Enable `unordered_write` for statedb which helps increase normal peak write rates during snap sync from about 1.5 million writes per sec to around 2.0 writes per sec.
- Increase write buffer count to the same as cpu cores, which eliminate stall due to waiting for flush at higher speed.
- Slightly increase maximum snap response latency from 3 second to 3.5 second.
- Overall impact is not much as it still take time to reach full speed and need some time to download large storage at the end, and I don't think I have a good enough internet.

- [X] Optimization

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

#### Notes on testing

- Mainnet sync ok.


LukaszRozmej:
https://github.com/NethermindEth/nethermind/pull/7733 will fix partitioning large storage tries

LukaszRozmej (approved):

PR #7921, opened 2024-12-16
fix tracing posdao blocks

LukaszRozmej:
## Changes

- Fixes tracing in posdao by using correct `AuraBlockProcessor`

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

## Remarks

@emlautarom1 @deffrian @flcl42 I see very similar approach in Optimism and Taiko plugins. I think we should unify and refactor it in a way that would allow us to inject `BlockProcessor` type into the traces (`ReadOnlyEnvironment`) better. WDYT?

### DIFF
--- a/src/Nethermind/Nethermind.Consensus.AuRa/InitializationSteps/RegisterAuRaRpcModules.cs
+++ b/src/Nethermind/Nethermind.Consensus.AuRa/InitializationSteps/RegisterAuRaRpcModules.cs
@@ -1,4 +1,4 @@
-﻿// SPDX-FileCopyrightText: 2024 Demerzel Solutions Limited
+// SPDX-FileCopyrightText: 2024 Demerzel Solutions Limited
 // SPDX-License-Identifier: LGPL-3.0-only
 
 using System.Collections.Generic;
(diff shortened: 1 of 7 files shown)


flcl42 (approved):

emlautarom1:
From a quick look I can't see how this refactor would look like. Could you share a small snippet of the end goal with such refactor?

Honestly, every time I look into initialization code I shudder: there is a lot going on and temporal coupling is noticeable. I still haven't seen the benefits of the WIP DI approach, but maybe it's just a "me" problem.

LukaszRozmej:
> From a quick look I can't see how this refactor would look like. Could you share a small snippet of the end goal with such refactor?
> 
> Honestly, every time I look into initialization code I shudder: there is a lot going on and temporal coupling is noticeable. I still haven't seen the benefits of the WIP DI approach, but maybe it's just a "me" problem.

Yeah, it isn't obvious, so I haven't done it yet. But you would have to have some factory for ReadOnlyChainProcessingEnv so it could also create a typed BlockProcessor.

Then you would only change the factory between standard/AuRa/Taiko/Optimism.

PR #7923, opened 2024-12-16
Snap: Extension Rlp value index is 1

benaadams:
## Changes

- We use 0 in one place and 1 in another; leading to over work

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Synchronization/SnapSync/SnapProviderHelper.cs
+++ b/src/Nethermind/Nethermind.Synchronization/SnapSync/SnapProviderHelper.cs
@@ -17,6 +17,8 @@ namespace Nethermind.Synchronization.SnapSync
 {
     public static class SnapProviderHelper
     {
+        private const int ExtensionRlpChildIndex = 1;
+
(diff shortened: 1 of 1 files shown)


damian-orzechowski (approved):

src/Nethermind/Nethermind.Synchronization/SnapSync/SnapProviderHelper.cs
```diff
@@ -17,6 +17,8 @@ namespace Nethermind.Synchronization.SnapSync
 {
     public static class SnapProviderHelper
     {
+        private const int ExtensionRlpChildIndex = 1;
```
asdacap, line 20: Shouldnt it always be 0?

asdacap: Ow, no, sometimes its 1 its 0. That can't be good.

benaadams: Some methods add 1 if extension (converting 0 -> 1); is second element in rlp for extension (first being key); these methods doesn't add one

PR #7925, opened 2024-12-17
Add filter for max transaction size

marcindsobczak:
## Changes

- add filter for max transaction size
- setting max tx size to 128 KiB which is inline with Geth https://github.com/ethereum/go-ethereum/blob/master/core/txpool/legacypool/legacypool.go#L56

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No


LukaszRozmej (commented):
We could add similar logic in transaction gossip, before we even deserialise RLP.
Do we want exception for local transactions? (probably not)

src/Nethermind/Nethermind.TxPool/Filters/SizeTxFilter.cs
```diff
@@ -0,0 +1,31 @@
...
+internal sealed class SizeTxFilter(ITxPoolConfig txPoolConfig, ILogger logger) : IIncomingTxFilter
+{
+    private readonly long _configuredMaxTxSize = txPoolConfig.MaxTxSize ?? long.MaxValue;
+
+    public AcceptTxResult Accept(Transaction tx, ref TxFilteringState state, TxHandlingOptions txHandlingOptions)
+    {
+        if (tx.SupportsBlobs)
+        {
+            return AcceptTxResult.Accepted;
+        }
```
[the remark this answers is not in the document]
marcindsobczak: I was sure it can't, but looking at the spec and code - there is no restriction on it. Just can't create a contract (`to` can't be `null`).

Geth is limiting non-blob part to 1MB https://github.com/ethereum/go-ethereum/blob/master/core/txpool/blobpool/blobpool.go#L63-L68

I think I will add here a limit of 8x standard limit (8x128KiB).

Great catch!

marcindsobczak:
Good idea with not requesting txs with size above max limit!

About skipping deserialization - we need to deserialise at least a part of tx to check tx type. But I think we can reject at the very beginning all txs with size exceeding maximally loaded blob tx (max number of blobs + max size of non-blob data).

About exception for local txs - I don't think so. If user would need to send such tx, there is always an option to increase config value

LukaszRozmej:
> Good idea with not requesting txs with size above max limit!
> 
> About skipping deserialization - we need to deserialise at least a part of tx to check tx type. But I think we can reject at the very beginning all txs with size exceeding maximally loaded blob tx (max number of blobs + max size of non-blob data).
> 
> About exception for local txs - I don't think so. If user would need to send such tx, there is always an option to increase config value

We can get the type without deserializing anything else, it is just first few bytes.

src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/PooledTxsRequestor.cs
```diff
@@ -69,6 +71,12 @@ public void RequestTransactionsEth68(Action<V66.Messages.GetPooledTransactionsMe
                 int txSize = discoveredTxHashesAndSizes[i].Size;
                 TxType txType = (TxType)discoveredTxHashesAndSizes[i].Type;
 
+                if (txType != TxType.Blob && txSize > _configuredMaxTxSize
+                    || txType == TxType.Blob && txSize > _configuredMaxTxSize * 8 + (long)Eip4844Constants.MaxBlobGasPerBlock)
```
LukaszRozmej, line 75:
```suggestion
                long maxSize = txType != TxType.Blob ? _configuredMaxTxSize : _configuredMaxTxSize * 8 + (long)Eip4844Constants.MaxBlobGasPerBlock;
                if (txSize > maxSize)
```

Also on tx we have `SupportsBlobs` which can be more than just `TxType.Blob` in the future. Would be good to have it as extension method on `TxType`.

src/Nethermind/Nethermind.Core/Transaction.cs
```diff
@@ -334,7 +334,7 @@ public class SystemTransaction : Transaction
     /// <remarks>Created because of cyclic dependencies between Core and Rlp modules</remarks>
     public interface ITransactionSizeCalculator
     {
-        int GetLength(Transaction tx);
+        int GetLength(Transaction tx, bool doNotCountBlobs = false);
```
LukaszRozmej: `doNotCountBlobs` here but `shouldCountBlobs` above - confusing. Also I would prefer positive over negative as it is easier to reason about.

src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/PooledTxsRequestor.cs
LukaszRozmej: Maybe also just calculate it once?

src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/PooledTxsRequestor.cs
```diff
@@ -18,13 +18,15 @@ public class PooledTxsRequestor : IPooledTxsRequestor
...
         private readonly bool _blobSupportEnabled;
+        private readonly long _configuredMaxTxSize;
 
         private readonly ClockKeyCache<ValueHash256> _pendingHashes = new(MemoryAllowance.TxHashCacheSize);
 
         public PooledTxsRequestor(ITxPool txPool, ITxPoolConfig txPoolConfig)
         {
             _txPool = txPool;
             _blobSupportEnabled = txPoolConfig.BlobsSupport.IsEnabled();
+            _configuredMaxTxSize = txPoolConfig.MaxTxSize ?? long.MaxValue;
```
LukaszRozmej, line 29: Maybe have `txPoolConfig.MaxBlobTxSize` which is just calculated?

marcindsobczak:
> > Good idea with not requesting txs with size above max limit!
> > About skipping deserialization - we need to deserialise at least a part of tx to check tx type. But I think we can reject at the very beginning all txs with size exceeding maximally loaded blob tx (max number of blobs + max size of non-blob data).
> > About exception for local txs - I don't think so. If user would need to send such tx, there is always an option to increase config value
> 
> We can get the type without deserializing anything else, it is just first byte.

There are some exceptions like legacy txs without tx type. For sure it's doable, but requires more insights, time, caution and extended testing, which I don't want to do in the scope of this PR. I added an issue for it: https://github.com/NethermindEth/nethermind/issues/7986

Here I added:
- `TxType` comparisons as `TxTypeExtension`
- limiting max blob tx size (excluding blobs)
- max blob tx size as config value
- not requesting too large txs in `PooledTxsRequestor`

src/Nethermind/Nethermind.TxPool/TxPoolConfig.cs
```diff
@@ -21,6 +21,7 @@ public class TxPoolConfig : ITxPoolConfig
         public int HashCacheSize { get; set; } = 512 * 1024;
         public long? GasLimit { get; set; } = null;
         public long? MaxTxSize { get; set; } = 128.KiB();
+        public long? MaxBlobTxSize { get; set; } = 1.MiB();
```
LukaszRozmej, line 24: why blopb tx's can be 8x bigger?

marcindsobczak: Doesn't have sense to me, but I wanted to do it similar to Geth implementation
https://github.com/ethereum/go-ethereum/blob/master/core/txpool/blobpool/blobpool.go#L63-L68

marcindsobczak: And it is not a big deal as blob txs are saved to disc. Memory is not affected as in case of non-blob txs

marcindsobczak: We can also go with standard 128 KiB and it will simplify this PR, but we will then reject some txs accepted by Geth

LukaszRozmej (approved):

PR #7931, opened 2024-12-18
Bls precompile: Use MUL for single point MSM

Marchhill:
## Changes

- For MSM with a single point use MUL

- [x] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

Covered by state tests

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


MarekM25 (commented):
Looks good. Do we need this though? MSM should be good enough?

Marchhill:
> Looks good. Do we need this though? MSM should be good enough?

It's more efficient this way as we don't have overhead of MSM

MarekM25 (approved):

PR #7939, opened 2024-12-19
feat: standardized error responses for payload validation

### LINKED ISSUE #7901: feat: add standardized error messages in engine API responses
As it currently stands, Geth and Nethermind have different engine API error response formats, especially the ones relating to new payload validation. Here is an example:

Geth: `blockhash mismatch, want %x, got %x`
Ref: https://github.com/ethereum/go-ethereum/blob/c7e740f40ca93cd2b8681b97af71e0b20207ca09/beacon/engine/types.go#L222

Nethermind: `InvalidHeaderHash: Header hash does not match.`
Ref: https://github.com/NethermindEth/nethermind/blob/3f19aeefffea18899890d110b7bd44c0630ffb0d/src/Nethermind/Nethermind.Consensus/Messages/BlockErrorMessages.cs#L50

The most relevant difference in these two messages is that Nethermind doesn't include the expected block hash whereas Geth does. This same pattern exists for other types, such as the following mismatching fields:

* gas used
* receipts root
* logs bloom
* state root

Standardizing these messages could be useful for two reasons: operators running different EL/CL combinations would have a more unified debugging experience, and as a secondary benefit receiving "hints" from EL block validation _could be used to build blocks out of the EL client_. 

As an example, for https://github.com/chainbound/bolt we use the engine API response errors to parse hints to generate a valid local payload that includes a set of transactions (i.e. similar to how a same-slot inclusion list protocol would work). Of course this is far from an ideal block building setup, however it's also very low effort compared to other alternatives (if you are interested, we explore some of these [here](https://github.com/chainbound/bolt/discussions/59)).

Happy to open a PR for this and answer any questions!

merklefruit:
Fixes #7901

## Changes

* Added more descriptive payload validation error responses in engine API

- [x] Optimization

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks

More context on this change can be found in #7901


LukaszRozmej (approved):

merklefruit:
Hi @MarekM25, can I help add context for a review? Would love to see this merged if possible

LukaszRozmej:
@kamilchodola It would be good to run `engine` and `pyspec` hive tests on this PR. But I cannot seem to run them on outside branches from actions.

MarekM25 (commented):
LGTM, but Hive tests will have to be updated to handle new error codes

merklefruit:
@MarekM25 I can do that, can you provide a quick pointer to where i should look?

flcl42 (approved):

MarekM25:
> Quote reply

I think you will have to search for either our old errors or geth errors in this repo and provide a PR: https://github.com/ethereum/hive here

I can contact you with maintainers if needed. Thx for your contribution! 🙏

smartprogrammer93 (approved):

merklefruit:
@MarekM25 hi, I've been able to run the `ethereum/engine` hive test suite for both this branch and nethermind's master.
Here are the results:

* http://135.181.191.125:8081/ here is the result on the latest nethermind master branch
* http://135.181.191.125:8082/ here is the result on my PR's branch

from this run, it seems like this specific branch is not causing any new errors, but as I previously mentioned I'm not familiar enough with Hive to be 100% sure that I didn't miss something. 

I have also tried to look for the hardcoded error messages in the hive codebase but haven't found them: the behaviour there seems to be to only check the "Status: Invalid" property, and not the actual validation error message

benaadams (approved):

MarekM25 (approved):

benaadams:
@merklefruit thank you for your contribution

PR #7946, opened 2024-12-20
Code fix Dockerfile

teenager-ETH:
# Fix: Surround variables with double quotes in Dockerfile

## Changes
- Updated the `dotnet publish` command to quote variables.

  - **Before:**
    ```dockerfile
    dotnet publish src/Nethermind/Nethermind.Runner -c $BUILD_CONFIG -a $arch -o /publish --sc false \
    ```

  - **After:**
    ```dockerfile
    dotnet publish src/Nethermind/Nethermind.Runner -c "$BUILD_CONFIG" -a "$arch" -o /publish --sc false \
    ```

## Purpose
- Prevents errors caused by spaces or special characters in variables.
- Follows shell scripting best practices.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ ARG TARGETARCH
 COPY src/Nethermind src/Nethermind
 
 RUN arch=$([ "$TARGETARCH" = "amd64" ] && echo "x64" || echo "$TARGETARCH") && \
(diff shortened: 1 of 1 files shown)


rubo:
No need for quotes in this particular case. Thank you for your attention.

PR #7953, opened 2024-12-21
Remove some closure allocations

benaadams:
## Changes

- Avoid closure
- Capture closure state in .ctor
- Use try/catch rather than .ContinueWith for async Task
- Change lambda sig to pass in value

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Network/P2P/Session.cs
+++ b/src/Nethermind/Nethermind.Network/P2P/Session.cs
@@ -366,7 +366,7 @@ namespace Nethermind.Network.P2P
             HandshakeComplete?.Invoke(this, EventArgs.Empty);
         }
 
-        public void InitiateDisconnect(DisconnectReason disconnectReason, string? details = null)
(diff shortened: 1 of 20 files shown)


src/Nethermind/Nethermind.Network/P2P/Session.cs
```diff
@@ -491,28 +491,32 @@ public void MarkDisconnected(DisconnectReason disconnectReason, DisconnectType d
             if (_context is null)
             {
                 //in case pipeline did not get to p2p - no disconnect delay
-                _channel.DisconnectAsync().ContinueWith(x =>
+                try
```
asdacap, line 494: Can split the async part and don't block on it? I'm concerned that this may have unintended consequence.

benaadams: Moved the async out

asdacap (approved):

PR #7958, opened 2024-12-22
Skip sstore gas metering read for system tx

benaadams:
## Changes

- Skip read in SSTORE for Beacon Root tx (happens very early in block and prewarming isn't quite fast enough)
- Move prewarming a bit earlier
- Add both Storage slots to Beacon Root access list (trie fill is still needed for the commit)

Before

image

After
image

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain/BeaconBlockRoot/BeaconBlockRootHandler.cs
+++ b/src/Nethermind/Nethermind.Blockchain/BeaconBlockRoot/BeaconBlockRootHandler.cs
@@ -20,6 +20,8 @@ public class BeaconBlockRootHandler(ITransactionProcessor processor, IWorldState
 
     public (Address? toAddress, AccessList? accessList) BeaconRootsAccessList(Block block, IReleaseSpec spec, bool includeStorageCells = true)
     {
+        const int HistoryBufferLength = 8191;
+
         BlockHeader? header = block.Header;
         bool canInsertBeaconRoot = spec.IsBeaconBlockRootAvailable
                                   && !header.IsGenesis
@@ -39,7 +41,13 @@ public class BeaconBlockRootHandler(ITransactionProcessor processor, IWorldState
 
         if (includeStorageCells)
         {
-            builder.AddStorage(block.Timestamp % 8191);
+            // https://eips.ethereum.org/EIPS/eip-4788
+            // Set the storage value at header.timestamp % HISTORY_BUFFER_LENGTH to be header.timestamp
+            UInt256 slotIndex = block.Timestamp % HistoryBufferLength;
+            builder.AddStorage(in slotIndex);
+            // Set the storage value at header.timestamp % HISTORY_BUFFER_LENGTH + HISTORY_BUFFER_LENGTH to be calldata[0:32]
+            slotIndex += HistoryBufferLength;
+            builder.AddStorage(in slotIndex);
         }
 
         return (eip4788ContractAddress, builder.Build());
@@ -65,7 +73,16 @@ public class BeaconBlockRootHandler(ITransactionProcessor processor, IWorldState
 
             transaction.Hash = transaction.CalculateHash();
 
-            processor.Execute(transaction, header, tracer);
+            // Skip reads in SSTORE for calculating gas cost
+            spec.IsSystemTransaction = true;
+            try
+            {
+                processor.Execute(transaction, header, tracer);
+            }
+            finally
+            {
+                spec.IsSystemTransaction = false;
+            }
         }
     }
 }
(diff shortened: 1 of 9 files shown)


benaadams:
Chain size failure https://github.com/NethermindEth/nethermind/pull/7959

src/Nethermind/Nethermind.Blockchain/BeaconBlockRoot/BeaconBlockRootHandler.cs
```diff
@@ -39,7 +41,14 @@ public class BeaconBlockRootHandler(ITransactionProcessor processor, IWorldState
...
         {
-            builder.AddStorage(block.Timestamp % 8191);
+            // https://eips.ethereum.org/EIPS/eip-4788
+            // Set the storage value at header.timestamp % HISTORY_BUFFER_LENGTH to be header.timestamp
+            ulong slotIndex = header.Timestamp % HistoryBufferLength;
+            UInt256 slot256 = slotIndex;
+            builder.AddStorage(in slot256);
+            // Set the storage value at header.timestamp % HISTORY_BUFFER_LENGTH + HISTORY_BUFFER_LENGTH to be calldata[0:32]
+            slot256 = slotIndex + HistoryBufferLength;
+            builder.AddStorage(in slot256);
```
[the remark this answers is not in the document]
benaadams: Spec doesn't mention Access Lists; no gas is charged 30M should be enough for x2 SSTOREs

Spec is actually more wild and says

> Clients may decide to omit an explicit EVM call and directly set the storage values. Note: While this is a valid optimization for Ethereum mainnet, it could be problematic on non-mainnet situations in case a different contract is used.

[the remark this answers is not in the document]
LukaszRozmej: @MarekM25 ?

src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.cs
```diff
@@ -107,44 +108,51 @@ the previous head state.*/
                 preWarmTask = null;
                 WaitForCacheClear();
                 Block suggestedBlock = suggestedBlocks[i];
-                if (blocksCount > 64 && i % 8 == 0)
-                {
-                    if (_logger.IsInfo) _logger.Info($"Processing part of a long blocks branch {i}/{blocksCount}. Block: {suggestedBlock}");
```
MarekM25, line 112: This log was helpful in determining that something went seriously wrong in the past :D

src/Nethermind/Nethermind.Core/Specs/IReleaseSpec.cs
```diff
@@ -419,5 +419,10 @@ public interface IReleaseSpec : IEip1559Spec, IReceiptSpec
         bool IsAuthorizationListEnabled => IsEip7702Enabled;
 
         public bool RequestsEnabled => ConsolidationRequestsEnabled || WithdrawalRequestsEnabled || DepositsEnabled;
+
+        /// <summary>
+        /// Skip read in SSTORE for calculating gas cost as not charged
+        /// </summary>
+        bool IsSystemTransaction { get; }
```
MarekM25, line 426: Looks weird, this is added to ReleaseSpec and the field seems to be referenced to Transaction object

src/Nethermind/Nethermind.Blockchain/BeaconBlockRoot/BeaconBlockRootHandler.cs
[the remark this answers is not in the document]
MarekM25: hmmm I think some chains use a different implementation of this contract, so not sure if we should add anything besides just calling sys call

LukaszRozmej (approved):

src/Nethermind/Nethermind.Core/Specs/IReleaseSpec.cs
LukaszRozmej: I agree this is a bit weird way of passing that information and is not universal

LukaszRozmej (commented):
Could we avoid `ReleaseSpec` changes?

src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.cs
benaadams: Was just moved down

benaadams:
> Could we avoid `ReleaseSpec` changes?

Is hard to get that info inside SStore otherwise though 🤔

LukaszRozmej:
> > Could we avoid `ReleaseSpec` changes?
> 
> Is hard to get that info inside SStore otherwise though 🤔

Maybe BlockExecutionContext?

benaadams:
Closing in favour of https://github.com/NethermindEth/nethermind/pull/8415

PR #7962, opened 2024-12-23
Feature/verifytrie admin rpc

asdacap:
- As requested. `admin_verifyTrie` RPC.
- Also, move blocking mechanism to TrieStore instead of processing queue.

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Tested manually. 
- Tested to cancel on shutdown.
- Tested to not run multiple verifytrie.


src/Nethermind/Nethermind.JsonRpc/Modules/Admin/AdminRpcModule.cs
```diff
@@ -139,4 +143,19 @@ public ResultWrapper<PruningStatus> admin_prune()
     {
         return ResultWrapper<PruningStatus>.Success(_pruningTrigger.Trigger());
     }
+
+    public ResultWrapper<string> admin_verifyTrie()
+    {
+        if (_blockTree.Head is null)
+        {
+            return ResultWrapper<string>.Fail("Head is null. Unable to know state root to verify.");
```
LukaszRozmej, line 151: Potentially can we pass the `BlockParameter` where we want to verify tries?
Also potentially should we add `admin_availableStateRoots`, where it would look-up where state root is available in some block range?

src/Nethermind/Nethermind.Synchronization/FastSync/BlockingVerifyTrie.cs
```diff
@@ -0,0 +1,69 @@
...
+            }
+            catch (OperationCanceledException)
+            {
+            }
+            catch (Exception e)
+            {
+                _logger.Error($"Error in verify trie", e);
+            }
+
+        }, TaskCreationOptions.LongRunning);
```
LukaszRozmej, line 65: I remember we were removing `LongRunning` because of some reason, @benaadams ?

LukaszRozmej (approved):

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/AdminRpcModule.cs
asdacap: Sure, why not.

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/AdminRpcModule.cs
```diff
@@ -139,19 +144,31 @@ public ResultWrapper<bool> admin_setSolc()
...
     {
         return ResultWrapper<PruningStatus>.Success(_pruningTrigger.Trigger());
     }
 
-    public ResultWrapper<string> admin_verifyTrie()
+    public ResultWrapper<string> admin_verifyTrie(BlockParameter block)
     {
-        if (_blockTree.Head is null)
+        SearchResult<BlockHeader> headerSearchResult = _blockTree.SearchForHeader(block);
+        if (headerSearchResult.Object is null)
```
LukaszRozmej: I think we need a check if state root is available?

asdacap: done

asdacap: done

PR #7975, opened 2024-12-26
Declarative RLP Encoding/Decoding

emlautarom1:
## Changes

- Introduce an alternative approach to RLP encoding and decoding, based on a declarative API with support for code generation through [Source Generators](https://devblogs.microsoft.com/dotnet/introducing-c-source-generators/)

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

The core library has 100% test coverage. Source generated code might not be fully covered.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks

When we started working on refactoring our `TxDecoder` one thing that came up was how unergonomic is to work with our current RLP API. We even have some comments on the code itself mentioning these difficulties, for example:

https://github.com/NethermindEth/nethermind/blob/b81070de7a905b52a6f605ebdd4f417fefe234c7/src/Nethermind/Nethermind.Serialization.Rlp/Eip2930/AccessListDecoder.cs#L17-L23

https://github.com/NethermindEth/nethermind/blob/b81070de7a905b52a6f605ebdd4f417fefe234c7/src/Nethermind/Nethermind.Serialization.Rlp/Eip2930/AccessListDecoder.cs#L76-L81

This PR introduces a new RLP API based on https://github.com/NethermindEth/nethermind/pull/7334#issuecomment-2329147512 with several improvements:

- Describe the structure of a record and get encoding and decoding for free. No code duplication required.
- Records can be described using other records. Supports conditional, exceptions, function calls, etc.
- Decoding and encoding are extensible through classes that can be defined anywhere, plus some extension methods.
- Minimal core library with 100% code coverage.
- Supports backtracking.
- All function calls are known ahead of time (no `virtual` or `override`). Interfaces are only used to enforce implementations.
- Despite the extensive usage of lambdas, no closures are required (all lambdas are `static`). You can still use them if you want to, but overloads are provided to avoid them.
- Automatically generate the required code through Source Generators.


LukaszRozmej:
Can you move some RLP decoders to this new approach for testing purposes?
Can you share their generated code for inspection?

LukaszRozmej (commented):
Would be great to move from string builders to some more readable format. Even giant string interpolation would be good here.

emlautarom1:
This is how the generated code looks like for the Rose Tree (formatted):

https://github.com/NethermindEth/nethermind/blob/8341a0e956b0c7995a2e50f3c1b8e4f9a5bf4310/src/Nethermind/Nethermind.Serialization.FluentRlp.Test/RlpDerivedTest.cs#L22-L23

```C#
// <auto-generated />
#nullable enable
using System;
using System.CodeDom.Compiler;
using Nethermind.Serialization.FluentRlp;
using Nethermind.Serialization.FluentRlp.Instances;

namespace Nethermind.Serialization.FluentRlp.Test;

[GeneratedCode("RlpSourceGenerator", "0.1")]
public abstract class TreeRlpConverter : IRlpConverter<Nethermind.Serialization.FluentRlp.Test.Tree>
{
    public static void Write(ref RlpWriter w, Nethermind.Serialization.FluentRlp.Test.Tree value)
    {
        w.WriteSequence(value, static (ref RlpWriter w, Nethermind.Serialization.FluentRlp.Test.Tree value) =>
        {
            w.Write(value.Value);
            w.Write(value.Children, static (ref RlpWriter w, Tree value) => { w.Write(value); });
        });
    }

    public static Nethermind.Serialization.FluentRlp.Test.Tree Read(ref RlpReader r)
    {
        return r.ReadSequence(static (scoped ref RlpReader r) =>
        {
            var Value = r.ReadString();
            var Children = r.ReadList(static (scoped ref RlpReader r) => { return r.ReadTree(); });
            return new Nethermind.Serialization.FluentRlp.Test.Tree(Value, Children);
        });
    }
}

[GeneratedCode("RlpSourceGenerator", "0.1")]
public static class TreeExt
{
    public static Nethermind.Serialization.FluentRlp.Test.Tree ReadTree(this ref RlpReader reader) => TreeRlpConverter.Read(ref reader);
    public static void Write(this ref RlpWriter writer, Nethermind.Serialization.FluentRlp.Test.Tree value) => TreeRlpConverter.Write(ref writer, value);
}
```

In Rider you can inspect the generated in the Solution explorer:

image

src/Nethermind/Nethermind.Serialization.FluentRlp/Rlp.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        return serialized;
+    }
+
+    public static T Read<T>(ReadOnlySpan<byte> source, RefRlpReaderFunc<T> func)
+        where T : allows ref struct
+    {
+        var reader = new RlpReader(source);
+        T result = func(ref reader);
+        // TODO: We might want to add an option to check for no trailing bytes.
```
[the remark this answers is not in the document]
emlautarom1: Added: now the `RLP` API will throw when there are trailing bytes.

emlautarom1:
Updated generators to use string interpolation. There are some places where we still use `StringBuilder` though.

LukaszRozmej:
> This is how the generated code looks like for the Rose Tree (formatted):

Ok but can we do something real-life for us? Like `BlockHeader` for example?

src/Nethermind/Nethermind.Serialization.FluentRlp/RlpWriter.cs
```diff
@@ -0,0 +1,203 @@
...
+public ref struct RlpWriter
+{
+    private const bool LengthMode = false;
+    private const bool ContentMode = true;
+
+    private bool _mode;
+
+    public int Length { get; private set; }
+
+    private byte[] _buffer;
```
LukaszRozmej, line 25: We often write into Netty arena-based buffers to avoid allocations, would be good to support that.

LukaszRozmej (commented):
What about `RlpBehaviors`?

src/Nethermind/Nethermind.Serialization.FluentRlp/Int32Primitive.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// <summary>
+    /// Reads a <see cref="int" /> from the beginning of a read-only span of bytes, as big endian.
+    /// </summary>
+    /// <param name="source">The read-only span to read.</param>
+    /// <returns>The big endian value.</returns>
+    /// <remarks>The span is padded with leading `0`s as needed.</remarks>
+    public static int Read(ReadOnlySpan<byte> source)
+    {
+        Span<byte> buffer = stackalloc byte[sizeof(Int32)];
+        source.CopyTo(buffer[^source.Length..]);
```
Scooletz, line 20: `Unsafe.ReadUnalaligned`? Why copy?

src/Nethermind/Nethermind.Serialization.FluentRlp/Rlp.cs
```diff
@@ -0,0 +1,33 @@
...
+        => Write(action, static (ref RlpWriter w, RefRlpWriterAction action) => action(ref w));
+
+    public static byte[] Write<TContext>(TContext ctx, RefRlpWriterAction<TContext> action)
+        where TContext : allows ref struct
+    {
+        var lengthWriter = RlpWriter.LengthWriter();
+        action(ref lengthWriter, ctx);
+        var serialized = new byte[lengthWriter.Length];
+        var contentWriter = RlpWriter.ContentWriter(serialized);
+        action(ref contentWriter, ctx);
```
Scooletz, line 20: When is it used? Do we write the data twice?

src/Nethermind/Nethermind.Serialization.FluentRlp/RlpWriter.cs
```diff
@@ -0,0 +1,203 @@
...
+
+    public static RlpWriter LengthWriter()
+    {
+        return new RlpWriter
+        {
+            _mode = LengthMode
+        };
+    }
+
+    public static RlpWriter ContentWriter(byte[] buffer)
```
Scooletz, line 36: Have you considered adding an option aligned with `Utf8JsonWriter` where there's a ctor that accepts `IBufferWriter<byte>`? This would mean that ufortunately dicontinued chunks should be supported, but could allow to provide a writer over anything. Maybe this could help to address Netty comment from @LukaszRozmej  

https://learn.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonwriter.-ctor?view=net-9.0#system-text-json-utf8jsonwriter-ctor(system-buffers-ibufferwriter((system-byte))-system-text-json-jsonwriteroptions)

Scooletz (commented):
What an interesting take on this topic 😍 A few remarks provided as comments. General:

1. Probably `[SkipLocalInit]` would be beneficial.
2. Benchmarking with potential ASM output
3. Converting some of the existing ones and performing mano-a-mano comparison.

src/Nethermind/Nethermind.Serialization.FluentRlp.Generator/RlpSourceGenerator.cs
```diff
@@ -0,0 +1,320 @@
...
+/// </summary>
+[Generator]
+public sealed class RlpSourceGenerator : IIncrementalGenerator
+{
+    private const string Version = "0.1";
+    private const string GeneratedCodeAttribute = $"""[GeneratedCode("{nameof(RlpSourceGenerator)}", "{Version}")]""";
+
+    public void Initialize(IncrementalGeneratorInitializationContext context)
+    {
+        var provider = context.SyntaxProvider.CreateSyntaxProvider(
```
Scooletz, line 46: Consider attribute based creation using [ForAttributeWithMetadataName](https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.syntaxvalueprovider.forattributewithmetadataname?view=roslyn-dotnet-4.9.0). It should be much more cheaper than scanning all the records and only then select on attribute basis.

src/Nethermind/Nethermind.Serialization.FluentRlp/Int32Primitive.cs
emlautarom1: This method was added due to `BinaryPrimitives.ReadInt32BigEndian` requiring exactly 4 byte to read a `Int32`, so we pad `source` with enough `0` so it can properly decode.

I've never used `Unsafe.ReadUnaligned`, not even `usafe` code. How would that look?

src/Nethermind/Nethermind.Serialization.FluentRlp/Rlp.cs
emlautarom1: When is what used? We "write" the data twice to first compute the length (`LengthWriter`), and then we actually write the bytes into a buffer (`ContentWriter`).

src/Nethermind/Nethermind.Serialization.FluentRlp/RlpWriter.cs
emlautarom1: Interesting. I picked `byte[]` as a safe default but there is no reason why other type could be used.

- `IBufferWriter<T>` is quite small, it's part of the std lib and it supports `Span`-based APIs.

- `NettyRlpStream` is based on `IByteBuffer`, the latter which is defined in `DotNetty.Buffers`. It supports the operations that we need but it's quite large.

Now, the issue is that `IByteBuffer` and `IBufferWriter` are unrelated so we would need to pick one (unless we start doing conversions).

LukaszRozmej: `IBufferWriter` is probably more useful, we can try making adapter to `IByteBuffer` if we want

src/Nethermind/Nethermind.Serialization.FluentRlp/Int32Primitive.cs
LukaszRozmej: Why not use `Bytes` class we already have?

emlautarom1:
@LukaszRozmej `RlpBehaviors` as currently defined are not supported by the API though you can get the same behavior by manually pass any "context" during read/write.

We recently changed the `Rlp` interface so trailing bytes throw by default (though you can avoid it by using the `RlpReader`/`RlpWriter` structs directly if needed).

emlautarom1:
@LukaszRozmej `RlpBehaviors` are not explicitly supported by the API but you can get the same behavior by manually passing any "context" when reading/writing.

Note that we recently changed the `Rlp` interface so trailing bytes throw by default. If you want more control over what happens before or after reading/writing you can use the `RlpReader` and `RlpWriter` APIs directly.

emlautarom1:
I've added a benchmark that encodes and decodes an `AccessList` as defined in:

https://github.com/NethermindEth/nethermind/blob/e0c4a599d8f0ee1c8e3ce13d1c8620d4e3255a23/src/Nethermind/Nethermind.Core/Eip2930/AccessList.cs#L15

Results on my machine are the following:

```
| Method  | Mean     | Error   | StdDev  | Ratio |
|-------- |---------:|--------:|--------:|------:|
| Current | 343.9 us | 1.43 us | 1.34 us |  1.00 |
| Fluent  | 834.9 us | 2.34 us | 2.19 us |  2.43 |
```

There is room for a possible optimization: some records like `Address` have a known fixed byte size which we can leverage to avoid having to copy bytes twice: once to figure out the length and the other to actually copy the bytes.

src/Nethermind/Nethermind.Serialization.FluentRlp/Int32Primitive.cs
emlautarom1: Didn't think about it, but seems like implementations are the same:

https://github.com/NethermindEth/nethermind/blob/c8748770fe88fecf3b0b36226296e5c0858da051/src/Nethermind/Nethermind.Core/Extensions/Bytes.cs#L439-L441

Seems a bit overkill to add a project reference for 3 LOCs.

emlautarom1:
Replacing `Marshal.SizeOf<T>()` with `sizeof(T)` and some unsafe annotations gives quite the boost at no cost:

```
| Method  | Mean     | Error   | StdDev  | Ratio | RatioSD |
|-------- |---------:|--------:|--------:|------:|--------:|
| Current | 359.8 us | 5.03 us | 4.70 us |  1.00 |    0.02 |
| Fluent  | 626.2 us | 2.90 us | 2.42 us |  1.74 |    0.02 |
```

src/Nethermind/Nethermind.Serialization.FluentRlp/RlpWriter.cs
```diff
@@ -0,0 +1,203 @@
...
+                break;
+        }
+    }
+
+    private unsafe void LengthWrite<T>(T value) where T : unmanaged, IBinaryInteger<T>
+    {
+        var size = sizeof(T);
+        Span<byte> bigEndian = stackalloc byte[size];
+        value.WriteBigEndian(bigEndian);
+        bigEndian = bigEndian.TrimStart((byte)0);
```
emlautarom1, line 63: `TrimStart` does not seem to be heavily optimized. There might be something better that we can use, specially considering that we're removing leading zeros.

emlautarom1:
After running some benchmarks I found that `UInt256` was getting boxed due to the usage of interface default bodies. Fixing that issue improves performance and drastically reduces memory allocation (added `[MemoryDiagnoser]` as requested by @Scooletz):

```
| Method  | Mean     | Error   | StdDev  | Ratio | Gen0     | Gen1     | Gen2     | Allocated  | Alloc Ratio |
|-------- |---------:|--------:|--------:|------:|---------:|---------:|---------:|-----------:|------------:|
| Current | 359.5 us | 0.98 us | 0.86 us |  1.00 | 166.5039 | 166.5039 | 166.5039 | 1033.87 KB |        1.00 |
| Fluent  | 530.1 us | 4.67 us | 4.14 us |  1.47 |  51.7578 |  51.7578 |  51.7578 |   617.5 KB |        0.60 |
```

src/Nethermind/Nethermind.Serialization.Rlp.Benchmark/Program.cs
```diff
@@ -0,0 +1,120 @@
...
+        _fluent = Benchmark.Fluent.BuildAccessList(new Random(42));
+    }
+
+    [Benchmark(Baseline = true)]
+    public Nethermind.Core.Eip2930.AccessList Current()
+    {
+        var decoder = Eip2930.AccessListDecoder.Instance;
+
+        var length = decoder.GetLength(_current, RlpBehaviors.None);
+        var stream = new RlpStream(length);
```
Scooletz, line 36: This line is responsible for all the allocations as it allocates a new array underneath https://github.com/NethermindEth/nethermind/blob/4e4455960a4d62ea3429e7bb143b45789ec7246f/src/Nethermind/Nethermind.Serialization.Rlp/RlpStream.cs#L44 Is this the case that we want to benchmark or rather it should be a reused `RlpStream` here?

emlautarom1: To be comparable with the `FluentRlp` approach we should allocate a new buffer. Since both will allocate the same buffer size then it should not matter.

Scooletz: I see. If they allocate the same buffer, what makes the current allocate over 400kb more then?

emlautarom1: Interestingly they're not allocating the same buffer size: the fluent approach uses a buffer of `170850` bytes while the current one uses `172845`. That does not account for the 400kb you mention though. Rider's profiler is not giving me anything useful so I'm kind of stuck now.

Maybe we should add other objects (ex. `LogEntry`, `BlockInfo`, etc.) to get more accurate benchmarks.

src/Nethermind/Nethermind.Serialization.FluentRlp/RlpWriter.cs
LukaszRozmej: Looks like .net doesn't use SIMD for it!
We could write Vector based way to find start index.
@benaadams

src/Nethermind/Nethermind.Serialization.FluentRlp/RlpWriter.cs
emlautarom1: Replaced `byte[]` buffer with `IBufferWriter<byte>`.

The return type of the `Rlp` interface is now `ReadOnlyMemory<byte>` instead of `byte[]` for more flexibility compared to returning `ReadOnlySpan<byte>`. Overloads for `Rlp.Read(ReadOnlyMemory<byte>)` were added.

src/Nethermind/Nethermind.Serialization.Rlp.Benchmark/Program.cs
LukaszRozmej: you can use NettyRlpStream that uses arena memory

src/Nethermind/Nethermind.Serialization.FluentRlp/Rlp.cs
Scooletz: This is costly. There might be a way to write it once if you do it in a way similar to Paprika. In Paprika, there's enough room left at the beginning to encode any length, then once the encoding is done, the length is written at the beginning and the position is adjusted. This would require a copy of the memory though as it would leave a gap at the beginning. This might be still cheaper if you copy the memory once more instead of going through the serialization two times.

Having this said, maybe adding some context instead of a writer could be useful. If there was a context you could borrow from it an array to write to, without getting into ArrayPool.Shared thing. In Paprika profiling this was costly and was eliminated by passing a context object that has a simple stack of arrays that you can rent from (single threaded, no need to deal with the shared thing).

https://github.com/NethermindEth/Paprika/blob/61c326a39e942ab9c550f6a6b9b5df59e67c0f0f/src/Paprika/Merkle/ComputeMerkleBehavior.cs#L750-L757

emlautarom1: This could work but I think it would be quite a redesign if it were to be adapted to all paths, not only here at the top level. Recursive calls in `WriteSequence` would need to be adjusted to first immediately write the bytes and then move them around to deal with writing the length first.

Could be worth exploring though.

src/Nethermind/Nethermind.Serialization.FluentRlp/Rlp.cs
```diff
@@ -37,3 +33,43 @@ public static T Read<T>(ReadOnlySpan<byte> source, RefRlpReaderFunc<T> func)
...
+    public T[] Buffer => _buffer;
+
+    /// <summary>
+    /// Creates an instance of an <see cref="FixedArrayBufferWriter{T}"/>, in which data can be written to,
+    /// with a fixed capacity specified.
+    /// </summary>
+    /// <param name="capacity">The capacity of the underlying buffer.</param>
+    public FixedArrayBufferWriter(int capacity)
+    {
+        _buffer = new T[capacity];
```
LukaszRozmej, line 55: ArrayPool?

emlautarom1: By default I think we should go with a plain array to match the `ArrayBufferWriter` behavior. At the end of the day, the `RLP` static class is like a "safe default" API.

If we want more control over the buffers we use we can write a custom `IBufferWriter` as you suggested earlier while using `RlpReader` and `RlpWriter` directly.

src/Nethermind/Nethermind.Serialization.FluentRlp/Rlp.cs
Scooletz: Moving once vs writing twice 👀 Clearly, I'm just asking as I don't have a clear answer. Also, in Paprika there are just a few cases where it's needed as majority of the written RLP is really short (beside branches, where it's applied).

stdevMac:
@emlautarom1 stale-PR check: parked on the 1.47x perf blocker, and #12011 changed the RLP ground under it. OK to close (the design notes stay linkable if FluentRlp gets revisited)? If there's no reply in the next few days, we'll close this as part of the stale-PR cleanup — it can always be reopened or resubmitted.

emlautarom1:
@stdevMac Feel free to close but do not delete the branch. This work is independent of any other RLP design.

LukaszRozmej:
Keeping this open as a draft rather than closing it — the design is worth preserving as a starting point even though it is not mergeable as-is.

**Status for anyone picking this up:**

- Parked on performance. The last numbers were 1.47x slower than the current encoder on the `AccessList` benchmark, after the `Marshal.SizeOf<T>()` -> `sizeof(T)` fix and the `UInt256` boxing fix. Allocations were better (0.60x), so the trade is CPU for GC pressure — not obviously the right trade for a decoder on the hot path.
- #12011 changed the RLP layer underneath this, so the comparison baseline has moved and the benchmark would need re-running before any of the above is trusted.
- The core library plus source generator is self-contained and independent of whatever direction the RLP API takes, so it stays useful as a reference for the declarative approach and for the generated-code shape.

Not merging without the perf gap closing. Leaving it in draft so the design and the discussion stay findable; the `feature/declarative-rlp` branch is not to be deleted.

PR #7976, opened 2024-12-28
Refactor/isolate worldstate

asdacap:
- Move `ITrieStore` and `IWorldState` outside of `INethermindApi`. These should not be used directly as it will make swapping state layout very hard.
- Move healing related code and `ISnapServer` from `Nethermind.Synchronization` to `Nethermind.State`, exposing them from `IWorldStateManager`. These uses `ITrieStore` so, to hide `ITrieStore`, it need to be in `IWorldStateManager`. Also, paprika probably need to have a completely different snap server implementation.
- Move all pruning trie store state related initialization code into `PruningTrieStateFactory`. This should make it easier to be used in tests and make it clear, which part of the initialization is for the pruning trie store and hopefully prevent future exposure of trie store specific interface.

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

#### Notes on testing

- Testing....


Scooletz (commented):
No input from my end. `INodeStorage` is still in the api, which in my understanding should be a subject of a further refactor when another storage (like Paprika) is used. Appreciate the removal of the `TrieStore` though.

asdacap:
Yea, it still there. I was thinking of just something like `IWorldStateManager.SyncFromNetwork(INetwork network)` or something. But that is still far.

benaadams:
@asdacap formatting

LukaszRozmej (approved):

tanishqjasoria (approved):

PR #7980, opened 2024-12-29
Change for 7702 in devnet 5 

ak88:
Chain id is now a u256 in the tuple.
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7702.md


LukaszRozmej (changes requested):
change `DecodeULong`

ak88:
> change `DecodeULong`

fixed

src/Nethermind/Nethermind.Core.Test/Encoding/AuthorizationTupleDecoderTests.cs
```diff
@@ -103,7 +103,7 @@ public static IEnumerable<RlpStream> WrongSizeFieldsEncodedCases()
 
         yield return TupleRlpStream(
             //Wrong R size
-            Enumerable.Range(0, 8).Select(static i => (byte)0xFF).ToArray(),
+            Enumerable.Range(0, 32).Select(static i => (byte)0xFF).ToArray(),
```
LukaszRozmej, line 106: shouldn't those be some constants?

LukaszRozmej (approved):

PR #7983, opened 2024-12-30
Reduce parallel prewarming overwork

benaadams:
## Changes

To prewarm faster than main thread

- `ExecutionOptions.Warmup` => `ExecutionOptions.SkipValidation`
- `ExecutionOptions.NoValidation` => `ExecutionOptions.SkipValidationAndCommit` (which is  `SkipValidation` | `Commit`)
- Then skip validations based on the `SkipValidation` flag rather than the previous `NoValidation` flag which also needed the tx to `Commit` and wasn't the flag prewarming was using
- Use actual tx don't clone them
- Don't clone block header

Fastest code is the code that doesn't run

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
+++ b/src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
@@ -69,7 +69,7 @@ namespace Nethermind.Evm.TransactionProcessing
             /// <summary>
             /// Skip potential fail checks and commit state after execution
             /// </summary>
-            System = Commit | SkipValidation,
+            SkipValidationAndCommit = Commit | SkipValidation,
 
             /// <summary>
             /// Commit and later restore state also skip validation, use for CallAndRestore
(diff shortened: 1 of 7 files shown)


src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -696,7 +696,7 @@ protected virtual void ExecuteEvmCall(
             WorldState.Restore(snapshot);
 
         Complete:
-            if (!opts.HasFlag(ExecutionOptions.NoValidation))
+            if (!opts.HasFlag(ExecutionOptions.SkipValidation))
```
LukaszRozmej, line 699: I was worried that we might modify the header, but we probably are not modifying it?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
benaadams: Not inside a single tx run as we aren't doing block processing

PR #7985, opened 2024-12-31
RBuilder rpc

asdacap:
## Changes
This PR adds support for NMC as _remote state provider_ for [rbuilder](https://github.com/flashbots/rbuilder) 

## Types of changes
We have added new JSON RPC namespace `rbuilder_` with following endpoints:
1. rbuilder_getCodeByHash
2. rbuilder_calculateStateRoot
3. rbuilder_getAccount
4. rbuilder_getBlockHash

The last 2 could've been handled with `eth_getProof` and `eth_getBlockByNumber`, but the reason we have added new, rbuilder-specific calls, is performance. We want to send as little data as possible (e.g. just block hash instead of the whole block (header)) as well as to execute this as quickly as possible.  This is because there are _a lot_ of these calls and we want  to maintain the the lowest latency possible (in range of couple hundred micro_s).

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

This was extensively tested on mainnet node & rbuilder for weeks, it works :)


mralj:
@asdacap I've just realised that we'll probably have to update the RPC call for the `rbuilder_getCodeByHash` to accept the `BlockParameter` param. 

When constructing `rbuilder` _provider_, the provider can be specified per block (so basically we are dealing with the state at some block in history). So this scenario should be possible:
1. EOA deploys contract at some address in block no. 100
2. On `rbuilder` we create provider for block  no. 99
3. Since we don't provide the `BlockParameter` to the RPC call, the call will refer to the latest state where contract exists and `rbuilder_getCodeByHash` will return code
4. Expected behaviour was to return _not found_

Am I correct?

asdacap:
Hmm... that is true. In that case you'll need to change rbuilder to use `eth_getCode` as internally, there is no way of knowing which block does a code get inserted.

asdacap:
The address would know the right code hash at which block, so it would have the history.

mralj:
Sorry I'm not sure I 100% understood you.  So to double-check:
If you receive both `code_hash` AND `BlockParameter` from `rbuilder` - this is still not enough, you'll need the `address` as well? 
Or I misunderstood you and having `code_hash` and `BlockParameter` is enough for this call to work properly?

asdacap:
There is no way to know when a `code_hash` is set. In fact, a lot of the time many account share the same code hash, so who knows which contract set it first. However, the account contains the code hash. It is possible to know the state of the account at a particular block, so if the `code_hash` of the account change, we could know it.

mralj:
> There is no way to know when a `code_hash` is set. In fact, a lot of the time many account share the same code hash, so who knows which contract set it first. However, the account contains the code hash. It is possible to know the state of the account at a particular block, so if the `code_hash` of the account change, we could know it.

Ok, let's then leave this as is, and if it provides problematic in practice, we'll tackle it then.

src/Nethermind/Nethermind.JsonRpc/Modules/Rbuilder/RbuilderRpcModule.cs
```diff
@@ -0,0 +1,162 @@
...
+    {
+        return ResultWrapper<byte[]>.Success(worldStateManager.GlobalStateReader.GetCode(hash));
+    }
+
+    public ResultWrapper<Hash256> rbuilder_calculateStateRoot(BlockParameter blockParam, IDictionary<Address, AccountChange> accountDiff)
+    {
+        BlockHeader? blockHeader = blockFinder.FindHeader(blockParam);
+        if (blockHeader is null)
+        {
+            return ResultWrapper<Hash256>.Fail("Block not found");
```
LukaszRozmej, line 32: error code?

```suggestion
            return ResultWrapper<Hash256>.Fail("Block not found", ErrorCodes.ResourceNotFound);
```

src/Nethermind/Nethermind.JsonRpc/Modules/Rbuilder/RbuilderRpcModule.cs
```diff
@@ -0,0 +1,162 @@
...
+                        {
+                        }
+                        else
+                        {
+                            worldState.SubtractFromBalance(address, originalBalance - accountChange.Balance.Value, releaseSpec);
+                        }
+
+                        if (worldState.GetBalance(address) != accountChange.Balance.Value)
+                        {
+                            throw new Exception("Balance is not same!? Why?");
```
LukaszRozmej, line 84: `ResultWrapper<Hash256>.Fail` rather than throwing base `Exception`?

src/Nethermind/Nethermind.JsonRpc/Modules/Rbuilder/RbuilderRpcModule.cs
```diff
@@ -0,0 +1,162 @@
...
+                    worldState.InsertCode(address, accountChange.CodeHash, Array.Empty<byte>(), releaseSpec, false);
+                }
+
+                if (accountChange.ChangedSlots is not null)
+                {
+                    foreach (KeyValuePair<UInt256, UInt256> changedSlot in accountChange.ChangedSlots)
+                    {
+                        ReadOnlySpan<byte> bytes = changedSlot.Value.ToBigEndian();
+                        bool newIsZero = bytes.IsZero();
+                        bytes = !newIsZero ? bytes.WithoutLeadingZeros() : [0];
```
LukaszRozmej, line 105: introduce static for `[0]` - there is one in `Bytes` but it is `ReadOnlyMemory`

src/Nethermind/Nethermind.JsonRpc/Modules/Rbuilder/RbuilderRpcModule.cs
```diff
@@ -0,0 +1,162 @@
...
+                {
+                    // Note, this set CodeDb, but since this is a read only world state, it should do nothing.
+                    worldState.InsertCode(address, accountChange.CodeHash, Array.Empty<byte>(), releaseSpec, false);
+                }
+
+                if (accountChange.ChangedSlots is not null)
+                {
+                    foreach (KeyValuePair<UInt256, UInt256> changedSlot in accountChange.ChangedSlots)
+                    {
+                        ReadOnlySpan<byte> bytes = changedSlot.Value.ToBigEndian();
```
LukaszRozmej, line 103:
```suggestion
                        byte[] bytes = changedSlot.Value.ToBigEndian();
```

src/Nethermind/Nethermind.JsonRpc/Modules/Rbuilder/RbuilderRpcModule.cs
```diff
@@ -0,0 +1,162 @@
...
+                        bool newIsZero = bytes.IsZero();
+                        bytes = !newIsZero ? bytes.WithoutLeadingZeros() : [0];
+                        worldState.Set(new StorageCell(address, changedSlot.Key), bytes.ToArray());
```
LukaszRozmej, line 106:
```suggestion
                        worldState.Set(new StorageCell(address, changedSlot.Key), bytes);
```

src/Nethermind/Nethermind.JsonRpc/Modules/Rbuilder/RbuilderRpcModule.cs
```diff
@@ -0,0 +1,162 @@
...
+        }
+    }
+
+
+    public ResultWrapper<AccountState> rbuilder_getAccount(Address address, BlockParameter block)
+    {
+        BlockHeader? blockHeader = blockFinder.FindHeader(block);
+        if (blockHeader is null)
+        {
+            return ResultWrapper<AccountState>.Fail("Block not found");
```
LukaszRozmej, line 127:
```suggestion
            return ResultWrapper<Hash256>.Fail("Block not found", ErrorCodes.ResourceNotFound);
```

src/Nethermind/Nethermind.JsonRpc/Modules/Rbuilder/RbuilderRpcModule.cs
```diff
@@ -0,0 +1,162 @@
...
+    {
+        public IOverridableWorldScope Create()
+        {
+            return worldStateManager.CreateOverridableWorldScope();
+        }
+
+        public bool Return(IOverridableWorldScope obj)
+        {
+            obj.WorldState.Reset();
+            obj.WorldState.ResetOverrides();
```
LukaszRozmej, line 158: Should one invoke the other? Shouldn't there be some unification?

src/Nethermind/Nethermind.JsonRpc/Modules/Rbuilder/RbuilderRpcModule.cs
LukaszRozmej: Maybe there should be some helper methods in `ResultWrapper` for most used error types?

LukaszRozmej (commented):
Shouldn't it be in its own plugin?

src/Nethermind/Nethermind.JsonRpc/Modules/Rbuilder/RbuilderRpcModule.cs
mralj: This should never happen, so I agree with @asdacap having exception thrown here, that is if we don't remove this code.
I think @asdacap that we should delete this altogether?

src/Nethermind/Nethermind.JsonRpc/Modules/Rbuilder/RbuilderRpcModule.cs
mralj: @LukaszRozmej the `WithoutLeadingZeros` excepts `ReadOnlySpan<bytes>, and we  can refactor these 4  lines to simply:
```
ReadOnlySpan<byte> bytes = changedSlot.Value.ToBigEndian().WithoutLeadingZeros();
worldState.Set(new StorageCell(address, changedSlot.Key), bytes.ToArray());
```

Wdyt? 

Or should I add method `WithoutLeadingZeros` which accepts & returns byte[]  I'm not sure how much perf are we loosing with conversion `byte[] -> ReadOnlySpan -> byte[]`

src/Nethermind/Nethermind.JsonRpc/Modules/Rbuilder/RbuilderRpcModule.cs
asdacap: This is more like a debug actually.

src/Nethermind/Nethermind.JsonRpc/Modules/Rbuilder/RbuilderRpcModule.cs
asdacap: @tanishqjasoria is working on a more defined worldstate scope for verkle. I'm thinking of unifying this together with that.

src/Nethermind/Nethermind.JsonRpc/Modules/Rbuilder/RbuilderRpcModule.cs
asdacap: Removed

mralj:
> Shouldn't it be in its own plugin?

I thik this is better answered by @asdacap as he chose this path :)

LukaszRozmej:
> > Shouldn't it be in its own plugin?
> 
> @LukaszRozmej I think this is better answered by @asdacap as he chose this path :)

Maybe should go under flashbots: https://github.com/NethermindEth/nethermind/tree/master/src/Nethermind/Nethermind.Flashbots ?

mralj:
> > > Shouldn't it be in its own plugin?
> > 
> > 
> > @LukaszRozmej I think this is better answered by @asdacap as he chose this path :)
> 
> Maybe should go under flashbots: https://github.com/NethermindEth/nethermind/tree/master/src/Nethermind/Nethermind.Flashbots ?

I'm OK with this, @asdacap can you please do it, I have some more pressing tasks, and you will be much faster than me with this :)

asdacap:
Moved to flashbots plugin

mralj:
@LukaszRozmej tested after @asdacap moved this to plugin and merged with latest master, it's ready for re-reveiw :)

LukaszRozmej (approved):

PR #7989, opened 2025-01-01
Fix links in CONTRIBUTING.md

bocalhky:
hi sers, found that the two Discord links in CONTRIBUTING.md we're invalid, replaced them with [the one from the README](https://github.com/NethermindEth/nethermind/blob/master/README.md?plain=1#L12)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,11 +8,11 @@ We expect project participants to adhere to our [code of conduct](./CODE_OF_COND
 
 ## Asking questions and providing feedback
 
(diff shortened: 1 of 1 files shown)


rubo:
The changes don't adhere to our [contributing guidelines](https://github.com/NethermindEth/nethermind/blob/master/CONTRIBUTING.md#dos-and-donts). Thank you for your time and attention.

bocalhky:
don't think this is a simple grammar fix but sure, would suggest you change it though as who knows what discord someone might end up in while looking for it

PR #7890, opened 2024-12-11
Bls precompiles: update gas costs

Marchhill:
Note: CI not passing since tests are not updated with EIP-2537 changes

## Changes

- Update bls gas costs to match [latest spec](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2537.md)

#### What types of changes does your code introduce?

- [ ] Bugfix (a non-breaking change that fixes an issue)
- [x] New feature (a non-breaking change that adds functionality)
- [ ] Breaking change (a change that causes existing functionality not to work as expected)
- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

#### If yes, did you write tests?

- [ ] Yes
- [x] No

#### Requires documentation update

- [ ] Yes
- [x] No

#### Requires explanation in Release Notes

- [ ] Yes
- [x] No

PR #7991, opened 2025-01-02
Updated eip 2537  gaspricing

ak88:
Updated gas pricings and discounts for BLS precompile:
https://github.com/ethereum/execution-specs/pull/1059


Marchhill (commented):
Already implemented in https://github.com/NethermindEth/nethermind/pull/7890 and https://github.com/NethermindEth/nethermind/pull/7932

PR #7998, opened 2025-01-03
Faster Keccak Xor

benaadams:
## Changes

- As 25 longs in state, 9 more long sizes to process after the vector sizes, so process as longs rather than bytes

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Core/Crypto/KeccakHash.cs
+++ b/src/Nethermind/Nethermind.Core/Crypto/KeccakHash.cs
@@ -435,6 +435,8 @@ namespace Nethermind.Core.Crypto
                     state256 = Vector512.Xor(state256, input256);
                 }
 
+                if (input.Length == vectorLength) return;
+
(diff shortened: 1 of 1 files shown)


Scooletz (commented):
> As 25 longs in state, 9 more to process after the vector sizes

I don't understand the comment. Could you rephrase it?

benaadams:
> > As 25 longs in state, 9 more to process after the vector sizes
> 
> I don't understand the comment. Could you rephrase it?

The size that is being Xor'd is 200 bytes, which is 200/8 = 25 ulongs

The vector portion at smallest size 128/8 = 16bytes will complete 192 bytes, leaving 200 - 192 = 8, 1 long to process

So process as 1 long rather than 8 bytes

benaadams:
> LGTM, assuming there are already existing tests covering this. Otherwise, please add them

Yes, everything would break if it was wrong 😅

PR #8000, opened 2025-01-03
Add SnapSync=true to EnergyWeb and Volta

kamilchodola:
Adding SnapSync = true to Energyweb as bootnodes are in process of resyncing to HalfPath with SnapServer
Same for Volta.

To be merged only when we will have final confirmation from EnergyWeb team that migration is completed and we will prove that everything works smoothly.

- [X] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

- [ ] Yes
- [X] No

- [X] Yes
- [ ] No

Snap Sync enabled in EnergyWeb chains (EnergyWeb and Volta). This will drastically reduce the sync time of those chains (from 10h+ to minutes of sync).


benaadams:
Fails check expecting it to be false

```
  Failed Snap_sync_settings_as_expected("energyweb.json",False) [114 ms]
  Error Message:
   Expected object to be False because energyweb.json: ISyncConfig.SnapSync, but found True.
```

benaadams (approved):

kamilchodola:
@benaadams this test doesn't makes sense imo :D

benaadams:
> @benaadams this test doesn't makes sense imo :D

Just need to change it to expect true?

kamilchodola:
> > @benaadams this test doesn't makes sense imo :D
> 
> Just need to change it to expect true?

Yes yes - just not sure if this test helps in anything :) Adjusted.

benaadams (approved):

LukaszRozmej (approved):

MarekM25 (approved):

PR #8001, opened 2025-01-03
Move prewarming as early as possible

benaadams:
## Changes

- Move prewarming as early as possible in block processing (every little helps)

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.cs
@@ -86,6 +86,12 @@ public partial class BlockProcessor(
     {
         if (suggestedBlocks.Count == 0) return [];
 
+        /* We need to save the snapshot state root before reorganization in case the new branch has invalid blocks.
+           In case of invalid blocks on the new branch we will discard the entire branch and come back to
+           the previous head state.*/
+        Hash256 previousBranchStateRoot = CreateCheckpoint();
+        InitBranch(newBranchStateRoot);
+
         Block suggestedBlock = suggestedBlocks[0];
         CancellationTokenSource? cancellationTokenSource = null;
         Task? preWarmTask = null;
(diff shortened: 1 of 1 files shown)


LukaszRozmej:
Is it worth it though? How much do we gain on normal flow?

benaadams:
> Is it worth it though? How much do we gain on normal flow?

Means it gets the prewarmer threads queued to ThreadPool before `TxHashCalculator` and all the callback events.

Every millisecond counts for the early prewarms (beacon root and first txs)

LukaszRozmej:
ok does it register on our grafana graphs?

benaadams:
> ok does it register on our grafana graphs?

Yes though can't find it atm (am doing a full sync)

We want the prewarming to front run the txs.

Block takes 10ms running 300 tx, so roughly 33 microseconds per tx. At that speed any and all edge is significant; as to whether the tx is prewarmed or not.

This I think is the maximum it can go ahead (I moved it earlier but was unhappy that some of the block wasn't set in state, so missing trienode errors)

benaadams:
Moved `TxHashCalculator.CalculateInBackground` to last item as that's not needed until forkchoice, so less over subscription on cpu threads

src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.cs
```diff
@@ -85,28 +86,45 @@ public Block[] Process(Hash256 newBranchStateRoot, List<Block> suggestedBlocks,
...
-                Block suggestedBlock = suggestedBlocks[i];
+                suggestedBlock = suggestedBlocks[i];
+
+                skipPrewarming = preWarmer is null || suggestedBlock.Transactions.Length < 3;
+                // If cancelationTokenSource is not null it means we are in first iteration of loop
+                // and started prewarming at method entry, so don't start it again
+                if (cancellationTokenSource is null && !skipPrewarming)
+                {
+                    (cancellationTokenSource, preWarmTask) = PreWarmTransactions(suggestedBlock, preBlockStateRoot);
+                }
```
LukaszRozmej, line 126: This code is very similar to the one above, can we have a helper method?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.cs
benaadams: Moved into PreWarm method

PR #8007, opened 2025-01-05
Add Triekey to reduce allocations

benaadams:
## Changes

- Add Triekey which can represent a key from two arrays rather than allocating a new one using concatenation
- Horrible

- [x] Optimization

#### Requires testing

- [x] Yes

- [x] Yes

### DIFF
--- a/src/Nethermind/Nethermind.Trie/Pruning/TreePath.cs
+++ b/src/Nethermind/Nethermind.Trie/Pruning/TreePath.cs
@@ -88,7 +88,17 @@ public struct TreePath : IEquatable<TreePath>
     public readonly TreePath Append(ReadOnlySpan<byte> nibbles)
     {
         if (nibbles.Length == 0) return this;
-        if (nibbles.Length == 1) return Append(nibbles[0]);
+        if (nibbles.Length == 1) return Append((int)nibbles[0]);
+
+        TreePath copy = this;
+        copy.AppendMut(nibbles);
+        return copy;
+    }
+
+    public readonly TreePath Append(TrieNodeKey nibbles)
+    {
+        if ((nibbles?.Length ?? 0) == 0) return this;
+        if (nibbles.Length == 1) return Append((int)nibbles[0]);
 
         TreePath copy = this;
         copy.AppendMut(nibbles);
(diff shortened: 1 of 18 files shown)


src/Nethermind/Nethermind.Trie/TrieKey.cs
```diff
@@ -0,0 +1,529 @@
...
+
+    /// <summary>
+    /// An empty TrieKey instance (no bytes).
+    /// </summary>
+    public static TrieKey Empty { get; } = new(Array.Empty<byte>());
+
+    // These two byte arrays together represent the key. 
+    // If _keyPart1 is null, the entire key is in _keyPart0.
+    private readonly byte[]? _keyPart0;
+    private readonly byte[]? _keyPart1;
```
LukaszRozmej, line 29: Before I dive deeper, maybe we can use something like ReadOnlySequence<byte>?

Scooletz: `ReadOnlySequence` would be much more fat. 2 objects + 2 ints.

benaadams: Also allocates if you read across the objects as a Span

benaadams: Also allocates if you read across the objects as a Span; works better for much larger sizes of data

src/Nethermind/Nethermind.Trie/Pruning/TreePath.cs
```diff
@@ -135,6 +144,40 @@ internal void AppendMut(ReadOnlySpan<byte> nibbles)
...
+            offset = 1;
+        }
+
+        int byteLength = (nibbles.Length - offset) / 2;
+        int pathSpanStart = Length / 2;
+        for (int i = 0; i < byteLength; i++)
+        {
+            pathSpan[i + pathSpanStart] = Nibbles.ToByte(nibbles[offset + i * 2], nibbles[offset + i * 2 + 1]);
+            Length += 2;
+        }
```
LukaszRozmej, line 172: SIMD?

benaadams:
Too many merge conficts

PR #8008, opened 2025-01-05
Back signature by Vector512 rather than byte[]

benaadams:
## Changes

- Drops one object reference per sig and `Vector512` is smaller than `byte[]` with one less indirection

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Core/Crypto/Signature.cs
+++ b/src/Nethermind/Nethermind.Core/Crypto/Signature.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Runtime.InteropServices;
+using System.Runtime.Intrinsics;
 using Nethermind.Core.Attributes;
 using Nethermind.Core.Extensions;
 using Nethermind.Int256;
@@ -12,12 +13,13 @@ namespace Nethermind.Core.Crypto
     public class Signature : IEquatable<Signature>
     {
         public const int VOffset = 27;
+        private Vector512<byte> _signature;
 
         public Signature(ReadOnlySpan<byte> bytes, int recoveryId)
         {
             ArgumentOutOfRangeException.ThrowIfNotEqual(bytes.Length, 64);
 
-            bytes.CopyTo(Bytes.AsSpan());
+            bytes.CopyTo(Bytes);
             V = (ulong)recoveryId + VOffset;
         }
 
@@ -25,7 +27,7 @@ namespace Nethermind.Core.Crypto
         {
             ArgumentOutOfRangeException.ThrowIfNotEqual(bytes.Length, 65);
 
-            bytes[..64].CopyTo(Bytes.AsSpan());
+            bytes[..64].CopyTo(Bytes);
             V = bytes[64];
         }
 
@@ -33,8 +35,9 @@ namespace Nethermind.Core.Crypto
         {
             ArgumentOutOfRangeException.ThrowIfLessThan(v, (ulong)VOffset);
 
-            r.CopyTo(Bytes.AsSpan(32 - r.Length, r.Length));
-            s.CopyTo(Bytes.AsSpan(64 - s.Length, s.Length));
+            Span<byte> span = Bytes;
+            r.CopyTo(span.Slice(32 - r.Length, r.Length));
+            s.CopyTo(span.Slice(64 - s.Length, s.Length));
             V = v;
         }
 
@@ -42,8 +45,9 @@ namespace Nethermind.Core.Crypto
         {
             ArgumentOutOfRangeException.ThrowIfLessThan(v, (ulong)VOffset);
 
-            r.ToBigEndian(Bytes.AsSpan(0, 32));
-            s.ToBigEndian(Bytes.AsSpan(32, 32));
+            Span<byte> span = Bytes;
+            r.ToBigEndian(span.Slice(0, 32));
+            s.ToBigEndian(span.Slice(32, 32));
 
             V = v;
         }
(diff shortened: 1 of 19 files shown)


src/Nethermind/Nethermind.JsonRpc.Test/Modules/Eth/EthRpcSimulateTestsBase.cs
```diff
@@ -127,7 +127,7 @@ protected static byte[] GenerateTransactionDataForEcRecover(Hash256 keccak, Sign
     {
         AbiDefinition call = new AbiDefinitionParser().Parse(GetEcRecoverContractJsonAbi(name));
         AbiEncodingInfo functionInfo = call.GetFunction(name).GetCallInfo();
-        return AbiEncoder.Instance.Encode(functionInfo.EncodingStyle, functionInfo.Signature, keccak, signature.V, signature.R, signature.S);
+        return AbiEncoder.Instance.Encode(functionInfo.EncodingStyle, functionInfo.Signature, keccak, signature.V, signature.R.ToArray(), signature.S.ToArray());
```
LukaszRozmej, line 130: We should be able to encode spans?

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
```diff
@@ -285,7 +285,7 @@ public ResultWrapper<byte[]> eth_sign(Address addressData, byte[] message)
         }
 
         if (_logger.IsTrace) _logger.Trace($"eth_sign request {addressData}, {message}, result: {sig}");
-        return ResultWrapper<byte[]>.Success(sig.Bytes);
+        return ResultWrapper<byte[]>.Success(sig.Bytes.ToArray());
```
LukaszRozmej, line 288: Can we either get Memory<byte> to the signature, or just simpler add JSON converter for whole signature object?

src/Nethermind/Nethermind.JsonRpc/Modules/Parity/ParityTransaction.cs
```diff
@@ -60,8 +60,8 @@ public ParityTransaction(Transaction transaction, byte[] raw, PublicKey publicKe
             Input = transaction.Data.AsArray();
             PublicKey = publicKey;
             ChainId = transaction.Signature.ChainId;
-            R = transaction.Signature.R;
-            S = transaction.Signature.S;
+            R = transaction.Signature.R.ToArray();
+            S = transaction.Signature.S.ToArray();
```
LukaszRozmej, line 64: We could back it up by signature object and make R and S spans here?

src/Nethermind/Nethermind.JsonRpc/Modules/Personal/PersonalRpcModule.cs
```diff
@@ -96,7 +96,7 @@ public ResultWrapper<byte[]> personal_sign(byte[] message, Address address, stri
             }
 
             message = ToEthSignedMessage(message);
-            return ResultWrapper<byte[]>.Success(_wallet.Sign(Keccak.Compute(message), address).Bytes);
+            return ResultWrapper<byte[]>.Success(_wallet.Sign(Keccak.Compute(message), address).Bytes.ToArray());
```
LukaszRozmej, line 99: Signature JSON converter

src/Nethermind/Nethermind.JsonRpc.Test/Modules/Eth/EthRpcSimulateTestsBase.cs
benaadams: Is test, didn't seem important

src/Nethermind/Nethermind.JsonRpc/Modules/Parity/ParityTransaction.cs
benaadams: Switched to Memory

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
benaadams: Switched to Memory

src/Nethermind/Nethermind.JsonRpc/Modules/Personal/PersonalRpcModule.cs
benaadams: Switched to Memory

src/Nethermind/Nethermind.Core/Crypto/Signature.cs
```diff
@@ -52,26 +57,27 @@ public Signature(string hexString)
             : this(Core.Extensions.Bytes.FromHexString(hexString))
         {
         }
+        public Span<byte> Bytes => MemoryMarshal.AsBytes(MemoryMarshal.CreateSpan(ref _signature, 1));
+        public override Memory<byte> Memory => CreateMemory(64);
```
LukaszRozmej, line 61: This doesn't allocate but return parts of binary representation of the class as `Memory<byte>`?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Core/Crypto/Signature.cs
benaadams: Yeah, class uses itself as the backing by inheriting from `MemoryManager<byte>`; then Memory when Span is called invokes the Span method. 

Is a bit weird and like a sledgehammer for a nut, but skips all the allocations 🙂

LukaszRozmej: Shouldn't we store it somewhere or is it better to create it on every call?

benaadams: Easy enough to create it; otherwise it will expand `Signature` itself by 16 bytes

PR #8009, opened 2025-01-05
Back Bloom with inline array rather than byte[]

benaadams:
## Changes

- Drops one object reference per Bloom and inline array is smaller than `byte[]` with one less indirection

- [x] Optimization

#### Requires testing

- [x] No


LukaszRozmej (approved):

rubo:
Maybe not in this specific case with `InlineArray`, but good to [know](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.inlinearrayattribute?view=net-9.0#remarks):

> .NET 9 and later, the default implementations of `Equals()` and `GetHashCode()` for types marked with this attribute throw `NotSupportedException`. You must override both `Equals(Object)` and `GetHashCode` if they will be used.

PR #8022, opened 2025-01-07
New filter restricting delegations in txpool

ak88:
This is a new filter for tx pool, introducing limitations to delegated accounts.
Originally mitigations are described in the EIP itself:
https://eips.ethereum.org/EIPS/eip-7702#transaction-propagation

As the EIP suggests this will limit delegated accounts to only one transaction in the pool at a time. 

Also will restrict any transaction with the same nonce and source as a pending authorization tuple in the pool.


src/Nethermind/Nethermind.TxPool/Filters/OnlyOneTxPerDelegatedAccountFilter.cs
```diff
@@ -0,0 +1,49 @@
...
+            if (pendingDelegations.HasPending(tx.SenderAddress!, tx.Nonce))
+            {
+                return AcceptTxResult.PendingDelegation;
+            }
+
+            if (!codeInfoRepository.TryGetDelegation(worldState, tx.SenderAddress!, out _))
+                return AcceptTxResult.Accepted;
+            Transaction[] currentTxs;
+
+            if (standardPool.TryGetBucket(tx.SenderAddress!, out currentTxs) || blobPool.TryGetBucket(tx.SenderAddress!, out currentTxs))
```
LukaszRozmej, line 34: `TryGetBucket` locks and then copies all the transactions. Maybe we can have some kind of `SearchInBucket(Predicate)` where we could avoid creating the array?

ak88: Alternatively returning a pooled array instead? Keeping the lock while searching inside the bucket, could have other side effects.
@benaadams  what do you think?

benaadams: Should keep the lock if changing the txs in the bucket? What if another tx came in from same sender doing same thing?

---

Why is it or?

If there are regular txs does it not matter if are blob txs? And do the blob txs only matter if not regular ones?

benaadams: Though probably lock at bucket level rather than whole pool

benaadams: Fair; add comment for 2.

src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -444,6 +444,14 @@ public AcceptTxResult SubmitTx(Transaction tx, TxHandlingOptions handlingOptions
...
                 if (accepted)
                 {
+                    if (tx.HasAuthorizationList)
+                    {
+                        foreach (var auth in tx.AuthorizationList)
+                        {
+                            if (auth.Authority is not null)
+                                _pendingDelegations.IncrementDelegationCount(auth.Authority!, auth.Nonce);
+                        }
+                    }
```
LukaszRozmej, line 454: Let's have `AddPendingDelegations` for symmetry

src/Nethermind/Nethermind.TxPool/DelegationCache.cs
```diff
@@ -0,0 +1,62 @@
...
+        }
+    }
+
+    private static UInt256 KeyMask(AddressAsKey key, UInt256 nonce)
+    {
+        //A nonce cannot exceed 2^64-1 and an address is 20 bytes, so we can pack them together in one u256
+        UInt256 addressPlusNonce = new(key.Value.Bytes);
+        nonce <<= 64 * 3;
+        addressPlusNonce += nonce;
+        return addressPlusNonce;
```
LukaszRozmej, line 60:
```suggestion
        //A nonce cannot exceed 2^64-1 and an address is 20 bytes, so we can pack them together in one u256
        ref byte baseRef = ref key.Value.Bytes[0];
        return new UInt256(Unsafe.ReadUnaligned<ulong>(ref baseRef),
          Unsafe.ReadUnaligned<ulong>(ref Unsafe.Add(ref baseRef, 8)),
          Unsafe.ReadUnaligned<uint>(ref Unsafe.Add(ref baseRef, 16)),
          nonce.u1);
```

src/Nethermind/Nethermind.TxPool/Filters/OnlyOneTxPerDelegatedAccountFilter.cs
LukaszRozmej: The search below is very fast. I would rather have this search under the lock as I think it won't be much slower than allocating an array?

src/Nethermind/Nethermind.TxPool/Collections/SortedPool.cs
```diff
@@ -530,6 +530,14 @@ public bool TryGetBucketsWorstValue(TGroupKey groupKey, out TValue? item)
             return false;
         }
 
+        public bool BucketEmptyExcept(TGroupKey groupKey, Func<TValue, bool> predicate)
```
LukaszRozmej, line 533: Maybe better name `BucketAny`?

LukaszRozmej (approved):

marcindsobczak (approved):
LGTM, left a few comments

src/Nethermind/Nethermind.TxPool.Test/OnlyOneTxPerDelegatedAccountFilterTest.cs
```diff
@@ -0,0 +1,161 @@
...
+        headInfoProvider.GetCurrentHeadSpec().Returns(Prague.Instance);
+        TxDistinctSortedPool standardPool = new TxDistinctSortedPool(MemoryAllowance.MemPoolSize, Substitute.For<IComparer<Transaction>>(), NullLogManager.Instance);
+        TxDistinctSortedPool blobPool = new BlobTxDistinctSortedPool(10, Substitute.For<IComparer<Transaction>>(), NullLogManager.Instance);
+        OnlyOneTxPerDelegatedAccountFilter filter = new(headInfoProvider, standardPool, blobPool, Substitute.For<IReadOnlyStateProvider>(), new CodeInfoRepository(), new DelegationCache());
+        Transaction transaction = Build.A.Transaction.SignedAndResolved(new EthereumEcdsa(0), TestItem.PrivateKeyA).TestObject;
+        TxFilteringState state = new();
+
+        AcceptTxResult result = filter.Accept(transaction, ref state, TxHandlingOptions.None);
+
+        Assert.That(result, Is.EqualTo(AcceptTxResult.Accepted));
```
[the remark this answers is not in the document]
ak88: Yes i forgot to delegate it

src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.cs
```diff
@@ -1770,6 +1770,89 @@ public void SubmitTx_CodeIsNotDelegationAndDelegation_DelegationIsAccepted((byte
...
+
+            result = _txPool.SubmitTx(secondTx, TxHandlingOptions.PersistentBroadcast);
+
+            result.Should().Be(AcceptTxResult.PendingDelegation);
+
+            _txPool.RemoveTransaction(firstTx.Hash);
+
+            result = _txPool.SubmitTx(secondTx, TxHandlingOptions.PersistentBroadcast);
+
+            result.Should().Be(AcceptTxResult.AlreadyKnown);
```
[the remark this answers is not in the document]
ak88: Added `withRemoval `

src/Nethermind/Nethermind.TxPool/Filters/OnlyOneTxPerDelegatedAccountFilter.cs
```diff
@@ -0,0 +1,38 @@
...
+            if (!spec.IsEip7702Enabled)
+                return AcceptTxResult.Accepted;
+
+            if (pendingDelegations.HasPending(tx.SenderAddress!, tx.Nonce))
+                return AcceptTxResult.PendingDelegation;
+
+            if (!codeInfoRepository.TryGetDelegation(worldState, tx.SenderAddress!, out _))
+                return AcceptTxResult.Accepted;
+            //Transactios from the same source can only be either blob transactions or some other type 
+            if (!standardPool.BucketEmptyExcept(tx.SenderAddress!, (t) => t.Nonce == tx.Nonce) || !blobPool.BucketEmptyExcept(tx.SenderAddress!, (t) => t.Nonce == tx.Nonce))
```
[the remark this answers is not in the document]
ak88: added

src/Nethermind/Nethermind.TxPool/AcceptTxResult.cs
```diff
@@ -89,6 +89,15 @@ namespace Nethermind.TxPool
         /// Ignores transactions if tx type is not supported
         /// </summary>
         public static readonly AcceptTxResult NotSupportedTxType = new(15, nameof(NotSupportedTxType));
+        /// <summary>
+        /// Only one tx is allowed per delegated account. 
+        /// </summary>
+        public static readonly AcceptTxResult OnlyOneTxPerDelegatedAccount = new(16, nameof(OnlyOneTxPerDelegatedAccount));
```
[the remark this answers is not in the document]
ak88: fixed

src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.cs
```diff
@@ -1805,11 +1805,12 @@ public void Delegated_account_can_only_have_one_tx()
...
             result = _txPool.SubmitTx(secondTx, TxHandlingOptions.PersistentBroadcast);
 
-            result.Should().Be(AcceptTxResult.OnlyOneTxPerDelegatedAccount);
+            result.Should().Be(AcceptTxResult.MoreThanOneTxPerDelegatedAccount);
         }
 
-        [Test]
-        public void Tx_with_pending_delegation_is_rejected_then_is_accepted_after_delegation_removal()
+        [TestCase(true)]
+        [TestCase(true)]
```
marcindsobczak, line 1812: twice true?

src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.cs
```diff
@@ -1842,15 +1843,20 @@ public void Tx_with_pending_delegation_is_rejected_then_is_accepted_after_delega
...
 
-            result = _txPool.SubmitTx(secondTx, TxHandlingOptions.PersistentBroadcast);
+                result.Should().Be(AcceptTxResult.Accepted);
+            }
+            else
+            {
+                result = _txPool.SubmitTx(secondTx, TxHandlingOptions.PersistentBroadcast);
 
-            result.Should().Be(AcceptTxResult.AlreadyKnown);
+                result.Should().Be(AcceptTxResult.AlreadyKnown);
```
marcindsobczak, line 1858: I think it will fail in `false` scenario and result will be `PendingDelegation`

marcindsobczak (approved):

src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.cs
ak88: Yes has been corrected now.

src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.cs
ak88: fixed

marcindsobczak (approved):

PR #8037, opened 2025-01-09
EvmStack native endian

Scooletz:
This PR changes the endianness of the `EvmStack` to be aligned with the platform it runs on. It means that no endianness conversions should be done when retrieving storing `UInt256`. The penalty though is related to pushing/poping bytes that now require reversing. According to the profiling this should happen much less frequent than regular operations.

Beside the endiannes, this PR add one method that allows to peek a `ref UInt256` to the top of the stack. This allows for any `Instruction` that has push/pop behavior to get a ref to the last parameter and directly set the value on the stack, without performing calculations for pop and push. Instead it will calculated the offset once for getting the `ref` and use it for pop.

## Changes

- _List the changes_

- [x] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Remarks

### DIFF
--- a/src/Nethermind/Nethermind.Evm/EvmStack.cs
+++ b/src/Nethermind/Nethermind.Evm/EvmStack.cs
@@ -11,8 +11,11 @@ using Nethermind.Evm.Tracing;
 using System.Diagnostics.CodeAnalysis;
 using System.Runtime.Intrinsics;
 using System.Diagnostics;
+using System.Numerics;
 using System.Runtime.Intrinsics.X86;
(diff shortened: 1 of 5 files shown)


Scooletz:
Benchmarks showed no more than 2% gain, which was much less than the standard deviation for the MGas/s. Closing in its current form.

PR #8040, opened 2025-01-10
Automatic history pruning

### LINKED ISSUE #7943: Historical blocks & receipts automated cleaner
There are several reasons to implement automated removal of historical blocks/receipts:

1. After dropping pre-merge history, users should be able to seamlessly reduce storage requirements without the need to resync the node.
2. During a full sync from genesis, if we are only interested in the last X amount of history, we should be able to drop anything older than X.
3. When a sliding window is introduced with [EIP-4444](https://eips.ethereum.org/EIPS/eip-4444), we should have the ability to easily drop history.

The key consideration is that dropping history should not interfere with current block processing.

Two possible solutions to achieve this are:
1. Drop X items with every new payload.
2. Start saving new blocks/receipts to a separate database and remove the entire database at a designated point in time.

Marchhill:
Resolves #7943 

## Changes

- Prunes block bodies and receipts outside of configurable window - rolling or up to ancient barriers
- Changes sync to not download outside of history expiry window
- Update peers of block range stored

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

Add section on configuring `History` settings. Users can configure `Pruning` and `RetentionEpochs` as they chose.

- [x] Yes
- [ ] No

Added support for automated history removal.


src/Nethermind/Nethermind.Blockchain/BlockTree.cs
```diff
@@ -55,7 +55,8 @@ public partial class BlockTree : IBlockTree
         private readonly IBloomStorage _bloomStorage;
         private readonly ISyncConfig _syncConfig;
         private readonly IChainLevelInfoRepository _chainLevelInfoRepository;
-
+        private readonly IHistoryConfig _historyConfig;
+        private readonly IBlocksConfig _blocksConfig;
```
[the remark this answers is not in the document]
Marchhill: Yep, is still draft will clean everything up and add some tests then ready for review ;)

src/Nethermind/Nethermind.Blockchain/Blocks/BlockStore.cs
```diff
@@ -96,4 +97,20 @@ public void Cache(Block block)
...
+        {
+            Block block = _blockDecoder.Decode(value.AsRlpStream());
+            if (block.Timestamp >= timestamp)
+            {
+                // exit early since the blocks are ordered by number
+                break;
+            }
+
+            yield return block;
+        }
```
LukaszRozmej: Not sure if this is optimal way?

flcl42 (commented):
Does it consider the state of block processing and synchronization? Should we prune only when are free from processing/production? Should we wait for blocks/receipts or even state  sync before pruning? Should we consider the settings in block/receipt sync?

src/Nethermind/Nethermind.Blockchain/Receipts/PersistentReceiptStorage.cs
```diff
@@ -357,19 +361,11 @@ private void EnsureCanonical(Block block, long? lastBlockNumber)
                 byte[] blockHash = block.Hash.BytesToArray();
                 foreach (Transaction tx in block.Transactions)
                 {
-                    Hash256 hash = (tx.Hash ??= tx.CalculateHash());
+                    tx.Hash ??= tx.CalculateHash();
+                    Hash256 hash = tx.Hash;
```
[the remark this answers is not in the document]
Marchhill: It's not needed but just cleaned this up since I was already making edits to the file. Doing multiple assignments on one line is [normally considered bad practice](https://softwareengineering.stackexchange.com/questions/197482/when-are-chained-assignments-i-e-a-b-c-bad-form) but ofc is subjective

Marchhill:
> Should we consider the state of block processing and synchronization? Should we prune only when are free from processing/production? Should we wait for blocks/receipts or even state sync before pruning? Should we consider the settings in block/receipt sync?

Good questions, I would appreciate more input from people who are more familiar with the sync process than I am. My thinking behind not checking if we are syncing:
- We only prune blocks older than 82125 epochs, outside the weak subjectivity period so it shouldn't affect anything
- If for some reason it does, then we may want to prune while syncing. If we didn't do this the size of the history DB could get very large during sync, before eventually being pruned. This loses the benefits of history expiry, as we will still need disk space to store history while syncing

You also mention block production, do you think we shouldn't prune while producing a block? Atm I can't see a problem with doing that

flcl42:
> We only prune blocks older than 82125 epochs, outside the weak subjectivity period so it shouldn't affect anything

We sync old blocks, so I suppose such pruning will prune lots of them

> * If for some reason it does, then we may want to prune while syncing. If we didn't do this the size of the history DB could get very large during sync, before eventually being pruned. This loses the benefits of history expiry, as we will still need disk space to store history while syncing

Yes, so what if we change sync to consider pruning border and do not sync old blocks at all? Then after syncing we turn on pruning and prune what became outdated(maybe a dozen of blocks or so)

> You also mention block production, do you think we shouldn't prune while producing a block? Atm I can't see a problem with doing that

Pruning is a secondary task that requires resources, seems like a good candidate for execution when ProcessingQueueEmpty. Purely optional for this request

src/Nethermind/Nethermind.Blockchain/HistoryPruning/HistoryPruner.cs
```diff
@@ -0,0 +1,126 @@
...
+    ulong secondsPerSlot,
+    ILogManager logManager) : IHistoryPruner
+{
+    private Task? _pruneHistoryTask;
+    private readonly Lock _pruneLock = new();
+    private ulong _lastPrunedTimestamp;
+    private readonly ILogger _logger = logManager.GetClassLogger();
+    private readonly bool _enabled = historyConfig.Enabled;
+    private readonly ulong _epochLength = secondsPerSlot * 32;
+    private readonly ulong _minHistoryRetentionEpochs = 82125;
```
smartprogrammer93, line 30: This is minimum is valid for mainnet, but might have other values in other chains. i prefer pulling such a value from chainspec

src/Nethermind/Nethermind.Config/IHistoryConfig.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Nethermind.Config;
+
+public interface IHistoryConfig : IConfig
+{
+    bool Enabled { get; }
+
+    // For EIP-4444 set default to 82125
+    [ConfigItem(
+        Description = "The number of epochs to retain historical blocks and receipts. Set to null for unlimited retention. Must be at least 82125.",
```
smartprogrammer93, line 12:
```suggestion
        Description = "The number of epochs to retain historical blocks and receipts. Set to null for unlimited retention. For mainnet this must be at least 82125.",
```

src/Nethermind/Nethermind.Taiko/Rpc/TaikoForkchoiceUpdatedHandler.cs
```diff
@@ -48,6 +48,7 @@ internal class TaikoForkchoiceUpdatedHandler(IBlockTree blockTree,
           specProvider,
           syncPeerPool,
           logManager,
+          null,
```
smartprogrammer93, line 51:
```suggestion
          _api.HistoryPruner,
```
just pass it?

smartprogrammer93 (changes requested):

smartprogrammer93:
there should definitely be changes to the syncer. why sync blocks that i will be pruning away later?! just stop syncing bodies and receipts when reaching the boundry

src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
```diff
@@ -83,11 +86,13 @@ public ForkchoiceUpdatedHandler(
         _syncPeerPool = syncPeerPool;
         _simulateBlockProduction = simulateBlockProduction;
         _secondsPerSlot = secondsPerSlot;
+        _historyPruner = historyPruner;
```
asdacap, line 89: The handler is merge specific. Can you make it work without modifying merge plugin? Think about pow network or network that does not use the rpc handler for consensus.

src/Nethermind/Nethermind.Blockchain/Blocks/BlockStore.cs
```diff
@@ -96,4 +97,20 @@ public void Cache(Block block)
     {
         _blockCache.Set(block.Hash, block);
     }
+
+    public IEnumerable<Block> GetBlocksOlderThan(ulong timestamp)
+    {
+        var blocks = blockDb.GetAll(true);
+        foreach ((byte[] _, byte[]? value) in blocks)
```
asdacap, line 104: The block record actually is not necessarily ordered. This is because previously the block keys are not prefixed with block number, so old running nethermind nodes will randomly not prune with this code. I suggest checking `blockInfo` for block by block number. Keep in mind, with portal network, the blocks are not continuous and it may need to store blocks even before the boundary becaus it still need to serve some block.

src/Nethermind/Nethermind.Blockchain/BlockTree.cs
```diff
@@ -887,6 +850,81 @@ private void DeleteBlocks(Hash256 deletePointer)
             }
         }
 
+        public IEnumerable<Block> DeleteBlocksBeforeTimestamp(ulong cutoffTimestamp, CancellationToken cancellationToken)
```
asdacap, line 853: Please move this method outside also. Timestamp based cutoff may not apply to all case, eg: portal. Also, plaese add a check so that no block is deleted after or at `BlockTree.SyncPivot`.

src/Nethermind/Nethermind.Blockchain/BlockTree.cs
```diff
@@ -887,6 +850,81 @@ private void DeleteBlocks(Hash256 deletePointer)
...
+        {
+            BlockAcceptingNewBlocks();
+            List<Block> blocks = [];
```
asdacap, line 856: Storing the blocks in a list is probably not a great idea as this might accidentally load a lot of blocks on existing node.

asdacap:
For the sync part of the code, the easiest way to integrate it is with some kind of interface `IBlockPersistenceStrategy` with a `ShouldPersistBlock(BlockInfo)`. `BodiesSyncFeed` and `ReceiptsSyncFeed` has a `SyncStatusList` with a `TryGetInfosForBatch` method that accept a function that you can put it to specify the logic if it should download the block or not. You can also update the `SyncConfigBarrierCalc` to have the progress log make more sense. Because the head, suggestedheader, bestnumbers are all confusing and painful to think about, I suggest just use `_blockTree.SyncPivot` as the assumed head. It make it much easier to reason about.

asdacap:
Actually, peers may not serve block if not done accurately. So the assumed head probably should be `BestSuggestedHeader`

src/Nethermind/Nethermind.Blockchain/Blocks/BlockStore.cs
Marchhill: changed to `continue` so all blocks will be scanned. I don't think `BlockInfo` can be used since it doesn't have timestamp

src/Nethermind/Nethermind.Blockchain/BlockTree.cs
Marchhill: I'm not sure it makes sense to move outside as we need to access lots of private state in `BlockTree`. For cases where it doesn't apply they can not call the method?

asdacap: Move it outside please. Try to think of a way to not use private state.

src/Nethermind/Nethermind.Blockchain/BlockTree.cs
```diff
@@ -887,6 +850,80 @@ private void DeleteBlocks(Hash256 deletePointer)
             }
         }
 
+        public IEnumerable<Block> DeleteBlocksBeforeTimestamp(ulong cutoffTimestamp, CancellationToken cancellationToken)
+        {
+            BlockAcceptingNewBlocks();
```
asdacap: This is probably not necessary and might accidentally block forward sync. Just don't delete anything after sync pivot.

src/Nethermind/Nethermind.Blockchain/Blocks/BlockStore.cs
asdacap: No, you can't scan all block 😢 . Those are potentially 800GB.

asdacap: Ok, here is one potential strategy, probably not here though. You need to save the last deleted point in a metadata somewhere. Then just iterate from there, one block at a time until you reach a new enough block.

asdacap: I guess at first run, it will do a very long delete, but that is fine, as the pointer is moving little by little.

src/Nethermind/Nethermind.Blockchain/HistoryPruning/HistoryPruner.cs
```diff
@@ -0,0 +1,136 @@
...
+        if (cutoffTimestamp <= _lastPrunedTimestamp)
+        {
+            return;
+        }
+
+        if (_logger.IsInfo) _logger.Info($"Pruning historical blocks up to timestamp {cutoffTimestamp}");
+
+        await Task.Run(() =>
+        {
+            var deletedBlocks = _blockTree.DeleteBlocksBeforeTimestamp(cutoffTimestamp, cancellationToken);
```
asdacap: Just iterate `IBlockStore` here.

src/Nethermind/Nethermind.Init/Steps/InitializeBlockchain.cs
```diff
@@ -166,6 +166,21 @@ protected virtual Task InitBlockchain()
...
+            if (historyConfig.Enabled)
+            {
+                HistoryPruner historyPruner = new(
+                    _api.BlockTree!,
+                    _api.ReceiptStorage!,
+                    _api.SpecProvider!,
+                    historyConfig,
+                    (long)blocksConfig.SecondsPerSlot,
+                    _api.LogManager);
+                historyPruner.CheckConfig();
```
asdacap: So.... just check in `HistoryPruner` constructor?

src/Nethermind/Nethermind.Blockchain/Blocks/BlockStore.cs
Marchhill: Two things here:

- I'm not sure how to store a pointer into db and continue searching from there
- If we are deleting all these old blocks then why do we need to store a pointer anyway? The first block in the db will be the deleted one after pruning

asdacap: The first block in the db is not necessarily the first block that should be deleted. This is because the block in the db is not always ordered because of legacy encoding.

asdacap: You can add a pointer somehow in metadata db.

Marchhill: What if we skip over x blocks because they are not old enough and then store the delete pointer. Then we will only iterate blocks after the delete pointer. Those x blocks will eventually become old enough to be deleted but they never will be

Marchhill: maybe we could have the delete pointer expire at this point but it gets very messy. Besides in most cases these out of order blocks would be very old so probably deleted anyway

Marchhill: when did we start storing ordered, was this pre-merge?

asdacap: We start storing ordered about 1 year post merge.

asdacap: > What if we skip over x blocks because they are not old enough and then store the delete pointer. Then we will only iterate blocks after the delete pointer. Those x blocks will eventually become old enough to be deleted but they never will be

This sound messier than just querying block number one by one.

Marchhill: I'm saying this is the reason why delete pointer approach won't work

Marchhill: we can't just start searching from after the pointer, because the blocks before the pointer will eventually have to be deleted too

Marchhill: only solution I can see atm is instead of supporting pre-merge pruning, we use the cutoff as one year post merge so we can assume the db is always ordered. even then it would depend on when the node was update which is awkward to work around

Marchhill: generally the delete pointer approach would also miss deleting some old blocks, as they are scattered randomly through the db

Marchhill: Ok I have a better idea how to solve: we have a config option like `ScanEntireDb` which users set to true if they have a long running node which was synced before 1.23. If this option is set we scan and delete from the entire db on the first round - if we find some old block which shouldn't be deleted yet we can add the block number prefix to the key. Once we've gone through one round we can store this in metadata and assume the db is ordered from then on

src/Nethermind/Nethermind.Api/NethermindApi.cs
```diff
@@ -89,6 +85,8 @@ public IBlockchainBridge CreateBlockchainBridge()
         public IEnode? Enode { get; set; }
         public IEthereumEcdsa EthereumEcdsa => Context.Resolve<IEthereumEcdsa>();
         public IFileSystem FileSystem { get; set; } = new FileSystem();
+        public IHistoryPruner? HistoryPruner { get; set; }
+        // public IHistoryPruner? HistoryPruner => Context.Resolve<IHistoryPruner>();
```
Marchhill: @asdacap better to resolve somehow like this?

src/Nethermind/Nethermind.Synchronization/FastBlocks/BodiesSyncFeed.cs
```diff
@@ -302,5 +311,19 @@ private void UpdateSyncReport()
...
         }
+
+        private class BodiesDownloadStrategy(IBlockTree blockTree, ISyncReport syncReport, IHistoryPruner? historyPruner) : IBlockDownloadStrategy
+        {
+            public bool ShouldDownloadBlock(BlockInfo info)
+            {
+                bool hasBlock = blockTree.HasBlock(info.BlockNumber, info.BlockHash);
+                long? cutoff = historyPruner?.CutoffBlockNumber;
+                // needed?
+                // cutoff = cutoff is null ? null : long.Max(cutoff.Value, blockTree.BestSuggestedHeader.Number);
```
Marchhill: @asdacap you mentioned using BestSuggestedHeader, do you think needs to be used here?

src/Nethermind/Nethermind.Synchronization/FastBlocks/ReceiptsSyncFeed.cs
```diff
@@ -32,18 +33,28 @@ public class ReceiptsSyncFeed : BarrierSyncFeed<ReceiptsSyncBatch?>
     {
         protected override long? LowestInsertedNumber => _syncPointers.LowestInsertedReceiptBlockNumber;
         protected override int BarrierWhenStartedMetadataDbKey => MetadataDbKeys.ReceiptsBarrierWhenStarted;
-        protected override long SyncConfigBarrierCalc => _syncConfig.AncientReceiptsBarrierCalc;
+        protected override long SyncConfigBarrierCalc
+        {
+            get
+            {
+                // could take a long time to return, safe to do?
+                long? cutoffBlockNumber = _historyPruner.CutoffBlockNumber;
```
Marchhill, line 40: @asdacap not sure if this is OK to do since getting `_historyPruner.CutoffBlockNumber` could take a long time of scanning the db. Especially if this is called before history pruning has happened.

asdacap: Then.... don't do it? Why is it doing it?

src/Nethermind/Nethermind.Synchronization/FastBlocks/BodiesSyncFeed.cs
asdacap: No? Why does it need that?

src/Nethermind/Nethermind.Synchronization.Test/E2ESyncTests.cs
```diff
@@ -158,9 +160,10 @@ private IContainer CreateNode(PrivateKey nodeKey, Action<IConfigProvider, ChainS
                 }
         }
 
-        var builder = new ContainerBuilder()
+        ContainerBuilder builder = new ContainerBuilder()
             .AddModule(new PseudoNethermindModule(spec, configProvider, new TestLogManager()))
             .AddModule(new TestEnvironmentModule(nodeKey, $"{nameof(E2ESyncTests)} {dbMode} {isPostMerge}"))
+            .AddSingleton(Substitute.For<IHistoryPruner>())
```
asdacap: Add to `BlockTreeModule`. Don't use substitute here.

src/Nethermind/Nethermind.Merge.Plugin.Test/BaseEngineModuleTests.cs
```diff
@@ -258,6 +261,16 @@ protected override IBlockProducer CreateTestBlockProducer()
             PostMergeBlockProducer? postMergeBlockProducer = blockProducerFactory.Create(blockProducerEnv);
             BlockProducer = postMergeBlockProducer;
 
+            HistoryPruner = new HistoryPruner(
```
asdacap: Add to `BlockTreeModule` and you don't have to construct it manually, and just use `Container.Resolve` here.

src/Nethermind/Nethermind.Api/NethermindApi.cs
asdacap: Yes, make it consistent with test.

src/Nethermind/Nethermind.Merge.Plugin.Test/BaseEngineModuleTests.cs
asdacap: Actually, but it in `NethermindModule` as `BlockTreeModule` is in test only, so that you don't have to make one in `InitializeBlockTree` also.

src/Nethermind/Nethermind.Blockchain/HistoryPruning/HistoryPruner.cs
```diff
@@ -0,0 +1,269 @@
...
+        _enabled = historyConfig.Enabled;
+        _epochLength = secondsPerSlot * 32;
+        _minHistoryRetentionEpochs = specProvider.GenesisSpec.MinHistoryRetentionEpochs;
+
+        CheckConfig();
+        LoadDeletePointer();
+    }
+
+    public void OnBlockProcessorQueueEmpty(object? sender, EventArgs e)
+        => _ = TryPruneHistory(CancellationToken.None);
```
asdacap: Add `IProcessExitSource` and pass in the token from there.

src/Nethermind/Nethermind.Blockchain/HistoryPruning/HistoryPruner.cs
```diff
@@ -0,0 +1,269 @@
...
+            {
+                return;
+            }
+
+            _pruneHistoryTask = PruneHistory(cancellationToken);
+        }
+        await _pruneHistoryTask;
+    }
+
+    public long? CutoffBlockNumber
```
asdacap: Right, I see where the problem is....

asdacap: Perhaps have a `bool ShouldDownload(long blockNumber)`, fetch the header there, check if timestamp > cut off timestamp, store a `_lowestKnownShouldDownload` as a hueristic. Or its probably the same as `_deletePointer`.

src/Nethermind/Nethermind.Blockchain/HistoryPruning/HistoryPruner.cs
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    private IEnumerable<Block> GetBlocksBeforeTimestamp(ulong cutoffTimestamp)
+        => GetBlocksByNumber(_deletePointer, b => b.Timestamp >= cutoffTimestamp, i => _deletePointer = i);
+
+    private IEnumerable<Block> GetBlocksByNumber(long from, Predicate<Block> endSearch, Action<long> onFirstBlock)
+    {
+        bool firstBlock = true;
+        long headNumber = _blockTree.Head!.Number;
+        for (long i = from; i <= headNumber; i++)
```
asdacap: Don't do anything after `_blockTree.Pivot`.

src/Nethermind/Nethermind.Blockchain/HistoryPruning/HistoryPruner.cs
```diff
@@ -0,0 +1,269 @@
...
+    private ulong CalculateCutoffTimestamp()
+    {
+        ulong cutoffTimestamp = 0;
+
+        if (_historyConfig.HistoryRetentionEpochs.HasValue)
+        {
+            cutoffTimestamp = _blockTree.Head!.Timestamp - (ulong)(_historyConfig.HistoryRetentionEpochs.Value * _epochLength);
+        }
+
+        if (_historyConfig.DropPreMerge)
```
asdacap: `DropPreMerge` has priority over `HistoryRetentionEpochs`? Or the other way round?

src/Nethermind/Nethermind.Blockchain/HistoryPruning/HistoryPruner.cs
```diff
@@ -0,0 +1,269 @@
...
+        ulong cutoffTimestamp = CalculateCutoffTimestamp();
+
+        if (cutoffTimestamp <= _lastPrunedTimestamp)
+        {
+            return;
+        }
+
+        if (_logger.IsInfo) _logger.Info($"Pruning historical blocks up to timestamp {cutoffTimestamp}");
+
+        await Task.Run(() => PruneBlocksAndReceipts(cutoffTimestamp, cancellationToken), cancellationToken);
```
asdacap: Probably run it with `IBackgroundTaskScheduler` so that it does not affect attestation. This will be long for existing user.

src/Nethermind/Nethermind.Blockchain/HistoryPruning/HistoryPruner.cs
```diff
@@ -0,0 +1,269 @@
...
+
+    private void PruneBlocksAndReceipts(ulong cutoffTimestamp, CancellationToken cancellationToken)
+    {
+        int deletedBlocks = 0;
+        try
+        {
+            using BatchWrite batch = _chainLevelInfoRepository.StartBatch();
+
+            IEnumerable<Block> blocks = GetBlocksBeforeTimestamp(cutoffTimestamp);
+            foreach (Block block in blocks)
```
asdacap: Probably have a limit so that the batch does not get too large.

src/Nethermind/Nethermind.Blockchain/BlockTree.cs
```diff
@@ -1672,35 +1670,43 @@ public int DeleteChainSlice(in long startNumber, long? endNumber = null, bool fo
                 newHeadBlock = newHeadHash is null ? null : FindBlock(newHeadHash, BlockTreeLookupOptions.None, blockNumber: startNumber - 1);
             }
 
-            using (_chainLevelInfoRepository.StartBatch())
+            BlockAcceptingNewBlocks();
```
asdacap: Remove the `BlockAcceptingNewBlocks` dont rely on it.

src/Nethermind/Nethermind.Blockchain/BlockTree.cs
```diff
@@ -807,15 +796,50 @@ public void DeleteInvalidBlock(Block invalidBlock)
...
+                _chainLevelInfoRepository.Delete(currentNumber, batch);
+            }
+            else if (currentLevel is not null)
+            {
+                _chainLevelInfoRepository.PersistLevel(currentNumber, currentLevel, batch);
+            }
+
+            _blockStore.Delete(currentNumber, currentHash);
+            _headerStore.Delete(currentHash);
+            if (isOldBlock) _blockInfoDb.Delete(currentHash);
```
asdacap: `_chainLevelInfoRepository` uses `_blockInfoDb`. So use `_chainLevelInfoRepository.Delete` instead.

src/Nethermind/Nethermind.Blockchain/IBlockTree.cs
```diff
@@ -157,6 +158,7 @@ AddBlockResult Insert(Block block, BlockTreeInsertBlockOptions insertBlockOption
         IOwnedReadOnlyList<BlockHeader> FindHeaders(Hash256 hash, int numberOfBlocks, int skip, bool reverse);
 
         void DeleteInvalidBlock(Block invalidBlock);
+        void DeleteBlock(long currentNumber, Hash256 currentHash, Hash256 nextHash, BatchWrite batch, ChainLevelInfo? currentLevel = null, bool isOldBlock = false);
```
asdacap: Try to reduce number of parameter here.

src/Nethermind/Nethermind.Blockchain/HistoryPruning/HistoryPruner.cs
Marchhill: yeah that could help if we haven't run pruning yet. For the first call it could be really slow though. Could do a binary search to speed it up

src/Nethermind/Nethermind.Blockchain/HistoryPruning/HistoryPruner.cs
Marchhill: there's a check that nothing from the sync pivot onwards will be deleted

src/Nethermind/Nethermind.Blockchain/HistoryPruning/HistoryPruner.cs
Marchhill: neither has priority, we just apply both. If `DropPreMerge` is set we will drop pre-merge history. If `HistoryRetentionEpochs` is set we will drop anything older than that many epochs, so they could of course overlap

src/Nethermind/Nethermind.Blockchain/HistoryPruning/HistoryPruner.cs
```diff
@@ -0,0 +1,358 @@
...
+using Nethermind.Serialization.Rlp;
+using Nethermind.State.Repositories;
+
+namespace Nethermind.Blockchain.HistoryPruning;
+
+public class HistoryPruner : IHistoryPruner
+{
+    private Task? _pruneHistoryTask;
+    private readonly Lock _pruneLock = new();
+    private readonly Lock _searchLock = new();
```
LukaszRozmej: why two locks needed?

src/Nethermind/Nethermind.Synchronization/FastBlocks/ReceiptsSyncFeed.cs
Marchhill: From your comment you said to update `SyncBarrierConfigCalc` https://github.com/NethermindEth/nethermind/pull/8040#issuecomment-2795483580

asdacap: Ow, okk. I thought a simple head-numOfBlocks would do it. But it seems to need by time.

src/Nethermind/Nethermind.History.Test/HistoryPrunerTests.cs
```diff
@@ -0,0 +1,336 @@
...
+        // n.b. technically invalid, should be at least 82125 epochs
+        // however not feasible to test this
+        IHistoryConfig historyConfig = new HistoryConfig
+        {
+            HistoryRetentionEpochs = 2,
+            DropPreMerge = false
+        };
+        ISpecProvider specProvider = Substitute.For<ISpecProvider>();
+
+        HistoryPruner historyPruner = new(
```
asdacap: Note: You can `testBlockchain.Container.Resolve<IHistoryPruner>()`. 

To change the config,  the `BasicTestBlockchain.Create()` or some overload, take a configurator func in which you can set the history config with sometthing like `.AddSingleton<IHistoryConfig>(theHistoryConfig)`.

src/Nethermind/Nethermind.Config/IHistoryConfig.cs
```diff
@@ -0,0 +1,20 @@
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+namespace Nethermind.Config;
+
+public interface IHistoryConfig : IConfig
```
asdacap, line 8: Ask the core slack channel should this have a dedicated config or just use `IBlocksConfig`.

src/Nethermind/Nethermind.Blockchain/Blocks/BlockStore.cs
Marchhill: outdated

src/Nethermind/Nethermind.Synchronization/FastBlocks/ReceiptsSyncFeed.cs
Marchhill: yeah we need to convert timestamp to block number which could always require a long db search. What do you think, is it important to update `SyncBarrierConfigCalc ` or should we just leave it?

src/Nethermind/Nethermind.Blockchain/HistoryPruning/HistoryPruner.cs
Marchhill: to make sure we only have max one pruning task, and one searching task (converting timestamp -> block number) at a time

src/Nethermind/Nethermind.Config/IHistoryConfig.cs
Marchhill: I prefer separate config here, happy to change if reviewers prefer to put in `IBlocksConfig` though

src/Nethermind/Nethermind.Blockchain/IBlockTree.cs
Marchhill: I think all of these are needed

src/Nethermind/Nethermind.Blockchain/HistoryPruning/HistoryPruner.cs
Marchhill: Limited to 64, not sure if that's a reasonable amount

src/Nethermind/Nethermind.Synchronization/FastBlocks/ReceiptsSyncFeed.cs
Marchhill: using binary search it should be fine even in the worst case

src/Nethermind/Nethermind.Api/NethermindApi.cs
```diff
@@ -87,6 +84,8 @@ public IBlockchainBridge CreateBlockchainBridge()
         public IEnode? Enode { get; set; }
         public IEthereumEcdsa EthereumEcdsa => Context.Resolve<IEthereumEcdsa>();
         public IFileSystem FileSystem { get; set; } = new FileSystem();
+        // public IHistoryPruner? HistoryPruner { get; set; }
```
asdacap: Missed

src/Nethermind/Nethermind.AuRa.Test/Contract/TxPriorityContractTests.cs
```diff
@@ -354,7 +355,7 @@ protected override ILocalDataSource<IEnumerable<Address>> GetWhitelistLocalDataS
             protected override ILocalDataSource<IEnumerable<TxPriorityContract.Destination>> GetMinGasPricesLocalDataStore() =>
                 LocalDataSource.GetMinGasPricesLocalDataSource();
 
-            protected override Task<TestBlockchain> Build(Action<ContainerBuilder> configurer = null)
+            protected override Task<TestBlockchain> Build(Action<ContainerBuilder> configurer = null, IEnumerable<IConfig>? configs = null)
```
asdacap: Missed also?

src/Nethermind/Nethermind.Blockchain/IBlockTree.cs
```diff
@@ -157,6 +158,7 @@ AddBlockResult Insert(Block block, BlockTreeInsertBlockOptions insertBlockOption
         IOwnedReadOnlyList<BlockHeader> FindHeaders(Hash256 hash, int numberOfBlocks, int skip, bool reverse);
 
         void DeleteInvalidBlock(Block invalidBlock);
+        void DeleteBlock(long currentNumber, Hash256 currentHash, Hash256? nextHash, BatchWrite batch, ChainLevelInfo? currentLevel = null, bool isOldBlock = false);
```
asdacap: Please cut down the number of exposed param. If its needed internally, just proxy it.

src/Nethermind/Nethermind.Init/Steps/EthereumStepsLoader.cs
```diff
@@ -32,7 +31,7 @@ public IEnumerable<StepInfo> ResolveStepsImplementations()
 
             return _stepsInfo
                 .GroupBy(s => s.StepBaseType)
-                .Select(g => SelectImplementation(g.ToArray()))
+                .Select(g => SelectImplementation([.. g]))
```
asdacap: Is this some kind of auto format?

src/Nethermind/Nethermind.Init/Steps/EthereumStepsManager.cs
```diff
@@ -62,16 +62,11 @@ private List<Task> CreateAndExecuteSteps(CancellationToken cancellationToken)
...
                 cancellationToken.ThrowIfCancellationRequested();
 
-                Func<IStep> stepFactory = () =>
-                {
-                    return CreateStepInstance(stepInfo);
-                };
-
                 Debug.Assert(!stepInfoMap.ContainsKey(stepInfo.StepBaseType), "Resolve steps implementations should have deduplicated step by base type");
-                stepInfoMap.Add(stepInfo.StepBaseType, new StepWrapper(stepFactory, stepInfo));
+                stepInfoMap.Add(stepInfo.StepBaseType, new StepWrapper(() => CreateStepInstance(stepInfo), stepInfo));
```
asdacap: Don't change things unnecessarily, it makes reviewer's job harder and therefore make people don't want to review your PR. One or two things is ok, but this whole file does not need to change.

src/Nethermind/Nethermind.Init/Steps/StartMonitoring.cs
```diff
@@ -19,7 +18,7 @@
 
 namespace Nethermind.Init.Steps;
 
-[RunnerStepDependencies(typeof(InitializeBlockTree))]
+[RunnerStepDependencies(typeof(InitializeBlockchain))]
```
asdacap, line 24: Why?

src/Nethermind/Nethermind.Synchronization/FastBlocks/BodiesSyncFeed.cs
```diff
@@ -302,5 +310,17 @@ private void UpdateSyncReport()
             _syncReport.FastBlocksBodies.Update(_pivotNumber - _syncStatusList.LowestInsertWithoutGaps);
             _syncReport.FastBlocksBodies.CurrentQueued = _syncStatusList.QueueSize;
         }
+
+        private class BodiesDownloadStrategy(IBlockTree blockTree, ISyncReport syncReport, IHistoryPruner? historyPruner) : IBlockDownloadStrategy
```
asdacap, line 313: Can be DI'ed. Portal will most likely override this.

src/Nethermind/Nethermind.History.Test/HistoryPrunerTests.cs
```diff
@@ -0,0 +1,307 @@
...
+    private static readonly ulong BeaconGenesisTimestamp = (ulong)new DateTimeOffset(TestBlockchain.InitialTimestamp).ToUnixTimeSeconds() + (BeaconGenesisBlockNumber * SecondsPerSlot);
+    private static readonly IBlocksConfig BlocksConfig = new BlocksConfig()
+    {
+        SecondsPerSlot = SecondsPerSlot
+    };
+
+    [Test]
+    public async Task Can_prune_blocks_older_than_specified_epochs()
+    {
+        const int Blocks = 100;
```
asdacap: NP: Should be camel case.

src/Nethermind/Nethermind.AuRa.Test/Contract/TxPriorityContractTests.cs
Marchhill: No this is needed, check `TestBlockchain:Build`. I needed to pass in the configs there as adding to the container didn't work

src/Nethermind/Nethermind.Init/Steps/StartMonitoring.cs
Marchhill: There are DI errors otherwise, Smoke tests fail on optimism chains. This should run after `InitializeBlockchain` since that's where `IBackgroundTaskScheduler` is initialised

src/Nethermind/Nethermind.AuRa.Test/Contract/TxPriorityContractTests.cs
asdacap: I mean you did not pass the config to the base in line 367.

Marchhill: Please check this actually, not sure if there is a better way to do it

asdacap: Just double check if just adding another `builder.AddSingleton<IHistoryConfig>(theChangedConfig)` works. It should work just fine.

src/Nethermind/Nethermind.Blockchain/Blocks/BlockStore.cs
flcl42: Can we reorder and then do stuff?

src/Nethermind/Nethermind.Blockchain/Receipts/PersistentReceiptStorage.cs
```diff
@@ -330,6 +330,21 @@ public void EnsureCanonical(Block block)
             EnsureCanonical(block, null);
         }
 
+        public void RemoveReceipts(Block block)
+        {
+            _receiptsCache.Delete(block.Hash);
+
+            Span<byte> blockNumPrefixed = stackalloc byte[40];
+            GetBlockNumPrefixedKey(block.Number, block.Hash, blockNumPrefixed);
+            _blocksDb.Remove(blockNumPrefixed);
```
flcl42: It seems like `RemoveReceipts` means also block deletion, which is not obvious from the name. Also can it happen so a user sets receipts barrier higher than block bodies barrier, but bodies disappear too?

src/Nethermind/Nethermind.History/HistoryPruner.cs
```diff
@@ -0,0 +1,454 @@
...
+        IDbProvider dbProvider,
+        IHistoryConfig historyConfig,
+        IBlocksConfig blocksConfig,
+        IProcessExitSource processExitSource,
+        IBackgroundTaskScheduler backgroundTaskScheduler,
+        ILogManager logManager)
+    {
+        _specProvider = specProvider;
+        _logger = logManager.GetClassLogger();
+        LoggingInterval = _logger.IsDebug ? 5 : 100000;
```
flcl42: Can be _loggingIntervalMs or so; it's not clear what is this choice between 5 : 100000? Remove every 5ms? Report every 27days? And the value is not really used below.

src/Nethermind/Chains/foundation.json
```diff
@@ -151,6 +151,7 @@
     "accountStartNonce": "0x0",
     "maximumExtraDataSize": "0x20",
     "minGasLimit": "0x1388",
+    "minHistoryRetetionEpochs": "0x140CD",
```
flcl42, line 154: Hex in low registry

flcl42 (changes requested):
It probably should not be aware of pruning (I mean this lock). Might be cool to have DbIntensiveBackgroundTaskLock/CpuIntensiveBackgroundTaskLocks usable by all such tasks.

alexb5dh:
Need to trigger `Eth69ProtocolHandler.NotifyOfNewRange` when earliest available block changes. Either as part of this or a follow-up PR.

src/Nethermind/Nethermind.Config/HistoryConfig.cs
```diff
@@ -0,0 +1,12 @@
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+namespace Nethermind.Config;
+
+public class HistoryConfig : IHistoryConfig
+{
+    public bool Enabled => RetentionEpochs is not null || DropPreMerge;
```
flcl42: For users it might be more convenient to just activate it, or have it active by default in the future. Something like this: 
- for now by default, Enabled=false, DropPreMerge=true, RetentionEpochs=picked from spec or default to 82125
- in the future, Enabled=true, DropPreMerge=true, RetentionEpochs=picked from spec or default to 82125.

Btw why should we care about premerge stuff? If rolling window does not include it, it should be dropped anyway, shouldn't it? I mean DropPreMerge  option

src/Nethermind/Nethermind.History/HistoryPruner.cs
Marchhill: It controls how often we show the deletion progress logs, so in debug we log every 5 blocks deleted but in prod every 100000 (around 2 weeks of logs)

Marchhill: It controls how often we show the deletion progress logs, so in debug we log every 5 blocks deleted but in prod every 100000 (around 2 weeks of logs)

src/Nethermind/Nethermind.Config/HistoryConfig.cs
Marchhill: `Enabled` isn't set, it's determined by the other two variables. `DropPreMerge` makes it easy to just prune pre-merge history without setting a rolling window - we might want to do this for a while until we have a good post-merge history storage solution

src/Nethermind/Chains/foundation.json
Marchhill: lowercase?

src/Nethermind/Nethermind.Blockchain/Receipts/PersistentReceiptStorage.cs
Marchhill: this variable is actually the receipts column from blocks db, so no blocks should be deleted. Renamed variable to make clearer

src/Nethermind/Chains/foundation.json
flcl42: Yes

src/Nethermind/Nethermind.History/HistoryPruner.cs
flcl42: 100000 of 12sec blocks is 333 days?

src/Nethermind/Nethermind.Config/HistoryConfig.cs
flcl42: Yeah I see, and seems like a problem. I guess people will just want to enable it and be not aware of 82125

Marchhill: When we enable it as default they won't have to be aware. Could add another config option like `DropOutside4444Window` or something to make it simpler though

src/Nethermind/Nethermind.Config/IHistoryConfig.cs
```diff
@@ -1,20 +1,35 @@
...
+        Description = "The number of epochs to retain historical blocks and receipts when using 'Rolling' pruning mode. For mainnet this must be at least 82125.",
+        DefaultValue = "82125")]
+    long RetentionEpochs { get; set; }
+}
+
+public enum PruningModes
+{
+    [Description("No history pruning.")]
+    Disabled,
+
```
flcl42, line 28: One less empty line

src/Nethermind/Nethermind.Synchronization/SyncServer.cs
```diff
@@ -429,36 +433,56 @@ private void OnNewHeadBlock(object? sender, BlockEventArgs blockEventArgs)
...
                 return;
 
             Block latestBlock = latestBlockEventArgs.Block;
 
             // Notify every 32 blocks
-            if (latestBlock.Number % BlockRangeUpdateFrequency != 0)
+            if (latestBlock.Number % NewHeadBlockRangeUpdateFrequency != 0)
+                return;
+
+            OnNewRange(Genesis, latestBlock.Header);
```
alexb5dh: `earliest` should be earliest available block here. As I understand it's not always `Genesis` now, once pruning is enabled.

Marchhill: Oh yeah thanks missed this one. I think even before history pruning it wouldn't be genesis though? Unless this is an archive node

src/Nethermind/Nethermind.History/HistoryPruner.cs
```diff
@@ -0,0 +1,484 @@
...
+            {
+                var cts = CancellationTokenSource.CreateLinkedTokenSource(backgroundTaskToken, cancellationToken);
+                return TryPruneHistory(cts.Token);
+            });
+
+    internal Task TryPruneHistory(CancellationToken cancellationToken)
+    {
+        lock (BackgroundTaskScheduler.DbIntensiveBackgroundTaskLock)
+        {
+            lock (_pruneLock)
```
benaadams: Do you want to create a queue of pruning tasks?

MIght be better to use `TryEnter` and quit if already locked (i.e. in progress)
```csharp
if (_pruneLock.TryEnter())
{
    try
    {
        // body
    }
    finally
    {
        _pruneLock.Exit();
    }
}
 ```

File: src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs [hunk 46]
```diff
@@ -31,6 +31,8 @@ namespace Nethermind.Consensus.Scheduler;
 /// </summary>
 public class BackgroundTaskScheduler : IBackgroundTaskScheduler, IAsyncDisposable
 {
+    public static readonly Lock DbIntensiveBackgroundTaskLock = new();
```
benaadams: This is only used in one place; which then immediately gets a second lock. Is it useful?

src/Nethermind/Nethermind.History/HistoryPruner.cs
```diff
@@ -0,0 +1,484 @@
...
+        => _backgroundTaskScheduler.ScheduleTask(1,
+            (_, backgroundTaskToken) =>
+            {
+                var cts = CancellationTokenSource.CreateLinkedTokenSource(backgroundTaskToken, cancellationToken);
+                return TryPruneHistory(cts.Token);
+            });
+
+    internal Task TryPruneHistory(CancellationToken cancellationToken)
+    {
+        lock (BackgroundTaskScheduler.DbIntensiveBackgroundTaskLock)
```
benaadams: Is this useful?

src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
Marchhill: this was suggested by @flcl42 . I think it makes sense to have even if only used here atm. If people add other db intensive background tasks in future it will remind them to use this lock

src/Nethermind/Nethermind.History/HistoryPruner.cs
Marchhill: responded above

src/Nethermind/Nethermind.History/HistoryPruner.cs
```diff
@@ -0,0 +1,528 @@
...
+        get
+        {
+            if (!_hasLoadedDeletePointer)
+            {
+                bool lockTaken = false;
+                // take lock before updating delete pointer
+                // avoids race conditions with pruning
+                try
+                {
+                    Monitor.TryEnter(_pruneLock, LockWaitTimeoutMs, ref lockTaken);
```
benaadams, line 215: Can rerrange to have try inside if

```csharp
Monitor.TryEnter(_pruneLock, LockWaitTimeoutMs, ref lockTaken)
if (lockTaken)
{
     try {} finally { Monitor.Exit(_pruneLock) }
}
```

src/Nethermind/Nethermind.History/HistoryPruner.cs
```diff
@@ -0,0 +1,528 @@
...
+                var cts = CancellationTokenSource.CreateLinkedTokenSource(backgroundTaskToken, cancellationToken);
+                return TryPruneHistory(cts.Token);
+            });
+
+    internal Task TryPruneHistory(CancellationToken cancellationToken)
+    {
+        bool lockTaken = false;
+        try
+        {
+            Monitor.TryEnter(_pruneLock, LockWaitTimeoutMs, ref lockTaken);
```
benaadams, line 255: Also here; TryEnter outside try; then try inside if

benaadams (approved):

src/Nethermind/Nethermind.History/HistoryPruner.cs
Marchhill: discussed offline

src/Nethermind/Nethermind.History/HistoryPruner.cs
```diff
@@ -0,0 +1,519 @@
...
+                if (_blockTree.Head is null ||
+                    _blockTree.SyncPivot.BlockNumber == 0 ||
+                    !TryLoadDeletePointer() ||
+                    !ShouldPruneHistory(out ulong? cutoffTimestamp))
+                {
+                    _logger.Info($"[prune] Skipping historical block pruning.");
+                    return Task.CompletedTask;
+                }
+
+                lock (BackgroundTaskScheduler.DbIntensiveBackgroundTaskLock)
```
asdacap: This is not how you use `IBackgroundTaskScheduler`. You might as well not use it at all.

asdacap (changes requested):

src/Nethermind/Nethermind.Blockchain/BlockTree.cs
```diff
@@ -1755,9 +1753,8 @@ public void ForkChoiceUpdated(Hash256? finalizedBlockHash, Hash256? safeBlockHas
             TryUpdateSyncPivot();
         }
 
-        public long GetLowestBlock()
-        {
-            return _syncConfig.AncientReceiptsBarrierCalc < _syncConfig.AncientBodiesBarrierCalc ? _syncConfig.AncientReceiptsBarrierCalc : _syncConfig.AncientBodiesBarrierCalc;
-        }
+        public long GetLowestBlock() => _oldestBlock;
```
asdacap, line 1764: Is this used anywhere?

src/Nethermind/Nethermind.Synchronization.Test/OldStyleFullSynchronizerTests.cs
```diff
@@ -52,6 +53,7 @@ public void Setup()
 
             IContainer container = new ContainerBuilder()
                 .AddModule(new TestNethermindModule(configProvider))
+                .AddSingleton<IHistoryPruner>(Substitute.For<IHistoryPruner>())
```
asdacap, line 56: Does it break something?

Marchhill: Yeah, without this the synchronizer tests run forever

asdacap: Why? Will it cause issue in prod?

src/Nethermind/Nethermind.Blockchain/BlockTree.cs
Marchhill: yes, `GetLowestBlock` is used to check if a block is pruned for RPC. Before it was set to ancient barriers so needed to change

src/Nethermind/Nethermind.History/HistoryPruner.cs
Marchhill: what's wrong with it, the lock?

asdacap (approved):

src/Nethermind/Chains/foundation.json
```diff
@@ -151,6 +151,7 @@
     "accountStartNonce": "0x0",
     "maximumExtraDataSize": "0x20",
     "minGasLimit": "0x1388",
+    "minHistoryRetetionEpochs": "0x140cd",
```
flcl42, line 154: Retention

flcl42 (approved):

smartprogrammer93 (approved):

PR #8041, opened 2025-01-10
Add `curl` as an explicit dependency

### LINKED ISSUE #6805: Nethermind PPA package will fail if curl is not installed
**Description**
Nethermind PPA package will fail if curl is not installed

**Steps to Reproduce**

on machine with no curl installed:

```bash
sudo apt-get install nethermind
```

**Actual behavior**

```
Downloading https://github.com/NethermindEth/nethermind/releases/download/1.25.4/nethermind-1.25.4-20b10b35-linux-x64.zip
/var/lib/dpkg/info/nethermind.postinst: line 32: curl: command not found
dpkg: error processing package nethermind (--configure):
 installed nethermind package post-installation script subprocess returned error exit status 127
```

**Expected behavior**

package installed with no fuss no muss

rubo:
Fixes #6805 

## Changes

- Added `curl` as an explicit dependency to PPA configuration
- Updated the description as per README

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

### DIFF
--- a/scripts/deployment/ppa/debian/control
+++ b/scripts/deployment/ppa/debian/control
@@ -11,6 +11,6 @@ Rules-Requires-Root: no
 
 Package: nethermind
 Architecture: amd64 arm64
-Depends: ${misc:Depends}, ${shlibs:Depends}, unzip
+Depends: ${misc:Depends}, ${shlibs:Depends}, curl, unzip
 Description: A robust execution client for Ethereum node operators.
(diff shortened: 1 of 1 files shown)


LukaszRozmej (commented):
Do we really need `curl` as dependency?

LukaszRozmej:
> Yes. We need to download Nethermind from GitHub.

can't we do it with something built-in?

MarekM25 (approved):

rubo:
> can't we do it with something built-in?

Usually, both `curl` and `wget` are available, but apparently, some people have different configurations. I'm not sure what your concern is. Including a dependency makes the package manager install it automatically without user interaction.

LukaszRozmej (approved):

PR #8044, opened 2025-01-13
Add Worldchain support

deffrian:
Fixes Closes Resolves #

Worldchain support

## Changes

- Adds worldchain mainnet & sepolia configs
- CodeHash based genesis file

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainSpecLoader.cs
+++ b/src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainSpecLoader.cs
@@ -355,15 +355,13 @@ public class ChainSpecLoader(IJsonSerializer serializer) : IChainSpecLoader
             return;
         }
 
-        Dictionary<Hash256, byte[]> codes = new();
-        codes[Hash256.Zero] = [];
         if (chainSpecJson.CodeHashes is not null)
         {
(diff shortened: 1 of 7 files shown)


src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainSpecLoader.cs
```diff
@@ -355,6 +355,17 @@ private static void LoadAllocations(ChainSpecJson chainSpecJson, ChainSpec chain
             return;
         }
 
+        Dictionary<Hash256, byte[]> codes = new();
```
LukaszRozmej, line 358: I think we don't need this dictionary and can just use `chainSpecJson.CodeHashes` directly later to conserve a bit of memory.

LukaszRozmej (approved):
Looking at Megabytes worth of JSON, maybe we should add support for them in zipped format, not to blow repository size?

LukaszRozmej:
Do we want to add any auto pivot updates like in #8043?

PR #8046, opened 2025-01-13
Add single block receipt migration functionality

kamilchodola:
Adds possibility to fix only single block with missing TxHash<---> Receipt connection

Current method goes from 0 to blockNum which is long and annoying if you have missing receipt on block 30mln.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [X] No

- [X] Yes
- [ ] No

New parameter to debug_migrateReceipts added: `migrateSingleBlock`. Default value = false.

When set, combined with blockNumber allows to execute migration only for single block rather entire range which helps fixing TxHash<->Receipt relation quickly.

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain/Receipts/IReceiptsMigration.cs
+++ b/src/Nethermind/Nethermind.Blockchain/Receipts/IReceiptsMigration.cs
@@ -7,6 +7,6 @@ namespace Nethermind.Blockchain.Receipts
 {
     public interface IReceiptsMigration
     {
-        Task<bool> Run(long blockNumber);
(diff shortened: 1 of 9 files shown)


LukaszRozmej:
I think #8047 is more flexible?

kamilchodola:
@LukaszRozmej Yes - focusing on the other one now with testing so to be closed.

PR #8047, opened 2025-01-14
Feature/range receipt migration

asdacap:
- Allow `debug_migrateReceipts` to specify `from` and `to`.
- Account for strange off by one migration. The `from` and `to` is not inclusive.

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Called manually.


asdacap:
I think its fine. CLI should not really need anything complicated as this is not normally done.

kamilchodola (approved):

PR #8048, opened 2025-01-14
Add action for bumping version on release

### LINKED ISSUE #4794: [Versioning] Automate version of package based on Release workflow and tag provided + consider adding "preview" and "unstable"
As for now, we need to manually set a version before releasing. We should consider automating it together with our Release Workflow - version should be based on tag.

While running Release workflow, as a very first stage, we should take a TAG provided, update Directory.Build.props file and push changes to master and release branches. Then rest of procedure should be applied.

No alternatives for this process but I see some extensions to it.
We may want to add to additional informations to Version number:
1. "unstable" - when package/docker image/builded version is created from master branch
2. "preview" - when package/docker image/builded version is created from release branch

Also it would be nice to automate next-version number.
Example is:
We will release 1.14.4. Every new change above 14.4 tag and release should be 14.5-unstable. 
It would be great to automate that in some proper way as well.

stdevMac:
Fixes #4794

## Changes

- Added new GitHub Action workflow to automatically bump version on master after release branch creation
- The workflow triggers when a new `release/*` branch is created
- Automatically increments the minor version and adds `-unstable` suffix
- Creates a pull request instead of direct push to master for better review process
- Handles all version types correctly:
  - Regular releases (e.g., `release/1.31.0` -> `1.32.0-unstable`)
  - Major releases (e.g., `release/2.0.0` -> `2.1.0-unstable`)
  - Patch releases (e.g., `release/1.30.3` -> `1.31.0-unstable`)

- [ ] Optimization
- [x] Build-related changes
- [ ] Refactoring
- [ ] Documentation update
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

The workflow should be tested by:
1. Creating a test release branch
2. Verifying that a PR is created automatically
3. Checking that the version is bumped correctly in `Directory.Build.props`
4. Verifying the PR content and format

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks

This automation will help maintain consistent version numbering across releases and development branches, reducing manual intervention and potential human error in version management.


.github/workflows/bump-version.yml
```diff
@@ -0,0 +1,52 @@
...
+
+on:
+  create:
+    branches:
+      - 'release/*'
+
+jobs:
+  bump-version:
+    runs-on: ubuntu-latest
+    if: startsWith(github.ref, 'refs/heads/release/')
```
kamilchodola, line 11: Probably not needed condition as triggering condition does the same

.github/workflows/bump-version.yml
```diff
@@ -0,0 +1,52 @@
+name: Bump Version on Release Branch Creation
```
kamilchodola, line 1: Won't this workflow will have unnecessary runs when we will create for example release/1.30.4?

It is beneficial to run it only when a release branch with bumped minor is created.

.github/workflows/bump-version.yml
```diff
@@ -0,0 +1,52 @@
...
+jobs:
+  bump-version:
+    runs-on: ubuntu-latest
+    if: startsWith(github.ref, 'refs/heads/release/')
+    
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
kamilchodola, line 17: Shouldn't we specify ref: master here? We want to make a change on master branch but triggering condition here will probably checkout current ref not default ref (so in this case release branch not master).

.github/workflows/bump-version.yml
```diff
@@ -0,0 +1,52 @@
...
+          # Always increment minor version
+          NEW_MINOR=$((MINOR + 1))
+          NEW_VERSION="$MAJOR.$NEW_MINOR.0"
+          echo "new_version=$NEW_VERSION" >> $GITHUB_OUTPUT
+          
+          # Update version in Directory.Build.props
+          sed -i "s/<VersionPrefix>.*<\/VersionPrefix>/<VersionPrefix>$NEW_VERSION<\/VersionPrefix>/" src/Nethermind/Directory.Build.props
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v5
```
kamilchodola, line 38: Create PR only if version actually changes - there is a scenario where we will probably try to create PR to bump version to exact same one,

kamilchodola (approved):

rubo (approved):
- Minor revisions
- Created the label; otherwise, the creation of PR fails
- Tested

PR #8051, opened 2025-01-14
Fix dispose block.AccountChanges ArrayPoolList

Marchhill:
## Changes

- Dispose ArrayPoolList

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks

Fixes the following error observed on chiado node:
```
Unhandled exception. System.InvalidOperationException: ArrayPoolList hasn't been disposed. Created    at Nethermind.Core.Collections.ArrayPoolList`1..ctor(ArrayPool`1 arrayPool, Int32 capacity, Int32 startingCount)
   at Nethermind.Core.Collections.ArrayPoolList`1..ctor(Int32 capacity)
   at Nethermind.State.StateProvider.ChangedAddresses()
   at Nethermind.State.WorldState.Nethermind.State.IWorldState.GetAccountChanges()
   at Nethermind.State.WorldStateMetricsDecorator.GetAccountChanges()
   at Nethermind.Consensus.Processing.BlockProcessor.ProcessBlock(Block block, IBlockTracer blockTracer, ProcessingOptions options)
   at Nethermind.Consensus.AuRa.AuRaBlockProcessor.PostMergeProcessBlock(Block block, IBlockTracer blockTracer, ProcessingOptions options)
   at Nethermind.Merge.AuRa.AuRaMergeBlockProcessor.ProcessBlock(Block block, IBlockTracer blockTracer, ProcessingOptions options)
   at Nethermind.Consensus.Processing.BlockProcessor.ProcessOne(Block suggestedBlock, ProcessingOptions options, IBlockTracer blockTracer)
   at Nethermind.Consensus.Processing.BlockProcessor.Process(Hash256 newBranchStateRoot, List`1 suggestedBlocks, ProcessingOptions options, IBlockTracer blockTracer)
   at Nethermind.Consensus.Processing.BlockchainProcessor.ProcessBranch(ProcessingBranch& processingBranch, ProcessingOptions options, IBlockTracer tracer, String& error)
   at Nethermind.Consensus.Processing.BlockchainProcessor.Process(Block suggestedBlock, ProcessingOptions options, IBlockTracer tracer, String& error)
   at Nethermind.Consensus.Processing.BlockchainProcessor.RunProcessingLoop()
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.AsyncStateMachineBox`1.MoveNext(Thread threadPoolThread)
   at System.Threading.Tasks.Sources.ManualResetValueTaskSourceCore`1.SetResult(TResult result)
   at System.Threading.Channels.ChannelReader`1.ReadAllAsync(CancellationToken cancellationToken)+MoveNext()
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.AsyncStateMachineBox`1.MoveNext(Thread threadPoolThread)

   at Nethermind.Core.Collections.ArrayPoolList`1.Finalize() in /root/nethermind/src/Nethermind/Nethermind.Core/Collections/ArrayPoolList.cs:line 386
   at System.GC.RunFinalizers()
```


benaadams:
Is Disposed here in TxPool; so this would mean it was disposed before the txPool could use it

https://github.com/NethermindEth/nethermind/blob/b0c51654dcb8df33c000ad1b30cdbe78017ef85d/src/Nethermind/Nethermind.TxPool/TxPool.cs#L226-L238

However; OnHeadChange isn't always fired to the txpool for every block 🤔

MarekM25 (commented):
Could you try to write unit tests that reproduce it? Of course, if possible

Marchhill:
> Is Disposed here in TxPool; so this would mean it was disposed before the txPool could use it
> 
> https://github.com/NethermindEth/nethermind/blob/b0c51654dcb8df33c000ad1b30cdbe78017ef85d/src/Nethermind/Nethermind.TxPool/TxPool.cs#L226-L238
> 
> However; OnHeadChange isn't always fired to the txpool for every block 🤔

I think this the TxPool would use it before my code disposes it. I added just after the `Process` function.

The call stack would look something like this `Process` -> `_blockTree.UpdateMainChain` -> `MoveToMain` -> invoke `BlockReplacementEventArgs` -> TxPool `ProcessNewHeads`

So the change should just dispose if this is never called

Marchhill:
> Could you try to write unit tests that reproduce it? Of course, if possible

I'll have a look into it, happens for me just from starting the node on chiado

src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
```diff
@@ -285,6 +285,9 @@ private async Task RunProcessingLoop()
                 _stats.Start();
                 Block processedBlock = Process(block, blockRef.ProcessingOptions, _compositeBlockTracer.GetTracer(), out string? error);
 
+                block.AccountChanges?.Dispose();
+                processedBlock?.AccountChanges?.Dispose();
```
LukaszRozmej, line 289: I think this is used in `TxPool` so cannot be disposed here

LukaszRozmej (changes requested):
@benaadams ?

src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
Marchhill: see my above comment, I think that's before

Marchhill:
btw what is the advantage of using ArrayPoolList here anyway? @benaadams @LukaszRozmej

LukaszRozmej:
> Is Disposed here in TxPool; so this would mean it was disposed before the txPool could use it
> 
> https://github.com/NethermindEth/nethermind/blob/b0c51654dcb8df33c000ad1b30cdbe78017ef85d/src/Nethermind/Nethermind.TxPool/TxPool.cs#L226-L238
> 
> However; OnHeadChange isn't always fired to the txpool for every block 🤔

BlockAddedToMain is fired on every block

LukaszRozmej:
> > Is Disposed here in TxPool; so this would mean it was disposed before the txPool could use it
> > https://github.com/NethermindEth/nethermind/blob/b0c51654dcb8df33c000ad1b30cdbe78017ef85d/src/Nethermind/Nethermind.TxPool/TxPool.cs#L226-L238
> > 
> > However; OnHeadChange isn't always fired to the txpool for every block 🤔
> 
> I think this the TxPool would use it before my code disposes it. I added just after the `Process` function.
> 
> The call stack would look something like this `Process` -> `_blockTree.UpdateMainChain` -> `MoveToMain` -> invoke `BlockReplacementEventArgs` -> TxPool `ProcessNewHeads`
> 
> So the change should just dispose if this is never called

Can you add a test for it? :)

Marchhill:
> > Is Disposed here in TxPool; so this would mean it was disposed before the txPool could use it
> > https://github.com/NethermindEth/nethermind/blob/b0c51654dcb8df33c000ad1b30cdbe78017ef85d/src/Nethermind/Nethermind.TxPool/TxPool.cs#L226-L238
> > 
> > However; OnHeadChange isn't always fired to the txpool for every block 🤔
> 
> BlockAddedToMain is fired on every block

I think there are some cases qhwew this doesn't happen. If you look at `BlockchainProcessor.cs:395` we do this check before updating the main chain, for example in the case of a NewPayload we wouldn't update the main chain and the ArrayPoolList wouldn't be disposed

PR #8052, opened 2025-01-14
JsonRpc modules cancellation only on Debugger detached

Scooletz:
When debugging JsonRpc modules, including tests, the default configuration causes a timeout after 20s. This PR proposes to not timeout when a debugger is attached. This is done by providing an extension method for `IJsonRpcConfig` and then using it instead of building `CancellationTokenSource` from the `.Timeout` property. All call-sites that are amendable (no `BoundedModule`) are refactored.

## Changes

- extraction of a timeout method

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [x] Other: Ease debugging of hive tests & other that depend on the timeout

#### Requires testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.JsonRpc/JsonRpcConfigExtension.cs
+++ b/src/Nethermind/Nethermind.JsonRpc/JsonRpcConfigExtension.cs
@@ -2,7 +2,9 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 
 using System.Collections.Generic;
+using System.Diagnostics;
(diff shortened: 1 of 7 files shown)


LukaszRozmej (approved):

MarekM25 (approved):
Smart!

PR #8067, opened 2025-01-16
Override empty state scope with the state of the current block

flcl42:
## Changes

- Always override scope, applying block header
- Use readonly block tree in taiko?

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Tracing/GethStyleTracer.cs
+++ b/src/Nethermind/Nethermind.Consensus/Tracing/GethStyleTracer.cs
@@ -84,7 +84,7 @@ public class GethStyleTracer : IGethStyleTracer
         try
         {
             Dictionary<Address, AccountOverride>? stateOverride = options.StateOverrides;
-            using IOverridableTxProcessingScope? scope = stateOverride != null ? _env.BuildAndOverride(block.Header, stateOverride) : null;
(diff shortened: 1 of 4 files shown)


LukaszRozmej (approved):

src/Nethermind/Nethermind.Consensus/Tracing/GethStyleTracer.cs
```diff
@@ -84,7 +84,7 @@ public GethLikeTxTrace Trace(Hash256 blockHash, int txIndex, GethTraceOptions op
         try
         {
             Dictionary<Address, AccountOverride>? stateOverride = options.StateOverrides;
-            using IOverridableTxProcessingScope? scope = stateOverride != null ? _env.BuildAndOverride(block.Header, stateOverride) : null;
+            using IOverridableTxProcessingScope? scope = _env.BuildAndOverride(block.Header, stateOverride);
```
LukaszRozmej, line 87: why do we need it always?

flcl42: We have to always apply `block.Header`. As for the state override, if it's null, then it's just ignored inside `BuildAndOverride`, but header should be always used, otherwise we just calculated it above without purpose

PR #8070, opened 2025-01-17
Fix verifytrie does not block processing.

asdacap:
- Made a mistake when passing verify trie code into world state manager.
- The triestore passed in to verify trie must not be a read only triestore because that can't block.
- Also remove `ReadOnlyWorldStateManager`. 

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

#### Notes on testing

- Tested manually.


src/Nethermind/Nethermind.State/WorldStateManager.cs
```diff
@@ -10,19 +10,45 @@
...
+        _logManager = logManager;
+
+        IReadOnlyDbProvider readOnlyDbProvider = dbProvider.AsReadOnly(false);
+        _readaOnlyCodeCb = readOnlyDbProvider.GetDb<IDb>(DbNames.Code).AsReadOnly(true);
+        GlobalStateReader = new StateReader(_readOnlyTrieStore, _readaOnlyCodeCb, _logManager);
+
+        if (processExitSource is not null)
+        {
+            _blockingVerifyTrie = new BlockingVerifyTrie(trieStore, GlobalStateReader, _readaOnlyCodeCb!, processExitSource!, logManager);
+        }
```
LukaszRozmej, line 49: It is bit weird that it depends on if exitsource is passed

asdacap: Yea, it should have been just one `void VerifyTrie(CancellationToken cancellation)` and the exit source is handled outside.

asdacap: Moved the exit source logic out.

LukaszRozmej (approved):

damian-orzechowski (approved):

PR #8076, opened 2025-01-20
Make sure prewarming and pruning trie store is running on ethereum tests

asdacap:
- Double check that ethereum tests passed with prewarmer and pruning trie store.
- Aside from needing to explicitly clear cache, seems to work fine.

#### Requires testing

- [X] No?

- [X] Yes

### DIFF
--- a/src/Nethermind/Ethereum.Test.Base/BlockchainTestBase.cs
+++ b/src/Nethermind/Ethereum.Test.Base/BlockchainTestBase.cs
@@ -8,13 +8,11 @@ using System.Linq;
 using System.Numerics;
 using System.Threading;
 using System.Threading.Tasks;
+using Autofac;
 using Nethermind.Blockchain;
-using Nethermind.Blockchain.BeaconBlockRoot;
-using Nethermind.Blockchain.Blocks;
 using Nethermind.Blockchain.Find;
-using Nethermind.Blockchain.Receipts;
+using Nethermind.Config;
 using Nethermind.Consensus;
-using Nethermind.Consensus.Comparers;
 using Nethermind.Consensus.Ethash;
 using Nethermind.Consensus.Processing;
 using Nethermind.Consensus.Rewards;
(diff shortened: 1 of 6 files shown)


src/Nethermind/Nethermind.State/IWorldState.cs
```diff
@@ -104,6 +104,8 @@ public interface IWorldState : IJournal<Snapshot>, IReadOnlyStateProvider
 
     void DecrementNonce(Address address) => DecrementNonce(address, UInt256.One);
 
+    void SetNonce(Address address, in UInt256 nonce);
```
LukaszRozmej, line 107: This is not a normal use-case method when interacting with `WorldState`. Might be useful on some setups, but not sure if should be just public on interface.

src/Nethermind/Nethermind.Core.Test/Modules/MainBlockProcessingContext.cs
```diff
@@ -19,7 +21,9 @@ namespace Nethermind.Core.Test.Modules;
 public record MainBlockProcessingContext(
     ILifetimeScope LifetimeScope,
     BlockchainProcessor BlockchainProcessor,
+    IWorldState WorldState,
     IBlockProcessor BlockProcessor,
+    ITransactionProcessor TransactionProcessor,
```
LukaszRozmej, line 26: Added parameters are not used?

src/Nethermind/Ethereum.Test.Base/BlockchainTestBase.cs
```diff
@@ -247,6 +204,13 @@ protected async Task<EthereumTestResult> RunTest(BlockchainTest test, Stopwatch?
         await blockchainProcessor.StopAsync(true);
         stopwatch?.Stop();
 
+        IBlockCachePreWarmer? preWarmer = container.Resolve<MainBlockProcessingContext>().LifetimeScope.ResolveOptional<IBlockCachePreWarmer>();
```
LukaszRozmej, line 207: Not sure if, but does this create any issue with parallel test invocation? If not can you be more specific about it? I think maybe `LifetimeScope` solves this but it isn't totally clear for me.

asdacap: It does not? Each invocation creates a completely different instance.

PR #8081, opened 2025-01-20
Distinguish by fields

### LINKED ISSUE #7987: 'Invalid Params' Error on Nethermind for Type-3 Transactions
**Description**

After the Cancun upgrade, Ethereum introduced Type-3 transactions, which can store “blob” data with significantly reduced gas costs. While eth_estimateGas should return the correct gas usage for a Type-3 transaction, it currently returns {'code': -32602, 'message': 'Invalid params'}.

**Steps to Reproduce**

To reproduce this issue, follow these steps:
1. Run a testnet or use an existing mainnet node provided by `Nethermind`.
2. Execute the script in the appendix with the appropriate environment configuration (i.e., `TESTNET_RPC_URL=<your-nethermind-node-rpc-url> python3 estimate.py`).
3. Observe that the script returns `{'code': -32602, 'message': 'Invalid params'}`.

**Actual behavior**

The error returned is {'code': -32602, 'message': 'Invalid params'}. After reviewing the following code:
https://github.com/NethermindEth/nethermind/blob/1a1e6025c3c23fca13d830d188c1f2207b06b6be/src/Nethermind/Nethermind.JsonRpc/JsonRpcService.cs#L154-L162
I suspect that the RPC module does not yet support Type-3 transactions, as suggested by [this snippet](https://github.com/NethermindEth/nethermind/blob/1a1e6025c3c23fca13d830d188c1f2207b06b6be/src/Nethermind/Nethermind.JsonRpc/Modules/Eth/SimulateTxExecutor.cs#L71-L90).

**Expected behavior**

Switching the `TESTNET_RPC_URL` to a endpoint of `geth`, `reth`, `besu`, or `erigon` allows the script to successfully return the estimated gas.

 - Operating System: Linux
 - Version: v1.30.1
 - Installation Method: Using Kurtosis to run testnet
 - Consensus Client: Lighthouse

**Appendix**

The python script (i.e., `estimate.py` used for poc):
```
#!/usr/bin/env python3

import os
import hashlib
import json
import traceback
from web3 import Web3, HTTPProvider
from eth_account import Account

def generate_demo_blob(blob_index: int, size_kb: int = 128) -> bytes:
    """
    Generates a demo blob of specified size filled with deterministic data for consistency.
    
    Args:
        blob_index (int): Index of the blob for pattern variation.
        size_kb (int): Size of the blob in kilobytes (default is 128 KB).
        
    Returns:
        bytes: Generated blob data.
    """
    # Each blob is size_kb KB
    pattern = (f"Blob{blob_index}Data").encode()
    blob_size = size_kb * 1024  # Convert KB to bytes
    blob = (pattern * (blob_size // len(pattern) + 1))[:blob_size]
    return blob

def compute_blob_hash(blob, version=1):
    sha256_hash = hashlib.sha256(blob).digest()
    versioned_hash = bytes([0x01]) + sha256_hash[1:]
    return '0x' + versioned_hash.hex()

def to_hex(value):
    """
    Converts an integer value to a hex string with '0x' prefix.
    
    Args:
        value (int): The integer value to convert.
        
    Returns:
        str: Hexadecimal string representation with '0x' prefix.
    """
    return hex(value)

def main():
    # Initialize Web3
    rpc_url = os.getenv("TESTNET_RPC_URL")
    if not rpc_url:
        raise EnvironmentError("TESTNET_RPC_URL environment variable not set.")
    w3 = Web3(HTTPProvider(rpc_url))
    assert w3.is_connected(), "Web3 provider is not connected."

    # Load account from private key
    private_key = os.getenv("PRIVATE_KEY", "0xbf3beef3bd999ba9f2451e06936f0423cd62b815c9233dd3bc90f7e02a1e8673")
    account = Account.from_key(private_key)

    # Get the nonce for the account
    nonce = w3.eth.get_transaction_count(account.address)

    # Number of blobs to include (1 blob of 128 KB)
    num_blobs = 1
    blob_size_kb = 128  # 128 KB per blob

    # Generate demo blobs and their hashes
    blobs = []
    blob_versioned_hashes = []
    for i in range(num_blobs):
        blob = generate_demo_blob(i, size_kb=blob_size_kb)
        blobs.append('0x' + blob.hex())  # Blob as hex string with '0x' prefix
        blob_hash = compute_blob_hash(blob)
        blob_versioned_hashes.append(blob_hash)

    # Validate blob sizes
    for idx, blob in enumerate(blobs):
        blob_bytes = bytes.fromhex(blob[2:])  # Remove '0x' prefix
        expected_size = blob_size_kb * 1024
        actual_size = len(blob_bytes)
        assert actual_size == expected_size, f"Blob {idx + 1} size is incorrect: {actual_size} bytes, expected {expected_size} bytes."

    # Validate blob hashes sizes
    for idx, hash_str in enumerate(blob_versioned_hashes):
        hash_hex = hash_str[2:]  # Remove '0x' prefix
        assert len(hash_hex) == 64, f"Blob {idx + 1} hash length is incorrect: {len(hash_hex)} characters, expected 64 characters."

    # Construct the Type-3 transaction with hexadecimal fields
    tx = {
        "type": 3,  # Type-3 Transaction
        "chainId": to_hex(w3.eth.chain_id),
        "nonce": to_hex(nonce),
        "from": account.address,  # Optional when signing
        "to": "0x0000000000000000000000000000000000000000",
        "value": to_hex(0),
        "maxPriorityFeePerGas": to_hex(w3.to_wei("2", "gwei")),
        "maxFeePerGas": to_hex(w3.to_wei("30", "gwei")),
        "maxFeePerBlobGas": to_hex(w3.to_wei("100", "gwei")),
        "gas": to_hex(1500000),
        "data": "0x",
        "blobVersionedHashes": blob_versioned_hashes,  # List containing one 32-byte hash
        "blobs": blobs,  # List containing one 128 KB blob as hex string
    }

    # Estimate Gas
    try:
        estimated_gas = w3.eth.estimate_gas(tx)
        print(f"Estimated Gas: {estimated_gas}")
    except Exception as e:
        print("Error estimating gas:", e)
        traceback.print_exc()

if __name__ == "__main__":
    main()
```

flcl42:
Fixes #7987

## Changes

- Detect tx type if certain fields present

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

@emlautarom1 do we need to handle OP deposit txs in RPC?


emlautarom1:
Directly from the OP team:

```
No, deposit-txs are solely accepted through the L1, by making a call to the optimism portal contract (this may be indirect, if you use higher level abstractions of the bridge contracts). The deposit is then authenticated by the L1 change (an event). The deposit tx itself does not have a signature, and cannot go into the tx-pool.
```

In summary, **no**: deposit transactions are not handled at the RPC level.

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
```diff
@@ -76,55 +81,45 @@ public JsonConverter()
             RegisterTransactionType<SetCodeTransactionForRpc>();
         }
 
-        internal static void RegisterTransactionType<T>() where T : TransactionForRpc, ITxTyped
-            => _types[(int)T.TxType] = typeof(T);
+        internal static void RegisterTransactionType<T>() where T : TransactionForRpc, IFromTransaction<T>, ITxTyped
+        {
+            Type txType = typeof(T);
+            _fromTransactionFuncs[(byte)T.TxType] = T.FromTransaction;
+            string[] uniqueProperties = txType.GetProperties(BindingFlags.Public | BindingFlags.Instance | BindingFlags.DeclaredOnly).Where(p => p.GetCustomAttribute<JsonDiscriminatorAttribute>() is not null).Select(p => p.Name).ToArray();
```
MarekM25, line 88: I am wondering if we can make it simpler, it's hard for me to understand how the the type is determined now based on fields. Let's discuss

MarekM25 (changes requested):
Let's add some simple tests

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/EIP1559TransactionForRpc.cs
```diff
@@ -15,9 +15,11 @@ public class EIP1559TransactionForRpc : AccessListTransactionForRpc, IFromTransa
 
     public override TxType? Type => TxType.EIP1559;
 
+    [JsonDiscriminator]
```
MarekM25, line 18: Discrimnator is a weird name

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
```diff
@@ -134,19 +129,21 @@ public static TransactionForRpc FromTransaction(Transaction tx, TransactionConve
...
             ChainId = chainId,
             TxIndex = txIndex,
             BlockHash = blockHash,
             BlockNumber = blockNumber,
             BaseFee = baseFee
         });
+
+    public static void RegisterTransactionType<T>() where T : TransactionForRpc, IFromTransaction<T>, ITxTyped => TransactionJsonConverter.RegisterTransactionType<T>();
+}
+
```
MarekM25, line 144: add a short comment?

MarekM25 (approved):

LukaszRozmej (commented):
I am missing why is it better to differentiate by properties rather than by explicit type?

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
```diff
@@ -76,55 +85,45 @@ public JsonConverter()
...
+            JsonObject untyped = JsonSerializer.Deserialize<JsonObject>(ref txTypeReader, options);
 
-            return _types.TryGetByTxType(discriminator, out Type concreteTxType)
+            return TryGetTxType(_requiredProps, untyped, out Type? concreteTxType)
                 ? (TransactionForRpc?)JsonSerializer.Deserialize(ref reader, concreteTxType, options)
                 : throw new JsonException("Unknown transaction type");
+
+            static bool TryGetTxType(OrderedDictionary<Type, string[]> props, JsonObject untyped, [NotNullWhen(true)] out Type? type)
+            {
+                type = props.FirstOrDefault(p => p.Value.Any(name => untyped.ContainsKey(name)), props.Last()).Key;
```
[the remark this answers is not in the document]
flcl42: If so we can mark only the different fields as discriminators

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
```diff
@@ -61,13 +66,17 @@ public virtual Transaction ToTransaction()
     public abstract void EnsureDefaults(long? gasCap);
     public abstract bool ShouldSetBaseFee();
 
-    internal class JsonConverter : JsonConverter<TransactionForRpc>
+    internal class TransactionJsonConverter : JsonConverter<TransactionForRpc>
     {
-        // NOTE: Should we default to a specific TxType?
-        private const TxType DefaultTxType = TxType.Legacy;
-        private static readonly Type[] _types = new Type[Transaction.MaxTxType + 1];
+        private static readonly OrderedDictionary<Type, string[]> _requiredProps = [];
```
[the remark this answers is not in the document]
flcl42: > shouldn't this be the other way around, from required properties to type?
may be a list of tuples I guess. Let me add an internal type

> Could potentially be FrozenDictionary
It can be changed from plugins.

flcl42:
> I am missing why is it better to differentiate by properties rather than by explicit type?

Because [agreed spec's type](https://github.com/ethereum/execution-apis/blob/main/src/schemas/transaction.yaml#L488) is not a required field and a lot of libraries simply does not pass it

LukaszRozmej:
> > I am missing why is it better to differentiate by properties rather than by explicit type?
> 
> Because [agreed spec's type](https://github.com/ethereum/execution-apis/blob/main/src/schemas/transaction.yaml#L488) is not a required field and a lot of libraries simply do not pass it

Ok but it is in the spec here: https://github.com/ethereum/execution-apis/blob/main/src/schemas/transaction.yaml#L493-L495
So should the logic be:
1. If there is tx type passed - use it explicitly
2. If there is none type passed - then infer it from what fields were passed?

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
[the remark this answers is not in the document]
LukaszRozmej: Sure it can be changed on startup - just rebuild FrozenDictionary.
After startup it shouldn't be changed, and lookup in the dictionary will be faster for the whole lifetime of the app.

flcl42:
> > > I am missing why is it better to differentiate by properties rather than by explicit type?
> > 
> > 
> > Because [agreed spec's type](https://github.com/ethereum/execution-apis/blob/main/src/schemas/transaction.yaml#L488) is not a required field and a lot of libraries simply do not pass it
> 
> Ok but it is in the spec here: https://github.com/ethereum/execution-apis/blob/main/src/schemas/transaction.yaml#L493-L495 So should the logic be:
> 
>     1. If there is tx type passed - use it explicitly
> 
>     2. If there is none type passed - then infer it from what fields were passed?

Yes, it's indeed in the spec but is ignored by geth. But seems like it's useful to take it into account, so let me check it too. Fixed

emlautarom1 (commented):
I think we're being overly clever on trying to figure out which transaction type we're dealing with, but it also seems like geth is doing the same thing (see https://github.com/ethereum/go-ethereum/blob/a840e9b59f28427d3c65de9f028853e075438201/internal/ethapi/transaction_args.go#L477)

Not a fan of this, and it seems like having a single `GenericTransaction` request type separated from the response types would avoid some challenges (it was an alternative design proposed during the `TransactionForRpc` refactor, see https://github.com/NethermindEth/nethermind/pull/7446#issue-2529488213) yet I don't think it would make a radical difference.

Proposed solution works but I would expect some long term maintenance cost and most likely some bugs will show up.

I would add some tests for the initiating issue with Blob transactions and then it should be good to merge.

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
```diff
@@ -295,6 +295,11 @@ private TransactionResult CallAndRestore(
                 callHeader.ExcessBlobGas = treatBlockHeaderAsParentBlock
                     ? BlobGasCalculator.CalculateExcessBlobGas(blockHeader, releaseSpec)
                     : blockHeader.ExcessBlobGas;
+
+                if (transaction.Type is TxType.Blob && transaction.MaxFeePerBlobGas is null && BlobGasCalculator.TryCalculateFeePerBlobGas(callHeader, out UInt256 blobBaseFee))
```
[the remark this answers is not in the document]
flcl42: Added

src/Nethermind/Nethermind.JsonRpc.Test/Data/Eip2930Tests.cs
```diff
@@ -219,17 +219,6 @@ public void can_deserialize_not_provided_txType()
         transactionForRpc.Type.Should().Be(TxType.Legacy);
     }
 
-    [Test]
-    public void can_deserialize_direct_null_txType()
```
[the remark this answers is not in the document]
flcl42: Added in a different way

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
```diff
@@ -61,13 +65,16 @@ public virtual Transaction ToTransaction()
     public abstract void EnsureDefaults(long? gasCap);
     public abstract bool ShouldSetBaseFee();
 
-    internal class JsonConverter : JsonConverter<TransactionForRpc>
+    internal class TransactionJsonConverter : JsonConverter<TransactionForRpc>
     {
-        // NOTE: Should we default to a specific TxType?
-        private const TxType DefaultTxType = TxType.Legacy;
-        private static readonly Type[] _types = new Type[Transaction.MaxTxType + 1];
+        private static readonly List<TxTypeInfo> _txTypes = [];
```
LukaszRozmej, line 70: Why not allocate static, never growing array with: 
`private static readonly TxTypeInfo[] _types = new TxTypeInfo[Transaction.MaxTxType + 1];`
So we will have tx type mapping from index

LukaszRozmej (approved):
Change internal structure to keep it index by tx type, rest is fine

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
flcl42: Because index there does not mean tx type any more, we check tx fields of tx type in ascending order. Such list will take less place also

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
```diff
@@ -76,77 +83,95 @@ public JsonConverter()
...
+            {
+                txType = JsonSerializer.Deserialize<TxType?>(node, options);
+            }
 
-            TxType discriminator = (TxType)(type.Type ?? (ulong)DefaultTxType);
+            Type concreteTxType =
+                (txType is not null
+                ? _txTypes.FirstOrDefault(p => p.TxType == txType)
+                : _txTypes.FirstOrDefault(p => p.DiscriminatorProperties.Any(name => untyped.ContainsKey(name)), _txTypes.Last()))?.Type
+                ?? throw new JsonException("Unknown transaction type");
```
LukaszRozmej, line 132: For readability:
```suggestion
            Type concreteTxType =
                (
                    txType is not null
                        ? _txTypes.FirstOrDefault(p => p.TxType == txType)
                        : _txTypes.FirstOrDefault(p => p.DiscriminatorProperties.Any(name => untyped.ContainsKey(name)), _txTypes[^1])
                )?.Type
                ?? throw new JsonException("Unknown transaction type");
```

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
```diff
@@ -76,77 +83,95 @@ public JsonConverter()
...
-            // The first parse extract the `Type` while the second parses the entire Transaction
+            // The first parse is used to check for fields, while the second parses the entire Transaction
             Utf8JsonReader txTypeReader = reader;
-            TransactionType type = JsonSerializer.Deserialize<TransactionType>(ref txTypeReader, options);
+            JsonObject untyped = JsonSerializer.Deserialize<JsonObject>(ref txTypeReader, options);
+
+            TxType? txType = null;
+            if (untyped.TryGetPropertyValue(TypeFieldKey, out JsonNode? node))
+            {
+                txType = JsonSerializer.Deserialize<TxType?>(node, options);
```
LukaszRozmej, line 125:
```suggestion
                txType = node.Deserialize<TxType?>(options);
```

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
```diff
@@ -76,77 +83,95 @@ public JsonConverter()
...
+
+            if (existingTypeInfo != -1)
+            {
+                // Adding in reverse order so new tx type are in priority
+                _txTypes[existingTypeInfo] = typeInfo;
+            }
+            else
+            {
+                _txTypes.Insert(0, typeInfo);
+            }
```
LukaszRozmej, line 111: I think we shouldn't depend on addition order, but tx type order. Then we need that list kept sorted. Either sort it after adding or use linked list and inserting while keeping it sorted.

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
LukaszRozmej: I actually think the tx type order is what is important not inserting order!
We can trim the array to smaller list if we don't want excessive items.

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
flcl42: > Properties marked with <see cref="JsonDiscriminatorAttribute" /> are checked in reverse order of tx type registration.

Looks like it's not enough to mention this, and others will assume different behavior. No problem I will change it.

emlautarom1 (approved):

PR #8082, opened 2025-01-20
Log indexing

Scooletz:
This PR speculatively proposes a new way of capturing logs and querying them. It does it by introducing a `LogBuilder` that is responsible for capturing the logs as they appear and later build them into an immutable file. To make it efficient, each address/topic is hashes using `XXHash64`. This is a fast non-cryptographic hash based on 64 bit ulong. This, due to the birthday paradox, can suffer from collisions after breaching `4 billion` of topics (`log_2 (64)`). This should be unlikely if we limit the range of blocks put in one file.

### Entries

Each `LogEntry` is indexed by its address and all the topics. To distinguish the topic by position, different seeds are used for hashing. This impacts the collision probability slightly. The block number and a transaction number are encoded under ulong->uint mapping. While `ulong` represents the hash, `uint` is used to encode tuple of `(block, tx)`. This gives 12 bytes of storage for each topic/address of a log entry.

### Writing and deduplication

When a builder is flushed to a `IBufferWriter<byte>`, first it has its entries sorted by their hashes. To make the lookup faster, key/values are kept as separate arrays. After sorting, if a key (topic/address) appears multiple times, it's encoded in a different way. First all the corresponding entries are sorted to later be encoded using diff encoding with a varint. They are written to the output buffer that is sealed by a special entry at the that points to the beginning of the sequence. The address of such entry is the thing that is mapped to the topic in a given file. This allows to encode frequently occurring topics in a very efficient way without sacrificing the case of a unique address/topic. The file is sealed by writing a single `int` that represents how many topics there are in it. The rest, can be derived.

### Block+tx vs block

Tests show that encoding block numbers only greatly reduce the size of the data. The question is what do we need from the logging. If it's block number + tx number, then the index is bigger. For block only numbers, we can squeeze and compress much more.

### IO

```shell
                0x0000 0x0001 0x0002 0x0003   ...                0xFFFE 0xFFFF                                            
               ┌──────┬──────┬──────┬──────┬──────┬──────┬──────┬──────┬──────┐                                           
               │      │      │      │      │      │      │      │      │      │                                           
    prefixes   │      │      │      │   ─┐ │ ...  │      │      │      │      │                                           
               │      │      │      │    │ │      │      │      │      │      │                                           
               └──────┴──────┴──────┴────┼─┴──────┴──────┴──────┴──────┴──────┘                                           
                                         │                                                                                
                                         │                                                                                
               ┌───┬───┬────┬───┬────┬───▼───┬───────┬───────────────────────────────────────────────────────────────────┐
               │   │   │    │   │    │       │       │                                                                   │
hashes & ptrs  │   │   │    │   │    │   │   │       │    ...                                                            │
               │   │   │    │   │    │   │   │       │                                                                   │
               └───┴───┴────┴───┴────┴───┼───┴───────┼───────────────────────────────────────────────────────────────────┘
                                         │                                                                                
                                         │                                                                                
                                         │                                                                                
                                         ▼                                                                                
               ┌────────────────────────┌───────────────────────────────┐────────────────────────────────────────────────┐
 compressed    │                        │                               │                                                │
               │                        │                               │                                                │
               └────────────────────────└───────────────────────────────┘────────────────────────────────────────────────┘
```

To check whether there are blocks that contain a given address/topic+index, we need to perform 2 IO operations:

1. read 8 bytes of `prefixes`
2. read ~1kb of `hashes`

To enumerate all the occurrences of a given address/topic+index, we need to perform 2 or at least 3 IO operations:

1. read 8 bytes of `prefixes`
2. read ~1kb of `hashes`
3. if topic/address occurred more than once, at least 1 more read. Reads are amortized and fetch 4kb in one go. For heavy topics, one occurrence will take 1 byte.

#### Block + tx number

image

#### Block number only

image

### Merging and querying

As a single builder can encode up to a 20-50 thousands blocks (probably not possible due to the memory), merging of files must be implemented. As they are ordered both by keys as well as entries, it can be thought of as merging two sorted enumerables (sorting).

### GC

If the builder builds files in block ranges, like 30-50blocks at a time, files could be named using the starting block number. Then GC would remove the oldest files.

### Issues

The diff encoding may be heavy to search through. If needed, when grouping 64k blocks is not enough, different encoding or a skip list can be added.

## Changes

- _List the changes_

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Remarks


Scooletz:
Just to compare:

- 1 billion of entries (250 of addresses and 750 of topics) takes 12GB now with the other approach but can be greatly reduced
- with this maximum size would be 12GBs as well, but it strongly depends on the distribution. The bottom line would be 2GB (with same topic repeated over and over) but this is unreasonable and should be thought of only as the lowest possible one

Issue #6481, opened 2024-01-08
[Old Bodies] On already synced node it switches to FastBodies and stays there for about 1 minute without progress


kamilchodola:
**Steps to Reproduce**
In order to replicate the behavior, please provide a detailed list of steps:
1. Have synced node (on older version)
2. Restart it on 1.25.0
3. Monitor logs

**Actual behavior**
These logs appears:
```
08 Jan 14:53:31 | Sync mode changed from DbLoad to WaitingForBlock, FastBodies
08 Jan 14:53:31 | SyncDispatcher<BodiesSyncBatch> has finished work.
08 Jan 14:53:31 | Fast blocks bodies task completed.
08 Jan 14:53:32 | Changing state WaitingForBlock, FastBodies to WaitingForBlock, FastReceipts at processed: 18962940 | state: 18962940 | block: 18963005 | header: 18963005 | target block: 18963007 | peer block: 18963007
08 Jan 14:53:32 | Changed pivot in receipts sync. Now using pivot 18526686 and barrier 1
08 Jan 14:53:33 | Peers | with best block: 5 | all: 5 | eth66 (20 %), eth68 (80 %) | Active: None | Sleeping: All | Geth (40 %), Nethermind (20 %), Erigon (20 %), Unknown (20 %)
08 Jan 14:53:33 | Old Bodies   18,526,686 / 18,526,686 (100.00 %) | queue         0 | current            0 Blk/s | total            0 Blk/s
08 Jan 14:53:38 | Old Bodies   18,526,686 / 18,526,686 (100.00 %) | queue         0 | current            0 Blk/s | total            0 Blk/s
08 Jan 14:53:43 | Old Bodies   18,526,686 / 18,526,686 (100.00 %) | queue         0 | current            0 Blk/s | total            0 Blk/s
08 Jan 14:53:48 | Old Bodies   18,526,686 / 18,526,686 (100.00 %) | queue         0 | current            0 Blk/s | total            0 Blk/s
08 Jan 14:53:53 | Old Bodies   18,526,686 / 18,526,686 (100.00 %) | queue         0 | current            0 Blk/s | total            0 Blk/s
08 Jan 14:53:58 | Old Bodies   18,526,686 / 18,526,686 (100.00 %) | queue         0 | current            0 Blk/s | total            0 Blk/s
08 Jan 14:54:01 | SyncDispatcher<ReceiptsSyncBatch> has finished work.
```

**Expected behavior**
[...]

PR #8083, opened 2025-01-21
Fix/fast header initialization

asdacap:
Fixes Closes Resolves #

- When `lowestInsertedHeader` is loaded at initialization, its total difficulty is not set, leading to the underflow error, unless it completely finish fast header.
- The error in `Initialize` is not apparent as it is called async.

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Testing
- Need to double check what causes #6481


src/Nethermind/Nethermind.Synchronization/FastBlocks/FastHeadersSyncFeed.cs
```diff
@@ -201,8 +201,25 @@ protected virtual void ResetPivot()
...
+                if (lowestInserted.TotalDifficulty is null)
+                {
+                    // When the LowestInsertedHeader is set in blockTree initializer, its TD is not set from block info.
+                    // So here we explicitly try to fetch it again.
+                    lowestInserted = _blockTree.FindHeader(lowestInserted.Number, BlockTreeLookupOptions.RequireCanonical);
+
+                    // In case of some strange corruption, we will have to reset the whole sync.
+                    if (lowestInserted!.TotalDifficulty is null)
+                    {
+                        _logger.Warn($"Missing total difficulty on lowest inserted header: {lowestInserted!.ToString(BlockHeader.Format.Short)}. Resetting header sync.");
```
LukaszRozmej, line 213: IsWarn

LukaszRozmej (approved):

src/Nethermind/Nethermind.Synchronization/FastBlocks/FastHeadersSyncFeed.cs
```diff
@@ -201,8 +201,25 @@ protected virtual void ResetPivot()
             BlockHeader? lowestInserted = _blockTree.LowestInsertedHeader;
             if (lowestInserted is not null && lowestInserted!.Number < _pivotNumber)
             {
-                SetExpectedNextHeaderToParent(lowestInserted);
-                _lowestRequestedHeaderNumber = lowestInserted.Number;
+                if (lowestInserted.TotalDifficulty is null)
```
Scooletz, line 204: So this is the crux, that the `lowestInserted` can be loaded without the `TotalDifficulty`. Why do we load it though this way. Would it not be easier to handle it where it's loaded and pass requirements https://github.com/NethermindEth/nethermind/blob/4fa9b3ffeeafd6f4715780348f02da37ec4bdc5a/src/Nethermind/Nethermind.Blockchain/BlockTree.Initializer.cs#L133 so that the `lowestInserted` is properly preloaded??

Scooletz (commented):
A few minor questions. Overall ok. Would be nice to have the answers to understand the setup fully.

src/Nethermind/Nethermind.Synchronization/FastBlocks/FastHeadersSyncFeed.cs
asdacap: Heck, I don't know. Strictly speaking, only the hash is stored.

PR #8090, opened 2025-01-22
Perf/directly specify StepsInfo

asdacap:
- Directly specify `StepsInfo` instead of iterating types in assembly.
- `IInitializationPlugin` explicitly declare `StepsInfo` with `GetSteps`. 
- I guess it make startup marginally faster, but more importantly initialization plugins are more explicit. 
- Double check by scanning assembly in unit tests. 

## Types of changes

- [X] Optimization
- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Testing


LukaszRozmej (approved):
A bit mixed feelings. Pro: It is explicit and more discoverable. Cons: I do like convention based registration and that you don't need to do it in multiple places.

But overall it is fine and we don't overload steps that often.

src/Nethermind/Nethermind.Optimism/OptimismPlugin.cs
```diff
@@ -61,6 +63,13 @@ public class OptimismPlugin(ChainSpec chainSpec) : IConsensusPlugin, ISynchroniz
     private OptimismCL? _cl;
     public bool Enabled => chainSpec.SealEngineType == SealEngineType;
 
+    public IEnumerable<StepInfo> GetSteps()
+    {
+        yield return new StepInfo(typeof(InitializeBlockchainOptimism));
+        yield return new StepInfo(typeof(InitializeBlockProducerOptimism));
+        yield return new StepInfo(typeof(RegisterOptimismRpcModules));
+    }
```
LukaszRozmej, line 71: Minor syntatic suggar add implicit casting from `Type` to `StepInfo` and change this to 
```suggestion
        yield return typeof(InitializeBlockchainOptimism);
        yield return typeof(InitializeBlockProducerOptimism);
        yield return typeof(RegisterOptimismRpcModules);
    }
```

src/Nethermind/Nethermind.Api/Steps/StepInfo.cs
```diff
@@ -4,19 +4,19 @@
...
+        public StepInfo(Type type)
         {
             if (type.IsAbstract)
             {
                 throw new ArgumentException("Step type cannot be abstract", nameof(type));
             }
 
             StepType = type;
-            StepBaseType = baseType;
+            StepBaseType = GetStepBaseType(type);
```
LukaszRozmej, line 19: I think we need to check if `Type` is `IsStepType` and throw if not?

src/Nethermind/Nethermind.Optimism/OptimismPlugin.cs
asdacap: Done

PR #8097, opened 2025-01-23
Devnet 6 change for 7702

ak88:
Change is described here:
https://github.com/ethereum/EIPs/pull/9248/files

All EXTCODE op codes will now act on the full designator instead of just the first two bytes.


smartprogrammer93 (approved):
LGTM

LukaszRozmej (approved):

PR #8098, opened 2025-01-24
Feature/end to end discovery test

asdacap:
- Create 5 node, assign 1 of them as boot nodes for other.
- Start discovery, then check if they can find all other nodes.
- Dicv5 does not work. Something with lantern session authentication handling keep failing on multiple concurrent requests.

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Seems to work.

### DIFF
--- a/src/Nethermind/Nethermind.Core.Test/Modules/InsecureProtectedPrivateKey.cs
+++ b/src/Nethermind/Nethermind.Core.Test/Modules/InsecureProtectedPrivateKey.cs
@@ -1,0 +1,17 @@
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using Nethermind.Core.Crypto;
+using Nethermind.Crypto;
+
+namespace Nethermind.Core.Test.Modules;
+
+public class InsecureProtectedPrivateKey(PrivateKey privateKey) : IProtectedPrivateKey
+{
+    public PublicKey PublicKey => privateKey.PublicKey;
+    public CompressedPublicKey CompressedPublicKey => privateKey.CompressedPublicKey;
+    public PrivateKey Unprotect()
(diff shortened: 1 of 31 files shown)


src/Nethermind/Nethermind.Core.Test/Modules/InsecureProtectedPrivateKey.cs
```diff
@@ -0,0 +1,17 @@
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using Nethermind.Core.Crypto;
+using Nethermind.Crypto;
+
+namespace Nethermind.Core.Test.Modules;
+
+public class InsecureProtectedPrivateKey(PrivateKey privateKey) : IProtectedPrivateKey
```
LukaszRozmej, line 9: NotProtetedPriveteKey :D

src/Nethermind/Nethermind.Init/Steps/InitializeNetwork.cs
```diff
@@ -257,8 +257,6 @@ private void InitDiscovery()
             _api.LogManager, _api.Timestamper, _api.CryptoRandom,
             _api.NodeStatsManager, _api.IpResolver
         );
-
-        _api.DiscoveryApp.Initialize(_api.NodeKey.PublicKey);
```
LukaszRozmej, line 261: This is moved somewhere esle?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Init/Steps/InitializeNetwork.cs
asdacap: Moved to CompositeDiscoveryApp's constructor.

PR #6227, opened 2023-10-26
Perf/NoWAL during OldBodies

asdacap:
- Most of the write IO during old bodies are actually due to WAL (Write Ahead Log) file, which is the mechanism for which rocksdb recover on unclean shutdown.
- This PR disable WAL writes for bodies writes from OldBodies.
- For recovery, the LowestInsertedBlockNumber is not updated on all request, instead it is updated every 100000 blocks, at which point an explicit flush is also triggered to make sure the memtables are written at that point.
- This reduces total writes during OldBodies by about 60-70% or so. This is more than 50% because the WAL file are not compressed and the bodies are compressed to about 60-70% on flush.
- No change in bodies sync time. Unless your SSD can't sustain about 350MB/s of writes before and you have really fast internet.
- Graph is after, before, after, before.
Screenshot_2023-10-26_14-07-00

## Changes

- Add WriteFlags to PutSpan.
- Add WriteFlags to BlockTree.Insert
- Set nowal during old bodies.

#### What types of changes does your code introduce?

- [X] Optimization

#### Requires testing

- [X] Yes

#### If yes, did you write tests?

- [X] Yes

#### Notes on testing

- Manually `kill -s KILL` nethermind a couple of time during sync. Then run a custom python script to verify all expected blocks is present. Verified, it resumed slightly before the point it was killed

PR #8103, opened 2025-01-24
FastHeadersSyncFeed explicitly flush before setting metadata

### LINKED ISSUE #8029: Chiado node after several early restarts crashed Old Headers with orphaned block exception
After few restarts at very early stage of chiado node suddenly my OldHeaders sync crashed with an exception:

```
System.InvalidOperationException: An orphaned block on the chain Hash: 0xf37b46445a182df8463f6527216324c37e03f2bbabc2215bd431bbfe35936063
Number: 13630060
Parent: 0xc4bed041a9c2e9be13f7f519730e2281d3ce3a0595613cd7b035714ac927747c
Beneficiary: 0x2408f8a0475d3823a1f4a4bfd86b2a4b80e6edb6
Gas Limit: 17099764
Gas Used: 29289
Timestamp: 1735859005
Extra Data: 4e65746865726d696e64
Difficulty: 0
Mix Hash: 0x15d7155b93d730602caef25c6bb11f33678d9d69e962c47d9ef0a800d25900b3
Nonce: 0
Uncles Hash: 0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347
Tx Root: 0x41899120ceccd9792baaad3ee7d9146582f64dd95434613ab7c88590970eb2ed
Receipts Root: 0x0c6e5517da8c4523421f7646f9f15baa2cde45e40e69bee240bf9ac17a3616a2
State Root: 0xfa627d2e6e1b48118c68ee4826e5d3f727619c40678b548ad82ebd6bfad79d4a
BaseFeePerGas: 7
WithdrawalsRoot: 0x65b241d97bf0e7a3ff408f24f24babaa96cdfc7fb976c295e7a3e308990c7d23
ParentBeaconBlockRoot: 0x3a1c35497b7e0f5bdb2fb7b4d7999b91928adbe9798f0d30069e644122ed2908
BlobGasUsed: 0
ExcessBlobGas: 0
IsPostMerge: False
TotalDifficulty: 

   at Nethermind.Blockchain.BlockTree.<SetTotalDifficulty>g__GetParentHeader|127_0(BlockHeader current) in /src/Nethermind/Nethermind.Blockchain/BlockTree.cs:line 1363
   at Nethermind.Blockchain.BlockTree.SetTotalDifficulty(BlockHeader header) in /src/Nethermind/Nethermind.Blockchain/BlockTree.cs:line 1414
   at Nethermind.Blockchain.BlockTree.FindHeader(Hash256 blockHash, BlockTreeLookupOptions options, Nullable`1 blockNumber) in /src/Nethermind/Nethermind.Blockchain/BlockTree.cs:line 485
   at Nethermind.Blockchain.Find.IBlockFinder.FindHeader(Hash256 blockHash, Nullable`1 blockNumber) in /src/Nethermind/Nethermind.Blockchain/Find/IBlockFinder.cs:line 70
   at Nethermind.Synchronization.FastBlocks.HeadersSyncFeed.ProcessPersistedPortion(HeadersSyncBatch batch) in /src/Nethermind/Nethermind.Synchronization/FastBlocks/FastHeadersSyncFeed.cs:line 508
   at Nethermind.Synchronization.FastBlocks.HeadersSyncFeed.ProcessPersistedHeadersOrBuildNewBatch() in /src/Nethermind/Nethermind.Synchronization/FastBlocks/FastHeadersSyncFeed.cs:line 348
   at Nethermind.Synchronization.FastBlocks.HeadersSyncFeed.PrepareRequest(CancellationToken cancellationToken) in /src/Nethermind/Nethermind.Synchronization/FastBlocks/FastHeadersSyncFeed.cs:line 319
   at Nethermind.Synchronization.ParallelSync.SyncDispatcher`1.DispatchLoop(CancellationToken cancellationToken) in /src/Nethermind/Nethermind.Synchronization/ParallelSync/SyncDispatcher.cs:line 112
   at Nethermind.Synchronization.ParallelSync.SyncDispatcher`1.Start(CancellationToken cancellationToken) in /src/Nethermind/Nethermind.Synchronization/ParallelSync/SyncDispatcher.cs:line 76
```

No following restart helped - just resync started it again.
Reproduced that on at least two nodes.

DEBUG logs: 
[chiado.logs.txt](https://github.com/user-attachments/files/18350252/chiado.logs.txt)

Scooletz:
Closes #8029 

This PR makes the flush of the underlying databases explicit for `HeadersSyncFeed` inserting headers. It's a change similar to a fix done in #6227 which explicitly calls the flush after a specific boundary is breached. This PR flushes the headers after each batch. If a performance penalty occurs, it should be done in an amortized way, every N batches set. 

In addition `BodiesSyncFeed` is modified, so that it Flushes the underlying database in the same way `HeadersSyncFeed` does it. It calls the `.Flush` with a specific reason.

## Changes

- explicit call out to `.Flush` of the underlying database

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Remarks


Scooletz:
@asdacap You authored https://github.com/NethermindEth/nethermind/pull/6227 and I think you'd be most suited to provide the initial review. If you could, it'd be great.

asdacap:
Do you have some performance figure? Also, where is the NoWAL flag in fast header? If that is not use, there is not much use for an explicit flush.

The thing about blocks and receipts is that, once the compaction writes was reduced enough, the portion of the writes for the WAL file was large enough that it is worth optimizing away.

The thing about headers is that, it is not parallelized when inserting because of the total difficulty calculation, so it straight up can't saturate the IO bandwidth. Latency, sure, but not bandwith. Having a way insert batch of headers instead of one by one would help it a lot.

Scooletz:
> Do you have some performance figure? Also, where is the NoWAL flag in fast header? If that is not use, there is not much use for an explicit flush.

I'll clear it out.

> The thing about blocks and receipts is that, once the compaction writes was reduced enough, the portion of the writes for the WAL file was large enough that it is worth optimizing away.

I see.

> The thing about headers is that, it is not parallelized when inserting because of the total difficulty calculation, so it straight up can't saturate the IO bandwidth. Latency, sure, but not bandwith. Having a way insert batch of headers instead of one by one would help it a lot.

This was not intended to be addressed here, but surely I'll take a look.

Scooletz:
@asdacap  I've reviewed the answers and I think I need more imperative input or a review that points to the missing ones. Specifically:

1. `NoWAL` there's no change in `FastHeadersSyncFeed` that changes the flush behavior. The only thing is that we flush before the the metadata are assigned. What do you mean by _If that is not use, there is not much use for an explicit flush._
1. _writes for the WAL_ I spent some time and I think I'm not following. This PR tries not to touch anything peformance/throughput related. It tries to ensure that we update metadata ONLY IF the underlying headers that we'll point to will be ok
1. _total difficulty_ I think that if we want to tackle it, a separate PR should be provided. This aims only in providing a proper capture of the last inserted header.

asdacap:
Discussed offline, so basically, there is no need for flush unless NoWAL flag is used, in which case, you MUST flush. Especially in the case that kamil gave as the system does not crash, only the app.

Issue #8065, opened 2025-01-16
Automate Superchain chains configurations


emlautarom1:
**Is your feature request related to a problem? Please describe.**
Currently adding support for Superchain chains requires manual work to convert the required configs and assets into something compatible with Nethermind. This usually results in PRs like #8044.

**Describe the solution you'd like**
We would like to automate this process as follows:
- Create a script that fetches the required assets from the [Superchain Registry](https://github.com/ethereum-optimism/superchain-registry) and generates corresponding config files.
- Add a Github action similar to https://github.com/NethermindEth/nethermind/blob/e3983dfcedb276fe45154d705ef00c74c898dfab/.github/workflows/update-fast-sync.yml with such script.
- Modify Nethermind to load the assets in their compressed format directly rather than preprocessing them.

**Describe alternatives you've considered**
1. Instead of fetching the assets we could add the superchain registry as a git submodule and then trigger a pull. This will most likely result in additional assets that we don't care about, but we avoid any potential issues regarding fetching.
2. We could add an additional step to the script to convert the assets to something that Nethermind can understand right now but we believe that this will result in a lot of used storage (see https://github.com/NethermindEth/nethermind/pull/8044#pullrequestreview-2546095567).
3. Fetching the assets could be done at runtime instead of build time. This could cause issues during initialization and makes harder to know what configurations is being used, alongside the fact that each run could end up with potential different configs. Overall we think it's a bad idea.

**Additional context**
- https://github.com/NethermindEth/core-scripts/tree/main/gen2spec

emlautarom1:
This topic was revisited by additional team members and it was decided that we will go back to the original plan, that is, adding support for all Superchain chains in Nethermind itself.

cc. @stdevMac @kamilchodola @jmederosalvarado

emlautarom1:
Example of an automated PR with updated configurations: https://github.com/NethermindEth/nethermind/pull/8201

PR #8105, opened 2025-01-24
Automate Superchain Registry chain support

emlautarom1:
Solves #8065 

## Changes

- Add a `superchain.py` script to fetch, process and generate Nethermind compatible files for all OP Sperchain chains (see https://github.com/ethereum-optimism/superchain-registry)

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

Added tests to verify that we're properly decompressing generated `ChainSpec` files. Some tests were updated to use a new `ChainSpecFileLoader` that is format-aware of ChainSpec files, thus is able to decompress them if needed.

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

Once this feature is properly working we need to document that Nethermind now supports all OP Superchain chains, not only a small subset (currently: OP, Base and Worldchain).

## Remarks

The generated `ChainSpec` files are stored in a compressed format due to their excessive size when in `.json` format. Just like in the Superchain Registry repository (https://github.com/ethereum-optimism/superchain-registry/blob/cb9a0ca8eda1608bf9958137a736fd2c18884fbd/superchain/README.md) we use [zstandard](http://facebook.github.io/zstd/) as the compression algorithm. This is due to the OP team already figuring out that this is one of the best algorithms for this use case, and also to reduce the required amount of code (our scripts already deal with decompression so might as well compress the resulting artifacts using the same algorithm).

Currently, 45 configs are generated: 28 in Mainnet and 17 in Sepolia.

We're still missing a GitHub action that automatically runs this script either periodically or when the Superchain Registry repository gets updated (the latter being preferable)


kamilchodola:
@stdevMac Can you take care of checking this and making sure it works accordingly with Sedge? Also add to our CI BUT limit those side chains to trigger once per week.

emlautarom1:
@kamilchodola @stdevMac Do not test this branch yet since it's not working ATM. I'll change the status to "ready to review" once I figure out what is the issue (currently Nethermind can't find peers).

src/Nethermind/Chains/arena-z-mainnet.json.zstd
rubo, line 1: Binary files aren't Git-friendly. Are those JSONs that large, or is this compression format kind of a standard for Superchain? Could you please elaborate?

scripts/superchain.py
```diff
@@ -0,0 +1,338 @@
...
+        nethermind["genesis"]["stateUnavailable"] = True
+        nethermind["params"]["terminalTotalDifficulty"] = 210470125
+        nethermind["params"]["eip2565Transition"] = "0x3C45B0"
+        nethermind["params"]["eip2929Transition"] = "0x3C45B0"
+        nethermind["params"]["eip2930Transition"] = "0x3C45B0"
+
+    return nethermind
+
+
+def to_nethermind_runner(chain_name, l1, chain):
```
emlautarom1, line 207: The WIP `OptimismCL` section is not included at the moment.

src/Nethermind/Chains/arena-z-mainnet.json.zstd
rubo, line 1: Binary files aren't Git-friendly, but I don't have better solutions. If you're referring to the [superchain-registry](https://github.com/ethereum-optimism/superchain-registry/tree/main/superchain/extra/genesis/mainnet) repo, let's keep the standard extension of `.zst` instead of `.zstd`, as they do.

src/Nethermind/Nethermind.Specs/ChainSpecStyle/ZstdChainSpecLoader.cs
```diff
@@ -0,0 +1,43 @@
...
+            stream.CopyTo(buffer);
+        }
+        else
+        {
+            using Stream stream = File.OpenRead(_dictionaryPath);
+            stream.CopyTo(buffer);
+        }
+
+        using var decompressedStream = new DecompressionStream(streamData);
+        decompressedStream.LoadDictionary(buffer.ToArray());
```
emlautarom1, line 40: Unfortunately `LoadDictionary` does not accept any async/Stream overload.
In practice it should not be an issue since the `dictionary` is ~3.5 Mb and it's only loaded once during initialization.

src/Nethermind/Chains/arena-z-mainnet.json.zstd
emlautarom1: For some reason I confused the file extension. Changed to `.zst`

rubo: You forgot the remove `.zstd` ones, though :)

scripts/copy-superchain.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
```
rubo, line 1: Add the missing file header:

```suggestion
#!/bin/bash
# SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
# SPDX-License-Identifier: LGPL-3.0-only
```

scripts/superchain.py
```diff
@@ -0,0 +1,338 @@
+import argparse
```
rubo, line 1: Add the missing file header:

```suggestion
# SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
# SPDX-License-Identifier: LGPL-3.0-only

import argparse
```

rubo (commented):
Added missing file headers

emlautarom1:
Changing to "draft". For more context see: https://github.com/NethermindEth/nethermind/issues/8065#issuecomment-2634128782

src/Nethermind/Nethermind.Runner/configs/op-mainnet.json
```diff
@@ -1,12 +1,10 @@
 {
   "$schema": "https://raw.githubusercontent.com/NethermindEth/core-scripts/refs/heads/main/schemas/config.json",
   "Init": {
-    "ChainSpecPath": "chainspec/op-mainnet.json",
-    "GenesisHash": "0x7ca38a1916c42007829c55e69d3e9a73265554b586a499015373241b8a3fa48b",
+    "ChainSpecPath": "chainspec/op-mainnet.json.zst",
+    "GenesisHash": "0xdbf6a80fef073de06add9b0d14026d6e5a86c85f6d102c36d3d8e9cf89c2afd3",
```
deffrian, line 5: Why is it changed? Did you try to run archive and snap? Maybe it will brake something

deffrian (approved):
LGTM, but I have few questions:
1. What will happen if they change format again?
2. Can we put all op stuff in separate folder? Something like `Cains/superchain` and `configs/superchain` 
3. We need to check op-mainnet genesis is working.

src/Nethermind/Nethermind.Runner/configs/op-mainnet.json
emlautarom1: It was changed since we're pulling the data from the Superchain registry

https://github.com/ethereum-optimism/superchain-registry/blob/8ff62ada16e14dd59d0fb94ffb47761c7fa96e01/superchain/configs/mainnet/op.toml#L34

Now, this is wrong: Nethermind is not able to start due to a hash mismatch (the previous one was correct). I'm not sure what to make out of this. We already have some special logic for OP-mainnet but it should not be necessary to hardcode this hash.

emlautarom1:
> 1. What will happen if they change format again?
If they change from `zstd` to something different then the script will break. This is the same in case they change directory structure, certain index files, etc.

Particularly, if they change the compression format we can either continue using `zstd` (we'll have to adjust the scripts to decompress the new format) or adopt the new format altogether. The later is not that complicated since this PR also introduces flexibility to extend Nethermind to load other Chainspec formats. We'll see what to do when it happens.

> 2. Can we put all op stuff in separate folder? Something like `Cains/superchain` and `configs/superchain` 

We could but we need to revisit all scripts that use the root directory (ex. `syncSettings.py`) and we make the initialization code a bit trickier with not a lot off benefit IMO. For now I suggest leaving it as is.

> 3. We need to check op-mainnet genesis is working.

Currently on it. We should be able to merge once we can ensure that it works as expected.

.github/workflows/update-superchain.yml
```diff
@@ -0,0 +1,56 @@
+name: Update OP Superchain chains
+
+on:
+  schedule:
+    - cron: '0 0 * * 0'
```
emlautarom1, line 5: The workflow will run weekly. Does this sound OK @kamilchodola ?

src/Nethermind/Nethermind.Specs/ChainSpecStyle/ZstdChainSpecLoader.cs
LukaszRozmej: Ok but you first load it to MemoryStream and then ToArray, can you just use one data in the first place?

src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainSpecFileLoader.cs
```diff
@@ -0,0 +1,109 @@
...
+                {
+                    missingChainspecFileMessage.AppendLine($"  * {configFile}");
+                }
+            }
+            catch (Exception)
+            {
+                // do nothing - the lines above just give extra info and config is loaded at the beginning so unlikely we have any catastrophic errors here
+            }
+
+            throw new FileNotFoundException(missingChainspecFileMessage.ToString());
```
LukaszRozmej, line 104: As you are throwing exception regardless, would be good to preserve that catched exception as inner exception

src/Nethermind/Nethermind.Specs/ChainSpecStyle/ZstdChainSpecLoader.cs
emlautarom1: You mean that each `Stream` inside the conditional produce a `byte[]` directly instead of going through `MemoryStream`? If so then I might be missing some API because I'm not sure how to do that.

flcl42 (approved):

src/Nethermind/Nethermind.Specs/ChainSpecStyle/ZstdChainSpecLoader.cs
LukaszRozmej:
```
        using Stream stream = _dictionaryPath is null
            ? typeof(IConfig).Assembly.GetManifestResourceStream(EmbeddedDictionaryPath)!
            : File.OpenRead(_dictionaryPath);

        byte[] buffer = new byte[stream.Length];
        stream.ReadExactly(buffer);

        using var decompressedStream = new DecompressionStream(streamData);
        decompressedStream.LoadDictionary(buffer);
```

emlautarom1: Didn't know that you could "just ask" for a Stream length. Changed.

LukaszRozmej (approved):

rubo:
It looks like after this commit, the fast sync settings update always [failed](https://github.com/NethermindEth/nethermind/actions/workflows/update-fast-sync.yml).

emlautarom1:
@rubo Fixed in https://github.com/NethermindEth/nethermind/pull/8422

PR #8106, opened 2025-01-25
Fix/storage recovery does not work

asdacap:
- Fix storage recovery does not work due a change in how `StorageTree` is called.

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Can reliably fail in unit tests.

### DIFF
--- a/src/Nethermind/Nethermind.Synchronization.Test/Trie/HealingTreeTests.cs
+++ b/src/Nethermind/Nethermind.Synchronization.Test/Trie/HealingTreeTests.cs
@@ -2,16 +2,33 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 
 using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
 using System.Linq.Expressions;
+using System.Threading;
 using System.Threading.Tasks;
+using Autofac;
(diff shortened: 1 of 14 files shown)


src/Nethermind/Nethermind.Synchronization.Test/Trie/HealingTreeTests.cs
```diff
@@ -106,4 +119,102 @@ private void recovery_works<T>(
...
+            for (int i = 1; i < 100; i++)
+            {
+                mainWorldState.Set(new StorageCell(storageAddress, (UInt256)i), i.ToBigEndianByteArray());
+            }
+
+            mainWorldState.Commit(Cancun.Instance);
+            mainWorldState.CommitTree(1);
+
+            stateRoot = mainWorldState.StateRoot;
+        }
```
LukaszRozmej, line 176: This test is bit hard to follow, maybe move some parts to private methods with good naming?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Synchronization.Test/Trie/HealingTreeTests.cs
asdacap: Done

PR #8107, opened 2025-01-25
Set MaxAttemptsToUpdatePivot to -1 for infinite pivot update trials

### LINKED ISSUE #5992: Refactor `PivotUpdator` to be really infinite
Right now we are using `MaxAttemptsToUpdatePivot` = `int.MaxValue`, which is enough, because it means that node is trying to update pivot for ~68 years before giving up. But making it really infinite would be nice.

Refactor code to make infinite attempts to update pivot if `MaxAttemptsToUpdatePivot` is e.g. `-1`

**Additional context**
Automatic pivot updates were introduced here: https://github.com/NethermindEth/nethermind/pull/5256
And changed to "infinite" here: https://github.com/NethermindEth/nethermind/pull/5988

Deeptanshu-sankhwar:
Fixes Closes Resolves #5992 

## Changes

- In `ISyncConfig.cs` update the default value for `MaxAttemptsToUpdatePivot` to -1 in an attempt to make the pivot update trials to infinite instead of ~68 years
- Since `_attemptsLeft ` now begins with a value of -1 and keeps decrementing by 1, ~~updated the condition in `PivotUpdator.cs`~~ , handle its update condition
- Since the default value for `MaxAttemptsToUpdatePivot` change to -1 ~~from positive, updated it in `MultiSyncModeSelector.cs`~~, handle both positive and negative value for `ShouldBeInUpdatingPivot` in `MultiSyncModeSelector.cs`
- ~~Thereby updated its test case in `MultiSyncModeSelectorTests.Scenario.cs`~~

- [ ] Optimization
- [X] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No


damian-orzechowski (changes requested):
I think the changes should work both with special value of `-1` as well as any value placed via  configuration.

src/Nethermind/Nethermind.Synchronization.Test/ParallelSync/MultiSyncModeSelectorTests.Scenario.cs
```diff
@@ -873,7 +873,7 @@ public ScenarioBuilder WhenMergeSyncPivotNotResolvedYet()
                     _syncProgressSetups.Add(
                         () =>
                         {
-                            SyncConfig.MaxAttemptsToUpdatePivot = 3;
+                            SyncConfig.MaxAttemptsToUpdatePivot = -1;
```
damian-orzechowski, line 876: With tests, I'd be cautious to use any settings potentially causing infinite work

src/Nethermind/Nethermind.Synchronization/ParallelSync/MultiSyncModeSelector.cs
```diff
@@ -355,7 +355,7 @@ private bool ShouldBeInBeaconHeaders(bool shouldBeInUpdatingPivot)
 
         private bool ShouldBeInUpdatingPivot()
         {
-            bool updateRequestedAndNotFinished = _syncConfig.MaxAttemptsToUpdatePivot > 0;
+            bool updateRequestedAndNotFinished = _syncConfig.MaxAttemptsToUpdatePivot == -1;
```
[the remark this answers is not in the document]
Deeptanshu-sankhwar: Hi @damian-orzechowski , sorry I didn't consider the possibility of any value coming from the config, and the logic shall hold true given any value for `MaxAttemptsToUpdatePivot ` from config. I have now updated the logic where retires happen infinite number of times if `MaxAttemptsToUpdatePivot  = -1` and finite number of times if `MaxAttemptsToUpdatePivot  = N`, N being a positive number.

src/Nethermind/Nethermind.Synchronization.Test/ParallelSync/MultiSyncModeSelectorTests.Scenario.cs
Deeptanshu-sankhwar: I've reverted back the test to set `MaxAttemptsToUpdatePivot  = 3`, so it can test by retrying pivot updates for exactly 3 number of times.

src/Nethermind/Nethermind.Synchronization/ParallelSync/MultiSyncModeSelector.cs
```diff
@@ -355,7 +355,7 @@ private bool ShouldBeInBeaconHeaders(bool shouldBeInUpdatingPivot)
 
         private bool ShouldBeInUpdatingPivot()
         {
-            bool updateRequestedAndNotFinished = _syncConfig.MaxAttemptsToUpdatePivot > 0;
+            bool updateRequestedAndNotFinished = _syncConfig.MaxAttemptsToUpdatePivot != 0;
```
damian-orzechowski, line 358: I don't think this is correct. We should be updating pivot with a matching condition to logic in `PivotUpdator`, that's when `-1` or `> 0`. Could simplify it to `>= -1`, but probably have to test.

damian-orzechowski (changes requested):

src/Nethermind/Nethermind.Synchronization/ParallelSync/MultiSyncModeSelector.cs
Deeptanshu-sankhwar: I see @damian-orzechowski , it also needs to handle the possibility when `_syncConfig.MaxAttemptsToUpdatePivot >= 0` to match to a condition in `PivotUpdator`. I was just considering the possibilities to be `0` or `<0(-1)`, which should've been encompassed within `!=0`, but yes I am it needs to inculcate positive values too. I have changed the condition to `>=-1`. Thanks for pointing it out.

Deeptanshu-sankhwar: Hello @damian-orzechowski, I tested this further, and >= -1 actually causes issues in `MultiSyncModeSelector`. The problem is that it treats -1 and 0 the same, allowing retries when MaxAttemptsToUpdatePivot == 0, which shouldn't happen.

!= 0 correctly allows infinite retries (-1) and limited retries (> 0) while stopping retries when 0. This aligns with logic in `PivotUpdator`, where 0 explicitly stops retries.

The tests also confirm this behavior, so I think `!= 0` is the correct choice. Let me know if you see another case I should check!

stdevMac:
Closing — changes-requested feedback was never addressed, branch is conflicting/dirty, and pivot-updator behaviour has likely shifted in the ~14 months since. Please reopen against current master if #5992 is still relevant.

PR #8109, opened 2025-01-27
Nethermind UI (initial)

benaadams:
## Changes

- First iteration of a Nethermind UI, so users don't need access to console, have to install Seq or Grafana to monitor what is going on with their node
- Mapped when Health checks UI is mapped

image

#### Requires testing

- [x] No

- [x] Yes


src/Nethermind/Nethermind.Blockchain/BlockTree.cs
```diff
@@ -1557,6 +1558,11 @@ public void ForkChoiceUpdated(Hash256? finalizedBlockHash, Hash256? safeBlockHas
                 _metadataDb.Set(MetadataDbKeys.FinalizedBlockHash, Rlp.Encode(FinalizedHash!).Bytes);
                 _metadataDb.Set(MetadataDbKeys.SafeBlockHash, Rlp.Encode(SafeHash!).Bytes);
             }
+
+            var finalizedNumber = FindHeader(FinalizedHash, BlockTreeLookupOptions.DoNotCreateLevelIfMissing)?.Number;
+            var safeNumber = FindHeader(safeBlockHash, BlockTreeLookupOptions.DoNotCreateLevelIfMissing)?.Number;
```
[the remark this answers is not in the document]
benaadams: The ForkChoiceHandler has already looked them up; however need to work out how to plumb it through

src/Nethermind/Nethermind.Runner/Ethereum/JsonRpcRunner.cs
```diff
@@ -76,6 +76,7 @@ public async Task Start(CancellationToken cancellationToken)
                     s.AddSingleton(_jsonRpcUrlCollection);
                     s.AddSingleton(_webSocketsManager);
                     s.AddSingleton(_rpcAuthentication);
+                    s.AddSingleton(_api);
```
LukaszRozmej, line 79: I don't like this, we want to remove/hide api not inject it

src/Nethermind/Nethermind.Runner/Monitoring/DataFeed.cs
```diff
@@ -0,0 +1,375 @@
...
+{
+    public static long StartTime { get; set; }
+
+    private readonly ITxPool _txPool;
+    private readonly ISpecProvider _specProvider;
+    private readonly IReceiptFinder _receiptFinder;
+    private readonly IBlockTree _blockTree;
+    private readonly BlockDecoder _blockDecoder = new();
+
+    public DataFeed(INethermindApi api)
```
LukaszRozmej, line 39: Just list dependencies: ITxPool, IBlockTree, ISpecProvider, IReceiptFinder

src/Nethermind/Nethermind.Runner/Monitoring/DataFeed.cs
```diff
@@ -0,0 +1,375 @@
...
+
+        await ctx.Response.WriteAsync($"event: txNodes\n", cancellationToken: ct);
+        await ctx.Response.WriteAsync($"data: ", cancellationToken: ct);
+        await ctx.Response.Body.WriteAsync(TxPoolFlow.NodeJson, cancellationToken: ct);
+        await ctx.Response.WriteAsync($"\n\n", cancellationToken: ct);
+
+        await ctx.Response.WriteAsync($"event: log\n", cancellationToken: ct);
+        await ctx.Response.WriteAsync($"data: ", cancellationToken: ct);
+        await ctx.Response.Body.WriteAsync(JsonSerializer.SerializeToUtf8Bytes(ConsoleHelpers.GetRecentMessages(), JsonSerializerOptions.Web), cancellationToken: ct);
+        await ctx.Response.WriteAsync($"\n\n", cancellationToken: ct);
```
LukaszRozmej, line 73: Wouldn't it be better to do that in 1 WriteAsync?

src/Nethermind/Nethermind.Runner/Monitoring/DataFeed.cs
```diff
@@ -0,0 +1,375 @@
...
+        var channel = Channel.CreateUnbounded<ChannelEntry>();
+
+        InitializeChannelSubscriptions(channel, ct);
+
+        await foreach (ChannelEntry entry in channel.Reader.ReadAllAsync(ct))
+        {
+            await ctx.Response.WriteAsync($"event: {entry.Type}\n", cancellationToken: ct);
+            await ctx.Response.WriteAsync($"data: ", cancellationToken: ct);
+            await ctx.Response.Body.WriteAsync(entry.Data, cancellationToken: ct);
+            await ctx.Response.WriteAsync($"\n\n", cancellationToken: ct);
```
LukaszRozmej, line 84: Same here

src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -375,6 +392,9 @@ private void RemoveProcessedTransactions(Block block)
                 Metrics.Eip7702TransactionsInBlock = eip7702Txs;
                 Metrics.BlobTransactionsInBlock = blobTxs;
                 Metrics.BlobsInBlock = blobs;
+
+                Metrics.TransactionsSourcedPrivateOrderFlow += notInMempoool;
+                Metrics.TransactionsSourcedMemPool += transactionsInBlock - notInMempoool;
```
LukaszRozmej, line 397: With DarkPoolRatio, do we need those 2 metrics?

src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -350,14 +353,28 @@ private void RemoveProcessedTransactions(Block block)
                     }
                 }
 
+                if (blockTx.Type == TxType.SetCode)
+                {
+                    eip7702Txs++;
+                }
+
+                bool isKnown = true;
```
LukaszRozmej, line 361:
```suggestion
                bool isKnown = IsKnown(txHash);
```

src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -350,14 +353,28 @@ private void RemoveProcessedTransactions(Block block)
...
                 if (!IsKnown(txHash))
                 {
                     discoveredForHashCache++;
+                    isKnown = false;
                 }
 
+                bool isPending = true;
```
LukaszRozmej, line 368:
```suggestion
                bool isPending = RemoveIncludedTransaction(blockTx);
```

LukaszRozmej (approved):
Apart of my comments I'm fine with rest from C#.
Someone needs to review TypeScript, @rubo ?

src/Nethermind/Nethermind.Runner/Nethermind.Runner.csproj
```diff
@@ -84,11 +84,36 @@
...
+    <TypeScriptCompile Include="scripts\ridgeLine.ts" />
+    <TypeScriptCompile Include="scripts\treeMap.ts" />
+    <TypeScriptCompile Include="scripts\types.ts" />
+    <TypeScriptCompile Include="scripts\utilities.ts" />
+  </ItemGroup>
+
+  <Target Name="JsBuild" BeforeTargets="PreBuildEvent">
+    <Exec Command="yarn build:dev" Condition="'$(Configuration)'=='Debug'" />
+    <Exec Command="yarn build:release" Condition="'$(Configuration)'=='Release'" />
+  </Target>
```
rubo, line 115: It is highly undesirable to make yarn, hence node.js, a required tool to build Nethermind for a non-essential and optional component as the UI. The build failed for me just because I didn't have yarn.

A few suggestions:

- Do not use node.js/yarn at all
- Provide already minified versions (do we really need it, given its local usage only?)
- Having the UI as a separate NuGet package, like the health checks. I think this is the best approach and similar to the plugin approach suggested by @LukaszRozmej. I can help with that.

src/Nethermind/Nethermind.Runner/wwwroot/fonts/dm-mono-latin.woff2
rubo, line 1: Why have all these fonts in the repo instead of downloading them from HTML?

src/Nethermind/Nethermind.Runner/wwwroot/favicon.webp
rubo, line 1: Could be an SVG.

src/Nethermind/Nethermind.Runner/wwwroot/nethermind.svg
rubo, line 1: Better to put this in the `logo` directory with the other images and rename it to `images`. So, all the graphics are in the same place.

rubo (changes requested):
Although the UI concept is cool, I have some concerns about the approach. I'd prefer this to be a separate component instead of being incorporated into the `Nethermind Runner`. Please see the comments.

src/Nethermind/Nethermind.Runner/Nethermind.Runner.csproj
benaadams: node is only used in docker to build; its not included in the final container

src/Nethermind/Nethermind.Runner/wwwroot/favicon.webp
benaadams: Not on a mac or iOS https://caniuse.com/link-icon-svg

image

src/Nethermind/Nethermind.Runner/wwwroot/fonts/dm-mono-latin.woff2
benaadams: Web site has 0 requirement for an internet connection (just to your node) and doesn't send beacons requests to google

src/Nethermind/Nethermind.Runner/wwwroot/nethermind.svg
benaadams: Is a special image and different dimensions; logos is 3rd party chain logos and all the same dimensions

src/Nethermind/Nethermind.Runner/wwwroot/fonts/dm-mono-latin.woff2
benaadams: Web site has 0 requirement for an internet connection (just to your node) and doesn't send beacon requests to google

Maybe you want to put it on a big TV, but don't want to give that TV full access to the Internet just to download some fonts

src/Nethermind/Nethermind.TxPool/TxPool.cs
benaadams: Done

src/Nethermind/Nethermind.TxPool/TxPool.cs
benaadams: Done

src/Nethermind/Nethermind.Runner/Monitoring/DataFeed.cs
benaadams: Done

src/Nethermind/Nethermind.Runner/Ethereum/JsonRpcRunner.cs
benaadams: Changed to inject them individually

src/Nethermind/Nethermind.Runner/ConsoleHelpers.cs
```diff
@@ -41,3 +61,114 @@ public static void EnableConsoleColorOutput()
     [DllImport("kernel32.dll")]
     private static extern bool SetConsoleMode(IntPtr hConsoleHandle, uint dwMode);
 }
+
+
+public sealed class LineInterceptingTextWriter : TextWriter
```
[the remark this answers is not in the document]
benaadams: Not really; that I could find :(

src/Nethermind/Nethermind.Runner/Nethermind.Runner.csproj
rubo: I agree that maintaining a NuGet package is somewhat annoying, but bringing the Node.js as a new build dependency for some optional components is also annoying. I believe not everyone in the team uses or even has Node.js & stuff, let alone third parties that build from source. Moreover, with a NuGet package, you'll have more liberties regarding the organization and approaches, as there will be much less nitpicking.
Of course, this is just my opinion, but I think I'm not alone in that. I'd like more opinions here.

src/Nethermind/Nethermind.Runner/wwwroot/fonts/dm-mono-latin.woff2
rubo: I see, but on the other hand, binary stuff is not git-friendly. If it was a separate package, it would be much less of a concern.

src/Nethermind/Nethermind.Runner/wwwroot/favicon.webp
rubo: I don't think Safari is important that much in our case, but anyway. This is a minor thing.

src/Nethermind/Nethermind.Runner/wwwroot/nethermind.svg
rubo: Still, images can go into their separate directory, and there you can organize them however you wish. As mentioned above, if it was a separate package, I wouldn't care.

src/Nethermind/Nethermind.Runner/scripts/utilities.ts
```diff
@@ -0,0 +1,307 @@
...
+  // "If the cleaned text is less than half length of input size, output it as hex, else output the text."
+  // The original code in the snippet has a minor mismatch in the condition, but we’ll follow the comment:
+  if (data.length > (bytes.length / 2)) {
+    // It's "mostly" valid text
+    return data;
+  } else {
+    // It's "mostly" not decodable text => show hex
+    return `0x${hex}`;
+  }
+}
```
emlautarom1, line 97: I would personally remove a lot of code from here:
- Return the original hex string if it's null, empty or its length is > 32.
- If not, convert the hex string into a `Uint8Array`.
- Then, use the native `TextDecoder` to convert a `Uint8Array` to a `string` using `UTF-8` encoding.
- Count the number of non-printable characters. If they exceed half of the string, return the original hex string.
- Otherwise, return the decoded string.

The main issue is that we might get some of the UTF-8 decoding wrong and I don't see the point of having such complex logic for the UI.

```suggestion
export function parseExtraData(extraData: string | null | undefined): string {
  if (!extraData || extraData === "0x") {
    return "0x";
  }

  // Remove the "0x" prefix if present
  let hex = extraData.startsWith("0x")
    ? extraData.slice(2)
    : extraData;

  // If there is nothing after removing "0x", return "0x"
  if (hex.length === 0) {
    return "0x";
  }

  // Convert hex -> bytes
  const bytes = hexStringToUint8Array(hex);

  // Decode to UTF-8
  let decoder = new TextDecoder('utf-8', { fatal: false });
  let decoded = decoder.decode(bytes);

  // Count the number of control characters
  let controlCount = decoded.split('').filter(c => isControlCharacter(c.charCodeAt(0))).length;

  // Return the original hex string if there are too many control characters
  if (controlCount >= decoded.length / 2) {
    return `0x${hex}`;

  // Return the decoded string
  return decoded;
}
```

src/Nethermind/Nethermind.Runner/scripts/app.ts
```diff
@@ -0,0 +1,400 @@
...
+const formatDec = d3.format(',.1f');
+
+let txPoolNodes: INode[] = null;
+/**
+ * Main function to start polling data and updating the UI.
+ */
+function updateTxPool(txPool: TxPool) {
+
+  if (txPool.pooledTx == 0) {
+    document.getElementById("txPoolFlow").classList.add("not-active");
```
emlautarom1, line 134: If this element is going to always be present in the DOM is better to have a reference to it at the start instead of querying in a loop.

src/Nethermind/Nethermind.Runner/scripts/app.ts
```diff
@@ -0,0 +1,400 @@
...
+    if (link.target === 'Received Txs') {
+      currentReceived += link.value;
+    }
+    if (link.target === 'Tx Pool') {
+      currentTxPool += link.value;
+    }
+    if (link.target === 'Added To Block') {
+      currentBlock += link.value;
+    }
+    if (link.target === 'Duplicate') {
```
emlautarom1, line 159: I suggest extracting the constants. They're used also in https://github.com/NethermindEth/nethermind/blob/8c039eb692cde02bac0aa0f83563543b7d16e62d/src/Nethermind/Nethermind.Runner/scripts/txPoolFlow.ts#L163 for example

src/Nethermind/Nethermind.Runner/scripts/app.ts
```diff
@@ -0,0 +1,400 @@
...
+  const data = JSON.parse(e.data) as NodeData;
+
+  const networkName = getNetworkName(data.network);
+  const newTitle = `Nethermind [${networkName}]${(data.instance ? ' - ' + data.instance : '')}`;
+  if (document.title != newTitle) {
+    document.title = newTitle;
+  }
+  updateText(version, data.version);
+  updateText(network, networkName);
+  (document.getElementById("network-logo") as HTMLImageElement).src = `logos/${getNetworkLogo(data.network)}`;
```
emlautarom1, line 223: Same as with previous `getElementById`

src/Nethermind/Nethermind.Runner/scripts/app.ts
```diff
@@ -0,0 +1,400 @@
...
+      mergedData,
+      // keyFn
+      d => d.hash,
+      // orderFn
+      d => d.order,
+      // sizeFn
+      d => parseInt(d.gasUsed, 16),
+      // colorFn
+      d => parseInt(d.effectiveGasPrice, 16) * parseInt(d.gasUsed, 16)
+  );
```
emlautarom1, line 301: Maybe refactor `updateTreeMap` to accept an object for named parameters instead of relying on comments?

```suggestion
  updateTreemap<TransactionReceipt>(
    {
      element: document.getElementById("block"), // or d3.select("#treemap"),
      height: 200,
      totalSize: parseInt(data.head.gasLimit, 16),
      data: mergedData,
      keyFn: d => d.hash,
      orderFn: d => d.order,
      sizeFn: d => parseInt(d.gasUsed, 16),
      colorFn: d => parseInt(d.effectiveGasPrice, 16) * parseInt(d.gasUsed, 16)
    });
```

src/Nethermind/Nethermind.Runner/scripts/app.ts
```diff
@@ -0,0 +1,400 @@
...
+  // e.data is a JSON array of Peer objects
+  const peersArray = JSON.parse(e.data) as Peer[];
+
+  // Aggregate by clientType
+  const countsByType: Record<string, number> = {};
+
+  peersArray.forEach((peer) => {
+    const nodeType = getNodeType(peer.clientType);
+    countsByType[nodeType] = (countsByType[nodeType] || 0) + 1;
+  });
```
emlautarom1, line 320:
```suggestion
  // Aggregate by clientType
  let countsByType = data.reduce((acc, peer) => {
    let nodeType = getNodeType(peer.clientType);
    acc[nodeType] = (acc[nodeType] || 0) + 1;
    return acc;
  }, {});
```

src/Nethermind/Nethermind.Runner/scripts/app.ts
```diff
@@ -0,0 +1,400 @@
...
+      scroll = true;
+    }
+    const frag = document.createDocumentFragment();
+    for (let i = 0; i < logs.length; i++) {
+      const newEntry = document.createElement('div');
+      newEntry.innerHTML = logs[i];
+      frag.appendChild(newEntry);
+    }
+    logs = [];
+    nodeLog.appendChild(frag);
```
emlautarom1, line 386: Do we drain this `nodeLog` element at some point? Otherwise if we keep appending then we'll run OOM.

src/Nethermind/Nethermind.Runner/scripts/utilities.ts
```diff
@@ -0,0 +1,307 @@
...
+  "Mainnet": "Ethereum mainnet",
+  "1": "Ethereum mainnet",
+  "480": "World Mainnet",
+  "8453": "Base mainnet"
+}
+const logos = {
+  "Mainnet": "ethereum-logo.svg",
+  "1": "ethereum-logo.svg",
+  "480": "world-logo.svg",
+  "8453": "base-logo.svg"
```
emlautarom1, line 282: I see that we have logos for Base and Worldchain. No Optimism?

emlautarom1 (commented):
Some comments on the TS side of things.

src/Nethermind/Nethermind.Runner/scripts/utilities.ts
```diff
@@ -0,0 +1,307 @@
...
+  const day = String(date.getDate()).padStart(2, '0');
+  const month = months[date.getMonth()];
+  const year = date.getFullYear();
+  const hours = String(date.getHours()).padStart(2, '0');
+  const minutes = String(date.getMinutes()).padStart(2, '0');
+  const seconds = String(date.getSeconds()).padStart(2, '0');
+
+  // Return the formatted date string
+  // Example: "09 Feb 2025 13:45:00"
+  return `${day} ${month} ${year} ${hours}:${minutes}:${seconds}`;
```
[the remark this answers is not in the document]
rubo: `en-uk` is invalid though. `en` is enough here ;)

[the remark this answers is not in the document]
emlautarom1: Is it invalid semantically or do you get an error when trying to use it? `en` gets me the `MM/DD/YYYY` format which is common in the US while `en-UK` gets me `DD/MM/YYYY`.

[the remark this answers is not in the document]
rubo: I mean the correct locale for the UK is `en-gb,` although the browser somehow recognizes the other one. As we use a neutral locale for Nethermind everywhere else, `en` is the one to go with.

benaadams:
Was having issue with github after rebase, so reopened at this PR https://github.com/NethermindEth/nethermind/pull/8503

src/Nethermind/Nethermind.Runner/scripts/utilities.ts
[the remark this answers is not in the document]
benaadams: NATO format

src/Nethermind/Nethermind.Runner/package.json
```diff
@@ -0,0 +1,23 @@
...
+  },
+  "devDependencies": {
+    "@types/d3": "7.4.3",
+    "esbuild": "^0.25.0",
+    "typescript": "5.7.3"
+  },
+  "dependencies": {
+    "ansi-to-html": "0.7.2",
+    "d3": "7.9.0",
+    "d3-sankey": "0.12.3"
```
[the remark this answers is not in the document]
benaadams: Removed

PR #8120, opened 2025-01-29
Feature/recover whole path

asdacap:
- Current trie recover recover one node then try again without recovery.
  - This means it does not recover when more than one node in a path is missing.
- This PR changes that by trying to recover the whole path. 
  - When snap peer is available it uses account range or storage range for the full path and attempt to assemble the needed nodes from the proof of the range.
  - This seems to be about 4 time faster than assemblying the path one node at a time.
- Then when combined with an additional code recovery (not included here) allow it to execute block with nodes entirely from network. Obviously its gonna be terrible, but how terrible? About a few minute for the first block, can be 3 minute, can be 15 minute, really depends on how close is the server. After that, it can take 10 to 30 seconds per block. Lowest I've seen is about 5 second per block. It does not reliably speed up enough to escape the snap 128 block state range.

- [X] Optimization

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Tested with a hacked state sync that only save root.


src/Nethermind/Nethermind.Core/Utils/AutoCancelTokenSource.cs
```diff
@@ -0,0 +1,29 @@
...
+using System;
+using System.Threading;
+
+namespace Nethermind.Core.Utils;
+
+/// <summary>
+/// Automatically cancel and dispose underlying cancellation token source.
+/// Make it easy to have golang style defer cancel pattern.
+/// </summary>
+public readonly struct AutoCancelTokenSource(CancellationTokenSource cancellationTokenSource) : IDisposable
```
LukaszRozmej, line 13: why not derive from CancellationTokenSource?

src/Nethermind/Nethermind.Core/Tasks/WaitForPassingTasks.cs
```diff
@@ -0,0 +1,48 @@
...
+public static class Wait
+{
+    /// <summary>
+    /// Wait for any of the task that passed the predicate and forward the result, or all of the task to complete.
+    /// </summary>
+    /// <param name="cond"></param>
+    /// <param name="tasks"></param>
+    /// <typeparam name="T"></typeparam>
+    /// <returns></returns>
+    public static async Task<T> ForPassingTask<T>(Func<T, bool> cond, params IEnumerable<Task<T>> tasks)
```
LukaszRozmej, line 21: Can we make it span?
LINQ style naming? `AnyWhere` or just `Any` - in LINQ it has optional parameter:https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.any?view=net-8.0#system-linq-enumerable-any-1(system-collections-generic-ienumerable((-0))-system-func((-0-system-boolean)))
```suggestion
    public static async Task<T> AnyWhere<T>(Func<T, bool> predicate, params ReadOnlySpan<Task<T>> tasks)
```

src/Nethermind/Nethermind.Core/Tasks/WaitForPassingTasks.cs
```diff
@@ -0,0 +1,48 @@
...
+            if (cond(result))
+            {
+                // Its ok, then immediately return.
+                return result;
+            }
+
+            if (!taskSet.Any())
+            {
+                // No more tasks, just return the last one.
+                return result;
```
LukaszRozmej, line 40: shouldn't we return null or just TaskCompleated with default value?

src/Nethermind/Nethermind.Network.Stats/NodeStatsLight.cs
```diff
@@ -204,7 +204,7 @@ private void UpdateValue(ref decimal? currentValue, decimal newValue)
     {
         return (long?)(transferSpeedType switch
         {
-            TransferSpeedType.Latency => _averageLatency,
+            TransferSpeedType.Latency => _averageLatency ?? 10000,
```
LukaszRozmej, line 207: const from some timeout? int.MaxValue?

src/Nethermind/Nethermind.Synchronization/Peers/ISyncPeerPool.cs
```diff
@@ -100,12 +100,26 @@ Task<SyncPeerAllocation> Allocate(
...
             CancellationToken cancellationToken)
+        {
+            return syncPeerPool.AllocateAndRun2(
+                (peerInfo) => func(peerInfo?.SyncPeer),
+                peerAllocationStrategy,
+                allocationContexts,
+                cancellationToken);
+        }
+
+        public static async Task<T> AllocateAndRun2<T>(
```
LukaszRozmej, line 117: confused about the naming, can we just name it `AllocateAndRun`?

src/Nethermind/Nethermind.Synchronization/Trie/NodeDataRecovery.cs
```diff
@@ -0,0 +1,192 @@
...
+        if (queryPath.Truncate(startingPath.Length) != startingPath)
+        {
+            int remainingLength = 64 - startingPath.Length;
+            TreePath newQueryPath = startingPath;
+            for (int i = 0; i < remainingLength; i++)
+            {
+                newQueryPath.Append(0);
+            }
+            queryPath = newQueryPath;
+        }
```
LukaszRozmej, line 55: Move to helper method

src/Nethermind/Nethermind.Synchronization/Trie/NodeDataRecovery.cs
```diff
@@ -0,0 +1,192 @@
...
+        do
+        {
+            // In case of deeper node that already exist.
+            byte[]? nodeRlp = nodeStorage.Get(address, currentPath, currentHash);
+            if (nodeRlp is null)
+            {
+                nodeRlp = await FetchRlp(rootHash, address, currentPath, currentHash, cts.Token);
+            }
+
+            if (nodeRlp == null)
```
LukaszRozmej, line 68: is null

src/Nethermind/Nethermind.Synchronization/Trie/NodeDataRecovery.cs
```diff
@@ -0,0 +1,192 @@
...
+            int remainingLength = 64 - startingPath.Length;
+            TreePath newQueryPath = startingPath;
+            for (int i = 0; i < remainingLength; i++)
+            {
+                newQueryPath.Append(0);
+            }
+            queryPath = newQueryPath;
+        }
+
+        Dictionary<TreePath, byte[]> recoveredNodes = new();
```
LukaszRozmej, line 57: Does it make sense to be `Dictionary`? Wouldn't list suffice? (could use pooledList)

src/Nethermind/Nethermind.Synchronization/Trie/NodeDataRecovery.cs
```diff
@@ -0,0 +1,192 @@
...
+                    catch (Exception ex)
+                    {
+                        if (_logger.IsDebug) _logger.Debug($"Error recovering node from {peer} {ex}");
+                        peerPool.ReportWeakPeer(peer, AllocationContexts.State);
+                    }
+
+                    return null;
+                }, NodePeerStrategy, AllocationContexts.State, cancellationToken);
+            })
+            .ToArray();
```
LukaszRozmej, line 126: `ToPooledList` instead of array

src/Nethermind/Nethermind.Synchronization/Trie/SnapRangeRecovery.cs
```diff
@@ -0,0 +1,245 @@
...
+                            }
+                            catch (Exception ex)
+                            {
+                                if (_logger.IsWarn) _logger.Warn($"Error recovering node from {peer} {ex}");
+                                peerPool.ReportWeakPeer(peer, AllocationContexts.Snap);
+                            }
+                            return null;
+                        }, SnapPeerStrategy, AllocationContexts.Snap, cts.Token);
+                })
+                .ToArray();
```
LukaszRozmej, line 73: ToPooledList

src/Nethermind/Nethermind.Synchronization/Trie/SnapRangeRecovery.cs
```diff
@@ -0,0 +1,245 @@
...
+        if (new TreePath(queryPath, 64).Truncate(startingPath.Length) != startingPath)
+        {
+            int remainingLength = 64 - startingPath.Length;
+            TreePath newQueryPath = startingPath;
+            for (int i = 0; i < remainingLength; i++)
+            {
+                newQueryPath.Append(0);
+            }
+            queryPath = newQueryPath.Path;
+        }
```
LukaszRozmej, line 107: Same code as the other one?

LukaszRozmej (approved):
Love the idea!

src/Nethermind/Nethermind.Core/Utils/AutoCancelTokenSource.cs
asdacap: I prefer not to use inheritance.

src/Nethermind/Nethermind.Core/Tasks/WaitForPassingTasks.cs
asdacap: Because its an async code.

src/Nethermind/Nethermind.Core/Tasks/WaitForPassingTasks.cs
asdacap: Hmmm... don't know.

asdacap: Well, it does make nullability simple.

src/Nethermind/Nethermind.Synchronization/Peers/ISyncPeerPool.cs
asdacap: `AllocateAndRun` already used.

src/Nethermind/Nethermind.Core/Utils/AutoCancelTokenSource.cs
LukaszRozmej: I somewhat agree, but it IS a CancellationTokenSource, so it makes some sense

src/Nethermind/Nethermind.Core/Tasks/WaitForPassingTasks.cs
LukaszRozmej: Ok the naming comment still stands

src/Nethermind/Nethermind.Synchronization/Peers/ISyncPeerPool.cs
LukaszRozmej: but has different parameters?

asdacap: Yes, one is `Func<ISyncPeer>` another is `Func<PeerInfo>`.

LukaszRozmej: So we should be able to name them the same

asdacap: Ow I forgot this is c#

src/Nethermind/Nethermind.Synchronization.Test/Trie/RecoveryTests.cs
```diff
@@ -108,14 +108,14 @@ public void TearDown()
     [Test]
     public async Task can_recover_eth66()
     {
-        IDictionary<TreePath, byte[]>? response = await Recover(_nodeDataDataRecovery, _peerEth66);
-        response![_path].Should().BeEquivalentTo(_nodeRlp);
+        IOwnedReadOnlyList<(TreePath, byte[])>? response = await Recover(_nodeDataDataRecovery, _peerEth66);
```
LukaszRozmej, line 111: You should add using in tests too! Otherwise they can fail in Debug.

PR #8123, opened 2025-01-29
Fix blob tx size calculation

### LINKED ISSUE #8121: Fix blob hive tests
We are failling a few blobs related tests in hive
:heavy_multiplication_x: Blob Transaction Ordering, Single Account, Single Blob (Cancun) (nethermind)
[373](https://github.com/NethermindEth/nethermind/actions/runs/12989206968/job/36221761931#step:12:374)
:heavy_multiplication_x: Request Blob Pooled Transactions Multiple (Cancun) (nethermind)
[422](https://github.com/NethermindEth/nethermind/actions/runs/12989206968/job/36221761931#step:12:423)
:heavy_multiplication_x: Request Blob Pooled Transactions Single (Cancun) (nethermind)

marcindsobczak:
Fixes #8121

## Changes

- save tx size only if blobs are included in calculations

When adding filter for max tx size (https://github.com/NethermindEth/nethermind/pull/7925) I introduced a regression. In `SizeTxFilter` we are calculating sizes of blob txs without including blobs. The value is saved, so we are not recalculating it later, when we need size of the whole tx, including blobs.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

Hive tests were failing because of regression, it is fixed now

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Core/Transaction.cs
+++ b/src/Nethermind/Nethermind.Core/Transaction.cs
@@ -197,8 +197,8 @@ namespace Nethermind.Core
         /// </summary>
         public int GetLength(ITransactionSizeCalculator sizeCalculator, bool shouldCountBlobs)
         {
-            return shouldCountBlobs 
-              ? (_size ??= sizeCalculator.GetLength(this, true))
+            return shouldCountBlobs
+              ? _size ??= sizeCalculator.GetLength(this, true)
               : sizeCalculator.GetLength(this, false);
         }
 
--- a/src/Nethermind/Nethermind.Core/Transaction.cs
+++ b/src/Nethermind/Nethermind.Core/Transaction.cs
@@ -195,9 +195,13 @@ namespace Nethermind.Core
         /// <summary>
         /// Encoded transaction length
         /// </summary>
-        public int GetLength(ITransactionSizeCalculator sizeCalculator, bool shouldCountBlobs = true)
+        public int GetLength(ITransactionSizeCalculator sizeCalculator, bool shouldCountBlobs)
         {
-            return _size ??= sizeCalculator.GetLength(this, shouldCountBlobs);
+            if (_size is not null) return _size.Value;
+
+            int size = sizeCalculator.GetLength(this, shouldCountBlobs);
+            if (shouldCountBlobs) _size = size;
+            return size;
         }
 
         public string ToShortString()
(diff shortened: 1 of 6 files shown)


marcindsobczak:
We are using incorrect blob tx sizes when announcing blob txs to our peers

src/Nethermind/Nethermind.Core/Transaction.cs
```diff
@@ -197,7 +197,11 @@ private void ClearPreHashInternal()
         /// </summary>
         public int GetLength(ITransactionSizeCalculator sizeCalculator, bool shouldCountBlobs)
         {
-            return _size ??= sizeCalculator.GetLength(this, shouldCountBlobs);
+            if (_size is not null) return _size.Value;
+
+            int size = sizeCalculator.GetLength(this, shouldCountBlobs);
+            if (shouldCountBlobs) _size = size;
+            return size;
```
LukaszRozmej, line 204: We shouldn't return cached `_size` when `shouldCountBlobs` is false
```suggestion
            return shouldCountBlobs 
              ? (_size ??= sizeCalculator.GetLength(this, true))
              : sizeCalculator.GetLength(this, false);
```

LukaszRozmej (changes requested):

src/Nethermind/Nethermind.Core/Transaction.cs
marcindsobczak: Right, thanks for catching it!
Probably not worth to add `_sizeWithoutBlobs`, we are using it only once, during initial filtering

src/Nethermind/Nethermind.Core/Transaction.cs
```diff
@@ -197,7 +197,9 @@ private void ClearPreHashInternal()
         /// </summary>
         public int GetLength(ITransactionSizeCalculator sizeCalculator, bool shouldCountBlobs)
```
flcl42, line 198: Might be cool in addition to Transaction to have NetworkTransaction class-wrapper with GetTransaction method or so to make it clearer which form is used here or there. So you can't call GetLength for transaction without NetworkWrapper. May be too much work though

flcl42 (approved):

PR #8125, opened 2025-01-29
No warnings with Engine Capabilites from V1 and V2

### LINKED ISSUE #8124: Consensus client missing capabilities V2 and V1
` Consensus client missing capabilities: engine_newPayloadV2, engine_getPayloadV1, engine_getPayloadV2, engine_forkchoiceUpdatedV2, engine_forkchoiceUpdatedV1, engine_newPayloadV1
`

It looks like Teku removed support for V2 and V1, which makes sense. However, Nethermind will still print a warning. We should ensure that the warning only applies to the latest version.

MarekM25:
## Fixes #8124

Some CL clients, such as Teku, have started removing support for very old versions of the engine API. I think it makes sense to stop printing warnings in this case.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Remarks


LukaszRozmej (commented):
Potentially this could be configurable per chain, which level of support that chain expects?

MarekM25:
> Potentially this could be configurable per chain, which level of support that chain expects?

yeah, was thinking about it, but let's keep it simple and it will be easy to change if needed

PR #8127, opened 2025-01-29
Add EOF tracing fields

shemnon:
## Changes

Add the "section" and "functionDepth" tracing fields to correspond to code section and CALLF depth (respectivly). This required some deep plumbing changes to add the field in the interfaces.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [X] No

- [ ] Yes
- [X] No

#### Notes on testing

nethermind is being fuzzed via an EOF focused fuzzer. These new fields are needed for the differential fuzzing.

- [ ] Yes
- [X] No

- [ ] Yes
- [X] No

### DIFF
--- a/src/Nethermind/Nethermind.Evm/EvmState.cs
+++ b/src/Nethermind/Nethermind.Evm/EvmState.cs
@@ -127,6 +127,7 @@ public sealed class EvmState : IDisposable // TODO: rename to CallState
         ReturnStackHead = 0;
         ExecutionType = executionType;
         ProgramCounter = 0;
+        FunctionIndex = 0;
         IsTopLevel = isTopLevel;
         _canRestore = !isTopLevel;
         IsStatic = isStatic;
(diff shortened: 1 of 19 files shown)


src/Nethermind/Nethermind.Evm/EvmState.cs
```diff
@@ -127,6 +127,7 @@ private void Initialize(
         ReturnStackHead = 0;
         ExecutionType = executionType;
         ProgramCounter = 0;
+        FunctionIndex = 0;
```
shemnon, line 130: This was the cause of some obscure fuzzing breakage.  If a transaction exited (such as Out of Gas) will in a CALLF then the old function index would be used at the start of the next transaction, resulting in line 1 having an incorrect section number.

benaadams:
That you for your contribution!

src/Nethermind/Nethermind.Evm/EvmState.cs
shemnon: And to clarify, this was a test instrumentation issue only, it did no impact consensus.

PR #8131, opened 2025-01-30
Always trace invalid block even if succeeds on trace

benaadams:
## Changes

- If an invalid block succeeds when tracing it (is valid) still output the traces

#### Requires testing

- [x] No


LukaszRozmej (approved):

src/Nethermind/Nethermind.Blockchain/LogTraceDumper.cs
```diff
@@ -47,10 +48,31 @@ public static void LogDiagnosticRlp(
...
+
+            if (blocks.Count == 1)
+            {
+                blockHash = blocks[0].Hash.ToString();
+            }
+            else
+            {
+                blockHash = string.Join("|", blocks.Select(b => b.Hash.ToString()));
+            }
+        }
```
LukaszRozmej, line 75: Maybe put that in separate method

LukaszRozmej:
After more thought, I am not sure if this is useful?

benaadams:
> After more thought, I am not sure if this is useful?

It does specify if valid or invalid logs; and some logs is better than none (so at least can see what type of thing is in the block)

LukaszRozmej:
should we add valid/invalid info to file name?

src/Nethermind/Nethermind.Blockchain/LogTraceDumper.cs
benaadams: Done

benaadams:
> should we add valid/invalid info to file name?

Added

PR #8144, opened 2025-01-31
Cleanup/move more sync integration tests to DI

asdacap:
- Move more sync tests to DI.
- Double check prod config works.
- Cut some code.

- [X] Refactoring

#### Requires testing

- [ ] Yes
- [X] No

- [X] Yes
- [ ] No

#### Notes on testing

- Seems to work.


src/Nethermind/Nethermind.Synchronization/Blocks/SyncBatchSize.cs
```diff
@@ -8,7 +8,7 @@
 namespace Nethermind.Synchronization.Blocks
 {
     [DebuggerDisplay("{Current}")]
-    public struct SyncBatchSize
+    public class SyncBatchSize
```
LukaszRozmej, line 11: why class?

LukaszRozmej (approved):

PR #8145, opened 2025-01-31
Fix/initsteps race condition

### LINKED ISSUE #8122: Sometimes client hangs in startup, when a required plugin fails to load.
**Description**
There is a race condition, that sometimes causes the client to hang on startup, when a required plugin fails. 

**Steps to Reproduce**
1. Setup a required plugin (MustInitialize == true)
2. Make it fail to start (throw exception etc.)
3. repeat until the client hangs and does not progress.

ak88:
Fixes #8122

## Changes

When a `IStep` fails it can cause the main process to hang in an Autoreset Event. 
I have refactored the flow in `EthereumStepsManager` to be based on awaitable tasks instead, avoiding any race conditions and greatly simplyfying it. 

- _List the changes_
- Steps and their dependencies are handled with `Task` instead of reset event.

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Remarks


src/Nethermind/Nethermind.Init/Steps/InitStep.cs
```diff
@@ -0,0 +1,41 @@
...
+
+    private TaskCompletionSource _taskCompletedSource = new TaskCompletionSource();
+    public InitStep()
+    {
+        StepCompleted = _taskCompletedSource.Task;
+    }
+
+    public async Task Execute(IEnumerable<Task> dependentSteps, CancellationToken cancellationToken)
+    {
+        cancellationToken.Register(() => _taskCompletedSource.TrySetCanceled());
```
asdacap, line 25: Can you move this logic into `EthereumStepsManager`? Like, keep `IStep`(s) as is.

src/Nethermind/Nethermind.Init/Steps/EthereumStepsManager.cs
```diff
@@ -36,138 +34,79 @@ public EthereumStepsManager(
...
-            } while (changedAnything);
         }
 
         public async Task InitializeAll(CancellationToken cancellationToken)
         {
-            while (_allSteps.Any(static s => s.Stage != StepInitializationStage.Complete))
+            List<Task> allRequiredSteps = CreateAndExecuteSteps(cancellationToken);
+            if (allRequiredSteps.Count == 0)
+                return;
+            Task current;
```
LukaszRozmej, line 44:
```suggestion
```

src/Nethermind/Nethermind.Init/Steps/EthereumStepsManager.cs
```diff
@@ -36,138 +34,79 @@ public EthereumStepsManager(
...
             {
-                cancellationToken.ThrowIfCancellationRequested();
-
-                RunOneRoundOfInitialization(cancellationToken);
-                await ReviewDependencies(cancellationToken);
-                ReviewFailedAndThrow();
-            }
-
-            await Task.WhenAll(_allPending);
+                current = await Task.WhenAny(allRequiredSteps);
```
LukaszRozmej, line 47:
```suggestion
                Task current = await Task.WhenAny(allRequiredSteps);
```

src/Nethermind/Nethermind.Init/Steps/InitStep.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Nethermind.Init.Steps;
+public abstract class InitStep
+{
+    public Task StepCompleted => _taskCompletedSource.Task;
```
LukaszRozmej, line 15:
```suggestion
    public Task StepTask => _taskCompletedSource.Task;
```

LukaszRozmej (commented):
I like going to task, I think we can avoid changing `IStep` interface and make `InitStep` wrapper (can be private class of StepManager) rather than base class.

src/Nethermind/Nethermind.Init/Steps/EthereumStepsManager.cs
```diff
@@ -201,13 +143,36 @@ private async Task ExecuteStep(IStep step, StepInfo stepInfo, CancellationToken
...
+
+                await Task.WhenAll(dependentSteps.Select(s => s.StepTask));
+                try
+                {
+                    await step.Execute(cancellationToken);
+                    _taskCompletedSource.SetResult();
+                }
+                catch
+                {
+                    _taskCompletedSource.SetCanceled();
```
asdacap, line 171: I think this should be set exception?

ak88: It would create a cascade effect  where all depending steps would write an error to the log.

src/Nethermind/Nethermind.Init/Steps/InitStep.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        cancellationToken.Register(() => _taskCompletedSource.TrySetCanceled());
+
+        await Task.WhenAll(dependentSteps);
+        try
+        {
+            await Setup(cancellationToken);
+            _taskCompletedSource.SetResult();
+        }
+        catch (Exception)
```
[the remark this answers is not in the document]
ak88: changed

src/Nethermind/Nethermind.Init/Steps/InitStep.cs
```diff
@@ -0,0 +1,41 @@
...
+using MathNet.Numerics.Differentiation;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Nethermind.Init.Steps;
+public abstract class InitStep
```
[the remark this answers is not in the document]
ak88: introduced wrapper instead, IStep is the same as before

src/Nethermind/Nethermind.Init/Steps/EthereumStepsManager.cs
ak88: fixed

src/Nethermind/Nethermind.Init/Steps/EthereumStepsManager.cs
ak88: fixed

src/Nethermind/Nethermind.Init/Steps/InitStep.cs
ak88: renamed

src/Nethermind/Nethermind.Init/Steps/InitStep.cs
ak88: It is now a wrapper class in `EthereumStepsManager`

src/Nethermind/Nethermind.Init/Steps/EthereumStepsManager.cs
```diff
@@ -36,153 +34,96 @@ public EthereumStepsManager(
...
-                stepInfo.Stage = StepInitializationStage.Executing;
-                startedThisRound++;
-                Task task = ExecuteStep(step, stepInfo, cancellationToken);
-
-                if (step.MustInitialize)
-                {
-                    _allPending.Enqueue(task);
-                }
-                else
+                createdSteps.Add(step.GetType(), new StepWrapper(step));
```
asdacap, line 67: The dependency is mapped by `StepBaseType`

src/Nethermind/Nethermind.Runner.Test/Ethereum/Steps/EthereumStepsManagerTests.cs
```diff
@@ -153,7 +153,7 @@ public StepA(NethermindApi runnerContext)
         }
     }
 
-    [RunnerStepDependencies(typeof(StepC))]
+    [RunnerStepDependencies(typeof(StepCStandard))]
```
asdacap, line 156: I think the previous test case is correct, the dependency is declared by base type not the subtype. Its a subtle thing that plugins rely on.

ak88: Ok I see. If that is the intended behavior there can be an issue if steps are inhering from another step like `InitDatabaseSnapshot` and something then uses that as a dependency.

asdacap: Many scary issue that I don't want to bring in my sleep.

src/Nethermind/Nethermind.Init/Steps/EthereumStepsManager.cs
LukaszRozmej: Can you add comment about why it is that way then?

src/Nethermind/Nethermind.Runner.Test/Ethereum/Steps/EthereumStepsManagerTests.cs
LukaszRozmej: In my original implementation, that wasn't an issue, steps were grouped by base step for dependency resolution

asdacap:
Can you address the `StepBaseType` thing? The PR is a good idea in general.

ak88:
> Can you address the `StepBaseType` thing? The PR is a good idea in general.

Yes, I need to finish another PR first, and then this one.

src/Nethermind/Nethermind.Init/Steps/EthereumStepsManager.cs
ak88: Added a comment

src/Nethermind/Nethermind.Init/Steps/EthereumStepsManager.cs
ak88: steps are mapped by base type with a corresponding dependency list now

src/Nethermind/Nethermind.Init/Steps/EthereumStepsManager.cs
```diff
@@ -36,153 +34,105 @@ public EthereumStepsManager(
...
             long startTime = Stopwatch.GetTimestamp();
             try
             {
-                await step.Execute(cancellationToken);
+                List<StepWrapper> dependencies = [];
+                foreach (Type type in stepInfo.Dependencies)
+                {
+                    if (!stepBaseTypeMap.ContainsKey(type))
+                        throw new StepDependencyException($"The dependent step {type.Name} for {stepInfo.StepType.Name} was not created.");
+                    foreach (Type dependency in stepInfo.Dependencies)
```
asdacap, line 105: Is this nested loop intended? Not just `dependencies.Add(stepBaseTypeMap[type])`?

asdacap (approved):

src/Nethermind/Nethermind.Init/Steps/EthereumStepsManager.cs
```diff
@@ -36,153 +34,102 @@ public EthereumStepsManager(
...
-
-                if (step.MustInitialize)
-                {
-                    _allPending.Enqueue(task);
+                    stepBaseTypeMap[stepInfo.StepBaseType].Add(new StepWrapper(step));
                 }
                 else
                 {
-                    stepInfo.Stage = StepInitializationStage.Complete;
+                    stepBaseTypeMap.Add(stepInfo.StepBaseType, new List<StepWrapper>() { new StepWrapper(step) });
```
LukaszRozmej, line 74: You can do it sligthly better with [CollectionsMarshal.GetValueRefOrNullRef](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.collectionsmarshal.getvaluerefornullref?view=net-9.0#system-runtime-interopservices-collectionsmarshal-getvaluerefornullref-3(system-collections-generic-dictionary((-0-1))-alternatelookup((-2))-2))

```
List<StepWrapper>? list = CollectionsMarshal.GetValueRefOrNullRef(stepBaseTypeMap, stepInfo.StepBaseType);
list ??= new();
list.Add(new StepWrapper(step))
```

src/Nethermind/Nethermind.Init/Steps/EthereumStepsManager.cs
```diff
@@ -36,153 +34,102 @@ public EthereumStepsManager(
...
+            List<Task> allRequiredSteps = new();
+            foreach (List<StepWrapper> steps in stepBaseTypeMap.Values)
             {
-                Interlocked.Increment(ref _foreverLoop);
-                if (_foreverLoop > 100)
+                foreach (StepWrapper stepWrapper in steps)
                 {
-                    if (_logger.IsWarn) _logger.Warn($"Didn't start any initialization steps during initialization round and all previous steps are already completed.");
+                    StepInfo stepInfo = stepInfoMap[stepWrapper.Step.GetType()];
+                    Task task = ExecuteStep(stepWrapper, stepInfo, stepBaseTypeMap, cancellationToken);
```
LukaszRozmej, line 83: This will execute all steps per base step type right?
That was not the logic before. Before we executed only one, the most overridden one from chosen consensus perspective.

EDIT: I think this is guaranteed through `EthereumStepsLoader` and not here. Here basetypes are for dependency resolution only.

src/Nethermind/Nethermind.Init/Steps/EthereumStepsManager.cs
```diff
@@ -36,153 +34,102 @@ public EthereumStepsManager(
...
                 }
                 else
                 {
-                    stepInfo.Stage = StepInitializationStage.Complete;
+                    stepBaseTypeMap.Add(stepInfo.StepBaseType, new List<StepWrapper>() { new StepWrapper(step) });
                 }
             }
-
-            if (startedThisRound == 0 && _allPending.All(static t => t.IsCompleted))
+            List<Task> allRequiredSteps = new();
```
LukaszRozmej, line 77: Could be a pooled list, but its one-time code, so maybe not worth it.

src/Nethermind/Nethermind.Init/Steps/EthereumStepsManager.cs
```diff
@@ -36,153 +34,102 @@ public EthereumStepsManager(
...
         }
 
-        private async Task ExecuteStep(IStep step, StepInfo stepInfo, CancellationToken cancellationToken)
+        private async Task ExecuteStep(StepWrapper stepWrapper, StepInfo stepInfo, Dictionary<Type, List<StepWrapper>> stepBaseTypeMap, CancellationToken cancellationToken)
         {
             long startTime = Stopwatch.GetTimestamp();
             try
             {
-                await step.Execute(cancellationToken);
+                List<StepWrapper> dependencies = [];
```
LukaszRozmej, line 100: Could be a pooled list, but its one-time code, so maybe not worth it.

LukaszRozmej (approved):

src/Nethermind/Nethermind.Init/Steps/EthereumStepsManager.cs
ak88: think it's overkill here

src/Nethermind/Nethermind.Init/Steps/EthereumStepsManager.cs
ak88: think it's overkill here

PR #8146, opened 2025-01-31
rpc-compat: remove Author from BlockForRpc

pnowosie:
## Changes

- Removes `Author` field from BlockForRpc

## Types of changes

Removes redundant field from RPC getBlockBy * response.

- [x] Refactoring

#### Requires testing

- [x] Yes - Required test changes

- [x] Yes


MarekM25 (commented):
Pls make CI green

MarekM25 (approved):

LukaszRozmej (commented):
I think `author` made some sense for AuRa networks, not sure if it is needed though

benaadams:
> I think `author` made some sense for AuRa networks, not sure if it is needed though

Looking at the removals; it's only ever going to be the same value as `Miner` which is already included in the output

benaadams (approved):

benaadams:
> I think `author` made some sense for AuRa networks, not sure if it is needed though

Could just change the 

`Author = block.Author ?? block.Beneficiary;`

To

`Author = block.Author;`

And let the Json nullability remove it from output then?

src/Nethermind/Nethermind.Facade/Eth/BlockForRpc.cs
```diff
@@ -30,7 +30,6 @@ protected BlockForRpc()
     public BlockForRpc(Block block, bool includeFullTransactionData, ISpecProvider specProvider)
     {
         _isAuRaBlock = block.Header.AuRaSignature is not null;
-        Author = block.Author ?? block.Beneficiary;
```
benaadams, line 33: Change to 
`Author = block.Author;`

Then let `[JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]` remove it?

benaadams (changes requested):

src/Nethermind/Nethermind.Facade/Eth/BlockForRpc.cs
LukaszRozmej: I like that better!

LukaszRozmej:
Closed for #8314

PR #8154, opened 2025-02-03
FCU should move GlobalWorldState Root

LukaszRozmej:
## Changes

- Change global world state state root on FCU

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

Please run nodes.

## Remarks

- Why didn't we implement it earlier? Probably omission.
- It doesn't matter much with current state management, it matters a lot with Paprika.


src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
```diff
@@ -259,6 +232,7 @@ protected virtual bool IsOnMainChainBehindHead(Block newHeadBlock, ForkchoiceSta
         if (shouldUpdateHead)
         {
             _blockTree.UpdateMainChain(blocks!, true, true);
+            globalWorldState.StateRoot = newHeadBlock.StateRoot!;
```
LukaszRozmej, line 235: Main change here

Scooletz (commented):
As pointed in the other PR, I'd prefer to keep one owner of the global state and leave it up to the `BlockProcessor` to move the root as it wants to. `FCU` should just notify the processor about the checkpoint being changed.

Scooletz (commented):
Rereviewing it after discussion. The main difference between #8143 and this PR is that #8143 moves the checkpoint to which the state is reset to after each branch and sets it to the `FCU.Finalized`. This ensures that if the options are `options.ContainsFlag(ProcessingOptions.DoNotUpdateHead)`, the point to which state is restored is the last one that was finalized. With this PR, on empty queue, it changes the `StateRoot` of the global state, so that it becomes the checkpoint that is captured before and restored to later on. Design-wise, I think this PR is simpler. 

One more difference is that these PRs set the checkpoint differently, this uses `FCU.Head`, while the other uses `FCU.Finalized`. If we're ok with this being set to head, then it might be even faster in regards to making it faster (no change on the next block without reorg).

If the previous is 👍 , I'm 👍 with closing #8143.

src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
```diff
@@ -259,6 +233,7 @@ protected virtual bool IsOnMainChainBehindHead(Block newHeadBlock, ForkchoiceSta
         if (shouldUpdateHead)
         {
             _blockTree.UpdateMainChain(blocks!, true, true);
+            _stateRootUpdater.StateRoot = newHeadBlock.StateRoot;
```
LukaszRozmej, line 236: Main change here

Scooletz (changes requested):

LukaszRozmej:
needs tests

Scooletz:
@LukaszRozmej Could you ack that setting the `StateRoot` to `FCU.Head` is fine? I think that either from the reorg pov or head just moving forward, we should be fine, but would be nice to have it explicitly stated in the PR.

LukaszRozmej:
> @LukaszRozmej Could you ack that setting the `StateRoot` to `FCU.Head` is fine? I think that either from the reorg pov or head just moving forward, we should be fine, but would be nice to have it explicitly stated in the PR.

ack, even desired

benaadams (approved):

src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
```diff
@@ -464,4 +439,52 @@ private BlockInfo[] GetBeaconChainBranch(Block newHeadBlock, BlockInfo newHeadBl
...
+        private Hash256? _stateRoot;
+
+        public StateRootUpdater(IWorldState globalWorldState, IBlockProcessingQueue queue)
+        {
+            _globalWorldState = globalWorldState;
+            _queue = queue;
+            _queue.ProcessingQueueEmpty += OnProcessingQueueEmpty;
+        }
+
+        public Hash256? StateRoot
```
flcl42, line 456: So _globalWorldState.StateRoot can be potentially directly changed from other place of the app. Seems a bit unsafe. Maybe we can have `_globalWorldState.EnqueueStateSelection(newStateRoot)` and read only `_globalWorldState.StateRoot`?

LukaszRozmej: It is under lock in general and state root shouldn't be changed from anywhere eslse.

src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
```diff
@@ -464,4 +439,52 @@ private BlockInfo[] GetBeaconChainBranch(Block newHeadBlock, BlockInfo newHeadBl
...
+        {
+            set
+            {
+                if (_stateRoot != value)
+                {
+                    Interlocked.Exchange(ref _stateRoot, value);
+                    if (value is not null && _queue.IsEmpty)
+                    {
+                        UpdateStateRoot();
```
Scooletz, line 465: @LukaszRozmej would you consider removing this fast path for the update? The reason would be to always run on the queue empty handler rather than potentially make an in-flight cross-thread change. It should not change much in performance (we finally have the queue empty and we run on it) but could address some potential issues.

LukaszRozmej: Tehn you can potentially wait next 12s

Scooletz: But we progress. The lack of moving forward was the initial cause to implement the update of the checkpoint. I don't see anything wrong with waiting for 1 block if this makes the threading reasoning simpler. Maybe I'm missing something here.

LukaszRozmej: well there is logic in what you say. I did want to make it consistent with FCU's received though.

src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
```diff
@@ -116,7 +90,7 @@ protected virtual bool IsOnMainChainBehindHead(Block newHeadBlock, ForkchoiceSta
             default : // Don't boost priority if we are definitely syncing
             Thread.CurrentThread.BoostPriority();
 
-        if (_invalidChainTracker.IsOnKnownInvalidChain(forkchoiceState.HeadBlockHash, out Hash256? lastValidHash))
+        if (invalidChainTracker.IsOnKnownInvalidChain(forkchoiceState.HeadBlockHash, out Hash256? lastValidHash))
```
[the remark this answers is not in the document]
benaadams: Doesn't understand primary constructors

src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
```diff
@@ -129,15 +103,15 @@ protected virtual bool IsOnMainChainBehindHead(Block newHeadBlock, ForkchoiceSta
 
             BlockHeader? headBlockHeader = null;
 
-            if (_blockCacheService.BlockCache.TryGetValue(forkchoiceState.HeadBlockHash, out Block? block))
+            if (blockCacheService.BlockCache.TryGetValue(forkchoiceState.HeadBlockHash, out Block? block))
```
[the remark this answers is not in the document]
benaadams: Doesn't understand primary constructors

src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
LukaszRozmej: BTW if queue is empty then the lock on NewPayload/FCU should prevent anything to be added to the queue

stdevMac:
@LukaszRozmej stale-PR check: the premise here was Paprika, which lost to FlatDb — though the idea itself was called desirable in review. OK to close this version (and redo FlatDb-era if still wanted)? If there's no reply in the next few days, we'll close this as part of the stale-PR cleanup — it can always be reopened or resubmitted.

src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
```diff
@@ -464,4 +439,52 @@ private BlockInfo[] GetBeaconChainBranch(Block newHeadBlock, BlockInfo newHeadBl
...
+                }
+            }
+        }
+
+        private void UpdateStateRoot()
+        {
+            Hash256? stateRoot = Interlocked.Exchange(ref _stateRoot, null);
+            if (stateRoot is not null)
+            {
+                _globalWorldState.StateRoot = stateRoot;
```
flcl42, line 476: **[HIGH] State-root application can race with newer FCUs and block processing**

`Interlocked.Exchange` protects only the pending hash, while `UpdateStateRoot` can run from both the processing-queue callback and the FCU fast path. A queue-empty callback can take root A, a later FCU can apply root B while that callback is paused, and then the callback can write A last; a payload can likewise start after the empty observation and receive this write during processing. `WorldState.StateRoot` updates its account and storage providers separately, so this interleaving can leave global state behind `BlockTree.Head` or internally split and cause a valid payload to fail state-root validation.

src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
```diff
@@ -34,57 +35,30 @@ namespace Nethermind.Merge.Plugin.Handlers;
...
+    IPayloadPreparationService payloadPreparationService,
+    IBlockProcessingQueue processingQueue,
+    IBlockCacheService blockCacheService,
+    IInvalidChainTracker invalidChainTracker,
+    IMergeSyncController mergeSyncController,
+    IBeaconPivot beaconPivot,
+    IPeerRefresher peerRefresher,
+    ISpecProvider specProvider,
+    ISyncPeerPool syncPeerPool,
+    IWorldState globalWorldState,
```
flcl42, line 51: **[MEDIUM] The public handler constructor loses binary compatibility**

`ForkchoiceUpdatedHandler` is public, and Nethermind publishes reference assemblies against which out-of-tree plugins can compile. Replacing its sole constructor with this new primary-constructor signature removes the previous metadata member, so an existing plugin binary that instantiates the handler reaches `MissingMethodException` during startup after upgrading. This makes the feature a binary compatibility break despite its non-breaking classification.

src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
```diff
@@ -464,4 +439,52 @@ private BlockInfo[] GetBeaconChainBranch(Block newHeadBlock, BlockInfo newHeadBl
...
+            Hash256? stateRoot = Interlocked.Exchange(ref _stateRoot, null);
+            if (stateRoot is not null)
+            {
+                _globalWorldState.StateRoot = stateRoot;
+            }
+        }
+
+        private void OnProcessingQueueEmpty(object? sender, EventArgs e)
+        {
+            UpdateStateRoot();
```
flcl42, line 482: **[MEDIUM] Deferred state root is applied when the queue drains even after processing has advanced the world state past it**

When an FCU sets the head to an already-processed block while the processing queue still holds later blocks, the new head's state root is stored as the pending value and is only applied once the queue reports empty. The pending value is never correlated with what the queue processes afterward. In a sequential flow such as newPayload(P1), newPayload(P2), then FCU(P1) arriving after P1 is processed but while P2 is still queued, the pending root becomes P1's root; the processing loop then executes P2, which moves BlockTree.Head to P2 and commits GlobalWorldState to P2's root. When the queue drains, this callback writes the stale P1 root back into the world state, leaving GlobalWorldState.StateRoot behind BlockTree.Head. A subsequent FCU for P2 does not repair this, because EnsureNewHead returns null when the head already matches, so shouldUpdateHead is false and the StateRoot setter is never invoked again. The mismatch persists until the next processed block resets the root via InitBranch (which also discards state caches and increments the reorganization metric as if a reorg had occurred), and any consumer reading GlobalWorldState at its current root in that window observes the older state. Applying the pending root only when it still corresponds to the current head, or invalidating it when processing advances the head beyond it, would avoid the stale write.

src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
flcl42, line 236: **[LOW] State-root movement on forkchoice update has no regression test coverage**

The change introduces `_stateRootUpdater.StateRoot = newHeadBlock.StateRoot` and a nested `StateRootUpdater` that mutates `GlobalWorldState.StateRoot` in response to Engine API forkchoice updates, but the diff only adjusts existing test wiring (`BaseEngineModuleTests` passes `chain.WorldStateManager.GlobalWorldState`, and `TaikoEngineApiTests` passes `Substitute.For<IWorldState>()`) without adding a test that exercises the behavior. No assertion verifies that a forkchoice update to an already-processed head moves the global world state root to that head's `StateRoot`, either on the immediate path (empty queue) or the deferred path (queue drains). A regression in either path would surface only indirectly through downstream state-reader or block-processing failures. A focused test that sends a forkchoice update for an already-processed head and asserts `GlobalWorldState.StateRoot` equals that head's `StateRoot` would establish the expected outcome.

flcl42 (commented):
Found 4 issues (1 high, 2 medium, and 1 low) in `src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs`.

wurdum (approved):

LukaszRozmej:
Closing as obsolete — the world state architecture this PR was fixing no longer exists.

The change was built on `IWorldStateManager.GlobalWorldState` being an `IWorldState` with a mutable `StateRoot` setter: an ambient global root that lagged behind `BlockTree.Head` until the next processed block reset it, which is what `StateRootUpdater` was pushing forward on FCU (and what #8143 addressed from the other side, by moving the reset checkpoint to `FCU.Finalized`).

On current master:

- `IWorldStateManager.GlobalWorldState` is an `IWorldStateScopeProvider`, not an `IWorldState`.
- `IWorldState` has no `StateRoot` setter at all — only `RecalculateStateRoot()`. There is nothing left to assign.
- Every consumer opens a scope against an explicit base block (`BeginScope(BlockHeader)`) — `BranchProcessor`, `GenesisLoader`, the read-only tx processing envs, the witness collectors. There is no ambient root that can fall behind the head.
- The `InitBranch` / branch-restore machinery whose behaviour motivated both PRs has been removed outright (no hits anywhere in the tree).

The Paprika motivation is also gone (FlatDb won), and the remaining diff is mostly the primary-constructor conversion, which master already has. The open review points — the state-root/queue race, the stale deferred root when processing advances past the FCU head, and the missing regression test — all concern code that has no counterpart to port to.

If a checkpoint-advance signal is ever wanted for the flat/scope-based backend, it should be designed against `IWorldStateScopeProvider` rather than resurrected from here.

PR #8156, opened 2025-02-03
requests validation hotfix

rjnrohit:
Incorporate changes from https://github.com/ethereum/execution-spec-tests/pull/1139

### DIFF
--- a/src/Nethermind/Nethermind.Merge.Plugin/Data/IExecutionPayloadParams.cs
+++ b/src/Nethermind/Nethermind.Merge.Plugin/Data/IExecutionPayloadParams.cs
@@ -51,7 +51,29 @@ public class ExecutionPayloadParams<TVersionedExecutionPayload>(
             if (ExecutionRequests.Length > ExecutionRequestExtensions.MaxRequestsCount)
             {
                 error = $"Execution requests must have less than {ExecutionRequestExtensions.MaxRequestsCount} items";
-                return ValidationResult.Invalid;
+                return ValidationResult.Fail;
+            }
+
+            // verification of the requests
+            for (int i = 0; i < ExecutionRequests.Length; i++)
+            {
+                if (ExecutionRequests[i] == null || ExecutionRequests[i].Length <= 1)
+                {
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Merge.Plugin/Data/IExecutionPayloadParams.cs
```diff
@@ -51,7 +51,29 @@ public ValidationResult ValidateParams(IReleaseSpec spec, int version, out strin
...
+                    return ValidationResult.Fail;
+                }
+
+                if (i > 0 && ExecutionRequests[i][0] <= ExecutionRequests[i - 1][0])
+                {
+                    error = "Execution requests must be ordered by request_type in ascending order";
+                    return ValidationResult.Fail;
+                }
+
+                if (ExecutionRequests.Skip(i + 1).Any(req => req != null && req[0] == ExecutionRequests[i][0]))
```
MarekM25, line 72: Proposition: can you use var called previousExecutionType and currentExecutionType, you update previousExecutionType at the end of the loop and currentExecutionType at the begining. You compare it here. Sounds much more simpler than Linq experssion. Think about it.

What else this could be used for the previous condition as well

MarekM25 (commented):
The PR looks good, but you have to add tests

src/Nethermind/Nethermind.Merge.Plugin/Data/IExecutionPayloadParams.cs
rjnrohit: I have just given a thought to the third check; it is unnecessary as strictly ascending order would confirm that a type would never appear twice

rjnrohit:
> The PR looks good, but you have to add tests

Okay adding few tests

src/Nethermind/Nethermind.Merge.Plugin/Data/IExecutionPayloadParams.cs
MarekM25: True, let's remove

LukaszRozmej (approved):

PR #8157, opened 2025-02-03
Proper Transaction `type` for `eth_*` calls when running Optimism

emlautarom1:
## Changes

- Properly map Transaction `type` instead of always using `0x7E` ("deposit")

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

Added unit tests to ensure that we're not changing `TxType` and we're including additional receipt information for "deposit" transactions only.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks

I'm curious as to why we didn't get this reported earlier considering that the issue has been present since #7483.


src/Nethermind/Nethermind.Optimism/Rpc/OptimismEthRpcModule.cs
```diff
@@ -160,76 +157,86 @@ public override async Task<ResultWrapper<Hash256>> eth_sendRawTransaction(byte[]
...
-        OptimismTransactionForRpc transactionModel = new OptimismTransactionForRpc(
-            transaction,
-            blockHash: block.Hash,
-            receipt: receipts.FirstOrDefault(r => r.TxHash == transaction.Hash));
+        var transactionModel = TransactionForRpc.FromTransaction(transaction, blockHash: block.Hash);
+        if (transactionModel is DepositTransactionForRpc depositTx)
+        {
+            OptimismTxReceipt? receipt = _receiptFinder
+                .Get(block)
+                .Cast<OptimismTxReceipt>()
```
emlautarom1, line 200: Are all receipts always of type `OptimismTxReceipt` or is it possible to have a `TxReceipt`? If the later is correct then this cast will fail at runtime.

cc. @flcl42 @jmederosalvarado

src/Nethermind/Nethermind.Optimism/Rpc/OptimismEthRpcModule.cs
```diff
@@ -249,17 +256,27 @@ public override async Task<ResultWrapper<Hash256>> eth_sendRawTransaction(byte[]
...
+            {
+                var transactionModel = TransactionForRpc.FromTransaction(
+                    transaction: block.Transactions[index],
+                    blockHash: block.Hash,
+                    blockNumber: block.Number,
+                    txIndex: index);
+                if (transactionModel is DepositTransactionForRpc depositTx)
+                {
+                    var receipt = _receiptFinder
+                        .Get(block)
```
LukaszRozmej, line 272: you are getting receipts for same block multiple times

LukaszRozmej (commented):
Potentially you could make `EthRpcModule` bit more extensible and avoid some code duplication.

deffrian (approved):

PR #8161, opened 2025-02-04
Fix PivotTotalDifficulty in update configs action

kamilchodola:
TD was removed probably from RPC endpoints and defaults to 0 which is wrong.
Made change so PoS chains will not get affected in future.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

### DIFF
--- a/scripts/syncSettings.py
+++ b/scripts/syncSettings.py
@@ -22,113 +22,135 @@ configs = {
     "mainnet": {
         "url": "api.etherscan.io",
         "blockReduced": 1000,
-        "multiplierRequirement": 1000
+        "multiplierRequirement": 1000,
+        "isPoS": True
     },
     "gnosis": {
         "url": "https://rpc.gnosischain.com",
         "blockReduced": 8192,
-        "multiplierRequirement": 10000
+        "multiplierRequirement": 10000,
+        "isPoS": True
     },
     "chiado": {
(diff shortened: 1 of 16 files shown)


scripts/syncSettings.py
```diff
@@ -22,113 +22,135 @@
...
     with open(f'{configsPath}/{configuration}.json', 'r') as mainnetCfg:
         data = json.load(mainnetCfg)
-        data['Sync']['PivotNumber'] = baseBlock
-        data['Sync']['PivotHash'] = pivotHash
+
+    data['Sync']['PivotNumber'] = baseBlock
+    data['Sync']['PivotHash'] = pivotHash
+  
+    if not isPoS:
         data['Sync']['PivotTotalDifficulty'] = str(pivotTotalDifficulty)
```
asdacap, line 149: So it does not set it now? And set to null?

asdacap (approved):

scripts/syncSettings.py
kamilchodola: It will simply not change it json - so keep current value for PoS chains as it is static since merge.

PR #8173, opened 2025-02-05
Optimize TryCalculateFeePerBlobGas

benaadams:
## Changes

- Skip expensive 256bit multiplication when factor is 1 (which it is currently)
- Use `ulong` iterator rather than `UInt256` iterator which has much more expensive increments
- Use `.IsZero` rather than `> 0` which is faster as it just vector checks everything is 0

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Evm/BlobGasCalculator.cs
+++ b/src/Nethermind/Nethermind.Evm/BlobGasCalculator.cs
@@ -50,35 +50,40 @@ public static class BlobGasCalculator
     {
         static bool FakeExponentialOverflow(UInt256 factor, UInt256 num, UInt256 denominator, out UInt256 feePerBlobGas)
         {
-            UInt256 output = UInt256.Zero;
-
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Evm/BlobGasCalculator.cs
```diff
@@ -50,35 +50,40 @@ public static bool TryCalculateFeePerBlobGas(ulong excessBlobGas, UInt256 blobGa
...
+            }
+            else if (UInt256.MultiplyOverflow(factor, denominator, out accumulator))
             {
                 feePerBlobGas = UInt256.MaxValue;
                 return true;
             }
 
-            for (UInt256 i = 1; numAccum > 0; i++)
+            UInt256 output = default;
+            for (ulong i = 1; !accumulator.IsZero; i++)
```
LukaszRozmej, line 66: in practice accumulator should never go pass `ulong`, should we have separate path for it?

LukaszRozmej (approved):

PR #8175, opened 2025-02-06
check requests type validity

rjnrohit:
Validate if the execution requests have invalid type


rjnrohit:
> LGTM, but for future try to cover such checks with tests

okay

PR #8179, opened 2025-02-07
Shorten boolean command line option notation

rubo:
## Changes

- Made the `true` value optional for boolean command line options, reducing clutter in configs.
  For example, `--metrics-enabled true` turns to just `--metrics-enabled`.
  This change brings Nethermind CLI closer to the standard behavior people would expect. Geth behaves the same with its `--http` flag. Besu as well.
  _This style is optional and non-breaking, keeping full backward compatibility._
- Updated the docs generator to reflect the changes.

- [ ] Optimization
- [ ] Refactoring
- [x] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

Needs manual testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No


tools/DocGen/ConfigGenerator.cs
```diff
@@ -100,27 +100,40 @@ private static void WriteMarkdown(StreamWriter file, Type configType)
...
+            if (prop.PropertyType == typeof(bool))
+            {
+                value = "true|false";
+                cliValue = $"[{value}]";
+            }
+            else
+            {
+                value = "<value>";
+                cliValue = value;
+            }
```
LukaszRozmej, line 115: Maybe extract to a method, result as (,) tuple and use deconstruct?

LukaszRozmej (approved):

PR #8183, opened 2025-02-10
Cleanup/remove last seen map

asdacap:
- The `_persistedLastSeen` map is used to keep track of recommitted persisted trie node so that it is not removed from DB which is then later not re-added causing trie exception.
- A simpler way is to simply not prune the persisted node from the cache if it is still needed which is also checked if a node was recommited. This remove the `_persistedLastSeen` map and the code related to it. This keep slightly more node in cached which has the benifit of slightly better cache hit rate at expense of more prune cycle.
- The effect is negligable in every way, but it cut code, so that is nice.

- [X] Optimization?
- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Can replay blocks from era of several weeks worth of blocks.

### DIFF
--- a/src/Nethermind/Nethermind.Trie.Test/Pruning/TreeStoreTests.cs
+++ b/src/Nethermind/Nethermind.Trie.Test/Pruning/TreeStoreTests.cs
@@ -941,11 +941,11 @@ namespace Nethermind.Trie.Test.Pruning
 
             if (_scheme == INodeStorage.KeyScheme.Hash)
             {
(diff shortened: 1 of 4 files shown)


LukaszRozmej (approved):

benaadams (approved):

benaadams:
The effect is negligable in every way, but it cut code, so that is nice.

95MB saved?

image

PR #8184, opened 2025-02-10
Fix some unit test

asdacap:
- These test fails more consistently so it could just be my setup.

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No


src/Nethermind/Nethermind.Core.Test/McsPriorityLock.cs
```diff
@@ -11,18 +11,18 @@ namespace Nethermind.Core.Test;
 
 public class McsPriorityLockTests
 {
-    private McsPriorityLock mcsLock;
+    private McsPriorityLock _mcsLock;
```
LukaszRozmej, line 14: Can you rename the file to end with `Tests.cs`?

src/Nethermind/Nethermind.Sockets/SocketClient.cs
```diff
@@ -15,15 +15,23 @@ public class SocketClient<TStream> : ISocketsClient where TStream : Stream, IMes
...
     public string ClientName { get; }
 
-    public SocketClient(string clientName, TStream stream, IJsonSerializer jsonSerializer)
+    public SocketClient(string clientName, TStream stream, IJsonSerializer jsonSerializer) :
+        // Optional maxRequestSize break some tests.
+        this(clientName, stream, jsonSerializer, MAX_REQUEST_BODY_SIZE_FOR_ENGINE_API)
+    {
+    }
+
+    public SocketClient(string clientName, TStream stream, IJsonSerializer jsonSerializer, int maxRequestSize)
```
LukaszRozmej, line 29: No need for 2 constructors?
```suggestion
    public SocketClient(string clientName, TStream stream, IJsonSerializer jsonSerializer, int maxRequestSize = MAX_REQUEST_BODY_SIZE_FOR_ENGINE_API)
```

LukaszRozmej (approved):

PR #8192, opened 2025-02-11
Add Nimbus-eth1 & EthereumJS to client types

benaadams:
## Changes

- Add the Nimbus-eth1 EL client to client types
- Add EthereumJS to client types 

Shows less Unknown in peers

image

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Network.Stats/Model/Node.cs
+++ b/src/Nethermind/Nethermind.Network.Stats/Model/Node.cs
@@ -213,6 +213,10 @@ namespace Nethermind.Stats.Model
             {
                 return NodeClientType.Reth;
             }
+            else if (clientId.Contains(nameof(NodeClientType.Nimbus), StringComparison.OrdinalIgnoreCase))
+            {
+                return NodeClientType.Nimbus;
+            }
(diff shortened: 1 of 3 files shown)


src/Nethermind/Nethermind.Network.Stats/Model/Node.cs
```diff
@@ -213,6 +213,14 @@ public static NodeClientType RecognizeClientType(string clientId)
...
                 return NodeClientType.Reth;
             }
+            else if (clientId.Contains(nameof(NodeClientType.Nimbus), StringComparison.OrdinalIgnoreCase))
+            {
+                return NodeClientType.Nimbus;
+            }
+            else if (clientId.Contains(nameof(NodeClientType.EthereumJS), StringComparison.OrdinalIgnoreCase))
+            {
+                return NodeClientType.EthereumJS;
+            }
```
LukaszRozmej, line 223: Can we do it based on enum and not if-else? (Can use FastEnum)

benaadams: Not sure I see how?

e.g. it wants to turn `geth go1.23.3 linux` to `NodeClientType.Geth` which isn't a straight parse

PR #8196, opened 2025-02-12
Refactor/isolate main transaction processor

asdacap:
- I've gotten question before on why `IBlockValidator` cannot accept `ITransactionProcessor` from `INethermindApi`.
- You can't because transaction runs on top of a world state somehow, and `IBlockValidator` is used in `BlockDownloader` for example and when you call it, which `IWorldState` is it running on top of? The last processed block may not be exactly the currrently being validated block in `BlockDownloader`.
- To prevent further confusion, this PR moved `ITransactionProcessor`, `IBlockProcessor` and `IBlockchainProcessor` (for the main block validation, not block building or RPC, see how this is a problem?) from `INethermindApi` into `IMainProcessingContext`. 
- Also, the methods in `InitializeBlockchain` explicitly pass `ITransactionProcessor` and `IWorldState` now. 
- Also, taiko block building seems to be using the main transaction processor. Need to double check to see if that is intentional.

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

#### Notes on testing

- Unit tests still passed.


src/Nethermind/Nethermind.Taiko/TaikoPlugin.cs
```diff
@@ -166,8 +166,8 @@ public async Task InitRpcModules()
                 new BlockInvalidTxExecutor(new BuildUpTransactionProcessorAdapter(scope.TransactionProcessor), scope.WorldState),
                 scope.WorldState,
                 _api.ReceiptStorage,
-                _api.TransactionProcessor,
-                new BeaconBlockRootHandler(_api.TransactionProcessor, _api.WorldStateManager.GlobalWorldState),
+                scope.TransactionProcessor,
```
asdacap, line 169: Which one is this? Its used by `TaikoPreparationService`. There is an explicitly created read only env also, but that is not used. The payload preparation service uses block processor instead of block builder?

src/Nethermind/Nethermind.Consensus.AuRa/InitializationSteps/StartBlockProducerAuRa.cs
```diff
@@ -158,7 +158,7 @@ private BlockProcessor CreateBlockProcessor(IReadOnlyTxProcessingScope changeabl
             _api.LogManager,
             _api.BlockTree,
             NullWithdrawalProcessor.Instance,
-            _api.TransactionProcessor,
+            changeableTxProcessingEnv.TransactionProcessor,
```
asdacap, line 161: This is different. Block producer using main transaction processor originally.

src/Nethermind/Nethermind.Core.Test/Modules/MainBlockProcessingContext.cs
```diff
@@ -20,13 +21,14 @@ namespace Nethermind.Core.Test.Modules;
 /// <param name="GenesisLoader"></param>
 public record MainBlockProcessingContext(
     ILifetimeScope LifetimeScope,
-    BlockchainProcessor BlockchainProcessor,
+    BlockchainProcessor BlockchainProcessorImpl,
```
LukaszRozmej, line 24: Can we change it to?

```suggestion
    IBlockchainProcessor BlockchainProcessor,
```

LukaszRozmej (approved):

src/Nethermind/Nethermind.Core.Test/Modules/MainBlockProcessingContext.cs
asdacap: Not in this case I'm afraid, as `IBlockchainProcessor` does not implement `IProcessingQueue`.

src/Nethermind/Nethermind.Taiko/TaikoPlugin.cs
flcl42: Looks like it should be writable here, because blockProcessor is used blockchainProcessor, maybe we need to change BlockhashStore/BlockProductionWithdrawalProcessor to writable too

asdacap: Ok then, I don't know why it is using the main state, but as long as it is intended then its fine.

flcl42: Or maybe you are right, it does not replace main blockchain processor, probably that scope is needed everywhere

asdacap: I'm not sure I follow. I'll just revert this part of the code and file an issue.

src/Nethermind/Nethermind.Core.Test/Caching/LruKeyCacheTests.cs
```diff
@@ -174,7 +174,8 @@ public void Beyond_capacity_lru_check()
         public void Beyond_capacity_lru_parallel()
         {
             LruKeyCache<AddressAsKey> cache = new(Capacity, Capacity / 2, "test");
-            Parallel.For(0, Math.Min(Environment.ProcessorCount * 8, 64), (iter) =>
+            int processorCount = Math.Min(Environment.ProcessorCount, 32);
```
flcl42, line 177: 32 but 4 might be supposed

PR #8203, opened 2025-02-13
Refactor/move sync pivot to IBlockTree.

asdacap:
- Move sync pivot from `ISyncConfig` into `IBlockTree` and ancient block barriers so `ISyncPointers`.
  - Why into `IBlockTree`? Well, blocktree initialization need it so to prevent adding a new dependency to it, just put it in `IBlockTree`.
  - Plus if you think of it as `FinalizedBlock`, which it kinda is, it make a lot more sense.
- Why ancient block barriers into `ISyncPointers`? Because `IBlockTree` is a bit fat, and the barriers are more of a sync concern, less important than sync pivot. 
  - Combined portal and sharded block, direct era attached storage, the block barriers have a lot of logic.
  - And since ancient barriers rely on sync pivot for some reason, it kinda need to be put outside.

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- [X] Hive sync test passed


asdacap:
Will redo

PR #8209, opened 2025-02-14
4444-ish: Cleanup/disable header only forward sync

asdacap:
- Post merge block downloader originally does not allow header only forward sync. I've modified it to allow it, which then reveal an undisired behaviour where it does not switch to full sync unless the state sync finished at exactly the best suggested header.
  - This is because, the full sync will start exactly at the block after it the best suggested header (downloaded by the header only forward fast sync), which is also the downloaded state, so it need to be exact. Unless bodies is also downloaded, in which case `IBlockchainProcessor` will process its parents also.
  - Also state sync should only finish once now.
  - Anyway, I'm inclined to just remove the header only forward sync as per previous post merge behaviour as it significantly clears up which part of the code is PoW header related.
- Additionally, this ensure that pos SyncPivot, all block is available, an assumption that make it simple to reason about when `SyncPivot` is movable.
- Additionally, this make the behaviour of why starting fast sync with  `--Sync.DownloadBodiesInFastSync false`, then restart with `true` after finished state sync will result in gap of block after the sync pivot to no longer be the case. The same is true with receipts, which I'm also inclined to just always download receipt for block after `SyncPivot`.
- Fix on long range state sync (restart on very old node) `BlockchainProcessor` will load all pending blocks anyway resulting in a very high memory and processing a lot of blocks.

- [X] Bugfix?
- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- [x] Mainnet sync
- [x] Mainnet full sync
- [X] Long range state sync catch up still works.
- [X] Energyweb sync
- [x] Hive engine tests.
- [x] Sync github action.


src/Nethermind/Nethermind.Synchronization/Blocks/FastSyncFeed.cs
```diff
@@ -25,18 +25,7 @@ public FastSyncFeed(ISyncConfig syncConfig)
...
-                options |= DownloaderOptions.WithReceipts;
-            }
-
-            if (_syncConfig.DownloadBodiesInFastSync)
-            {
-                options |= DownloaderOptions.WithBodies;
-            }
-
-            return options;
+            return DownloaderOptions.Insert | DownloaderOptions.WithReceipts;
```
LukaszRozmej, line 28: Why does it simplify?

src/Nethermind/Nethermind.Synchronization/ParallelSync/MultiSyncModeSelector.cs
```diff
@@ -432,17 +431,15 @@ private bool ShouldBeInFullSyncMode(Snapshot best)
...
 
             // Shared with full sync archive
             bool desiredPeerKnown = AnyDesiredPeerKnown(best);
 
             // Full sync specific
-            bool hasFastSyncBeenActive = best.Header >= _pivotNumber;
+            bool hasFastSyncBeenActive = best.Header >= best.PivotNumber;
             bool notInFastSync = !best.IsInFastSync;
             bool notInStateSync = !best.IsInStateSync;
-            bool stateSyncFinished = best.State > 0
```
LukaszRozmej, line 444: no need to check for that?

src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
```diff
@@ -643,41 +639,29 @@ private ProcessingBranch PrepareProcessingBranch(Block suggestedBlock, Processin
...
-
             // TODO: there is no test for the second condition
             // generally if we finish fast sync at block, e.g. 8 and then have 6 blocks processed and close Neth
             // then on restart we would find 14 as the branch head (since 14 is on the main chain)
             // we need to dig deeper to go all the way to the false (reorg boundary) head
             // otherwise some nodes would be missing
             bool notFoundTheBranchingPointYet = !_blockTree.IsMainChain(branchingPoint.Hash!);
-            bool notReachedTheReorgBoundary = branchingPoint.Number > (_blockTree.Head?.Header.Number ?? 0);
+            bool notReachedTheReorgBoundary = branchingPoint.Number > (_blockTree.Head?.Header.Number);
+            bool hasState = toBeProcessed?.StateRoot is null || _stateReader.HasStateForBlock(toBeProcessed.Header);
```
LukaszRozmej, line 649: Can we make the logic here, so it would ALWAYS go deeper until it finds state or until it goes past our `Pruning.PersistenceInterval`? That recover so many issues. Would allow us to recover from long reorgs, would allow us to SetHead to random block on prunned nodes - node would figure out where it stored last state root.

src/Nethermind/Nethermind.Synchronization/Blocks/FastSyncFeed.cs
asdacap: The receipt one, it does not make much of a difference, but it does make it less confusing why if you start a node without old block and receipts, and then restart with old block and receipts it would have a gap after the sync pivot.

src/Nethermind/Nethermind.Synchronization/ParallelSync/MultiSyncModeSelector.cs
asdacap: Should be covered by fast sync and state sync.

src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
asdacap: Done

src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
```diff
@@ -576,13 +576,17 @@ private ProcessingBranch PrepareProcessingBranch(Block suggestedBlock, Processin
...
-        bool suggestedBlockIsPostMerge = suggestedBlock.IsPostMerge;
 
         Block toBeProcessed = suggestedBlock;
         long iterations = 0;
         do
         {
             iterations++;
+            if (iterations > MaxBranchSize)
+            {
+                throw new InvalidOperationException($"Maximum size of branch reached ({MaxBranchSize}). This is unexpected.");
```
MarekM25, line 587: What if we want to do reorg with more blocks? I know that it is currently impossible, but we should aim to be able reorg to the last finalized block and we might have long non-finality events like in Holesky

MarekM25 (commented):
Have you tested it on a pre-merge network? EnergyWeb or Joc?

src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
asdacap: Increase MaxBranchSize?

asdacap: Actually, assuming that a beacon sync was initiated, this should not get called as block downloader would suggest the block one by one.

src/Nethermind/Nethermind.Synchronization/Blocks/BlockDownloader.cs
```diff
@@ -296,7 +207,7 @@ public virtual async Task<long> DownloadBlocks(PeerInfo? bestPeer, BlocksRequest
                         blocksSynced++;
                     }
 
-                    currentNumber += 1;
+                    _currentNumber += 1;
```
benaadams, line 210: Could this get out of sync with actual block numbers being requested?

benaadams (approved):

src/Nethermind/Nethermind.Synchronization/Blocks/BlockDownloader.cs
asdacap: Line 155 reset it and only one of this `DownloadBlocks` is executed at any time.

src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
```diff
@@ -72,7 +72,7 @@ public sealed class BlockchainProcessor : IBlockchainProcessor, IBlockProcessing
 
     private int _currentRecoveryQueueSize;
     private bool _isProcessingBlock;
-    private const int MaxBlocksDuringFastSyncTransition = 8192;
+    private const int MaxBranchSize = 8192;
```
LukaszRozmej, line 75: Should this be same as `Reorganization.PersistanceInterval` or maybe better `PruningConfig.PersistanceInterval`?

LukaszRozmej: We are even using `Reorganization.PersistenceInterval)` in `PrepareProcessingBranch` here

src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
LukaszRozmej: By keeping `notFoundTheBranchingPointYet` in line 635 we can break the loop even if we didn't find the state but processed the block earlier, which is exactly the case I wanted to handle. Let's ignore being on main chain if we don't have the state, until at least `PruningConfig.PersistanceInterval` is reached?

PR #8214, opened 2025-02-16
Optimize BlockchainProcessor Channel use

benaadams:
## Changes

- Optimize BlockchainProcessor Channel use
   - Single reader (lighter synchronisation)

- [x] Optimization

#### Requires testing

- [x] No


benaadams:
Makes tests unhappy :(

LukaszRozmej:
Did you investigate tests? I think there are only 2 type of failures there

benaadams:
> Did you investigate tests? I think there are only 2 type of failures there

Made .Count property invalid for one of the Channels; follow up https://github.com/NethermindEth/nethermind/pull/8218

PR #8216, opened 2025-02-17
fix gnosis state / blockchain tests runner

yerke26:
Gnosis state, blockchain tests runner is not configured properly
## Changes

- use CancunGnosis, PragueGnosis spec instead of Cancun and Prague for gnosis tests
- fix AuthorizationListForRpc where Signature was constructed wrong for AuthorizationTuple

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Remarks


src/Nethermind/Ethereum.Test.Base/BlockchainTestBase.cs
```diff
@@ -78,16 +78,16 @@ protected async Task<EthereumTestResult> RunTest(BlockchainTest test, Stopwatch?
             TestContext.Out.WriteLine($"Network after transition: [{test.NetworkAfterTransition.Name}] at {test.TransitionForkActivation}");
         Assert.That(test.LoadFailure, Is.Null, "test data loading failure");
 
+        var genesisSpec = test.ChainId == GnosisSpecProvider.Instance.ChainId ? GnosisSpecProvider.Instance.GenesisSpec : MainnetSpecProvider.Instance.GenesisSpec;
```
LukaszRozmej, line 81: Can we encapsulate that logic in some spec loader, rather than passing chain id everywhere?

yerke26: yes, I created ChainUtils and adjust spec to gnosis spec there

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/AuthorizationListForRpc.cs
```diff
@@ -64,7 +64,7 @@ public AuthorizationTuple[] ToAuthorizationList() => _tuples
             (ulong)tuple.ChainId,
             tuple.Address,
             tuple.Nonce,
-            new Signature(tuple.R, tuple.S, (ulong)tuple.YParity))
+            new Signature(tuple.R, tuple.S, (ulong)tuple.YParity + Signature.VOffset))
```
yerke26, line 67: Is this correct? I changed because in other constructor it adds `Signature.VOffset` here: https://github.com/NethermindEth/nethermind/blob/master/src/Nethermind/Nethermind.Core/AuthorizationTuple.cs#L22
This change help me to solve problem with EEST generation

src/Nethermind/Ethereum.Test.Base/GeneralTestBase.cs
```diff
@@ -62,10 +64,12 @@ protected EthereumTestResult RunTest(GeneralStateTest test, ITxTracer txTracer)
             TestContext.Out.Write($"Running {test.Name} at {DateTime.UtcNow:HH:mm:ss.ffffff}");
             Assert.That(test.LoadFailure, Is.Null, "test data loading failure");
 
-            ISpecProvider specProvider = test.ChainId == GnosisSpecProvider.Instance.ChainId
-                ? GnosisSpecProvider.Instance
-                : new CustomSpecProvider(
-                    ((ForkActivation)0, Frontier.Instance), // TODO: this thing took a lot of time to find after it was removed!, genesis block is always initialized with Frontier
+            ChainUtils.AdjustSpecToGnosisChain(test.Fork, test.ChainId);
```
LukaszRozmej, line 67: You are not doing anything with the result.
Maybe pass the whole `test` and adjust internally?

Call it: `ResolveSpec`, can be made extension method on `test`

src/Nethermind/Ethereum.Test.Base/GeneralTestBase.cs
```diff
@@ -62,10 +64,12 @@ protected EthereumTestResult RunTest(GeneralStateTest test, ITxTracer txTracer)
...
+
+            var genesisSpec = ChainUtils.GetGenesisSpec(test.ChainId);
```
LukaszRozmej, line 69: Then `test` could have a property for GenesisSpec too and all can be done in one call?

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/AuthorizationListForRpc.cs
LukaszRozmej: @ak88 @smartprogrammer93 can you check ASAP?

ak88: @yerke26 that should be correct yes. `Signature` is expecting the V value to be passed, and not the raw y parity.

Here is a bit more information about V if you are interested:
https://bitcoin.stackexchange.com/questions/38351/ecdsa-v-r-s-what-is-v/38909#38909

And [EIP-155](https://eips.ethereum.org/EIPS/eip-155) explains the relation between V and chain id for legacy tx.

src/Nethermind/Ethereum.Test.Base/GeneralTestBase.cs
yerke26: yes, I think IEthereumTest interface should be abstract class, and added common ChainId and GenesisSpec properties to the EthereumTest abstract class in order to not copy in child classes: GeneralStateTest, BlockchainTest

src/Nethermind/Ethereum.Test.Base/GeneralTestBase.cs
yerke26: fixed, left as it is because passing test and adjusting internally is not right idea, because fork is named `Network`, `NetworkAfterTransition` in `BlockchainTest` and `Fork` in `GeneralStateTest`

LukaszRozmej (approved):

PR #8218, opened 2025-02-18
Blockchainprocessor channel use

benaadams:
## Changes

- Optimize BlockchainProcessor Channel use
   - Single reader (lighter synchronisation)
   - Continue on same thread (don't requeue to threadpool, carry priority)

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
@@ -43,11 +43,27 @@ public sealed class BlockchainProcessor : IBlockchainProcessor, IBlockProcessing
     private readonly IBlockTree _blockTree;
     private readonly ILogger _logger;
 
-    private readonly Channel<BlockRef> _recoveryQueue = Channel.CreateUnbounded<BlockRef>();
+    private readonly Channel<BlockRef> _recoveryQueue = Channel.CreateUnbounded<BlockRef>(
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
```diff
@@ -386,7 +402,7 @@ private void FireProcessingQueueEmpty()
         {
             long blockProcessingTimeInMicrosecs = _stopwatch.ElapsedMicroseconds();
             Metrics.LastBlockProcessingTimeInMs = blockProcessingTimeInMicrosecs / 1000;
-            Metrics.RecoveryQueueSize = _recoveryQueue.Reader.Count;
+            Metrics.RecoveryQueueSize = _queueCount;
```
LukaszRozmej, line 405: I don't think this is correct. `_queueCount` is for the whole queue, so `Recovery+Processing+(isProcessingOneNow ? 1 : 0)`

LukaszRozmej (approved):

rjnrohit (approved):

src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
benaadams: Is still wrong, at least at time when logs as processing decrements; will do follow up fix

recover     1 | process     0

PR #8221, opened 2025-02-18
Trie visitor adjustment

asdacap:
Fixes Closes Resolves #

- Convert all `ITrieVisitor` to `ITrieVisitorr<OldStyleTrievisitContext>`
- Remove `TrieVisitContext` from `ITrieVisitor<TCtx>`.
- Fix missed some storage due to concurrent access to `IsStorage` in `TrieVisitContext` by removing `TrieVisitContext` from the main visitor code.
- Encapsulate visitor code to `TrieNodeTraverser`.
- Increase GC threshold from 1Mil to 100Mil nodes.

- [X] Optimization
- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes?
- [ ] No

#### Notes on testing

- VerifyTrie stats is now consistent.

### DIFF
--- a/src/Nethermind/Nethermind.Trie/TrieNode.Visitor.cs
+++ b/src/Nethermind/Nethermind.Trie/TrieNode.Visitor.cs
@@ -292,7 +292,7 @@ namespace Nethermind.Trie
 
                         TNodeContext leafContext = nodeContext.Add(Key!);
 
-                        if (!trieVisitContext.IsStorage && trieVisitContext.ExpectAccounts) // can combine these conditions
+                        if (!trieVisitContext.IsStorage && visitor.ExpectAccounts) // can combine these conditions
                         {
                             Account account = _accountDecoder.Decode(Value.AsRlpStream());
                             if (account.HasCode && visitor.ShouldVisit(leafContext, account.CodeHash))
--- a/src/Nethermind/Nethermind.Trie/TrieNode.Visitor.cs
+++ b/src/Nethermind/Nethermind.Trie/TrieNode.Visitor.cs
@@ -2,9 +2,7 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 
 using System;
-using System.Collections.Generic;
 using System.IO;
-using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Threading.Tasks;
 using Nethermind.Core;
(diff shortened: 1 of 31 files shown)


src/Nethermind/Nethermind.Trie/TrieNode.Visitor.cs
```diff
@@ -289,58 +100,155 @@ static void VisitAllSingleThread(TrieNode currentNode, ref TreePath path, ITreeV
...
+            }
+        }
+
+        private void VisitAllSingleThread(TrieNode currentNode, ref TreePath path, TNodeContext nodeContext, ITrieNodeResolver nodeResolver, bool isStorage)
+        {
+            using ArrayPoolList<TrieNode?> output = new ArrayPoolList<TrieNode>(TrieNode.BranchesCount, TrieNode.BranchesCount);
+            currentNode.ResolveAllChildBranch(nodeResolver, ref path, output.AsSpan());
+
+            path.AppendMut(0);
+            for (int i = 0; i < 16; i++)
```
LukaszRozmej, line 178: TrieNode.BranchesCount

LukaszRozmej (commented):
Can we just migrate `OldStyleTrieVisitContext` visitors to new one? Do we even need a generic context then?

src/Nethermind/Nethermind.Trie/TrieNode.Visitor.cs
```diff
@@ -289,58 +100,155 @@ static void VisitAllSingleThread(TrieNode currentNode, ref TreePath path, ITreeV
...
+            });
+
+        private void AddVisited()
+        {
+            int visitedNodes = Interlocked.Increment(ref _visitedNodes);
+
+            // TODO: Fine tune interval? Use TrieNode.GetMemorySize(false) to calculate memory usage?
+            if (visitedNodes % 100_000_000 == 0)
+            {
+                GC.Collect();
```
LukaszRozmej, line 251: How does this change memory usage of a node?

asdacap: It should use more I guess, although at the rate of tree iteration nowadays, it should get GCed once every few minute or so, so it should be fine.

asdacap:
> Can we just migrate `OldStyleTrieVisitContext` visitors to new one? Do we even need a generic context then?

`OldStyleTrieVisitContext` is the new interface. I guess individual visitor can be rewrote to use custom context for efficiency reason, but that is for another time perhaps for intern or external contributer.

src/Nethermind/Nethermind.Trie/TrieNode.Visitor.cs
LukaszRozmej: Can you share some graph of memory usage?

LukaszRozmej:
> > Can we just migrate `OldStyleTrieVisitContext` visitors to new one? Do we even need a generic context then?
> 
> `OldStyleTrieVisitContext` is the new interface. I guess individual visitor can be rewrote to use custom context for efficiency reason, but that is for another time perhaps for intern or external contributer.

but do we really need multiple contexts?

asdacap:
Yes? We need multiple of them? Thats why it become as is.

LukaszRozmej (approved):

src/Nethermind/Nethermind.Trie/TrieNode.Visitor.cs
asdacap: Screenshot_2025-02-25_15-50-40

1 Mil, 100 Mil and 10 Mil. Twice peak throughput though.

PR #8231, opened 2025-02-19
Limit task count in tests

asdacap:
- Unit tests slows down with high core count machine which seems to be due to the spawing of tasks.
- This PR limits those tasks creation.

## Changes

- Add config to limit prewarming task count.
- Reduce default limit of concurrent task for dispatcher in tests.
- Disable `AllowSynchronousContinuations` as it randomly deadlock tests.

- [X] Optimization

#### Requires testing

- [ ] Yes
- [X] No

- [X] Yes?
- [ ] No

#### Notes on testing

- Seems to be faster but still not passing completely.

### DIFF
--- a/src/Nethermind/Nethermind.Config/BlocksConfig.cs
+++ b/src/Nethermind/Nethermind.Config/BlocksConfig.cs
@@ -24,6 +24,7 @@ namespace Nethermind.Config
         public ulong SecondsPerSlot { get; set; } = 12;
 
         public bool PreWarmStateOnBlockProcessing { get; set; } = true;
+        public int PreWarmStateConcurrency { get; set; } = 0;
 
         public int BlockProductionTimeoutMs { get; set; } = 4_000;
 
--- a/src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
@@ -48,8 +48,6 @@ public sealed class BlockchainProcessor : IBlockchainProcessor, IBlockProcessing
         {
             // Optimize for single reader concurrency
             SingleReader = true,
-            // Share thread with request, if waiting
-            AllowSynchronousContinuations = true
         });
 
     private readonly Channel<BlockRef> _blockQueue = Channel.CreateBounded<BlockRef>(
@@ -59,8 +57,6 @@ public sealed class BlockchainProcessor : IBlockchainProcessor, IBlockProcessing
             SingleReader = true,
             // Optimize for single writer concurrency (recovery queue)
             SingleWriter = true,
-            // Share thread with request, if waiting
-            AllowSynchronousContinuations = true
         });
 
     private bool _recoveryComplete = false;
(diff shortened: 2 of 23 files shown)


src/Nethermind/Nethermind.Config/BlocksConfig.cs
```diff
@@ -24,6 +24,7 @@ public class BlocksConfig : IBlocksConfig
         public ulong SecondsPerSlot { get; set; } = 12;
 
         public bool PreWarmStateOnBlockProcessing { get; set; } = true;
+        public int PreWarmStateConcurrency { get; set; } = 0;
```
LukaszRozmej, line 27: This could replace `PreWarmStateOnBlockProcessing` flag if set to negative?

src/Nethermind/Nethermind.Trie/TrieNode.Decoder.cs
```diff
@@ -166,7 +166,7 @@ private static int GetChildrenRlpLengthForBranchParallel(ITrieNodeResolver tree,
             private static int GetChildrenRlpLengthForBranchNonRlpParallel(ITrieNodeResolver tree, TreePath rootPath, TrieNode item, ICappedArrayPool bufferPool)
             {
                 int totalLength = 0;
-                ParallelUnbalancedWork.For(0, BranchesCount, RuntimeInformation.ParallelOptionsLogicalCores,
+                ParallelUnbalancedWork.For(0, BranchesCount, RuntimeInformation.ParallelOptionsLogicalCoresUpTo16,
```
LukaszRozmej, line 169: why up to 16?

src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
```diff
@@ -48,8 +48,6 @@ public sealed class BlockchainProcessor : IBlockchainProcessor, IBlockProcessing
         {
             // Optimize for single reader concurrency
             SingleReader = true,
-            // Share thread with request, if waiting
-            AllowSynchronousContinuations = true
```
LukaszRozmej, line 52: Not convinced about this one, I think it should work well in production when we only get single request? @benaadams

src/Nethermind/Nethermind.Config/BlocksConfig.cs
asdacap: Sure. That would be a possible user facing breaking change though. Is that fine?

src/Nethermind/Nethermind.Trie/TrieNode.Decoder.cs
asdacap: No more than 16 child.

src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
asdacap: This one in particular reliably locks `BlockchainProcessorTests` when run on its own. I don't recommend these kind per block optimization, it took a lot of time to trace this kind of problem.

benaadams: Bigger issue is test parameter discovery triggering all the static test parameter values, which then start blockchains, when then setup background pruning tasks etc, which then soak up all the threads (as happening with many tests in parallel), just to decide it needs to put 0x521 in as a parameter to a test case 🤷‍♂️

src/Nethermind/Nethermind.Config/BlocksConfig.cs
LukaszRozmej: @rubo ?

src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
LukaszRozmej: @benaadams I have understood none of that. I will leave it up to you.

LukaszRozmej (approved):

PR #8233, opened 2025-02-20
Refactor ShutterTime to SlotTime

Marchhill:
## Changes

- Renamed `ShutterTime` to `SlotTime` and moved to `Nethermind.Core`. Since this class is not specific to Shutter functionality this could better facilitate code reuse

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Core/SlotTime.cs
+++ b/src/Nethermind/Nethermind.Core/SlotTime.cs
@@ -5,7 +5,7 @@ using System;
 
 namespace Nethermind.Core;
 
-public class SlotTime(ulong genesisTimestampMs, ITimestamper timestamper, TimeSpan slotLength, TimeSpan blockUpToDateCutoff)
+public readonly struct SlotTime(ulong genesisTimestampMs, ITimestamper timestamper, TimeSpan slotLength, TimeSpan blockUpToDateCutoff)
 {
     public class SlotCalulationException(string message, Exception? innerException = null) : Exception(message, innerException);
 
--- a/src/Nethermind/Nethermind.Core/SlotTime.cs
+++ b/src/Nethermind/Nethermind.Core/SlotTime.cs
@@ -2,13 +2,12 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 
 using System;
-using Nethermind.Core;
 
-namespace Nethermind.Shutter;
+namespace Nethermind.Core;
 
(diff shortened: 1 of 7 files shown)


src/Nethermind/Nethermind.Core/SlotTime.cs
```diff
@@ -2,13 +2,12 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 
 using System;
-using Nethermind.Core;
 
-namespace Nethermind.Shutter;
+namespace Nethermind.Core;
 
-public class ShutterTime(ulong genesisTimestampMs, ITimestamper timestamper, TimeSpan slotLength, TimeSpan blockUpToDateCutoff)
+public class SlotTime(ulong genesisTimestampMs, ITimestamper timestamper, TimeSpan slotLength, TimeSpan blockUpToDateCutoff)
```
LukaszRozmej, line 8:
```suggestion
public readonly struct SlotTime(ulong genesisTimestampMs, ITimestamper timestamper, TimeSpan slotLength, TimeSpan blockUpToDateCutoff)
```

or at least `readonly class`

LukaszRozmej (approved):

src/Nethermind/Nethermind.Core/SlotTime.cs
LukaszRozmej: If we go with readonly struct maybe pass it with `in` parameter where possible?

Marchhill: If we did this we may as well make a class right? Also since it's just one `ulong` I think it makes sense to pass by value, the overhead could be greater from indirection when passing by reference

LukaszRozmej: well it is also ITimestamper reference  and 2 timespans

Marchhill: Ah right I can make a new PR. Is there any reason not to make a class?

Marchhill: Ah right I can make a new PR. Is there any reason not to make a class?

PR #8237, opened 2025-02-21
Change FCU update head message to be less confusing

LukaszRozmej:
## Changes

- Rename `Synced Chain Head` -> `Updated Chain Head` to not confuse users that we are still syncing. User was confused his node is not yet synced.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [x] Other: logs

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
+++ b/src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
@@ -284,7 +284,7 @@ public class ForkchoiceUpdatedHandler : IForkchoiceUpdatedHandler
         if (shouldUpdateHead)
         {
             _poSSwitcher.ForkchoiceUpdated(newHeadBlock.Header, finalizedBlockHash);
-            if (_logger.IsInfo) _logger.Info($"Synced Chain Head to {newHeadBlock.ToString(Block.Format.Short)}");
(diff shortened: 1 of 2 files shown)


MarekM25 (commented):
Hmm, not sure. 'Synced' might make users feel calm.

src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
```diff
@@ -284,7 +284,7 @@ protected virtual bool IsOnMainChainBehindHead(Block newHeadBlock, ForkchoiceSta
         if (shouldUpdateHead)
         {
             _poSSwitcher.ForkchoiceUpdated(newHeadBlock.Header, finalizedBlockHash);
-            if (_logger.IsInfo) _logger.Info($"Synced Chain Head to {newHeadBlock.ToString(Block.Format.Short)}");
+            if (_logger.IsInfo) _logger.Info($"Updated Chain Head to {newHeadBlock.ToString(Block.Format.Short)}");
```
benaadams, line 287: Adds an extra char to the logs; so now the block number for Updated will be out of alignment with all the other logs

image

Either need to add an additional space to all those logs or use a word with one less character

benaadams: Everything aligned (red line)

image

PR #8240, opened 2025-02-21
Add message with exact error to `AcceptTxResult.Invalid`

marcindsobczak:
## Changes

- adding message with exact error to `AcceptTxResult.Invalid`
so there will be e.g.:
`Invalid, BlobTxGasLimitExceeded: Transaction's totalDataGas=393216 exceeded MaxBlobGas per transaction=262144.`
instead of just:
`Invalid`

We need to ensure that it will be compatible with our RPC users as error code will be changed (extended) now.

## Types of changes

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.JsonRpc.Test/Modules/Eth/EthRpcModuleTests.cs
+++ b/src/Nethermind/Nethermind.JsonRpc.Test/Modules/Eth/EthRpcModuleTests.cs
@@ -1070,7 +1070,7 @@ public partial class EthRpcModuleTests
         string serialized = await ctx.Test.TestEthRpc("eth_sendRawTransaction", rawTransaction);
         Transaction tx = Rlp.Decode<Transaction>(Bytes.FromHexString(rawTransaction));
         await txSender.Received().SendTransaction(tx, TxHandlingOptions.PersistentBroadcast);
-        Assert.That(serialized, Is.EqualTo("{\"jsonrpc\":\"2.0\",\"error\":{\"code\":-32010,\"message\":\"Invalid\"},\"id\":67}"));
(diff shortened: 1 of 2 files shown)


LukaszRozmej (approved):

benaadams (approved):

emlautarom1 (approved):

Scooletz (commented):
> We need to ensure that it will be compatible with our RPC users as error code will be changed (extended) now.

Have you ensured it? This looks like a potential breaking change especially if the callers assume one of given values.  I'd delegate it to a separate field like `Reason: InvalidTxSignature: Signature is invalid` or similar so that the previously existing property preserves it semantics and given set of values.

marcindsobczak:
> > We need to ensure that it will be compatible with our RPC users as error code will be changed (extended) now.
> 
> Have you ensured it? This looks like a potential breaking change especially if the callers assume one of given values. I'd delegate it to a separate field like `Reason: InvalidTxSignature: Signature is invalid` or similar so that the previously existing property preserves it semantics and given set of values.

I don’t have a way to ensure that.

On the other hand, there are error codes (as numbers) and messages meant to be descriptive. Geth provides decent descriptions under `message`:
[txpool/errors.go](https://github.com/ethereum/go-ethereum/blob/master/core/txpool/errors.go)
[core/error.go](https://github.com/ethereum/go-ethereum/blob/master/core/error.go)

Currently, we return a very generic `Invalid`, which isn’t particularly helpful. For RPC users, relying on error codes for general purposes makes more sense. If someone is still depending on messages, perhaps it’s time to reconsider that approach.

So IMO `message` is already our `Reason` field.

Scooletz:
> On the other hand, there are error codes (as numbers) and messages meant to be descriptive.

Three folks already accepted this PR. I don't believe it's worth to block it. BTW, this whole discussion reminds me of _Public APIs, like diamonds, are forever_ by Joshua Bloch 😄

benaadams (approved):

PR #8241, opened 2025-02-21
Return empty logs for future blocks

flcl42:
## Changes

- Return empty logs for future blocks, as others do

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No


src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
```diff
@@ -664,6 +664,33 @@ public ResultWrapper<IEnumerable<FilterLog>> eth_getLogs(Filter filter)
...
+        {
+            blockResult = blockFinder.SearchForHeader(blockParameter);
+
+            if (blockResult.IsError)
+            {
+                sourceBlockNumber = blockParameter.Type is BlockParameterType.BlockNumber ? blockParameter.BlockNumber.Value : 0;
+
+                if (blockParameter.Type is BlockParameterType.BlockNumber &&
+                    blockFinder.Head?.Number is not null &&
+                    blockFinder.Head.Number < blockParameter.BlockNumber)
```
alexb5dh, line 681: What if `blockFinder.Head.Number == blockParameter.BlockNumber`?

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
```diff
@@ -589,57 +589,57 @@ public ResultWrapper<IEnumerable<FilterLog>> eth_getFilterLogs(UInt256 filterId)
...
+        BlockParameter fromBlock = filter.FromBlock;
+        BlockParameter toBlock = filter.ToBlock;
+
         // because of lazy evaluation of enumerable, we need to do the validation here first
         CancellationTokenSource timeout = BuildTimeoutCancellationTokenSource();
         CancellationToken cancellationToken = timeout.Token;
 
+        if (!TryFindOrUseLatest(_blockFinder, ref toBlock, out SearchResult<BlockHeader> toBlockResult, out long sourceToBlockNumber))
+        {
+            timeout.Dispose();
```
alexb5dh, line 601: What do you think about adding a `finally` with `timeout.Dispose()` and remove all other occurrences?

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
```diff
@@ -589,57 +589,57 @@ public ResultWrapper<IEnumerable<FilterLog>> eth_getFilterLogs(UInt256 filterId)
...
-        cancellationToken.ThrowIfCancellationRequested();
-
-        BlockHeader fromBlock = fromBlockResult.Object!;
-        BlockHeader toBlock = toBlockResult.Object!;
-
-        long fromBlockNumber = fromBlock.Number;
-        long toBlockNumber = toBlock.Number;
-
-        if (fromBlockNumber > toBlockNumber && toBlockNumber != 0)
+        if (_blockFinder.Head?.Number is not null && sourceFromBlockNumber > _blockFinder.Head.Number)
```
alexb5dh, line 627: I see that  `&& toBlockNumber != 0` condition was removed, is this case handled properly?

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
flcl42: It's OK, logs will be returned for a single block

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
flcl42: Good idea

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
```diff
@@ -664,6 +659,33 @@ public ResultWrapper<IEnumerable<FilterLog>> eth_getLogs(Filter filter)
...
+        {
+            blockResult = blockFinder.SearchForHeader(blockParameter);
+
+            if (blockResult.IsError)
+            {
+                sourceBlockNumber = blockParameter.Type is BlockParameterType.BlockNumber ? blockParameter.BlockNumber.Value : null;
+
+                if (blockParameter.Type is BlockParameterType.BlockNumber &&
+                    blockFinder.Head?.Number is not null &&
+                    blockFinder.Head.Number < blockParameter.BlockNumber)
```
LukaszRozmej, line 676: `blockParameter.Type is BlockParameterType.BlockNumber` check is done twice.

I am not sure if I understand intention of that `if` condition. Can you make it clearer?

flcl42: It's not optimal in rare corner cases, but is surprisingly less complicated

LukaszRozmej: Can you change the code so the intentions are more readable?

alexb5dh (approved):

LukaszRozmej (approved):

PR #8247, opened 2025-02-23
Use rocksdb_get_pinned

benaadams:
Use PinnableSlice; for less native memcpy with point lookups https://rocksdb.org/blog/2017/08/24/pinnableslice.html

## Changes

- Switch to rocksdb_get_pinned for gets; temporary change
- Made PR upstream https://github.com/curiosity-ai/rocksdb-sharp/pull/61

- [x] Optimization

#### Requires testing

- [x] No


src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
```diff
@@ -643,6 +635,67 @@ public byte[]? this[ReadOnlySpan<byte> key]
...
+        IntPtr valuePtr = Native.Instance.rocksdb_pinnableslice_value(handle, out UIntPtr valueLength);
+        if (valuePtr == IntPtr.Zero)
+        {
+            return null;
+        }
+
+        int length = (int)valueLength;
+        byte[] result = new byte[length];
+        Marshal.Copy(valuePtr, result, 0, length);
+        Native.Instance.rocksdb_pinnableslice_destroy(handle);
```
LukaszRozmej, line 695: should be in finally clause

benaadams: Added to finally

asdacap (approved):

src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
```diff
@@ -643,6 +637,74 @@ public byte[]? this[ReadOnlySpan<byte> key]
...
+            }
+
+            int length = (int)valueLength;
+            byte[] result = new byte[length];
+            new ReadOnlySpan<byte>((void*)valuePtr, length).CopyTo(new Span<byte>(result));
+            return result;
+        }
+        finally
+        {
+            Native.Instance.rocksdb_pinnableslice_destroy(handle);
```
LukaszRozmej, line 697: shouldn't it be `valuePtr` and 

https://github.com/NethermindEth/nethermind/pull/8247/files#diff-37e7eb8b529265d0cd0f25ad1f5eed170ca123883a8dc9ba13d1cef5a2409549R684

should be outside of `try`?

Or maybe it should be both?

src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
```diff
@@ -643,6 +637,74 @@ public byte[]? this[ReadOnlySpan<byte> key]
...
+        {
+            IntPtr valuePtr = Native.Instance.rocksdb_pinnableslice_value(handle, out UIntPtr valueLength);
+            if (valuePtr == IntPtr.Zero)
+            {
+                return null;
+            }
+
+            int length = (int)valueLength;
+            byte[] result = new byte[length];
+            new ReadOnlySpan<byte>((void*)valuePtr, length).CopyTo(new Span<byte>(result));
```
LukaszRozmej, line 692: I thought getting through Spans was all about avoiding copying, that's why we have `GetSpan` and `DangerousReleaseMemory` in the first place?

LukaszRozmej (commented):
Is it any better? Can we measure it? What about `GetSpan`?

src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
benaadams: doesn't care about `valuePtr` its `handle` we are releasing

PR #8248, opened 2025-02-23
Skip SLOAD and SSTORE lookups on empty storage

benaadams:
## Changes

- If an account's storage is empty; then we can skip db reads for all slots for `SLOAD` and `SSTORE` (extending the self-destruct approach)
    - New contract creates, set code, etc

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
+++ b/src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
@@ -43,7 +43,7 @@ internal sealed class PersistentStorageProvider : PartialStorageProviderBase
     private readonly Dictionary<StorageCell, byte[]> _originalValues = new();
 
     private readonly HashSet<StorageCell> _committedThisRound = new();
-    private readonly Dictionary<AddressAsKey, SelfDestructDictionary<byte[]>> _blockCache = new(4_096);
+    private readonly Dictionary<AddressAsKey, DefaultableDictionary<byte[]>> _blockCache = new(4_096);
     private readonly ConcurrentDictionary<StorageCell, byte[]>? _preBlockCache;
     private readonly Func<StorageCell, byte[]> _loadFromTree;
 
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
```diff
@@ -501,23 +516,31 @@ public StorageTree Create(Address address, IScopedTrieStore trieStore, Hash256 s
...
     {
-        private bool _selfDestruct;
+        private bool _missingAreDefault;
         private readonly Dictionary<UInt256, TValue> _dictionary = new(Comparer.Instance);
 
+        public void MarkEmpty()
+        {
+            _missingAreDefault = true;
+        }
+
```
LukaszRozmej, line 528: We potentially don't need this method, we will have one additional clear on dictionary but it is a no-op, so no big deal. We also need to rename `SelfDestruct` to `MarkEmpty` or `MarkCleared`.

LukaszRozmej (approved):

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
```diff
@@ -501,23 +516,31 @@ public StorageTree Create(Address address, IScopedTrieStore trieStore, Hash256 s
             => new(trieStore, storageRoot, logManager);
     }
 
-    private sealed class SelfDestructDictionary<TValue>(TValue destructedValue)
+    private sealed class SelfDestructDictionary<TValue>(TValue defaultValue)
```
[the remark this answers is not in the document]
benaadams: Renamed

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
benaadams: `ClearAndSetMissingAsDefault`

PR #8251, opened 2025-02-24
Capture block stats per log (no block stats bleed)

### LINKED ISSUE #8163: Invalid block details after single block reorg
Image

After single block reorg new block seems to be adding block size and tx details from both original and new block.
On Image above - Seems like block have  34,81MGas while in reality it is 4,8 MGas.

benaadams:
Resolves #8163

## Changes

- Pass the block stats via params rather than reading them from fields which may have been overwritten by time of log
- Also add blob count when chain supports blobs

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Processing/ProcessingStats.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/ProcessingStats.cs
@@ -17,43 +17,43 @@ using Nethermind.State;
 namespace Nethermind.Consensus.Processing
 {
     //TODO Consult on disabling of such metrics from configuration
-    internal class ProcessingStats : IThreadPoolWorkItem
+    internal class ProcessingStats
     {
+        private readonly Action<BlockData> _executeFromThreadPool;
         private readonly IStateReader _stateReader;
         private readonly ILogger _logger;
         private readonly Stopwatch _runStopwatch = new();
 
-        private Block? _lastBlock;
-        private Hash256? _lastBranchRoot;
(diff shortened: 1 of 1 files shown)


flcl42 (approved):

src/Nethermind/Nethermind.Consensus/Processing/ProcessingStats.cs
```diff
@@ -319,5 +370,30 @@ public void Start()
             new Address("0x9FC3da866e7DF3a1c57adE1a97c9f00a70f010c8"), // Extra data as: Titan (titanbuilder.xyz)
             new Address("0x0b92619DdE55C0cbf828d32993a7fB004E00c84B"), // Extra data as: Builder+ www.btcs.com/builder
         };
+
+        private struct BlockData
```
LukaszRozmej, line 374: As it will always end up on heap due to multithreading, maybe better to make it a class and use ObjectPool?

Probably still miniscule impact compared to those strings though.

LukaszRozmej (approved):

src/Nethermind/Nethermind.Consensus/Processing/ProcessingStats.cs
benaadams: Changed to pooling; is big enough to bother me 😉

LukaszRozmej: Pool the strings!

PR #8255, opened 2025-02-24
Perf/parallel block downloader

asdacap:
- Include #8254 
- Currently forward sync only download from a single peer. 
- This is a problem as the upload of a peer is more limited than download therefore the download speed really depends on the peer that the forward sync is being pinned to.
  - Depending on the peer, the download speed of the forward sync can be 0 (hang), 15 blk/s or 150 blks. 
  - Since nethermind can consistently process at least 30blk/s a lot of the time when catching up, the node is waiting until it got lucky with a peer with fast upload.
  - This makes it hard to know if something is wrong with it or not.
  - Also makes performance benchmarking inconsistent as most of the time, it waiting for block download.
- This PR add the ability to dowload blocks and receipts in parallel.
- This provide a much more consistent forward sync download speed of around 200 block per sec.
- Works by having a map of hash->(block, receipts) that stores download requests. 

- [X] Optimization

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- [x] Mainnet snap sync
- [x] Mainnet snap sync with disable pivot updater
- [x] Energyweb
- [x] Sync github action
- [x] Hive test

### DIFF
--- a/src/Nethermind/Nethermind.Synchronization.Test/BlockDownloaderTests.cs
+++ b/src/Nethermind/Nethermind.Synchronization.Test/BlockDownloaderTests.cs
@@ -929,15 +929,27 @@ public partial class BlockDownloaderTests
         {
             IPeerAllocationStrategy peerAllocationStrategy = Substitute.For<IPeerAllocationStrategy>();
 
+            AutoResetEvent autoResetEvent = new(true);
+
             peerAllocationStrategy
                 .Allocate(Arg.Any<PeerInfo?>(), Arg.Any<IEnumerable<PeerInfo>>(), Arg.Any<INodeStatsManager>(), Arg.Any<IBlockTree>())
                 .Returns(peerInfo);
             SyncPeerAllocation peerAllocation = new(peerAllocationStrategy, AllocationContexts.Blocks, null);
+
+            // Set the current here
             peerAllocation.AllocateBestPeer(new List<PeerInfo>(), Substitute.For<INodeStatsManager>(), BlockTree);
 
             PeerPool
                 .Allocate(Arg.Any<IPeerAllocationStrategy>(), Arg.Any<AllocationContexts>(), Arg.Any<int>(), Arg.Any<CancellationToken>())
-                .Returns(Task.FromResult(peerAllocation));
+                .Returns((c) =>
+                {
+                    if (!autoResetEvent.WaitOne(100)) return Task.FromResult<SyncPeerAllocation?>(null)!;
+                    return Task.FromResult(peerAllocation);
+                });
+
+            PeerPool
+                .When((p) => p.Free(peerAllocation))
+                .Do((c) => autoResetEvent.Set());
         }
 
         public async Task SyncUntilNoRequest(SyncFeedComponent<BlocksRequest> component, PeerInfo peerInfo)
@@ -980,6 +992,7 @@ public partial class BlockDownloaderTests
                 e => BlockTree.NewBestSuggestedBlock -= e,
                 (arg) => arg.Block.Number == untilBestSuggestedHeaderIs);
 
+            if (BlockTree.BestSuggestedHeader?.Number == untilBestSuggestedHeaderIs) return;
             Task dLoop = FullSyncFeedComponent.Dispatcher.Start(cts.Token);
             FullSyncFeedComponent.Feed.Activate();
 


benaadams:
Needs rebasing

src/Nethermind/Nethermind.Blockchain.Test/BlockhashProviderTests.cs
```diff
@@ -286,4 +286,34 @@ private static void AssertGenesisHash(IReleaseSpec spec, BlockhashProvider provi
...
+        tree.SuggestHeader(current.Header);
+
+        IWorldState worldState = CreateWorldState();
+        var specProvider = new CustomSpecProvider(
+            (new ForkActivation(0, genesis.Timestamp), Frontier.Instance),
+            (new ForkActivation(0, current.Timestamp), Prague.Instance));
+        BlockhashStore store = new(specProvider, worldState);
+
+        // 1. Set some code to pass IsContract check
+        byte[] code = [1, 2, 3];
```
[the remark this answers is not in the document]
benaadams: Doesn't know new syntax

asdacap:
Request size adjustment is a bit flaking. Need to find a solution first or batch size is going to be limited by the slowest node.

src/Nethermind/Nethermind.Synchronization/Blocks/BlockDownloader.cs
```diff
@@ -80,91 +109,52 @@ private void BlockTreeOnNewHeadBlock(object? sender, BlockEventArgs e)
...
+                    // Note: Could be change in peer or a fork.
+                    throw new InvalidOperationException("Forward header starting block number did not changed. This is unexpected");
                 }
+                previousStartingHeaderNumber = headers[0].Number;
 
-                AdjustSyncBatchSize(Stopwatch.GetElapsedTime(startTime));
-
-                Block[]? blocks = context.Blocks;
-                TxReceipt[]?[]? receipts = context.ReceiptsForBlocks;
+                (bool shouldProcess, bool downloadReceipts) = ReceiptEdgeCase(bestProcessedBlock, headers[1].Number, originalShouldProcess, originalDownloadReceiptOpts);
```
LukaszRozmej, line 134: Question about the edge case - if we processed some block, we shouldn't need to download receipt for it. If we store receipts that receipt should be already stored.

src/Nethermind/Nethermind.Synchronization/Blocks/BlockDownloader.cs
```diff
@@ -175,22 +165,287 @@ private async Task<long> DownloadBlocks(PeerInfo? bestPeer, BlocksRequest blocks
...
+            }
+
+            return new BlocksRequest()
+            {
+                BodiesRequests = bodiesToDownload,
+                ReceiptsRequests = receiptsToDownload,
+            };
+        }
+
+        private ArrayPoolList<BlockEntry> AssembleSatisfiedEntries(IOwnedReadOnlyList<BlockHeader?>? headers, bool shouldDownloadReceipt)
```
LukaszRozmej, line 280:
```suggestion
        private ArrayPoolList<BlockEntry> AssembleSatisfiedEntries(IOwnedReadOnlyList<BlockHeader?> headers, bool shouldDownloadReceipt)
```

src/Nethermind/Nethermind.Synchronization/Blocks/BlockDownloader.cs
```diff
@@ -175,22 +165,287 @@ private async Task<long> DownloadBlocks(PeerInfo? bestPeer, BlocksRequest blocks
...
+        {
+            ArrayPoolList<BlockEntry>? satisfiedEntry = null;
+            try
+            {
+                satisfiedEntry = new ArrayPoolList<BlockEntry>(headers.Count);
+                foreach (var blockHeader in headers.Skip(1))
+                {
+                    if (!_downloadRequests.TryGetValue(blockHeader.Hash, out BlockEntry blockEntry)) break;
```
LukaszRozmej, line 288:
```suggestion
                    if (blockHeader is null) break;
                    if (!_downloadRequests.TryGetValue(blockHeader.Hash, out BlockEntry blockEntry)) break;
```

src/Nethermind/Nethermind.Synchronization/Blocks/BlockDownloader.cs
```diff
@@ -175,22 +165,287 @@ private async Task<long> DownloadBlocks(PeerInfo? bestPeer, BlocksRequest blocks
...
+
+        public void PruneDownloadBuffer()
+        {
+            _downloadRequests.Clear();
+        }
+
+        private void PruneRequestMap(IOwnedReadOnlyList<BlockHeader> currentHeaders)
+        {
+            HashSet<Hash256> currentHeaderHashes = currentHeaders.Select(h => h.Hash).ToHashSet();
+            HashSet<Hash256> toRemove = new HashSet<Hash256>();
```
LukaszRozmej, line 197: You don't need a set here.
```suggestion
            using ArrayPoolList<Hash256> toRemove = new ArrayPoolList<Hash256>();
```

src/Nethermind/Nethermind.Synchronization/Blocks/BlockDownloader.cs
```diff
@@ -175,22 +165,287 @@ private async Task<long> DownloadBlocks(PeerInfo? bestPeer, BlocksRequest blocks
...
+            foreach (var kv in _downloadRequests)
+            {
+                if (!currentHeaderHashes.Contains(kv.Key))
+                {
+                    toRemove.Add(kv.Key);
```
LukaszRozmej, line 202: To be honest, you don't need a `toRemove` collection at all. `ConcurrentDictionary` supports iterating and removing in performant way. It iterates on a snapshot-like (but light) enumerator.

```suggestion
                    _downloadRequests.Remove(kv.Key, out _);
```

src/Nethermind/Nethermind.Synchronization/Blocks/BlocksSyncPeerSelectionStrategy.cs
```diff
@@ -68,14 +68,8 @@ public BlocksSyncPeerAllocationStrategy(long? minBlocksAhead)
...
-                // TODO: this is not needed -> always wtapped
-                // if (info.TotalDifficulty <= localTotalDiff)
-                // {
-                //     // if we require higher difficulty then we need to discard peers with same diff as ours
-                //     continue;
-                // }
-
-                if (info.TotalDifficulty - localTotalDiff <= 2 && (info.PeerClientType == NodeClientType.Parity || info.PeerClientType == NodeClientType.OpenEthereum))
+                UInt256 remoteTd = info.TotalDifficulty;
+                if (remoteTd >= localTotalDiff && remoteTd - localTotalDiff <= 2 && (info.PeerClientType == NodeClientType.Parity || info.PeerClientType == NodeClientType.OpenEthereum))
```
LukaszRozmej, line 72: Do consistent naming `Td` vs `TotalDiff`

src/Nethermind/Nethermind.Synchronization/Blocks/BlocksSyncPeerSelectionStrategy.cs
```diff
@@ -109,7 +103,9 @@ public BlocksSyncPeerAllocationStrategy(long? minBlocksAhead)
             }
 
             averageSpeed /= peersCount;
-            UInt256 difficultyDifference = bestDiffPeer.Info.TotalDifficulty - localTotalDiff;
+            UInt256 difficultyDifference = bestDiffPeer.Info.TotalDifficulty > localTotalDiff
+                ? bestDiffPeer.Info.TotalDifficulty - localTotalDiff
+                : 0;
```
LukaszRozmej, line 108:
```suggestion
                : Uint256.Zero;
```

src/Nethermind/Nethermind.Synchronization/Blocks/PowForwardHeaderProvider.cs
```diff
@@ -165,7 +165,7 @@ private void OnNewBestPeer(PeerInfo newBestPeer)
                 }
 
                 // Remember, we start downloading from currentNumber+1
-                if (!CheckAncestorJump(bestPeer, headers[1], ref _currentNumber)) continue;
+                if (!CheckAncestorJump(bestPeer, headers[0], ref _currentNumber)) continue;
```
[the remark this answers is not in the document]
asdacap: Nope, it still not obsolete, just change how it work.

src/Nethermind/Nethermind.Synchronization/Blocks/FullSyncFeed.cs
```diff
@@ -8,26 +8,32 @@
 
 namespace Nethermind.Synchronization.Blocks
 {
-    public class FullSyncFeed : ActivatedSyncFeed<BlocksRequest?>
+    public class FullSyncFeed(IForwardSyncController forwardSyncController) : ActivatedSyncFeed<BlocksRequest?>
```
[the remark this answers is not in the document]
asdacap: I'm more likely to merge the two sync mode completely.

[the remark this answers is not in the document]
asdacap: But in generally, not gonna change that side of the code in this  PR.

src/Nethermind/Nethermind.Synchronization/Blocks/BlockDownloader.cs
```diff
@@ -39,7 +45,28 @@ public class BlockDownloader : ISyncDownloader<BlocksRequest>
         private readonly IFullStateFinder _fullStateFinder;
         private readonly IForwardHeaderProvider _forwardHeaderProvider;
         private readonly ILogger _logger;
-        protected SyncBatchSize _syncBatchSize;
+
+        // Estimated maximum tx in buffer used to estimate memory limit. Each tx is on average about 1KB.
+        private readonly int _maxTxInBuffer = 200000;
```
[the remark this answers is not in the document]
asdacap: Correct, but its unclean to pass in the the size  somehow right now. I'd prefer to have #8528 first.

src/Nethermind/Nethermind.Blockchain/Synchronization/ISyncConfig.cs
```diff
@@ -170,4 +170,7 @@ public interface ISyncConfig : IConfig
 
     [ConfigItem(Description = "_Technical._ Enable storage range split.", DefaultValue = "false", HiddenFromDocs = true)]
     bool EnableSnapSyncStorageRangeSplit { get; set; }
+
+    [ConfigItem(Description = "_Technical._ Max tx in forward sync buffer.", DefaultValue = "200000", HiddenFromDocs = true)]
```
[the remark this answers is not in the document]
asdacap: Removed the one in `BlockDownloader`.

src/Nethermind/Nethermind.Synchronization/Blocks/BlockDownloader.cs
asdacap: If `shouldProcess` is originally true, then `shouldDownloadReceipt` is originally false (FullSync). Otherwise in FastSync, it download receipt depending on sync config. This is to detect edge case where FullSync is on but the state is not at just the right block, resulting in missing receipt. Yes, this does mean that all along the mode could just be `FullSync` all the time and it would probably be just fine.

LukaszRozmej (approved):

PR #8264, opened 2025-02-24
Request Hash calculation no allocations

LukaszRozmej:
## Changes

- Use RecyclableMemoryStream instead of ArrayPoolList<byte> to avoid ToArray when computing hash

- [x] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Core/ExecutionRequest/ExecutionRequestExtensions.cs
+++ b/src/Nethermind/Nethermind.Core/ExecutionRequest/ExecutionRequestExtensions.cs
@@ -45,7 +45,7 @@ public static class ExecutionRequestExtensions
         }
 
         using SHA256 sha256 = SHA256.Create();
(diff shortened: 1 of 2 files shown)


MarekM25 (commented):
insane :D

LukaszRozmej:
> insane :D

Nah, normal. You forgot to approve BTW.

rjnrohit (approved):

benaadams (approved):

tanishqjasoria (approved):

LukaszRozmej:
> Don't we have other cases where we can use this as well?

Not sure, depends on the API, if the API could take `Span` or `IReadOnlyList` then we could keep `ArrayPoolList`, but as it can only take arrays or `Streams` this was the problem.

PR #8269, opened 2025-02-25
hash casting cleanup

LukaszRozmej:
## Changes

- Fixes `Hash`/`ValueHash` distinction by removing implicit cast from `ValueHash` to Hash and adjusting the code
- Uses `ValueHash` where possible, defaulting to Hash where it may live longer on the heap, trying to minimize conversions that were previously implicit
- Introduces `Hash256AsKeyComparer` to be able to do `HashSet`/`Dictionary` lookups by `ValueHash`
- Moves block processing to `IReadOnlyList` and accumulates branch in `ArrayPoolList` to avoid allocation
- Adds `Reverse` to `ArrayPoolList` not to avoid LINQ reverse that might allocate
- Makes `ProcessingBranch` a `ref struct`, changing `List`s to `ArrayPoolList`s and making it `Dispose`able

- [x] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


src/Nethermind/Nethermind.Trie/INodeStorage.cs
```diff
@@ -18,14 +18,14 @@ public interface INodeStorage
     /// </summary>
     public bool RequirePath { get; }
 
-    byte[]? Get(Hash256? address, in TreePath path, in ValueHash256 keccak, ReadFlags readFlags = ReadFlags.None);
-    void Set(Hash256? address, in TreePath path, in ValueHash256 hash, ReadOnlySpan<byte> data, WriteFlags writeFlags = WriteFlags.None);
+    byte[]? Get(in ValueHash256? address, in TreePath path, in ValueHash256 keccak, ReadFlags readFlags = ReadFlags.None);
```
asdacap, line 21: Address should be kept a reference type as the triestore should have it as `Hash256` so that it can be deduplicated across keys of the same storage.

src/Nethermind/Nethermind.Trie/Pruning/ITrieStoreExtensions.cs
```diff
@@ -1,12 +1,17 @@
...
+
 namespace Nethermind.Trie.Pruning
 {
     // ReSharper disable once InconsistentNaming
     public static class ITrieStoreExtensions
     {
         public static IReadOnlyTrieStore AsReadOnly(this ITrieStore trieStore, INodeStorage? readOnlyStore = null) =>
             trieStore.AsReadOnly(readOnlyStore);
+
+        public static IScopedTrieStore GetTrieStore(this ITrieStore trieStore, ValueHash256 address) =>
```
asdacap, line 14: Should be removed so that less mistake on top level where the address is stored as `ValueHash256` which may result in increase memory usage in pruning cache.

LukaszRozmej: So besides tests it has 3 usages and all of them allocated before `Hash256` from `Address.ToAccountPath`. I think we can't easily avoid allocating here. I can remove but the result will always be the same + like 20 explicit allocations (or 3 if I make this method available in tests).

src/Nethermind/Nethermind.Trie/INodeStorage.cs
LukaszRozmej: You are right those were unnecessary

src/Nethermind/Nethermind.Trie/Pruning/ITrieStoreExtensions.cs
LukaszRozmej: Changed it to take `Address`

src/Nethermind/Nethermind.State/Healing/HealingStateTree.cs
```diff
@@ -67,11 +67,11 @@ public override void Set(ReadOnlySpan<byte> rawKey, byte[] value)
         }
     }
 
-    private bool Recover(in TreePath missingNodePath, in ValueHash256 hash, Hash256 fullPath)
+    private bool Recover(in TreePath missingNodePath, in Hash256 hash, Hash256 fullPath)
```
benaadams, line 70: No `in` with ref type; or is a pointer of a pointer for double pointer chase
```suggestion
    private bool Recover(in TreePath missingNodePath, Hash256 hash, Hash256 fullPath)
```

src/Nethermind/Nethermind.State/Healing/HealingStorageTree.cs
```diff
@@ -65,11 +65,11 @@ public override void Set(ReadOnlySpan<byte> rawKey, byte[] value)
         }
     }
 
-    private bool Recover(in TreePath missingNodePath, in ValueHash256 hash, Hash256 fullPath)
+    private bool Recover(in TreePath missingNodePath, in Hash256 hash, Hash256 fullPath)
```
benaadams, line 68: No `in` with ref type; or is a pointer of a pointer for double pointer chase
```suggestion
    private bool Recover(in TreePath missingNodePath, Hash256 hash, Hash256 fullPath)
```

asdacap (approved):

PR #8252, opened 2025-02-24
secp256r1 improvements

alexb5dh:
**[WIP] Requires few other PRs to be merged**

Improvements to secp256r1 precompile.

## Changes

- Switches to Go ECDSA using P/Invoke to improve speed
- Adds precompile benchmarks

#### What types of changes does your code introduce?

- [x] Optimization

#### Requires testing

- [x] Yes

#### If yes, did you write tests?

- [x] Yes

PR #8284, opened 2025-02-28
Secp256r1 precompile optimization & tests

alexb5dh:
Reincarnation of #8252 (secp256r1 precompile improvements)

## Changes

- Switches to [Go ECDSA using P/Invoke](https://github.com/NethermindEth/secp256r1-bindings) for improved performance
- Adds [test vectors from op-geth](https://github.com/ethereum-optimism/op-geth/blob/7017b54770d480b5c8be63dc40eac9da166150f5/core/vm/testdata/precompiles/p256Verify.json)

- [x] Optimization

#### Requires testing

- [x] Yes

- [x] Yes


LukaszRozmej (approved):

benaadams (approved):

src/Nethermind/Nethermind.Evm/Precompiles/Secp256r1Precompile.cs
```diff
@@ -19,24 +19,12 @@ public class Secp256r1Precompile : IPrecompile<Secp256r1Precompile>
...
-        ReadOnlySpan<byte> hash = bytes[..32], sig = bytes[32..96];
-        ReadOnlySpan<byte> x = bytes[96..128], y = bytes[128..160];
-
-        using var ecdsa = ECDsa.Create(new ECParameters
-        {
-            Curve = ECCurve.NamedCurves.nistP256,
-            Q = new() { X = x.ToArray(), Y = y.ToArray() }
-        });
-        var isValid = ecdsa.VerifyHash(hash, sig);
+        var isValid = Secp256r1.VerifySignature(in inputData);
```
Marchhill, line 27: Minor point - I don't think `in` is needed here as `ReadOnlyMemory` is designed to be passed by value: https://stackoverflow.com/questions/49297751/should-readonlyspant-parameters-use-the-in-modifier

Marchhill (approved):

src/Nethermind/Nethermind.Evm/Precompiles/Secp256r1Precompile.cs
alexb5dh: Not sure about "designed", but there's an interesting [benchmark](https://stackoverflow.com/a/49299718/3542151), showing there doesn't seem to be any performance benefit of using `in`.

Yet I'll prefer to leave this as is for now, as there a lot of places where we use still use `in` with `ReadOnlyMemory`.

PR #8285, opened 2025-03-01
Disable warmup by default in StateTestRunner

### LINKED ISSUE #8227: Disable WarmUp in stateTests by default
This [StateTestRunner.cs#L76](https://github.com/NethermindEth/nethermind/blob/master/src/Nethermind/Nethermind.Test.Runner/StateTestRunner.cs#L76) is relevant only when we want to benchmark, but warmup should be disabled by default, with a command-line parameter available to turn it on:

[Program.cs#L17](https://github.com/NethermindEth/nethermind/blob/master/src/Nethermind/Nethermind.Test.Runner/Program.cs#L17)

Antrikshgwal:
Closes #8227

## Changes

- Add --warmup command-line option to enable warmup when needed for benchmarking
- Modify StateTestsRunner to only perform warmup when explicitly enabled
- Default behavior now skips warmup for faster test execution
- Maintain warmup capability for benchmarking scenarios

## Types of changes

- [ ] Optimization

## Testing

- [ ] Yes

### DIFF
--- a/src/Nethermind/Nethermind.Test.Runner/StateTestRunner.cs
+++ b/src/Nethermind/Nethermind.Test.Runner/StateTestRunner.cs
@@ -30,6 +30,7 @@ namespace Nethermind.Test.Runner
         private readonly bool _traceStack;
         private readonly string? _filter;
         private readonly ulong _chainId;
+        private readonly bool _enableWarmup;
         private static readonly IJsonSerializer _serializer = new EthereumJsonSerializer();
 
         public StateTestsRunner(ITestSourceLoader testsSource, WhenTrace whenTrace, bool traceMemory, bool traceStack, ulong chainId, string? filter = null)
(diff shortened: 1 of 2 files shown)


LukaszRozmej (approved):
Please fix indentation.

Antrikshgwal:
> Please fix indentation.

done

src/Nethermind/Nethermind.Test.Runner/StateTestRunner.cs
```diff
@@ -40,6 +41,7 @@ public StateTestsRunner(ITestSourceLoader testsSource, WhenTrace whenTrace, bool
             _traceStack = traceStack;
             _filter = filter;
             _chainId = chainId;
+            _enableWarmup = enableWarmup;
```
LukaszRozmej, line 44: Where `enableWarmup` comes from? It is not a constructor parameter.

Antrikshgwal: Hey, I, somehow, missed to intialize enableWarmup in the constructor. Sorry about that. I have initialized it now.

PR #8290, opened 2025-03-03
fix txValidator for State test, fix PragueGnosis spec

yerke26:
Before GeneralTestBase was using TxValidator for Mainnet, which is incorrect for Gnosis
## Changes

- fix GeneralTestBase: choose appropriate chainId for TxValidator
- Add FeeCollector to PragueGnosis

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Remarks

### DIFF
--- a/src/Nethermind/Ethereum.Test.Base/GeneralTestBase.cs
+++ b/src/Nethermind/Ethereum.Test.Base/GeneralTestBase.cs
@@ -4,14 +4,11 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Linq;
-using Nethermind.Consensus.Validators;
 using Nethermind.Core;
 using Nethermind.Core.Crypto;
 using Nethermind.Core.Extensions;
 using Nethermind.Core.Specs;
 using Nethermind.Crypto;
-using Nethermind.Db;
 using Nethermind.Int256;
 using Nethermind.Evm;
 using Nethermind.Evm.Tracing;
@@ -21,15 +18,11 @@ using Nethermind.Specs;
 using Nethermind.Specs.Forks;
 using Nethermind.Specs.Test;
 using Nethermind.State;
-using Nethermind.Trie.Pruning;
 using NUnit.Framework;
 using System.Threading.Tasks;
 using Autofac;
 using Nethermind.Config;
-using Nethermind.Consensus.AuRa.InitializationSteps;
-using Nethermind.Core.Test.Builders;
 using Nethermind.Core.Test.Modules;
-using Nethermind.Specs.GnosisForks;
 
 namespace Ethereum.Test.Base
 {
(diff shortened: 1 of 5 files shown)


src/Nethermind/Nethermind.Specs/GnosisForks/18_PragueGnosis.cs
```diff
@@ -16,6 +16,7 @@ protected PragueGnosis() : base()
         BlobBaseFeeUpdateFraction = 0x10fafa;
         TargetBlobCount = 1;
         MaxBlobCount = 2;
+        FeeCollector = GnosisSpecProvider.FeeCollector;
```
LukaszRozmej, line 19: should this be also done for previous forks?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Specs/GnosisForks/18_PragueGnosis.cs
Marchhill: I don't see why we need this, it should already be set in chainspec? It is not changing in Prague

yerke26: yes its already set in chainspec, but this spec classes are used for testing mostly, in state/blockchain tests you specify the fork and it uses those classes where FeeCollector is not specified

src/Nethermind/Ethereum.Test.Base/GeneralTestBase.cs
```diff
@@ -134,7 +126,7 @@ protected EthereumTestResult RunTest(GeneralStateTest test, ITxTracer txTracer)
                 header.ExcessBlobGas = BlobGasCalculator.CalculateExcessBlobGas(parent, spec);
             }
 
-            ValidationResult txIsValid = _txValidator.IsWellFormed(test.Transaction, spec);
+            ValidationResult txIsValid = ChainUtils.ValidateTransaction(test.Transaction, test.Fork);
```
Marchhill, line 129: I'd prefer not to hardcode this function in `ChainUtils` as it only works for Gnosis and mainnet. Also why use `test.Fork` instead of `spec`
```suggestion
            ValidationResult txIsValid = new TxValidator(test.ChainId).ValidateTransaction(test.Transaction, spec);
```

src/Nethermind/Nethermind.Specs/GnosisForks/18_PragueGnosis.cs
yerke26: > should this be also done for previous forks?

@LukaszRozmej  It should be on PragueCancun I guess?

MarekM25 (approved):

LukaszRozmej (approved):

PR #8292, opened 2025-03-04
Perf/concurrent processing jsonrpc socket

### LINKED ISSUE #8257: JSON RPC over IPC should support multiple simultaneous clients
The way the  `JsonRpcIpcRunner` is implemented now doesn't allow multiple simultaneous clients to be connected. This is because in _accept connection loop_ we do the following

```CS
while (true)
{
    if (_logger.IsInfo) _logger.Info("Waiting for an IPC connection...");

    Socket socket = await _server.AcceptAsync(cancellationToken);

    socket.ReceiveTimeout = _jsonRpcConfig.Timeout;
    socket.SendTimeout = _jsonRpcConfig.Timeout;

    using JsonRpcSocketsClient<IpcSocketMessageStream>? socketsClient = new(
        string.Empty,
        new IpcSocketMessageStream(socket),
        RpcEndpoint.IPC,
        _jsonRpcProcessor,
        _jsonRpcLocalStats,
        _jsonSerializer,
        maxBatchResponseBodySize: _jsonRpcConfig.MaxBatchResponseBodySize);

    await socketsClient.ReceiveLoopAsync();
}

```
The issue is the line `await socketsClient.ReceiveLoopAsync();` because as long as we are serving one client, we cannot accept others.

To be able to use JSON RPC over IPC/UnixDomainSockets with multiple clients simultaneously.

**Additional context**
Additionally the way this whole section is written, more specifically thanks to [this call](https://github.com/NethermindEth/nethermind/blob/b8034f1e93863ac2f7532fc5be4ad6e826f6ea67/src/Nethermind/Nethermind.Runner/JsonRpc/JsonRpcIpcRunner.cs#L110) to `Dispose` if the client Disconnects in erroneous way  the whole IPC is _shut down_ and no more clients are able to connect unless the node is restarted.

### LINKED ISSUE #8258: IPC should leverage the fact that UnixDomainSockets are fully duplex
**IMPORTANT**:
I could be misunderstanding the code; maybe we already _do have_ support for the improvement I'm proposing here!

For JSON RPC over IPC, we are using Unix Domain Sockets (UDS) ([link](https://github.com/NethermindEth/nethermind/blob/b8034f1e93863ac2f7532fc5be4ad6e826f6ea67/src/Nethermind/Nethermind.Runner/JsonRpc/JsonRpcIpcRunner.cs#L67)). These UnixDomainSockets are[ available since ](https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/)Windows 10/Windows Server 2019 (and ofc. much longer on Unix/Linux). They offer fur duplex communication support, meaning we can read and write simultaneously using a single socket connection. 

My C# is a bit Rusty 🦀, but unless I misunderstand the code, we have the following:
1. Reader loop starts: [link](https://github.com/NethermindEth/nethermind/blob/b8034f1e93863ac2f7532fc5be4ad6e826f6ea67/src/Nethermind/Nethermind.Runner/JsonRpc/JsonRpcIpcRunner.cs#L89)
2. If there is enough bytes, _process them_ (and `await`): [link](https://github.com/NethermindEth/nethermind/blob/b8034f1e93863ac2f7532fc5be4ad6e826f6ea67/src/Nethermind/Nethermind.Sockets/SocketClient.cs#L68)
3. Here is the code for process&write loops: [link](https://github.com/NethermindEth/nethermind/blob/e2043a92273543692c0f619ef46a76d123abfafa/src/Nethermind/Nethermind.JsonRpc/WebSockets/JsonRpcSocketsClient.cs#L57)

As far as I can see, while all of these calls are `async` there is no concurrency/parallelism _here_, meaning this is how I read the execution:
1.  Read message 1
2. Process message 1
3. Write a response to message 1
4. Read message 2
5. Process message 2
6. ...

As far as I can tell, all new reads are blocked till the previous write(s) are finished. IMO, this is suboptimal; the UDS (and, by extension, WebSockets) offers a fully duplex connection, meaning we can read and write simultaneously. How this is implemented, one s_low-to-process(or bad)-request_ can strain (block) the whole system. 

We should take advantage of UDS's full-duplex connection and handle reading and processing/writing separately. 

**Additional context**
[Here is how I implemented](https://github.com/nethermindeth/reipc) this for `rbuilder` - in a standalone Rust crate(library), it has (hopefully) an easy-to-understand graphical representation of the system I'm proposing.

The system depicted in the issue is, I _think,_ a bit more complicated than what we actually need in NMC. This is because `reipc` acts as a client and needs to attribute incoming data as responses to corresponding requests correctly. 

Whereas NMC acts as a server, it should:
1. Have reader loop
2. Once the message is received & parsed, it can be offloaded to  _the processor_ (which shouldn't block the reader loop)
3. Once _the processor_ is finished, it writes the response 

Some of the open Qs (if we go with the architecture proposed above):

- What will be the boundaries? Will the reader just read bytes and leave the parsing to _the processor_? Will the processor also write a response, or shall we have a separate writer loop (as we do with the reader)
- As far as I see, _the processor_ will be the bottleneck. I think we should have multiple _processors_ (e.g., at least 2) so that one bad request doesn't hog the whole IPC
- I propose using channels as impl. solution as they should be well suited for this task. Moreover using channels should simplify some of the code we already have (e.g. [reading this ](https://devblogs.microsoft.com/dotnet/an-introduction-to-system-threading-channels/) they already use the same primitives for the _data protection_ as [we do here](https://github.com/NethermindEth/nethermind/blob/e2043a92273543692c0f619ef46a76d123abfafa/src/Nethermind/Nethermind.JsonRpc/WebSockets/JsonRpcSocketsClient.cs#L26))

asdacap:
Resolves #8257
Resolves #8258

- Add support for processing json rpc request from websockets and IPC connection concurrently.
- Two additional config both set to 1 by default.
  - `--JsonRpc.WebSocketsProcessingConcurrency 32`
  - `--JsonRpc.IpcProcessingConcurrency 32`
- Fix IPC connection might hang on last request when it is in the same buffer as its previous request.

- [X] Optimization

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Added unit tests.
- Websocket and IPC connection works from rbuilder but I don't have any perf number.


LukaszRozmej:
Does it address #8257 and #8258 ?

asdacap:
Yes

src/Nethermind/Nethermind.JsonRpc/WebSockets/JsonRpcSocketsClient.cs
```diff
@@ -54,17 +68,62 @@ public override void Dispose()
...
+
+        await _processChannel.Writer.WriteAsync(new ProcessRequest(memory, memoryOwner), cancellationToken);
+    }
+
+    public override async Task ReceiveLoopAsync(CancellationToken cancellationToken)
+    {
+        using AutoCancelTokenSource cts = cancellationToken.CreateChildTokenSource();
+
+        using ArrayPoolList<Task> allTasks = new((int)(_processConcurrency + 1));
+        allTasks.Add(Task.Run(async () =>
```
mralj, line 87: WDYT: Does it make sense to make this threaded/_long running task_, e.g.:
`Task.Factory.StartNew(f, cancellationToken, TaskCreationOptions.LongRunning);`

src/Nethermind/Nethermind.JsonRpc/WebSockets/JsonRpcSocketsClient.cs
```diff
@@ -54,17 +68,62 @@ public override void Dispose()
...
+            finally
+            {
+                _processChannel.Writer.Complete();
+            }
+        }));
+
+        for (int i = 0; i < _processConcurrency; i++)
+        {
+            int idx = i;
+            allTasks.Add(WorkerLoop(cts.Token, idx));
```
mralj, line 102: Same as the above: 

> WDYT: Does it make sense to make this threaded/_long running task_, e.g.:
> `Task.Factory.StartNew(f, cancellationToken, TaskCreationOptions.LongRunning);`

LukaszRozmej:
@mralj what are the numbers for Geth or Reth?

mralj:
> @mralj what are the performance numbers for Geth or Reth?

I haven't checked these, since I would have to implement couple of `rbuilder` specific RPC calls. But now that you ask, I will try to find some time and do it. But, unless you think otherwise, I plan to treat this comparison as low-priority task, do you agree?

src/Nethermind/Nethermind.JsonRpc/WebSockets/JsonRpcSocketsClient.cs
```diff
@@ -33,13 +44,16 @@ public JsonRpcSocketsClient(
...
-        long? maxBatchResponseBodySize = null)
+        long? maxBatchResponseBodySize = null,
+        long concurrency = 1)
         : base(clientName, stream, jsonSerializer)
     {
         _jsonRpcProcessor = jsonRpcProcessor;
         _jsonRpcLocalStats = jsonRpcLocalStats;
         _maxBatchResponseBodySize = maxBatchResponseBodySize;
         _jsonRpcContext = new JsonRpcContext(endpointType, this, url);
+        _processChannel = Channel.CreateBounded<ProcessRequest>(1);
```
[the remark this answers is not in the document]
asdacap: Its deliberate, although setting it to concurrency is probably fine.

src/Nethermind/Nethermind.JsonRpc/WebSockets/JsonRpcSocketsClient.cs
asdacap: Not sure. Probably.

src/Nethermind/Nethermind.JsonRpc/IJsonRpcConfig.cs
```diff
@@ -169,4 +169,10 @@ public interface IJsonRpcConfig : IConfig
 
     [ConfigItem(Description = "The JSON-RPC server CORS origins.", DefaultValue = "*")]
     string[] CorsOrigins { get; set; }
+
+    [ConfigItem(Description = "Concurrency level of websocket connection.", DefaultValue = "1")]
+    int WebSocketsProcessingConcurrency { get; set; }
+
+    [ConfigItem(Description = "Concurrency level of IPC connection.", DefaultValue = "1")]
+    int IpcProcessingConcurrency { get; set; }
```
LukaszRozmej, line 177: Why defaults are 1? I think we should increase it?

src/Nethermind/Nethermind.JsonRpc/WebSockets/JsonRpcSocketsClient.cs
```diff
@@ -33,13 +44,16 @@ public JsonRpcSocketsClient(
...
-        long? maxBatchResponseBodySize = null)
+        long? maxBatchResponseBodySize = null,
+        int concurrency = 1)
         : base(clientName, stream, jsonSerializer)
     {
         _jsonRpcProcessor = jsonRpcProcessor;
         _jsonRpcLocalStats = jsonRpcLocalStats;
         _maxBatchResponseBodySize = maxBatchResponseBodySize;
         _jsonRpcContext = new JsonRpcContext(endpointType, this, url);
+        _processChannel = Channel.CreateBounded<ProcessRequest>(concurrency);
```
LukaszRozmej, line 55: Is it single writer, multiple reader? Let's provide that context for optimization.

src/Nethermind/Nethermind.JsonRpc/WebSockets/JsonRpcSocketsClient.cs
```diff
@@ -54,17 +68,61 @@ public override void Dispose()
...
+            await HandleRequest(request.Buffer, cancellationToken);
+        }
+    }
+
+    public async Task HandleRequest(Memory<byte> data, CancellationToken cancellationToken = default)
+    {
+        PipeReader request = PipeReader.Create(new ReadOnlySequence<byte>(data));
         int allResponsesSize = 0;
 
         await foreach (JsonRpcResult result in _jsonRpcProcessor.ProcessAsync(request, _jsonRpcContext))
```
LukaszRozmej, line 121: Should we cancel it here?
```suggestion
        await foreach (JsonRpcResult result in _jsonRpcProcessor.ProcessAsync(request, _jsonRpcContext).WithCancellation(cancellationToken))
```

src/Nethermind/Nethermind.JsonRpc/WebSockets/JsonRpcSocketsClient.cs
```diff
@@ -111,7 +169,7 @@ private void IncrementBytesSentMetric(int size)
         }
     }
 
-    public virtual async Task<int> SendJsonRpcResult(JsonRpcResult result)
+    public virtual async Task<int> SendJsonRpcResult(JsonRpcResult result, CancellationToken cancellationToken = default)
     {
         await _sendSemaphore.WaitAsync();
```
LukaszRozmej, line 174:
```suggestion
        await _sendSemaphore.WaitAsync(cancellationToken);
```

src/Nethermind/Nethermind.JsonRpc/WebSockets/JsonRpcSocketsClient.cs
```diff
@@ -121,15 +179,15 @@ public virtual async Task<int> SendJsonRpcResult(JsonRpcResult result)
...
+                await using JsonRpcBatchResultAsyncEnumerator enumerator = result.BatchedResponses!.GetAsyncEnumerator(cancellationToken);
                 while (await enumerator.MoveNextAsync())
                 {
                     JsonRpcResult.Entry entry = enumerator.Current;
                     using (entry)
                     {
                         if (!isFirst)
                         {
-                            await _stream.WriteAsync(_jsonComma);
+                            await _stream.WriteAsync(_jsonComma, cancellationToken);
```
LukaszRozmej, line 190: Missed `cancellationToken` in line 205

LukaszRozmej: IMO `SerializeAsync` should now also take `cancellationToken` (need to add to our json serializer)

src/Nethermind/Nethermind.JsonRpc/WebSockets/JsonRpcSocketsClient.cs
```diff
@@ -121,15 +179,15 @@ public virtual async Task<int> SendJsonRpcResult(JsonRpcResult result)
                 int responseSize = 1;
                 bool isFirst = true;
                 await _stream.WriteAsync(_jsonOpeningBracket);
```
LukaszRozmej, line 181:
```suggestion
                await _stream.WriteAsync(_jsonOpeningBracket, cancellationToken);
```

src/Nethermind/Nethermind.JsonRpc/WebSockets/JsonRpcSocketsClient.cs
```diff
@@ -54,17 +68,61 @@ public override void Dispose()
...
+        Memory<byte> memory = memoryOwner.Memory[..data.Count];
+
+        await _processChannel.Writer.WriteAsync(new ProcessRequest(memory, memoryOwner), cancellationToken);
+    }
+
+    public override async Task ReceiveLoopAsync(CancellationToken cancellationToken)
+    {
+        using AutoCancelTokenSource cts = cancellationToken.CreateChildTokenSource();
+
+        using ArrayPoolList<Task> allTasks = new(_processConcurrency + 1);
```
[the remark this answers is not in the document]
mralj: Because 1 task is reader task and `_processConcurrency` is number of workers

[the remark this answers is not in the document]
mralj: @asdacap maybe better naiming would help here, smth. like:
`_processConcurrency` -> `_workerTaskCount` (or just `workerCount`) 
and  name this constant 1, eg.

```
const int readerTaskCount = 1
using ArrayPoolList<Task> allTasks = new(_workerTaskCount + readerTaskCount );

````

File: src/Nethermind/Nethermind.JsonRpc/WebSockets/JsonRpcSocketsClient.cs (hunk 5 above)
Comment (mralj) [reply src/Nethermind/Nethermind.JsonRpc/WebSockets/JsonRpcSocketsClient.cs:55]: If I'm not mistaken, it's actually single reader AND single writer, but multiple workers.

File: src/Nethermind/Nethermind.JsonRpc/IJsonRpcConfig.cs (hunk 4 above)
Comment (asdacap) [reply src/Nethermind/Nethermind.JsonRpc/IJsonRpcConfig.cs:177]: Increasing it will cause the response to be handled potentially out of order which might break some client. So it set to 1 by default.

File: src/Nethermind/Nethermind.JsonRpc/WebSockets/JsonRpcSocketsClient.cs (hunk 5 above)
Comment (asdacap) [reply src/Nethermind/Nethermind.JsonRpc/WebSockets/JsonRpcSocketsClient.cs:55]: single writer, multiple reader.

**Comment by benaadams** [comment]
`dotnet format` issue; though I think its a bit rogue currently

File: src/Nethermind/Nethermind.JsonRpc/WebSockets/JsonRpcSocketsClient.cs [hunk 11]
```diff
@@ -54,17 +68,61 @@ public override void Dispose()
...
+    }
+
+    public override async Task ReceiveLoopAsync(CancellationToken cancellationToken)
+    {
+        using AutoCancelTokenSource cts = cancellationToken.CreateChildTokenSource();
+
+        using ArrayPoolList<Task> allTasks = new(_workerTaskCount + 1);
+        allTasks.Add(Task.Factory.StartNew(async () =>
+        {
+            try
```
benaadams, line 89: Maybe throw in some thread names since live forever (till process end)
```suggestion
            Thread.CurrentThread.Name = "..."
            try
```

benaadams (approved):

src/Nethermind/Nethermind.JsonRpc/WebSockets/JsonRpcSocketsClient.cs
```diff
@@ -54,17 +68,61 @@ public override void Dispose()
...
+            {
+                await base.ReceiveLoopAsync(cts.Token);
+            }
+            finally
+            {
+                _processChannel.Writer.Complete();
+            }
+        }, TaskCreationOptions.LongRunning));
```
LukaszRozmej, line 97: I thought we discouraged using long running?

mralj:
I've been testing this for past 1.5/2 days in various scenarios, and I'm getting better results than initial ones. It seems that I was too conservative with `IpcProcessingConcurrency` flag (when setting it to 4), bumping this number to 32/64 helped, here are some ScreenShots:

CleanShot 2025-03-06 at 10 00 55@2x

CleanShot 2025-03-07 at 08 51 26@2x

The test are bit shorter than the one I'm comparing it with, but the number of req is 1M+, IMO this should be pretty good sample.
Now Average and p95/99/99.9 are IMO significantly improved at 15-20% latency drop :)

mralj:
@LukaszRozmej  regarding your comment about `LongRunning` flag, I [suggested](https://github.com/NethermindEth/nethermind/pull/8292#discussion_r1981489427) it to @asdacap because I misunderstood how it works. You are correct, it should be removed.

Reading [this blog](https://blog.i3arnon.com/2015/07/02/task-run-long-running/) I see it's useless in this scenario, in fact unless I'm misunderstanding things, both `Task.Run` and `Task.Factory.New with LongRunning` are useless in scenarios where task hits `.await` _quickly_ (like here), so I think @asdacap  original implementation is actually the correct one.

src/Nethermind/Nethermind.JsonRpc/WebSockets/JsonRpcSocketsClient.cs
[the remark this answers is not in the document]
benaadams: Actually no point in being LongRunning since `base.ReceiveLoopAsync` has awaits and the threads will be lost anyway

LukaszRozmej (approved):

src/Nethermind/Nethermind.JsonRpc.Test/JsonRpcSocketsClientTests.cs
```diff
@@ -146,6 +150,79 @@ static async Task<int> CountNumberOfMessages(Socket socket, CancellationToken to
...
+                Interlocked.Increment(ref concurrentCall);
+                await completeSource.Task;
+                yield return JsonRpcResult.Single(new JsonRpcSuccessResponse(null), new RpcReport());
+            }
+
+            jsonRpcProcessor
+                .ProcessAsync(Arg.Any<PipeReader>(), Arg.Any<JsonRpcContext>())
+                .Returns(ResponseFunc);
+
+            for (int i = 0; i < concurrencyLevel; i++)
```
emlautarom1, line 206: I'm not sure that I understand what we're testing here: we are sending `concurrencyLevel` messages sequentially, but we don't know what we're not checking what we're getting back.

src/Nethermind/Nethermind.Sockets/IpcSocketMessageStream.cs
```diff
@@ -33,9 +35,18 @@ public class IpcSocketMessageStream(Socket socket) : NetworkStream(socket), IMes
             catch { }
         }
 
-        int read = _bufferedDataLength + await Socket.ReceiveAsync(buffer[_bufferedDataLength..], SocketFlags.None);
+        int delimiter = ((IList<byte>)buffer[.._bufferedDataLength]).IndexOf(Delimiter);
```
emlautarom1, line 38: At the end of the day I think they fall back to the same implementation:

```suggestion
        int delimiter = buffer[.._bufferedDataLength].AsSpan().IndexOf(Delimiter);
```

emlautarom1 (changes requested):
We should fix the tests before merging

src/Nethermind/Nethermind.JsonRpc/WebSockets/JsonRpcSocketsClient.cs
```diff
@@ -54,17 +68,61 @@ public override void Dispose()
...
+        using ArrayPoolList<Task> allTasks = new(_workerTaskCount + 1);
+        allTasks.Add(Task.Factory.StartNew(async () =>
+        {
+            try
+            {
+                await base.ReceiveLoopAsync(cts.Token);
+            }
+            finally
+            {
+                _processChannel.Writer.Complete();
```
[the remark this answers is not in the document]
asdacap: So that the worker exit.

mralj (approved):

src/Nethermind/Nethermind.JsonRpc.Test/JsonRpcSocketsClientTests.cs
asdacap: We don't care about the response, just that we can have `concurrencyLevel` of in flight processing.

emlautarom1 (approved):

fipso:
Hello, just want to point out that the newly added config options are not shown when using --help and are not part of the online docs yet

PR #8293, opened 2025-03-04
Refactor/refactor metric controller

asdacap:
- Refactor metric controller to make it easier to add new metric type.
- At high level, guage and dictionary handling is abstracted behind an `IMetricUpdater`.

#### What types of changes does your code introduce?

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

#### If yes, did you write tests?

- [X] Yes
- [ ] No

PR #8294, opened 2025-03-04
Feature/detailed rpc metric

asdacap:
- Require #8293 
- Add support for prometheus's summary metric type.
- Create an option `--JsonRpc.EnablePerMethodMetrics true` which would enable per-method latency metric with 50p, 90p, 95p, 99p percentiles

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

#### Notes on testing

- Seems to work.
Screenshot_2025-03-04_11-04-17


src/Nethermind/Nethermind.Monitoring/Metrics/MetricsController.cs
```diff
@@ -264,6 +285,37 @@ private bool TryCreateMetricUpdater(Type type, Meter? meter, MemberInfo memberIn
...
+                }
+                else
+                {
+                    throw new UnreachableException();
+                }
+
+                _individualUpdater.Add(GetGaugeNameKey(type.Name, memberInfo.Name), metricUpdater);
+                return true;
+            }
+
```
LukaszRozmej, line 318: Can we organize this method better, maybe local functions or something?

src/Nethermind/Nethermind.Monitoring/Metrics/MetricsController.cs
```diff
@@ -264,6 +285,37 @@ private bool TryCreateMetricUpdater(Type type, Meter? meter, MemberInfo memberIn
...
+                    p.SetValue(null, metricUpdater);
+                }
+                else if (memberInfo is FieldInfo f)
+                {
+                    f.SetValue(null, metricUpdater);
+                }
+                else
+                {
+                    throw new UnreachableException();
+                }
```
LukaszRozmej, line 313: Looks like code duplication

LukaszRozmej (approved):

tmeinlschmidt:
btw - docs is missing, the same as show this new config option in `--help`

rubo:
@tmeinlschmidt Docs are published with the release of the respective version. However, if you build from the master branch, `--help` should display the new option. Doesn't it?

tmeinlschmidt:
I've thought that it was already in 1.31.4, but it's not

PR #8298, opened 2025-03-04
Fix/taiko executor

ak88:
Fixes taiko debug and trace modules not using the correct `IBlockProcessor.IBlockTransactionsExecutor`

And adds some default options for taiko db.


ak88:
superceded by #8311

PR #8303, opened 2025-03-05
match on deposit contract log topic

rjnrohit:
This resolves a situation on the Sepolia testnet, which has a different deposit contract. The contract on that network emits two kinds of logs, instead of only deposit events like the deposit contract on mainnet. So we need to skip events with mismatched topics.

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/ExecutionRequests/ExecutionRequestProcessor.cs
+++ b/src/Nethermind/Nethermind.Consensus/ExecutionRequests/ExecutionRequestProcessor.cs
@@ -8,6 +8,7 @@ using System.Linq;
 using Nethermind.Abi;
 using Nethermind.Core;
 using Nethermind.Core.Collections;
+using Nethermind.Core.Crypto;
 using Nethermind.Core.ExecutionRequest;
 using Nethermind.Core.Specs;
 using Nethermind.Crypto;
@@ -25,6 +26,8 @@ public class ExecutionRequestsProcessor : IExecutionRequestsProcessor
     private readonly AbiEncoder _abiEncoder = AbiEncoder.Instance;
 
     private const long GasLimit = 30_000_000L;
+
+    private Hash256 DepositTopic  = new ("0x649bbc62d0e31342afea4e5cd82d4049e7e1ee912fc0889aa790803be39038c5");
 
     private readonly ITransactionProcessor _transactionProcessor;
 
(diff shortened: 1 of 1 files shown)


LukaszRozmej (approved):

src/Nethermind/Nethermind.Consensus/ExecutionRequests/ExecutionRequestProcessor.cs
```diff
@@ -26,6 +27,8 @@ public class ExecutionRequestsProcessor : IExecutionRequestsProcessor
 
     private const long GasLimit = 30_000_000L;
 
+    private Hash256 DepositTopic = new("0x649bbc62d0e31342afea4e5cd82d4049e7e1ee912fc0889aa790803be39038c5");
```
LukaszRozmej, line 30: You can use `_depositEventABI.Hash` instead - just validate it matches this hash.

smartprogrammer93 (approved):

MarekM25 (commented):
I am not sure if this is a correct fix for chiado as well

https://gnosis-chiado.blockscout.com/tx/0xf3f87ff98d88e20b2ae852c637dddad2b3cff6f8646bf0685da507b8e1191409?tab=logs

rjnrohit:
https://discord.com/channels/595666850260713488/892088344438255616/1346780623712550922

MarekM25 (approved):

src/Nethermind/Nethermind.Consensus/ExecutionRequests/ExecutionRequestProcessor.cs
```diff
@@ -73,7 +74,7 @@ public void ProcessDeposits(TxReceipt[] receipts, IReleaseSpec spec, ArrayPoolLi
                 for (var j = 0; j < logEntries.Length; j++)
                 {
                     LogEntry log = logEntries[j];
-                    if (log.Address == spec.DepositContractAddress)
+                    if (log.Address == spec.DepositContractAddress && log.Topics.Length >= 1 && log.Topics[0] == DepositEventAbi.Hash)
```
MarekM25, line 77: Let's use hardcoded version

LukaszRozmej: We are already using ABI for parsing, this is actually safer.

Marchhill, line 77: Will it always be the first topic?

Marchhill (approved):

LukaszRozmej:
Use #8305

PR #8310, opened 2025-03-05
Add client id privacy option

### LINKED ISSUE #4566: Possibility to hide Client Name from Peers for Privacy
As per Discord thread, user want to hide his Client name for privacy purpose:
"can i remove the client name from the p2p identity messages for privacy?"

Maybe it would be nice to have config property with this value and to be editable in any specific way (keeping info about Nethermind and its version but removing rest of the path)

We may have some flag where on startup we can define, what should be hidden to the external audience keeping all functionalities working and keeping all stats on services up to date.

octogene:
Resolves #4566

## Changes

-  Add --network.PublicClientIdFormat
- Update sending HelloMessage with anonymized client id if enabled
- Update web3_clientVersion rpc method to return anonymized client id if enabled

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

## Remarks

Are there other places where the public client id should be used ? (Shutter ?)

### DIFF
--- a/src/Nethermind/Nethermind.Core/ProductInfo.cs
+++ b/src/Nethermind/Nethermind.Core/ProductInfo.cs
@@ -18,7 +18,9 @@ public static class ProductInfo
         var metadataAttrs = assembly.GetCustomAttributes<AssemblyMetadataAttribute>()!;
         var productAttr = assembly.GetCustomAttribute<AssemblyProductAttribute>()!;
         var versionAttr = assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()!;
-        var timestamp = metadataAttrs?.FirstOrDefault(static a => a.Key.Equals("BuildTimestamp", StringComparison.Ordinal))?.Value;
+        var timestamp = metadataAttrs
+            ?.FirstOrDefault(static a => a.Key.Equals("BuildTimestamp", StringComparison.Ordinal))
+            ?.Value;
 
         BuildTimestamp = long.TryParse(timestamp, out var t)
             ? DateTimeOffset.FromUnixTimeSeconds(t)
@@ -94,7 +96,7 @@ public static class ProductInfo
     private static Dictionary<string, string> ClientIdParts { get; }
 
     public static string PublicClientId { get; private set; }
-    
+
     public const string DefaultPublicClientIdFormat = "{name}/{version}/{os}/{runtime}";
 
     public static void InitializePublicClientId(string formatString)
--- a/src/Nethermind/Nethermind.Core/ProductInfo.cs
+++ b/src/Nethermind/Nethermind.Core/ProductInfo.cs
@@ -99,8 +99,6 @@ public static class ProductInfo
 
     public const string DefaultPublicClientIdFormat = "{name}/{version}/{os}/{runtime}";
 
-    public static void InitializePublicClientId(string formatString)
-    {
+    public static void InitializePublicClientId(string formatString) =>
         PublicClientId = FormatClientId(formatString);
-    }
 }
--- a/src/Nethermind/Nethermind.Core/ProductInfo.cs
+++ b/src/Nethermind/Nethermind.Core/ProductInfo.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.InteropServices;
+using System.Text;
 
 namespace Nethermind.Core;
 
@@ -44,7 +45,7 @@ public static class ProductInfo
             { "runtime", $"dotnet{Runtime[5..]}" }
         };
 
-        ClientId = FormatClientId("{name}/{version}/{os}/{runtime}");
+        ClientId = FormatClientId(DefaultPublicClientIdFormat);
         PublicClientId = ClientId;
     }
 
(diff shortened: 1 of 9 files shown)


src/Nethermind/Nethermind.Network.Contract/Config/INetworkConfig.cs
```diff
@@ -103,4 +103,7 @@ public interface INetworkConfig : IConfig
 
     [ConfigItem(DefaultValue = "false", HiddenFromDocs = true, Description = "[TECHNICAL] Shutdown timeout when closing TCP port.")]
     long RlpxHostShutdownCloseTimeoutMs { get; set; }
+
+    [ConfigItem(DefaultValue = null, Description = "A comma-separated list of the part names of the client id to be masked.")]
+    string? ClientIdHiddenParts { get; set; }
```
LukaszRozmej, line 108: It needs some more information how the format looks like and what parts are there.

src/Nethermind/Nethermind.Core/ProductInfo.cs
```diff
@@ -35,11 +35,43 @@ static ProductInfo()
...
+            $"{OS.ToLowerInvariant()}-{OSArchitecture}",
+            $"dotnet{Runtime[5..]}"
+        ];
+
+        ClientId = string.Join("/", ClientIdParts);
     }
 
     public static DateTimeOffset BuildTimestamp { get; }
 
+    public static string FormatClientId(string? hiddenPartsName = null)
```
LukaszRozmej, line 51: Instead of passing everywhere `hiddenPartsName` and calling this, better to configure it once here and cache it in some `PublicClientId` property.

src/Nethermind/Nethermind.Network.Contract/Config/INetworkConfig.cs
LukaszRozmej: Other idea is instead of providing list of parts to hide, provide a template string, that the parts would be interpolated to. Then use our standard format as default value. I think it would be more flexible.

octogene: Good idea, I've simplified the implementation to effectively provide a public client id template. It is pretty permissive but if allowing users to put anything as the public client id is not an issue i can keep it as is.

src/Nethermind/Nethermind.Core/ProductInfo.cs
octogene: I've initialized the PublicClientId in the network initialization, so no need to pass any value anymore. Furthermore the PublicClientId is initialized to the ClientId by default.
But i'm not sure if it's the best place to initialize it as i don't have a good grasp of the full initialization process. So is it ok or is there another place to initialize it in order to avoid having some discrepancy between the usage of the PublicClientId ?

src/Nethermind/Nethermind.Init/Steps/InitializeNetwork.cs
```diff
@@ -203,8 +203,11 @@ await StartDiscovery().ContinueWith(initDiscoveryTask =>
             throw new InvalidOperationException("Cannot initialize network without knowing own enode");
         }
 
+        ProductInfo.InitializePublicClientId(_networkConfig.PublicClientIdFormat);
+
         ThisNodeInfo.AddInfo("Ethereum     :", $"tcp://{_api.Enode.HostIp}:{_api.Enode.Port}");
         ThisNodeInfo.AddInfo("Client id    :", ProductInfo.ClientId);
+        ThisNodeInfo.AddInfo("Public client id    :", ProductInfo.PublicClientId);
```
LukaszRozmej, line 210: Is this nicely aligned?

src/Nethermind/Nethermind.Core/ProductInfo.cs
```diff
@@ -35,11 +36,27 @@ static ProductInfo()
...
+
+        ClientId = FormatClientId("{name}/{version}/{os}/{runtime}");
+        PublicClientId = ClientId;
     }
 
     public static DateTimeOffset BuildTimestamp { get; }
 
+    private static string FormatClientId(string formatString)
+    {
+        return ClientIdParts.Aggregate(formatString, (current, placeholder) =>
```
LukaszRozmej, line 55: minor: could use StringBuilder

src/Nethermind/Nethermind.Network.Contract/Config/INetworkConfig.cs
```diff
@@ -10,6 +10,7 @@ public interface INetworkConfig : IConfig
     public const int DefaultNettyArenaOrder = -1;
     public const int MaxNettyArenaOrder = 14;
     public const int DefaultMaxNettyArenaCount = 8;
+    public const string DefaultPublicClientIdFormat = "{name}/{version}/{os}/{runtime}";
```
LukaszRozmej, line 13: Define that const in `ProductInfo` as it is duplicated there

src/Nethermind/Nethermind.Network/ProtocolsManager.cs
```diff
@@ -116,6 +116,7 @@ public ProtocolsManager(
                 _clientIdPattern = new Regex(networkConfig.ClientIdMatcher, RegexOptions.Compiled);
             }
 
+
```
LukaszRozmej, line 119: no need to modify

LukaszRozmej (approved):

src/Nethermind/Nethermind.Init/Steps/InitializeNetwork.cs
octogene: Changed it to "Public id" to avoid having to make changes in all the other places where AddInfo is used, should be explicit enough. If not i'll make the changes to align everything.

LukaszRozmej (approved):

rubo (approved):

PR #8311, opened 2025-03-05
Use Taiko's Transaction Executors for RPC modules

smartprogrammer93:
## Changes

- Use TaikoBlockValidationTransactionExecutor.cs for trace, debug and proof modules so anchor tx wont fail in generate proof

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

## Documentation

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

now the nethermind client can be used to generate proofs for the sgx and zkProvers for Taiko stack


LukaszRozmej (approved):
We could probably simplify it by providing TxProcessingEnvFactory to the original module?

benaadams (approved):

src/Nethermind/Nethermind.Taiko/TaikoReadOnlyTxProcessingEnvFactory.cs
ak88, line 1: this is not used

ak88 (approved):
LGTM just that one unused class

src/Nethermind/Nethermind.Taiko/TaikoReadOnlyTxProcessingEnvFactory.cs
smartprogrammer93: true, but didnt want to remove it in case we want to use it for something later.

flcl42 (approved):

PR #8313, opened 2025-03-06
fix shutter test execute engine tests

asdacap:
- Some shutter test subclasses `EngineModuleTests` without actually changing any behaviour. It just need the rpc factory method and some util function. This causes it to execute some engine module tests also.
- This PR moves those util class ot `BaseEngineModuleTests` so that shutter tests does not execute the engine tests again.

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Merge.Plugin.Test/BaseEngineModuleTests.cs
+++ b/src/Nethermind/Nethermind.Merge.Plugin.Test/BaseEngineModuleTests.cs
@@ -1,10 +1,11 @@
-// SPDX-FileCopyrightText: 2022 Demerzel Solutions Limited
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
 // SPDX-License-Identifier: LGPL-3.0-only
 
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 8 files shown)


src/Nethermind/Nethermind.Merge.Plugin.Test/BaseEngineModuleTests.cs
```diff
@@ -150,6 +154,75 @@ protected IEngineRpcModule CreateEngineModule(MergeTestBlockchain chain, ISyncCo
...
+        {
+            BlockNumber = head.Number,
+            BlockHash = head.Hash!,
+            StateRoot = head.StateRoot!,
+            ReceiptsRoot = head.ReceiptsRoot!,
+            GasLimit = head.GasLimit,
+            Timestamp = head.Timestamp,
+            BaseFeePerGas = head.BaseFeePerGas,
+        };
+    }
```
LukaszRozmej, line 224: Wouldn't those be better as extension methods?

LukaszRozmej (approved):

PR #8321, opened 2025-03-07
Feature/detailed metric

asdacap:
- Add `DetailedMetricAttribute` and  `--Metric.EnableDetailedMetric` which are only published when set.
  - Some candidate include any metric heavy number that is not really important for non dev such as time in sync dispatcher, time for each message, size of each message time, etc.
- Should this be leveled instead of only "detail"? I don't know. Maybe. Tags I guess? Separate config for specific use case?
- Add `ExponentialPowerHistogramMetric` and set to block processing time and mgas per sec per block.
- Fix index out of bound exception on some labelled metric.
- Note: With histogram you need to specify the buckets of value. So the ballpark figure need to be known ahead of time. Additionally, the value resolution depends on the number of bucket and each bucket appears as a separate tag in prometheus. This means it is kinda heavy on the metric server side.

Screenshot_2025-03-07_07-59-18
Screenshot_2025-03-07_07-55-17

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Seems to work.

### DIFF
--- a/src/Nethermind/Nethermind.Core/Metric/IMetricObserver.cs
+++ b/src/Nethermind/Nethermind.Core/Metric/IMetricObserver.cs
@@ -6,4 +6,5 @@ namespace Nethermind.Core.Metric;
 public interface IMetricObserver
 {
     public void Observe(IMetricLabels labels, double value);
+    public void Observe(double value);
 }
--- a/src/Nethermind/Nethermind.Blockchain/Metrics.cs
+++ b/src/Nethermind/Nethermind.Blockchain/Metrics.cs
@@ -4,6 +4,7 @@
 using System.ComponentModel;
 using System.Runtime.Serialization;
 using Nethermind.Core.Attributes;
+using Nethermind.Core.Metric;
 using Nethermind.Int256;
 // ReSharper disable InconsistentNaming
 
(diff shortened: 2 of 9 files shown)


src/Nethermind/Nethermind.Core/Metric/IMetricObserver.cs
```diff
@@ -6,4 +6,5 @@ namespace Nethermind.Core.Metric;
 public interface IMetricObserver
 {
     public void Observe(IMetricLabels labels, double value);
+    public void Observe(double value);
```
LukaszRozmej, line 9: Can both methods be collapsed into:
```suggestion
    public void Observe(double value, IMetricLabels? labels = null);
```

emlautarom1 (approved):
LGTM.

emlautarom1:
> Should this be leveled instead of only "detail"? I don't know. Maybe. Tags I guess? Separate config for specific use case?

I agree, collapsing metrics options would be better than having multiple flags (ex. `--Metrics=Off|On|Detailed`) but for now I think it's fine.

PR #8324, opened 2025-03-07
Perf/Faster SynchronizerTests

asdacap:
- ~~SyncDispatcher does not exit on dispose randomly.~~ still happening.
- ConfigProvider's config discovery is very slow.
- Disable pruning triestore in the test as it is slow to start with all those dirty caches.
- Increase E2E setup timeout.
- `NSubsitute<ISyncPeer>` is confusingly really slow
- Fix a legitimate reorg bug.

- [X] Optimization

#### Requires testing

- [ ] Yes
- [X] No

- [X] Yes
- [ ] No


src/Nethermind/Nethermind.Config/ConfigProvider.cs
```diff
@@ -86,14 +88,32 @@ public void AddSource(IConfigSource configSource)
         _configSource.Add(configSource);
     }
 
-    public void Initialize()
+    private static IList<(Type ConfigType, Type DirectImplementation)> GatherConfigsFromAssembly()
     {
+        if (_configTypesCache is not null) return _configTypesCache;
```
LukaszRozmej, line 93: Can you make it to self-initialized property that calls this method internally? IMO it is more readable and standard.

PR #8326, opened 2025-03-07
Revert secp256r1 optimization

alexb5dh:
Reverts #8284 until crash is fixed.


src/Nethermind/Nethermind.Evm/Precompiles/Secp256r1Precompile.cs
```diff
@@ -19,12 +19,34 @@ public class Secp256r1Precompile : IPrecompile<Secp256r1Precompile>
...
+        {
+            ecdsa = ECDsa.Create(new ECParameters
+            {
+                Curve = ECCurve.NamedCurves.nistP256,
+                Q = new() { X = x.ToArray(), Y = y.ToArray() }
+            });
+        }
+        catch
+        {
+            // Invalid x/y parameters
```
alexb5dh, line 45: No option to validate this in .NET without exception handling.

LukaszRozmej (approved):

PR #8332, opened 2025-03-10
Feature/Load ISteps with a container.

asdacap:
- Load `ISteps` with a container instead of activator.
- Allow constructor to have component other than `NethermindApi`.
- Move `EraModule` to top level container.

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- [X] Mainnet
- [X] Taiko
- [X] Optimism


src/Nethermind/Nethermind.Core/Container/FallbackToFieldFromApi.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        PropertyInfo? property;
+        Type serviceType = ts.ServiceType;
+        if (registrationAccessor(service).Any())
+        {
+            // Already have registration
+            if (!_allowRedundantRegistration && _availableTypes.TryGetValue(serviceType, out property) && property.SetMethod != null)
+            {
+                // To prevent mistake, a service that already have registration via dependency injection must not also
+                // have a setter in api. This is to prevent the assumption that the setter will caause the service
```
LukaszRozmej, line 69:
```suggestion
                // have a setter in api. This is to prevent the assumption that the setter will cause the service
```

src/Nethermind/Nethermind.Runner/Ethereum/Api/ApiBuilder.cs
```diff
@@ -24,18 +28,62 @@ public class ApiBuilder
...
+        FollowOtherMiners gasLimitCalculator = new FollowOtherMiners(specProvider);
+
+        IConsensusPlugin? enginePlugin = consensusPlugins.FirstOrDefault();
+        INethermindApi nethermindApi =
+            enginePlugin?.CreateApi(_configProvider, _jsonSerializer, _logManager, ChainSpec) ??
+            new NethermindApi(_configProvider, _jsonSerializer, _logManager, ChainSpec);
+        nethermindApi.SpecProvider = specProvider;
+        nethermindApi.GasLimitCalculator = gasLimitCalculator;
+        nethermindApi.ProcessExit = _processExitSource;
+        ((List<INethermindPlugin>)nethermindApi.Plugins).AddRange(plugins);
```
LukaszRozmej, line 69: This looks kind of off, can we design not to need this cast?

LukaszRozmej (approved):
minor `is null` `== null` inconsistencies.

src/Nethermind/Nethermind.Runner/Ethereum/Api/ApiBuilder.cs
LukaszRozmej: Yeah I know, maybe there should exist some plugin collection that should be accessed separately, and only have read-only access through api?

PR #8333, opened 2025-03-10
Fix/TrieException after full pruning

asdacap:
- Fix triexception after full pruning due to persisted node not being removed on `PersistCache`.

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Four consecutive mainnet full prune


LukaszRozmej (approved):

LukaszRozmej:
Exception was after FullPruning?

asdacap:
Yes full pruning, not memory pruning.

PR #8334, opened 2025-03-10
Fix/random debug rpc module test failure

asdacap:
- Fix random debug rpc failure due to the use of global IWorldState.
- Removed global IWorldState from TestBlockchain, replaced with ReadOnlyState.
- Fix inconsistent block number after add block. Does not seems to fix the test though.

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Core.Test/Blockchain/TestBlockchain.cs
+++ b/src/Nethermind/Nethermind.Core.Test/Blockchain/TestBlockchain.cs
@@ -136,37 +136,36 @@ public class TestBlockchain : IDisposable
         EthereumEcdsa = new EthereumEcdsa(SpecProvider.ChainId);
         DbProvider = await CreateDbProvider();
         TrieStore = new TrieStore(StateDb, LogManager);
-        var State = new WorldState(TrieStore, DbProvider.CodeDb, LogManager, PreBlockCaches);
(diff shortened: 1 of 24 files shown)


asdacap:
First try!

src/Nethermind/Nethermind.Core.Test/Blockchain/TestBlockchain.cs
```diff
@@ -135,7 +136,7 @@ protected virtual async Task<TestBlockchain> Build(ISpecProvider? specProvider =
         EthereumEcdsa = new EthereumEcdsa(SpecProvider.ChainId);
         DbProvider = await CreateDbProvider();
         TrieStore = new TrieStore(StateDb, LogManager);
-        State = new WorldState(TrieStore, DbProvider.CodeDb, LogManager, PreBlockCaches);
+        var State = new WorldState(TrieStore, DbProvider.CodeDb, LogManager, PreBlockCaches);
```
LukaszRozmej, line 139:
```suggestion
        var state = new WorldState(TrieStore, DbProvider.CodeDb, LogManager, PreBlockCaches);
```

LukaszRozmej (approved):

src/Nethermind/Nethermind.Core.Test/Blockchain/TestBlockchain.cs
asdacap: Done

PR #8344, opened 2025-03-10
Fix 2935 - fetch trimmed hashes correctly

tanishqjasoria:
- pad left for blockHash.Length < 32

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Remarks


asdacap (approved):

LukaszRozmej (approved):
Lets wait for @MarekM25 review

MarekM25 (approved):

MarekM25 (commented):
What about other values we read from the state, do we use PadLeft?

MarekM25 (commented):
Another potential issue: if value is not present, in SLOAD we will return byte[0] (if I am not mistaken) while here we return Bytes32

@tanishqjasoria could you double check what is correct?

tanishqjasoria:
> Another potential issue: if value is not present, in SLOAD we will return byte[0] (if I am not mistaken) while here we return Bytes32
> 
> @tanishqjasoria could you double check what is correct?

yes, it is correct. Here we return according to the BlockHash opCode - where we push a empty array of 32 bytes on the stack.

PR #8348, opened 2025-03-11
Make TxPool pending thread safe

LukaszRozmej:
## Changes

- Waits till Tx pool gets updated from background thread before giving pending information

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [x] Other: thread safety consistency

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.TxPool/TxPool.cs
+++ b/src/Nethermind/Nethermind.TxPool/TxPool.cs
@@ -239,14 +239,14 @@ namespace Nethermind.TxPool
 
             try
             {
-                if (++_headsQueueCount == 1)
+                if (Interlocked.Increment(ref _headsQueueCount) == 1)
                 {
                     _headBlocksResetEvent.Reset();
                 }
 
                 if (!_headBlocksChannel.Writer.TryWrite(e))
                 {
-                    _headsQueueCount--;
+                    Interlocked.Decrement(ref _headsQueueCount);
                     _headBlocksResetEvent.Set();
                 }
             }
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -262,7 +299,16 @@ private void ProcessNewHeads()
...
                         catch (Exception e)
                         {
-                            if (_logger.IsDebug) _logger.Debug($"TxPool failed to update after block {args.Block.ToString(Block.Format.FullHashAndNumber)} with exception {e}");
+                            if (_logger.IsDebug)
+                                _logger.Debug(
+                                    $"TxPool failed to update after block {args.Block.ToString(Block.Format.FullHashAndNumber)} with exception {e}");
+                        }
+                        finally
+                        {
+                            if (--_headsQueueCount == 0)
```
Scooletz, line 308: This is a non-atomic field that is updated from several places: `ProcessNewHeads` and `OnHeadChange`, why this should work?

Scooletz (changes requested):
What is the case where we want to wait on the event here? It's not clear from the description. No tests provided. There should be some.

src/Nethermind/Nethermind.TxPool/TxPool.cs
LukaszRozmej: yeah true, I was thinking that single reader and writer is enough, but we do need to make it interlocked

LukaszRozmej:
> What is the case where we want to wait on the event here? It's not clear from the description. No tests provided. There should be some.

Check assosiated PR

marcindsobczak (commented):
Code looks good, but we should consider if such changes are justified just for a micro-optimization. If we decide to go with it, it would be good to add some tests.

stdevMac:
Closing along with its companion #8345: the stale-snapshot race in TxPool pending that this draft targeted was fixed by merged #10604 (2026-05-20), and the review feedback here was never picked back up.

PR #8352, opened 2025-03-11
Accept tx with same nonce when pending or is delegated

ak88:
Change `DelegatedAccountFilter` to have the same rule for pending or active delegated account. 

## Changes

A tx with a pending delegation will also be accepted, if the nonce corresponds to the current account nonce. 

This will keep the max allowed number of tx for pending or delegated account to 1, and also ensures that an account is always able to send a replacement. 

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_


marcindsobczak (approved):
Seems reasonable. Code looks good

benaadams (approved):

PR #8353, opened 2025-03-11
fix Trace leak

LukaszRozmej:
## Changes

- fix Trace leak

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Tracing/Tracer.cs
+++ b/src/Nethermind/Nethermind.Consensus/Tracing/Tracer.cs
@@ -6,59 +6,46 @@ using Nethermind.Consensus.Processing;
 using Nethermind.Core;
 using Nethermind.Core.Crypto;
 using Nethermind.Evm.Tracing;
-using Nethermind.State;
+using Nethermind.Evm.TransactionProcessing;
 using Nethermind.Trie;
 
 namespace Nethermind.Consensus.Tracing
 {
-    public class Tracer : ITracer
+    public class Tracer(
+        IReadOnlyTxProcessingScope scope,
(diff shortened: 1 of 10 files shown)


MarekM25 (commented):
My understanding after reading the code: every request increases the memory state in the database, and there was no reset, which is why we had a memory leak. Is this understanding correct?

If yes, LGTM

src/Nethermind/Nethermind.Consensus/Tracing/Tracer.cs
```diff
@@ -6,59 +6,46 @@
...
 
-            blockTracer.StartNewBlockTrace(block);
-
             try
             {
+                blockTracer.StartNewBlockTrace(block);
                 processor.Process(block, options, blockTracer);
+                blockTracer.EndBlockTrace();
             }
-            catch (Exception)
```
MarekM25, line 43: hmmm catch might be still needed, no?

brbrr (approved):
LGTM from the testing perspective

src/Nethermind/Nethermind.Consensus/Tracing/Tracer.cs
LukaszRozmej: not needed, Reset in finally suffices

MarekM25 (approved):

asdacap (approved):

rjnrohit (approved):

PR #7930, opened 2024-12-18
StorageRange request refactoring for parallel execution

damian-orzechowski:
Refactoring Storage Range requests in snap sync to pass through original request object when spliting accounts storage range for parallel execution. Additional change to #7733 

## Changes

- _List the changes_

#### What types of changes does your code introduce?

- [ ] Bugfix (a non-breaking change that fixes an issue)
- [ ] New feature (a non-breaking change that adds functionality)
- [ ] Breaking change (a change that causes existing functionality not to work as expected)
- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [ ] No

#### If yes, did you write tests?

- [ ] Yes
- [ ] No

#### Notes on testing

_Optional. Remove if not applicable._

#### Requires documentation update

- [ ] Yes
- [ ] No

_If yes, link the PR to the docs update or the issue with the details labeled `docs`. Remove if not applicable._

#### Requires explanation in Release Notes

- [ ] Yes
- [ ] No

_If yes, fill in the details here. Remove if not applicable._

## Remarks

_Optional. Remove if not applicable._

PR #8356, opened 2025-03-11
Feature/disable storage range split flag

asdacap:
- Causes snap sync hangs sometimes. Unfortunately, not always. 

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No


damian-orzechowski (commented):
I'd probably prefer to have an enable flag and defaulted to `false`

MarekM25 (commented):
I agree with Łukasz. Let's disable by default broken feature and fix later

asdacap:
I'll wait for some time in case @damian-orzechowski figure it out.

damian-orzechowski:
Let's merge this for the time being.
I managed to recreate the scenario (on energyweb) and there is one storage ranges requests stuck in the queue. The request was created by splitting the storage trie in chunks, so it might be related to #7930. The request always get an empty message as a response which is interpreted as expired root hash and forces indefinite pivot update.

PR #8359, opened 2025-03-13
Fix flashbots prewarmer not wired correctly

asdacap:
- Fix flashbots's rpc module prewarmer not wired correctly. Implicitly the read only processing env factory should have the world state to warm up.
- Fix block processor re-created on each call, which make sense for some application, but generally we keep the instance and recycle it.
- Create `ReadOnlyProcessingEnvFactory.CreateForWarmingUp` which require passing the world state to warm up.
- Prewarmer now need the world state to warm up passed in its constructor to make it clear which world state it is warming up instead of implicitly in the instance of `ReadOnlyProcessingEnvFactory`.

- [X] Optimization

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

#### Notes on testing

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
@@ -23,13 +23,13 @@ using Nethermind.Trie;
 
 namespace Nethermind.Consensus.Processing;
 
-public sealed class BlockCachePreWarmer(ReadOnlyTxProcessingEnvFactory envFactory, ISpecProvider specProvider, int concurrency, ILogManager logManager, PreBlockCaches? preBlockCaches = null) : IBlockCachePreWarmer
+public sealed class BlockCachePreWarmer(ReadOnlyTxProcessingEnvFactory envFactory, IWorldState worldStateToWarmup, ISpecProvider specProvider, int concurrency, ILogManager logManager, PreBlockCaches? preBlockCaches = null) : IBlockCachePreWarmer
 {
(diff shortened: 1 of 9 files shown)


rjnrohit (approved):
Looks good, needs to be tested though @mralj

mralj:
> Looks good, needs to be tested though @mralj
I will let you know once I test :)

mralj (approved):
Works! 😁

asdacap:
Nice, any change in latency? How do you run it?

mralj:
> Nice, any change in latency? How do you run it?

I'm only checking latency for IPC calls and this one is HTTP (and it's not even derenctly from `rbuilder` but 
`rbuilder -> test_relay -> validation call` so I cannot say :(

This is how I'm running node: 

```
./nethermind -c mainnet  --data-dir ~/execution-data --jsonrpc-jwtsecretfile ~/jwt.hex  --healthchecks-enabled true --jsonrpc-enabledmodules [Eth,Subscribe,Trace,TxPool,Web3,Personal,Proof,Net,Parity,Health,Rpc,rbuilder,admin] --JsonRpc.Host 0.0.0.0 --jsonrpc-ipcunixdomainsocketpath ~/execution-data/nm.ipc --sync-nonvalidatornode true --jsonrpc-websocketsport 8546 --Sync.DownloadReceiptsInFastSync false  --JsonRpc.IpcProcessingConcurrency 16 --JsonRpc.WebSocketsProcessingConcurrency 2 --TxPool.BlobsSupport InMemory
```

And this is [the branch](https://github.com/NethermindEth/nethermind/tree/mralj/rbuilder-ipc-test-latest) I used to test

PR #8396, opened 2025-03-19
Include blob count in block production logs

benaadams:
## Changes

- Include blob count in block production logs

image

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Transactions/CompositeTxSource.cs
+++ b/src/Nethermind/Nethermind.Consensus/Transactions/CompositeTxSource.cs
@@ -11,7 +11,9 @@ namespace Nethermind.Consensus.Transactions
 {
     public class CompositeTxSource : ITxSource
     {
-        private readonly IList<ITxSource> _transactionSources;
+        private readonly List<ITxSource> _transactionSources;
+
+        public bool SupportsBlobs => _transactionSources.Any(s => s.SupportsBlobs);
 
         public CompositeTxSource(params ITxSource[] transactionSources)
         {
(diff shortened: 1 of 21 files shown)


rubo (approved):

src/Nethermind/Nethermind.Consensus/Transactions/CompositeTxSource.cs
```diff
@@ -11,7 +11,9 @@ namespace Nethermind.Consensus.Transactions
 {
     public class CompositeTxSource : ITxSource
     {
-        private readonly IList<ITxSource> _transactionSources;
+        private readonly List<ITxSource> _transactionSources;
+
+        public bool SupportsBlobs => _transactionSources.Any(s => s.SupportsBlobs);
```
LukaszRozmej, line 16: Would be better to calculate once on startup than to recalculate on every call

src/Nethermind/Nethermind.Merge.Plugin/BlockProduction/PayloadPreparationService.cs
```diff
@@ -180,7 +181,16 @@ private void CleanupOldPayloads(object? sender, EventArgs e)
...
-                _logger.Info($" Produced  {blockFees.ToDecimal(null) / weiToEth,5:N3}{BlocksConfig.GasTokenTicker,4} {block.ToString(block.Difficulty != 0 ? Block.Format.HashNumberDiffAndTx : Block.Format.HashNumberMGasAndTx)} | {time.TotalMilliseconds,7:N2} ms");
+                bool supportsBlobs = _blockProducer.SupportsBlobs;
+                int blobs = 0;
+                if (supportsBlobs)
+                {
+                    foreach (Transaction tx in block.Transactions)
+                    {
+                        blobs += tx.BlobVersionedHashes?.Length ?? 0;
+                    }
+                }
```
LukaszRozmej, line 192: Put into a method `CalculateBlobCount`?

PR #8400, opened 2025-03-20
Pack blobs optimally during block production

### LINKED ISSUE #8399: Blob packing in block production is not optimal
**Description**

As raised by @nerolation in research post https://ethresear.ch/t/max-blobs-flag-economic-perspective/21798

@dataalways https://x.com/Data_Always/status/1902697542086721671

And also in the research paper https://arxiv.org/pdf/2502.12966

Blob inclusion is sub-optimal for the builder not taking into account priority fee and packing

This resolves to a [0-1 knapsack problem](https://en.wikipedia.org/wiki/Knapsack_problem#0-1_knapsack_problem); and solving that would produce the optimal solution

benaadams:
Resolves #8399

## Changes

- Solve the 0/1 Knapsack problem to produce optimal blob packing

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Producers/TxPoolTxSource.cs
+++ b/src/Nethermind/Nethermind.Consensus/Producers/TxPoolTxSource.cs
@@ -212,54 +212,66 @@ namespace Nethermind.Consensus.Producers
             int leftoverCapacity,
             ArrayPoolList<Transaction> selectedBlobTxs)
         {
-            // dpFees[c] = the maximum total fee achievable with capacity
-            ulong[] dpFees = new ulong[leftoverCapacity + 1];
+            int size = leftoverCapacity + 1;
+            // The maximum total fee achievable with capacity
+            using ArrayPoolList<ulong> dpFeesPooled = new(capacity: size, count: size);
+            Span<ulong> dpFees = dpFeesPooled.AsSpan();
 
-            // choices[iTx, c] tracks whether the i-th transaction (1-based index) was picked 
-            // when building the solution for capacity c.
-            // 0 means "not picked", 1 means "picked".
-            int[,] choices = new int[candidateTxs.Count, leftoverCapacity + 1];
+            // Index of the transaction that gave an improved dpFees or -1 if none
+            using ArrayPoolList<int> picksPooled = new(capacity: size, count: size, fill: -1);
+            Span<int> picks = picksPooled.AsSpan();
 
             // Build up the DP table to find the maximum total fee for each capacity.
-            // Outer loop: go through each transaction (1-based index).
-            // Inner loop: iterate capacity in descending order to avoid overwriting data needed for the calculation.
-            for (int i = 0; i < candidateTxs.Count; i++)
+            int txIndex = 0;
+            foreach (Transaction tx in candidateTxs.AsSpan())
             {
-                Transaction tx = candidateTxs[i];
                 int blobCount = tx.GetBlobCount();
                 ulong feeValue = (ulong)tx.MaxPriorityFeePerGas * (ulong)tx.GasLimit;
 
-                // Iterate backward from maxBlobCapacity down to blobCount 
-                // so we only compute for valid capacities that can fit this transaction.
-                for (int capacity = leftoverCapacity; capacity >= blobCount; capacity--)
+                // Early prune: skip if it doesn't fit at all or has no fee benefit.
+                if (blobCount > leftoverCapacity || feeValue == 0)
                 {
-                    // If we can fit this item in capacity, see if it improves dpFees[capacity]
-                    ulong candidateFee = dpFees[capacity - blobCount] + feeValue;
-                    if (candidateFee > dpFees[capacity])
+                    txIndex++;
+                    continue;
+                }
+
+                // Iterate in descending order so we don't overwrite 
+                // dpFees entries we haven't yet used this iteration
+                for (int i = leftoverCapacity; i >= blobCount; i--)
+                {
+                    // If including txIndex yields a better fee total, update dpFees and picks.
+                    ulong candidateFee = dpFees[i - blobCount] + feeValue;
+                    if (candidateFee > dpFees[i])
                     {
-                        dpFees[capacity] = candidateFee;
-                        choices[i, capacity] = 1; // we picked item i-1
+                        dpFees[i] = candidateFee;
+                        picks[i] = txIndex;
                     }
                 }
+                txIndex++;
             }
 
-            int start = selectedBlobTxs.Count;
-            // Backtrack through 'choices' to find which transactions were actually chosen.
-            int remainingCapacity = leftoverCapacity;
-            for (int i = candidateTxs.Count - 1; i >= 0; i--)
+            // Backtrack to see which transactions were chosen.
+            int addStart = selectedBlobTxs.Count;
+            // We add them in reverse order
+            // Walk downward in capacity, picking whichever transaction 'picks' says was used.
+            while (leftoverCapacity > 0)
             {
-                if (choices[i, remainingCapacity] == 1)
+                int i = picks[leftoverCapacity];
+                if (i == -1)
                 {
-                    Transaction tx = candidateTxs[i];
-                    int blobCount = tx.GetBlobCount();
-                    selectedBlobTxs.Add(tx);
-                    remainingCapacity -= blobCount;
+                    // No transaction was used at this capacity => we can't backtrack further.
+                    break;
                 }
+
+                Transaction chosenTx = candidateTxs[i];
+                selectedBlobTxs.Add(chosenTx);
+
+                leftoverCapacity -= chosenTx.GetBlobCount();
             }
 
             // The newly added items were added in reverse
             // restore original picking order.
-            selectedBlobTxs.AsSpan()[start..].Reverse();
+            selectedBlobTxs.AsSpan()[addStart..].Reverse();
         }
 
         private bool TryGetFullBlobTx(Transaction blobTx, [NotNullWhen(true)] out Transaction? fullBlobTx)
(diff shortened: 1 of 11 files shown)


src/Nethermind/Nethermind.Merge.Plugin/BlockProduction/PayloadPreparationService.cs
```diff
@@ -183,14 +183,20 @@ private void LogProductionResult(Task<Block?> t, Block currentBestBlock, UInt256
...
                 {
                     foreach (Transaction tx in block.Transactions)
                     {
-                        blobs += tx.BlobVersionedHashes?.Length ?? 0;
+                        int blobCount = tx.GetBlobCount();
+                        if (blobCount > 0)
+                        {
+                            blobs += blobCount;
+                            blobTx++;
+                        }
```
LukaszRozmej, line 196: Move this code to helper method?

src/Nethermind/Nethermind.Consensus/Producers/TxPoolTxSource.cs
```diff
@@ -162,14 +153,116 @@ private void SelectBlobTransactions(IEnumerable<Transaction> blobTransactions, B
...
+        {
+            int size = leftoverCapacity + 1;
+            // The maximum total fee achievable with capacity
+            using ArrayPoolList<ulong> dpFeesPooled = new(capacity: size, count: size);
+            Span<ulong> dpFees = dpFeesPooled.AsSpan();
+
+            // choices[iTx, c] tracks whether the i-th transaction (1-based index) was picked 
+            // when building the solution for capacity c.
+            // 0 means "not picked", 1 means "picked".
+            int[,] choices = new int[candidateTxs.Count, leftoverCapacity + 1];
```
LukaszRozmej, line 223: pool?

benaadams: Added pooling

src/Nethermind/Nethermind.Merge.Plugin/BlockProduction/PayloadPreparationService.cs
benaadams: Is only used in one place though; which is this log 🤔

smartprogrammer93 (approved):

src/Nethermind/Nethermind.Consensus/Producers/TxPoolTxSource.cs
```diff
@@ -264,5 +359,27 @@ internal static IEnumerable<Transaction> Order(IDictionary<AddressAsKey, Transac
...
+            public ArrayPoolBitMap(int size)
+            {
+                _array = ArrayPool<ulong>.Shared.Rent(GetLengthOfBitLength(size));
+                _array.AsSpan().Clear();
+            }
+
+            public bool this[int i]
+            {
+                get => (_array[i >> BitShiftPerInt64] & (1UL << i)) != 0;
+                set => _array[(uint)i >> BitShiftPerInt64] |= (1UL << i);
```
Scooletz, line 379: This won't work for `false` unless you do the clearance first. I'd name it differently because in a month or two another person uses it without knowing that they can't `[i] = false` it. Paprika uses the first option https://github.com/NethermindEth/Paprika/blob/5792d74fe1596265059aecd8b017f139403df305/src/Paprika/Data/BitVector.cs#L216-L230

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -171,6 +171,7 @@ protected virtual TransactionResult Execute(Transaction tx, in BlockExecutionCon
             long gasAvailable = tx.GasLimit - intrinsicGas.Standard;
             ExecuteEvmCall(tx, header, spec, tracer, opts, delegationRefunds, intrinsicGas.FloorGas, accessTracker, gasAvailable, env, out TransactionSubstate? substate, out GasConsumed spentGas, out byte statusCode);
             PayFees(tx, header, spec, tracer, substate, spentGas.SpentGas, premiumPerGas, blobBaseFee, statusCode);
+            tx.SpentGas = spentGas.SpentGas;
```
Scooletz, line 174: How `SpentGas` was not needed before ?

src/Nethermind/Nethermind.Core/Transaction.cs
```diff
@@ -52,6 +52,8 @@ public class Transaction
         public bool SupportsBlobs => Type.SupportsBlobs();
         public bool SupportsAuthorizationList => Type.SupportsAuthorizationList();
         public long GasLimit { get; set; }
+        [JsonIgnore]
+        public long? SpentGas { get; set; }
```
Scooletz, line 56: We always fallback on null with `??`. Could we make it long maybe? One less `?`

src/Nethermind/Nethermind.Merge.Plugin/BlockProduction/PayloadPreparationService.cs
```diff
@@ -175,6 +175,7 @@ private void CleanupOldPayloads(object? sender, EventArgs e)
     private void LogProductionResult(Task<Block?> t, Block currentBestBlock, UInt256 blockFees, TimeSpan time)
     {
         const long weiToEth = 1_000_000_000_000_000_000;
+        const long weiToGwei = 1_000_000_000;
```
Scooletz, line 178: Don't we have a shared denomination elsewhere?

src/Nethermind/Nethermind.Blockchain.Test/TransactionSelectorTests.cs
```diff
@@ -183,6 +184,41 @@ public static IEnumerable EnoughShardBlobTransactionsSelectedTestCases
...
+                        .WithMaxFeePerGas(20).WithMaxFeePerBlobGas(1).WithGasLimit(20).WithBlobVersionedHashes([[0]]).SignedAndResolved(TestItem.PrivateKeyF).TestObject,
+                };
+
+                higherPriorityTransactionsSelected.ExpectedSelectedTransactions.AddRange(
+                    higherPriorityTransactionsSelected.Transactions.Where(tx => tx.GetBlobCount() == 1)
+                    .OrderByDescending(t => t.MaxFeePerGas).Take(5));
+                for (int i = 0; i < 20; i++)
+                {
+                    yield return new TestCaseData(higherPriorityTransactionsSelected).SetName($"Correct priority blobs {i:00}");
+                    higherPriorityTransactionsSelected.Transactions.Shuffle(Random.Shared);
```
Scooletz, line 220: `Random.Shared` is not stable. `new Random(seed)` over some const `seed`?

src/Nethermind/Nethermind.Blockchain.Test/TransactionSelectorTests.cs
```diff
@@ -183,6 +184,41 @@ public static IEnumerable EnoughShardBlobTransactionsSelectedTestCases
...
+                    Build.A.Transaction.WithSenderAddress(TestItem.AddressD).WithType(TxType.Blob).WithNonce(1)
+                        .WithMaxFeePerGas(17).WithMaxFeePerBlobGas(1).WithGasLimit(20).WithBlobVersionedHashes([[0]]).SignedAndResolved(TestItem.PrivateKeyD).TestObject,
+                    Build.A.Transaction.WithSenderAddress(TestItem.AddressE).WithType(TxType.Blob).WithNonce(1)
+                        .WithMaxFeePerGas(19).WithMaxFeePerBlobGas(1).WithGasLimit(20).WithBlobVersionedHashes([[0]]).SignedAndResolved(TestItem.PrivateKeyE).TestObject,
+                    Build.A.Transaction.WithSenderAddress(TestItem.AddressF).WithType(TxType.Blob).WithNonce(1)
+                        .WithMaxFeePerGas(20).WithMaxFeePerBlobGas(1).WithGasLimit(20).WithBlobVersionedHashes([[0]]).SignedAndResolved(TestItem.PrivateKeyF).TestObject,
+                };
+
+                higherPriorityTransactionsSelected.ExpectedSelectedTransactions.AddRange(
+                    higherPriorityTransactionsSelected.Transactions.Where(tx => tx.GetBlobCount() == 1)
```
Scooletz, line 215: My understanding is that you omit the biggest one as the others provide the same fee without occupying 5 blobs.

I'd add a test with more variety in ranges to select from. Why not differentiate on the max price, ordering and more. Also, this could be tested in isolation.

Scooletz (changes requested):
Minors

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
benaadams: Is using the prior build build to inform the next block build; isn't needed outside block production

src/Nethermind/Nethermind.Merge.Plugin/BlockProduction/PayloadPreparationService.cs
benaadams: UInt256 based

src/Nethermind/Nethermind.Core/Transaction.cs
benaadams: Made non nullable

LukaszRozmej (approved):

src/Nethermind/Nethermind.Blockchain.Test/TransactionSelectorTests.cs
benaadams: Used fixed seed

src/Nethermind/Nethermind.Blockchain.Test/TransactionSelectorTests.cs
benaadams: Random iterations changes submission ordering (as there are different paths based on how many single blob txs come first)

Other 5 provide higher total fee; isn't how many blobs they use as that's irrelevant to the producer; however the blob cap means can't have the 5 blob tx and the others

Scooletz (approved):
Much better tests now with a stable random, a few cases and a non-trivial distribution of blob fees.

Issue #8394, opened 2025-03-19
eth_simulateV1 fails for op-nmc


mralj:
**Description**
Calling `eth_simulateV1` on op-NMC node fails. 

**Steps to Reproduce**
```Bash
curl https://url.dev \
        -X POST \
        -H "Content-Type: application/json" \
        --data '{"jsonrpc":"2.0","method":"eth_simulateV1","params":[{"blockStateCalls":[{"blockOverrides":{"baseFeePerGas":"0x9"},"stateOverrides":{"0xc000000000000000000000000000000000000000":{"balance":"0x4a817c420"}},"calls":[{"from":"0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045","to":"0x014d023e954bAae7F21E56ed8a5d81b12902684D","maxFeePerGas":"0xf","value":"0x1"}]}],"validation":true,"traceTransfers":true},"latest"],"id":1}'
```

**Actual behavior**

Request errors with:

```JSON
[...]

PR #8423, opened 2025-03-26
Correct timestamp in `eth_simulateV1`

emlautarom1:
Solves #8394

## Changes

- Fix `eth_simulateV1` bug where simulating a non "latest" block would result in the incorrect parent being used.
- Allow `OptimismReceiptMessageDecoder` to decode non `Deposit` receipts.
- Use Optimism's `TransactionProcessor` in `eth_simulateV1` calls.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

Minimal unit tests were added based on the spec for `eth_simulateV1`, which is referenced directly in the code.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks

Please take a look at the associated issue for more details.


src/Nethermind/Nethermind.Optimism/OptimismReceiptMessageDecoder.cs
```diff
@@ -12,10 +12,8 @@ namespace Nethermind.Optimism;
 public class OptimismReceiptTrieDecoder() : OptimismReceiptMessageDecoder(true) { }
 
 [Rlp.Decoder]
-public class OptimismReceiptMessageDecoder(bool isEncodedForTrie = false) : IRlpStreamDecoder<OptimismTxReceipt>, IRlpStreamDecoder<TxReceipt>
+public class OptimismReceiptMessageDecoder(bool isEncodedForTrie = false) : IRlpStreamDecoder<TxReceipt>
```
Scooletz, line 15: Ok, so this is probably the main change. Instead of going with decode/encode op specific receipt and the regular one, we use the common denominator here and handle `TxReceipt` that might be (dynamic check `is`) of the optimistic type.

src/Nethermind/Nethermind.Optimism/SimulateOptimismTransactionProcessor.cs
```diff
@@ -0,0 +1,70 @@
...
+using Nethermind.Evm;
+using Nethermind.Evm.Tracing;
+using Nethermind.Evm.TransactionProcessing;
+using Nethermind.Logging;
+using Nethermind.State;
+using Nethermind.Facade.Simulate;
+
+namespace Nethermind.Optimism;
+
+public sealed class SimulateOptimismTransactionProcessor(
```
Scooletz, line 15: I see that in the future this might require more implementation details. At the same time we create a processor type only to act on one flag: `validate` that could be embedded in the original `OptimismTransactionProcessor` as a special case (with the same flag). Have you considered this option?

Scooletz (commented):
Minor comments provided. I read the description of the PR and yes, I agree that it'd be nice to have some coarse grained tests for op-stack.

src/Nethermind/Nethermind.Optimism/SimulateOptimismTransactionProcessor.cs
emlautarom1: Before considering adding an additional flag to the original `OptimismTransactionProcessor` I thought of adding a decorator `SimulateTransactionProcessor(ITransactionProcessor inner)` but this cannot be easily done since we need to `override` methods that are not part of the interface (ex. `ShouldValidate`), and there is already a hierarchy implemented.

I think this change is the less invasive of all (it only removes the `sealed` modifier to `OptimismTransactionProcessor`) and it's consistent with the existing `SimulateTransactionProcessor`. Personally, I dislike modifying the original code to add special cases or using implementation inheritance to override behavior (unless there is a very good reason to do so).

src/Nethermind/Nethermind.Optimism/OptimismReceiptMessageDecoder.cs
emlautarom1: As you correctly pointed out, this is the main change that avoids triggering an exception. We had a similar case in the past (https://github.com/NethermindEth/nethermind/pull/8157) where we always processed items as if the were Deposit only.

src/Nethermind/Nethermind.Optimism/SimulateOptimismTransactionProcessor.cs
Scooletz: I think it's opinions. If you considered the flag, and don't like this approach, let's go with what is here.  Yes, the derivation structure in the simulation is a bit prohibitive to easily decorate it a bit more.

src/Nethermind/Nethermind.Facade/Simulate/SimulateBridgeHelper.cs
```diff
@@ -185,7 +185,7 @@ private static BlockHeader GetParent(BlockHeader parent, SimulatePayload<Transac
         ulong lastKnown = (ulong)latestBlockNumber;
         if (firstBlock?.BlockOverrides?.Number > 0 && firstBlock.BlockOverrides?.Number < lastKnown)
         {
-            Block? searchResult = blockTree.FindBlock((long)firstBlock.BlockOverrides.Number);
+            Block? searchResult = blockTree.FindBlock((long)firstBlock.BlockOverrides.Number - 1);
```
emlautarom1, line 188: This is the main issue. Due to a off-by-one error we were not picking the parent but rather the child of the parent. This only happens when we don't use `"latest"` as the block parameter.

LukaszRozmej (commented):
Overall looks fine, but maybe something like `ISimulateTransactionProcessorFactory` should come from DI and n Optimism Module?

src/Nethermind/Nethermind.Facade/Simulate/SimulateReadOnlyBlocksProcessingEnv.cs
```diff
@@ -73,6 +74,7 @@ public SimulateReadOnlyBlocksProcessingEnv(
         CodeInfoRepository = new OverridableCodeInfoRepository(new CodeInfoRepository());
         SimulateVirtualMachine virtualMachine = new SimulateVirtualMachine(new VirtualMachine(blockhashProvider, specProvider, CodeInfoRepository, logManager));
         _transactionProcessor = new SimulateTransactionProcessor(SpecProvider, StateProvider, virtualMachine, CodeInfoRepository, _logManager, validate);
```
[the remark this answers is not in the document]
emlautarom1: Weird that the IDE did not detect one. Removed

src/Nethermind/Nethermind.Optimism/OptimismReceiptMessageDecoder.cs
```diff
@@ -95,21 +93,23 @@ public OptimismTxReceipt Decode(RlpStream rlpStream, RlpBehaviors rlpBehaviors =
...
         }
 
-        if (item.TxType == TxType.DepositTx && item.DepositNonce is not null &&
-            (item.DepositReceiptVersion is not null || !_isEncodedForTrie))
+        if (item.TxType == TxType.DepositTx && item is OptimismTxReceipt opItem)
         {
-            contentLength += Rlp.LengthOf(item.DepositNonce);
-
-            if (item.DepositReceiptVersion is not null)
+            if (opItem.DepositNonce is not null && (opItem.DepositReceiptVersion is not null || !isEncodedForTrie))
```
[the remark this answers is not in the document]
emlautarom1: Not sure, this was mostly an automated refactor. The goal was to change the signature of the method to accept `TxReceipt` instead of `OptimismTxReceipt`. It's no clear to me what is the `isEncodedForTrie` flag for.

deffrian (approved):

emlautarom1:
@LukaszRozmej I'm not sure I'm following. You mean that the following:

https://github.com/NethermindEth/nethermind/blob/28cb76b7ffdd7b8b508f7a1f686549909e6c51b2/src/Nethermind/Nethermind.Api/NethermindApi.cs#L254

should come as part of the `dependencies` parameter used in the constructor?

Also, to initialize `SimulateOptimismTransactionProcessorFactory` we have to do it here

https://github.com/NethermindEth/nethermind/blob/28cb76b7ffdd7b8b508f7a1f686549909e6c51b2/src/Nethermind/Nethermind.Optimism/InitializeBlockchainOptimism.cs#L42

otherwise dependencies are not yet initialized (`SpecHelper` and `L1CostHelper`).

LukaszRozmej:
> @LukaszRozmej I'm not sure I'm following. You mean that the following:
> 
> https://github.com/NethermindEth/nethermind/blob/28cb76b7ffdd7b8b508f7a1f686549909e6c51b2/src/Nethermind/Nethermind.Api/NethermindApi.cs#L254
> 
> should come as part of the `dependencies` parameter used in the constructor?
> 
> Also, to initialize `SimulateOptimismTransactionProcessorFactory` we have to do it here
> 
> https://github.com/NethermindEth/nethermind/blob/28cb76b7ffdd7b8b508f7a1f686549909e6c51b2/src/Nethermind/Nethermind.Optimism/InitializeBlockchainOptimism.cs#L42
> 
> otherwise dependencies are not yet initialized (`SpecHelper` and `L1CostHelper`).

Ignore it for now

PR #8426, opened 2025-03-27
Peers ignore self and recent ips

### LINKED ISSUE #2157: Excessive peering

benaadams:
Resolves #2157

Simpler https://github.com/NethermindEth/nethermind/pull/7026

## Changes

- Disfavour nodes with same ip addresses of nodes tried recently
- Don't connect to self even with external ip

#### Requires testing

- [x] Yes

- [x] Yes - changed existing to use different ips

### DIFF
--- a/src/Nethermind/Nethermind.Network.Discovery/DiscoveryApp.cs
+++ b/src/Nethermind/Nethermind.Network.Discovery/DiscoveryApp.cs
@@ -240,6 +240,12 @@ public class DiscoveryApp : IDiscoveryApp
                 break;
             }
 
+            if (!_discoveryManager.NodesFilter.Set(networkNode.HostIp))
+            {
+                // Already seen this node ip recently
+                continue;
+            }
+
             Node node;
             try
             {
--- a/src/Nethermind/Nethermind.Network.Discovery/DiscoveryManager.cs
+++ b/src/Nethermind/Nethermind.Network.Discovery/DiscoveryManager.cs
@@ -52,7 +52,7 @@ public class DiscoveryManager : IDiscoveryManager
         _createNodeLifecycleManager = GetLifecycleManagerFunc(isPersisted: false);
         _createNodeLifecycleManagerPersisted = GetLifecycleManagerFunc(isPersisted: true);
 
-        NodesFilter = new(networkConfig?.MaxActivePeers ?? 256);
+        NodesFilter = new((networkConfig?.MaxActivePeers * 4) ?? 200);
     }
 
     public NodeRecord SelfNodeRecord => _nodeLifecycleManagerFactory.SelfNodeRecord;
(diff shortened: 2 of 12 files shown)


src/Nethermind/Nethermind.Network.Discovery/DiscoveryManager.cs
```diff
@@ -24,6 +26,8 @@ public class DiscoveryManager : IDiscoveryManager
     private readonly ConcurrentDictionary<Hash256, INodeLifecycleManager> _nodeLifecycleManagers = new();
     private readonly INodeTable _nodeTable;
     private readonly INetworkStorage _discoveryStorage;
+    // Size not too large that don't retry; but not too small that we retry too often
+    public ClockKeyCache<IDiscoveryManager.IpAddressAsKey> NodesFilter { get; } = new(256);
```
LukaszRozmej, line 30: maybe use MaxPeers * 4 or something?

src/Nethermind/Nethermind.Network.Discovery/DiscoveryApp.cs
```diff
@@ -240,6 +240,12 @@ private void AddPersistedNodes(CancellationToken cancellationToken)
                 break;
             }
 
+            if (!_discoveryManager.NodesFilter.Set(networkNode.HostIp))
```
LukaszRozmej, line 243: Why not use `Enode` instead of `Ip`?

benaadams: Enode is unique; don't want to repeatedly connect to the same ip/node just on different ports; rather than trying a different node entirely?

LukaszRozmej: But those would be different nodes, right? Also can there be multiple nodes on multiple machines under same public IP?

Maybe argument about geographical diversity of connected nodes would be more compelling, but not sure.

src/Nethermind/Nethermind.Network.Discovery/DiscoveryManager.cs
benaadams: Changed

src/Nethermind/Nethermind.Network.Discovery/DiscoveryApp.cs
benaadams: Potentially same node, different public key or port, but same bandwidth. Better to connect to a different node with different bandwidth when syncing rather than all going to the same ip?

flcl42 (approved):

PR #8438, opened 2025-03-28
Lookup TxDecoder once per block

benaadams:
## Changes

- Only lookup the TxDecoder once per block, rather than for each tx
- Optimize ExecutionPayloadParams.ValidateParams
- Devirtualize shared array pool

Before
image

After
image

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Trie/TrackingCappedArrayPool.cs
+++ b/src/Nethermind/Nethermind.Trie/TrackingCappedArrayPool.cs
@@ -13,10 +13,10 @@ namespace Nethermind.Trie;
 /// <summary>
 /// Track every rented CappedArray<byte> and return them all at once
 /// </summary>
-public class TrackingCappedArrayPool : ICappedArrayPool, IDisposable
+public sealed class TrackingCappedArrayPool : ICappedArrayPool, IDisposable
 {
     private readonly List<byte[]> _rentedBuffers;
-    private readonly ArrayPool<byte> _arrayPool;
+    private readonly ArrayPool<byte>? _arrayPool;
 
     public TrackingCappedArrayPool() : this(0)
     {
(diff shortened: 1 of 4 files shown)


src/Nethermind/Nethermind.Trie/TrackingCappedArrayPool.cs
```diff
@@ -25,7 +25,7 @@ public TrackingCappedArrayPool() : this(0)
     public TrackingCappedArrayPool(int initialCapacity, ArrayPool<byte> arrayPool = null)
     {
         _rentedBuffers = new List<byte[]>(initialCapacity);
-        _arrayPool = arrayPool ?? ArrayPool<byte>.Shared;
+        _arrayPool = arrayPool;
```
LukaszRozmej, line 28: simpler to just do this here and remove other conditionals:
```suggestion
        _arrayPool = arrayPool ?? ArrayPool<byte>.Shared;
```

src/Nethermind/Nethermind.Trie/TrackingCappedArrayPool.cs
```diff
@@ -13,10 +13,10 @@ namespace Nethermind.Trie;
 /// <summary>
 /// Track every rented CappedArray<byte> and return them all at once
 /// </summary>
-public class TrackingCappedArrayPool : ICappedArrayPool, IDisposable
+public sealed class TrackingCappedArrayPool : ICappedArrayPool, IDisposable
 {
     private readonly List<byte[]> _rentedBuffers;
-    private readonly ArrayPool<byte> _arrayPool;
+    private readonly ArrayPool<byte>? _arrayPool;
```
LukaszRozmej, line 19:
```suggestion
    private readonly ArrayPool<byte> _arrayPool;
```

LukaszRozmej (approved):

src/Nethermind/Nethermind.Trie/TrackingCappedArrayPool.cs
benaadams: Simpler codewise, but as a class field it is hidden behind a different type and the Jit con't devirtualize its access to underlying concrete array pool and inline it

PR #8439, opened 2025-03-29
Shrink Netty page size 16MB => 8MB

benaadams:
## Changes

- We decreased peers from 200 to 50; but kept the large pool sizes

This means most of the arenas are wasted space (fill rate); don't want to go too much smaller as are used during receipts and body sync

Before

After

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Init/Steps/InitializeNetwork.cs
+++ b/src/Nethermind/Nethermind.Init/Steps/InitializeNetwork.cs
@@ -6,12 +6,10 @@ using System.Collections.Generic;
 using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
-using Autofac;
 using Nethermind.Api;
 using Nethermind.Api.Extensions;
 using Nethermind.Api.Steps;
 using Nethermind.Blockchain.Synchronization;
-using Nethermind.Blockchain.Utils;
 using Nethermind.Core;
 using Nethermind.Crypto;
 using Nethermind.Db;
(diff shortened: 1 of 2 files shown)


LukaszRozmej (approved):
@asdacap WDYT?

src/Nethermind/Nethermind.Init/Steps/InitializeNetwork.cs
```diff
@@ -29,21 +27,24 @@
...
 
 public static class NettyMemoryEstimator
 {
-    // Environment.SetEnvironmentVariable("io.netty.allocator.pageSize", "8192");
     private const uint PageSize = 8192;
 
+    public static void SetPageSize()
+    {
+        // For some reason needs to be half page size to get page size
+        Environment.SetEnvironmentVariable("io.netty.allocator.pageSize", (PageSize / 2).ToString((IFormatProvider?)null));
```
Scooletz, line 45: The reason is the default size set in the netty to 8K https://github.com/Azure/DotNetty/blob/78c5757062a9eb62e2aab2bbb35d3983710f9651/src/DotNetty.Buffers/PooledByteBufferAllocator.cs#L38 and also 4K is the smallest that you can set https://github.com/Azure/DotNetty/blob/78c5757062a9eb62e2aab2bbb35d3983710f9651/src/DotNetty.Buffers/PooledByteBufferAllocator.cs#L33

Why not to set PageSize to 4k literally above.

asdacap:
Hello, please double check that there is no increase in LOH allocation during old bodies.

PR #8456, opened 2025-04-02
Differentiate Pruning metrics

benaadams:
## Changes

- Differentiate between Persisted and Transient nodes in metrics

#### Requires testing

- [x] No


src/Nethermind/Nethermind.Trie/Pruning/TrieStoreDirtyNodesCache.cs
```diff
@@ -324,15 +324,22 @@ Hash256 GenerateKeccak(in Key key, TrieNode node)
             return keccak;
         }
 
-        void RemoveNodeFromCache(in Key key, TrieNode node)
+        void RemoveNodeFromCache(in Key key, TrieNode node, bool isPersisted)
```
LukaszRozmej, line 327: better to pass `ref` to metrics field?

asdacap (approved):

PR #8461, opened 2025-04-02
Deserialize JSON of a decimal number > MAX_UINT64

flcl42:
## Changes

- Handle cases of big integers in form of decimal number passed in json

Example:

```
{"jsonrpc":"2.0","id":2,"method":"taikoAuth_txPoolContentWithMinTip","params": ["0x0000000000000000000000000000000000000000", 

15974323677659503371917880, 

10000, 10000, [], 1, 1 ]}
```

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Serialization.Json/UInt256Converter.cs
+++ b/src/Nethermind/Nethermind.Serialization.Json/UInt256Converter.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Buffers;
+using System.Buffers.Text;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
@@ -29,12 +30,13 @@ public class UInt256Converter : JsonConverter<UInt256>
     {
         if (reader.TokenType == JsonTokenType.Number)
         {
-            if (reader.TryGetUInt64(out ulong smallValue))
+            ReadOnlySpan<byte> bytes = reader.HasValueSequence ? reader.ValueSequence.ToArray() : reader.ValueSpan;
+
+            if (Utf8Parser.TryParse(bytes, out ulong shortValue, out int bytesConsumed) && bytes.Length == bytesConsumed)
             {
-                return smallValue;
+                return new UInt256(shortValue);
             }
 
-            ReadOnlySpan<byte> bytes = reader.HasValueSequence ? reader.ValueSequence.ToArray() : reader.ValueSpan;
             Span<char> chars = stackalloc char[bytes.Length + 1];
             Encoding.UTF8.GetChars(bytes, chars);
 
--- a/src/Nethermind/Nethermind.Serialization.Json/UInt256Converter.cs
+++ b/src/Nethermind/Nethermind.Serialization.Json/UInt256Converter.cs
@@ -12,7 +12,6 @@ using System.Runtime.CompilerServices;
 using System.Text;
 using System.Text.Json;
 using System.Text.Json.Serialization;
-
 using Nethermind.Core.Extensions;
 using Nethermind.Int256;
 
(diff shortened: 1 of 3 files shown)


benaadams:
Can be an issue here as Json should only technically be able to safely represent integers up to 9007199254740991; then it would be float rounded

> The Number.MAX_SAFE_INTEGER static data property represents the maximum safe integer in JavaScript (2^53 – 1).

smartprogrammer93:
> Can be an issue here as Json should only technically be able to safely represent integers up to 9007199254740991; then it would be float rounded
> 
> > The Number.MAX_SAFE_INTEGER static data property represents the maximum safe integer in JavaScript (2^53 – 1).

@benaadams why is that. I can understand that this is a deviation from the standard, but why cant it safely be transmitted that way and then decentralized with a deviation from the standard?

smartprogrammer93:
Javascript support is not really important here at all

src/Nethermind/Nethermind.Serialization.Json/UInt256Converter.cs
```diff
@@ -29,8 +29,18 @@ private static UInt256 ReadInternal(ref Utf8JsonReader reader, JsonTokenType all
...
         {
-            return reader.GetUInt64();
+            if (reader.TryGetUInt64(out ulong smallValue))
+            {
+                return smallValue;
+            }
+
+            ReadOnlySpan<byte> bytes = reader.HasValueSequence ? reader.ValueSequence.ToArray() : reader.ValueSpan;
+            Span<char> chars = stackalloc char[bytes.Length + 1];
+            Encoding.UTF8.GetChars(bytes, chars);
```
LukaszRozmej, line 39: `Bytes.FromUtf8HexString` ?

benaadams:
> txPoolContentWithMinTip

Any reason its not being passed as quoted hex which is standard for all large numbers?

smartprogrammer93:
> > txPoolContentWithMinTip
> 
> Any reason its not being passed as quoted hex which is standard for all large numbers?

Taiko did it this way

src/Nethermind/Nethermind.Serialization.Json/UInt256Converter.cs
```diff
@@ -29,8 +29,18 @@ private static UInt256 ReadInternal(ref Utf8JsonReader reader, JsonTokenType all
...
         if (reader.TokenType == JsonTokenType.Number)
         {
-            return reader.GetUInt64();
+            if (reader.TryGetUInt64(out ulong smallValue))
+            {
+                return smallValue;
+            }
+
+            ReadOnlySpan<byte> bytes = reader.HasValueSequence ? reader.ValueSequence.ToArray() : reader.ValueSpan;
+            Span<char> chars = stackalloc char[bytes.Length + 1];
```
Scooletz, line 38: This is Uint256, or can it be something bigger? If Uint256, should we have a fixed stackalloc here maybe? Can be a bit faster.

src/Nethermind/Nethermind.Serialization.Json/UInt256Converter.cs
```diff
@@ -29,8 +29,18 @@ private static UInt256 ReadInternal(ref Utf8JsonReader reader, JsonTokenType all
     {
         if (reader.TokenType == JsonTokenType.Number)
         {
-            return reader.GetUInt64();
+            if (reader.TryGetUInt64(out ulong smallValue))
+            {
+                return smallValue;
+            }
+
+            ReadOnlySpan<byte> bytes = reader.HasValueSequence ? reader.ValueSequence.ToArray() : reader.ValueSpan;
```
Scooletz, line 37: `reader.ValueSequence.ToArray()` probably won't happen to often and might not be the biggest cost, but always there's a way to `stackalloc byte[expected]` and copying it there.

src/Nethermind/Nethermind.Serialization.Json/UInt256Converter.cs
```diff
@@ -29,8 +29,18 @@ private static UInt256 ReadInternal(ref Utf8JsonReader reader, JsonTokenType all
     {
         if (reader.TokenType == JsonTokenType.Number)
         {
-            return reader.GetUInt64();
+            if (reader.TryGetUInt64(out ulong smallValue))
```
Scooletz, line 32: If this fails, and the request was chunked, TryGetUint64 will materialize the sequence and then it will be materialized one more time below. I'd either add a length condition here (the lenght of encoded uint64 max value) to not parse through or materialize the `bytes` first and then try parse first Uint64 and then Uint256 if this fails.

flcl42:
> Can be an issue here as Json should only technically be able to safely represent integers up to 9007199254740991; then it would be float rounded
> 
> > The Number.MAX_SAFE_INTEGER static data property represents the maximum safe integer in JavaScript (2^53 – 1).

JSON standard states the numbers can be of any precision, only constrained by implementation. Standard library handles up to max ulong, I'm trying to cover uint256 range additionally.

src/Nethermind/Nethermind.Serialization.Json/UInt256Converter.cs
flcl42: It's not hex, and it also outputs to a bytes span, but we need chars for `UInt256.Parse`

src/Nethermind/Nethermind.Serialization.Json/UInt256Converter.cs
flcl42: Length is not a good test, because decimal numbers of the same length can be more or less than max long. But as for reallocating, looks good after extracting it from standard methods, adding it

src/Nethermind/Nethermind.Serialization.Json/UInt256Converter.cs
```diff
@@ -29,8 +30,19 @@ private static UInt256 ReadInternal(ref Utf8JsonReader reader, JsonTokenType all
...
         {
-            return reader.GetUInt64();
+            ReadOnlySpan<byte> bytes = reader.HasValueSequence ? reader.ValueSequence.ToArray() : reader.ValueSpan;
+
+            if (Utf8Parser.TryParse(bytes, out ulong shortValue, out int bytesConsumed) && bytes.Length == bytesConsumed)
+            {
+                return new UInt256(shortValue);
+            }
+
+            Span<char> chars = stackalloc char[bytes.Length + 1];
```
benaadams, line 40: Unbounded stack allocation from user input; means some could take down the server by adding very long numeric values to UInt256 fields

benaadams (changes requested):

src/Nethermind/Nethermind.Serialization.Json/UInt256Converter.cs
benaadams: Also if variable length, move to different method as will make the parsing of hex UInt256 horrible even when this isn't reached

flcl42: Yep, let's make it safer. Not sure about separate method, like something like in the latest commit?

benaadams (approved):

Scooletz (approved):

PR #8464, opened 2025-04-03
Log index

alexb5dh:
**[WIP]**


src/Nethermind/Nethermind.Db/LogIndex/ILogIndexStorage.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace Nethermind.Db;
+
+public readonly record struct BlockReceipts(int BlockNumber, TxReceipt[] Receipts);
+
+public interface ILogIndexStorage : IAsyncDisposable
+{
+    int GetLastKnownBlockNumber();
+    IEnumerable<int> GetBlockNumbersFor(Address address, int from, int to);
+
+    IEnumerable<int> GetBlockNumbersFor(Hash256 topic, int from, int to);
```
Scooletz, line 19: Topic does not include the index. What if I'm searching for transfer from or to an account 
`Transfer (index_topic_1 address src, index_topic_2 address dst, uint256 wad)`. Will I get the same data and potentially some false positives?

alexb5dh: Yes, false-positives are possible.

alexb5dh: - For randomly generated 20k blocks - `175.48 MB` vs `206.92 MB` (`+18%`).
- For 1M blocks (19M to 20M) from Mainnet - `1.71 GB` vs `1.92 GB` (`+12%`)

Very rough estimate would be `5 - 10%` size increase for the whole Mainnet.

alexb5dh:
> Is it possible to prove correctness / completeness of eth_getLogs response with this index? Or is the goal to simply answer eth_getLogs more efficiently?

@etan-status only the latter, ATM at least.

src/Nethermind/Nethermind.Blockchain/Receipts/IReceiptStorage.cs
```diff
@@ -19,8 +19,13 @@ public interface IReceiptStorage : IReceiptFinder
...
         /// <summary>
-        /// Receipts for a block are inserted
+        /// Receipts for a new main block are inserted
         /// </summary>
         event EventHandler<BlockReplacementEventArgs> ReceiptsInserted;
+
+        /// <summary>
+        /// Any receipts are inserted
+        /// </summary>
+        event EventHandler<ReceiptsEventArgs> AnyReceiptsInserted;
```
alexb5dh: Naming is open for discussion.

Problem is that method above (`ReceiptsInserted`) is only invoked for the new head block (regular sync or reorg), while this one is actually triggered for any new receipts, including backward sync. Also, args are different - as needed by index.

src/Nethermind/Nethermind.Core/IWriteBatch.cs
```diff
@@ -5,5 +5,9 @@
 
 namespace Nethermind.Core
 {
-    public interface IWriteBatch : IDisposable, IWriteOnlyKeyValueStore;
+    public interface IWriteBatch : IDisposable, IWriteOnlyKeyValueStore
+    {
+        void Merge(ReadOnlySpan<byte> key, ReadOnlySpan<byte> value, WriteFlags flags = WriteFlags.None);
+        void Clear();
```
alexb5dh, line 10: `Clear` is needed to reset data in case of an error in the middle of the transaction, when no data should actually be written. Current version with write on `Dispose` will submit it regardless.

I would prefer to make the batch follow explicit-commit approach, but that would affect a lot of code.

src/Nethermind/Nethermind.Core/ProgressLogger.cs
```diff
@@ -160,7 +160,8 @@ public void SetFormat(Func<ProgressLogger, string> formatter)
             _formatter = formatter;
         }
 
-        public void LogProgress()
+        // TODO: should it actually ever be true?
+        public void LogProgress(bool resetCompletion = true)
```
alexb5dh: Currently, call to `LogProgress` will mark completed `ProgressLogger` as uncompleted. Looks like a bug, but I made fix hidden behind a parameter for now - in case this is intended.

src/Nethermind/Nethermind.Runner/Nethermind.Runner.csproj
```diff
@@ -32,6 +32,7 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
     <PackageReference Include="Microsoft.VisualStudio.Azure.Containers.Tools.Targets" />
+    <PackageReference Include="Nethermind.TurboPForBindings" />
```
alexb5dh: This is a temporary workaround for [this bug](https://github.com/NethermindEth/nethermind/actions/runs/18462738401/job/52597723980?pr=8464) and should be removed before merge.

src/Nethermind/Nethermind.Db.Rocks/ColumnDb.cs
```diff
@@ -130,7 +145,7 @@ public bool KeyExists(ReadOnlySpan<byte> key)
 
     public void Flush(bool onlyWal)
     {
-        _mainDb.Flush(onlyWal);
+        _mainDb.FlushWithColumnFamily(_columnFamily);
```
alexb5dh: `onlyWal` will be ignored here but looks liks it's not supported by CF flush.

src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
```diff
@@ -66,6 +66,11 @@ public partial class DbOnTheRocks : IDb, ITunableDb, IReadOnlyNativeKeyValueStor
 
     private readonly DbSettings _settings;
 
+    // ReSharper disable once CollectionNeverQueried.Local
+    // Need to keep options from GC in case of merge operator applied, as they are used in callback
+    // TODO: find better way?
+    private readonly ConcurrentBag<OptionsHandle> _doNotGcOptions = [];
```
alexb5dh: Open for discussion.

Need to preserve created `OptionsHandle` from GC as they are used for native merge callbacks.

src/Nethermind/Nethermind.Facade/Find/LogIndexBuilder.cs
```diff
@@ -0,0 +1,521 @@
...
+using Nethermind.Blockchain.Synchronization;
+using Nethermind.Core;
+using Nethermind.Db;
+using Nethermind.Logging;
+using Timer = System.Timers.Timer;
+using static System.Threading.Tasks.TaskCreationOptions;
+
+namespace Nethermind.Facade.Find;
+
+// TODO: reduce periodic logging
```
alexb5dh, line 22: Index builder logs progress update 30 seconds:
- this is logged on a chain-following node:
  Image
- and similar 2 entries while backward index building is in progress:
  Image

All or some of these entries can be removed or made less frequent.

src/Nethermind/Nethermind.Db/LogIndex/LogIndexStorage.cs
```diff
@@ -0,0 +1,1044 @@
...
+using System.Threading.Tasks;
+using Nethermind.Core;
+using Nethermind.Core.Collections;
+using Nethermind.Core.Crypto;
+using Nethermind.Core.Extensions;
+using Nethermind.Logging;
+
+namespace Nethermind.Db
+{
+    // TODO: test on big-endian system
```
alexb5dh: Needs independent testing on:
- Any big-endian CPU
- MacOS
- Old x86 where AVX2 is not supported

src/Nethermind/Nethermind.Db/RocksDbSettings.cs
```diff
@@ -17,6 +20,9 @@ public DbSettings(string name, string path)
         public bool DeleteOnStart { get; set; }
         public bool CanDeleteFolder { get; set; } = true;
 
+        public IMergeOperator? MergeOperator { get; set; }
+        public Dictionary<string, IMergeOperator>? MergeOperatorByColumnFamily { get; set; }
```
alexb5dh: Option to provide custom merge operator per column family. Naming and location is open for discussion.

src/Nethermind/Nethermind.Facade/Find/LogIndexStorageRpcExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+    {
+        (int from, int to) = ((int)fromBlock, (int)toBlock);
+
+        List<int>? addressNumbers = null;
+        if (filter.AddressFilter.Address is { } address)
+            addressNumbers = storage.GetBlockNumbersFor(address, from, to);
+        else if (filter.AddressFilter.Addresses is { Count: > 0 } addresses)
+            addressNumbers = AscListHelper.UnionAll(addresses.Select(a => storage.GetBlockNumbersFor(a, from, to)));
+
+        // TODO: consider passing storage directly to keep abstractions
```
alexb5dh: Updates to topic-related interfaces can potentially be simplified to use just a single method accepting `ILogIndexStorage` parameter. This will make interfaces simpler, but storage dependency will look out-of-place.

nuget.config
```diff
@@ -3,14 +3,14 @@
   <packageSources>
     <clear />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
-    <!--<add key="nugettest.org" value="https://apiint.nugettest.org/v3/index.json" />-->
+    <add key="nugettest.org" value="https://apiint.nugettest.org/v3/index.json" /> <!-- TODO: comment before merging! -->
```
alexb5dh: Temporary measure while PR is being tested on different machines.
Package will be published on Nuget after and this needs to be removed before merge.

src/Nethermind/Nethermind.Facade/Find/LogIndexBuilder.cs
alexb5dh: Reduced a bit - now 2-3 entries are logged every 30 seconds:
image

src/Nethermind/Nethermind.Core/ProgressLogger.cs
asdacap: Move whatever you can to another branch so that this PR is smaller.

asdacap:
Can you describe how it works in the description of the PR?

src/Nethermind/Nethermind.Core/ProgressLogger.cs
alexb5dh: https://github.com/NethermindEth/nethermind/pull/9481

asdacap:
What is the maximum size of the value (in bytes)

src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.cs
```diff
@@ -2415,7 +2415,7 @@ private async Task RaiseBlockAddedToMainAndWaitForNewHead(Block block, Block pre
                 : new BlockReplacementEventArgs(block ?? Build.A.Block.TestObject, previousBlock);
 
             Task waitTask = Wait.ForEventCondition<Block>(
-                default,
+                CancellationToken.None,
```
flcl42: They are the same, no need to change

alexb5dh:
> What is the maximum size of the value (in bytes)

@asdacap, no limit in general, but background compression tries to keep it under  `LogIndexConfig.CompressionDistance`, by default `128 * 4` bytes.

src/Nethermind/Nethermind.Db/LogIndex/Compactor.cs
```diff
@@ -0,0 +1,154 @@
...
+using Nethermind.Logging;
+
+namespace Nethermind.Db.LogIndex;
+
+partial class LogIndexStorage
+{
+    /// <summary>
+    /// Periodically forces background log index compaction for every N added blocks.
+    /// </summary>
+    private interface ICompactor
```
alexb5dh: Keeps disk usage a bit more reasonable during index building.
Can probably be replaced with some DB configuration tweaks and will most probably be removed in one of the subsequent PRs. As of now, space difference between having it enabled and not can be 50% and more.

src/Nethermind/Nethermind.Db/LogIndex/Compactor.cs
```diff
@@ -0,0 +1,154 @@
...
+    {
+        private int? _lastAtMin;
+        private int? _lastAtMax;
+
+        private CompactingStats _stats = new();
+        private readonly LogIndexStorage _storage;
+        private readonly ILogger _logger;
+        private readonly int _compactionDistance;
+
+        // TODO: simplify concurrency handling
```
alexb5dh: Compaction runs in background. Need to:
- ensure no concurrent compaction calls;
- have an option to force another compaction at any point in time (running after current if any).

Can probably do this with less number of primitives, but not sure how.

src/Nethermind/Nethermind.Db/LogIndex/LogIndexStorage.cs
```diff
@@ -0,0 +1,948 @@
...
+        {
+            IDb db = GetDb(index);
+            ISortedKeyValueStore? sortedDb = db as ISortedKeyValueStore
+                ?? throw new NotSupportedException($"{db.GetType().Name} DB does not support sorted lookups.");
+
+            return new LogIndexEnumerator(this, sortedDb, key, from, to);
+        }
+
+        // TODO: discuss potential optimizations
+        public LogIndexAggregate Aggregate(IReadOnlyList<BlockReceipts> batch, bool isBackwardSync, LogIndexUpdateStats? stats)
```
alexb5dh, line 544: This may need some optimization (in terms of both time and memory usage). Aggregates batchs of blocks with receipts to a dictionary, mapping each address/topic to an ordered list of blocks it occurs in.

Some batches with tons of spam transactions can make resulting dictionaries reach over 600k keys and aggregation to take several seconds.

From basic tracing looks like main time consumer is comparison of `Hash256` keys (not so much hashing, but an exact equality check). So replacing hashing function with something very basic, like comparing first 4-8 bytes as integer/long, doesn't help.

src/Nethermind/Nethermind.Db/LogIndex/LogIndexStorage.cs
```diff
@@ -0,0 +1,946 @@
...
+using System.Threading.Tasks;
+using Nethermind.Core;
+using Nethermind.Core.Collections;
+using Nethermind.Core.Crypto;
+using Nethermind.Core.Extensions;
+using Nethermind.Logging;
+
+namespace Nethermind.Db.LogIndex
+{
+    // TODO: test on big-endian system or optimize for little-endian
```
alexb5dh: Need to decide whether big-endian systems should be supported.
- If yes, this needs to be tested on such a machine.
- If no, I can optimize a few methods to use `MemoryMarshal` insteady of a loop. Would prefer to avoid doing if's with an endianness check.

src/Nethermind/Nethermind.Blockchain.Test/Find/LogFinderTests.cs
```diff
@@ -323,7 +329,7 @@ public async Task Throw_log_finder_operation_canceled_after_given_timeout([Value
         CancellationToken cancellationToken = cancellationTokenSource.Token;
 
         StoreTreeBlooms(true);
-        _logFinder = new LogFinder(_blockTree, _receiptStorage, _receiptStorage, _bloomStorage, LimboLogs.Instance, _receiptsRecovery);
+        _logFinder = CreateLogFinder();
```
flcl42, line 323: Do we need to create it if it's done in SetUp?

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/Filter.cs
```diff
@@ -65,6 +67,16 @@ public void ReadJson(JsonElement filter, JsonSerializerOptions options)
             {
                 Topics = GetTopics(topicsElement, options);
             }
+
+            if (filter.TryGetProperty("useIndex"u8, out JsonElement useIndex))
```
flcl42, line 71: Seems a bit redundant, why would it be used if it was applied?

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
```diff
@@ -633,6 +635,10 @@ public ResultWrapper<IEnumerable<FilterLog>> eth_getLogs(Filter filter)
         {
             LogFilter logFilter = _blockchainBridge.GetFilter(fromBlock, toBlock, filter.Address, filter.Topics);
 
+            // ReSharper disable once ConditionIsAlwaysTrueOrFalse - can be null in tests
+            if (logFilter is not null && filter is not null)
+                logFilter.UseIndex = filter.UseIndex;
```
flcl42, line 652: It's better to use {} around the code in if branches(and below)

src/Nethermind/Nethermind.Db/LogIndex/AverageStats.cs
```diff
@@ -0,0 +1,31 @@
...
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System.Threading;
+
+namespace Nethermind.Db.LogIndex;
+
+/// <summary>
+/// Aggregates average of multiple incrementally-added values.
+/// </summary>
+public class AverageStats
```
flcl42, line 11: Can all this statistics become standard grafana metrics instead? Not insisting at all, can be another pr if grafana approach sounds nicer

src/Nethermind/Nethermind.Facade/Find/LogIndexBuilder.cs
```diff
@@ -0,0 +1,492 @@
...
+            _aggregateBlock = aggregateBlock;
+            _addReceiptsBlock = addReceiptsBlock;
+        }
+    }
+
+    private readonly IBlockTree _blockTree;
+    private readonly ISyncConfig _syncConfig;
+    private readonly ILogger _logger;
+
+    private readonly CancellationTokenSource _cancellationSource = new();
```
flcl42, line 51: Should we dispose it?

src/Nethermind/Nethermind.Init/Steps/StartLogIndex.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    [RunnerStepDependencies(typeof(InitDatabase), typeof(StartBlockProcessor))]
+    public class StartLogIndex(IBasicApi api, IServiceStopper serviceStopper, ILogIndexBuilder logIndexBuilder) : IStep
+    {
+        public Task Execute(CancellationToken cancellationToken)
+        {
+            if (api.Config<ILogIndexConfig>().Enabled)
+            {
+                _ = logIndexBuilder.StartAsync();
+                serviceStopper.AddStoppable(logIndexBuilder);
```
asdacap: `IStoppableService` and `IDisposable` automatically detected by the container so no need to explicitly add it if `ILogIndexBuilder` is also created via container.

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/Filter.cs
```diff
@@ -21,6 +21,8 @@ public class Filter : IJsonRpcParam
 
     public IEnumerable<Hash256[]?>? Topics { get; set; }
 
+    public bool UseIndex { get; set; } = true;
```
asdacap, line 24: Is it meant to be selectable via RPC?

src/Nethermind/Nethermind.Facade/Find/LogFinder.cs
```diff
@@ -29,42 +30,50 @@ public class LogFinder : ILogFinder
         private readonly IReceiptsRecovery _receiptsRecovery;
         private readonly int _maxBlockDepth;
         private readonly int _rpcConfigGetLogsThreads;
+        private readonly int _minBlocksToUseIndex;
         private readonly IBlockFinder _blockFinder;
         private readonly ILogger _logger;
+        private readonly ILogIndexStorage _logIndexStorage;
 
         public LogFinder(IBlockFinder? blockFinder,
```
asdacap: NP: Perhaps make a decorator to `ILogFinder`

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/Filter.cs
alexb5dh: Do you mean why would anyone utilize `useIndex` if index is enabled via config? I would say it's a temporary measure while index is still in "beta" stage:
- we can utilize it for testing of some kind, for example compare requests with and without this flag;
- if there's in issue with `eth_getLogs` with enabled index, changing this flag can show if the problem is with the index itself or something else.

But I won't argue much against removing it :)

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
alexb5dh: It's subjective. I'm ok adding braces when there are other `if`s with them in the same scope, but otherwise prefer to have a single line. Or do we have some convention on that?

src/Nethermind/Nethermind.Db/LogIndex/AverageStats.cs
alexb5dh: Yeah, Grafana defenitely sounds better. I just haven't touched it yet and it was easier for me to check stats from console logs during testing. Not sure I'll be able to change it in this PR, will check how hard this will be to move.

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/Filter.cs
alexb5dh: Yes, this is an option to "disable" log index for a single request. Explained intention [here](https://github.com/NethermindEth/nethermind/pull/8464#discussion_r2686189308).

src/Nethermind/Nethermind.Facade/Find/LogFinder.cs
alexb5dh: Decorator using `ILogIndexStorage`, which replaces default `LogFinder` only when index is enabled?

src/Nethermind/Nethermind.Blockchain.Test/Find/LogFinderTests.cs
alexb5dh: Underlying setup seems to be different in some tests.

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/Filter.cs
flcl42: Less code is better. If it really useful let it stay

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
flcl42: {} makes it more readable and safer

flcl42 (approved):

src/Nethermind/Nethermind.Db/LogIndex/Compactor.cs
asdacap: What is the database size?

src/Nethermind/Nethermind.Db/LogIndex/LogIndexEnumerator.cs
```diff
@@ -0,0 +1,152 @@
...
+                _storage.DecompressDbValue(viewValue, _value.AsSpan());
+            }
+            else
+            {
+                length = viewValue.Length / BlockNumberSize;
+                _value = new(capacity: length, count: length);
+                ReadBlockNumbers(viewValue, _value.AsSpan());
+            }
+
+            ReverseBlocksIfNeeded(_value.AsSpan());
```
asdacap, line 133: @copilot find `ReverseBlocksIfNeeded`

asdacap: Well... I tried.

src/Nethermind/Nethermind.Db/LogIndex/LogIndexStorage.cs
asdacap: Nope. Many other things will probably break.

src/Nethermind/Nethermind.Db/LogIndex/Compactor.cs
alexb5dh: Sorry, missed this one. It's ~30 GB for mainnet, as mentioned in the description.

src/Nethermind/Nethermind.Db.Test/LogIndex/LogIndexStorageTestExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+            var result = new List<int>();
+            using IEnumerator<int> enumerator = storage.GetEnumerator(index, topic, from, to);
+
+            while (enumerator.MoveNext())
+                result.Add(enumerator.Current);
+
+            return result;
+        }
+    }
+}
```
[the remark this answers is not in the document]
alexb5dh: [It is](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-14#extension-members)

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
```diff
@@ -825,4 +834,29 @@ public ResultWrapper<JsonNode> eth_config()
...
+    private void VerifyLogsResponse(IList<FilterLog> response, LogFilter filter,
+        BlockHeader from, BlockHeader to, CancellationToken cancellation)
+    {
+        filter.UseIndex = false;
+        IEnumerable<FilterLog>? expectedResponse = _blockchainBridge.GetLogs(filter, from, to, cancellation);
+
+        using IEnumerator<FilterLog> expectedEnum = expectedResponse.GetEnumerator();
+
+        var i = -1;
+        while (++i < response.Count | expectedEnum.MoveNext())
```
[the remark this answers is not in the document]
alexb5dh: This is intended

src/Nethermind/Nethermind.Facade/Find/LogFinder.cs
alexb5dh: Didn't manage to do it via a decorator, as there are internal dependencies, but moved all log-index logic to a descendant, with minimal changes here.

src/Nethermind/Nethermind.Db/LogIndex/LogIndexStorage.cs
alexb5dh: Replaced another loop with `MemoryMarshal`, probably making it big-endian-incompatible for now. At least copying storage from one system to another won't work.
@LukaszRozmej FYI

src/Nethermind/Nethermind.Db/LogIndex/LogIndexAggregate.cs
```diff
@@ -0,0 +1,33 @@
...
+/// </summary>
+public struct LogIndexAggregate(int firstBlockNum, int lastBlockNum)
+{
+    private Dictionary<Address, List<int>>? _address;
+    private Dictionary<Hash256, List<int>>[]? _topic;
+
+    public int FirstBlockNum { get; } = firstBlockNum;
+    public int LastBlockNum { get; } = lastBlockNum;
+
+    public Dictionary<Address, List<int>> Address => _address ??= new();
```
asdacap, line 23: NP: There is a fancy `field` keyword.

asdacap:
Can you add some doc on the database encoding within the PR description itself?

src/Nethermind/Nethermind.Db/LogIndex/LogIndexAggregate.cs
alexb5dh: Don't think I can use it, given that `IsEmpty` method below needs to directly access `_address`.

src/Nethermind/Nethermind.JsonRpc/Modules/LogIndex/ILogIndexRpcModule.cs
```diff
@@ -0,0 +1,19 @@
...
+[RpcModule(ModuleType.LogIndex)]
+public interface ILogIndexRpcModule : IRpcModule
+{
+    [JsonRpcMethod(Description = "Retrieves log index block number for the given filter.", IsImplemented = true, IsSharable = true)]
+    ResultWrapper<IEnumerable<long>> logIndex_blockNumbers([JsonRpcParameter] Filter filter);
+
+    [JsonRpcMethod(Description = "Retrieves log index status.", IsImplemented = true, IsSharable = true)]
+    ResultWrapper<LogIndexStatus> logIndex_status();
+
+    // TODO: add compaction RPC
```
LukaszRozmej: Are you planning on adding them or not for now?

src/Nethermind/Nethermind.Db/LogIndex/MergeOperator.cs
```diff
@@ -0,0 +1,175 @@
...
+using System.Diagnostics;
+using System.Threading;
+using Nethermind.Core;
+using Nethermind.Core.Collections;
+
+namespace Nethermind.Db.LogIndex;
+
+partial class LogIndexStorage
+{
+    // TODO: check if success=false + paranoid_checks=true is better than throwing an exception
```
LukaszRozmej, line 14: address or leave it ?

src/Nethermind/Nethermind.Db/LogIndex/ILogIndexConfig.cs
```diff
@@ -0,0 +1,106 @@
...
+
+    [ConfigItem(
+        Description = "Log index is reset on startup if enabled.",
+        DefaultValue = "false"
+    )]
+    public bool Reset { get; set; }
+
+    [ConfigItem(
+        Description = "Max allowed reorg depth for the index.",
+        DefaultValue = "64",
```
LukaszRozmej, line 25: `Reorganization.MaxDepth` as default?
Or `PruningConfig.PruningBoundary`?

LukaszRozmej: Some chains have longer reorgs we support, up to 256 for prod and even 1000 for test chains.

src/Nethermind/Nethermind.Db/LogIndex/LogIndexConfig.cs
```diff
@@ -0,0 +1,32 @@
...
+
+    public int MaxReorgDepth { get; set; } = 64;
+
+    public int MaxBatchSize { get; set; } = 256;
+    public int MaxAggregationQueueSize { get; set; } = 16;
+    public int MaxSavingQueueSize { get; set; } = 16;
+
+    public int MaxReceiptsParallelism { get; set; } = Math.Max(Environment.ProcessorCount / 2, 1);
+    public int MaxAggregationParallelism { get; set; } = Math.Max(Environment.ProcessorCount / 2, 1);
+    public int MaxCompressionParallelism { get; set; } = Math.Max(Environment.ProcessorCount / 2, 1);
```
LukaszRozmej, line 24: Can you use `RuntimeInformation.PhysicalCoreCount`?

src/Nethermind/Nethermind.Db/LogIndex/DisabledLogIndexStorage.cs
```diff
@@ -0,0 +1,46 @@
...
+        throw new NotSupportedException();
+
+    public Task AddReceiptsAsync(LogIndexAggregate aggregate, LogIndexUpdateStats? stats = null) =>
+        throw new NotSupportedException();
+
+    public Task RemoveReorgedAsync(BlockReceipts block) =>
+        throw new NotSupportedException();
+
+    public Task CompactAsync(bool flush = false, int mergeIterations = 0, LogIndexUpdateStats? stats = null) =>
+        throw new NotSupportedException();
```
LukaszRozmej: Should it throw or just don't do anything?

asdacap (approved):

src/Nethermind/Nethermind.Db/LogIndex/MergeOperator.cs
```diff
@@ -0,0 +1,175 @@
...
+                throw new LogIndexStateException($"Invalid order during merge: {lastBlock} -> {nextBlock} (backward: {isBackward}).", key);
+
+            result.AddRange(value);
+        }
+
+        // TODO: avoid array copying in case of a single value?
+        private ArrayPoolList<byte>? Merge(ReadOnlySpan<byte> key, RocksDbMergeEnumerator enumerator, bool isPartial)
+        {
+            var success = false;
+            ArrayPoolList<byte>? result = null;
```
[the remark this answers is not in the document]
alexb5dh: Dispose is conditional

src/Nethermind/Nethermind.Db/LogIndex/LogIndexStorage.cs
```diff
@@ -0,0 +1,993 @@
...
+        {
+            if (_stopped)
+                return;
+
+            await _writeSemaphores[false].WaitAsync();
+            await _writeSemaphores[true].WaitAsync();
+
+            try
+            {
+                if (_stopped)
```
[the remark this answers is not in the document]
alexb5dh: Concurrent calls are possible

src/Nethermind/Nethermind.Db.Test/LogIndex/LogIndexStorageIntegrationTests.cs
```diff
@@ -0,0 +1,999 @@
...
+using Nethermind.TurboPForBindings;
+using NUnit.Framework;
+using NUnit.Framework.Interfaces;
+using static Nethermind.Db.LogIndex.LogIndexStorage;
+
+namespace Nethermind.Db.Test.LogIndex
+{
+    // TODO: test for reorg out-of-order
+    // TODO: test for concurrent reorg and backward sync
+    // TODO: test for background job failure
```
LukaszRozmej: Adding before merging?

src/Nethermind/Nethermind.Db/LogIndex/LogIndexStorage.cs
```diff
@@ -0,0 +1,996 @@
...
+using System.Threading.Tasks;
+using Nethermind.Core;
+using Nethermind.Core.Collections;
+using Nethermind.Core.Crypto;
+using Nethermind.Core.Extensions;
+using Nethermind.Logging;
+
+namespace Nethermind.Db.LogIndex
+{
+    // TODO: use uint instead of int for block number?
```
LukaszRozmej, line 22: ugh yes we kind of should, would give us 2x the space, which with sub-second blocks might be important.
Will it be hard to do in the future? (Will we break layout?)

src/Nethermind/Nethermind.Db/LogIndex/LogIndexAggregate.cs
LukaszRozmej, line 23: If we would use PooledDictionaries and ArrayLists we could probably decrease memory preassure a bunch here?

src/Nethermind/Nethermind.Db/LogIndex/LogIndexEnumerator.cs
```diff
@@ -0,0 +1,159 @@
...
+    // TODO: pre-fetch next value?
+    /// <summary>
+    /// Enumerates block numbers from <see cref="LogIndexStorage"/> for the given <c>key</c>,
+    /// within the specified <c>from</c>/<c>to</c> range.
+    /// </summary>
+    public sealed class LogIndexEnumerator : IEnumerator<int>
+    {
+        private const int CompletedIndex = int.MinValue;
+
+        private readonly LogIndexStorage _storage;
```
LukaszRozmej: minor: reference to storage seems excessive just to call one function (decompress).

src/Nethermind/Nethermind.Facade/Find/LogFinder.cs
```diff
@@ -103,30 +103,31 @@ private static bool ShouldUseBloomDatabase(BlockHeader fromBlock, BlockHeader to
 
         private IEnumerable<FilterLog> FilterLogsWithBloomsIndex(LogFilter filter, BlockHeader fromBlock, BlockHeader toBlock, CancellationToken cancellationToken)
         {
-            BlockHeader FindBlockHeader(long blockNumber, CancellationToken token)
+            IEnumerable<long> EnumerateBlockNumbers(LogFilter f, long from, long to, CancellationToken token)
```
LukaszRozmej: token not used?

src/Nethermind/Nethermind.Db/LogIndex/MergeOperator.cs
LukaszRozmej: Leave it

src/Nethermind/Nethermind.JsonRpc/Modules/LogIndex/ILogIndexRpcModule.cs
LukaszRozmej: Addressed in #10428

src/Nethermind/Nethermind.Db/LogIndex/ILogIndexConfig.cs
alexb5dh: Where can I find this value for such chains?

src/Nethermind/Nethermind.Db/LogIndex/LogIndexConfig.cs
alexb5dh: But I'm interested in max number of parallel threads which doesn't always equal to number of physical CPUs (Hyper-Threading, etc.).

If question is about replacing `Environment.ProcessorCount / 2` with `RuntimeInformation.PhysicalCoreCount` - also not sure how reliable this will be. 

Intention here is basically to limit any background job activity to 50% of available CPU.

src/Nethermind/Nethermind.Db/LogIndex/DisabledLogIndexStorage.cs
alexb5dh: Not used ATM apart from tests, was intended for RPC, but not anymore. Can even remove it from the interface.

src/Nethermind/Nethermind.Db/LogIndex/LogIndexAggregate.cs
alexb5dh: Regarding `ArrayPoolLists` - maybe, but not sure about performance impact of pooling/returning ton of arrays. In some rare cases it can be 100k+ in a single batch.
Regarding `PooledDictionary` - not familiar with it, need to check. But it will also need to handle 100k+ keys sometimes.

src/Nethermind/Nethermind.Db/LogIndex/LogIndexEnumerator.cs
alexb5dh: Need a reference to currently selected `CompressionAlgorithm` to run decompression if key is compressed.

src/Nethermind/Nethermind.Facade/Find/LogFinder.cs
alexb5dh: Yes, cancellation now happens in `FilterLogsInBlocksParallel` (`ParallelizeWithLock`). Will remove param.

src/Nethermind/Nethermind.Db/LogIndex/LogIndexStorage.cs
alexb5dh: https://github.com/NethermindEth/nethermind/pull/10430

src/Nethermind/Nethermind.Db/LogIndex/DisabledLogIndexStorage.cs
alexb5dh: Removed

src/Nethermind/Nethermind.Db/LogIndex/LogIndexEnumerator.cs
LukaszRozmej: yes but potentially this could be done in different way?

alexb5dh: Now passing compression algorithm directly and calling static `LogIndexStorage` method. Don't want to make it agorithm extension, as it's kind-of index-specific.

PR #8474, opened 2025-04-06
Perf/More accurate request limit.

asdacap:
- To reduce timeout and high allocation, we limit the size of the bodies, receipts and snap request at protocol handler level.
- This information however is not exposed to the sync code, so sync code still try to request in a large batch, potentially causing a large portion of the request to be retried later.
- This is not a big deal for downloads that is paralelizable, bit it can be a problem for sync that requires order. 

- [X] Optimization

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- No change in performance, but helpful for #8255

### DIFF
--- a/src/Nethermind/Nethermind.Serialization.Rlp/OwnedBlockBodies.cs
+++ b/src/Nethermind/Nethermind.Serialization.Rlp/OwnedBlockBodies.cs
@@ -3,6 +3,8 @@
 
 using System;
 using System.Buffers;
+using System.Collections;
+using System.Collections.Generic;
 using Nethermind.Core.Crypto;
 using Nethermind.Serialization.Rlp;
 
@@ -13,7 +15,7 @@ namespace Nethermind.Core;
 /// BlockBody may contain `Memory<byte>` that is explicitly managed. Reusing `BlockBody` from this object after Dispose
 /// is likely to cause corrupted `BlockBody`.
 /// </summary>
-public class OwnedBlockBodies : IDisposable
+public class OwnedBlockBodies : IDisposable, IReadOnlyList<BlockBody?>
 {
     private readonly BlockBody?[]? _rawBodies = null;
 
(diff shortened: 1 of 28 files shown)


benaadams (approved):

src/Nethermind/Nethermind.Serialization.Rlp/OwnedBlockBodies.cs
```diff
@@ -67,4 +69,24 @@ public void Dispose()
...
         _memoryOwner = null;
     }
+
+    public IEnumerator<BlockBody?> GetEnumerator()
+    {
+        foreach (var blockBody in _rawBodies)
+        {
+            yield return blockBody;
+        }
+    }
```
LukaszRozmej, line 79: Why not just 
```suggestion
    public IEnumerator<BlockBody?> GetEnumerator() => _rawBodies.GetEnumerator();
```

src/Nethermind/Nethermind.Synchronization/Peers/ISyncPeerPool.cs
```diff
@@ -30,6 +32,16 @@ Task<SyncPeerAllocation> Allocate(
 
         void ReportWeakPeer(PeerInfo peerInfo, AllocationContexts allocationContexts);
 
+        /// <summary>
+        /// Estimate the request limit for a specific request type for the peer which get allocated next based
+        /// on the allocation strategy and context. May not be accurate as different peer may get allocated.
+        /// </summary>
+        Task<int?> EstimateRequestLimit(
+            RequestType bodies,
```
LukaszRozmej, line 40:
```suggestion
            RequestType requestType,
```

src/Nethermind/Nethermind.Synchronization/Peers/ISyncPeerPool.cs
```diff
@@ -30,6 +32,16 @@ Task<SyncPeerAllocation> Allocate(
...
+            IPeerAllocationStrategy peerAllocationStrategy,
+            AllocationContexts blocks,
```
LukaszRozmej, line 42: allocationContext

PR #8481, opened 2025-04-08
Skip block build if no new tx have been added

benaadams:
## Changes

- If no new txs have been added to pool since last block build; skip the improvement and wait for next delay.
- Is no reason to build a block if it will build the same as previous

- [x] Optimization

#### Requires testing

- [x] No

- [x] Yes - modified existing test

### DIFF
--- a/src/Nethermind/Nethermind.Merge.Plugin/BlockProduction/PayloadPreparationService.cs
+++ b/src/Nethermind/Nethermind.Merge.Plugin/BlockProduction/PayloadPreparationService.cs
@@ -181,12 +181,7 @@ public class PayloadPreparationService : IPayloadPreparationService, IDisposable
                 // If we reach here, we still have time for an improvement build (which still responds to cancellation)
                 try
                 {
-                    // If the dynamic delay is still positive, await that
-                    if (dynamicDelay > TimeSpan.Zero)
-                    {
-                        if (_logger.IsTrace) _logger.Trace($"Block for payload {payloadId} with parent {parentHeader.ToString(BlockHeader.Format.FullHashAndNumber)} will be improved in {dynamicDelay.TotalMilliseconds}ms");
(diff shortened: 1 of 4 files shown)


src/Nethermind/Nethermind.Merge.Plugin/BlockProduction/PayloadPreparationService.cs
```diff
@@ -159,33 +160,34 @@ private IBlockImprovementContext CreateBlockImprovementContext(string payloadId,
...
-                    if (_logger.IsTrace) _logger.Trace($"Block for payload {payloadId} with parent {parentHeader.ToString(BlockHeader.Format.FullHashAndNumber)} will be improved in {dynamicDelay.TotalMilliseconds}ms");
                     await Task.Delay(dynamicDelay, token);
                 }
-            }
-            catch (OperationCanceledException) { }
+                catch (OperationCanceledException) { }
+
+                // Loop the delay if no new txs have been added, and while not cancelled.
+                // Is no point in rebuilding an identical block.
+            } while (added == TxPool.Metrics.PendingTransactionsAdded);
```
LukaszRozmej, line 190: Should `TxPool.Metrics.PendingTransactionsAdded` be changed to a field and we should do `Volatile.Read` of it?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Merge.Plugin/BlockProduction/PayloadPreparationService.cs
benaadams: I thought about this; decided not to

But you are probably right 😄

benaadams: Done

tanishqjasoria (approved):

PR #8486, opened 2025-04-08
Cap block build size in bytes

benaadams:
## Changes

- Don't build blocks larger than 9728.KiB (Configurable)

10MB is the max L1 CL uncompress block gossip size and also includes the block headers etc https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#max_message_size
image

So we shouldn't build passed this size

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Config/BlocksConfig.cs
+++ b/src/Nethermind/Nethermind.Config/BlocksConfig.cs
@@ -1,7 +1,9 @@
 // SPDX-FileCopyrightText: 2022 Demerzel Solutions Limited
 // SPDX-License-Identifier: LGPL-3.0-only
 
+using System;
 using System.Text;
+using Nethermind.Core;
 using Nethermind.Core.Exceptions;
 using Nethermind.Core.Extensions;
 using Nethermind.Int256;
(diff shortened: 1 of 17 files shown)


src/Nethermind/Nethermind.Config/BlocksConfig.cs
```diff
@@ -58,5 +58,7 @@ public byte[] GetExtraDataBytes()
         public string GasToken { get => GasTokenTicker; set => GasTokenTicker = value; }
 
         public static string GasTokenTicker { get; set; } = "ETH";
+
+        public long BlockProductionMaxTxKilobytes { get; set; } = 9728;
```
MarekM25, line 62: are you sure it is correct value? how much CL can have on worst-case?

src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.BlockProductionTransactionPicker.cs
```diff
@@ -4,24 +4,29 @@
...
     {
         public class BlockProductionTransactionPicker : IBlockProductionTransactionPicker
         {
+            private readonly long _maxTxLengthBytes;
+
             protected readonly ISpecProvider _specProvider;
             private readonly bool _ignoreEip3607;
 
-            public BlockProductionTransactionPicker(ISpecProvider specProvider, bool ignoreEip3607 = false)
+            public BlockProductionTransactionPicker(ISpecProvider specProvider, long maxTxLengthKilobytes = 9728, bool ignoreEip3607 = false)
```
MarekM25, line 26: nullable value in in constructor parameters and set it to const value inside? shared with the config value?

src/Nethermind/Nethermind.Core/Block.cs
```diff
@@ -122,6 +122,10 @@ public Transaction[] Transactions
 
     [JsonIgnore]
     public ArrayPoolList<AddressAsKey>? AccountChanges { get; set; }
+
+    [JsonIgnore]
+    public long TxByteLength { get; internal set; }
```
MarekM25, line 127: Can it be used in BlockToProduce instead?

src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.BlockProductionTransactionPicker.cs
MarekM25: or pass config in constructor?

MarekM25 (commented):
This is a good PR, but it will still have attack vectors with higher gas-limit. Of course, next improvements can be done in a seperate PRs. Let's discuss

src/Nethermind/Nethermind.Config/BlocksConfig.cs
benaadams: Docs say 127kB in worse case; but not sure they are outdated so increased to 512kB (as also EL payload framing)

1024*10-512 = 9728

src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.BlockProductionTransactionPicker.cs
LukaszRozmej: or use same named const in both places

LukaszRozmej (commented):
use named const

src/Nethermind/Nethermind.Core/Block.cs
benaadams: Moved

src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.BlockProductionTransactionPicker.cs
benaadams: Used const

LukaszRozmej (approved):

PR #8492, opened 2025-04-08
Include version number in default extra data

benaadams:
## Changes

- Include version number in default extra data, meaning it is easier to determine if any issues are fixed issues or different
- Convert `-unsable` to `a` so as not to use up too much space
```
JSON RPC     : http://127.0.0.1:8545 ; http://localhost:8551
Genesis hash : 0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3
Gaslimit     : 36,000,000
ExtraData    : Nethermind v1.32.0a
```

#### Requires testing

- [x] No - will throw exception if wrong on startup and outputs value in startup banner

### DIFF
--- a/src/Nethermind/Nethermind.Config/BlocksConfig.cs
+++ b/src/Nethermind/Nethermind.Config/BlocksConfig.cs
@@ -1,6 +1,7 @@
 // SPDX-FileCopyrightText: 2022 Demerzel Solutions Limited
 // SPDX-License-Identifier: LGPL-3.0-only
 
+using System;
 using System.Text;
 using Nethermind.Core;
 using Nethermind.Core.Exceptions;
@@ -11,6 +12,8 @@ namespace Nethermind.Config
 {
     public class BlocksConfig : IBlocksConfig
     {
+        public static bool AddVersionToExtraData { get; set; }
+
         public const string DefaultExtraData = "Nethermind";
         private byte[] _extraDataBytes = [];
         private string _extraDataString;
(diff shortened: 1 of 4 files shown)


src/Nethermind/Nethermind.Core/ProductInfo.cs
```diff
@@ -36,7 +36,7 @@ static ProductInfo()
         if (index != -1)
         {
             Commit = Version[(index + 1)..];
-            Version = Version[..Math.Min(index + 9, Version.Length - 1)];
```
smartprogrammer93, line 39: Pretty sure there is a reason why this is done like this. I just dont know it 😅 @rubo ?

smartprogrammer93 (approved):

src/Nethermind/Nethermind.Config/BlocksConfig.cs
```diff
@@ -11,8 +12,17 @@ namespace Nethermind.Config
     public class BlocksConfig : IBlocksConfig
     {
         public const string DefaultExtraData = "Nethermind";
-        private byte[] _extraDataBytes = Encoding.UTF8.GetBytes(DefaultExtraData);
-        private string _extraDataString = DefaultExtraData;
+        private byte[] _extraDataBytes = [];
```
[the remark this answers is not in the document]
benaadams: Doesn't understand new syntax

src/Nethermind/Nethermind.Core/ProductInfo.cs
rubo: Good catch. There is. Needs to be reverted to the original.

rubo (changes requested):

src/Nethermind/Nethermind.Core/ProductInfo.cs
benaadams: Why? Original never trims off the commit

rubo: This line truncates the commit hash to 8 characters if any. With your change, it simply drops the `+abcd1234` part from the version string.

benaadams: Ah ok

src/Nethermind/Nethermind.Config/BlocksConfig.cs
```diff
@@ -10,9 +12,41 @@ namespace Nethermind.Config
 {
     public class BlocksConfig : IBlocksConfig
     {
+        public static bool AddVersionToExtraData { get; set; }
+
         public const string DefaultExtraData = "Nethermind";
-        private byte[] _extraDataBytes = Encoding.UTF8.GetBytes(DefaultExtraData);
-        private string _extraDataString = DefaultExtraData;
+        private byte[] _extraDataBytes = [];
```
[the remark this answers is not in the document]
rubo: Moreover, it doesn't want to take a thumbs-down. It insists on this :)

rubo (approved):

cbermudez97 (approved):

PR #8495, opened 2025-04-09
Add metrics for `engine_getBlobsV1` full and partial responses

marcindsobczak:
## Changes

- add metrics `NumberOfGetBlobsSuccesses` and `NumberOfGetBlobsFailures`
A successful response to `engine_getBlobsV1` may include some (or even all) `null` values, indicating that the requested blobs are not available. We already track metrics for the number of blobs requested and the number of blobs sent, and based on that, we estimate the return rate on mainnet to be around 85%. For now, partial responses (some blobs returned, some `nulls`) are useful for the CL.
However, after Fusaka's PeerDAS, partial responses will no longer be useful - CL will require either all requested blobs or none. This PR introduces new metrics to track fully successful responses (where all requested blobs are returned) versus successful but incomplete responses (where some blobs are missing). These metrics will help us estimate the rate of fully successful responses, which are critical for PeerDAS.
These metrics are already included in the PeerDAS branch, but we also want to start collecting data from mainnet.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Merge.Plugin/Handlers/GetBlobsHandler.cs
+++ b/src/Nethermind/Nethermind.Merge.Plugin/Handlers/GetBlobsHandler.cs
@@ -26,6 +26,7 @@ public class GetBlobsHandler(ITxPool txPool) : IAsyncHandler<byte[][], IEnumerab
 
     private IEnumerable<BlobAndProofV1?> GetBlobsAndProofs(byte[][] request)
     {
+        bool allBlobsAvailable = true;
         Metrics.NumberOfRequestedBlobs += request.Length;
 
         foreach (byte[] requestedBlobVersionedHash in request)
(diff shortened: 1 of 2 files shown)


benaadams (approved):
Wonder if we should  rationalize (remove) some metrics 🤔

flcl42 (approved):

marcindsobczak:
After Fusaka, we can remove `NumberOfRequestedBlobs` and `NumberOfSentBlobs`, as partial responses will no longer be useful. For now, however, they remain important. The metrics introduced in this PR aren't critical at the moment, but they will help us collecting data for PeerDAS and we’ll need to add them eventually anyway.

benaadams:
> After Fusaka..

Mostly meant other metrics 😀

Issue #8484, opened 2025-04-08
Return transactions back to TxPool on orphaned block


kamilchodola:
With spamoor tool we noticed, that when block is orphaned, transactions from it probably are not returned back to TXPool which results with only empty blocks being produced afterwards (assuming no direct rawTx sent to node but only gossipped via p2p).

Would be good to improve that on because of the chains which are using mostly Nethermind as EL client.

benaadams:
Could you retry this on latest master? May have been fixed by https://github.com/NethermindEth/nethermind/pull/8679

kamilchodola:
@claude verify if this was solved from code perspective and if we have aunit test for that

PR #8507, opened 2025-04-10
Handle tx reorged blocks better

benaadams:
Contributes to #8484; test passed so can't recreate

## Changes

- Added test for regular tx reorgs
- Changed reorged txs to PersistentBroadcast
- Cleared more caches on reorg
- Increased "Not syncing" distance
- Changed Debug message to Warn message when TxPool can't be updated
- Add ReaderWriterLock around HeadUpdate and AddTx; so new tx don't mess with the changes from changing head
  - Update head takes a write lock
  - Submitting Tx takes a read lock

- [x] Refactoring
- [x] Other: Test

#### Requires testing

- [x] Yes

- [x] Yes - Is test

### DIFF
--- a/src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.cs
+++ b/src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.cs
@@ -2340,5 +2340,71 @@ namespace Nethermind.TxPool.Test
             _blockTree.BlockAddedToMain += Raise.EventWith(blockReplacementEventArgs);
             await waitTask;
         }
+
+        [Test]
+        public async Task should_bring_back_reorganized_txs()
+        {
+            const long blockNumber = 358;
+
+            ITxPoolConfig txPoolConfig = new TxPoolConfig()
+            {
+                Size = 128,
+                BlobsSupport = BlobsSupportMode.Disabled
+            };
+            _txPool = CreatePool(txPoolConfig, GetCancunSpecProvider());
+
+            EnsureSenderBalance(TestItem.AddressA, UInt256.MaxValue);
+            EnsureSenderBalance(TestItem.AddressB, UInt256.MaxValue);
+            EnsureSenderBalance(TestItem.AddressC, UInt256.MaxValue);
+
+            Transaction[] txsA = { GetTx(TestItem.PrivateKeyA), GetTx(TestItem.PrivateKeyB) };
+            Transaction[] txsB = { GetTx(TestItem.PrivateKeyC) };
+
+            _txPool.SubmitTx(txsA[0], TxHandlingOptions.None).Should().Be(AcceptTxResult.Accepted);
+            _txPool.SubmitTx(txsA[1], TxHandlingOptions.None).Should().Be(AcceptTxResult.Accepted);
+            _txPool.SubmitTx(txsB[0], TxHandlingOptions.None).Should().Be(AcceptTxResult.Accepted);
+
+            _txPool.GetPendingTransactionsCount().Should().Be(txsA.Length + txsB.Length);
+            _txPool.GetPendingBlobTransactionsCount().Should().Be(0);
+
+            // adding block A
+            Block blockA = Build.A.Block.WithNumber(blockNumber).WithTransactions(txsA).TestObject;
+            await RaiseBlockAddedToMainAndWaitForNewHead(blockA);
+
+            _txPool.GetPendingTransactionsCount().Should().Be(txsB.Length);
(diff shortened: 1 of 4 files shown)


src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -307,7 +307,10 @@ private void ReAddReorganisedTransactions(Block? previousBlock)
                         continue;
                     }
                     _hashCache.Delete(tx.Hash!);
-                    SubmitTx(tx, isEip155Enabled ? TxHandlingOptions.None : TxHandlingOptions.PreEip155Signing);
+                    // For reorg we will submit with PersistentBroadcast
+                    SubmitTx(tx, isEip155Enabled ?
+                        TxHandlingOptions.PersistentBroadcast :
+                        TxHandlingOptions.PersistentBroadcast | TxHandlingOptions.PreEip155Signing);
```
LukaszRozmej, line 313: why persistent broadcast?

benaadams:
- Add ReaderWriterLock around HeadUpdate and AddTx; so new tx don't mess with the changes from changing head
  - Update head takes a write lock
  - Submitting Tx takes a read lock

src/Nethermind/Nethermind.TxPool/TxPool.cs
LukaszRozmej: It doesn't really give boost for inclusion on this node, just sends to other nodes more aggresively

marcindsobczak: I'm not a fan of the idea of adding reorged transactions to the persistent broadcast. We had this feature a few years ago but removed it as an optimization.
Persistent broadcast is a strong booster for local transactions, and I don't think that we should care about reorged transactions that much.
That said, I'm not opposed to implementing it if you have a strong preference - I just personally prefer not to.

src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.cs
```diff
@@ -2340,5 +2340,71 @@ private async Task RaiseBlockAddedToMainAndWaitForNewHead(Block block, Block pre
...
+            _txPool.TryGetPendingTransaction(txsB[0].Hash!, out _).Should().BeFalse();
+
+            // txs from reorganized blockA should be readded to tx pool
+            _txPool.GetPendingTransactionsCount().Should().Be(txsA.Length);
+            _txPool.TryGetPendingTransaction(txsA[0].Hash!, out Transaction tx1).Should().BeTrue();
+            _txPool.TryGetPendingTransaction(txsA[1].Hash!, out Transaction tx2).Should().BeTrue();
+
+            tx1.Should().BeEquivalentTo(txsA[0], static options => options
+                .Excluding(static t => t.GasBottleneck)    // GasBottleneck is not encoded/decoded...
+                .Excluding(static t => t.PoolIndex));      // ...as well as PoolIndex
```
marcindsobczak, line 2394: Nitpick: I think we don't need to exclude `GasBottleneck` here, it should remain the same after readding. Value of `GasBottleneck` is added in submit tx pipeline.
It was probably copy-pasted from tests about blob txs loaded from db - then we needed to exclude it.

marcindsobczak (approved):
Looks good, but I'm not a fan of the part with adding reorged transactions to the persistent broadcast.

src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.cs
benaadams: Removed (or rather added back)

src/Nethermind/Nethermind.TxPool/TxPool.cs
benaadams: Removed the PersistentBroadcast

src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -456,36 +468,52 @@ public AcceptTxResult SubmitTx(Transaction tx, TxHandlingOptions handlingOptions
...
+                }
+            }
+            finally
+            {
+                _newHeadLock.ExitReadLock();
+            }
 
             if (!accepted)
             {
                 Metrics.PendingTransactionsDiscarded++;
```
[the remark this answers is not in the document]
benaadams: Unintentional from trying to keep lock small

marcindsobczak (approved):

PR #8512, opened 2025-04-11
Prefer mev reward tx value when calculating mev

benaadams:
## Changes

- Prefer the value of the mev tx `.Value` when calculating mev-boost block mev; as sometimes the beneficiary has their own tx in the block making straight balance calc unreliable (can currently sometimes outputs 0)
- Change block mev value for local production and block to 4 decimals as it tracks the individual improvements better since are more often now

image

- [x] Other: Logging

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Processing/ProcessingStats.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/ProcessingStats.cs
@@ -379,12 +379,7 @@ namespace Nethermind.Consensus.Processing
             {
                 UInt256 beforeBalance = _stateReader.GetBalance(beforeRoot, beneficiary);
                 UInt256 afterBalance = _stateReader.GetBalance(afterRoot, beneficiary);
-                if (beforeBalance < afterBalance)
-                {
-                    return afterBalance - beforeBalance;
-                }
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Consensus/Processing/ProcessingStats.cs
```diff
@@ -364,6 +374,18 @@ private void GenerateReport(BlockData data)
...
+            UInt256 CalculateBalanceChange(Hash256 beforeRoot, Hash256 afterRoot, Address beneficiary)
+            {
+                UInt256 beforeBalance = _stateReader.GetBalance(beforeRoot, beneficiary);
+                UInt256 afterBalance = _stateReader.GetBalance(afterRoot, beneficiary);
+                if (beforeBalance < afterBalance)
+                {
+                    return afterBalance - beforeBalance;
+                }
+
+                return default;
```
LukaszRozmej, line 387:
```suggestion
                return beforeBalance < afterBalance ? afterBalance - beforeBalance : UInt256.Zero;
```

LukaszRozmej (approved):

PR #8516, opened 2025-04-11
Safely cast Optimism subtypes

### LINKED ISSUE #8515: `eth_getBlockByNumber` can trigger `InvalidCastException`
**Description**
Invoking the method `eth_getBlockByNumber` in an OP chain can trigger a `InvalidCastException`.

**Steps to Reproduce**
1. Invoke the method `eth_getBlockByNumber(["finalized",true])`

**Actual behavior**
Nethermind produces the following response:

```json
{
  "jsonrpc": "2.0",
  "error": {
    "code": -32603,
    "message": "Internal error",
    "data": "System.InvalidCastException: Unable to cast object of type 'Nethermind.Core.TxReceipt' to type 'Nethermind.Optimism.OptimismTxReceipt'.\n   at System.Linq.Enumerable.CastICollectionIterator1.ToArray()\n   at Nethermind.Optimism.Rpc.OptimismEthRpcModule.GetBlock(BlockParameter blockParameter, Boolean returnFullTransactionObjects) in /app/src/Nethermind/Nethermind.Optimism/Rpc/OptimismEthRpcModule.cs:line 252\n   at System.Reflection.MethodInvoker.InvokeImpl(Object obj, Object arg1, Object arg2, Object arg3, Object arg4)\n   at System.Reflection.MethodInvoker.Invoke(Object obj, Span1 arguments)\n   at Nethermind.JsonRpc.JsonRpcService.ExecuteAsync(JsonRpcRequest request, String methodName, ResolvedMethodInfo method, JsonRpcContext context) in /app/src/Nethermind/Nethermind.JsonRpc/JsonRpcService.cs:line 175"
  },
  "id": 1
}
```

**Expected behavior**
Nethermind should not throw an exception and instead return a spec-compliant result.

 - Operating System: Irrelevant
 - Version: Nethermind v1.31.6
 - Installation Method: Irrelevant
 - Consensus Client: op-node v1.12.2

emlautarom1:
Fixes #8515 

## Changes

- Cast only when appropriate avoiding `InvalidCastException`s.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks

The PR is initially marked as draft due to an issue with subtypes and serialization.


src/Nethermind/Nethermind.Optimism/Rpc/OptimismEthRpcModule.cs
```diff
@@ -77,31 +78,35 @@ public OptimismEthRpcModule(
...
+                            optimismTxReceipt,
+                            tx.GetGasInfo(spec, block.Header),
+                            l1BlockGasInfo.GetTxGasInfo(tx),
+                            receipts.GetBlockLogFirstIndex(receipt.Index))
+                        : new ReceiptForRpc(
+                            tx.Hash!,
+                            receipt,
+                            tx.GetGasInfo(spec, block.Header),
+                            receipts.GetBlockLogFirstIndex(receipt.Index)))];
+        return ResultWrapper<ReceiptForRpc[]?>.Success(result);
```
emlautarom1, line 109: We have a problem here:
- If we set the return type to `OptimismReceiptForRpc`, then we cannot store a "default" `ReceiptForRpc`
- If we set the return type to `ReceiptForRpc`, when serializing, subtypes are treated as the static type so subtype-only fields like `DepositReceiptVersion` are not serialized.

emlautarom1: I ended up solving it by reusing the `OptimismReceiptForRpc` since all its fields can be ignored when null, so it essentially looks like the base class `ReceiptForRpc`. An additional constructor was enough.

src/Nethermind/Nethermind.Optimism/Rpc/OptimismEthRpcModule.cs
```diff
@@ -77,31 +78,35 @@ public OptimismEthRpcModule(
...
+                            optimismTxReceipt,
+                            tx.GetGasInfo(spec, block.Header),
+                            l1BlockGasInfo.GetTxGasInfo(tx),
+                            receipts.GetBlockLogFirstIndex(receipt.Index))
+                        : new OptimismReceiptForRpc(
+                            tx.Hash!,
+                            receipt,
+                            tx.GetGasInfo(spec, block.Header),
+                            receipts.GetBlockLogFirstIndex(receipt.Index)))];
+        return ResultWrapper<ReceiptForRpc[]?>.Success(result);
```
emlautarom1, line 109: There is a warning regarding covariant conversions but it should be fine considering that the resulting array is intended to be read-only.

LukaszRozmej (approved):

src/Nethermind/Nethermind.Optimism/Rpc/OptimismEthRpcModule.cs
Scooletz: Yeah, the usual case for it. Having something like `IEnumerable<>` would be better for the result type (that would address the warning) but that would change a lot of the API surface.

src/Nethermind/Nethermind.Optimism/Rpc/OptimismEthRpcModule.cs
```diff
@@ -235,7 +251,7 @@ protected override ResultWrapper<TransactionForRpc> GetTransactionByBlockAndInde
                     txIndex: index);
                 if (transactionModel is DepositTransactionForRpc depositTx)
                 {
-                    OptimismTxReceipt? receipt = receipts.FirstOrDefault(r => r.TxHash?.Equals(hash) ?? false);
+                    OptimismTxReceipt? receipt = (OptimismTxReceipt?)receipts.FirstOrDefault(r => r.TxHash?.Equals(hash) ?? false);
```
Scooletz, line 254: Could you provide a comment or spec link why is this always the case?

Scooletz (commented):
Looks good. One minor question

src/Nethermind/Nethermind.Optimism/Rpc/OptimismEthRpcModule.cs
emlautarom1: According to the yellow paper, the Receipt type is the same as the Transaction type, so a Deposit transaction should be associated with a Deposit receipt.

If you take a look at op-geth's codebase, when working with a Deposit transaction, the associated receipt contains Optimism's specific fields (ex. `DepositNonce`), which matches our `OptimismTxReceipt` type.

https://github.com/ethereum-optimism/op-geth/blob/d1b66c4722e7fb5525efbcadda4c1ab3f7dacaa8/internal/ethapi/api.go#L1611-L1616

Unfortunately, a lot of these invariants are not part of the type system and I haven't been able to find them expressed in the specs so I might be wrong about them.

PR #8523, opened 2025-04-14
Limit time any single block can be produced for

benaadams:
## Changes

- We can improve blocks until asked for the payload, however no single block should consume the whole slot to run, as it then needs to be propagated, validated, attested to etc

#### Requires testing

- [x] No - Config change

### DIFF
--- a/src/Nethermind/Nethermind.Config/BlocksConfig.cs
+++ b/src/Nethermind/Nethermind.Config/BlocksConfig.cs
@@ -58,6 +58,7 @@ namespace Nethermind.Config
         public int PreWarmStateConcurrency { get; set; } = 0;
 
         public int BlockProductionTimeoutMs { get; set; } = 4_000;
+        public double SingleBlockImprovementOfSlot { get; set; } = 0.25;
 
         public int GenesisTimeoutMs { get; set; } = 40_000;
 
--- a/src/Nethermind/Nethermind.Merge.Plugin/MergePlugin.cs
+++ b/src/Nethermind/Nethermind.Merge.Plugin/MergePlugin.cs
@@ -264,7 +264,7 @@ public partial class MergePlugin(ChainSpec chainSpec, IMergeConfig mergeConfig) 
 
             // Single block shouldn't take a full slot to run
             // We can improve the blocks until requested, but the single block still needs to be run in a timely manner
-            double maxSingleImprovementTimePerSlot = _blocksConfig.SecondsPerSlot * 0.25;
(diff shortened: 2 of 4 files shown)


smartprogrammer93 (approved):

src/Nethermind/Nethermind.Merge.Plugin/MergePlugin.cs
```diff
@@ -262,16 +262,19 @@ public Task InitRpcModules()
             }
             Thread.Sleep(5000);
 
+            // Single block shouldn't take a full slot to run
+            // We can improve the blocks until requested, but the single block still needs to be run in a timely manner
+            double maxSingleImprovementTimePerSlot = _blocksConfig.SecondsPerSlot * 0.25;
```
smartprogrammer93, line 267: this should be configurable actually. for L2s, nethermind can use the whole slot time for building without issues.

smartprogrammer93 (changes requested):

src/Nethermind/Nethermind.Merge.Plugin/MergePlugin.cs
smartprogrammer93: should be specified in BlocksConfig (the percentage of time out of the whole slot for block building)

benaadams: Made it confiurable

src/Nethermind/Nethermind.Config/BlocksConfig.cs
```diff
@@ -58,6 +58,7 @@ private static string GetDefaultVersionExtraData()
         public int PreWarmStateConcurrency { get; set; } = 0;
 
         public int BlockProductionTimeoutMs { get; set; } = 4_000;
+        public double SingleBlockImprovementOfSlot { get; set; } = 0.25;
```
smartprogrammer93, line 61: Given this is the default, now you need to add this field in every chain config file that is not PoS, (Taiko Alethia, Taiko Hekla, Optimism, Base, etc etc)

smartprogrammer93 (changes requested):

src/Nethermind/Nethermind.Config/BlocksConfig.cs
benaadams: Why?

Taiko doesn't use timeout anyway; as it doesn't improve blocks

smartprogrammer93: okay, ben brought up that taiko doesnt do block improvement which is true, and OP can be adjusted later

smartprogrammer93 (approved):

PR #8524, opened 2025-04-14
Feature/Allow multiple class for the same rpc module.

asdacap:
- Allow multiple rpc module class of the same rpc module name (Admin/Debug/Eth) to work add at same time.
  - This allow splitting the same rpc module across different class which reduces change. 
  - Eg: The era import/export command does not need to be in `AdminRpcModule`.
  - This allow plugins to add/replace method to existing rpc module without changing the original module.
- Extract era history command outside of `AdminRpcModule.`
- Allow passing in rpc module factory via rpc module provider constructor.
  - This allow rpc modules to be added via autofac modules.
- Move flashbots plugin to use modules.
- Move some easily movable rpc module to `RpcModules`.
- Remove some `INethermindApi` fields.

- [X] New feature
- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- [x] Still sync fine.
- [X] `admin_verifyTrie` still works.
- [X] `admin_exportHistory` still works.


src/Nethermind/Nethermind.JsonRpc/Modules/RpcModuleProvider.cs
```diff
@@ -52,22 +65,36 @@ public RpcModuleProvider(IFileSystem fileSystem, IJsonRpcConfig jsonRpcConfig, I
...
+                var poolRecord = GetPool(pool);
+
                 // FrozenDictionary can't be directly updated (which makes it fast for reading) so we combine the two sets of
                 // data as an Enumerable create an new FrozenDictionary from it and then update the reference
-                _pools = GetPools<T>(moduleType, pool).ToFrozenDictionary(StringComparer.Ordinal);
-                _methods = _methods.Concat(GetMethods<T>(moduleType)).ToFrozenDictionary(StringComparer.Ordinal);
+                _pools = _pools.Concat(methods
+                    .Select(method => new KeyValuePair<string, Pool>(method.Key, poolRecord)))
+                    .ToFrozenDictionary<string, Pool>(StringComparer.Ordinal);
+                _methods = _methods.Concat(methods).ToFrozenDictionary(StringComparer.Ordinal);
```
LukaszRozmej, line 97: Maybe the methods and pools should be concatenated in normal dictionary and Frozen when first read?

src/Nethermind/Nethermind.JsonRpc/Modules/RpcModuleProvider.cs
```diff
@@ -52,22 +65,36 @@ public RpcModuleProvider(IFileSystem fileSystem, IJsonRpcConfig jsonRpcConfig, I
...
-                if (_logger.IsWarn) _logger.Warn($"Cannot register {typeof(T).Name} as a JSON RPC module because it does not have a {nameof(RpcModuleAttribute)} applied.");
+                if (_logger.IsWarn) _logger.Warn($"Cannot register {moduleClass.Name} as a JSON RPC module because it does not have a {nameof(RpcModuleAttribute)} applied.");
                 return;
             }
 
+            if (_logger.IsTrace) _logger.Trace($"Registering module {moduleClass.Name} as part of {attribute.ModuleType} module");
             string moduleType = attribute.ModuleType;
             lock (_updateRegistrationsLock)
             {
+                using var methods = GetMethods<T>(moduleType).ToPooledList(0);
```
LukaszRozmej, line 89: ToPooledList(0) is weird usage

LukaszRozmej (approved):

src/Nethermind/Nethermind.JsonRpc/Modules/RpcModuleProvider.cs
asdacap: Fixed

src/Nethermind/Nethermind.JsonRpc/Modules/RpcModuleProvider.cs
asdacap: Changed to `ToArray`

PR #8525, opened 2025-04-15
Scoped worldstate

tanishqjasoria:
Fixes Closes Resolves #

## Changes

- _List the changes_

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Remarks


src/Nethermind/Nethermind.State/WorldState.cs
```diff
@@ -248,32 +316,106 @@ public Snapshot TakeSnapshot(bool newTransactionStart = false)
...
+        private void EnsureInScope()
+        {
+            if (!_inScope) throw new InvalidOperationException("WorldState used outside of its scope.");
+        }
+
+        private bool _inScope;
+
+        public IDisposable BeginScope(Hash256 stateRoot)
+        {
+            if (_inScope)
```
asdacap, line 366: Do this case happen? Is this expected?

asdacap:
LGTM

src/Nethermind/Nethermind.Blockchain/Contracts/Contract.ConstantContract.cs
```diff
@@ -85,7 +85,7 @@ public override object[] Call(CallInfo callInfo)
 
                 lock (_readOnlyTxProcessorSource)
                 {
-                    using var scope = _readOnlyTxProcessorSource.Build(GetState(callInfo.ParentHeader));
+                    using var scope = _readOnlyTxProcessorSource.BuildAndInit(GetState(callInfo.ParentHeader));
```
tanishqjasoria, line 88: here we create a new scope, that calls WorldState.BeginScope()

then nested inside the call there is another SystemCall to read the contract address for this call, there we create another scope and that is what create the issues of calling BeginScope twice on the same WorldState
@asdacap

src/Nethermind/Nethermind.State/IWorldState.cs
```diff
@@ -17,6 +17,10 @@ namespace Nethermind.State;
 /// </summary>
 public interface IWorldState : IJournal<Snapshot>, IReadOnlyStateProvider
 {
+
+    public IDisposable BeginScope(Hash256 stateRoot);
```
asdacap, line 21: Maybe `BlockHeader? parent` would be better? Could support world state that does not calculate state root.

tanishqjasoria: yes, that is the idea
this is just a try to see if things work with stateRoot - easy to do in tests - and then i can modify it to take blockHeader+spec instead of stateRoot.

src/Nethermind/Nethermind.State/WorldState.cs
Scooletz: I'd capture the current scope as a field instead of boolean and bump up the counter there. Otherwise you can run into situations where nested scopes are not disposed and you won't get any notification about it. With a counter you get all the book keeping you need.

src/Nethermind/Nethermind.State/WorldState.cs
```diff
@@ -248,32 +314,110 @@ public Snapshot TakeSnapshot(bool newTransactionStart = false)
...
+
+        private void EndScope()
+        {
+            Reset();
+            _inScope = false;
+        }
+
+        private class ScopeGuard(WorldState state) : IDisposable
+        {
+            private bool _disposed;
```
Scooletz, line 398: I'd make it `int _counter` and do ref counting here. On each `BeginScope` bump up the current, on each dispose -1. When goes to 0, `state.EndScope`.

src/Nethermind/Nethermind.State/WorldState.cs
Scooletz: Ok. One more thing. Why don't we capture the previous root hash and reset to it at the end of the scope? That would be a better scoping mechanism?

src/Nethermind/Nethermind.Consensus/Processing/ReadOnlyTxProcessingScope.cs
```diff
@@ -9,20 +10,31 @@ namespace Nethermind.Consensus.Processing;
...
     public ITransactionProcessor TransactionProcessor => transactionProcessor;
     public IWorldState WorldState => worldState;
     public void Reset()
     {
-        worldState.StateRoot = originalStateRoot;
-        worldState.Reset();
+        // this means the scope was not initialized here, so don't need to reset anything
+        if (_worldStateScopeGuard is not null)
+        {
+            worldState.StateRoot = _originalStateRoot;
```
Scooletz, line 36: And here's the missing rootHash capturing I believe. If the `_worldStateScopeGuard ` captured the previous hash that it resets to on `.Dispose`, you'd need no manual capturing of the original state root.

Scooletz (commented):
A PR description would be beneficial for the reviewer. A review provided. I'd go with capturing the previous root hash and creating a single scope.

LukaszRozmej:
I somewhat like the concept, but in the current form I think it can be more confusing than helpful. Basically after `BeginScope` all other state operations should happen on the scope, rather than on the same component `BeginScope` was called. That would ensure API that guarantees the separation of scopes and correctness of usage.

tanishqjasoria:
> I somewhat like the concept, but in the current form I think it can be more confusing than helpful. Basically after `BeginScope` all other state operations should happen on the scope, rather than on the same component `BeginScope` was called. That would ensure API that guarantees the separation of scopes and correctness of usage.

yeah, but creating the scope and then passing the scope as a argument results in a HUGE refactor and this approach avoids that refactor.
Another thing that maybe we can do is, pass IWorldStateManager everywhere and when do IWorldStateManager.BeginScope like we do in this PR and then we want to access the state we have to first do IWorldState = IWorldStateManager.CurrentScope and then do the state operations
Thoug, i dont like this idea that much.

src/Nethermind/Nethermind.Consensus/Producers/BlockProducerBase.cs
```diff
@@ -192,7 +192,11 @@ protected bool TrySetState(Hash256? parentStateRoot)
         {
             if (parentStateRoot is not null && StateProvider.HasStateForRoot(parentStateRoot))
             {
-                StateProvider.StateRoot = parentStateRoot;
+                // TODO: do we need this here?
+                using (StateProvider.BeginScope())
+                {
+                    StateProvider.StateRoot = parentStateRoot;
+                }
```
benaadams, line 199: This is a pretty weird pattern that emerges

asdacap: Should make StateRoot read only

tanishqjasoria: > Should make StateRoot read only
then how do we change it when we need to

tanishqjasoria: > This is a pretty weird pattern that emerges
yes, i don't think we need this

PR #8531, opened 2025-04-17
Feature/better dicovery metrics

asdacap:
- Add metric for total number of candidate peer
- Add metric for rate of candiate peer filter.
- Add metric for incoming and outgoing discovery message counts.
- Add option to disable static labels.

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

#### Notes on testing

- Tested manually. Looks like this:
Screenshot_2025-04-17_20-39-57


src/Nethermind/Nethermind.Core/Collections/DictionaryExtensions.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Nethermind.Core.Collections;
+
+public static class DictionaryExtensions
+{
+    public static void Increment<TKey>(this IDictionary<TKey, int> dictionary, TKey key)
+    {
+        if (!dictionary.TryAdd(key, 1)) dictionary[key]++;
+    }
+}
```
LukaszRozmej, line 14: This doesn't guarantee thread-safety.

Maybe instead of using dictionaries in some places, we could use Arrays of ints, get ref to the array element and use Interlocked? If array indexes would be mapped to enum values, we could get labels with some attribute?

asdacap: This is not for the concurrent dictionary or the metric. This is for the counters in filtering candidate peers.

LukaszRozmej: If they are regular dictionaries you would get better results using [GetValueRefOrAddDefault](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.collectionsmarshal.getvaluereforadddefault?view=net-9.0#system-runtime-interopservices-collectionsmarshal-getvaluereforadddefault-3(system-collections-generic-dictionary((-0-1))-alternatelookup((-2))-2-system-boolean@)) and then incrementing the ref, we already use it in some places in our codebase

src/Nethermind/Nethermind.Core/Nethermind.Core.csproj
```diff
@@ -13,6 +13,7 @@
     <PackageReference Include="Microsoft.IdentityModel.JsonWebTokens" />
     <PackageReference Include="Nethermind.Crypto.SecP256k1" />
     <PackageReference Include="Nethermind.Numerics.Int256" />
+    <PackageReference Include="NonBlocking" />
```
LukaszRozmej, line 16: In most of our test regular ConcurrentDictionary was better than NonBlocking one. Only in very high contention scenarios non blocking can be better

LukaszRozmej (approved):

PR #8550, opened 2025-04-23
WIP: block-stm

LukaszRozmej:
This PR introduces block-STM.
Based on paper: https://arxiv.org/abs/2203.06871 : Paper has spin-loop and 2 multi-threading bugs though(!)

Current status: abstract algorithm is implemented and tested. It is not yet integrated with block processing in any way.

## Changes

- _List the changes_

- [x] Optimization

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

Large blocks - perfnet?

- [x] Yes
- [ ] No


src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/ParallelRunner.cs
```diff
@@ -0,0 +1,84 @@
...
+        using var handle = Thread.CurrentThread.BoostPriorityHighest();
+        TxTask task = scheduler.NextTask();
+        do
+        {
+            if (task.IsEmpty)
+            {
+                // Alternatives:
+                // - Use ManualResetEventSlim and block threads, but don't have this complicated TCS stuff
+                // - Spin this loop until scheduler.Done (what paper proposes)
+                await scheduler.WorkAvailable.WaitAsync();
```
LukaszRozmej, line 46: This is questionable, other mentioned approaches can be found:
- ManualResetEventSlim: [d21026c](https://github.com/NethermindEth/nethermind/pull/8550/commits/d21026cbf5b944d864c12fe47075e95007d90639) and later fix in [0710541](https://github.com/NethermindEth/nethermind/pull/8550/commits/071054173838db6a845e166da9c2877ca70db8ba)
- Spinning: [b1e015f](https://github.com/NethermindEth/nethermind/pull/8550/commits/b1e015ffecfe6fd172f2bc0fde57062e7adbaa9a)

I'm happy to hear more about your thoughts

LukaszRozmej (commented):
For now please focus on reviewing most important classes:
src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/ParallelRunner.cs
src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/ParallelScheduler.cs
src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/MultiVersionMemory.cs

with supporting:
src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/Version.cs
src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/ParallelTrace.cs
src/Nethermind/Nethermind.Core/Tasks/AsyncManualResetEventSlim.cs
src/Nethermind/Nethermind.Core/Threading/InterlockedEx.cs

While tests are:
src/Nethermind/Nethermind.Consensus.Test/Processing/ParallelProcessing/ParallelRunnerTests.cs

Rest is temporary and will be reworked

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/MultiVersionMemory.cs
```diff
@@ -0,0 +1,194 @@
...
+
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using Nethermind.Core.Collections;
+using Nethermind.Core.Extensions;
+
+namespace Nethermind.Consensus.Processing.ParallelProcessing;
+
+public class MultiVersionMemory<TLocation, TLogger>(ushort txCount, ParallelTrace<TLogger> parallelTrace) where TLogger : struct, IIsTracing where TLocation : notnull
```
LukaszRozmej, line 12: Another question is what Location should be?
We can have a map from Adress/StorageSlot to int and use that, or we can use complex types more directly.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/Version.cs
```diff
@@ -0,0 +1,17 @@
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+namespace Nethermind.Consensus.Processing.ParallelProcessing;
+
+public readonly record struct Version(ushort TxIndex, ushort Incarnation)
+{
+    public bool IsEmpty => TxIndex == ushort.MaxValue && Incarnation == ushort.MaxValue;
```
Scooletz, line 8: Go with default meaning 0? It'd be more natural.

src/Nethermind/Nethermind.Core/Tasks/AsyncManualResetEventSlim.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        TaskCompletionSource<bool>? newTcs = null;
+        while (true)
+        {
+            TaskCompletionSource<bool> tcs = _tcs;
+            if (!tcs.Task.IsCompleted)
+                return;
+
+            newTcs ??= CreateNewTcs();
+            if (Interlocked.CompareExchange(ref _tcs, newTcs, tcs) == tcs)
```
Scooletz, line 37: You interlock here but in `Set` you just read the field with `volatile`. This could result in Reset swaping the underlying `tcs` while `Set` would get the previous and try set result on it. What is the contract here? Who's gonna call Set/Reset?

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/MultiVersionMemory.cs
```diff
@@ -0,0 +1,194 @@
...
+{
+    private readonly record struct Value(ushort Incarnation, byte[] Bytes);
```
Scooletz, line 14: Why `Value` and this structure in general captures `byte[]`?

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/MultiVersionMemory.cs
Scooletz: The ongoing question that I have (trying it out under #8482 partially) is would we benefit from having a concurrent map of Address/StorageCell -> int and then use it for navigation and addressing within a block. This additional mapping would incur some cost, but at the same time it would decrease the friction in all the underlying structures making either their keys atomic (for concurrent access) or simply smaller for hash sets etc. Having a uniform mapping would make everything generic of `<int>`.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/Locations.cs
```diff
@@ -0,0 +1,31 @@
...
+
+public interface ILocations<in TKey>
+{
+    int GetLocation(TKey address);
+}
+
+public class Locations : ILocations<Address>, ILocations<StorageCell>
+{
+    private int _index = -1;
+    private readonly ConcurrentDictionary<AddressAsKey, int> _addressLocations = new();
```
Scooletz, line 19: So this is the mapping that I mentioned elsewhere. I think, if we made it in a uniform way, we could make it potentially cheaper and totally alloc free (https://github.com/NethermindEth/nethermind/blob/7357f676685cf009c40dc84c8b46a8719a7ecba6/src/Nethermind/Nethermind.State/StorageValueMap.cs from #8482 )

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/MultiVersionMemory.cs
```diff
@@ -0,0 +1,194 @@
...
+
+    private bool UpdateWrittenLocations(int txIndex, Dictionary<TLocation, byte[]> writeSet)
+    {
+        ConcurrentDictionary<TLocation, Value> txData = _data[txIndex];
+        ref HashSet<TLocation>? lastWritten = ref _lastWrittenLocations[txIndex];
+        lastWritten ??= new HashSet<TLocation>();
+
+        if (lastWritten.Count != 0)
+        {
+            using ArrayPoolList<TLocation> toRemove = new(lastWritten.Count);
```
Scooletz, line 42: I think I do understand that we remove what was not written this round. I don't understand why.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/MultiVersionMemory.cs
```diff
@@ -0,0 +1,194 @@
...
+            }
+        }
+    }
+
+    public Status TryRead(TLocation location, ushort txIndex, out Version version, out byte[]? value)
+    {
+        long id = parallelTrace.ReserveId();
+        ushort prevTx = txIndex;
+        prevTx--;
+        while (prevTx != ushort.MaxValue)
```
Scooletz, line 91: A magic const.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/ParallelScheduler.cs
```diff
@@ -0,0 +1,274 @@
...
+            }
+        }
+
+        Interlocked.Decrement(ref _activeTasks);
+        return Version.Empty;
+    }
+
+    public TxTask NextTask()
+    {
+        bool validating = Volatile.Read(ref _validationIndex) < Volatile.Read(ref _executionIndex);
```
Scooletz, line 101: You make two volatile reads. I don't follow the ordering semantics here.

Scooletz (commented):
Initial comments provide. We need some comments in this code and PR because it's not easy to understand thoroughly what it does.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/Version.cs
LukaszRozmej: Transactions are indexed from 0...N, so special value is easier as MaxValue.
I might change it to int to support more than 65k transaction blocks, then special value can be -1.

src/Nethermind/Nethermind.Core/Tasks/AsyncManualResetEventSlim.cs
LukaszRozmej: This intends to be replacement for `ManualResetEventSlim` but with async wait

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/MultiVersionMemory.cs
LukaszRozmej: Read the paper ;)

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/MultiVersionMemory.cs
LukaszRozmej: not really, it is about underflow (going below 0)

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/MultiVersionMemory.cs
```diff
@@ -0,0 +1,340 @@
...
+    /// </summary>
+    private readonly HashSet<TLocation>?[] _lastWrittenLocations = new HashSet<TLocation>?[txCount];
+
+    /// <summary>
+    /// Mapping between TransactionIndex -> HashSet that will store reads by last incarnation of the transaction.
+    /// </summary>
+    private readonly HashSet<Read<TLocation>>?[] _lastReads = new HashSet<Read<TLocation>>[txCount];
+
+    // For given transaction incarnation it stores it's writeset into _data and updates _lastWrittenLocations
+    private bool ApplyWriteSet(Version version, Dictionary<TLocation, byte[]> writeSet)
```
Scooletz, line 63: I counted how many times the location is accessed here. By access, I mean using it in a data structure that requires comparisons and hash calculation. 

1. a loop over `writeSet` first with setting them by the key
2. `writeSet.ContainsKey`
3. `toRemove.Add` copies the location, so the smaller it is, the better
4. `lastWritten.Remove(id)`
5. `txData.Dictionary.Remove(id, out _);`
6. `lastWritten.UnionWith(writeSet.Keys)`

This means 6 places where keys are compared/hashed. I think if we want to make it work nice and smooth we need to have this identity map so that it can map an arbitrary blittable to something that is a pointer/ref like in #8482. This would make all this operations trivial (pointers hash or equality) and would reduce the sizes of all the structures used. 

BUT, this does not impact correctness and we can optimize it at the end. I can provide a concurrent, non-allocating, lock-free map. I'll familiarize myself with the paper now.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/MultiVersionMemory.cs
```diff
@@ -0,0 +1,340 @@
...
+                    return Status.Ok;
+                }
+            }
+            else
+            {
+                prevTransactionData.Lock.ExitReadLock();
+            }
+
+            // if not we are going backwards until we find it or checked all transactions
+            prevTx--;
```
Scooletz, line 197: If a value was set in transaction number 1 with an arbitrary incarnation (not important), and we retrieve it from 60, do we need to spin through all the transaction dictionaries to get the value?

LukaszRozmej: No if `prevTransactionData.Dictionary.TryGetValue` is true we have a `return` statement. So we go until we find one, but need precedence of highest index smaller than current transaction.

This definitively could be optimized somehow.

PR #8557, opened 2025-04-26
Maybe fix incorrect total difficulty

asdacap:
- There seems to be a chance that the next header difficulty is not updated by about one batch.
- Unfortunately this is not reliably reproducable nor can I figureout how can that be.
- Anyway, this combine the next header hash and next header difficulty into one so that they are both updated at the same time.
- Also set the next expected header in a finally block in case an exception happen in blocktree maybe. But that would also mean next header hash also not updated, so what happen? I don't know.

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

#### Notes on testing

- Can't reproduce issue.

### DIFF
--- a/src/Nethermind/Nethermind.Synchronization/FastBlocks/FastHeadersSyncFeed.cs
+++ b/src/Nethermind/Nethermind.Synchronization/FastBlocks/FastHeadersSyncFeed.cs
@@ -44,11 +44,8 @@ namespace Nethermind.Synchronization.FastBlocks
 
         private readonly ulong _fastHeadersMemoryBudget;
         protected long _lowestRequestedHeaderNumber;
-
-        protected Hash256 _nextHeaderHash;
-        protected UInt256? _nextHeaderTotalDifficulty;
-
         protected long _pivotNumber;
+        protected (Hash256 Hash, UInt256? TotalDifficulty) _expectedNextHeader;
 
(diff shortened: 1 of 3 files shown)


src/Nethermind/Nethermind.Synchronization/FastBlocks/FastHeadersSyncFeed.cs
```diff
@@ -44,11 +44,8 @@ public class HeadersSyncFeed : ActivatedSyncFeed<HeadersSyncBatch?>
 
         private readonly ulong _fastHeadersMemoryBudget;
         protected long _lowestRequestedHeaderNumber;
-
-        protected Hash256 _nextHeaderHash;
-        protected UInt256? _nextHeaderTotalDifficulty;
-
         protected long _pivotNumber;
+        protected (Hash256 Hash, UInt256? TotalDifficulty) _expectedNextHeader;
```
LukaszRozmej, line 48: This is a complex struct, I think it's assignment isn't guaranteed to be atomic. If you want atomic `=` than make it a reference type.

asdacap: Changed to record.

LukaszRozmej (approved):

PR #8564, opened 2025-04-28
Skip reloads of accounts for txpool delegations

benaadams:
## Changes

- Account has been already loaded, so don't load 2 more times checking code hash

Before

image

After

image

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Evm/CodeInfoRepository.cs
+++ b/src/Nethermind/Nethermind.Evm/CodeInfoRepository.cs
@@ -84,8 +84,13 @@ public class CodeInfoRepository : ICodeInfoRepository
 
     private static ICodeInfo InternalGetCachedCode(IReadOnlyStateProvider worldState, Address codeSource, IReleaseSpec vmSpec)
     {
-        ICodeInfo? cachedCodeInfo = null;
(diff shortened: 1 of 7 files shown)


src/Nethermind/Nethermind.Evm/CodeInfoRepository.cs
```diff
@@ -113,9 +118,9 @@ private static ICodeInfo InternalGetCachedCode(IReadOnlyStateProvider worldState
 
         [DoesNotReturn]
         [StackTraceHidden]
-        static void MissingCode(Address codeSource, in ValueHash256 codeHash)
+        static void MissingCode(in ValueHash256 codeHash)
         {
-            throw new NullReferenceException($"Code {codeHash} missing in the state for address {codeSource}");
+            throw new NullReferenceException($"Code {codeHash} missing in the state");
```
LukaszRozmej, line 123: Can we change the exception? It is confusing to throw `NullReferenceException` explicitly

LukaszRozmej (approved):

src/Nethermind/Nethermind.Evm/CodeInfoRepository.cs
benaadams: Throwing better exception

PR #8567, opened 2025-04-28
Commit state according to the regular spec, before system tx spec usage

flcl42:
## Changes

- Commit additionally between the applications of two different specs

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No (tested by hive)

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.cs
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: 2022 Demerzel Solutions Limited
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
 // SPDX-License-Identifier: LGPL-3.0-only
 
 using System;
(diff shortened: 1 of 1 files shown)


benaadams:
Should it be more like this?

image

flcl42:
@benaadams yep I've tried this and it does not work 🤷

benaadams (approved):
Guess ship it as fix now; and then refactor later

LukaszRozmej:
> Should it be more like this?
> 
> image

I think both are needed?
MinerRewards and Withdrawals are system transactions

LukaszRozmej (approved):

flcl42:
@LukaszRozmej would you advice to also commit between ApplyMinerRewards and ProcessWidthdrawals?

LukaszRozmej:
> @LukaszRozmej would you advice to also commit between ApplyMinerRewards and ProcessWidthdrawals?

No as both are system transactions

LukaszRozmej:
This should only be a problem on AuRa chains

PR #8571, opened 2025-04-30
Handle zero bloom in decoded receipts

### LINKED ISSUE #8508: Account for zero bloom
**Description**
Erigon send receipt's bloom as zero bloom. Bloom can be regenerated locally and we do that, but if the bloom is null, not zero. This causes erigon to get disconnected during old receipts.

**Steps to Reproduce**
Full sync, monitor the peer allocation graph, wait for old receipts.

**Actual behavior**
Erigon got disconnected.

**Expected behavior**
Erigon do not get disconnected.

**Screenshots**
I'm out right now.

 - Operating System: Any
 - Version: Who knows.
 - Installation Method: Whatever ways you use.
 - Consensus Client: I sure hope all.

**Additional context**
- Quite an old isse.

**Logs**
- No log appear.

Hyodar:
Fixes #8508

## Changes

- Add a method `HasInvalidZeroBloom` to `ReceiptMessageDecoder` to detect null-like zero blooms based on the amount of log entries when decoding (see #8508 for context)
- Add a check to `ReceiptMessageDecoder`'s `Decode` to trigger bloom calculation in the case of an invalid zero bloom

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


Hyodar:
This seems like one way to fix the issue in that specific case, but zero is also a 'valid' value and not strictly null, so maybe not a valid approach.
Is there any other factor we may use for identifying when we should patch the receipt and when we shouldn't?

stdevMac:
Closing: the underlying issue #8508 was resolved on master by commit 4140b9bce ("fix(sync): accept receipts with zero bloom from peers"), which normalizes the bloom only when it is exactly `Bloom.Empty` with logs present — addressing the concern raised here that zero is also a valid value.

PR #8579, opened 2025-05-02
Rename peerdas getBlobsV2 metrics

mininny:
## Changes
This PR renames execution metrics for the new engine_getBlobsV2 rpc for peerdas.
This rename is for unifying the execution metric naming, shown in [this proposal](https://testinprod.notion.site/Proposal-for-Unified-EL-metrics-for-PeerDAS-1d28fc57f54680f2a3cbfe408d7db4b8?pvs=4).

To correctly identify the number of blobs in the blobpool, this PR modifies `AreBlobsAvailable` to `GetBlobCounts` which will return the number of blobs in the blobpool, then use that blob count to compare against the number of blobhashes requested.

Because this is naming change for metric that isn't public yet, so I think the naming changes won't be breaking change. 

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Merge.Plugin/Metrics.cs
+++ b/src/Nethermind/Nethermind.Merge.Plugin/Metrics.cs
@@ -39,5 +39,17 @@ namespace Nethermind.Merge.Plugin
         [GaugeMetric]
         [Description("Number of responses to engine_getBlobsV1 and engine_getBlobsV2 without all requested blobs")]
         public static int NumberOfGetBlobsFailures { get; set; }
+
+        [CounterMetric]
+        [Description("Number of Blobs requested by engine_getBlobsV2")]
+        public static int ExecutionGetBlobsRequestedFromCLTotal { get; set; }
+
+        [CounterMetric]
+        [Description("Number of Blobs requested by engine_getBlobsV2 that are present in the blobpool")]
+        public static int ExecutionGetBlobsRequestedFromCLHit { get; set; }
+
+        [GaugeMetric]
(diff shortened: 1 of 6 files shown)


src/Nethermind/Nethermind.Merge.Plugin/Metrics.cs
```diff
@@ -34,10 +34,22 @@ public static class Metrics
...
-        public static int NumberOfGetBlobsFailures { get; set; }
+        public static int GetBlobsRequestsFailureTotal { get; set; }
+
+        [CounterMetric]
+        [Description("Number of Blobs requested by engine_getBlobsV2")]
+        public static int GetBlobsRequestsTotal { get; set; }
+
+        [CounterMetric]
+        [Description("Number of Blobs requested by engine_getBlobsV2 that are present in the blobpool")]
+        public static int GetBlobsRequestsInBlobpoolTotal { get; set; }
```
marcindsobczak, line 49: We have similar metrics already - `GetBlobsRequestsTotal` is like `NumberOfRequestedBlobs` and `GetBlobsRequestsInBlobpoolTotal` is like `NumberOfSentBlobs`.
But I think it is reasonable to introduce new, standardized metrics for `engine_getBlobsV2`, while keep using old ones only for `engine_getBlobsV1`

marcindsobczak (approved):
Looks good!

mininny:
Thank you! :)

PR #8582, opened 2025-05-03
Consider dependency chains in optimal blob packing

benaadams:
## Changes

- Evaultate less (Perf good): When adding txs to consider during block production, stop adding from same address (don't add later nonces) when 
   - The sum of the dependency chain exceeds blob count or block gaslimit
   -  As soon as one of the txs fails tx filters e.g. gas below base fee
- Evaluate deeper (Bug fix): If dependency on blobs tx, only include blob tx in optimal list if also included parent tx

- [x] Optimization

#### Requires testing

- [x] Yes

- [x] Yes

#### Notes on testing

Two hive tests check for a suboptimal built block result; testing for a lower profit blocks (will submit fix to hive)

Have had added variant tests around that; including the variant hive thinks its testing (but it's taking into account profit, so changed priority fees to produce same output), and the max profit result of what it is testing.

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain.Test/TransactionSelectorTests.cs
+++ b/src/Nethermind/Nethermind.Blockchain.Test/TransactionSelectorTests.cs
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: 2022 Demerzel Solutions Limited
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
 // SPDX-License-Identifier: LGPL-3.0-only
 
 using System;
@@ -221,14 +221,43 @@ namespace Nethermind.Blockchain.Test
                     // as the packing rules should win
                     higherPriorityTransactionsSelected.Transactions.Shuffle(rnd);
                 }
+            }
+        }
 
-                static Transaction CreateBlobTransaction(Address address, PrivateKey key, UInt256 maxFee, int blobCount)
-                {
-                    return Build.A.Transaction.WithSenderAddress(address).WithType(TxType.Blob).WithNonce(1)
-                        .WithMaxFeePerGas(maxFee).WithMaxFeePerBlobGas(1).WithGasLimit(20)
-                        .WithBlobVersionedHashes([.. Enumerable.Range(0, blobCount).Select(i => new byte[1] { 0 })])
-                        .SignedAndResolved(key).TestObject;
-                }
+        private static Transaction CreateBlobTransaction(Address address, PrivateKey key, UInt256 maxFee, int blobCount)
+        {
+            return Build.A.Transaction.WithSenderAddress(address).WithType(TxType.Blob).WithNonce(1)
+                .WithMaxFeePerGas(maxFee).WithMaxFeePerBlobGas(1).WithGasLimit(20)
+                .WithBlobVersionedHashes([.. Enumerable.Range(0, blobCount).Select(i => new byte[1] { 0 })])
+                .SignedAndResolved(key).TestObject;
+        }
+
+        public static IEnumerable NoNonceGapWhenShardBlobTransactionsSelectedTestCases
+        {
+            get
+            {
+                ProperTransactionsSelectedTestCase higherPriorityTransactionsSelected = ProperTransactionsSelectedTestCase.Eip1559Default;
+                var accounts = higherPriorityTransactionsSelected.AccountStates;
+                accounts[TestItem.AddressA] = (1000000, 0);
+                higherPriorityTransactionsSelected.ReleaseSpec = Cancun.Instance;
+                higherPriorityTransactionsSelected.BaseFee = 1;
+                higherPriorityTransactionsSelected.Transactions =
+                [
+                    // This tx should be rejected in preference for the other 5 even though its fee is much higher
+                    CreateBlobTransaction(TestItem.AddressA, TestItem.PrivateKeyA, maxFee: 1000, blobCount: 5),
+                    CreateBlobTransaction(TestItem.AddressA, TestItem.PrivateKeyA, maxFee: 1000, blobCount: 5),
+                    // As total of other 5 below is higher
+                    CreateBlobTransaction(TestItem.AddressA, TestItem.PrivateKeyA, maxFee: 1000, blobCount: 1),
+                    CreateBlobTransaction(TestItem.AddressA, TestItem.PrivateKeyA, maxFee: 1000, blobCount: 1),
+                    CreateBlobTransaction(TestItem.AddressA, TestItem.PrivateKeyA, maxFee: 1000, blobCount: 1),
+                    CreateBlobTransaction(TestItem.AddressA, TestItem.PrivateKeyA, maxFee: 1000, blobCount: 1),
+                    CreateBlobTransaction(TestItem.AddressA, TestItem.PrivateKeyA, maxFee: 1000, blobCount: 1),
+                ];
+
+                higherPriorityTransactionsSelected.ExpectedSelectedTransactions.AddRange(
+                    higherPriorityTransactionsSelected.Transactions.Take(1));
+
+                yield return higherPriorityTransactionsSelected;
             }
         }
 
@@ -236,6 +265,7 @@ namespace Nethermind.Blockchain.Test
         [TestCaseSource(nameof(Eip1559LegacyTransactionTestCases))]
         [TestCaseSource(nameof(Eip1559TestCases))]
         [TestCaseSource(nameof(EnoughShardBlobTransactionsSelectedTestCases))]
+        [TestCaseSource(nameof(NoNonceGapWhenShardBlobTransactionsSelectedTestCases))]
         public void Proper_transactions_selected(ProperTransactionsSelectedTestCase testCase)
         {
             MemDb stateDb = new();
--- a/src/Nethermind/Nethermind.Blockchain.Test/TransactionSelectorTests.cs
+++ b/src/Nethermind/Nethermind.Blockchain.Test/TransactionSelectorTests.cs
@@ -227,10 +227,13 @@ namespace Nethermind.Blockchain.Test
         private static Transaction CreateBlobTransaction(Address address, PrivateKey key, UInt256 maxFee, int blobCount)
             => CreateBlobTransaction(address, key, maxFee, blobCount, nonce: 1);
 
-        private static Transaction CreateBlobTransaction(Address address, PrivateKey key, UInt256 maxFee, int blobCount, UInt256 nonce)
+        private static Transaction CreateBlobTransaction(Address address, PrivateKey key, UInt256 maxFee, int blobCount, UInt256 nonce, uint priority = 1)
         {
             return Build.A.Transaction.WithSenderAddress(address).WithType(TxType.Blob).WithNonce(nonce)
-                .WithMaxFeePerGas(maxFee).WithMaxFeePerBlobGas(1).WithGasLimit(20)
+                .WithMaxFeePerGas(maxFee)
+                .WithMaxFeePerBlobGas(2)
+                .WithMaxPriorityFeePerGas(priority)
+                .WithGasLimit(20)
                 .WithBlobVersionedHashes([.. Enumerable.Range(0, blobCount).Select(i => new byte[1] { 0 })])
                 .SignedAndResolved(key).TestObject;
         }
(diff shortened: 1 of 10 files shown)


src/Nethermind/Nethermind.Consensus/Producers/TxPoolTxSource.cs
```diff
@@ -320,32 +371,50 @@ internal static IEnumerable<Transaction> Order(IDictionary<AddressAsKey, Transac
...
                     // we take first transaction from sorting order, on first call: N4_P4 from B
-                    (Transaction tx, IEnumerator<Transaction> enumerator) = transactions.Min;
+                    (Transaction addedTx, (IEnumerator<Transaction> enumerator, long gasUsed, int blobCount)) = transactions.Min;
+
+                    transactions.Remove(addedTx);
+                    if (filter is not null && !filter.Invoke(addedTx))
+                    {
+                        // Cannot be added, so stop adding more txs from this sender
+                        continue;
+                    }
```
LukaszRozmej, line 397: We now invoke this filter 3x per transaction - seems wasteful?

src/Nethermind/Nethermind.Consensus/Producers/TxPoolTxSource.cs
```diff
@@ -301,13 +352,13 @@ private bool TryUpdateFeePerBlobGas(Transaction lightBlobTx, BlockHeader parent,
...
-        protected virtual IEnumerable<Transaction> GetOrderedTransactions(IDictionary<AddressAsKey, Transaction[]> pendingTransactions, IComparer<Transaction> comparer) =>
-            Order(pendingTransactions, comparer);
+        protected virtual IEnumerable<Transaction> GetOrderedTransactions(IDictionary<AddressAsKey, Transaction[]> pendingTransactions, IComparer<Transaction> comparer, long gasLimit, Func<Transaction, bool>? filter = null, int maxBlobs = 0) =>
+            Order(pendingTransactions, comparer, gasLimit, filter, maxBlobs);
 
         protected virtual IComparer<Transaction> GetComparer(BlockHeader parent, BlockPreparationContext blockPreparationContext)
             => _transactionComparerProvider.GetDefaultProducerComparer(blockPreparationContext);
 
-        internal static IEnumerable<Transaction> Order(IDictionary<AddressAsKey, Transaction[]> pendingTransactions, IComparer<Transaction> comparerWithIdentity)
+        internal static IEnumerable<Transaction> Order(IDictionary<AddressAsKey, Transaction[]> pendingTransactions, IComparer<Transaction> comparerWithIdentity, long gasLimit, Func<Transaction, bool>? filter = null, int maxBlobs = 0)
```
LukaszRozmej, line 361: Does it make sense to keep the filter nullable?

src/Nethermind/Nethermind.Consensus/Producers/TxPoolTxSource.cs
```diff
@@ -215,37 +217,86 @@ private static void ChooseBestBlobTransactions(
             in UInt256 baseFee,
             ArrayPoolList<Transaction> selectedBlobTxs)
         {
-            int size = leftoverCapacity + 1;
+            int maxSize = leftoverCapacity + 1;
             // The maximum total fee achievable with capacity
-            using ArrayPoolList<ulong> dpFeesPooled = new(capacity: size, count: size);
+            using ArrayPoolList<ulong> dpFeesPooled = new(capacity: maxSize, count: maxSize);
```
LukaszRozmej, line 222: what this does?

src/Nethermind/Nethermind.Consensus/Producers/TxPoolTxSource.cs
```diff
@@ -215,37 +217,86 @@ private static void ChooseBestBlobTransactions(
...
             Span<ulong> dpFees = dpFeesPooled.AsSpan();
 
-            using ArrayPoolBitMap isChosen = new(candidateTxs.Count * size);
+            using ArrayPoolBitMap isChosen = new(candidateTxs.Count * maxSize);
```
LukaszRozmej, line 225: Why this is better than `ArrayPoolList` with indexes of chosen transactions?

LukaszRozmej (commented):
In general I like quick bailout in `Order`.

To be honest I don't understand what is going on in `ChooseBestBlobTransactions` and why it needs to be so complicated.

src/Nethermind/Nethermind.Consensus/Producers/TxPoolTxSource.cs
benaadams: Changed to only run once

benaadams:
> To be honest I don't understand what is going on in `ChooseBestBlobTransactions` and why it needs to be so complicated.

Is the Knapsack problem https://en.wikipedia.org/wiki/Knapsack_problem

> Given a set of items, each with a weight and a value, determine which items to include in the collection so that the total weight is less than or equal to a given limit and the total value is as large as possible.

With the additional constraint (change in this PR); that if you include an item it needs to have included all the prior ones it is dependent on (i.e. txs from same sender with prior nonces)

src/Nethermind/Nethermind.Consensus/Producers/TxPoolTxSource.cs
benaadams: Best discovered sum of priority fees at a given capacity, so far

src/Nethermind/Nethermind.Consensus/Producers/TxPoolTxSource.cs
benaadams: Don't know what the chosen transactions are at this point.

The [Dynamic programming solution](https://en.wikipedia.org/wiki/Knapsack_problem#0-1_knapsack_problem) to the 0/1 Knapsack problem wants a 2d matrix of `int` fees; of size `candidateTxs.Count * maxCapacity`; and then you walk back through the matrix to workout the chosen txs having the best fees. However that's kinda huge e.g.

`10,000 candidate tx * 56 blob capacity * sizeof(int)` = 2.1MiB

So instead I'm using a 1d array of fees of size `maxCapacity` and a bit array of `candidateTxs.Count * maxCapacity` to say if picked at that capacity, then walk back through th picked

`56 blob capacity * sizeof(int)` = 224 bytes
`10,000 candidate tx * 56 blob capacity * sizeof(bit)` = 68KiB

So 68KiB rather than 2.1MiB in the standard approach

src/Nethermind/Nethermind.Blockchain.Test/TransactionSelectorTests.cs
```diff
@@ -221,13 +221,243 @@ public static IEnumerable EnoughShardBlobTransactionsSelectedTestCases
...
+
+                    UInt256 nonce = 1;
+                    AddTxs(txCount: 5, blobsPerTx: 5, account: 0, txs, ref nonce);
+                    nonce = 1;
+                    AddTxs(txCount: 5, blobsPerTx: 1, account: 1, txs, ref nonce);
+
+                    blobTxs.Transactions = txs;
+
+                    blobTxs.ExpectedSelectedTransactions.AddRange(blobTxs.Transactions.Take(1));
+                    blobTxs.ExpectedSelectedTransactions.AddRange(blobTxs.Transactions.Skip(5).Take(1));
```
flcl42, line 292: Probably "2 txs with 6 blobs are better than 5 txs with 5 blobs" is not a typical case. Can you also add a case with bigger max fee (like we usually have big gas price and low blob gas price)

src/Nethermind/Nethermind.Blockchain.Test/TransactionSelectorTests.cs
```diff
@@ -221,13 +221,243 @@ public static IEnumerable EnoughShardBlobTransactionsSelectedTestCases
...
+
+                    UInt256 nonce0 = 1;
+                    AddTxs(txCount: 2, blobsPerTx: 2, account: 0, txs, ref nonce0, priority: 1);
+                    UInt256 nonce1 = 1;
+                    AddTxs(txCount: 1, blobsPerTx: 5, account: 1, txs, ref nonce1, priority: 1);
+                    AddTxs(txCount: 3, blobsPerTx: 1, account: 1, txs, ref nonce1, priority: 1);
+
+                    blobTxs.Transactions = txs;
+
+                    blobTxs.ExpectedSelectedTransactions.AddRange(blobTxs.Transactions.Take(2));
```
flcl42, line 389: Why 2 txs with 4 blobs are better than 2 txs with 6 blobs?

src/Nethermind/Nethermind.Consensus/Producers/TxPoolTxSource.cs
```diff
@@ -215,37 +214,86 @@ private static void ChooseBestBlobTransactions(
...
                 }
+
+                // How many blobs does this tx actually consume?
                 int blobCount = tx.GetBlobCount();
-                // Use actual gas used when available as the tx may be using over-estimated gaslimit
-                ulong feeValue = (ulong)premiumPerGas * (ulong)(tx.SpentGas);
+                // If this tx has explicit dependencies (i.e. it requires k prior blobs
+                // from the *same address* to be in the block before it), include them here.
+                // We'll need a capacity of blobDependenciesCount slots *plus* its own blobCount.
+                int blobCapacityNeeded = tx.BlobDependenciesCount + blobCount;
```
flcl42, line 242: Can be a dictionary<tx, int>, uglier but without adding tx field needed only here

src/Nethermind/Nethermind.Blockchain.Test/TransactionSelectorTests.cs
benaadams: AddTxs bumps priority fee so are higher with more blobs (`priority fee = blobs * 2`) if not specified, so 5 + 1 would be better in this case; have another test where priorty fees are all the same where it switches (and is what the hive test is checking and saying it should be 5+1 rather than 5 x 1; which as you point out isn't optimal)

src/Nethermind/Nethermind.Consensus/Producers/TxPoolTxSource.cs
benaadams: Problematic when testing tx pools with 250k txs? That dictionary will be massive

src/Nethermind/Nethermind.Blockchain.Test/TransactionSelectorTests.cs
benaadams: Though why the algo chooses it is "value" (priority fees) are the same and is remaining space to add more; just aren't any more

src/Nethermind/Nethermind.Consensus/Producers/TxPoolTxSource.cs
benaadams: Actually could return with enumerator; as pair though would be more ugly 🤔

benaadams: Removed the extra property via two sort methods

marcindsobczak (approved):
Looks good.

One potential issue:
We have full blob txs saved on disc and only light equivalences in memory.
We used to pick blob txs first and then load from disc only txs selected for inclusion.
Right now we are loading many blob txs as candidates.
Our full blob tx cache is 512 by default so not an issue with less than 512 blob txs in the pool, but with thousands it will be an issue - we will load hundreds or thousands txs from disc and probably we will just time out

Maybe we should add to candidates only in-memory form of blob txs and load from db at the end, when txs for inclusion are picked?

Or add to candidates up to e.g. 10x block blob space and skip worse txs?

LukaszRozmej (approved):
One last thing:

Do we need to call `TryUpdateFeePerBlobGas` for every blob transaction? I think not? It is not super cheap call.

LukaszRozmej:
> Looks good.
> 
> One potential issue: We have full blob txs saved on disc and only light equivalences in memory. We used to pick blob txs first and then load from disc only txs selected for inclusion. Right now we are loading many blob txs as candidates. Our full blob tx cache is 512 by default so not an issue with less than 512 blob txs in the pool, but with thousands it will be an issue - we will load hundreds or thousands txs from disc and probably we will just time out
> 
> Maybe we should add to candidates only in-memory form of blob txs and load from db at the end, when txs for inclusion are picked?
> 
> Or add to candidates up to e.g. 10x block blob space and skip worse txs?

I consider this to be critical!

benaadams:
> Or add to candidates up to e.g. 10x block blob space and skip worse txs?

Added x5 cap; as having the actual blob is part of the filtering criteria

benaadams:
> Do we need to call `TryUpdateFeePerBlobGas` for every blob transaction? I think not? It is not super cheap call.

Won't be that many (now with cap); and isn't that expensive

LukaszRozmej:
> > Do we need to call `TryUpdateFeePerBlobGas` for every blob transaction? I think not? It is not super cheap call.
> 
> Won't be that many (now with cap); and isn't that expensive

1. Still why call it n times when we can call it once? Or do we actually need to call it n times?
2. Cap is still suboptimal, why load from disk even 1 unnecessary blob?

I thought we are optimizing it to the max :D

benaadams:
> 2\. Cap is still suboptimal, why load from disk even 1 unnecessary blob?

Because if we can't load it from disk then we can't include the blob; is one of the filter criteria and then the selection of everything else after will be wrong. Optimizing for profit not performance

benaadams:
> 1. Still why call it n times when we can call it once? Or do we actually need to call it n times?

Oh weird; thought it was per tx; but is for block

LukaszRozmej:
> > 2. Cap is still suboptimal, why load from disk even 1 unnecessary blob?
> 
> Because if we can't load it from disk then we can't include the blob; is one of the filter criteria and then the selection of everything else after will be wrong. Optimizing for profit not performance

Ok but in what scenarios, we can't load from disk? Those would probably be exceptional/catastrophic scenarios, in which we can build a less profitable block. If they are expected scenarios than I can agree with you.

benaadams:
> Ok but in what scenarios, we can't load from disk?

If cache is 512 and its loading blobs for the most profitable blob txs, should be ok? Would only be issue if/when blobs per block goes over 102 blobs?

marcindsobczak:
If there is no bug, we should always be able to load full tx (with blobs). Failure at this point means exception/bug/whatever.
But 5x cap is way better than no cap - in worst case (nothing useful in cache) we will just load 5x more txs than necessary

LukaszRozmej:
Can we have cap and delayed blob loading?
Do we even need to load blobs from disk to include blob transactions in the block?

benaadams:
> Do we even need to load blobs from disk to include blob transactions in the block?

I assume so because hive broke when didn't switch the blobs to actual ones

benaadams:
> Can we have cap and delayed blob loading?

Sure; @kamilchodola is going to do some testing on perfnet and will look at follow up

PR #8590, opened 2025-05-05
Fix hive

flcl42:
Fixes some hive tests

## Changes

Covers the following errors:
- Out of gas execution or failure of withdrawals/consolidations contract
- No contract deployed

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No (covered by hive)


src/Nethermind/Nethermind.Consensus/ExecutionRequests/ExecutionRequestProcessor.cs
```diff
@@ -131,25 +133,37 @@ private void ReadRequests(Block block, IWorldState state, IReleaseSpec spec, Add
 
         int requestsByteSize = isWithdrawalRequests ? ExecutionRequestExtensions.WithdrawalRequestsBytesSize : ExecutionRequestExtensions.ConsolidationRequestsBytesSize;
 
-        if (!(isWithdrawalRequests ? spec.WithdrawalRequestsEnabled : spec.ConsolidationRequestsEnabled) || !state.AccountExists(contractAddress))
+        if (!(isWithdrawalRequests ? spec.WithdrawalRequestsEnabled : spec.ConsolidationRequestsEnabled))
```
LukaszRozmej, line 136: can we remove `isWithdrawalRequests` and change all those `if's` to come from method parameters?

flcl42: Nice idea

src/Nethermind/Nethermind.Consensus/ExecutionRequests/ExecutionRequestProcessor.cs
```diff
@@ -125,31 +130,35 @@ private void DecodeDepositRequest(LogEntry log, Span<byte> buffer)
...
-            return;
+        if (!state.HasCode(contractAddress))
+        {
+            throw new InvalidBlockException(block, contractEmptyError);
+        }
 
         CallOutputTracer tracer = new();
 
-        _transactionProcessor.Execute(isWithdrawalRequests ? _withdrawalTransaction : _consolidationTransaction, new BlockExecutionContext(block.Header, spec), tracer);
+        _transactionProcessor.Execute(systemTx, new BlockExecutionContext(block.Header, spec), tracer);
```
LukaszRozmej, line 143: btw we don't need tracer, we could just use TransactionResult that is being returned

MarekM25 (commented):
So far it looks good

src/Nethermind/Nethermind.Consensus/ExecutionRequests/ExecutionRequestProcessor.cs
flcl42: We need ReturnValue

LukaszRozmej: still you can use error code

src/Nethermind/Nethermind.Core/Transaction.cs
```diff
@@ -257,6 +257,8 @@ public string ToString(string indent)
 
         public bool MayHaveNetworkForm => Type is TxType.Blob;
 
+        public bool IsSystemCall { get; set; } = false;
```
LukaszRozmej, line 260: use `TransactionExtensions.IsSystem` extension method

flcl42: The problem is that IsSystem != IsSystemCall. Gas estimation uses system tx by default and expect gas spendings to be the same as for regular tx. By introducing IsSystemCall we also prevent potential problems for chains that already use system transactions and it's dangerous to change gas calculations for them. Additionally, what's is real expected is "system call" but not real transaction.

flcl42: Hm maybe instead of field we can make a systemcallliketransaction type and check for it

flcl42:
Waiting for hive tests update. Prague contracts now require consolidations/withdrawals contracts in pre-state

LukaszRozmej (approved):

PR #8591, opened 2025-05-06
Delay resolving to full blob when picking

benaadams:
## Changes

- Don't resolve to full blob in candidate phase but delay until blob is being added to block

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Producers/TxPoolTxSource.cs
+++ b/src/Nethermind/Nethermind.Consensus/Producers/TxPoolTxSource.cs
@@ -74,7 +74,10 @@ namespace Nethermind.Consensus.Producers
 
                 foreach (Transaction blobTx in PickBlobTxsBetterThanCurrentTx(selectedBlobTxs, tx, comparer))
                 {
-                    yield return blobTx;
+                    if (TryGetFullBlobTx(blobTx, out Transaction fullBlobTx))
+                    {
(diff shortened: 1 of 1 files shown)


LukaszRozmej (approved):

src/Nethermind/Nethermind.Consensus/Producers/TxPoolTxSource.cs
```diff
@@ -87,11 +90,28 @@ public IEnumerable<Transaction> GetTransactions(BlockHeader parent, long gasLimi
...
+                    if (ResolveBlob(blobTx, out Transaction fullBlobTx))
+                    {
+                        yield return fullBlobTx;
+                    }
                 }
             }
 
             if (_logger.IsDebug) _logger.Debug($"Potentially selected {selectedTransactions} out of {checkedTransactions} pending transactions checked.");
+
+            bool ResolveBlob(Transaction blobTx, out Transaction fullBlobTx)
```
marcindsobczak, line 102: nitpick: `TryResolveBlob` would better reflect its bool return type. Furthermore, since it doesn't actually resolve anything, `TryLoadFullBlobTx` would be more accurate. As the PR is already merged and this is a minor detail, feel free to ignore.

marcindsobczak (commented):
Looks great!

Do we need `maxBlobsToConsider` any more?

PR #8593, opened 2025-05-06
Refactor/separate pruning trieestore

asdacap:
- Create a `RawScopedTrieStore` to be used by `PatriciaTrie` separate from `TrieStore`.
- `TrieStore` contains a lot of logic that deals with pruning and reorg boundary while also being used for general purpose hash calculation and snap sync, which does not need the dirty nodes cache and reorg boundary logic.
- This make the code path hard to follow and because it can be heavy to construct, `SnapProvider` uses a pool of it.
- This PR separate the general use case to another class, leaving `TrieStore` to deal with main processing trie handling.

## Changes

- Create `RawScopedTrieStore` which take only an `INodeStorage`. 
- Remove `IScopedTrieStore.Set`, as it is only used by healing code. Healing code write directly to `INodeStorage`.
- Remove the pool of `TrieStore` in `SnapProvider`.

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- [x] Mainnet can sync.


src/Nethermind/Nethermind.Trie/Pruning/RawScopedTrieStore.cs
```diff
@@ -0,0 +1,62 @@
...
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using Nethermind.Core;
+using Nethermind.Core.Crypto;
+
+namespace Nethermind.Trie.Pruning;
+
+public class RawScopedTrieStore(INodeStorage nodeStorage, Hash256 address) : IScopedTrieStore
```
LukaszRozmej, line 11: I've seen it used a lot, so maybe:

```suggestion
public class RawScopedTrieStore(INodeStorage nodeStorage, Hash256? address = null) : IScopedTrieStore
```

src/Nethermind/Nethermind.Synchronization/SnapSync/SnapProvider.cs
```diff
@@ -91,54 +90,46 @@ public AddRangeResult AddAccountRange(
         {
             if (accounts.Count == 0)
                 throw new ArgumentException("Cannot be empty.", nameof(accounts));
-            ITrieStore store = _trieStorePool.Get();
-            try
-            {
-                StateTree tree = new(store.GetTrieStore(null), _logManager);
+            StateTree tree = new(new RawScopedTrieStore(_nodeStorage, null), _logManager);
```
LukaszRozmej, line 93: Do we need to create instance per response then? Can't we have one `RawScopedTrieStore` as a field of the class?

LukaszRozmej: So maybe reuse the one with `null` address?

LukaszRozmej (approved):

PR #8602, opened 2025-05-08
Fix E2ESyncTest

asdacap:
- E2ESyncTests fails after pectra as post merge EIP is activated. 
- ParentBeaconBlockRoot must not be null so that header is encoded correctly.
- Now the `ITimestamper` changed so that any EIP post merge is not activated in pre merge tests.

#### Requires testing

- [ ] Yes
- [X] No

- [X] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Synchronization.Test/E2ESyncTests.cs
+++ b/src/Nethermind/Nethermind.Synchronization.Test/E2ESyncTests.cs
@@ -114,17 +114,7 @@ public class E2ESyncTests(E2ESyncTests.DbMode dbMode, bool isPostMerge)
         // TODO: Need to double check which code part does not pass in timestamp from header.
         spec.Parameters.Eip658Transition = 0;
 
-        if (!isPostMerge)
-        {
-            // Disable as the built block always don't have withdrawal (it came from engine) so it fail validation.
-            spec.Parameters.Eip4895TransitionTimestamp = null;
-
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Synchronization.Test/E2ESyncTests.cs
```diff
@@ -172,12 +163,19 @@ private IContainer CreateNode(PrivateKey nodeKey, Action<IConfigProvider, ChainS
                 ))
                 .AddDecorator<ITestEnv, PostMergeTestEnv>()
                 ;
-
+        }
+        else
+        {
+            // So that any EIP after the merge is not activated.
+            DateTime mergeTime = new DateTime(2022, 9, 15, 13, 45, 0, DateTimeKind.Utc);
+            ManualTimestamper timestamper = new ManualTimestamper(mergeTime.AddDays(-1));
```
LukaszRozmej, line 171: Maybe abstract in `static ManualTimestamper.CreatePreMerge`?
Might be useful somewhere else in the future.

LukaszRozmej (approved):

src/Nethermind/Nethermind.Synchronization.Test/E2ESyncTests.cs
asdacap: Sure

PR #8603, opened 2025-05-08
HealthChecks: Track CL calls from fcu & np methods

deffrian:
Fixes Closes Resolves #
In Optimism CL we call engine api directly avoiding json-rpc. This causes Nethermind to incorrectly identify CL as unavailable

## Changes

- Now all fcus & nps are tracked directly without json-rpc proxy

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


src/Nethermind/Nethermind.HealthChecks/ClHealthRequestsTracker.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            if (logger.IsWarn)
+                logger.Warn("Not receiving ForkChoices from the consensus client that are required to sync.");
+        }
+    }
+
+    private bool IsRequestTooOld(DateTime now, DateTime requestTime)
+    {
+        int diff = (int)(Math.Floor((now - requestTime).TotalSeconds));
+        return diff > maxIntervalClRequestTime;
```
emlautarom1, line 56: Prefer to use `TimeSpan` over `int` for more control over durations (requires changing the `maxIntervalClRequestTime` parameter type too):

```suggestion
        TimeSpan diff = (now - requestTime).Duration();
        return diff > maxIntervalClRequestTime;
```

src/Nethermind/Nethermind.HealthChecks/IClHealthTracker.cs
```diff
@@ -0,0 +1,11 @@
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System;
+
+namespace Nethermind.HealthChecks;
+
+public interface IClHealthTracker : IAsyncDisposable
```
emlautarom1, line 8: Not sure about having `IAsyncDisposable` in the interface itself since "being disposable" is up to each implementation.

src/Nethermind/Nethermind.HealthChecks/HealthChecksPlugin.cs
```diff
@@ -140,7 +141,12 @@ public Task InitRpcModules()
 
             _nodeHealthService = new NodeHealthService(_api.SyncServer,
                 _api.MainProcessingContext!.BlockchainProcessor, _api.BlockProducerRunner!, _healthChecksConfig, _api.HealthHintService!,
-                _api.EthSyncingInfo!, _api.RpcCapabilitiesProvider, _api, drives, _initConfig.IsMining);
+                _api.EthSyncingInfo!, _engineRequestsTracker, _api.SpecProvider!.TerminalTotalDifficulty, drives, _initConfig.IsMining);
+
+            if (_mergeConfig.Enabled)
+            {
+                _engineRequestsTracker.StartAsync();
```
emlautarom1, line 148: Missing `await` here? (requires making the function `async`)

```suggestion
                await _engineRequestsTracker.StartAsync();
```

src/Nethermind/Nethermind.HealthChecks/ClHealthRequestsTracker.cs
```diff
@@ -0,0 +1,74 @@
...
+        _timer.Change(Timeout.Infinite, 0);
+
+        return Task.CompletedTask;
+    }
+
+    public async ValueTask DisposeAsync()
+    {
+        await StopAsync();
+        await _timer.DisposeAsync();
+    }
```
emlautarom1, line 42: I think we can inline `StopAsync` and make it sync:

```suggestion
    public async ValueTask DisposeAsync()
    {
        _timer.Change(Timeout.Infinite, 0);
        await _timer.DisposeAsync();
    }
```

Even then, what is the point of `.Change` if it's going to be immediately disposed? I might be ignoring some `Timer` best practices though.

src/Nethermind/Nethermind.HealthChecks/NodeHealthService.cs
```diff
@@ -24,16 +23,16 @@ public class CheckHealthResult
         public IEnumerable<string> Errors { get; set; }
     }
 
-    public class NodeHealthService : INodeHealthService
+    public class NodeHealthService : INodeHealthService, IAsyncDisposable
```
emlautarom1, line 26: Let's remove the implementation of `IAsyncDisposable`:

- Since we're injecting `IClHealthTracker` in the constructor, this class is not responsible for managing the lifetime of the dependency.
- The place where we inject a concrete implementation is `HealthChecksPlugin.InitRpcModules`, and that class already disposes of the implementation, so right now we're disposing twice. In this case it's fine sine we're essentially disposing a `Timer` twice.

src/Nethermind/Nethermind.HealthChecks.Test/ClHealthTrackerTests.cs
```diff
@@ -0,0 +1,53 @@
...
+
+namespace Nethermind.HealthChecks.Test;
+
+public class ClHealthTrackerTests
+{
+    [Test]
+    public void ClHealthRequestsTracker_multiple_requests()
+    {
+        ManualTimestamper timestamper = new(DateTime.Parse("18:23:00"));
+        ClHealthRequestsTracker healthTracker = new(timestamper, 300, NullLogger.Instance);
```
emlautarom1, line 18: Missing dispose (surprisingly not detected by the IDE or compiler):

```suggestion
        await using ClHealthRequestsTracker healthTracker = new(timestamper, 300, NullLogger.Instance);
```

src/Nethermind/Nethermind.Api/IEngineRequestsTracker.cs
```diff
@@ -0,0 +1,23 @@
...
+public interface IEngineRequestsTracker
+{
+    void OnForkchoiceUpdatedCalled();
+    void OnNewPayloadCalled();
+}
+
+public class NullEngineRequestsTracker : IEngineRequestsTracker
+{
+    public static readonly NullEngineRequestsTracker Instance = new();
+
```
emlautarom1, line 15: Add a private parameterless constructor to prevent future missuse:

```suggestion
    private NullEngineRequestsTracker() { }
    public static readonly NullEngineRequestsTracker Instance = new();
```

src/Nethermind/Nethermind.Merge.Plugin.Test/BaseEngineModuleTests.cs
```diff
@@ -161,6 +162,7 @@ protected IEngineRpcModule CreateEngineModule(MergeTestBlockchain chain, ISyncCo
             new ExchangeTransitionConfigurationV1Handler(chain.PoSSwitcher, chain.LogManager),
             new ExchangeCapabilitiesHandler(capabilitiesProvider, chain.LogManager),
             new GetBlobsHandler(chain.TxPool),
+            NullEngineRequestsTracker.Instance,
```
emlautarom1, line 165: Would a substitute be enough?

```suggestion
            Substitute.For<IEngineRequestsTracker>(),
```

src/Nethermind/Nethermind.Merge.Plugin.Test/MergePluginTests.cs
```diff
@@ -81,6 +81,7 @@ private IContainer BuildContainer(IConfigProvider? configProvider = null)
                     api.TransactionComparerProvider!,
                     ctx.Resolve<IBlocksConfig>(),
                     api.LogManager!);
+                api.EngineRequestsTracker = NullEngineRequestsTracker.Instance;
```
emlautarom1, line 84: Same as in `BaseEngineModuleTests`, would a substitute be enough?

```suggestion
                api.EngineRequestsTracker = Substitute.For<IEngineRequestsTracker>();
```

Given that `NullEngineRequestsTracker` is only used in tests, if we replace all its usages with substitutes we can remove the class entirely.

src/Nethermind/Nethermind.Runner/Ethereum/Modules/NethermindRunnerModule.cs
```diff
@@ -13,6 +13,7 @@
 using Nethermind.Core;
 using Nethermind.Core.Container;
 using Nethermind.Core.Specs;
+using Nethermind.HealthChecks;
```
emlautarom1, line 16: Unused imports

emlautarom1 (changes requested):
A couple of things that could be removed but overall LGTM, I don't see any potential issue.

src/Nethermind/Nethermind.Api/IEngineRequestsTracker.cs
emlautarom1: This change might be irrelevant if we can remove the class altogether.

src/Nethermind/Nethermind.HealthChecks/ClHealthRequestsTracker.cs
deffrian: We cant change it's type because of backwards compatibility

src/Nethermind/Nethermind.HealthChecks/HealthChecksPlugin.cs
deffrian: This should be fire and forget

src/Nethermind/Nethermind.HealthChecks/ClHealthRequestsTracker.cs
deffrian: As I remember it will still work for some time. Anyways, let's keep it like it was before

src/Nethermind/Nethermind.HealthChecks/HealthChecksPlugin.cs
emlautarom1: Then let's make it explicit with `_ = _engineRequestsTracker.StartAsync(); // Fire and forget`

src/Nethermind/Nethermind.HealthChecks/ClHealthRequestsTracker.cs
emlautarom1: Without changing the `maxIntervalClRequestTime` parameter type we can do the following:

```suggestion
        TimeSpan diff = (now - requestTime).Duration();
        return diff > TimeSpan.FromSeconds(maxIntervalClRequestTime);
```

src/Nethermind/Nethermind.Api/IEngineRequestsTracker.cs
```diff
@@ -0,0 +1,10 @@
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+namespace Nethermind.Api;
+
+public interface IEngineRequestsTracker
+{
+    void OnForkchoiceUpdatedCalled();
```
Scooletz, line 8: I think that the introduction of this tracker is not needed. Its methods are called in the `EngineRpcModule` which is constructed in each plugin according to its needs. For `Optimism`, one could introduce a decorator for the original components like `IForkchoiceUpdatedHandler` and pass it down to the `EngineRpcModule`. This would keep the original shape of the `EngineRpcModule`. The seam seems to be ready to be extended without introduction of additional steps.

Scooletz (commented):
I'd use a decorator in the actual `Optimism` plugin instead of introducing a new entity of `IEngineRequestsTracker`. More in the file specific comments.

emlautarom1 (approved):
I think it's good enough, but I would explore the decorator approach as suggested.

Scooletz (approved):
We discussed offline the introduced capability and its placement. The main point that I missed earlier is `EngineRpcModule` initialization and its placement in various places (Taiko, Optimism, MergePlugin) that later is passed to `RpcModuleProvider.RegisterSingle`. Having this in mind 👍

PR #8604, opened 2025-05-08
Refactor/Explicit message serializer registration

asdacap:
- Remove `IMessageSerializationService.Register` passing the serializers via constructor instead.
- Explicitly register serializers in `NetworkModule` and `DiscoveryModule` instead of via assembly scanning.
- Added unit test that scan assembly in case of missed serializer.

- [X] (very slight) Optimization
- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- [x] Mainnet can sync.


src/Nethermind/Nethermind.Runner.Test/EthereumRunnerTests.cs
```diff
@@ -244,6 +248,36 @@ public async Task Smoke_CanResolveAllSteps((string file, ConfigProvider configPr
...
             api.Context.Resolve<IRpcModuleProvider>();
+            api.Context.Resolve<IMessageSerializationService>();
+
+            // The following types should not have a global unnamed singleton registration. This is because
+            // They are ambiguous by nature. Eg: For `IProtectedPrivateKey`, is it signer key or node key?
+            // Consider wrapping them in an type that is clearly global eg: `IWorldStateManager.GlobalWorldState`
+            // or using a named registration, or create an explicit child lifetime for that particular instance.
+            HashSet<Type> bannedTypeForRootScope =
+            [
+                typeof(IWorldState),
```
LukaszRozmej, line 259: Maybe add:
`ITransactionProcessor`
`IVirtualMachine`
`IDb`

ect?

LukaszRozmej (approved):

PR #8606, opened 2025-05-08
HexPrefixed turned into NibblePath

Scooletz:
Tentative. WIP

This PR reduces allocation and encoding/decoding overhead of `HexPrefix` by turning it into a `NibblePath`. `NibblePath` construct is heavily inspired by , making it aligned with the one proposed in Paprika's [NibblePath](https://github.com/NethermindEth/Paprika/blob/main/src/Paprika/Data/NibblePath.cs) 

## Changes

- replace `byte[]` representing the `HexPrefix` with an explicit construct of `NibblePath`
- use dense, 2 nibbles per byte encoding for `NibblePath`
- provide `Singles` that cache short paths (useful for extensions mostly)

- [x] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

> New path encoding that decreases the memory needed to keep `TrieNode`s in memory. Should result in smaller memory footprint and higher number of nodes that can be kept in memory.


src/Nethermind/Nethermind.Trie/NibblePath.cs
```diff
@@ -358,40 +385,85 @@ public Ref Slice(int start)
...
+            return new(ref Unsafe.Add(ref _span, (_odd + start) / 2),
+                (byte)((start & 1) ^ _odd), (byte)length);
         }
 
         public string ToHexString()
         {
-            throw new NotImplementedException();
+            const int prefixLength = 2;
+
+            // TODO: optimize
```
benaadams, line 399: `String.Create<TState>(Int32, TState, SpanAction<Char,TState>)` https://learn.microsoft.com/en-us/dotnet/api/system.string.create?view=net-9.0#system-string-create-1(system-int32-0-system-buffers-spanaction((system-char-0)))

Scooletz:
Replaced by smaller #8625.

src/Nethermind/Nethermind.Trie/NibblePath.cs
Scooletz: Will update the other one.

PR #8609, opened 2025-05-09
Feature/clef tx

### LINKED ISSUE #8546: Support eth_sendTransaction with Clef enabled
I'm trying to use the eth_sendTransaction RPC method while the Nethermind node is configured to use Clef for transaction signing. However, the request currently fails or is not processed as expected.
The error message: `Can only sign without passphrase when account is unlocked`.

**Additional context**

ak88:
Fixes #8546

Adds clef support in json RPC , when it's configured.

Also fixes an issue encoding issue with the existing eth_sign, where the prefix is wrongly escaped.

## Changes

fixes character escape issue with the prefix "\x19Ethereum Signed Message:\n" used in eth_sign 
eth_sign and eth_sendTransaction now uses clef as the wallet when its configured. 

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_


src/Nethermind/Nethermind.Wallet/IWallet.cs
```diff
@@ -18,6 +20,23 @@ public interface IWallet
...
+            {
+                throw new CryptographicException($"Failed to sign tx {tx.Hash} using the {tx.SenderAddress} address.");
+            }
+
+            tx.Signature.V = tx.Type == TxType.Legacy ? tx.Signature.V + 8 + 2 * chainId : (ulong)(tx.Signature.RecoveryId + 27);
+        }
+        Signature SignMessage(byte[] message, Address address)
+        {
+            const string signatureTemplate = "\x19" + "Ethereum Signed Message:\n{0}{1}";
+            string signatureText = string.Format(signatureTemplate, message.Length, message);
```
LukaszRozmej, line 37: use string interpolation

src/Nethermind/Nethermind.Runner/configs/mainnet.json
```diff
@@ -35,5 +35,9 @@
   },
   "Merge": {
     "Enabled": true
+  },
+    "Mining": {
+    "Enabled": true,
+    "Signer":"http://localhost:8550"
```
LukaszRozmej, line 41: Revert? We don't want to change mainnet config

src/Nethermind/Nethermind.ExternalSigner.Plugin/ClefSigner.cs
```diff
@@ -63,56 +51,27 @@ public Signature Sign(Hash256 message)
...
 
-    private static async Task<Address> GetSignerAddress(IJsonRpcClient rpcClient, Address? blockAuthorAccount)
+    private static Address GetSignerAddress(ClefWallet clefWallet, Address? blockAuthorAccount)
     {
-        var accounts = await rpcClient.Post<string[]>("account_list") ?? throw new InvalidOperationException("Remote signer 'account_list' response is invalid.");
+        Address[] accounts = clefWallet.GetAccounts();
         if (accounts.Length == 0) throw new InvalidOperationException("Remote signer has not been configured with any signers.");
         return blockAuthorAccount is not null
-            ? accounts.Any(a => new Address(a).Bytes.SequenceEqual(blockAuthorAccount.Bytes))
+            ? accounts.Any(a => a.Bytes.SequenceEqual(blockAuthorAccount.Bytes))
```
LukaszRozmej, line 65: Addess doesn't have `==` operator overloaded?

src/Nethermind/Nethermind.ExternalSigner.Plugin/ClefWallet.cs
```diff
@@ -0,0 +1,191 @@
...
+                    transactionForRpc = new EIP1559TransactionForRpc(transaction);
+                    break;
+                case TxType.Blob:
+                    transactionForRpc = new BlobTransactionForRpc(transaction);
+                    break;
+                case TxType.SetCode:
+                    transactionForRpc = new SetCodeTransactionForRpc(transaction);
+                    break;
+                default:
+                    throw new NotImplementedException($"Cannot send unknown tx type '{transaction.Type}'");
```
LukaszRozmej, line 143: Refactor factory out from `TransactionJsonConverter` where we can support any tx type

LukaszRozmej (changes requested):

src/Nethermind/Nethermind.ExternalSigner.Plugin/ClefSigner.cs
ak88: changed to ==

src/Nethermind/Nethermind.ExternalSigner.Plugin/ClefWallet.cs
ak88: fixed

src/Nethermind/Nethermind.Runner/configs/mainnet.json
ak88: forgot about that one :)

src/Nethermind/Nethermind.Wallet/IWallet.cs
ak88: Changed it to use explicit unicode escape sequence

ak88:
@LukaszRozmej found another bug in eth_sign, since it was not returning the V in the signature. 
Changed signature of the method to string, but since it's a rarely used method i think its fine.

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
```diff
@@ -269,28 +269,24 @@ public ResultWrapper<byte[]> eth_getCode(Address address, BlockParameter? blockP
                     : []);
     }
 
-    public ResultWrapper<Memory<byte>> eth_sign(Address addressData, byte[] message)
+    public ResultWrapper<string> eth_sign(Address addressData, byte[] message)
```
LukaszRozmej, line 272: why not keep it as Memory<byte> or even ReadOnlyMemory<byte> it should be serialized the same and we avoid allocation?

LukaszRozmej (approved):

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
ak88: Internally `Signature` keeps R and S in a single byte array and V as a separate byte. So if we have to concatenate somehow or allocate a new array it doesn't make much of a difference.

LukaszRozmej: hm... we could keep it internally in one array then.
We could also return `ReadOnlySequence<byte>`
But overall probably not worth it.

ak88: Will do a separate PR so we can get this merged.

PR #8614, opened 2025-05-12
Fix trie tests

asdacap:
- Fix trie tests failing as previously `TrackPastKeys` was not enabled in test.
- Rename `IfBlockOlderThan` to `EveryNBlock`.

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing 

- Reproducable with `Repeat(10)`.

### DIFF
--- a/src/Nethermind/Nethermind.Serialization.Json/StreamPipeWriter.cs
+++ b/src/Nethermind/Nethermind.Serialization.Json/StreamPipeWriter.cs
@@ -243,11 +243,13 @@ public sealed class CountingStreamPipeWriter : CountingWriter
 
     private void ReturnSegmentUnsynchronized(BufferSegment segment)
     {
+        /*
         segment.Reset();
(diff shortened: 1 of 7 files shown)


src/Nethermind/Nethermind.Serialization.Json/StreamPipeWriter.cs
```diff
@@ -244,6 +244,7 @@ private BufferSegment CreateSegmentUnsynchronized()
     private void ReturnSegmentUnsynchronized(BufferSegment segment)
     {
         segment.Reset();
+        Interlocked.MemoryBarrier();
```
LukaszRozmej, line 247: Why needed?

asdacap: Heck I don't know. I'm guessing something within `segment.Reset()` run after `_bufferSegmentPool.Push`.

PR #8615, opened 2025-05-12
Fix long running test fuzzing OOM

LukaszRozmej:
Fixes https://github.com/holiman/goevmlab/issues/173

## Changes

- Use console over TestContext in EthereumTests as they can be run from Test.Runner and not flushed for a long time

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No


benaadams (approved):

flcl42 (approved):

holiman (approved):
Seems to work fine!

PR #8232, opened 2025-02-20
Reintroduce missing 4844 RPC fields (GasPrice and V)

LukaszRozmej:
Fixes #8230 introduced in #7483

## Changes

- Reintroduce missing 4844 RPC fields (GasPrice and V)

#### What types of changes does your code introduce?

- [x] Bugfix (a non-breaking change that fixes an issue)
- [ ] New feature (a non-breaking change that adds functionality)
- [ ] Breaking change (a change that causes existing functionality not to work as expected)
- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

#### If yes, did you write tests?

- [ ] Yes
- [ ] No

#### Requires documentation update

- [ ] Yes
- [x] No

#### Requires explanation in Release Notes

- [x] Yes
- [ ] No

Potentially hotfix release?

PR #8626, opened 2025-05-15
Some fixes for clef integration in RPC

ak88:
Clef can be very particular about what you send it, and if you  run it without the --advanced it will complain a lot.

This is some fixes for that, including one for `Eip1559TransactionForRpc` that will affect RPC in general. 

We will no longer serialize gasPrice, since it's a legacy/accesslist tx field only, but this is **potentially changed behavior** for some RPC consumers.

### DIFF
--- a/src/Nethermind/Nethermind.ExternalSigner.Plugin/ClefWallet.cs
+++ b/src/Nethermind/Nethermind.ExternalSigner.Plugin/ClefWallet.cs
@@ -122,12 +122,15 @@ namespace Nethermind.ExternalSigner.Plugin
             ArgumentNullException.ThrowIfNull(transaction);
 
             TransactionForRpc transactionForRpc = TransactionForRpc.FromTransaction(transaction);
+            //Clef will complain about certain fields if they are serialized
+            if (transactionForRpc is EIP1559TransactionForRpc eip1559ForRpc)
+                eip1559ForRpc.GasPrice = null;
             SignTransactionResponse? signed = rpcClient.Post<SignTransactionResponse>(
                 "account_signTransaction",
                 transactionForRpc).GetAwaiter().GetResult();
             if (signed is null || signed.Tx is null) ThrowInvalidOperationSignFailed();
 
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.ExternalSigner.Plugin/ClefWallet.cs
```diff
@@ -127,7 +127,7 @@ public void Sign(Transaction transaction, ulong chainId)
                 transactionForRpc).GetAwaiter().GetResult();
             if (signed is null || signed.Tx is null) ThrowInvalidOperationSignFailed();
 
-            transaction.Signature = new Signature(signed.Tx.R!.Value, signed.Tx.S!.Value, (ulong)(signed.Tx.V! + Signature.VOffset));
+            transaction.Signature = new Signature(signed.Tx.R!.Value, signed.Tx.S!.Value, signed.Tx.V != 0 ? (ulong)(signed.Tx.V!) : (ulong)(signed.Tx.V!) + Signature.VOffset);
```
flcl42, line 130: Is it OK not to add the offset in case of signed.Tx.V=1?

ak88: Clef acts strangely - it returns  V set to zero without offset and also yParity for tx type > 1 and for legacy and access list it returns V with chainId encoded as expected. 
I will change to check on tx type instead.

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/EIP1559TransactionForRpc.cs
```diff
@@ -23,6 +23,9 @@ public class EIP1559TransactionForRpc : AccessListTransactionForRpc, IFromTransa
     [JsonIgnore(Condition = JsonIgnoreCondition.Never)]
     public UInt256? MaxFeePerGas { get; set; }
 
+    [JsonIgnore(Condition = JsonIgnoreCondition.Always)]
```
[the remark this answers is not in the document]
emlautarom1: We should not remove them, we had an issue in the past where trying to be spec compliant resulted in failing tests and user reports: https://github.com/NethermindEth/nethermind/pull/8232

PR #8632, opened 2025-05-16
Fix infinite loop in tx pool

asdacap:
- Fix memory leak when the tx pool processing is too slow causing the background tasks to accumulate.

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

-


benaadams:
👍

PR #8635, opened 2025-05-16
Fixes one discv5 test

ak88:
The rest of the tests can only be fixed in the library we use.

I opened a PR here :
https://github.com/Pier-Two/Lantern.Discv5/pull/114

### DIFF
--- a/src/Nethermind/Nethermind.Network.Discovery/TalkReqAndRespHandler.cs
+++ b/src/Nethermind/Nethermind.Network.Discovery/TalkReqAndRespHandler.cs
@@ -1,0 +1,27 @@
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using Lantern.Discv5.WireProtocol.Messages;
+using Lantern.Discv5.WireProtocol.Messages.Requests;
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Network.Discovery/TalkReqAndRespHandler.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Nethermind.Network.Discovery;
+internal class TalkReqAndRespHandler : ITalkReqAndRespHandler
+{
+    public byte[][]? HandleRequest(byte[] protocol, byte[] request)
+    {
+        //We currently don't advertise any supported protocols
+        return [Array.Empty<byte>()];
```
LukaszRozmej, line 19: put into static not to allocate every time?

LukaszRozmej (approved):

flcl42:
Seems like it being more like a discv5 protocol rule can look better in the library, instead of fixed externally.

ak88:
> Seems like it being more like a discv5 protocol rule can look better in the library, instead of fixed externally.

Probably would make sense to have a default handler like this in the library, and expose some configuration method for it.

benaadams (approved):

PR #8647, opened 2025-05-18
Fix for `Block.AccountChanges` not being disposed in some cases

alexb5dh:
## Changes

- Fixed `Block.AccountChanges` not being disposed when rerunning blocks during syncing.

#### Requires testing

- [x] No

- [x] No

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.TxPool/TxPool.cs
+++ b/src/Nethermind/Nethermind.TxPool/TxPool.cs
@@ -1023,7 +1023,7 @@ Db usage:
         // Cleanup ArrayPoolList AccountChanges as they are not used anywhere else
         private static void DisposeBlockAccountChanges(Block block)
         {
-            if (block.AccountChanges == null) return;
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -1015,6 +1019,15 @@ private static void WriteTxPoolReport(in ILogger logger)
 ------------------------------------------------
 ");
         }
+
+        // Cleanup ArrayPoolList AccountChanges as they are not used anywhere else
+        private static void DisposeBlockAccountChanges(Block block)
+        {
+            if (block.AccountChanges == null) return;
```
LukaszRozmej, line 1026:
```suggestion
            if (block.AccountChanges is null) return;
```

LukaszRozmej (approved):

benaadams (approved):

PR #8649, opened 2025-05-18
Implement Eip7823

benaadams:
## Changes

- Implement https://eips.ethereum.org/EIPS/eip-7823

#### Requires testing

- [x] Yes

- [x] Yes

### DIFF
--- a/src/Nethermind/Nethermind.Evm.Test/Eip7823Tests.cs
+++ b/src/Nethermind/Nethermind.Evm.Test/Eip7823Tests.cs
@@ -1,0 +1,74 @@
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System;
+using System.Buffers.Binary;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using Nethermind.Core.Specs;
+using Nethermind.Evm.Precompiles;
+using Nethermind.Int256;
+using Nethermind.Specs.Forks;
+using NUnit.Framework;
+
+namespace Nethermind.Evm.Test;
+
+public class Eip7823Tests
+{
+    private const int SizeOfUInt256 = UInt256.Len * sizeof(ulong);
+    private const uint ModExpMaxInputSizeEip7823 = ModExpPrecompile.ModExpMaxInputSizeEip7823;
+    private const uint ModExpMaxInputSizeEip7823PlusOne = ModExpPrecompile.ModExpMaxInputSizeEip7823 + 1;
+
+    [TestCase(true, true, ModExpMaxInputSizeEip7823, ModExpMaxInputSizeEip7823, ModExpMaxInputSizeEip7823)]
+    [TestCase(false, true, ModExpMaxInputSizeEip7823, ModExpMaxInputSizeEip7823, ModExpMaxInputSizeEip7823)]
+    [TestCase(true, false, ModExpMaxInputSizeEip7823PlusOne, ModExpMaxInputSizeEip7823, ModExpMaxInputSizeEip7823)]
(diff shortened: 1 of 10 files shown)


src/Nethermind/Nethermind.Evm/Precompiles/ModExpPrecompile.cs
```diff
@@ -75,6 +86,13 @@ public long DataGasCost(ReadOnlyMemory<byte> inputData, IReleaseSpec releaseSpec
             }
         }
 
+        private static bool ExceedsMaxInputSize(IReleaseSpec releaseSpec, in UInt256 baseLength, in UInt256 expLength, in UInt256 modulusLength)
+            => releaseSpec.IsEip7823Enabled && ExceedsMaxInputSize(releaseSpec, (int)(uint)baseLength, (int)(uint)expLength, (int)(uint)modulusLength);
```
flcl42, line 90: Spec checked inside, methods look like duplicates. 
The comparison probably can be done between two uint256s.

flcl42 (changes requested):
What's wrong with?

    Empty inputs
    Inputs consisting of only 0x9e5faafc or 0x85474728
    A large, but invalid input: 0x9e281a98000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000021e19e0c9bab2400000

Do we need to cover it?

src/Nethermind/Nethermind.Evm.Test/Eip7823Tests.cs
```diff
@@ -0,0 +1,74 @@
...
+
+namespace Nethermind.Evm.Test;
+
+public class Eip7823Tests
+{
+    private const int SizeOfUInt256 = UInt256.Len * sizeof(ulong);
+    private const uint ModExpMaxInputSizeEip7823 = ModExpPrecompile.ModExpMaxInputSizeEip7823;
+    private const uint ModExpMaxInputSizeEip7823PlusOne = ModExpPrecompile.ModExpMaxInputSizeEip7823 + 1;
+
+    [TestCase(true, true, ModExpMaxInputSizeEip7823, ModExpMaxInputSizeEip7823, ModExpMaxInputSizeEip7823)]
```
flcl42, line 22: Can be test case source ig

src/Nethermind/Nethermind.Evm/Precompiles/ModExpPrecompile.cs
Marchhill: this seems unsafe in terms of overflows, could we instead make `ModExpMaxInputSizeEip7823` a `UInt256` and compare them directly with one function?

src/Nethermind/Nethermind.Evm.Test/Eip7823Tests.cs
```diff
@@ -0,0 +1,74 @@
...
+using Nethermind.Specs.Forks;
+using NUnit.Framework;
+
+namespace Nethermind.Evm.Test;
+
+public class Eip7823Tests
+{
+    private const int SizeOfUInt256 = UInt256.Len * sizeof(ulong);
+    private const uint ModExpMaxInputSizeEip7823 = ModExpPrecompile.ModExpMaxInputSizeEip7823;
+    private const uint ModExpMaxInputSizeEip7823PlusOne = ModExpPrecompile.ModExpMaxInputSizeEip7823 + 1;
```
Marchhill, line 20: Also try a much larger value that could overflow

Marchhill (changes requested):
.

src/Nethermind/Nethermind.Evm.Test/Eip7823Tests.cs
benaadams: Changed

src/Nethermind/Nethermind.Evm.Test/Eip7823Tests.cs
benaadams: Should be covered by regular ModExp tests?

flcl42 (approved):

LukaszRozmej (approved):

PR #8650, opened 2025-05-18
PatriciaTree - small refactoring of TraverseNode

MarekM25:
## Changes

- When I call TraverseNode, I expect it to traverse all nodes. In the old code TraverseNode means TraverseExtensionOrLeaf

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [ ] No


benaadams (approved):

src/Nethermind/Nethermind.Trie/PatriciaTree.cs
```diff
@@ -536,9 +534,7 @@ ref TraverseBranches(startNode, ref startingPath, traverseContext) :
                     TreePath startingPath = TreePath.Empty;
                     ResolveNode(RootRef, in traverseContext, in startingPath);
                     if (_logger.IsTrace) TraceNode(in traverseContext);
-                    return ref RootRef.IsBranch ?
-                        ref TraverseBranches(RootRef, ref startingPath, traverseContext) :
-                        ref TraverseNode(RootRef, traverseContext, ref startingPath);
+                    return ref TraverseNode(RootRef, traverseContext, ref startingPath);
```
Scooletz, line 537: This changes the callstack by pushing the `TraverseBranches` one level deeper (it will be `TraverseNode`->`TraverseBranches` now. Not sure if it impacts the performance, but looks like something that could make the stack much higher.

Scooletz (commented):
A few minor comments provided.

src/Nethermind/Nethermind.Trie/PatriciaTree.cs
benaadams: I think this was explicit as Branches are common, and node then does the same resolve so it cuts out the cost of additional method call

Scooletz: Having that said, this refactorization might introduce some deeper stacks / more calls.

asdacap (dismissed):

src/Nethermind/Nethermind.Trie/PatriciaTree.cs
MarekM25: Hmm, if this refactor affects performance, then we should move everything into one method since it's on the critical path in PatriciaTree

But intresting comment, thx @Scooletz

src/Nethermind/Nethermind.Trie/PatriciaTree.cs
```diff
@@ -617,22 +618,13 @@ static ref readonly CappedArray<byte> TraverseInvalid(TrieNode node)
             {
                 switch (node.NodeType)
                 {
-                    case NodeType.Branch:
-                        return ref TraverseBranch(node);
```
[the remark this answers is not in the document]
MarekM25: yes, it can be changed to if statemets

src/Nethermind/Nethermind.Trie/PatriciaTree.cs
Scooletz: I'd `run node` or even `compare` even though some comments sound smart. So many times my intuition was disproved by GH actions 😓

benaadams (approved):

src/Nethermind/Nethermind.Trie/PatriciaTree.cs
```diff
@@ -577,7 +573,20 @@ static void ThrowDecodingError(TrieNode node, in TreePath path, RlpException rlp
             }
         }
 
-        private SpanSource TraverseNode(TrieNode node, scoped in TraverseContext traverseContext, scoped ref TreePath path)
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
Scooletz, line 576: This is different from what I recall before. Should have no impact.

src/Nethermind/Nethermind.Trie/PatriciaTree.cs
```diff
@@ -874,7 +875,7 @@ static void ThrowInvalidNullNode(TrieNode node)
                 => throw new InvalidOperationException($"An attempt to set a null node as a child of the {node}");
         }
 
-        private SpanSource TraverseBranches(TrieNode node, scoped ref TreePath path, TraverseContext traverseContext)
+        private SpanSource TraverseBranches(TrieNode node, TraverseContext traverseContext, scoped ref TreePath path)
```
Scooletz, line 874: I see that we update it below so that we want to pass by value not `scoped in`. I tested this in a limited environment as I wanted to probe for the potential to pass by ref then to copy to a var, but it looks that this is nicely JITted away.

Scooletz (approved):

stdevMac:
@MarekM25 stale-PR check: twice-approved but never merged, and a year of PatriciaTree churn has made it conflict — likely quicker to redo than rebase. OK to close? If there's no reply in the next few days, we'll close this as part of the stale-PR cleanup — it can always be reopened or resubmitted.

LukaszRozmej:
Closing as obsolete — the code this refactors no longer exists.

`PatriciaTree` traversal was rewritten since this was opened: `TraverseNode`, `TraverseBranches` and the `TraverseExtensionOrLeaf` split are all gone, replaced by the `TraverseStack` / `SetNew` machinery. There is no longer a `TraverseNode` that only handles extensions and leaves, so the naming problem this fixed is not present on master.

The point was a good one and it was approved twice — it just got overtaken. Branch is kept.

PR #8655, opened 2025-05-20
Perf/allow archive node to use prune cache

asdacap:
- Remove archive code path from  `TrieStore` and reimplement it by snapshotting all block commit set with obsolete key deletion disabled.
- This streamline the code for `TrieStore` while giving archive node access to pruning cache.
- Since key removal from the in-memory cache is still active, the in-memory-cache should be largely close to head (not archive) which allow archive node to have close access to state close to head, which should make archive node not slow down too much on head.
- This almost double mgas/sec/block on archive node when archive mode is enabled at around block 21 mil. Unfortunately it also introduce memory pruning time which mean forward sync only improved by about 10-20%. I don't have numbers on actual archive node.
- Graph is (archive after, archive before, non-archive). Memory limited to 96Gb via systemd-run. Archive given 20Gb of pruning cache or memory budget (--Init.MemoryBudget). Special rocksdb flag added on archive to offset the high writes involved.
Screenshot_2025-05-20_09-25-43
- Much higher RSS though.
Screenshot_2025-05-20_09-38-13
- Since memory pruning need to store all snapshot, memory pruning time doubles in archive mode.
- `--PruningConfig.PersistenceInterval` is repurposed to pick which block to persist (set to 1 by default which mean everything). It only active in archive mode now. Also allow archiving every n block instead of every block but I did not tests its impact or use case.
- `--PruningConfig.MaxUnpersistedBlockCount` and `--PruningConfig.MinUnpersistedBlockCount` is added as a config to prevent too long or too short time between memory pruning.

- [x] Optimization
- [x] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [ ] No

#### Notes on testing

- [x] Mainnet can sync.
- [x] Mainnet archive from genesis.
- [x] Mainnet archive, but not from genesis.
- [x] Pruning still works.


LukaszRozmej:
Only somewhat related: can we track TreeNodes in separate collection by LastSeen? So when pruning we don't have to iterate all the nodes, but already have a collection based on LastSeen we can remove?

asdacap:
Pruning broke. Need to find a solution.

src/Nethermind/Nethermind.Db/IPruningConfig.cs
```diff
@@ -91,4 +91,10 @@ The max number of parallel tasks that can be used by full pruning.
 
     [ConfigItem(Description = "Minimum persisted cache prune target", DefaultValue = "50000000")]
     long PrunePersistedNodeMinimumTarget { get; set; }
+
+    [ConfigItem(Description = "Maximimum number of block worth of unpersisted state in memory. Default is 297, which is number of mainnet block per hour.", DefaultValue = "297")]
```
LukaszRozmej, line 95: Isn't mainnet like 300 blc/h? (297 is 1% missed slots)

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -66,30 +67,28 @@ public TrieStore(
...
-            // 30 because of the 1 << 31 become negative
-            if (_shardBit is <= 0 or > 30)
-            {
-                throw new InvalidOperationException($"Shard bit count must be between 0 and 30.");
-            }
+        // 30 because of the 1 << 31 become negative
+        if (_shardBit is <= 0 or > 30)
+        {
+            throw new InvalidOperationException($"Shard bit count must be between 0 and 30.");
+        }
```
LukaszRozmej: Can we move that check into the config class?

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -66,30 +67,28 @@ public TrieStore(
...
+        _persistedHashes = new ConcurrentDictionary<HashAndTinyPath, Hash256?>[_shardedDirtyNodeCount];
+        for (int i = 0; i < _shardedDirtyNodeCount; i++)
+        {
+            _dirtyNodes[i] = new TrieStoreDirtyNodesCache(this, !_nodeStorage.RequirePath, _logger);
+            _persistedHashes[i] = new ConcurrentDictionary<HashAndTinyPath, Hash256>();
         }
 
-        if (pruningStrategy.PruningEnabled && pruningConfig.TrackPastKeys && nodeStorage.RequirePath)
+        _deleteOldNodes = _pruningStrategy.DeleteObsoleteKeys;
+        if (pruningConfig.TrackPastKeys && nodeStorage.RequirePath)
```
LukaszRozmej:
```suggestion
        _pastKeyTrackingEnabled = pruningConfig.TrackPastKeys && nodeStorage.RequirePath;
```

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -506,30 +449,32 @@ public void Dump()
         }
     }
 
+    private TrieStoreState GatherState()
```
LukaszRozmej:
```suggestion
    private TrieStoreState CaptureCurrentState()
```

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -570,64 +513,95 @@ private void PersistAndPruneDirtyCache()
...
+            for (int i = 0; i < _shardedDirtyNodeCount; i++)
+            {
+                if (!_persistedHashes[i].IsEmpty)
+                {
+                    _logger.Error($"Shard {i} is not empty and contain {_persistedHashes[i].Count} item");
+                }
+            }
+        }
+
+        Action<TreePath, Hash256?, TrieNode>? persistedNodeRecorder = shouldTrackPastKey ? _persistedNodeRecorder : null;
```
LukaszRozmej: maybe it would be better to have a no-op action rather than handling nullable?

src/Nethermind/Nethermind.Db/IPruningConfig.cs
asdacap: Don't know. This says its 297. https://bitinfocharts.com/ethereum/?utm_source=chatgpt.com

PR #8656, opened 2025-05-20
EIP-7918: Blob base fee bounded by execution cost

Marchhill:

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainSpecBasedSpecProvider.cs
+++ b/src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainSpecBasedSpecProvider.cs
@@ -110,6 +110,11 @@ namespace Nethermind.Specs.ChainSpecStyle
                         throw new ArgumentException($"Blob settings are scheduled at {settings.Timestamp}, before EIP-4844, activated at {chainSpec.Parameters.Eip4844TransitionTimestamp}");
                     }
 
+                    if (settings.Target > settings.Max)
+                    {
+                        throw new ArgumentException($"Blob schedule target ({settings.Target}) should not exceed max ({settings.Max}).");
+                    }
+
                     transitions.Add(settings.Timestamp);
                 }
             }


src/Nethermind/Nethermind.Evm/BlobGasCalculator.cs
```diff
@@ -106,10 +106,31 @@ static bool FakeExponentialOverflow(UInt256 factor, UInt256 num, UInt256 denomin
...
-        return excessBlobGas < targetBlobGasPerBlock
-            ? 0
-            : excessBlobGas - targetBlobGasPerBlock;
+
+        if (excessBlobGas + blobGasUsed < targetBlobGasPerBlock)
+        {
+            return 0;
+        }
+
+        var excessBlobGasChange = blobGasUsed - targetBlobGasPerBlock;
```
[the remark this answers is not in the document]
LukaszRozmej: I think it can

src/Nethermind/Nethermind.Evm/BlobGasCalculator.cs
```diff
@@ -106,10 +106,31 @@ static bool FakeExponentialOverflow(UInt256 factor, UInt256 num, UInt256 denomin
...
+            TryCalculateFeePerBlobGas(parentBlockHeader, releaseSpec.BlobBaseFeeUpdateFraction, out UInt256 feePerBlobGas);
+            UInt256 floorCost = Eip7918Constants.BlobBaseCost * parentBlockHeader.BaseFeePerGas;
+            UInt256 targetCost = targetBlobGasPerBlock * feePerBlobGas;
+
+            // if below floor cost then increase excess blob gas
+            if (floorCost > targetCost)
+            {
+                var target = releaseSpec.TargetBlobCount;
+                var max = releaseSpec.MaxBlobCount;
+                excessBlobGasChange = blobGasUsed * (max - target) / max;
```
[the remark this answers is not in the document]
Marchhill: max should always be greater than or equal to target but added a check just in case.
For excessBlobGas the assumption (also in the spec) seems to be that this will never overflow as the base fee would have to be crazy

src/Nethermind/Nethermind.Core/Eip7918Constants.cs
```diff
@@ -0,0 +1,10 @@
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+namespace Nethermind.Core;
+
+public static class Eip7918Constants
+{
+    // floor cost in execution gas
```
benaadams, line 8:
```suggestion
    // floor cost in execution gas 16_384
```

benaadams (approved):
Nit, add comment on what `1 << 14` is :)

src/Nethermind/Nethermind.Evm/BlobGasCalculator.cs
```diff
@@ -106,10 +106,34 @@ static bool FakeExponentialOverflow(UInt256 factor, UInt256 num, UInt256 denomin
...
+            {
+                ulong target = releaseSpec.TargetBlobCount;
+                ulong max = releaseSpec.MaxBlobCount;
+                return max > target
+                    ? excessBlobGas + (blobGasUsed * (max - target) / max)
+                    : excessBlobGas;
+            }
+        }
+
+        return parentBlobGas < targetBlobGasPerBlock
```
LukaszRozmej, line 135: this check is duplicated with one in line 113?

src/Nethermind/Nethermind.Evm/BlobGasCalculator.cs
```diff
@@ -106,10 +106,34 @@ static bool FakeExponentialOverflow(UInt256 factor, UInt256 num, UInt256 denomin
...
+            TryCalculateFeePerBlobGas(parentBlockHeader, releaseSpec.BlobBaseFeeUpdateFraction, out UInt256 feePerBlobGas);
+            UInt256 floorCost = Eip7918Constants.BlobBaseCost * parentBlockHeader.BaseFeePerGas;
+            UInt256 targetCost = targetBlobGasPerBlock * feePerBlobGas;
+
+            // if below floor cost then increase excess blob gas
+            if (floorCost > targetCost)
+            {
+                ulong target = releaseSpec.TargetBlobCount;
+                ulong max = releaseSpec.MaxBlobCount;
+                return max > target
```
LukaszRozmej, line 129: ah this is spec related, so we can remove this check, we shouldn't pay it for every call (we can throw when parsing chainspec)

Marchhill: added a check for this when loading blob schedule

PR #8657, opened 2025-05-20
Kute: order by filename

benaadams:
## Changes

- Write time is random when copying a directory and then using the files

#### Requires testing

- [x] No


kamilchodola (approved):

emlautarom1:
Only issue I see is that the naming format used by Nethermind it's not friendly to lexicographically ordering. For example, this is how files would be sorted using the proposed approach:

```
rpc.0.txt
rpc.1.txt
rpc.10.txt
rpc.12.txt
rpc.2.txt
rpc.20.txt
rpc.3.txt
```

I think this was the reason to use last modification time.

kamilchodola:
Proper fix then should allow you to decide on how to sort files

emlautarom1:
@kamilchodola Please create a ticket with the feature request so we can track it.

PR #8658, opened 2025-05-20
Optimism CL: Split finalized/safe blocks

deffrian:
Fixes Closes Resolves #
Right now we process only finalized L1 blocks. This causes Safe and Finalized blocks to be equal. This pr extends processing to unsafe L1 blocks.

## Changes

- Process L1 blocks up to current unsafe head.
- Handle L1 unsafe block reorgs
- L1Bridge no longer works in different thread

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


src/Nethermind/Nethermind.Optimism/CL/Decoding/DecodingPipeline.cs
```diff
@@ -69,4 +78,28 @@ public async Task Run(CancellationToken token)
...
+        _frameQueue.Clear();
+    }
+
+    private TaskCompletionSource _resetCompleted = new();
+
+    public async Task Reset(CancellationToken token)
+    {
+        Interlocked.Exchange(ref _resetRequested, 1);
+        await _resetCompleted.Task;
+        _resetCompleted = new();
```
emlautarom1, line 103: I have a hard time making sure that this assignment is actually thread safe. I'd suggest the following:

- Add a `Channel<TaskCompletionSource> _resetRequests = Channel.CreateBounded<TaskCompletionSource>(1);` field
- Change `Reset` to:

```C#
async Task Reset(CancellationToken token)
{
    var resetRequest = new TaskCompletionSource();
    await _resetRequests.Writer.WriteAsync(resetRequest, token);
    await resetRequest.Task;
}
```
- Change the check for resets in `Run` to:

```C#
if (_resetRequests.Reader.TryRead(out var resetRequest))
{
    try
    {
        await Clear(token)
        resetRequest.SetResult();
    }
    catch (OperationCanceledException)
    {
        resetRequest.SetCanceled(token);
        throw;
    }
    catch (Exception e)
    {
        resetRequest.SetException(e);
        throw;
    }
}
```

With this we support cancellation of resets and we don't need to reassign any field. I'm not 100% sure about the exception propagation though.

src/Nethermind/Nethermind.Optimism/CL/Decoding/DecodingPipeline.cs
```diff
@@ -69,4 +78,28 @@ public async Task Run(CancellationToken token)
...
+    {
+        while (await _inputChannel.Reader.WaitToReadAsync(token))
+        {
+            while (_inputChannel.Reader.TryRead(out _)) { }
+        }
+
+        while (await _outputChannel.Reader.WaitToReadAsync(token))
+        {
+            while (_outputChannel.Reader.TryRead(out _)) { }
+        }
```
emlautarom1, line 92: We're draining the channels here but I don't see how we're preventing other components from continue writing (is this OK behavior)? For example:

https://github.com/NethermindEth/nethermind/blob/4b898515feef499e63a61bf918cdc7d7eecbe76c/src/Nethermind/Nethermind.Optimism/CL/Driver.cs#L84

src/Nethermind/Nethermind.Optimism/CL/L1Bridge/IL1Bridge.cs
```diff
@@ -1,21 +1,55 @@
...
+    public static L1BridgeStepResult Reorg => new() { Type = L1BridgeStepResultType.Reorg };
+
+    public static L1BridgeStepResult Block(DaDataSource[] data) =>
+        new() { Type = L1BridgeStepResultType.Block, NewData = data };
+
+    public static L1BridgeStepResult Finalization(ulong blockNumber) => new()
+        { Type = L1BridgeStepResultType.Finalization, NewFinalized = blockNumber };
+
+    public L1BridgeStepResultType Type { private set; get; }
+    public DaDataSource[]? NewData { private set; get; }
```
emlautarom1, line 43: We can change this to `List<DaDataSource>` to avoid unnecessary conversions:

```suggestion
    public List<DaDataSource>? NewData { private set; get; }
```

src/Nethermind/Nethermind.Optimism/CL/L1Bridge/EthereumL1Bridge.cs
```diff
@@ -110,14 +122,16 @@ private async Task ProcessBlock(L1Block block, CancellationToken token)
                     if (_batcherInboxAddress == transaction.To &&
                         _batchSubmitter == transaction.From)
                     {
-                        await ProcessCalldataBatcherTransaction(transaction, token);
+                        result.Add(ProcessCalldataBatcherTransaction(transaction, block.Number, token));
                     }
                 }
             }
+            _currentHead = BlockId.FromL1Block(block);
+            return result.Count == 0 ? null : L1BridgeStepResult.Block(result.ToArray());
```
emlautarom1, line 130: What does returning `null` mean here? Something failed?

src/Nethermind/Nethermind.Optimism/CL/L1Bridge/EthereumL1Bridge.cs
```diff
@@ -149,28 +163,71 @@ private async Task ProcessBlobBatcherTransaction(L1Transaction transaction, int
...
+            L1BridgeStepResult? result = await ProcessBlock(block, cancellationToken);
+            if (result is not null) _unfinalizedL1BlocksQueue.Enqueue(result);
+        }
+        return _unfinalizedL1BlocksQueue.Count == 0 ? null : _unfinalizedL1BlocksQueue.Dequeue();
+    }
+
+    private void LogReorg()
+    {
+        if (_logger.IsInfo) _logger.Info("L1 reorg detected. Resetting pipeline");
     }
```
emlautarom1, line 231: Unused code

src/Nethermind/Nethermind.Optimism/CL/L1Bridge/IL1Bridge.cs
```diff
@@ -1,21 +1,55 @@
...
+
+public enum L1BridgeStepResultType
+{
+    Block,
+    Finalization,
+    Skip,
+    Reorg,
+}
+
+public class L1BridgeStepResult
```
emlautarom1, line 31: As an alternative design, we could use sealed classes as follows:

```c#
public abstract class L1BridgeStepResultV2
{
    private L1BridgeStepResultV2() { }

    public sealed class Skip
    {
        public override string ToString() => "L1 Event Skip";
    }

    public sealed class Reorg
    {
        public override string ToString() => "L1 Event Skip";
    }

    public sealed class Block
    {
        public required List<DaDataSource> NewData { get; init; }

        public override string ToString() => $"L1 Event Block. Blobs count {NewData.Count}";
    }

    public sealed class Finalization
    {
        public required ulong NewFinalized { get; init; }

        public override string ToString() => $"L1 Event Finalization. Block Number: {NewFinalized}";
    }
}
```

When doing `switch` cases we can use `is` patterns and extract the correct cases, avoiding `!` annotations, plus I think it's less code.

src/Nethermind/Nethermind.Optimism/CL/Driver.cs
```diff
@@ -47,49 +51,65 @@ public Driver(IL1Bridge l1Bridge,
...
+                        {
+                            await ProcessReorg(token);
+                            l1BridgeStep = _l1Bridge.Step(token);
+                            break;
+                        }
+                        case L1BridgeStepResultType.Skip:
+                        {
+                            l1BridgeStep = Task.Run(async () =>
+                            {
+                                await Task.Delay(120000, token);
```
emlautarom1, line 105: Extract this delay into a constant

src/Nethermind/Nethermind.Optimism/CL/Driver.cs
```diff
@@ -47,49 +51,65 @@ public Driver(IL1Bridge l1Bridge,
...
+                                return await _l1Bridge.Step(token);
+                            });
```
emlautarom1, line 107: Could we keep the `Task.Delay` inside the `case` statement, and extract the common `l1BridgeStep = _l1BridgeStep.Step(token)` outside the `switch` block?

src/Nethermind/Nethermind.Optimism/CL/Driver.cs
```diff
@@ -104,10 +124,88 @@ public async Task Run(CancellationToken token)
...
+    }
+
+    private readonly Queue<(ulong L1BatchOrigin, BlockId LastL2Block)> _safeBlocksQueue = new();
+
+    private async Task ProcessNewFinalized(ulong newFinalized, CancellationToken token)
+    {
+        if (_logger.IsInfo) _logger.Info($"Processing new finalized L1 block. {newFinalized}");
+        _currentFinalizedL1Block = newFinalized;
+        BlockId? lastL2Block = null;
+        while(_safeBlocksQueue.Any())
```
emlautarom1, line 139:
```suggestion
        while (_safeBlocksQueue.Count > 0)
```

emlautarom1 (approved):
Some comments, but approved as discussed internally.

PR #8660, opened 2025-05-21
Optimize RlpStream.Encode(ulong)

benaadams:
## Changes

- Call single `virtual Write(ReadOnlySpan<byte>)` rather than multiple `virtual WriteByte(byte)` calls

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Serialization.Rlp/RlpStream.cs
+++ b/src/Nethermind/Nethermind.Serialization.Rlp/RlpStream.cs
@@ -387,7 +387,7 @@ namespace Nethermind.Serialization.Rlp
 
         public void Encode(bool value) => Encode(value ? (byte)1 : (byte)0);
 
-        public void Encode(int value) => Encode((long)value);
+        public void Encode(int value) => Encode((ulong)(long)value);
 
         public void Encode(long value) => Encode((ulong)value);
 
--- a/src/Nethermind/Nethermind.Serialization.Rlp/RlpStream.cs
+++ b/src/Nethermind/Nethermind.Serialization.Rlp/RlpStream.cs
@@ -385,136 +385,48 @@ namespace Nethermind.Serialization.Rlp
             }
         }
 
-        public void Encode(bool value)
-        {
-            Encode(value ? (byte)1 : (byte)0);
-        }
+        public void Encode(bool value) => Encode(value ? (byte)1 : (byte)0);
 
-        public void Encode(int value)
-        {
-            Encode((long)value);
-        }
+        public void Encode(int value) => Encode((long)value);
 
-        public void Encode(long value)
-        {
-            if (value == 0L)
-            {
-                EncodeEmptyByteArray();
-                return;
-            }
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Serialization.Rlp/RlpStream.cs
```diff
@@ -385,136 +385,48 @@ public void Encode(byte value)
...
-        {
-            Encode(value ? (byte)1 : (byte)0);
-        }
+        public void Encode(bool value) => Encode(value ? (byte)1 : (byte)0);
 
-        public void Encode(int value)
-        {
-            Encode((long)value);
-        }
+        public void Encode(int value) => Encode((long)value);
```
LukaszRozmej, line 390:
```suggestion
        public void Encode(int value) => Encode((ulong)value);
```

src/Nethermind/Nethermind.Serialization.Rlp/RlpStream.cs
```diff
@@ -385,136 +385,48 @@ public void Encode(byte value)
...
+                // Reuse space in valueSpan for prefix rather than copying
+                valueSpan[leadingZeroBytes - 1] = prefix;
+                output = valueSpan.Slice(leadingZeroBytes - 1, 1 + valueLength);
+            }
+            else
+            {
+                // Build final output: prefix + value bytes
+                output[0] = prefix;
+                valueSpan.Slice(leadingZeroBytes, valueLength).CopyTo(output.Slice(1));
+                output = output.Slice(0, 1 + valueLength);
```
LukaszRozmej, line 426: wouldn't it be better to not using `output` at all and have one Write(byte) for prefix and then one Write(valuespan.Slice)?

benaadams: Issue is more that they are virtual calls and have very different behaviour if NettyStream or KeccakStream; so faster just to do the work on the local spans

src/Nethermind/Nethermind.Serialization.Rlp/RlpStream.cs
benaadams: Would need to be `(ulong)(long)value` due to negative number extension behaviour

benaadams, line 390:
```suggestion
        public void Encode(int value) => Encode((ulong)(long)value);
```

src/Nethermind/Nethermind.Serialization.Rlp/RlpStream.cs
benaadams: 2 non inlinable virtual calls; each calling 2 more non-inlinable interface calls (for `NettyRlpStream`); potentially reallocating and moving memory, span is just a pointer bump due to fixed small size and `SkipLocalsInit`

asdacap (approved):
Straightforward enough

PR #8661, opened 2025-05-21
Use RocksDB PinnableSlice when available

benaadams:
## Changes

- Reduces allocation and additional copy from RocksDB; which is good for large objects like Blocks
- Better performance when downloading receipts (as they load blocks)

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Serialization.Rlp/KeyValueStoreRlpExtensions.cs
+++ b/src/Nethermind/Nethermind.Serialization.Rlp/KeyValueStoreRlpExtensions.cs
@@ -50,12 +50,12 @@ public static class KeyValueStoreRlpExtensions
             if (decoder is IRlpValueDecoder<TItem> valueDecoder)
             {
                 item = db is IReadOnlyNativeKeyValueStore native
-                    ? GetViaNative(native, key, valueDecoder, rlpBehaviors)
-                    : GetViaSpan(db, key, valueDecoder, rlpBehaviors);
+                    ? Get(native, key, valueDecoder, rlpBehaviors)
+                    : Get(db, key, valueDecoder, rlpBehaviors);
             }
             else
             {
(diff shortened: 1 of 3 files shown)


src/Nethermind/Nethermind.Serialization.Rlp/KeyValueStoreRlpExtensions.cs
```diff
@@ -49,42 +49,13 @@ public static class KeyValueStoreRlpExtensions
...
-                    var rlpValueContext = data.AsRlpValueContext();
-                    item = valueDecoder.Decode(ref rlpValueContext, rlpBehaviors | RlpBehaviors.AllowExtraBytes);
-                }
-                finally
-                {
-                    db.DangerousReleaseMemory(data);
-                }
+                item = db is IReadOnlyNativeKeyValueStore native
+                    ? GetViaNative(native, key, valueDecoder, rlpBehaviors)
+                    : GetViaSpan(db, key, valueDecoder, rlpBehaviors);
```
LukaszRozmej, line 54: Could we utilize generics here?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Serialization.Rlp/KeyValueStoreRlpExtensions.cs
benaadams: Don't think so as is pattern matching to get type.

Have changed to have same names though and just use overloading

LukaszRozmej: But you should have made those overloads public, they are still private

benaadams: Could also make the method calling them private as not called outside the extension

PR #8663, opened 2025-05-21
Implement EIP-7825: Transaction Gas Limit Cap

Marchhill:
## Changes

- TxPool filtering
- Enforce in block processing

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


LukaszRozmej (commented):
One more place `TxValidator`

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
```diff
@@ -336,3 +341,16 @@ public ValidationResult IsWellFormed(Transaction transaction, IReleaseSpec relea
...
+
+public sealed class GasLimitCapTxValidator : ITxValidator
+{
+    public static readonly GasLimitCapTxValidator Instance = new();
+    private GasLimitCapTxValidator() { }
+
+    public ValidationResult IsWellFormed(Transaction transaction, IReleaseSpec releaseSpec)
+    {
+        long gasLimitCap = Eip7825Constants.GetTxGasLimitCap(releaseSpec);
+        return releaseSpec.IsEip7825Enabled && transaction.GasLimit > gasLimitCap ?
```
LukaszRozmej, line 355: To be honest this seems a bit redundant. Either use the default value here directly or `GetTxGasLimitCap` should return `long?` and the `IsEip7825Enabled` check should be inside and return `null` if not set.

LukaszRozmej (approved):

src/Nethermind/Nethermind.Specs/Forks/19_Osaka.cs
```diff
@@ -15,6 +14,7 @@ protected Osaka()
     {
         Name = "Osaka";
         IsEofEnabled = true;
+        // IsEip7825Enabled = true;
```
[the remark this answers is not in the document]
Marchhill: pyspec test in CI was failing otherwise

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
Marchhill: Sure I think either way works, will change it

Marchhill: will just return long.MaxValue, seems a bit cleaner. Feel free to change back if you prefer the other way

PR #8616, opened 2025-05-12
Feature/Rework DiscV4

asdacap:
- Rework DiscV4 to use a somewhat reusable Kademlia implementation that is also used for Portal.
  - This kademlia implementation uses proper tree routing table.
- `INodeSource` from discv4 now actively iterate the kad network for random target instead of randomly find nodes on another thread and waiting for new nodes event to send to `INodeSource`.
  - Discovery traffic now properly go down as it uses `IAsyncEnumerable` to block further request when peer manager is full.
- Additionally, peer now get pinged first before sending to `PeerManager`. 
- Did not add forkid check in this PR. Experiment shows that confusingly not many peer expose forkid.  
- Although it does populate peer candidate list faster, this is largely due to much higher limits. In fact the efficiency in terms of new candidate per message is lower. 
- Graph is (After, Before).
Screenshot_2025-05-22_08-10-29

## Changes

- Almost everything Discv4 except for serialization.

## How to review.

- Probably start with `DiscoveryApp`, follow `DiscV4KademliaModule`, and then `KademliaModule`.
- The important part are probably `KademliaDiscv4Adapter`.

#### What types of changes does your code introduce?

- [X] Optimization
- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

#### If yes, did you write tests?

- [X] Yes
- [ ] No

#### Notes on testing

- [x] Hive discv4
- [ ] Mainnet 
- [ ] Energyweb

PR #8664, opened 2025-05-21
Slight discovery refactor.

asdacap:
- Trying to reduce change to #8616 .
- Use `ArraySegment<Node>` in `Neighbours` msg as the underlying `k` array need to be sliced to two part, `12` and `4` because of MTU limit.
- Mode node persistence logic to `NodePersistenceManager`.

- [ ] Optimization
- [X] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Network.Discovery/Serializers/NeighborsMsgSerializer.cs
+++ b/src/Nethermind/Nethermind.Network.Discovery/Serializers/NeighborsMsgSerializer.cs
@@ -44,10 +44,10 @@ public class NeighborsMsgSerializer : DiscoveryMsgSerializerBase, IZeroInnerMess
         PrepareBufferForSerialization(byteBuffer, totalLength, (byte)msg.MsgType);
         NettyRlpStream stream = new(byteBuffer);
         stream.StartSequence(contentLength);
-        if (msg.Nodes.Length != 0)
+        if (msg.Nodes.Count != 0)
         {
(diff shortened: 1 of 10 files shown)


LukaszRozmej (approved):

src/Nethermind/Nethermind.Network.Discovery/Serializers/NeighborsMsgSerializer.cs
```diff
@@ -82,10 +82,10 @@ private static Node[] DeserializeNodes(RlpStream rlpStream)
         return rlpStream.DecodeArray<Node>(_decodeItem);
     }
 
-    private static int GetNodesLength(Node[] nodes, out int contentLength)
+    private static int GetNodesLength(ArraySegment<Node> nodes, out int contentLength)
     {
         contentLength = 0;
-        for (int i = 0; i < nodes.Length; i++)
+        for (int i = 0; i < nodes.Count; i++)
```
benaadams, line 88: Would be faster to [AsSpan](https://learn.microsoft.com/en-us/dotnet/api/system.memoryextensions.asspan?view=net-9.0#system-memoryextensions-asspan-1(system-arraysegment((-0))-system-index)) the `ArraySegment<Node>` and iterate the span; but if not many items probably not too worth it

PR #8665, opened 2025-05-21
Slice span rather than array

benaadams:
## Changes

- We slice an array to array, when could just slice to span for less allocations

- [x] Optimization

#### Requires testing

- [x] No


Scooletz:
This kicked me so many times. Should we have it renamed @benaadams to something more mouthful? `SliceToArray` ?

PR #8674, opened 2025-05-22
Mild performance tweaks

benaadams:
## Changes

- Switches ContainsKey+indexer to TryGetValue for single-dictionary lookups  
- Replaces LINQ `.Any()` with `Count` or `IsEmpty` checks on collections  
- Uses char overloads instead of string for simple checks

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainSpecLoader.cs
+++ b/src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainSpecLoader.cs
@@ -398,11 +398,11 @@ public class ChainSpecLoader(IJsonSerializer serializer) : IChainSpecLoader
             if (account.Value.CodeHash is not null)
             {
                 string codeHashString = account.Value.CodeHash.ToString();
-                if (chainSpecJson.CodeHashes is null || !chainSpecJson.CodeHashes.TryGetValue(codeHashString, out var value)) throw new ArgumentException($"CodeHash {account.Value.CodeHash} is not found");
+                if (chainSpecJson.CodeHashes is null || !chainSpecJson.CodeHashes.TryGetValue(codeHashString, out var codeHash)) throw new ArgumentException($"CodeHash {account.Value.CodeHash} is not found");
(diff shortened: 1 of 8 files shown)


src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainSpecLoader.cs
```diff
@@ -398,11 +398,11 @@ private static void LoadAllocations(ChainSpecJson chainSpecJson, ChainSpec chain
             if (account.Value.CodeHash is not null)
             {
                 string codeHashString = account.Value.CodeHash.ToString();
-                if (chainSpecJson.CodeHashes is null || !chainSpecJson.CodeHashes.ContainsKey(codeHashString)) throw new ArgumentException($"CodeHash {account.Value.CodeHash} is not found");
+                if (chainSpecJson.CodeHashes is null || !chainSpecJson.CodeHashes.TryGetValue(codeHashString, out var value)) throw new ArgumentException($"CodeHash {account.Value.CodeHash} is not found");
```
asdacap, line 401:
```suggestion
                if (chainSpecJson.CodeHashes is null || !chainSpecJson.CodeHashes.TryGetValue(codeHashString, out var codeHash)) throw new ArgumentException($"CodeHash {account.Value.CodeHash} is not found");
```

src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainSpecLoader.cs
```diff
@@ -398,11 +398,11 @@ private static void LoadAllocations(ChainSpecJson chainSpecJson, ChainSpec chain
...
                 chainSpec.Allocations[address] = new ChainSpecAllocation(
                     account.Value.Balance ?? UInt256.Zero,
                     account.Value.Nonce,
-                    chainSpecJson.CodeHashes[codeHashString],
+                    value,
```
asdacap, line 405:
```suggestion
                    codeHash,
```

asdacap (approved):

PR #8687, opened 2025-05-24
Optimize 12 x 20byte  & 32byte OpCodes

benaadams:
## Changes

There as some opcodes that we upfront know their sizes but are pushing to stack as variable size, incurring additional work:

Which are in our worse performing opcodes

image

- Align EvmStack to 32 bytes

- Reduce large stack valuetype copies by passing param `readonly ref`/`in` of the 32 byte value type (so register)
  - `ValueHash256`
  - `Uint256`
  - "Word" aka `Vector256<byte>`

- Optimize the 20 byte (`Address`) opcodes to use fixed size push20
  - OpAddress
  - OpOrigin
  - OpCaller
  - OpCoinbase

- Optimize the 32 byte (`Hash256`) opcodes to use fixed size push32
  - OpPrevRandao
  - OpChainId
  - OpExtCodeHash

- Optimize the 32 byte (`UInt256`) opcodes to use fixed size push32 (+vectorised endian swap)
  - OpCallValue
  - OpGasPrice 
  - OpBlobBaseFee
  - OpBaseFee 
  - OpBalance
  - OpSelfBalance

~Refactor out some of `Unsafe.`~

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Core/Account.cs
+++ b/src/Nethermind/Nethermind.Core/Account.cs
@@ -16,14 +16,15 @@ namespace Nethermind.Core
 
         private readonly Hash256? _codeHash;
         private readonly Hash256? _storageRoot;
-        private readonly UInt256 _balance;
+        public readonly UInt256 Nonce;
+        public readonly UInt256 Balance;
 
         public Account(in UInt256 balance)
         {
             _codeHash = null;
             _storageRoot = null;
             Nonce = default;
-            _balance = balance;
+            Balance = balance;
         }
 
         public Account(in UInt256 nonce, in UInt256 balance)
@@ -31,7 +32,7 @@ namespace Nethermind.Core
             _codeHash = null;
             _storageRoot = null;
             Nonce = nonce;
-            _balance = balance;
+            Balance = balance;
         }
 
         private Account()
@@ -39,7 +40,7 @@ namespace Nethermind.Core
             _codeHash = null;
             _storageRoot = null;
             Nonce = default;
-            _balance = default;
+            Balance = default;
         }
 
         public Account(in UInt256 nonce, in UInt256 balance, Hash256 storageRoot, Hash256 codeHash)
@@ -47,7 +48,7 @@ namespace Nethermind.Core
             _codeHash = codeHash == Keccak.OfAnEmptyString ? null : codeHash;
             _storageRoot = storageRoot == Keccak.EmptyTreeHash ? null : storageRoot;
             Nonce = nonce;
-            _balance = balance;
+            Balance = balance;
         }
 
         private Account(Account account, Hash256? storageRoot)
@@ -55,7 +56,7 @@ namespace Nethermind.Core
             _codeHash = account._codeHash;
             _storageRoot = storageRoot == Keccak.EmptyTreeHash ? null : storageRoot;
             Nonce = account.Nonce;
-            _balance = account.Balance;
+            Balance = account.Balance;
         }
 
         private Account(Hash256? codeHash, Account account)
@@ -63,7 +64,7 @@ namespace Nethermind.Core
             _codeHash = codeHash == Keccak.OfAnEmptyString ? null : codeHash;
             _storageRoot = account._storageRoot;
             Nonce = account.Nonce;
-            _balance = account.Balance;
+            Balance = account.Balance;
         }
 
         private Account(Account account, in UInt256 nonce, in UInt256 balance)
@@ -71,15 +72,13 @@ namespace Nethermind.Core
             _codeHash = account._codeHash;
             _storageRoot = account._storageRoot;
             Nonce = nonce;
-            _balance = balance;
+            Balance = balance;
         }
 
         public bool HasCode => _codeHash is not null;
 
         public bool HasStorage => _storageRoot is not null;
 
-        public UInt256 Nonce { get; }
-        public ref readonly UInt256 Balance => ref _balance;
         public Hash256 StorageRoot => _storageRoot ?? Keccak.EmptyTreeHash;
         public Hash256 CodeHash => _codeHash ?? Keccak.OfAnEmptyString;
         public bool IsTotallyEmpty => _storageRoot is null && IsEmpty;


src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -63,7 +65,7 @@ public sealed unsafe partial class VirtualMachine(
     internal static readonly PrecompileExecutionFailureException PrecompileExecutionFailureException = new();
     internal static readonly OutOfGasException PrecompileOutOfGasException = new();
 
-    private readonly byte[] _chainId = ((UInt256)specProvider.ChainId).ToBigEndian();
+    private readonly Word _chainId = Vector256.Create(((UInt256)specProvider.ChainId).ToBigEndian());
```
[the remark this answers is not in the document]
benaadams: What you talking about, fool?

src/Nethermind/Nethermind.Core/Extensions/Bytes.cs
```diff
@@ -151,6 +151,34 @@ public static void SetBit(this ref byte b, int bitNumber)
             b |= (byte)mask;
         }
 
+        private unsafe static int GetAlignmentOffset(this byte[] array, uint alignment)
```
LukaszRozmej, line 154: I am not convinced that this is best place for this methods or should it be extension methods.
I think they should live in some `PinnedExtensions` class or something, so it won't be accidentally called on non-pinned objects.

src/Nethermind/Nethermind.Core/Account.cs
```diff
@@ -78,7 +78,7 @@ private Account(Account account, in UInt256 nonce, in UInt256 balance)
         public bool HasStorage => _storageRoot is not null;
 
         public UInt256 Nonce { get; }
-        public UInt256 Balance { get; }
+        public readonly UInt256 Balance;
```
LukaszRozmej, line 81: Why this pattern?
Why not this pattern for everything else?

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Environment.cs
```diff
@@ -87,17 +87,17 @@ public interface IOpEnvUInt64
...
-        where TOpEnv : struct, IOpEnvBytes
+    public static EvmExceptionType InstructionEnvAddress<TOpEnv>(VirtualMachine vm, ref EvmStack stack, ref long gasAvailable, ref int programCounter)
+        where TOpEnv : struct, IOpEnvAddress
     {
         // Deduct the gas cost as defined by the operation implementation.
         gasAvailable -= TOpEnv.GasCost;
 
         // Execute the operation and retrieve the result.
-        TOpEnv.Operation(vm.EvmState, out Span<byte> result);
+        Address result = TOpEnv.Operation(vm.EvmState);
```
LukaszRozmej, line 97: should we have address struct?

src/Nethermind/Nethermind.Evm/StackPool.cs
```diff
@@ -42,7 +46,8 @@ public void ReturnStacks(byte[] dataStack, ReturnState[] returnStack)
 
         return
         (
-            new byte[(EvmStack.MaxStackSize + EvmStack.RegisterLength) * 32],
+            // Include extra Vector256<byte>.Count and pin so we can align to 32 bytes
+            GC.AllocateUninitializedArray<byte>(StackLength + Vector256<byte>.Count, pinned: true),
```
LukaszRozmej, line 50: will it ever be unpinned?

src/Nethermind/Nethermind.Evm/EvmStack.cs
```diff
@@ -18,6 +19,7 @@ namespace Nethermind.Evm;
 
 using Word = Vector256<byte>;
 using HalfWord = Vector128<byte>;
+using static Unsafe;
```
LukaszRozmej, line 22: I don't like this change to be honest, I would prefer each unsafe operation to be explicitly marked in code. Makes me analyze them more, now I can loose them easier.

src/Nethermind/Nethermind.Evm/EvmStack.cs
```diff
@@ -185,6 +186,27 @@ public void Push20Bytes(ref byte value)
...
+        // First 4+8 bytes are zero
+        Unsafe.As<byte, ulong>(ref bytes) = 0;
+        Unsafe.As<byte, uint>(ref Unsafe.Add(ref bytes, sizeof(ulong))) = 0;
+
+        // 20 bytes which is uint+Vector128
+        Unsafe.As<byte, uint>(ref Unsafe.Add(ref bytes, sizeof(uint) + sizeof(ulong)))
+            = Unsafe.As<byte, uint>(ref value);
+
+        Unsafe.As<byte, HalfWord>(ref Unsafe.Add(ref bytes, sizeof(ulong) + sizeof(ulong)))
+            = Unsafe.As<byte, HalfWord>(ref Unsafe.Add(ref value, sizeof(uint)));
```
LukaszRozmej, line 207: If we just use Address.ToHash() it will be worse performance?

benaadams: Could maybe do a shared method

src/Nethermind/Nethermind.Core/Account.cs
benaadams: Only changed the ones I needed a `ref ` to

src/Nethermind/Nethermind.Evm/EvmStack.cs
benaadams: Passed through to other push

src/Nethermind/Nethermind.Core/Account.cs
benaadams: Changed back to property, but is more changes

src/Nethermind/Nethermind.Core/Extensions/Bytes.cs
benaadams: Moved to VirtualMachine as only place these methods are used

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Environment.cs
benaadams: Doesn't really help as the array backing is used everywhere; would be a large change to alter it

src/Nethermind/Nethermind.Core/Account.cs
LukaszRozmej: Yes I've seen it later. This was only a question :)
 IMO it is fine to be readonly field, but maybe we need more readonly fields ? :)

benaadams: Change back to field and also the other one

LukaszRozmej (approved):
Maybe good to run some hive test before merging?

benaadams:
Hive tests https://github.com/NethermindEth/nethermind/actions/runs/15231337456

PR #8691, opened 2025-05-25
Perform stack writes as single 32 vector

benaadams:
## Changes

- Rather than doing multiple memory writes; prepare value in Vector256<byte> register and do single write to memory.

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Evm/EvmStack.cs
+++ b/src/Nethermind/Nethermind.Evm/EvmStack.cs
@@ -57,6 +57,10 @@ public ref struct EvmStack
     private ref Word PushedHead()
         => ref Unsafe.As<byte, Word>(ref PushBytesRef());
 
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    private static Word CreateWordFromUInt64(ulong value)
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Evm/EvmStack.cs
```diff
@@ -94,9 +94,12 @@ public void PushByte(byte value)
...
         ref byte bytes = ref PushBytesRef();
-        // Not full entry, clear first
-        Unsafe.As<byte, Word>(ref bytes) = default;
-        Unsafe.Add(ref bytes, WordSize - sizeof(byte)) = value;
+
+        // Build a 256-bit vector: [ 0, 0, 0, (value << 56) ]
+        // - when viewed as bytes: all zeros except byte[31] == value
+
+        // Single 32-byte store: last byte as value
+        Unsafe.As<byte, Word>(ref bytes) = Vector256.Create(0UL, 0UL, 0UL, (ulong)value << 56).AsByte();
```
LukaszRozmej, line 102: Can you reuse code? There are 4 times this code is reused, 6 times if you count no shift.

LukaszRozmej (approved):

PR #8692, opened 2025-05-25
Reintroduce Jit time branch elimination

benaadams:
## Changes

- Reintroduce Jit time branch elimination for EvmStack; it was removed during Eof due to complexity on Type and using function pointers. However the generic can be more easily added to the methods rather than the type and achieve same effect.

- [x] Optimization

#### Requires testing

- [x] No


LukaszRozmej (approved):
Can we move `IFlag` to `Nethermind.Core`? I think there will be more uses for it later.

benaadams:
> Can we move `IFlag` to `Nethermind.Core`? I think there will be more uses for it later.

Moved

PR #8699, opened 2025-05-27
Warm up evm instructions

benaadams:
## Changes

- Ensure all OpCodes are Tier1 Jitted by the time they are actually used

- [x] Optimization

#### Requires testing

- [x] No


benaadams:
> Won't it be the case that it will happen automatically within a few blocks? Or are you trying to minimize the penalty for specifically these few first blocks here?

Don't trust that the function pointers are backpatching direct to the Tier1 version when they are re-compiled

LukaszRozmej:
Might want to use spec from #8698

benaadams:
> Might want to use spec from #8698

Appears to be a draft?

Scooletz (approved):

PR #8712, opened 2025-05-29
Fix/disconnect without sending reason

asdacap:
- On incoming connection, if the max active peer count is full, we immediately disconnect the peer. 
- This happen so early in the connection that the disconnect reason message is sent uncompressed as p2p handler does not have the opportunity to turn on compression.

## Changes

- PeerManager does not disconnect the peer unless active peer count exceed max active peer + margin.
- IProtocolValidator now check peer manager for peer limit.

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Testing...

### DIFF
--- a/src/Nethermind/Nethermind.Network/PeerManager.cs
+++ b/src/Nethermind/Nethermind.Network/PeerManager.cs
@@ -85,7 +85,14 @@ namespace Nethermind.Network
         public IReadOnlyCollection<Peer> ConnectedPeers => _peerPool.ActivePeers.Values.Where(IsConnected).ToList();
 
         public int MaxActivePeers => _networkConfig.MaxActivePeers + _peerPool.StaticPeerCount;
+        public int ActivePeersCount => _peerPool.ActivePeerCount;
         private int AvailableActivePeersCount => MaxActivePeers - _peerPool.ActivePeers.Count;
+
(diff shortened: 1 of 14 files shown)


src/Nethermind/Nethermind.Network/PeerManager.cs
```diff
@@ -724,26 +731,11 @@ private void ProcessIncomingConnection(ISession session)
...
-                }
-
-                if (initCount >= MaxActivePeers)
-                {
-                    if (_logger.IsTrace) _logger.Trace($"Initiating disconnect with {session} {DisconnectReason.TooManyPeers} {DisconnectType.Local}");
-                    session.InitiateDisconnect(DisconnectReason.TooManyPeers, $"{initCount}");
-                    return;
-                }
+                if (_logger.IsTrace) _logger.Trace($"Initiating disconnect with {session} {DisconnectReason.HardLimitTooManyPeers} {DisconnectType.Local}");
+                session.InitiateDisconnect(DisconnectReason.HardLimitTooManyPeers, $"{ActivePeersCount}");
```
LukaszRozmej, line 737: But what if we exceed that limit, we still disconnect here, so is it a proper fix?

PR #8716, opened 2025-05-29
Optimize JournalSet

benaadams:
## Changes

- Reduce interface implementation scope and remove unused methods
- Use faster `CollectionsMarshal.SetCount` rather than `RemoveRange`
- Iterate using span rather than indexer

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Core/Collections/JournalSet.cs
+++ b/src/Nethermind/Nethermind.Core/Collections/JournalSet.cs
@@ -27,13 +27,10 @@ namespace Nethermind.Core.Collections
         [SkipLocalsInit]
         public void Restore(int snapshot)
         {
-            int count = _set.Count;
-            if (snapshot >= count)
+            if (snapshot >= _set.Count)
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Core/Collections/JournalSet.cs
```diff
@@ -4,41 +4,51 @@
...
-            if (snapshot >= Count)
+            int count = _set.Count;
+            if (snapshot >= count)
             {
-                throw new InvalidOperationException($"{nameof(JournalCollection<T>)} tried to restore snapshot {snapshot} beyond current position {Count}");
+                ThrowInvalidRestore(snapshot);
             }
 
-            int current = Position;
+            if (count == 0) return;
```
LukaszRozmej, line 36: is it worth to have this as special case? I doubt it happens often?

LukaszRozmej (approved):

PR #8717, opened 2025-05-30
Fix exception on exit for incomplete nethermind startup.

asdacap:
- Fix exception on exit for incomplete nethermind startup.
- Added an interface `IStoppableService` to declare a service that can be stopped in parallel on app shutdown.
- Replace the current hardcoded stop mechanism that is causing the issue.
- Hook up dispose stack with autofac's built in disposer.

- [X] Optimization
- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Can exit early without throwing.
- Can exit without throwing.


src/Nethermind/Nethermind.Runner/Ethereum/EthereumRunner.cs
```diff
@@ -33,31 +33,19 @@ public async Task Start(CancellationToken cancellationToken)
...
-        Task blockProducerTask = Stop(() => _api.BlockProducerRunner?.StopAsync(), "Stopping block producer");
-        Task peerPoolTask = Stop(() => _api.PeerPool?.StopAsync(), "Stopping peer pool");
-        Task peerManagerTask = Stop(() => _api.PeerManager?.StopAsync(), "Stopping peer manager");
-        Task blockchainProcessorTask = Stop(() => _api.MainProcessingContext?.BlockchainProcessor?.StopAsync(), "Stopping blockchain processor");
-        Task rlpxPeerTask = Stop(() => _api.RlpxPeer?.Shutdown(), "Stopping RLPx peer");
-        await Task.WhenAll(discoveryStopTask, rlpxPeerTask, peerManagerTask, peerPoolTask, blockchainProcessorTask, blockProducerTask);
+        await serviceStopper.StopAllServices();
 
-        foreach (INethermindPlugin plugin in _api.Plugins)
+        foreach (INethermindPlugin plugin in api.Plugins)
```
LukaszRozmej, line 38: Can or should we dispose plugins in parallel?

LukaszRozmej (approved):
Does this improve shutdown time?

src/Nethermind/Nethermind.Runner/Ethereum/EthereumRunner.cs
asdacap: I thought of removing this also, but that would introduce more change unnecessary to the issue. Maybe next time.

PR #8718, opened 2025-05-30
Async code db

LukaszRozmej:
## Changes

- CodeDB shouldn't write changes immediately on `StateProvider.Commit`
- Instead it writes to a temporary in-memory cache.
- This cache can be read from
- This cache is then immediately persisted in the background - so it should be emptied very fast

- [x] Optimization

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [ ] No


LukaszRozmej:
Closing, causing too many potential issues

PR #8719, opened 2025-05-30
SpanSource for TrieNode rlp

Scooletz:
This PR reduces usage of the `CappedArray<byte>` in `TrieNode` and `PatriciaTree` and replaces it with a `SpanSource`. A span source is a construct similar to a `ValueTask` that uses a single `object` field with a type check as a discriminated union. This allows to keep `byte[]` without having them wrapped in capped array as well as perform an atomic assignment when writing `_rlp`. This in addition removes the need of the `RlpFactory` as writes of `SpanSource` are atomic. 

### Memory

Managed memory by 10-20%

image

### Throughput

image

## Changes

- `TreeNode`, `PatriciaTree` augumented
- `RlpFactory` removed
- atomically assigned `SpanSource` introduced

- [x] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


Scooletz:
Run: https://github.com/NethermindEth/nethermind/actions/runs/15349239103/job/43193152746

Scooletz:
One more run: https://github.com/NethermindEth/nethermind/actions/runs/15390014290

Scooletz:
The node has synced and is progressing. Moving forward with comparing against master:

https://github.com/NethermindEth/nethermind/actions/runs/15400416614

Scooletz:
Could I get a preliminary pre-review here?

src/Nethermind/Nethermind.Core/Buffers/ISpanSource.cs
```diff
@@ -0,0 +1,37 @@
...
+
+    /// <summary>
+    /// Like <see cref="MemoryExtensions.SequenceEqual{T}(System.ReadOnlySpan{T},System.ReadOnlySpan{T})"/>
+    /// </summary>
+    bool SequenceEqual(ReadOnlySpan<byte> other);
+
+    /// <summary>
+    /// Like <see cref="MemoryExtensions.CommonPrefixLength{T}(System.ReadOnlySpan{T},System.ReadOnlySpan{T})"/>
+    /// </summary>
+    int CommonPrefixLength(ReadOnlySpan<byte> other);
```
LukaszRozmej: Do we need those when we have `Span`? Maybe they should be extension methods?

src/Nethermind/Nethermind.Core/Buffers/ISpanSourcePool.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            return new SpanSource(new byte[size]);
+        }
+
+        return new SpanSource(pool.Rent(size));
+    }
+
+    public static void SafeReturnBuffer(this ICappedArrayPool? pool, SpanSource buffer)
+    {
+        if (pool != null && buffer.TryGetCappedArray(out CappedArray<byte> capped))
```
LukaszRozmej:
```suggestion
            return new SpanSource(new byte[size]);
        }

        return new SpanSource(pool.Rent(size));
    }

    public static void SafeReturnBuffer(this ICappedArrayPool? pool, SpanSource buffer)
    {
        if (pool is not null && buffer.TryGetCappedArray(out CappedArray<byte> capped))
```

src/Nethermind/Nethermind.Core/Buffers/TinyArray.cs
```diff
@@ -0,0 +1,338 @@
...
+            ref byte first = ref Unsafe.Add(ref _data, ValueOffset);
+            ref byte second = ref MemoryMarshal.GetReference(other);
+
+            int offset = length - Vector128<byte>.Count;
+
+            return
+                ((Vector128.LoadUnsafe(ref first) ^ Vector128.LoadUnsafe(ref second)) |
+                 (Vector128.LoadUnsafe(ref Unsafe.Add(ref first, offset)) ^
+                  Vector128.LoadUnsafe(ref Unsafe.Add(ref second, offset)))) ==
+                Vector128<byte>.Zero;
```
LukaszRozmej: Why Vector128 and not 256 or 512 for example?

src/Nethermind/Nethermind.Serialization.Rlp/AccountDecoder.cs
```diff
@@ -33,6 +33,18 @@ public AccountDecoder(bool slimFormat = false)
             return (codeHash, storageRoot);
         }
 
+        public (Hash256 CodeHash, Hash256 StorageRoot) DecodeHashesOnly(ref Rlp.ValueDecoderContext context)
```
LukaszRozmej, line 36: consider ValueHashes?

src/Nethermind/Nethermind.Trie/TrieNode.cs
```diff
@@ -925,20 +937,25 @@ public void CallRecursively(
...
                 {
-                    if (logger.IsTrace) logger.Trace($"Persist recursively on storage root {leafData.StorageRoot} of {this}");
+                    if (logger.IsTrace)
+                        logger.Trace($"Persist recursively on storage root {leafData.StorageRoot} of {this}");
                     Hash256 storagePathAddr;
                     using (currentPath.ScopedAppend(Key))
                     {
-                        if (currentPath.Length != 64) throw new Exception($"unexpected storage path length. Total nibble count should add up to 64. Got {currentPath.Length}.");
+                        if (currentPath.Length != 64)
+                            throw new Exception(
```
LukaszRozmej: We should change exceptions to some concrete type

src/Nethermind/Nethermind.Core/Buffers/ISpanSource.cs
Scooletz: The idea was to have them as a special case of `TinyArray` can provide a better performance (unrolled vectorized comparison for example). I think I'll have them removed as well as the `TinyArray` implementations.

src/Nethermind/Nethermind.Serialization.Rlp/AccountDecoder.cs
Scooletz: Good point. I was porting to match the previous. Will look for the usages.

src/Nethermind/Nethermind.Core/Buffers/ISpanSource.cs
LukaszRozmej: Sure if it is faster than keep it.

src/Nethermind/Nethermind.Trie/TrieNode.cs
Scooletz: What type? This is not a part of this PR changes 😅

src/Nethermind/Nethermind.Serialization.Rlp/AccountDecoder.cs
Scooletz: Won't do. They are used in the `StateSyncItem`.

src/Nethermind/Nethermind.Core/Buffers/TinyArray.cs
Scooletz: Totally removed.

src/Nethermind/Nethermind.Core/Buffers/ISpanSource.cs
Scooletz: Removed as the `TinyArray` is 👋 . We stay with a wrapper for the `CappedArray` and the one that wraps `byte[]`

Scooletz:
@LukaszRozmej added benchmarks of throughput. I used the RPC comparison with the same hardware. Not sure if deviations look good. Could you take a look

src/Nethermind/Nethermind.Trie/TrieNode.cs
benaadams:
```suggestion
                            throw new TrieNodeException(
```

src/Nethermind/Nethermind.Trie/TrieNode.cs
```diff
@@ -1049,7 +1066,8 @@ private async ValueTask CallRecursivelyLeafAsync(
                 Hash256 storagePathAddr;
                 using (currentPath.ScopedAppend(Key))
                 {
-                    if (currentPath.Length != 64) throw new Exception("unexpected storage path length. Total nibble count should add up to 64.");
+                    if (currentPath.Length != 64)
+                        throw new Exception("unexpected storage path length. Total nibble count should add up to 64.");
```
benaadams:
```suggestion
                        throw new TrieNodeException("unexpected storage path length. Total nibble count should add up to 64.");
```

src/Nethermind/Nethermind.Core/Buffers/SpanSource.cs
```diff
@@ -0,0 +1,214 @@
...
+    {
+        Span<byte> comparand = other.Span;
+
+        var obj = _obj;
+        if (obj is byte[] array)
+        {
+            return array.AsSpan().SequenceEqual(comparand);
+        }
+
+        return Unsafe.As<CappedArraySource>(obj).SequenceEqual(comparand);
```
benaadams, line 126: Question I'd have is `is CappedArraySource` as its a simple sealed class a faster type check than `is byte[]`

asdacap (approved):

src/Nethermind/Nethermind.Core/Buffers/SpanSource.cs
Scooletz: Let me test it. This `Unsafe.As` was put when the `ITinyArray` was in this PR, while it's no longer the case. As it's no longer the case, maybe with the `sealed` we can get as fast as this Unsafe. Will report.

Scooletz: I changed it for Span and `MemorySize`. It seems that it plays with the JIT. For the `MemorySize` it looses one jump in the `CappedArray` size. For the `Span` it no longer inlines the callsite but removes `int 3` for overflow and returns immediately after testing for being the capped array. Pushed already.

benaadams (approved):

PR #8727, opened 2025-06-03
Cleanur/remove create heath hint service

asdacap:
- `IHealthHintService` is created in `IInitializeBlockchain.CreateHealthHintService`. 
- This PR removes that method and put it in health check plugin module.
- Additionally, replace `INethermindServicesPlugin` with `IJsonRpcServiceConfigurer`, and extract it to its own class.
- Move the rest of health check plugin to autofac module.

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- [x] Health check url show something
- [x] Mainnet can start
- [x] Taiko
- [x] Optimism

### DIFF
--- a/src/Nethermind/Nethermind.Api/NethermindApi.cs
+++ b/src/Nethermind/Nethermind.Api/NethermindApi.cs
@@ -10,7 +10,6 @@ using Nethermind.Blockchain;
 using Nethermind.Blockchain.Blocks;
 using Nethermind.Blockchain.Filters;
 using Nethermind.Blockchain.Receipts;
-using Nethermind.Blockchain.Services;
 using Nethermind.Config;
 using Nethermind.Consensus;
 using Nethermind.Consensus.Comparers;
@@ -38,7 +37,6 @@ using Nethermind.JsonRpc.Modules.Subscribe;
 using Nethermind.KeyStore;
 using Nethermind.Logging;
 using Nethermind.Network;
-using Nethermind.Network.P2P.Analyzers;
 using Nethermind.Network.Rlpx;
 using Nethermind.Serialization.Json;
 using Nethermind.Specs.ChainSpecStyle;
@@ -121,7 +119,7 @@ namespace Nethermind.Api
         public CompositeBlockPreprocessorStep BlockPreprocessor { get; } = new();
         public IBlockProcessingQueue? BlockProcessingQueue { get; set; }
         public IBlockProducer? BlockProducer { get; set; }
-        public IBlockProducerRunner? BlockProducerRunner { get; set; }
+        public IBlockProducerRunner? BlockProducerRunner { get; set; } = new NoBlockProducerRunner();
         public IBlockTree? BlockTree { get; set; }
         public IBlockValidator BlockValidator => Context.Resolve<IBlockValidator>();
         public IBloomStorage? BloomStorage { get; set; }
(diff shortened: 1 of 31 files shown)


src/Nethermind/Nethermind.Api/NethermindApi.cs
```diff
@@ -121,7 +119,7 @@ public IBlockchainBridge CreateBlockchainBridge()
         public CompositeBlockPreprocessorStep BlockPreprocessor { get; } = new();
         public IBlockProcessingQueue? BlockProcessingQueue { get; set; }
         public IBlockProducer? BlockProducer { get; set; }
-        public IBlockProducerRunner? BlockProducerRunner { get; set; }
+        public IBlockProducerRunner? BlockProducerRunner { get; set; } = new NoBlockProducerRunner();
```
LukaszRozmej:
```suggestion
        public IBlockProducerRunner BlockProducerRunner { get; set; } = new NoBlockProducerRunner();
```

src/Nethermind/Nethermind.Consensus/NoBlockProducerRunner.cs
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    public bool IsProducingBlocks(ulong? maxProducingInterval)
+    {
+        return false;
+    }
+
+#pragma warning disable CS0067
+    public event EventHandler<BlockEventArgs>? BlockProduced;
+#pragma warning restore CS0067
```
LukaszRozmej:
```suggestion
    public event EventHandler<BlockEventArgs>? BlockProduced { add { } remove { } };
```

src/Nethermind/Nethermind.HealthChecks/FreeDiskSpaceChecker.cs
```diff
@@ -23,14 +25,14 @@ public class FreeDiskSpaceChecker : IHostedService, IAsyncDisposable
         private readonly double _checkPeriodMinutes;
 
         public FreeDiskSpaceChecker(IHealthChecksConfig healthChecksConfig,
-            IDriveInfo[] drives,
+            [KeyFilter(nameof(IInitConfig.BaseDbPath))] IDriveInfo[] drives,
```
LukaszRozmej, line 28: why needs key?

src/Nethermind/Nethermind.Init/Steps/RegisterRpcModules.cs
```diff
@@ -27,7 +27,7 @@
 
 namespace Nethermind.Init.Steps;
 
-[RunnerStepDependencies(typeof(InitializeNetwork), typeof(SetupKeyStore), typeof(InitializeBlockchain), typeof(InitializePlugins))]
+[RunnerStepDependencies(typeof(InitializeNetwork), typeof(SetupKeyStore), typeof(InitializeBlockchain), typeof(InitializeBlockProducer), typeof(InitializePlugins))]
```
LukaszRozmej, line 30: why additional dependencies?

asdacap: `IHealthRpcModule` previously was created at `RegisterPluginRpcModules` which has access to block producer. Now it is in `RegisterRpcModules` so need to add block producer also.

src/Nethermind/Nethermind.HealthChecks/FreeDiskSpaceChecker.cs
asdacap: List of `IDriveInfo[]` is too ambiguous for me.

src/Nethermind/Nethermind.Init/Steps/RegisterRpcModules.cs
LukaszRozmej: Not sure if it won't cause even more problems when trying to make RPC start as early as possible.

LukaszRozmej (approved):
apart of some reservation about steps dependencies looks good

PR #8731, opened 2025-06-04
Fix runner test

asdacap:
- Fix aura hive startup.
- Fix runner test not failing when startup failed.
- Fix column memdb does not populate column correctly.
- Make test run faster.

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Test passed.

### DIFF
--- a/src/Nethermind/Nethermind.Init/Steps/InitializePrecompiles.cs
+++ b/src/Nethermind/Nethermind.Init/Steps/InitializePrecompiles.cs
@@ -15,6 +15,7 @@ namespace Nethermind.Init.Steps;
 public class InitializePrecompiles : IStep
 {
     private readonly INethermindApi _api;
+    private static bool _wasSetup = false;
 
     public InitializePrecompiles(INethermindApi api)
     {
@@ -30,7 +31,9 @@ public class InitializePrecompiles : IStep
 
             try
             {
+                if (_wasSetup) return;
                 await KzgPolynomialCommitments.InitializeAsync(logger, initConfig.KzgSetupPath);
+                _wasSetup = true;
             }
             catch (Exception e)
             {
--- a/src/Nethermind/Nethermind.Optimism/OptimismPlugin.cs
+++ b/src/Nethermind/Nethermind.Optimism/OptimismPlugin.cs
@@ -180,11 +180,14 @@ public class OptimismPlugin(ChainSpec chainSpec) : IConsensusPlugin
         ArgumentNullException.ThrowIfNull(_invalidChainTracker);
         ArgumentNullException.ThrowIfNull(_blockFinalizationManager);
 
-        // Ugly temporary hack to not receive engine API messages before end of processing of all blocks after restart.
-        // Then we will wait 5s more to ensure everything is processed
-        while (!_api.BlockProcessingQueue.IsEmpty)
-            await Task.Delay(100);
-        await Task.Delay(5000);
+        if (!_api.Config<IInitConfig>().InRunnerTest)
+        {
+            // Ugly temporary hack to not receive engine API messages before end of processing of all blocks after restart.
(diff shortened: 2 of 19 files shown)


src/Nethermind/Nethermind.Init/Steps/InitializePrecompiles.cs
```diff
@@ -30,7 +31,9 @@ public async Task Execute(CancellationToken cancellationToken)
 
             try
             {
+                if (_wasSetup) return;
```
LukaszRozmej: maybe better to have some kind of static lock here?

src/Nethermind/Nethermind.Optimism/OptimismPlugin.cs
```diff
@@ -180,11 +180,14 @@ public async Task InitRpcModules()
...
-            await Task.Delay(100);
-        await Task.Delay(5000);
+        if (!_api.Config<IInitConfig>().InRunnerTest)
+        {
+            // Ugly temporary hack to not receive engine API messages before end of processing of all blocks after restart.
+            // Then we will wait 5s more to ensure everything is processed
+            while (!_api.BlockProcessingQueue.IsEmpty)
+                await Task.Delay(100);
+            await Task.Delay(5000);
+        }
```
LukaszRozmej: Maybe this should be some extension method? It is in 3 places.

LukaszRozmej (approved):

src/Nethermind/Nethermind.Optimism/OptimismPlugin.cs
asdacap: I don't know why this is here at all.

src/Nethermind/Nethermind.Init/Steps/InitializePrecompiles.cs
asdacap: switched to `Interlocked`

src/Nethermind/Nethermind.Optimism/OptimismPlugin.cs
LukaszRozmej: well read the comment :D

src/Nethermind/Nethermind.Init/Steps/InitializePrecompiles.cs
```diff
@@ -30,7 +31,10 @@ public async Task Execute(CancellationToken cancellationToken)
 
             try
             {
-                await KzgPolynomialCommitments.InitializeAsync(logger, initConfig.KzgSetupPath);
+                if (!Interlocked.CompareExchange(ref _wasSetup, true, false))
+                {
+                    await KzgPolynomialCommitments.InitializeAsync(logger, initConfig.KzgSetupPath);
+                }
```
LukaszRozmej: ok but now if there are 2 threads accessing that, only one will wait for initialization and the other will just proceed.

I think the idea here is to block both of them until initialization finishes.

LukaszRozmej (changes requested):

src/Nethermind/Nethermind.Init/Steps/InitializePrecompiles.cs
LukaszRozmej: might have made it worse

src/Nethermind/Nethermind.Init/Steps/InitializePrecompiles.cs
asdacap: Switched to semaphore

src/Nethermind/Nethermind.Optimism/OptimismPlugin.cs
asdacap: Moved to `RegisterPluginsRpcModules`.

PR #8732, opened 2025-06-04
Improve performance of CALL opcodes

benaadams:
## Changes

Rather than setting the 2 large structs (can copying) per `call` or `create` opcode:
- Only set blockcontext once per block (or tx if `eth_call` etc); moving to VirtualMachine
- Only set transactioncontext once per transaction; moving to VirtualMachine

Order parameters and constructor values in order of memory layout

- [x] Optimization
- [x] Refactoring

#### Requires testing

- [x] No


Scooletz (approved):
Lots is lost. The deferred block context creation is a plus

src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.BlockValidationTransactionsExecutor.cs
```diff
@@ -32,27 +32,30 @@ public BlockValidationTransactionsExecutor(ITransactionProcessor transactionProc
...
-            public TxReceipt[] ProcessTransactions(Block block, in BlockExecutionContext blkCtx, ProcessingOptions processingOptions, BlockReceiptsTracer receiptsTracer, IReleaseSpec spec, CancellationToken token)
+            public void SetBlockExecutionContext(in BlockExecutionContext blockExecutionContext)
+                => transactionProcessor.SetBlockExecutionContext(in blockExecutionContext);
+
+            public TxReceipt[] ProcessTransactions(Block block, ProcessingOptions processingOptions, BlockReceiptsTracer receiptsTracer, IReleaseSpec spec, CancellationToken token)
             {
                 Metrics.ResetBlockStats();
 
-                var enhanced = EnhanceBlockExecutionContext(blkCtx);
+                EnhanceBlockExecutionContext(block);
```
LukaszRozmej, line 42: Should this be called now here, or on `SetBlockExecutionContext`?

src/Nethermind/Nethermind.Consensus/Processing/IBlockProcessor.cs
```diff
@@ -51,8 +51,9 @@ Block[] Process(
 
         public interface IBlockTransactionsExecutor
         {
-            TxReceipt[] ProcessTransactions(Block block, in BlockExecutionContext blkCtx, ProcessingOptions processingOptions, BlockReceiptsTracer receiptsTracer, IReleaseSpec spec, CancellationToken token = default);
+            TxReceipt[] ProcessTransactions(Block block, ProcessingOptions processingOptions, BlockReceiptsTracer receiptsTracer, IReleaseSpec spec, CancellationToken token = default);
             event EventHandler<TxProcessedEventArgs> TransactionProcessed;
+            void SetBlockExecutionContext(in BlockExecutionContext blockExecutionContext);
```
LukaszRozmej, line 56: Here it is still only 1 call to `ProcessTransactions`, maybe keep it as it was for simplicity and change only on `TransactionProcessor` level?

src/Nethermind/Nethermind.Core.Test/AssertionsSetup.cs
```diff
@@ -20,9 +20,9 @@ public void RunBeforeAnyTests()
         AssertionOptions.AssertEquivalencyUsing(static options =>
             {
                 options
-                    .Using<Memory<byte>>(static context =>
+                    .Using<ReadOnlyMemory<byte>>(static context =>
```
LukaszRozmej, line 27: Why not both to avoid explicit casting?

src/Nethermind/Nethermind.Evm/Instruction.cs
```diff
@@ -394,12 +394,11 @@ private static bool IsValidInstruction(Instruction instruction, bool isEofContex
         };
     }
 
-    public static string? GetName(this Instruction instruction, bool isPostMerge = false, IReleaseSpec? spec = null)
+    public static string? GetName(this Instruction instruction, IReleaseSpec? spec = null)
     {
         spec ??= Frontier.Instance;
         return instruction switch
         {
-            Instruction.PREVRANDAO when !isPostMerge => "DIFFICULTY",
```
LukaszRozmej, line 402: Why remove? We do support unmerged chains

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Call.cs
```diff
@@ -247,18 +246,15 @@ public static EvmExceptionType InstructionCall<TOpCall, TTracingInst>(
         // Load call data from memory.
         ReadOnlyMemory<byte> callData = vm.EvmState.Memory.Load(in dataOffset, dataLength);
         // Construct the execution environment for the call.
-        ExecutionEnvironment callEnv = new
-        (
-            txExecutionContext: in env.TxExecutionContext,
-            callDepth: env.CallDepth + 1,
+        ExecutionEnvironment callEnv = new(
```
LukaszRozmej, line 249: would it be better if `ExecutionEnvironment` was a class?

src/Nethermind/Nethermind.Serialization.Rlp/OwnedBlockBodies.cs
```diff
@@ -42,10 +44,8 @@ public void Disown()
             foreach (Transaction tx in blockBody.Transactions)
             {
                 Hash256? _ = tx.Hash; // Just need to trigger hash calculation
-                if (tx.Data is not null)
-                {
-                    tx.Data = tx.Data.Value.ToArray();
-                }
+                // Disconnect from any backing
+                tx.Data = tx.Data.ToArray();
```
LukaszRozmej, line 48: Not like that?

```suggestion
                // Disconnect from any backing
                tx.Data ??= tx.Data.ToArray();
```

LukaszRozmej (commented):
Well I don't like a bit that now you can forget to call `SetBlockExecutionContext` and just have thing not working correctly.

src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.BlockValidationTransactionsExecutor.cs
benaadams: Is virtual method; so simulate overrides and changes the context

src/Nethermind/Nethermind.Consensus/Processing/IBlockProcessor.cs
benaadams: Can't due to system txs and other things like receipts parsing so is a multi chunk on running txs

src/Nethermind/Nethermind.Serialization.Rlp/OwnedBlockBodies.cs
benaadams: Was a nullable memory before `Memory<byte>?` which is 24 bytes vs regular `Memory<byte>` at 16 bytes; and didn't really do anything other than be annoying in code as we always then `?? default` or converted to 0-length regular memory in some way.

So change is its no longer nullable

Goal of this here is to disconnect from a portion of block into independent array, so the `.ToArray` conversion is necessary

src/Nethermind/Nethermind.Core.Test/AssertionsSetup.cs
benaadams: Added back

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Call.cs
benaadams: I think think this is right approach (struct) as can be 5k created per block (or 10k including prewarming)

image

benaadams: Although.. is always just stuffed into EvmState; so maybe just unwrap? 🤔

src/Nethermind/Nethermind.Evm/Instruction.cs
benaadams: Block header isn't coming into the tracer, seemed a minor loss?

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Call.cs
benaadams: Would be big refactor on tracers, can follow up see if like it

benaadams:
HIve test https://github.com/NethermindEth/nethermind/actions/runs/15489816102

src/Nethermind/Nethermind.Evm/Instruction.cs
LukaszRozmej: so we can't keep it?

LukaszRozmej (approved):

PR #8741, opened 2025-06-06
Refactor/Deduplicate Proof, Trace, Debug, BlockchainBridge RPC

asdacap:
- Require #8750 
- In ~~today's~~ some other day's tech chat, we will see how to use DI to create RPC method that adapt to additional plugins.
- Arbitrum is gonna add a custom `ICodeRepository`, `IVirtualMachine`, `ITransactionProcessor` and probably other components and I'd rather do this now before that.
- Remove manual Proof, Trace, Debug RPC, and `IBlockchainBridge` factory for taiko, and optimism.  
  - Automatically add Proof RPC support to optimism. 
  - Any plugin that specify `ITransactionProcessor` and `IBlockProcessor` in their plugin module would get `Proof`, `Trace`, `Debug`, and `IBlockchainBridge` implementation for free. 
  - Except for `SimulateReadOnlyBlocksProcessingEnv` related calls, but one at a time.
  - Aura uses a custom implementation.
- Double check overridable world state scope. Some RPC missed wrapping the env. This may cause memory leak as the `IWorldState` from overridable env MUST be reset. 

### In general
- Autofac is a dependency injection framework. It automatically wires component together depending on their types. Plugin can override these types to change the behaviour of nethermind. 
- These configurations are specified as a collection of `Autofac.Module` classes. 
  - At startup, plugin's can provide their own autofac module via `INethermindPlugin.Module`. 
- Note: Nethermind's module uses a custom DSL to make the registration neater.
- The main nethermind's module is the `NethermindModule`, which add other modules.
- `NethermindModule` is not complete. Many components are not available purely from DI and constructed in `IStep`(s)
   - In test, `PseudoNethermindModule` and `TestNethermindModule` is available for a convenient module that override certain things such as rocksdb, sync config and some default configuration for tests. These two module are complete. Note: This also means that certain things that is not obvious in prod will be active also such as receipt recovery on log finder which is usually not implemented in test but running in prod.
- `NethermindRunnerModule` is the actual module that is used in prod. It includes plugins, steps and backward compatibility by checking the field of `INethermindApi` when a component is missing. Preferably, plugin author should not care about this. But when you got a resolve error when running, it generally means some components are not set yet in `INethermindApi`.
- `EthereumRunnerTests` have test that start and stop nethermind. Add your plugin there for convenient test.

### Block Processing and LifetimeScope
- An Autofac's `ILifetimeScope` is a collection of classes that get created and disposed together.
  - You can create a lifetime scope, and modify the component configuration just for that lifetime.
  - This is the mechanism for creating rpc modules and block processing components as each rpc have slightly different configuration (different transaction executor, shared `IWorldState`, some `IWorldState` is within an overridable state that must be disposed).
  - Preferably components that use overridable state should not be accessible without an explicit scope clause (see `ITracerEnv`), but I'll do that later.
  - Since the configuration are hierarchical, plugin configuration (such as change in `ITransactionProcessor`) that is not changed in the rpc module, is retained, fully adapting plugin with the rpc module. 
- At `BlockProcessingModule`, the implementation of classes that involve within processing is specified. Some component is deliberately excluded such as `IWorldState`. This is deliberate, as each lifetime scope have a shared instance of these block processing components. Additionally, each RPC can have different component such as `ITransactionProcessorAdapter`.
  - Plugins that need to modify how they process block is expected to override these predefined components within their plugin module. See `OptimismModule`.

### Rpc modules
- The main entry point of rpc module in nethermind is `IRpcModuleProviders` which accept a list of `RpcModuleInfo` which contains an rpc pool which may create rpc module on start. We have not make it lazy, but we could some day.
- These are declared in `RpcModules` class.
- Plugins can register rpc modules via `RegisterSingletonJsonRpcModule` or `RegisterBoundedJsonRpcModule`. The singleton module must be thread safe, while the bounded module are modules that cannot be shared between thread.
  - This is most likely because they have some block processing component. And block processing components can't be shared because they use `IWorldState` that is not thread safe. 
  - This means, that the rpc module provider will create a fixed number of rpc module instance via the rpc module factory. It is therefore the job of the rpc module factory to create the appropriate block processing components without crashing in between components. These are:
    - `TraceModuleFactory`
    - `DebugModuleFactory`
    - `ProofModuleFacttory`.
    - `IBlockchainBridge` which is created new for each rpc module that use it.
 - If you need to modify them whole for some reason, just replace `IRpcModuleFactory<ITraceModule>` like `AutoOptimismTraceModuleFactory` which uses `BlockProductionWithdrawalProcessor` for some reason but not globally (why? Why is it not using `OptimismWithdrawalProcessor` like in `InitializeBlockchainOptimism`). 
- If you want to **add** new method, then just create a new rpc module and just add it. Multiple rpc module of the same module name can be registered at the same time.
  - I guess you can replace some method like this also.
 - Technically, I could wrap the overridable env in a nicer class so that it is very clear of its scope, but trying to make this PR smaller.

- [X] Bugfix (a non-breaking change that fixes an issue)d
- [X] Refactoringd

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes: Test modified to integrate with the factory.
- [ ] No

#### Notes on testing

- [X] Mainnet can run just fine.


src/Nethermind/Nethermind.Consensus/Tracing/Tracer.cs
```diff
@@ -45,7 +42,40 @@ public void Accept<TCtx>(ITreeVisitor<TCtx> visitor, Hash256 stateRoot) where TC
...
+        public ITracerScope RunInProcessingScope(BlockHeader block) => new TracerScope(theTracer, envSource.Build(block.StateRoot));
+
+        public ITracerScope RunInProcessingScope(BlockHeader header, Dictionary<Address, AccountOverride>? stateOverride) => new TracerScope(theTracer, envSource.BuildAndOverride(header, stateOverride));
+
+        private class TracerScope(ITracer theTracer, IOverridableTxProcessingScope scope) : ITracerScope
+        {
+            public ITracer Tracer => theTracer;
+            public void Dispose()
+            {
+                scope.Dispose();
```
asdacap, line 77: It reset here now.

src/Nethermind/Nethermind.Consensus/Tracing/Tracer.cs
```diff
@@ -45,7 +42,40 @@ public void Accept<TCtx>(ITreeVisitor<TCtx> visitor, Hash256 stateRoot) where TC
...
+
+    public interface ITracerEnv
+    {
+        ITracerScope RunInProcessingScope(BlockHeader block);
+        ITracerScope RunInProcessingScope(BlockHeader block, Dictionary<Address, AccountOverride>? stateOverride);
+    }
+
+    public interface ITracerScope : IDisposable
+    {
+        ITracer Tracer { get; }
```
asdacap, line 57: Put it in here so that it only have access to `ITracer` after an explicit `RunInProcessingScope`.

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
```diff
@@ -68,11 +66,11 @@ public BlockchainBridge(OverridableTxProcessingEnv processingEnv,
...
             ISpecProvider specProvider,
             IBlocksConfig blocksConfig,
-            bool isMining)
+            IMiningConfig miningConfig)
         {
             _processingEnv = processingEnv ?? throw new ArgumentNullException(nameof(processingEnv));
-            _blockTree = processingEnv.BlockTree;
-            _stateReader = processingEnv.StateReader;
+            _blockTree = blockTree;
+            _stateReader = stateReader;
```
asdacap, line 73: Change from the overridable state reader to the global state reader. This was likely a mistake.

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
```diff
@@ -250,7 +248,7 @@ private TransactionResult CallAndRestore(
 
             if (transaction.Nonce == 0)
             {
-                transaction.Nonce = GetNonce(stateRoot, transaction.SenderAddress);
+                transaction.Nonce = scope.StateReader.GetNonce(stateRoot, transaction.SenderAddress);
```
asdacap, line 251: See, this is within the scope, so it make sense for this, but not others.

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
```diff
@@ -433,4 +426,44 @@ public IEnumerable<FilterLog> FindLogs(LogFilter filter, CancellationToken cance
...
     }
+
+    public interface IBlockchainBridgeFactory
+    {
+        IBlockchainBridge CreateBlockchainBridge();
+    }
+
+    public class BlockchainBridgeFactory(
+        IWorldStateManager worldStateManager,
+        Func<ICodeInfoRepository> codeInfoRepositoryFunc,
```
asdacap, line 437: Using a factory because arbitrum want to change the implementation. This and the `IWorldStateManager` can be done much neater, but I'm gonna leave it like this for now.

src/Nethermind/Nethermind.JsonRpc.Test/Modules/Proof/ProofRpcModuleTests.cs
```diff
@@ -151,8 +160,8 @@ public async Task On_incorrect_params_returns_correct_error_code()
...
...000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000c0\"],\"blockHeader\":\"0xf901f9a0a3e31eb259593976b3717142a5a9e90637f614d33e2ad13f01134ea00c24ca5aa01dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347940000000000000000000000000000000000000000a056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421a009e11c477e0a0dfdfe036492b9bce7131991eb23bcf9575f9bff1e4016f90447a0e1b1585a222beceb3887dc6701802facccf186c2d0f6aa69e26ae0c431fc2b5db9010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000830f424001833d090080830f424183010203a02ba5557a4c62a513c7e56d1bf13373e0da6bec016755483e91589fe1c6d212e28800000000000003e8\"},\"id\":67}")]
```
asdacap, line 163: added `from` because receipt recovery is wired now.

src/Nethermind/Nethermind.State/IWorldStateManager.cs
```diff
@@ -50,11 +50,7 @@ public interface IWorldStateManager
 
 public interface IOverridableWorldScope
 {
-    IOverridableWorldState WorldState { get; }
+    IWorldState WorldState { get; }
```
asdacap, line 58: Split scope and world state. Make it clearer which is is the scope, which is one is the state.

src/Nethermind/Nethermind.Optimism/Rpc/RegisterOptimismRpcModules.cs
```diff
@@ -91,38 +91,4 @@ protected override void RegisterEthRpcModule(IRpcModuleProvider rpcModuleProvide
         rpcModuleProvider.RegisterBounded(optimismEthModuleFactory,
             JsonRpcConfig.EthModuleConcurrentInstances ?? Environment.ProcessorCount, JsonRpcConfig.Timeout);
     }
-
-    protected override void RegisterTraceRpcModule(IRpcModuleProvider rpcModuleProvider)
```
hudem1, line 95: can't we do the same for EthRpcModule ? i mean, making it overridable from the plugin's Load method ?

src/Nethermind/Nethermind.Taiko/TaikoPlugin.cs
```diff
@@ -298,6 +298,8 @@ protected override void Load(ContainerBuilder builder)
 
             .AddSingleton<IHealthHintService, IBlocksConfig>((blocksConfig) =>
                 new ManualHealthHintService(blocksConfig.SecondsPerSlot * 6, HealthHintConstants.InfinityHint))
+
+            .AddScoped<IValidationTransactionExecutor, TaikoBlockValidationTransactionExecutor>()
```
hudem1: no need to register debug/proof/traceModuleFactory as well ? why were they overriden if they don't introduce anything different from the base version ?

src/Nethermind/Nethermind.Init/Modules/NethermindModule.cs
```diff
@@ -37,21 +41,31 @@ protected override void Load(ContainerBuilder builder)
...
+
+            .AddSingleton<IReceiptsRecovery, IEthereumEcdsa, ISpecProvider, IReceiptConfig>((ecdsa, specProvider, receiptConfig) =>
+                new ReceiptsRecovery(ecdsa, specProvider, !receiptConfig.CompactReceiptStore))
+            .AddSingleton<IReceiptFinder, FullInfoReceiptFinder>()
             .AddSingleton<IEthereumEcdsa, ISpecProvider>((specProvider) => new EthereumEcdsa(specProvider.ChainId))
             .Bind<IEcdsa, IEthereumEcdsa>()
+
+            .AddSingleton<SimulateReadOnlyBlocksProcessingEnvFactory>()
+            .AddSingleton<IBlockchainBridgeFactory, BlockchainBridgeFactory>()
+            .AddScoped<IBlockchainBridge>((ctx) => ctx.Resolve<IBlockchainBridgeFactory>().CreateBlockchainBridge())
```
hudem1, line 67: why have this line if we already added the singleton `BlockchainBridgeFactory` above ?

src/Nethermind/Nethermind.Init/Modules/NethermindModule.cs
```diff
@@ -37,21 +41,31 @@ protected override void Load(ContainerBuilder builder)
...
             .AddSingleton<IEciesCipher, EciesCipher>()
             .AddSingleton<ICryptoRandom, CryptoRandom>()
+
+            .AddSingleton<IReceiptsRecovery, IEthereumEcdsa, ISpecProvider, IReceiptConfig>((ecdsa, specProvider, receiptConfig) =>
+                new ReceiptsRecovery(ecdsa, specProvider, !receiptConfig.CompactReceiptStore))
+            .AddSingleton<IReceiptFinder, FullInfoReceiptFinder>()
             .AddSingleton<IEthereumEcdsa, ISpecProvider>((specProvider) => new EthereumEcdsa(specProvider.ChainId))
             .Bind<IEcdsa, IEthereumEcdsa>()
+
+            .AddSingleton<SimulateReadOnlyBlocksProcessingEnvFactory>()
```
hudem1, line 65: when is this used again ?

hudem1 (approved):

src/Nethermind/Nethermind.Optimism/Rpc/RegisterOptimismRpcModules.cs
asdacap: Yes, but I'm trying to make this PR reasonably sized.

src/Nethermind/Nethermind.Taiko/TaikoPlugin.cs
asdacap: Previously they were overridden because the overridable chain env that they have change the tx and block processor and this. This is now automatic.

src/Nethermind/Nethermind.Init/Modules/NethermindModule.cs
asdacap: One is for `IBlockchainBridgeFactory` and another is for `IBlockchainBridge`. Two different thing. but its perfectly fine to just inject `IBlockchainBridgeFactory`, then call create to get the bridge, before passing it to the rpc lifecycle. That work too.

src/Nethermind/Nethermind.Init/Modules/NethermindModule.cs
asdacap: Used by blockchain bridge.

src/Nethermind/Nethermind.Optimism/Rpc/RegisterOptimismRpcModules.cs
asdacap: If #8769 is merged before this PR, I'll deduplicate `EthRpcModule` also.

asdacap: Canceled. Optimism does some weird stuff.

src/Nethermind/Nethermind.Optimism/Rpc/RegisterOptimismRpcModules.cs
```diff
@@ -91,38 +91,4 @@ protected override void RegisterEthRpcModule(IRpcModuleProvider rpcModuleProvide
...
-            _api.BlockPreprocessor,
-            _api.RewardCalculatorSource,
-            _api.ReceiptStorage,
-            _api.SpecProvider,
-            _poSSwitcher,
-            _api.LogManager,
-            _api.L1CostHelper,
-            _api.SpecHelper,
-            new Create2DeployerContractRewriter(_api.SpecHelper, _api.SpecProvider, _api.BlockTree),
-            new BlockProductionWithdrawalProcessor(new NullWithdrawalProcessor()));
```
asdacap, line 124: Why is this not `OptimismWithdrawalProcessor`?

emlautarom1: Looks like something that was missed in https://github.com/NethermindEth/nethermind/issues/8260.

It should be like

https://github.com/NethermindEth/nethermind/blob/aa2482c33ea56ddc7a72fb7503132ddadd21e0e4/src/Nethermind/Nethermind.Optimism/InitializeBlockchainOptimism.cs#L80

src/Nethermind/Nethermind.Consensus/Tracing/GethStyleTracer.cs
```diff
@@ -159,8 +159,8 @@ public IEnumerable<string> TraceBlockToFile(Hash256 blockHash, GethTraceOptions
                 throw new InvalidOperationException("Cannot trace blocks with invalid parents");
         }
 
+        using var _ = _env.BuildAndOverride(block.Header, options.StateOverrides);
```
emlautarom1, line 162: It's not immediately clear why we are constructing an anonymous disposable object. From the implementation it looks like we intend to modify the `block.Header.StateRoot`. Could we make this a bit more explicit?

src/Nethermind/Nethermind.Core/ContainerBuilderExtensions.cs
```diff
@@ -331,6 +344,75 @@ public static ContainerBuilder AddDecorator<T>(this ContainerBuilder builder, Fu
...
+    /// Use with caution.
+    /// Uses <see cref="AddSingletonWithAccessToPreviousRegistration{T}"/> to create another registration that re-instantiate
+    /// current <see cref="T"/> in an inner lifecycle that can be re-configured. The use of inner lifetime scope may
+    /// result in unexpected (or expected) behaviour. Use with caution.
+    /// </summary>
+    /// <param name="builder"></param>
+    /// <param name="configurer"></param>
+    /// <typeparam name="T"></typeparam>
+    /// <returns></returns>
+    public static ContainerBuilder UpdateSingleton<T>(this ContainerBuilder builder, Action<ContainerBuilder> configurer) where T : class
```
emlautarom1: This method is only used in tests, and it depends on `AddSingletonWithAccessToPreviousRegistration`. Given their complexity, could we move some of this work to some workaround in tests rather than having them as extensions in our DI code? My assumption is that eventually someone will start using these methods in production.

src/Nethermind/Nethermind.Optimism/Rpc/OptimismTraceModuleFactory.cs
```diff
@@ -1,66 +1,21 @@
...
-                worldStateManager.GlobalStateReader,
-                _logManager,
-                opSpecHelper,
-                contractRewriter,
-                withdrawalProcessor,
-                transactionsExecutor);
+    protected override ContainerBuilder ConfigureCommonBlockProcessing<T>(ContainerBuilder builder)
+    {
+        return base.ConfigureCommonBlockProcessing<T>(builder)
+            .AddScoped<IWithdrawalProcessor>(new BlockProductionWithdrawalProcessor(new NullWithdrawalProcessor())); // Why? Is this global all the time
```
emlautarom1:
```suggestion
            .AddSingleton<IWithdrawalProcessor>(new OptimismWithdrawalProcessor(api.worldStateManager.GlobalWorldState, api.LogManager, api.SpecHelper))
```

`OptimismWithdrawalProcessor` is stateless by itself so it should be a Singleton. Not sure how you can get its required parameters from a `ContainerBuilder` though.

asdacap: I'll move it to optimism plugin.

src/Nethermind/Nethermind.Core/ContainerBuilderExtensions.cs
asdacap: Move it to test then.

src/Nethermind/Nethermind.Consensus/Tracing/GethStyleTracer.cs
asdacap: Yea, its clearer if `_processor` is behind the created scope that is being disposed, but I'm trying to not change too much here. The processor uses world state from the overridable world state which has a temporary memory backing. This was not the case before so it was fine, but with overridable state if the scope is not properly disposed, a memory leak would happen.

emlautarom1: Understood. Maybe add a TODO or some comment pointing this out for future refactoring.

PR #8743, opened 2025-06-06
chore(evm): fix 7918 blob gas calculator

spencer-tb:
## Description

Fixes small bug within the EIP-7918 blob gas calculator.

The bug essentially triggered the reserve price 6x higher than it should, due to the incorrect variable in the calculation. We caught this with a boundary test, afaik essentially any test with the block base fee set < 48 will catch it.

## Testing

Now passes all the latest fusaka-devnet-1 EEST tests:

- https://github.com/ethereum/execution-spec-tests/releases/tag/fusaka-devnet-1%40v1.0.0


Marchhill (approved):

Marchhill:
> @Marchhill can you add test for it?

Yep will add a regression test. For now LGTM

PR #8757, opened 2025-06-10
Implement EIP-7934: RLP Execution Block Size Limit

Marchhill:
## Changes

- Implement EIP-7934
- Store encoded block size on NewPayload and when downloading from debp2p

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


LukaszRozmej (commented):
Maybe all this is an overkill? We don't really need to actually encode block RLP to get the lenght, `IRlpDecoder` has a `GetLength`, so maybe it will suffice to just calculate it instead of doing full encoding?

PR #8759, opened 2025-06-10
EIP-7939 CLZ

LukaszRozmej:
## Changes

- Implement EIP-7939 CLZ

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Core/Extensions/UInt256Extensions.cs
+++ b/src/Nethermind/Nethermind.Core/Extensions/UInt256Extensions.cs
@@ -66,7 +66,7 @@ public static class UInt256Extensions
         return Unsafe.As<Word, ValueHash256>(ref result);
     }
 
-    public static int CountLeadingZeros(this UInt256 uInt256)
+    public static int CountLeadingZeros(this in UInt256 uInt256)
     {
         // Scan from the highest limb down to the lowest
         for (int i = 3; i >= 0; i--)
--- a/src/Nethermind/Nethermind.Core/Extensions/Bytes.Vector.cs
+++ b/src/Nethermind/Nethermind.Core/Extensions/Bytes.Vector.cs
@@ -146,7 +146,7 @@ namespace Nethermind.Core.Extensions
             return result;
         }
 
-        public static int CountLeadingZeroBits(this Vector256<byte> v)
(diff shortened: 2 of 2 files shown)


src/Nethermind/Nethermind.Core/Extensions/UInt256Extensions.cs
```diff
@@ -64,4 +65,20 @@ public static ValueHash256 ToValueHash(this in UInt256 value)
 
         return Unsafe.As<Word, ValueHash256>(ref result);
     }
+
+    public static int CountLeadingZeros(this UInt256 uInt256)
+    {
+        // Scan from the highest limb down to the lowest
+        for (int i = 3; i >= 0; i--)
+        {
+            ulong limb = uInt256[i];
```
Marchhill, line 74: nit: might be easier to read if instead counted up and did `uInt256[3 - i]`

src/Nethermind/Nethermind.Core/Specs/IReleaseSpec.cs
```diff
@@ -488,5 +488,12 @@ public interface IReleaseSpec : IEip1559Spec, IReceiptSpec
         public Array? EvmInstructionsTraced { get; set; }
 
         public ProofVersion BlobProofVersion => IsEip7594Enabled ? ProofVersion.V1 : ProofVersion.V0;
+
+        /// <summary>
+        /// EIP7939 - CLZ - Clear leading zeros instruction
```
Marchhill:
```suggestion
        /// EIP-7939 - CLZ - Count leading zeros instruction
```

src/Nethermind/Nethermind.Core/Extensions/UInt256Extensions.cs
LukaszRozmej: but then you need more substractions

Marchhill: not really, we wouldn't have the `(3-i)` a few lines below. Maybe one less subtraction but pretty minor

LukaszRozmej: that 3-1 is done once, the other one would be done in every loop iteration.

PR #8765, opened 2025-06-11
eip 7939 count leading zeros optimization

LukaszRozmej:
## Changes

- Tries to avoid going to Uint256 and uses avx2

- [x] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Math1Param.cs
+++ b/src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Math1Param.cs
@@ -93,27 +93,15 @@ internal static partial class EvmInstructions
     }
 
     /// <summary>
-    /// Implements the CLZ opcode.
-    /// Counts leading 0's of 256‐bit vector
+    /// Implements the ISZERO operation.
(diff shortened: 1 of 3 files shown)


src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Math1Param.cs
```diff
@@ -93,27 +93,15 @@ public struct OpIsZero : IOpMath1Param
     }
 
     /// <summary>
-    /// Implements the CLZ opcode.
-    /// Counts leading 0's of 256‐bit vector
+    /// Implements the ISZERO operation.
+    /// Compares the input 256‐bit vector to zero and returns a predefined marker if the value is zero;
+    /// otherwise, returns a zero vector.
```
LukaszRozmej:
```suggestion
    /// Implements the CLZ opcode.
    /// Counts leading 0's of 256‐bit vector
```

PR #8767, opened 2025-06-11
`Span`-based ABI encoder/decoder

### LINKED ISSUE #7591: Optimize ABI API to have less allocation
Encoding and Decoding ABI allocates a lot, rewrite it to a version that will allocate a lot less (maybe streams?) and update usages.

emlautarom1:
Resolves #7591 

## Changes

- Alternative ABI encoding/decoding library that works with `Span` and `ReadOnlySpan`, minimizing allocations.
- Include test suite based on the ABI spec's examples: https://docs.soliditylang.org/en/latest/abi-spec.html#formal-specification-of-the-encoding
- Include benchmarks against the current ABI implementation.

- [x] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

Unit tests were added based on own judgment and examples were copied from the ABI's spec site.

Current code has ~85% coverage. The only parts that are tricky to cover are N-Tuples due to it's duplication (discussed later).

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks

The proposed ABI library (`Nethermind.Experimental.Abi`) is intended to be a replacement for the existing ABI implementation (`Nethermind.Abi`). The experimental library provides the following features:

- Minimal allocations due to the usage of `Span` for encoding and `ReadOnlySpan` for decoding.
- Fully type-safe API: based on a signature, the compiler can infer the type when encoding and decoding.
- All function calls are known ahead of time (no `virtual` or `override`).
- Despite the extensive usage of lambdas, no closures are required (all lambdas are marked as `static`).

---

Unlike the current library, this does not support the non-standard `packed` mode (https://docs.soliditylang.org/en/latest/abi-spec.html#non-standard-packed-mode). I don't think this could be problematic considering that there is a single usage of this format outside tests, which can easily be replaced by very short custom code:

https://github.com/NethermindEth/nethermind/blob/c1dcc73f35e797c0cb30ddbaf05d9a39b33dc27c/src/Nethermind/Nethermind.Facade/Simulate/SimulateTxMutatorTracer.cs#L53-L54

Another downside is the code duplication for tuples: C# does not provide (as far as I know) tools to generalize over tuples of different sizes and different types (`(T1, T2, ..., Tn)`), and the way the ABI format is specified limits any possibility of code reuse without resorting to reflection or excessive boxing. This results in duplicated code as can be seen in:

https://github.com/NethermindEth/nethermind/blob/8c8c48ef78f65ca496633d2068c06ef929b0817d/src/Nethermind/Nethermind.Experimental.Abi/AbiType.Tuple.cs#L6

This code duplication could be addressed with Source Generators, or we could manually write it once and forget about it.


emlautarom1:
[Benchmarks are included in the PR](https://github.com/NethermindEth/nethermind/blob/8c8c48ef78f65ca496633d2068c06ef929b0817d/src/Nethermind/Nethermind.Abi.Benchmarks/results.md). On my machine, results are as follow:

```
BenchmarkDotNet v0.15.0, Linux Fedora Linux 42 (Workstation Edition)
AMD Ryzen 9 7950X3D 5.76GHz, 1 CPU, 32 logical and 16 physical cores
.NET SDK 9.0.106
  [Host]     : .NET 9.0.2 (9.0.225.6610), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI
  DefaultJob : .NET 9.0.5 (9.0.525.21509), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI
```

---

### F

| Method       |     Mean |     Error |    StdDev | Ratio |   Gen0 | Allocated | Alloc Ratio |
| ------------ | -------: | --------: | --------: | ----: | -----: | --------: | ----------: |
| Current      | 1.396 μs | 0.0081 μs | 0.0075 μs |  1.00 | 0.0744 |   3.69 KB |        1.00 |
| Experimental | 1.150 μs | 0.0083 μs | 0.0077 μs |  0.82 | 0.0401 |   1.99 KB |        0.54 |

### F_Encode

| Method       |       Mean |   Error |  StdDev | Ratio |   Gen0 | Allocated | Alloc Ratio |
| ------------ | ---------: | ------: | ------: | ----: | -----: | --------: | ----------: |
| Current      | 1,082.0 ns | 4.65 ns | 4.12 ns |  1.00 | 0.0591 |   2.91 KB |        1.00 |
| Experimental |   689.0 ns | 2.21 ns | 1.84 ns |  0.64 | 0.0334 |   1.66 KB |        0.57 |

### G

| Method       |     Mean |     Error |    StdDev | Ratio |   Gen0 | Allocated | Alloc Ratio |
| ------------ | -------: | --------: | --------: | ----: | -----: | --------: | ----------: |
| Current      | 2.618 μs | 0.0080 μs | 0.0071 μs |  1.00 | 0.1640 |   8.21 KB |        1.00 |
| Experimental | 1.460 μs | 0.0040 μs | 0.0033 μs |  0.56 | 0.0591 |   2.95 KB |        0.36 |

### G_Encode_Precomputed

| Method       |       Mean |   Error |  StdDev | Ratio |   Gen0 | Allocated | Alloc Ratio |
| ------------ | ---------: | ------: | ------: | ----: | -----: | --------: | ----------: |
| Current      | 1,342.5 ns | 4.20 ns | 3.73 ns |  1.00 | 0.1125 |    5664 B |        1.00 |
| Experimental |   175.7 ns | 0.42 ns | 0.39 ns |  0.13 |      - |         - |        0.00 |

emlautarom1:
Right now, the library is incomplete due missing types (ex. signed integers, `fixed`, `address`), but these are quite easy to implement if there is enough interest.

src/Nethermind/Nethermind.Experimental.Abi/BinarySpan.cs
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    public void Scoped<TCtx>(TCtx ctx, BinarySpanWriterAction<TCtx> inner)
+    {
+        var writer = new BinarySpanWriter(_span[_position..]);
+        inner(ctx, ref writer);
+        _position += writer._position;
+    }
+
+    public void WriteOffset<TCtx>(int offset, TCtx ctx, BinarySpanWriterAction<TCtx> inner)
```
Scooletz: What could be the context here? Small value tuples or some big structs? Have you considered `in TCtx` maybe?

src/Nethermind/Nethermind.Experimental.Abi/BinarySpan.cs
```diff
@@ -0,0 +1,133 @@
...
+    {
+        Span<byte> offsetLocation = _span[offset..(offset + 32)];
+        var position = (UInt256)_position;
```
Scooletz: You know that position is a positive int. Write it manually as big endian and zero out the rest. No need to cast up only to big endian it later.

src/Nethermind/Nethermind.Experimental.Abi/AbiType.cs
```diff
@@ -0,0 +1,255 @@
...
+            var bytes = r.ReadBytesPadded(length);
+            return Encoding.UTF8.GetString(bytes);
+        },
+        Write = (ref BinarySpanWriter w, String v) =>
+        {
+            Span<byte> buffer = stackalloc byte[Encoding.UTF8.GetByteCount(v)];
+            Encoding.UTF8.GetBytes(v, buffer);
+            int length = buffer.Length;
+
+            UInt256.Write(ref w, (UInt256)length);
```
Scooletz: You'll never have a string bigger than int. Consider writing it manually maybe?

src/Nethermind/Nethermind.Experimental.Abi/AbiType.cs
```diff
@@ -0,0 +1,255 @@
...
+        }
+    };
+
+    public static IAbi<String> String => new()
+    {
+        Name = "string",
+        IsDynamic = true,
+        Read = (ref BinarySpanReader r) =>
+        {
+            int length = (int)UInt256.Read(ref r);
```
Scooletz, line 228: You'll never have a string bigger than int. Consider reading it manually maybe?

src/Nethermind/Nethermind.Experimental.Abi/AbiType.cs
```diff
@@ -0,0 +1,255 @@
...
+            v.ToBigEndian(bytes);
+            w.Write(bytes);
+        },
+        Size = _ => 32,
+    };
+
+    public static readonly IAbi<UInt64> UInt64 = new()
+    {
+        Name = "uint64",
+        Read = (ref BinarySpanReader r) => (UInt64)UInt256.Read(ref r),
```
Scooletz, line 30: Could we specialize these methods and read Uint64 only and have a check that the rest is 0?

src/Nethermind/Nethermind.Experimental.Abi/AbiType.cs
```diff
@@ -0,0 +1,255 @@
...
+        Write = (ref BinarySpanWriter w, UInt64 v) => UInt256.Write(ref w, v),
+        Size = _ => 32,
+    };
+
+    public static readonly IAbi<UInt32> UInt32 = new()
+    {
+        Name = "uint32",
+        Read = (ref BinarySpanReader r) => (UInt32)UInt256.Read(ref r),
```
Scooletz, line 38: Could we specialize these methods and read Uint32 only and have a check that the rest is 0?

src/Nethermind/Nethermind.Experimental.Abi/AbiType.cs
```diff
@@ -0,0 +1,255 @@
...
+        Read = (ref BinarySpanReader r) =>
+        {
+            var bytes = r.ReadBytes(32);
+            return new UInt256(bytes, isBigEndian: true);
+        },
+        Write = (ref BinarySpanWriter w, UInt256 v) =>
+        {
+            Span<byte> bytes = stackalloc byte[32];
+            v.ToBigEndian(bytes);
+            w.Write(bytes);
```
Scooletz: Provide a method renting buffer from the `BinarySpanWriter`, a similar to https://learn.microsoft.com/en-us/dotnet/api/system.buffers.ibufferwriter-1.getspan?view=net-9.0#system-buffers-ibufferwriter-1-getspan(system-int32) . I'd name it `.GetAndAdvance` and you'd use it here to `v.ToBigEndian(w.GetAndAvance(32))`. No double copy. This smells like a good potential improvement.

src/Nethermind/Nethermind.Experimental.Abi/AbiType.cs
```diff
@@ -0,0 +1,255 @@
...
+        Name = "uint8",
+        Read = (ref BinarySpanReader r) => (Byte)UInt256.Read(ref r),
+        Write = (ref BinarySpanWriter w, Byte v) => UInt256.Write(ref w, v),
+        Size = _ => 32,
+    };
+
+    public static readonly IAbi<Boolean> Bool = new()
+    {
+        Name = "bool",
+        Read = (ref BinarySpanReader r) => UInt256.Read(ref r) != 0,
```
Scooletz, line 62: I'd specialize it.

Scooletz (commented):
A few remarks, mostly about specialing implementations.

src/Nethermind/Nethermind.Experimental.Abi/BinarySpan.cs
emlautarom1: Most usages of context are for small tuples (4 elements at most right now) in order to avoid closures.

I'll look into the usage of `in`.

Scooletz: If you mean `ValueTuple` like https://learn.microsoft.com/en-us/dotnet/api/system.valuetuple-4?view=net-9.0 then you don't want to pass by value probably. As usual, benchmark will tell the truth.

Scooletz:
@emlautarom1 The gains after introducing `.Take`, right? 😄

src/Nethermind/Nethermind.Experimental.Abi/Abi.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.Diagnostics;
+using Nethermind.Core.Extensions;
+
+namespace Nethermind.Experimental.Abi;
+
+public class AbiException : Exception;
+
+public static class Abi
+{
+    public static byte[] Encode<T>(AbiSignature<T> signature, T arg)
```
LukaszRozmej: Can we make it ?
```suggestion
    public static void Encode<T>(AbiSignature<T> signature, T arg, Span<byte> encoded)
```

src/Nethermind/Nethermind.Experimental.Abi/Abi.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        w.Write(signature.MethodId());
+        signature.Abi.Write(ref w, arg);
+
+        Debug.Assert(w.Written == buffer.Length, "Abi encoding did not write the expected number of bytes");
+
+        return buffer;
+    }
+
+    public static T Decode<T>(AbiSignature<T> signature, byte[] source)
```
LukaszRozmej:
```suggestion
    public static T Decode<T>(AbiSignature<T> signature, ReadOnlySpan<byte> source)
```

src/Nethermind/Nethermind.Experimental.Abi/Abi.cs
LukaszRozmej: We can have helper method that will preallocate.
But with that you can use stackalloc from outside, you can use `ArrayPool`, you can avoid allocations at all

src/Nethermind/Nethermind.Experimental.Abi/IAbi.cs
```diff
@@ -0,0 +1,19 @@
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+namespace Nethermind.Experimental.Abi;
+
+public delegate T IAbiReadFunc<out T>(ref BinarySpanReader r);
+public delegate void IAbiWriteAction<in T>(ref BinarySpanWriter w, T value);
+public delegate int IAbiSizeFunc<in T>(T value); // TODO: Use `UInt256` when dealing with sizes according to the ABI spec
+
+public class IAbi<T>
```
LukaszRozmej: `I` prefix makes me thing it is interface?
```suggestion
public class Abi<T>
```

src/Nethermind/Nethermind.Experimental.Abi/IAbi.cs
```diff
@@ -0,0 +1,19 @@
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+namespace Nethermind.Experimental.Abi;
+
+public delegate T IAbiReadFunc<out T>(ref BinarySpanReader r);
+public delegate void IAbiWriteAction<in T>(ref BinarySpanWriter w, T value);
+public delegate int IAbiSizeFunc<in T>(T value); // TODO: Use `UInt256` when dealing with sizes according to the ABI spec
```
LukaszRozmej:
```suggestion
public delegate T AbiReadFunc<out T>(ref BinarySpanReader r);
public delegate void AbiWriteAction<in T>(ref BinarySpanWriter w, T value);
public delegate int AbiSizeFunc<in T>(T value); // TODO: Use `UInt256` when dealing with sizes according to the ABI spec
```

LukaszRozmej (commented):
Overall I like it very much and I think we could ship it

src/Nethermind/Nethermind.Abi.Benchmarks/F.cs
```diff
@@ -0,0 +1,54 @@
...
+    {
+        var signature = new Exp.AbiSignature("f")
+            .With(
+                Exp.AbiType.UInt256,
+                Exp.AbiType.Array(Exp.AbiType.UInt32),
+                Exp.AbiType.BytesM(10),
+                Exp.AbiType.Bytes);
+
+        (UInt256, uint[], byte[], byte[]) arguments = (0x123, [0x456, 0x789], "1234567890"u8.ToArray(),
+            "Hello, world!"u8.ToArray());
```
LukaszRozmej, line 47: This makes your benchmark bad, prepare data outside benchmark, so we can compare allocations properly.

src/Nethermind/Nethermind.Abi.Benchmarks/F.cs
```diff
@@ -0,0 +1,54 @@
...
+        byte[] encoded = Exp.Abi.Encode(signature, arguments);
+
+        var decoded = Exp.Abi.Decode(signature, encoded);
```
LukaszRozmej: When decoding sequence, might be good to have an API, that would allow you to decode next item only instead the whole sequence

Also if the sequence is array of simple type, would be good to have a way to put a buffer to decode to.

Would allow to decode to `Span<byte>` ,`Span<int>` ect without allocating

LukaszRozmej: might be a tradeoff of more complex code for decoding, but with the upside of no-alloc

src/Nethermind/Nethermind.Experimental.Abi/Abi.cs
emlautarom1: The idea of the `Abi` class is to have a "simple" default that could be used during a migration from the existing API. If you want more control over the output buffer you can use the `IAbi.Write` method directly as we do in one of the benchmarks:

https://github.com/NethermindEth/nethermind/blob/f2a15bd85f5d896aad2b8c7a65b2a456f0a1ab29/src/Nethermind/Nethermind.Abi.Benchmarks/G_Encode_Precomputed.cs#L47-L48

Nevertheless, we can add some extra overloads in the `Abi` class for "common" use cases, but I would like to try it out and see what we need before committing to any particular API.

src/Nethermind/Nethermind.Experimental.Abi/Abi.cs
emlautarom1: Not sure why it was `byte[]` in the first place. Adjusted.

src/Nethermind/Nethermind.Abi.Benchmarks/F.cs
emlautarom1: Indeed, the run time gets affected by the data preparation, but the intention was to also compare the required boxing of the existing API.

There are other tests that are more accurate regarding to just encoding where we have both the signature and parameters prepared, like `G_Encode_Precomputed`:

https://github.com/NethermindEth/nethermind/blob/f2a15bd85f5d896aad2b8c7a65b2a456f0a1ab29/src/Nethermind/Nethermind.Abi.Benchmarks/G_Encode_Precomputed.cs#L44-L51

src/Nethermind/Nethermind.Abi.Benchmarks/F.cs
emlautarom1: I've been thinking about that, and there are some pathological cases where it would be really nice: for example, a method `g(uint8, byte[])` where the first parameter is a version code, and the second is some sort of encoding. With this API (and the existing one, AFAIK) we need to decode everything (which requires allocating an array for the second parameter) even in cases where we only care about a single parameter (we might want to know that the version code is `== 42`).

The design for that would be something akin to lenses (https://www.youtube.com/watch?v=3kduOmZ2Wxw, I don't have a good resource in C#), where we can compose decoders until we can decode exactly what we're interested in. Unfortunately I haven't figured out how that would look like so it's something to tackle in a different PR.

src/Nethermind/Nethermind.Experimental.Abi/IAbi.cs
emlautarom1: It's an artifact from when it was actually an interface. Adjusted.

src/Nethermind/Nethermind.Abi.Benchmarks/F.cs
LukaszRozmej: well easy way is to decode to lazy iterator instead of an array, iterator can be a struct so it won't allocate.

stdevMac:
@emlautarom1 stale-PR check: this was left at 'if there is enough interest' and tracking issue #7591 preserves the intent. OK to close the PR itself? If there's no reply in the next few days, we'll close this as part of the stale-PR cleanup — it can always be reopened or resubmitted.

emlautarom1:
@stdevMac Feel free to close but do not delete the branch.

LukaszRozmej:
Keeping this open as a draft rather than closing it — thanks for the note about the branch, but the review thread here is worth more than the branch alone, so it stays findable.

**Status for anyone picking this up:**

- The approach is sound and the benchmarks in the PR are favourable. My view has not changed from the review: this is shippable in principle.
- What blocks merge is completeness, not design: signed integers, `fixed`, and `address` are not implemented yet, and non-standard packed mode is unsupported. The single production use of packed mode is in `SimulateTxMutatorTracer`, and the PR notes it is easy to replace.
- The N-tuple duplication is the other loose end — either generate it or write it out once deliberately.
- It conflicts with master now, so the benchmarks want re-running against a current baseline before the numbers are quoted anywhere.

Tracked by #7591. Draft, not to be deleted.

PR #8769, opened 2025-06-12
Refactor/taiko extended eth module

asdacap:
- Previously it is not possible to have multiple class for the same rpc module name.
- Taiko then had to replace the whole eth rpc module.
- This PR split the extra RPC into its own interface `ITaikoExtendedEthRpcModule`.
- This remove some class and code.

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- `taiko_getSyncMode` can be called.


src/Nethermind/Nethermind.Era1/EraModule.cs
```diff
@@ -31,15 +31,15 @@ protected override void Load(ContainerBuilder builder)
 
             // The admin export/import history method is here
             .RegisterSingletonJsonRpcModule<IEraAdminRpcModule, EraAdminRpcModule>()
-            ;
 
-        builder.RegisterBuildCallback((ctx) =>
-        {
-            IEraConfig eraConfig = ctx.Resolve<IEraConfig>();
-            if (string.IsNullOrWhiteSpace(eraConfig.NetworkName))
+            .AddDecorator<IEraConfig>((ctx, eraConfig) =>
```
asdacap, line 35: So that `ISpecProvider` is not always resolved. Taiko replace `ISpecProvider` with its own implementation.

LukaszRozmej (approved):

PR #8773, opened 2025-06-12
Include Debug Symbols

benaadams:
## Changes

- Include debug symbols in build, so traces resolve

#### Requires testing

- [x] No


rubo (commented):
We explicitly turn it off when publishing Nethermind. Why do we need it now?

benaadams:
> > Because the traces from machines in the wild for investigation are very unhelpful
> 
> It should then be in the build script, not in general project settings, as embedding negatively affects incremental build performance.

I might need to defer to your knowledge of the build script on this one 🙏

PR #8781, opened 2025-06-13
Recover signer direct from keccak without intermediary array

benaadams:
## Changes

- Encode to KeccakRlpStream rather than RlpStream, create array, keccak array

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Crypto/EthereumEcdsaExtensions.cs
+++ b/src/Nethermind/Nethermind.Crypto/EthereumEcdsaExtensions.cs
@@ -12,6 +12,7 @@ namespace Nethermind.Crypto
 {
     public static class EthereumEcdsaExtensions
     {
+        private static readonly TxDecoder _txDecoder = TxDecoder.Instance;
         public static AuthorizationTuple Sign(this IEthereumEcdsa ecdsa, PrivateKey signer, ulong chainId, Address codeAddress, ulong nonce)
         {
             using NettyRlpStream rlp = AuthorizationTupleDecoder.Instance.EncodeWithoutSignature(chainId, codeAddress, nonce);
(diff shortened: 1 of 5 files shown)


src/Nethermind/Nethermind.Crypto/EthereumEcdsaExtensions.cs
```diff
@@ -75,9 +76,11 @@ public static bool Verify(this IEthereumEcdsa ecdsa, Address sender, Transaction
                 TxType.Legacy => ecdsa.ChainId,
                 _ => tx.ChainId!.Value,
             };
-            Hash256 hash = Keccak.Compute(Rlp.Encode(tx, true, applyEip155, chainId).Bytes);
 
-            return ecdsa.RecoverAddress(tx.Signature, hash);
+            KeccakRlpStream stream = new();
```
LukaszRozmej, line 80: should we pool those?

src/Nethermind/Nethermind.Crypto/EthereumEcdsaExtensions.cs
```diff
@@ -75,9 +76,11 @@ public static bool Verify(this IEthereumEcdsa ecdsa, Address sender, Transaction
                 TxType.Legacy => ecdsa.ChainId,
                 _ => tx.ChainId!.Value,
             };
-            Hash256 hash = Keccak.Compute(Rlp.Encode(tx, true, applyEip155, chainId).Bytes);
```
LukaszRozmej, line 78: should tx pool use keccak cache?

benaadams: No as txs are too large to fit in the cache (cache needs to store value for equality, so is limited to 64bytes)

src/Nethermind/Nethermind.Crypto/EthereumEcdsaExtensions.cs
benaadams: Are fairly lightweight are doing pooling internally

PR #8789, opened 2025-06-15
Move high volume logs from Debug to Trace

benaadams:
## Changes

- Regular operation activity messages should be moved from Debug to Trace, otherwise it makes Debug traces 90% noise and hard to use to diagnose issues for Ops.

As requested by @barnabasbusa

- [x] Optimization
- [x] Other: Logging

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
+++ b/src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
@@ -225,7 +225,7 @@ public class JsonRpcProcessor : IJsonRpcProcessor
                 // Handles a single JSON RPC request.
                 if (model is not null)
                 {
-                    if (_logger.IsDebug) _logger.Debug($"JSON RPC request {model}");
+                    if (_logger.IsTrace) _logger.Trace($"JSON RPC request {model}");
 
                     // Processes the individual request.
                     JsonRpcResult.Entry result = await HandleSingleRequest(model, context);
@@ -238,7 +238,7 @@ public class JsonRpcProcessor : IJsonRpcProcessor
                 // Processes a collection of JSON RPC requests.
                 if (collection is not null)
                 {
-                    if (_logger.IsDebug) _logger.Debug($"{collection.Count} JSON RPC requests");
+                    if (_logger.IsTrace) _logger.Trace($"{collection.Count} JSON RPC requests");
 
                     // Checks for authentication and batch size limit.
                     if (!context.IsAuthenticated && collection.Count > _jsonRpcConfig.MaxBatchSize)
@@ -265,7 +265,7 @@ public class JsonRpcProcessor : IJsonRpcProcessor
                     errorResponse.AddDisposable(() => jsonDocument.Dispose());
 
                     TraceResult(errorResponse);
-                    if (_logger.IsDebug) _logger.Debug($"  Failed request handled in {Stopwatch.GetElapsedTime(startTime).TotalMilliseconds:N0}ms");
+                    if (_logger.IsTrace) _logger.Trace($"  Failed request handled in {Stopwatch.GetElapsedTime(startTime).TotalMilliseconds:N0}ms");
                     deserializationFailureResult = JsonRpcResult.Single(RecordResponse(errorResponse, new RpcReport("# parsing error #", (long)Stopwatch.GetElapsedTime(startTime).TotalMilliseconds, false)));
                     yield return deserializationFailureResult.Value;
                     break;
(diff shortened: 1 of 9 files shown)


barnabasbusa (approved):

rubo (approved):

src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
```diff
@@ -225,7 +225,7 @@ JsonRpcResult GetParsingError(ref readonly ReadOnlySequence<byte> buffer, string
                 // Handles a single JSON RPC request.
                 if (model is not null)
                 {
-                    if (_logger.IsDebug) _logger.Debug($"JSON RPC request {model}");
+                    if (_logger.IsTrace) _logger.Trace($"JSON RPC request {model}");
```
asdacap: All the log in `JsonRpcProcessor` here should be debug.

src/Nethermind/Nethermind.JsonRpc/JsonRpcService.cs
```diff
@@ -90,7 +90,7 @@ private async Task<JsonRpcResponse> ExecuteAsync(JsonRpcRequest request, string
     {
         JsonElement providedParameters = request.Params;
 
-        LogRequest(methodName, providedParameters, method.ExpectedParameters);
+        if (_logger.IsTrace) LogRequest(methodName, providedParameters, method.ExpectedParameters);
```
asdacap, line 93: Should be debug IMO

src/Nethermind/Nethermind.Network/ProtocolsManager.cs
```diff
@@ -130,7 +130,7 @@ private void SessionDisconnected(object sender, DisconnectEventArgs e)
                 _txPool.RemovePeer(removed.Node.Id);
                 if (session.BestStateReached == SessionState.Initialized)
                 {
-                    if (_logger.IsDebug) _logger.Debug($"{session.Direction} {session.Node:s} disconnected {e.DisconnectType} {e.DisconnectReason} {e.Details}");
+                    if (_logger.IsTrace) _logger.Trace($"{session.Direction} {session.Node:s} disconnected {e.DisconnectType} {e.DisconnectReason} {e.Details}");
```
asdacap: If it happen once per connection, it probbaly should be debug.

src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
benaadams: These are literally the main logs being complained about 😅

The point is there is nothing unusual happening and these logs are megabytes of hex; so any error info is completely drowned by the over logging

src/Nethermind/Nethermind.JsonRpc/JsonRpcService.cs
benaadams: What should Debug be showing and what should Trace be doing?

Debug should be more issues that aren't errors; like timeouts, disconnects etc? Then trace more info about business as usual and this fits into that camp?

Question, do you use debug or trace logs?

src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
asdacap: Then the problem is the body/response. But there should still be a debug log when an rpc happen.

src/Nethermind/Nethermind.JsonRpc/JsonRpcService.cs
asdacap: IMO Debug should show up when something meaningful happen but we dont want to show it to the end user. It is something that dev can reasonably have turned on all the time at least in file for bugs that is very hard to replicate. Eg, we want to know when an rpc started and when it ended. Trace is when we dont have a choice.

src/Nethermind/Nethermind.JsonRpc/JsonRpcService.cs
```diff
@@ -233,7 +233,7 @@ private async Task<JsonRpcResponse> ExecuteAsync(JsonRpcRequest request, string
 
     private void LogRequest(string methodName, JsonElement providedParameters, ExpectedParameter[] expectedParameters)
     {
-        if (_logger.IsDebug && !_methodsLoggingFiltering.Contains(methodName))
+        if (!_methodsLoggingFiltering.Contains(methodName))
```
LukaszRozmej:
```suggestion
        if _logger.IsTrace &&(!_methodsLoggingFiltering.Contains(methodName))
```

LukaszRozmej (approved):
I'm fine either way and optimizing for what makes sense as "users", but discuss with @asdacap

MarekM25 (approved):

src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
kamilchodola: What if when in Debug mode RPC will be flushed but to separate files?

kamilchodola (approved):

src/Nethermind/Nethermind.JsonRpc/JsonRpcService.cs
benaadams: This one is Trace in LogRequest currently; and just moving it out so it doesn't call method and then exist

PR #8796, opened 2025-06-17
Base prewarming on logical cores

benaadams:
## Changes

- Base prewarming on logical cores rather than physical (max at 16); regression which switches prewarming off

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
@@ -24,7 +24,7 @@ namespace Nethermind.Consensus.Processing;
 
 public sealed class BlockCachePreWarmer(IReadOnlyTxProcessingEnvFactory envFactory, IWorldState worldStateToWarmup, ISpecProvider specProvider, int concurrency, ILogManager logManager, PreBlockCaches? preBlockCaches = null) : IBlockCachePreWarmer
 {
-    private int _concurrencyLevel = (concurrency == 0 ? Environment.ProcessorCount - 1 : concurrency);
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
```diff
@@ -25,7 +24,7 @@ namespace Nethermind.Consensus.Processing;
 
 public sealed class BlockCachePreWarmer(IReadOnlyTxProcessingEnvFactory envFactory, IWorldState worldStateToWarmup, ISpecProvider specProvider, int concurrency, ILogManager logManager, PreBlockCaches? preBlockCaches = null) : IBlockCachePreWarmer
 {
-    private int _concurrencyLevel = (concurrency == 0 ? RuntimeInformation.PhysicalCoreCount - 1 : concurrency);
+    private int _concurrencyLevel = (concurrency == 0 ? Math.Min(Environment.ProcessorCount - 1, 16) : concurrency);
```
LukaszRozmej, line 27: PR description says it maxes out on 16, while you Min 16 here?

LukaszRozmej: On the other hand I see potential go over 16 in some extreme cases

benaadams: Min(cores, 16) => max 16

benaadams: > On the other hand I see potential go over 16 in some extreme cases

Issue is its logical cores not physical and is very heavy on memory bandwidth (which is why it was using physical cores before)

PR #8797, opened 2025-06-17
Fix adaptive code resolver

marcindsobczak:
## Changes

- if protocol version is updated during session, resolver is not updated, because key was just `ProtocolCode`. Changed to `ProrocolCode/ProtocolVersion`.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Network/P2P/Session.cs
+++ b/src/Nethermind/Nethermind.Network/P2P/Session.cs
@@ -589,7 +589,7 @@ namespace Nethermind.Network.P2P
 
         private AdaptiveCodeResolver GetOrCreateResolver()
         {
-            string key = string.Join(":", _protocols.Select(static p => p.Key).OrderBy(static x => x).ToArray());
+            string key = string.Join(":", _protocols.Select(static p => string.Join("/", p.Value.ProtocolCode, p.Value.ProtocolVersion)).OrderBy(static x => x).ToArray());
             if (!_resolvers.TryGetValue(key, out AdaptiveCodeResolver? value))
             {
                 value = new AdaptiveCodeResolver(_protocols);


LukaszRozmej (approved):

src/Nethermind/Nethermind.Network/P2P/Session.cs
```diff
@@ -589,7 +589,7 @@ public void AddProtocolHandler(IProtocolHandler handler)
 
         private AdaptiveCodeResolver GetOrCreateResolver()
         {
-            string key = string.Join(":", _protocols.Select(static p => p.Key).OrderBy(static x => x).ToArray());
+            string key = string.Join(":", _protocols.Select(static p => string.Join("/", p.Value.ProtocolCode, p.Value.ProtocolVersion)).OrderBy(static x => x).ToArray());
```
LukaszRozmej: with nested string.join, maybe better to use a stringbuilder?

src/Nethermind/Nethermind.Network/P2P/Session.cs
```diff
@@ -588,8 +588,8 @@ public void AddProtocolHandler(IProtocolHandler handler)
 
         private AdaptiveCodeResolver GetOrCreateResolver()
         {
-            string key = string.Join(":", _protocols.Select(static p => p.Key).OrderBy(static x => x).ToArray());
-            if (!_resolvers.TryGetValue(key, out AdaptiveCodeResolver? value))
+            string key = string.Join(":", _protocols.Select(static p => p.Value.Name).OrderBy(static x => x));
+            if (!_resolvers.TryGetValue(key, out AdaptiveCodeResolver value))
```
alexb5dh, line 592: Still has a potential race condition here - `_protocols` is used twice (for key building and in `AdaptiveCodeResolve` ctor) and may theoretically change between calls. 

Seems like this was present for a long time, but didn't cause any noticeable issues, so may be ok to left as is. Can be handled, but will require either more allocations (making a dictionary copy) or a more complex code (probably something like Compare-and-Swap loop).

benaadams: Added atomic add

PR #8801, opened 2025-06-18
Skip recovery queue when no blocks queued

benaadams:
## Changes

- When no blocks queued (regular processing) skip the recovery queue and queue direct to the block processing queue to reduce latency

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
@@ -140,11 +140,11 @@ public sealed class BlockchainProcessor : IBlockchainProcessor, IBlockProcessing
         if (!_recoveryComplete)
         {
             Interlocked.Increment(ref _queueCount);
+            _lastProcessedBlock = DateTime.UtcNow;
             try
             {
                 if (blockRef.Resolve(_blockTree))
                 {
-                    _lastProcessedBlock = DateTime.UtcNow;
                     if (_logger.IsTrace) _logger.Trace($"A new block {block.ToString(Block.Format.Short)} enqueued for processing.");
                     if (_queueCount > 1)
                     {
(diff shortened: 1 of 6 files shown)


src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
```diff
@@ -203,48 +221,40 @@ private async Task RunRecovery()
...
         if (_logger.IsDebug) _logger.Debug($"Starting recovery loop - {_blockQueue.Reader.Count} blocks waiting in the queue.");
         _lastProcessedBlock = DateTime.UtcNow;
         await foreach (BlockRef blockRef in _recoveryQueue.Reader.ReadAllAsync(CancellationToken))
         {
             try
             {
-                if (blockRef.Resolve(_blockTree))
+                Interlocked.Add(ref _currentRecoveryQueueSize, -blockRef.Block!.Transactions.Length);
+                if (_logger.IsTrace) _logger.Trace($"Recovering addresses for block {blockRef.BlockHash}.");
+                _recoveryStep.RecoverData(blockRef.Block);
```
LukaszRozmej, line 241: This won't be run when skipping recovery?
Looks like potential issue

src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
```diff
@@ -142,10 +140,30 @@ public void Enqueue(Block block, ProcessingOptions processingOptions)
...
+                if (blockRef.Resolve(_blockTree))
+                {
+                    if (_logger.IsTrace) _logger.Trace($"A new block {block.ToString(Block.Format.Short)} enqueued for processing.");
+                    if (_queueCount > 1)
+                    {
+                        _recoveryQueue.Writer.TryWrite(blockRef);
+                    }
+                    else
+                    {
+                        // Skip recovery queue if nothing in queue
```
LukaszRozmej, line 155: sure but run recovery step somewhere?

src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
benaadams: Is run in `PrepareBlocksToProcess` for full chain of blocks; as may be run as part of reorg

src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
benaadams: Is also run on processing just in case from reorgs

PR #8806, opened 2025-06-19
Further isolate overridable env

asdacap:
- Futher isolate overridable world state behind `IOverridableEnv<T>` which is created by `IOverridableEnvFactory`.
- Instead of having a scope with desired component, on build only `IDisposable` is returned and the component is an out param.
  - eg: `IOverridableTxProccessingScope Build(Hash256 root)` changed to `IDisposable Build(Hash256 root, out T component)` where `T` is the needed component.
- This reduce the risk of skipping the dispose, and the generic make it easy to adapt the env with the use case. 
- Fix one missed dispose in blockchain bridge.
- Removed the use of block processing components outside of the given out parameter.

## What is an `IOverridableEnv<T>`?

- An env have a `Build` (and variant) method that expose an `out T` and return a dispose.
- The `T` is a protected service that must be used between `Build` and dispose of the returned disposable. 
- Use of `T` outside of this may result in memory leak or crash.
- `IOverridableEnv` (no T, give me a better name) is created by `IOverridableEnvFactory`. It is an autofac module which mean it can be passed to a lifetime configurer to configure the lifetime with the overridden components.
  - The overridden components are:
    - An `IWorldState` with a special backing KV.
    - An `IStateReader` which use that special backing KV.
    - An `ICodeInfoRepository` that is actually an `OverridableCodeInfoRepository` with a special backing KV. 
    - But I guess it can be something else too one day. Who knows... hint: you don't want to know.
- `IOverridableEnv<T>` has the `Build` variant with an `out T` parameter. It is created automatically from `IOverridableEnv` (no T) and an injected `T`, so `T` is expected to be available in the lifetime. Do not use `T` directly outside of `IOverridableEnv<T>`.

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Testing...


LukaszRozmej (commented):
I'm not sure about this to be honest. All being one disposable component actually tied together the scope and the created components, which made accidental usage of the components after disposing the scope unlikely and easier to spot. Now if the scope is separated from the components, they can easier leak. That part is more of a downgrade to me.

src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
```diff
@@ -56,7 +56,7 @@ public BackgroundTaskScheduler(IBlockProcessor blockProcessor, int concurrency,
         _blockProcessor.BlocksProcessing += BlockProcessorOnBlocksProcessing;
         _blockProcessor.BlockProcessed += BlockProcessorOnBlockProcessed;
 
-        _tasksExecutors = Enumerable.Range(0, concurrency).Select(_ => Task.Factory.StartNew(StartChannel)).ToArray();
+        _tasksExecutors = Enumerable.Range(0, concurrency).Select(_ => Task.Factory.StartNew(StartChannel, TaskCreationOptions.LongRunning)).ToArray();
```
LukaszRozmej: I remember discussion where we decided not to use `LongRunning`

src/Nethermind/Nethermind.Evm/OverridableEnv/IOverridableEnv.cs
```diff
@@ -0,0 +1,37 @@
...
+/// <see cref="T"/> here to that components or a wrapper if many component, then call one of the method here to get it.
+/// Always dispose the scope when finished.
+/// </summary>
+/// <typeparam name="T"></typeparam>
+public interface IOverridableEnv<T>
+{
+    IDisposable Build(Hash256 stateRoot, out T component);
+    IDisposable BuildAndOverride(BlockHeader header, out T component);
+    IDisposable BuildAndOverride(BlockHeader header, Dictionary<Address, AccountOverride>? stateOverride, out T component);
+}
```
LukaszRozmej:
```suggestion
public interface IOverridableEnv<T> where T : IDisposable
{
    IDisposable Build(Hash256 stateRoot, out T component);
    IDisposable BuildAndOverride(BlockHeader header, out T component);
    T BuildAndOverride(BlockHeader header, Dictionary<Address, AccountOverride>? stateOverride);}
```

LukaszRozmej (approved):

src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
asdacap: Yea, I'm thinking that this may be the cause of random failure in test. could be wrong though.

src/Nethermind/Nethermind.Evm/OverridableEnv/IOverridableEnv.cs
asdacap: Can T be refstruct? That way it must stay in  call stack. But then it make it hard to use async.

src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
benaadams: Is pointless, creates and adds a thread that immediately gets released on first `await`; so LongRunning just adds some additional overhead (thread creation) for async methods

benaadams: i.e. LongRunning only works where the method is full sync with no awaits

src/Nethermind/Nethermind.Evm/OverridableEnv/IOverridableEnv.cs
LukaszRozmej: I think you can allow for ref struct but you can't constraint to it `where T : allows ref struct`
https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/generics/constraints-on-type-parameters

asdacap: Combined to a generic `Scope`.

src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
asdacap: Removed

PR #8809, opened 2025-06-19
set V minus offset

### LINKED ISSUE #8685: JSON-RPC returns incorrect yParity for type 0x4 transaction
**Description**
The `eth_getBlockByHash` and `eth_transactionByHash` rpcs are returning an incorrect yParity in the `authorizationList` section of a type `0x4` transaction.

When used with OP Stack's op-node without `--l1.trustrpc`, op-node will validate that the data for the block retrieved from the RPC validates against its block hash and Nethermind fails that check due to the invalid yParity value.

**Steps to Reproduce**

Compare:
```
cast tx --rpc-url <nethermind-node> 0x4de037f490091665b7d81e63f2b3ef32a4d7627388d257c2295d404730a6eb0b --json | jq '.authorizationList[].yParity'
```

with the results from public node:
```
cast tx --rpc-url https://ethereum-sepolia-rpc.publicnode.com 0x4de037f490091665b7d81e63f2b3ef32a4d7627388d257c2295d404730a6eb0b --json | jq '.authorizationList[].yParity'
```

Nethermind gives `0x1` but public node gives `0xe5`.

**Actual behavior**
Nethermind gives an incorrect yParity value.

**Expected behavior**
Get the correct yParity value.

 - Operating System: Linux (docker)
 - Version: 1.31.11
 - Installation Method: Docker
 - Consensus Client: Teku 25.4.1

**Logs**
Nethermind logs look completely normal.

ak88:
Fixes #8685

y parity can be set to nonsense in a setcode tx. Instead of returning a calculated y parity in RPC set  V - V.offset instead.

### DIFF
--- a/src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/AuthorizationListForRpc.cs
+++ b/src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/AuthorizationListForRpc.cs
@@ -55,7 +55,7 @@ public class AuthorizationListForRpc
             new Tuple(tuple.ChainId,
                     tuple.Nonce,
                     tuple.CodeAddress,
-                    tuple.AuthoritySignature.RecoveryId,
+                    tuple.AuthoritySignature.V - Signature.VOffset,
                     new UInt256(tuple.AuthoritySignature.S.Span, true),
                     new UInt256(tuple.AuthoritySignature.R.Span, true))));
 


LukaszRozmej (approved):

src/Nethermind/Nethermind.JsonRpc.Test/Modules/Eth/EthRpcModuleTests.cs
```diff
@@ -1356,6 +1356,42 @@ public async Task eth_sendRawTransaction_returns_correct_error_if_AuthorityTuple
...
+    public async Task eth_getTransactionByHash_returns_correct_values_on_SetCode_tx()
+    {
+        TestSpecProvider specProvider = new TestSpecProvider(Prague.Instance);
+        specProvider.AllowTestChainOverride = false;
+
+        TestRpcBlockchain test = await TestRpcBlockchain.ForTest(SealEngineType.NethDev).Build(specProvider);
+
+        const int BadYparity = 229;
+        const int BadR = 123;
+        const int BadS = 123;
```
emlautarom1, line 1369: Just a nitpick, but why are these called `BadXXX`?

ak88: Well because they are bad values for a signature but they are still legal in an authorization.

PR #8814, opened 2025-06-20
Cleanup/Deduplicate aura block processor.

asdacap:
- Remove `ReadOnlyChainProcessingEnv`, only used by aura rpc factory now.
- Consolidate aura block processor factory between rpc and `InitializeAuraBlockchainAuRa`. 
- Remove `InitializeBlockchainAuraMerge` as the only different between merge and pre merge is the withdrawal processor which is now set in `AuRaBlockProcessorFactory`.
- Fix missed `MergeSealer` decorator in `AuraMerge`.
- Probably fix `AuraMerge` rpc's uses block processor without correct withdrawal processor.

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

#### Notes on testing

- [x] Gnosis can sync
- [x] Gnosis simulate block production does not throw.


src/Nethermind/Nethermind.Consensus.AuRa/InitializationSteps/RegisterAuRaRpcModules.cs
```diff
@@ -37,134 +32,92 @@ public class AuRaTraceModuleFactory(IWorldStateManager worldStateManager, Func<I
...
 {
-    private readonly ISpecProvider _specProvider = specProvider;
-    private readonly ILogManager _logManager = logManager;
-
-    protected override IBlockProcessor CreateBlockProcessor(IReadOnlyTxProcessingScope scope, IBlockTree blockTree,
-        IBlockValidator blockValidator, IRewardCalculator rewardCalculator, IReceiptStorage receiptStorage,
-        ISpecProvider specProvider, ILogManager logManager, IBlockProcessor.IBlockTransactionsExecutor transactionsExecutor)
+    public IBlockProcessor CreateBlockProcessor()
     {
         ITxFilter auRaTxFilter = new ServiceTxFilter(specProvider);
```
asdacap: In `InitializeBlockchainAuRa`, this is
```
ITxFilter auRaTxFilter = _api.TxAuRaFilterBuilders.CreateAuRaTxFilter(
            new ServiceTxFilter(_api.SpecProvider));
```
Is this expected?

LukaszRozmej: I think it doesn't matter much for RPC

asdacap:
I'll move it to block processor factory then

PR #8820, opened 2025-06-21
Don't broadcast txs while processing a block

benaadams:
## Changes

-  Don't broadcast txs while producing a block as it allocates a lot and we want to minimize GCs at this time also head about to change which could change the filter results; reducing variance during long blocks

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.TxPool/TxBroadcaster.cs
+++ b/src/Nethermind/Nethermind.TxPool/TxBroadcaster.cs
@@ -80,7 +80,7 @@ namespace Nethermind.TxPool
             _headInfo = chainHeadInfoProvider;
             _txGossipPolicy = transactionsGossipPolicy ?? ShouldGossip.Instance;
             // Allocate closure once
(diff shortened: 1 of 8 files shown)


src/Nethermind/Nethermind.TxPool/TxBroadcaster.cs
```diff
@@ -305,7 +305,13 @@ void NotifyPeers()
                 _txsToSend.Reset();
             }
 
-            NotifyPeers();
+            // Don't broadcast txs while producing a block as it
+            // allocates a lot and we want to minimize GCs at this time
+            // also head about to change which could change the filter results
+            if (!_headInfo.IsProcessingBlock)
```
asdacap, line 311: You can put it behind 'IBackgroundTaskScheduler` which prevent having to add `IsProcessingBlock`.

benaadams: Not sure I understand how that would work

PR #8823, opened 2025-06-21
fix: implement latest block validation in FilterManager

crStiv:
Resolves TODO comment in FilterManager.cs

## Changes

- Added IBlockFinder dependency to FilterManager constructor
- Implemented latest block validation in CreateLog method for BlockParameterType.Latest filters
- Updated dependency injection in InitializeBlockchain and test modules
- Added tests for latest block validation scenarios

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

Added comprehensive tests for latest block validation including edge cases when latest block is null or receipt is from different block.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks

This change resolves the TODO comment `//TODO: check if is last mined block` by properly validating that log filters with `BlockParameterType.Latest` only include logs from the actual latest mined block, preventing potential inconsistencies in filter results.

### DIFF
--- a/src/Nethermind/Nethermind.Facade/Filters/FilterManager.cs
+++ b/src/Nethermind/Nethermind.Facade/Filters/FilterManager.cs
@@ -28,6 +28,7 @@ namespace Nethermind.Blockchain.Filters
 
         private Hash256? _lastBlockHash;
         private readonly IFilterStore _filterStore;
+        private readonly IBlockFinder _blockFinder;
         private readonly ILogger _logger;
         private long _logIndex;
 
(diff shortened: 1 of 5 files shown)


src/Nethermind/Nethermind.Facade/Filters/FilterManager.cs
```diff
@@ -251,7 +254,12 @@ private void StoreLogs(LogFilter filter, TxReceipt txReceipt, long logIndex)
 
             if (logFilter.FromBlock.Type == BlockParameterType.Latest || logFilter.ToBlock.Type == BlockParameterType.Latest)
             {
-                //TODO: check if is last mined block
+                // Check if the receipt is from the last mined block
+                Block? latestBlock = _blockFinder.FindLatestBlock();
+                if (latestBlock is null || txReceipt.BlockNumber != latestBlock.Number)
+                {
+                    return null;
+                }
```
LukaszRozmej, line 262: This is an interesting question if this is correct in engine API driven consensus and how to handle it. Now moving block tree head is decoupled from new payloads and can come only when multiple new payloads pass. Would have to develop some custom pipelining to handle correctly as user would expect each log to be sent. It could  be done with `BlockTree.BlockAddedToMain` event and not current `BlockProcessor.TransactionProcessed`, but probably better is not to change it if no-one complains.

LukaszRozmej (commented):
Unfortunetly it is probably obsolete and not a good change in POS engine API driven world (check other comment)

PR #8841, opened 2025-06-23
feat: add dockerfile for kute tool

cbermudez97:
## Changes

- Add Dockerfile for building Kute tool Docker image

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [X] No

- [ ] Yes
- [X] No

- [ ] Yes
- [X] No

### DIFF
--- a/tools/Nethermind.Tools.Kute/Config.cs
+++ b/tools/Nethermind.Tools.Kute/Config.cs
@@ -101,4 +101,22 @@ public static class Config
         Description = "A comma separated list of additional tags to be added to the metrics in the format of key=value",
         HelpName = "value",
     };
+
+    public static Option<string> PrometheusEndpoint { get; } = new("--prometheus")
+    {
+        Description = "Endpoint to push metrics to Prometheus",
+        HelpName = "value",
+    };
+
+    public static Option<string> PrometheusBasicAuthUsername { get; } = new("--prometheus-auth-username")
+    {
+        Description = "Username for basic authentication to Prometheus",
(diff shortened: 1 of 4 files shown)


tools/Nethermind.Tools.Kute/Program.cs
```diff
@@ -134,12 +152,39 @@ private static IServiceProvider BuildServiceProvider(ParseResult parseResult)
...
+                );
+            }
+        );
+        collection.AddSingleton(
+            new MetricsPrometheusTextOutputFormatter(
+                new MetricsPrometheusOptions()
+                {
+                    MetricNameFormatter = (_, metricName) => PrometheusFormatterConstants.MetricNameFormatter("kute", metricName),
+                }
+            )
```
emlautarom1, line 180: I don't see how we're configuring this class based on other classes or arguments. If such configuration is not needed we can inline it where it's used

tools/Nethermind.Tools.Kute/Config.cs
```diff
@@ -77,4 +77,46 @@ public static class Config
     {
         Description = "If true then each batched request will be unwraped to single requests"
     };
+
+    public static Option<IEnumerable<(string, string)>> Tags { get; } = new("--tags", "-t")
```
emlautarom1, line 81: We should update the README with examples on how to use `--tags` and prometheus options.

PR #8844, opened 2025-06-24
Refactor/Move the rest of built in rpc modules to DI

asdacap:
- Move the rest of the built in Rpc modules to DI.
- Also move some of their dependencies that is simple.
- Fix randomly failing hive test due to `IContextAwareRpcModule` being shared.

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- [x] Mainnet can sync
- [x] RPC hive test passed


src/Nethermind/Nethermind.Init/Modules/NethermindModule.cs
```diff
@@ -62,10 +63,8 @@ protected override void Load(ContainerBuilder builder)
             .AddSingleton<IEthereumEcdsa, ISpecProvider>((specProvider) => new EthereumEcdsa(specProvider.ChainId))
             .Bind<IEcdsa, IEthereumEcdsa>()
 
-            .AddSingleton<SimulateReadOnlyBlocksProcessingEnvFactory>()
-            .AddSingleton<IBlockchainBridgeFactory, BlockchainBridgeFactory>()
-            .AddScoped<IBlockchainBridge>((ctx) => ctx.Resolve<IBlockchainBridgeFactory>().CreateBlockchainBridge())
-
+            .AddSingleton<IChainHeadSpecProvider, ChainHeadSpecProvider>()
+            .AddSingleton<IChainHeadInfoProvider, ChainHeadInfoProvider>()
```
LukaszRozmej, line 67: Will they be same instance?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Init/Modules/NethermindModule.cs
asdacap: Yes. Singleton.

LukaszRozmej: not 2 signletons?

asdacap: Well, yea, one is spec provider, one is info provider so they are different thing.

PR #8859, opened 2025-06-25
Reduce use of InitializeNetworkProtocol and review step dependencies

asdacap:
- `INethermindPlugin.InitializeNetworkProtocol` is often abused as the part of the code that execute after most of the dependencies are executed.
  - This is not necessary anymore as plugins can declare their own `IStep` and specify at which point should it execute.
- Convert plugins from using `InitializeNetworkProtocol` to using custom `IStep` with `InitializeNetwork` as dependencies or something else.
- `AnalyticsPlugin` to `AnalyticsStep`.
  - Require `ITxPool` which is set in `InitializeBlockchain` so it is set there. 
  - Removed `StartLogProducer` as it automatically created.
  - Moved `StartGrpc` to DI as one of its component is  an `IPublisher` which is used by `AnalyticsStep`.
  - Grpc components are now not configured if disabled.
- `EthStatsPlugin` to `EthStatsSteps`.
  - Same as `AnalyticsPlugin`, just need the `InitializeBlockchain`.
- `HivePlugin` to `HiveStep`.
  - Uses `CompositeTxGossipPolicy` which is modified in `InitializeNetwork`.
- `UPnPPlugin` to `UPnPStep`.
  - New step. Does not need anything.
- `EraStep` moved to after `ReviewBlockTree`. This is probably a mistake, it should not run in parallel of blocktree review something bad could happen.
- `StartMonitoring` moved ahead to after `InitializeBlockTree`. So the only reason it can't start earlier is because `IEthSyncingInfo` from syncing codes require database.
- Merge plugin does weird stuff in its InitializeNetworkProtocol, so can't move `InitializeBlockProducer` ahead.

- Add a debug log to show configured step. Looks like this:
```
● LogHardwareInfo                                                                           
● StartGrpc                             
● UPnPStep                                                                                                                                                                               
○ InitializePrecompiles                                                                     
○ MigrateConfigs                                                                            
○ ResolveIps                                                                                
  ○ ApplyMemoryHint (depends on MigrateConfigs)
  ○ SetupKeyStore (depends on ResolveIps)                                                   
    ○ InitTxTypesAndRlp (depends on ApplyMemoryHint)
    ○ InitDatabase (depends on ApplyMemoryHint)                
      ● StartMonitoring (depends on InitDatabase)
      ○ InitializeBlockTree (depends on InitTxTypesAndRlp, InitDatabase, SetupKeyStore)
        ○ InitializePlugins (depends on InitializeBlockTree)
          ○ InitializeBlockchain (depends on InitializePlugins, InitializePrecompiles)
            ● AnalyticsSteps (depends on InitializeBlockchain)
            ● EthStatsStep (depends on InitializeBlockchain)
            ○ StartBlockProcessor (depends on InitializeBlockchain)
              ○ LoadGenesisBlock (depends on StartBlockProcessor)                                                                                                                        
                ○ ReviewBlockTree (depends on LoadGenesisBlock)
                  ○ EraStep (depends on ReviewBlockTree)
                    ○ InitializeNetwork (depends on EraStep)
                      ● DatabaseMigrations (depends on InitializeNetwork)
                      ● HiveStep (depends on InitializeNetwork)                   
                      ○ InitializeBlockProducer (depends on InitializeNetwork)
                        ● StartBlockProducer (depends on InitializeBlockProducer)  
                        ○ RegisterRpcModules (depends on InitializeBlockProducer)
                          ○ RegisterPluginRpcModules (depends on RegisterRpcModules)
                            ● StartRpc (depends on RegisterPluginRpcModules)     
```

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

#### Notes on testing

- [X] Mainnet can start with hive, stats, grpc,

### DIFF
--- a/src/Nethermind/Nethermind.Init/Steps/EthereumStepsManager.cs
+++ b/src/Nethermind/Nethermind.Init/Steps/EthereumStepsManager.cs
@@ -90,7 +90,7 @@ namespace Nethermind.Init.Steps
                 }
             }
 
-            if (_logger.IsDebug) _logger.Debug($"Ethereum steps dependency tree:\n{BuildStepDependencyTree(stepInfoMap}");
+            if (_logger.IsDebug) _logger.Debug($"Ethereum steps dependency tree:\n{BuildStepDependencyTree(stepInfoMap)}");
             List<Task> allRequiredSteps = new();
             foreach (StepWrapper stepWrapper in stepInfoMap.Values)
             {


LukaszRozmej:
hm... init methods are easier to understand than steps, maybe we need `InitBlockchain` method?

asdacap:
Hmm, plugins are no dependency injected itself. We can have `IPluginHooks` though, or `IAfterInitializeBlockchain`. Only analytics and eth stats run right after initialize blockchain though.

src/Nethermind/Nethermind.EthStats/EthStatsPlugin.cs
```diff
@@ -1,118 +1,27 @@
...
-    private readonly bool _isOn = ethStatsConfig.Enabled;
-
-    public bool Enabled => _isOn;
-
+    public bool Enabled => ethStatsConfig.Enabled;
     public string Name => "EthStats";
     public string Description => "Ethereum Statistics";
     public string Author => "Nethermind";
+    public ValueTask DisposeAsync() => ValueTask.CompletedTask;
+    public Task Init(INethermindApi nethermindApi) => Task.CompletedTask;
```
LukaszRozmej: Maybe those should be default implementations in the interface?

src/Nethermind/Nethermind.Init/Steps/EthereumStepsManager.cs
```diff
@@ -89,6 +90,12 @@ private List<Task> CreateAndExecuteSteps(CancellationToken cancellationToken)
                 }
             }
 
+            if (_logger.IsDebug)
+            {
+                string dot = BuildStepDependencyTree(stepInfoMap);
+                _logger.Debug("Ethereum steps dependency tree:\n" + dot);
+            }
+
```
LukaszRozmej:
```suggestion
            if (_logger.IsDebug) _logger.Debug($"Ethereum steps dependency tree:\n{BuildStepDependencyTree(stepInfoMap}");
```

src/Nethermind/Nethermind.Init/Steps/EthereumStepsManager.cs
```diff
@@ -162,6 +169,93 @@ private void ReviewFailedAndThrow(Task task)
                 ExceptionDispatchInfo.Capture(task.Exception.GetBaseException()).Throw();
         }
 
+        /// <summary>
+        /// Recursively prints roots (steps with no dependencies) and their dependents.
+        /// </summary>
+        private string BuildStepDependencyTree(Dictionary<Type, StepWrapper> stepInfoMap)
```
LukaszRozmej, line 170: Can't we reuse existing code for building dependency trie?

src/Nethermind/Nethermind.UPnP.Plugin/UPnPStep.cs
```diff
@@ -0,0 +1,93 @@
...
+{
+    private static readonly TimeSpan ExpirationRate = TimeSpan.FromMinutes(10);
+    private readonly ILogger _logger = logManager.GetClassLogger<UPnPStep>();
+
+    // Routers tend to clean mapping, so we need to periodically
+    private readonly PeriodicTimer _timer = new(ExpirationRate);
+
+    public Task Execute(CancellationToken cancellationToken)
+    {
+        Task.Factory.StartNew(RunRefreshLoop, TaskCreationOptions.LongRunning);
```
LukaszRozmej:
```suggestion
        Task.Factory.StartNew(RunRefreshLoop);
```

LukaszRozmej (approved):

src/Nethermind/Nethermind.EthStats/EthStatsPlugin.cs
asdacap: Removed init. Dispose don't have default though.

src/Nethermind/Nethermind.Init/Steps/EthereumStepsManager.cs
asdacap: Does not look so actually.

PR #8862, opened 2025-06-25
Fix a few Hive tests of `eth_simulatev1`

rubo:
## Changes

- Fixed some of the failing Hive tests
- Revised the workflow

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No


src/Nethermind/Nethermind.Facade/Eth/BlockForRpc.cs
```diff
@@ -77,7 +74,6 @@ public BlockForRpc(Block block, bool includeFullTransactionData, ISpecProvider s
         Size = _blockDecoder.GetLength(block, RlpBehaviors.None);
         StateRoot = block.StateRoot;
         Timestamp = block.Timestamp;
-        TotalDifficulty = block.TotalDifficulty ?? 0;
```
LukaszRozmej, line 79: this is still usefull for pre-merge networks.

Maybe `TotalDifficulty = block.Difficulty.IsZero ? null : block.TotalDifficulty ?? 0;` and make `TotalDifficulty` nullable and ignore nulls?

src/Nethermind/Nethermind.Facade/Simulate/SimulateBridgeHelper.cs
```diff
@@ -211,13 +211,13 @@ private static bool TryGetBlock(
...
-            if (args.Action is BlockProcessor.TxAction.Stop or BlockProcessor.TxAction.Skip && payload.Validation)
-            {
-                error = $"invalid transaction index: {index} at block number: {callHeader.Number}, Reason: {args.Reason}";
-                return false;
-            }
+            //if (args.Action is BlockProcessor.TxAction.Stop or BlockProcessor.TxAction.Skip && payload.Validation)
+            //{
+            //    error = $"invalid transaction index: {index} at block number: {callHeader.Number}, Reason: {args.Reason}";
+            //    return false;
+            //}
```
LukaszRozmej, line 221: This is not needed? If not - delete.
But what about issues like nonce gaps?

src/Nethermind/Nethermind.Facade/Eth/BlockForRpc.cs
rubo: Since they mentioned that `totalDifficulty` has been removed from other JSON-RPC methods, I just removed it completely.

LukaszRozmej: It was on main Ethereum. But let's keep it on pre-merge chains.

rubo: Done

src/Nethermind/Nethermind.Facade/Simulate/SimulateBridgeHelper.cs
rubo: I don't know about nonce gaps, but this code surely was messing up with Hive tests' expectations. I want to keep the commented-out code until we finish this completely.

src/Nethermind/Nethermind.Facade/Eth/BlockForRpc.cs
```diff
@@ -73,10 +73,13 @@ public BlockForRpc(Block block, bool includeFullTransactionData, ISpecProvider s
         Size = _blockDecoder.GetLength(block, RlpBehaviors.None);
         StateRoot = block.StateRoot;
         Timestamp = block.Timestamp;
-        Transactions = (includeFullTransactionData
-                ? block.Transactions.Select((t, idx) => TransactionForRpc.FromTransaction(t, block.Hash, block.Number, idx, block.BaseFeePerGas, specProvider.ChainId))
-                : block.Transactions.Select(t => t.Hash).OfType<object>())
-            .ToArray();
+        TotalDifficulty = _isAuRaBlock ? block.TotalDifficulty ?? 0 : null;
```
LukaszRozmej: it is also for clique (like JOC).
But difficulty can't be 0 in normal case, only post-merge, so:
```suggestion
        TotalDifficulty = block.Difficulty.IsZero ? null : block.TotalDifficulty ?? UInt256.Zero;
```

PR #8866, opened 2025-06-26
Remove State from Evm dependencies

deffrian:
Fixes Closes Resolves #
This pr is part of EVM zkfication 

## Changes

- Move IWorldState and related interfaces to EVM package.
- Removes Accept function from IWorldState. Adds IVisitingWorldStateInterface with this function only.
- State and Storage tracer interfaces are part of EVM now

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


LukaszRozmej:
Can you provide justification about those decisions?
Especially why some things needed to be made public, why can't warmup stay in EVM?

deffrian:
@LukaszRozmej 
Moved Warmup back. Problem is that we can't access WorldState, CodeInfoRepository and Db related classes from EVM anymore. Came up with workaround, probably this one is better

src/Nethermind/Nethermind.Evm/State/Snapshot.cs
```diff
@@ -0,0 +1,34 @@
+// SPDX-FileCopyrightText: 2022 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Nethermind.Evm.Test")]
```
LukaszRozmej: why needed on this type?

src/Nethermind/Nethermind.Init/Modules/WorldStateModule.cs
```diff
@@ -66,6 +67,7 @@ protected override void Load(ContainerBuilder builder)
             .RegisterSingletonJsonRpcModule<IPruningTrieStateAdminRpcModule>()
 
             .AddSingleton<IReadOnlyStateProvider, ChainHeadReadOnlyStateProvider>()
+            .AddSingleton<IVisitingWorldState>(ctx => ctx.Resolve<IWorldStateManager>().GlobalWorldState)
```
LukaszRozmej: This might be bad? @asdacap

src/Nethermind/Nethermind.Init/Steps/InitializeBlockchain.cs
```diff
@@ -158,7 +159,15 @@ protected virtual Task InitBlockchain()
...
         }
 
-        protected virtual ITransactionProcessor CreateTransactionProcessor(CodeInfoRepository codeInfoRepository, IVirtualMachine virtualMachine, IWorldState worldState)
+        private void WarmupEvm()
+        {
+            IKeyValueStoreWithBatching db = new MemDb();
+            TrieStore trieStore = new(new NodeStorage(db), No.Pruning, Persist.EveryBlock, new PruningConfig(), NullLogManager.Instance);
+            WorldState state = new(trieStore, db, NullLogManager.Instance);
+            VirtualMachine.WarmUpEvmInstructions(state, new CodeInfoRepository());
+        }
```
LukaszRozmej: Nice!

src/Nethermind/Nethermind.Evm/State/PreBlockCaches.cs
```diff
@@ -0,0 +1,17 @@
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System;
+using Nethermind.Core;
+using Nethermind.Core.Extensions;
+
+namespace Nethermind.Evm.State;
+
+public readonly struct PrecompileCacheKey(Address address, ReadOnlyMemory<byte> data) : IEquatable<PrecompileCacheKey>
```
LukaszRozmej: Shouldn't it be at `Nethermind.State`?

src/Nethermind/Nethermind.State/CodeInfoRepository.cs
```diff
@@ -12,14 +12,15 @@
...
 using Nethermind.Evm.CodeAnalysis;
 using Nethermind.Evm.EvmObjectFormat;
 using Nethermind.Evm.Precompiles;
 using Nethermind.Evm.Precompiles.Bls;
 using Nethermind.Evm.Precompiles.Snarks;
-using Nethermind.State;
+using Nethermind.Evm.State;
 
-namespace Nethermind.Evm;
+namespace Nethermind.State;
```
LukaszRozmej, line 23: Why `State`? Doesn't sound right?

src/Nethermind/Nethermind.State/CodeInfoRepository.cs
```diff
@@ -198,7 +199,7 @@ private static bool TryGetDelegatedAddress(ReadOnlySpan<byte> code, [NotNullWhen
 
     private static PrecompileInfo CreateCachedPrecompile(
         in KeyValuePair<AddressAsKey, PrecompileInfo> originalPrecompile,
-        ConcurrentDictionary<PreBlockCaches.PrecompileCacheKey, (byte[], bool)> cache) =>
+        ConcurrentDictionary<PrecompileCacheKey, (byte[], bool)> cache) =>
```
LukaszRozmej: `PrecompileCacheKey` should be at same project as this

src/Nethermind/Nethermind.State/IVisitingWorldState.cs
```diff
@@ -0,0 +1,30 @@
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using Nethermind.Core.Crypto;
+using Nethermind.Evm.State;
+using Nethermind.Trie;
+
+namespace Nethermind.State;
+
+public interface IVisitingWorldState : IWorldState
```
LukaszRozmej, line 10: Do we want this to be under same interface hierarchy or use interface segregation?

LukaszRozmej (commented):
OH MY GOD! The namespace changes in this PR are abysmal :P

I think that in some places where you use `IVisitingWorldState` just simple `IWorldState` might suffice?

src/Nethermind/Nethermind.Init/Modules/WorldStateModule.cs
deffrian:
```
Autofac.Core.DependencyResolutionException : None of the constructors found with 'Nethermind.Core.Container.NethermindConstructorFinder' on type 'Nethermind.Core.Test.Modules.MainBlockProcessingContext' can be invoked with the available services and parameters:
Cannot resolve parameter 'Nethermind.State.IVisitingWorldState WorldState'
```
Without this line some tests fail with this exception. I tried to understand how it works with just IWorldState. No success. Maybe you know or @asdacap

src/Nethermind/Nethermind.State/IVisitingWorldState.cs
deffrian: I tried to make separate interface for that initially. But in some places we need both IWorldState and Accept function. I thought splitting them will add too much complexity with questionable upsides

src/Nethermind/Nethermind.Core.Test/Modules/MainBlockProcessingContext.cs
```diff
@@ -23,7 +23,7 @@ public record MainBlockProcessingContext(
     ILifetimeScope LifetimeScope,
     IBlockProcessingQueue BlockProcessingQueue,
     IBlockchainProcessor BlockchainProcessor,
-    IWorldState WorldState,
+    IVisitingWorldState WorldState,
```
asdacap, line 26: Do this have to be an 'IVisitingWorldState`?

deffrian: Yes, we need to call `DumpState`

src/Nethermind/Nethermind.Init/Steps/InitializeBlockchain.cs
```diff
@@ -158,7 +159,15 @@ protected virtual Task InitBlockchain()
             return Task.CompletedTask;
         }
 
-        protected virtual ITransactionProcessor CreateTransactionProcessor(CodeInfoRepository codeInfoRepository, IVirtualMachine virtualMachine, IWorldState worldState)
+        private void WarmupEvm()
+        {
+            IKeyValueStoreWithBatching db = new MemDb();
+            TrieStore trieStore = new(new NodeStorage(db), No.Pruning, Persist.EveryBlock, new PruningConfig(), NullLogManager.Instance);
+            WorldState state = new(trieStore, db, NullLogManager.Instance);
```
asdacap: I recommend `IWorldStateManager.CreateResettableWorldState` or something, set the root to `Keccak.EmptyTreeHash`. Way to create world state may change depending on type of statedb.

src/Nethermind/Nethermind.Api/IMainProcessingContext.cs
```diff
@@ -21,7 +21,7 @@ public interface IMainProcessingContext : IStoppableService
     ITransactionProcessor TransactionProcessor { get; }
     IBlockProcessor BlockProcessor { get; }
     IBlockchainProcessor BlockchainProcessor { get; }
-    IWorldState WorldState { get; }
+    IVisitingWorldState WorldState { get; }
```
asdacap, line 24: Do this need to be `IVisitingWorldState`?

src/Nethermind/Nethermind.Init/Modules/WorldStateModule.cs
asdacap: This is bad.

src/Nethermind/Nethermind.Api/IMainProcessingContext.cs
deffrian: Same reason. We need to call DumpState

src/Nethermind/Nethermind.State/IWorldStateManager.cs
```diff
@@ -12,7 +13,7 @@ namespace Nethermind.State;
 
 public interface IWorldStateManager
 {
-    IWorldState GlobalWorldState { get; }
+    IVisitingWorldState GlobalWorldState { get; }
```
asdacap, line 16: Only the main world state is an `IVisitingWorldState`?

src/Nethermind/Nethermind.State/IVisitingWorldState.cs
asdacap: This seems like something that can be moved to `IStateReader`.

src/Nethermind/Nethermind.Init/Modules/WorldStateModule.cs
asdacap: It is passed in here `https://github.com/NethermindEth/nethermind/pull/8866/files#diff-dd529e2fd71ffdf68c00eef5df37831e82170157dd2b006a91155f9ca37ddde2R93` but I recommend moving it to `IStateReader` instead.

src/Nethermind/Nethermind.State/IWorldStateManager.cs
deffrian: Wdym?

src/Nethermind/Nethermind.Init/Modules/WorldStateModule.cs
asdacap: Not sure if link work. Page hangs. Its in `TestBlockProcessingModule`.

src/Nethermind/Nethermind.State/IWorldStateManager.cs
asdacap: Like, there are many `IWorldState`, but only one implement `IVisitingWorldState`? And like only one need to?

deffrian: Only `WorldStateMetricsDecorator` is not `IVisitingWorldState`. It compiles so we do't need to make it `IVisitingWorldState`. I can add Accept function back no problems

deffrian: And we will have to change state to `IVisitingWorldState` in many other places to do that. So i prefer to not do it tbh

src/Nethermind/Nethermind.JsonRpc/Modules/Proof/ProofModuleFactory.cs
```diff
@@ -10,27 +9,30 @@
...
-        IReadOnlyTxProcessingEnvFactory readOnlyTxProcessingEnvFactory
+        IOverridableEnvFactory overridableEnvFactory
     ) : ModuleFactoryBase<IProofRpcModule>
     {
 
         public override IProofRpcModule Create()
         {
             // Note: No overridable world scope here. So there aren't any risk of leaking KV store.
-            IReadOnlyTxProcessingScope txProcessingEnv = readOnlyTxProcessingEnvFactory.Create().Build(Keccak.EmptyTreeHash);
+            var env = overridableEnvFactory.Create();
```
tanishqjasoria: it would be good use the explicit type here

src/Nethermind/Nethermind.JsonRpc/Modules/Proof/ProofModuleFactory.cs
```diff
@@ -10,27 +9,30 @@
...
         ILifetimeScope rootLifetimeScope,
-        IReadOnlyTxProcessingEnvFactory readOnlyTxProcessingEnvFactory
+        IOverridableEnvFactory overridableEnvFactory
     ) : ModuleFactoryBase<IProofRpcModule>
     {
 
         public override IProofRpcModule Create()
         {
             // Note: No overridable world scope here. So there aren't any risk of leaking KV store.
-            IReadOnlyTxProcessingScope txProcessingEnv = readOnlyTxProcessingEnvFactory.Create().Build(Keccak.EmptyTreeHash);
```
asdacap: The `IOverridableEnv` need special handling. See the comment. Double check in `ITracer` that it is wrapping any component with scope.

asdacap:
Consider having a `IZkSafeWorldState` instead of `IVisitingWorldState`

asdacap:
Question, is this ZK safe? https://github.com/NethermindEth/nethermind/pull/8525/files#diff-cd3f2dce1659790061b5effa2b3cd868639402ccc2e3ad3b1ecb029d13754458R21

src/Nethermind/Nethermind.State/CodeInfoRepository.cs
deffrian: It's a IWorldState extension in some sense. Not sure where else it can go

tanishqjasoria:
> Question, is this ZK safe? https://github.com/NethermindEth/nethermind/pull/8525/files#diff-cd3f2dce1659790061b5effa2b3cd868639402ccc2e3ad3b1ecb029d13754458R21

yes, i dont think it would matter. We would seperate out the evm and only feed the world state after setting the correct state root.

asdacap (approved):
Consider adding `IWorldState.DumpState()` instead. Having an `IWorldStateManager` that output `IVisitingWorldState` is weird. But I'd rather have the import changes done.

PR #8867, opened 2025-06-26
Optimise OpOrigin

benaadams:
## Changes

- Optimise OpOrigin by using 32 byte address (rather than 20 byte) when pushing to stack

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Environment.cs
+++ b/src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Environment.cs
@@ -34,7 +34,24 @@ internal static partial class EvmInstructions
     }
 
     /// <summary>
-    /// Defines an environment introspection operation that returns a byte span.
+    /// Defines an environment introspection operation that returns a big endian word.
+    /// Implementations should provide a static gas cost and a static Operation method.
+    /// </summary>
+    public interface IOpEnv32Bytes
+    {
+        /// <summary>
+        /// The gas cost for the operation.
+        /// </summary>
+        virtual static long GasCost => GasCostOf.Base;
+        /// <summary>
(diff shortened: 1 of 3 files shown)


src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Environment.cs
```diff
@@ -286,6 +303,29 @@ public static EvmExceptionType InstructionBlkUInt64<TOpEnv, TTracingInst>(Virtua
...
+    [SkipLocalsInit]
+    public static EvmExceptionType InstructionEnv32Bytes<TOpEnv, TTracingInst>(VirtualMachine vm, ref EvmStack stack, ref long gasAvailable, ref int programCounter)
+        where TOpEnv : struct, IOpEnv32Bytes
+        where TTracingInst : struct, IFlag
+    {
+        gasAvailable -= TOpEnv.GasCost;
+
+        ref readonly ValueHash256 result = ref TOpEnv.Operation(vm);
+
+        stack.Push32Bytes<TTracingInst>(in result);
```
LukaszRozmej, line 324: pushing ValueHash256 is the same as pushing a Vector256?

LukaszRozmej:
Do we have results?

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Environment.cs
benaadams: Yeah pretty much

benaadams:
> Do we have results?

Is less bouncy (more consistent) but no real gain 
image


Is probably worth committing for consistency, but not sure worth making changes to others in same way

PR #8868, opened 2025-06-26
feat: Make tx execution context overridable

hudem1:
For Arbitrum, we wanna make the TxExecutionContext overridable because we wanna add some fields there.

I'm not sure it's the best way to override it because i had to remove some `ref readonly`.
Maybe if we make it a record class (that we can then inherit from) instead of interface + struct, it'd be better ?

### DIFF
--- a/src/Nethermind/Nethermind.Evm/IVirtualMachine.cs
+++ b/src/Nethermind/Nethermind.Evm/IVirtualMachine.cs
@@ -13,8 +13,8 @@ namespace Nethermind.Evm
         TransactionSubstate ExecuteTransaction<TTracingInst>(EvmState state, IWorldState worldState, ITxTracer txTracer)
             where TTracingInst : struct, IFlag;
         ref readonly BlockExecutionContext BlockExecutionContext { get; }
-        ref readonly TxExecutionContext TxExecutionContext { get; }
+        ref readonly ITxExecutionContext TxExecutionContext { get; }
         void SetBlockExecutionContext(in BlockExecutionContext blockExecutionContext);
-        void SetTxExecutionContext(in TxExecutionContext txExecutionContext);
+        void SetTxExecutionContext(in ITxExecutionContext txExecutionContext);
     }
 }
--- a/src/Nethermind/Nethermind.Evm/TxExecutionContext.cs
+++ b/src/Nethermind/Nethermind.Evm/TxExecutionContext.cs
@@ -10,11 +10,21 @@ namespace Nethermind.Evm
         Address origin,
         ICodeInfoRepository codeInfoRepository,
         byte[][] blobVersionedHashes,
-        in UInt256 gasPrice)
+        in UInt256 gasPrice) : ITxExecutionContext
     {
-        public readonly Address Origin = origin;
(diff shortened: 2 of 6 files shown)


src/Nethermind/Nethermind.Evm/TxExecutionContext.cs
```diff
@@ -10,11 +10,21 @@ public readonly struct TxExecutionContext(
...
         ICodeInfoRepository codeInfoRepository,
         byte[][] blobVersionedHashes,
-        in UInt256 gasPrice)
+        in UInt256 gasPrice) : ITxExecutionContext
     {
-        public readonly Address Origin = origin;
-        public readonly ICodeInfoRepository CodeInfoRepository = codeInfoRepository;
-        public readonly byte[][]? BlobVersionedHashes = blobVersionedHashes;
-        public readonly UInt256 GasPrice = gasPrice;
+        private readonly UInt256 _gasPrice = gasPrice;
```
hudem1, line 15: i had to create a field so that i could keep the `in` keyword when passing `gasPrice`

src/Nethermind/Nethermind.Evm/TxExecutionContext.cs
```diff
@@ -10,11 +10,21 @@ public readonly struct TxExecutionContext(
...
+        public readonly byte[][]? BlobVersionedHashes => blobVersionedHashes;
+        public readonly UInt256 GasPrice => _gasPrice;
+    }
+
+    public interface ITxExecutionContext
+    {
+        Address Origin { get; }
+        ICodeInfoRepository CodeInfoRepository { get; }
+        byte[][]? BlobVersionedHashes { get; }
+        UInt256 GasPrice { get; }
```
benaadams, line 28:
```suggestion
        ref readonly UInt256 GasPrice { get; }
```

src/Nethermind/Nethermind.Evm/IVirtualMachine.cs
```diff
@@ -13,8 +13,8 @@ public interface IVirtualMachine
         TransactionSubstate ExecuteTransaction<TTracingInst>(EvmState state, IWorldState worldState, ITxTracer txTracer)
             where TTracingInst : struct, IFlag;
         ref readonly BlockExecutionContext BlockExecutionContext { get; }
-        ref readonly TxExecutionContext TxExecutionContext { get; }
+        ref readonly ITxExecutionContext TxExecutionContext { get; }
         void SetBlockExecutionContext(in BlockExecutionContext blockExecutionContext);
-        void SetTxExecutionContext(in TxExecutionContext txExecutionContext);
+        void SetTxExecutionContext(in ITxExecutionContext txExecutionContext);
```
benaadams, line 18: This will cause it to box an allocate as well as taking a hit in performance

Could instead go via generics; but might be complicated

Or a an additional nullable "extensions" property to TxExecutionContext for one extra Interface field and then extend via that?

LukaszRozmej (changes requested):
Either do it by generics or pass those parameters other way

benaadams:
Could live on VirtualMachine also; might be easier there

PR #8871, opened 2025-06-27
Refactor/deduplicate block producer env

asdacap:
- Convert `BlockProducerEnvFactory` to DI except aura (pre merge)
- Most configuration is already from #8741 
- Delete optimism related factory.
- Unit test modified to reflect.

#### What types of changes does your code introduce?

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

#### If yes, did you write tests?

- [X] Yes
- [ ] No

#### Notes on testing

- [x] Mainnet can run normally with simulate block production

PR #8872, opened 2025-06-27
Feature/enable prewarm for rpc and block producer

asdacap:
- Require #8871 
- Configure prewarmer for RPC and block producer except for simulate and AuRa.

- [X] Optimization

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- [ ] Mainnet with simulate block production run fine


src/Nethermind/Nethermind.Init/Modules/BlockProcessingModule.cs
```diff
@@ -74,5 +75,20 @@ protected override void Load(ContainerBuilder builder)
...
+        {
+            builder
+                .AddScoped<PreBlockCaches, IWorldState>(worldState =>
+                {
+                    if (worldState is IPreBlockCaches preBlockCaches && preBlockCaches.Caches is not null)
+                    {
+                        return preBlockCaches.Caches;
+                    }
+                    return new PreBlockCaches();
+                })
```
LukaszRozmej:
```suggestion
                .AddScoped<PreBlockCaches, IWorldState>(worldState => (worldState as IPreBlockCaches)?.Caches ?? new PreBlockCaches())
```

src/Nethermind/Nethermind.State/WorldStateManager.cs
```diff
@@ -23,18 +24,21 @@ public class WorldStateManager : IWorldStateManager
     private readonly IDbProvider _dbProvider;
     private readonly BlockingVerifyTrie? _blockingVerifyTrie;
     private readonly ILastNStateRootTracker _lastNStateRootTracker;
+    private readonly IBlocksConfig _blocksConfig;
 
     public WorldStateManager(
         IWorldState worldState,
         IPruningTrieStore trieStore,
         IDbProvider dbProvider,
+        IBlocksConfig blocksConfig,
```
LukaszRozmej, line 33: Not sure if we should make WorldState depend on config class?

LukaszRozmej (commented):
Good start but some design questions come to light:
1. For RPC, I am not sure if we want to use pre-warming. RPC providers have multiple concurrent calls and pre-warming could increase the overall load on the node - mostly in CPU though.
2. When building block we build it transaction-by-transaction, lazely picking them from TxSource, so unless we redesign building to be multithreaded and eagerly picking next transactions to pre-warm I think we don't get any benefit, but pay some cost.

For building I think better would be to have a cache that is persisted between building a block and then multiple further improvements (if we don't have one already? cc: @benaadams)

asdacap:
Fair, we could have a configuration that limits the thread for rpc or disable it.

PR #8875, opened 2025-06-27
Optimize JwtAuthentication.Authenticate

benaadams:
## Changes

- Move out cold code
- Validate initial non-async before creating state machine to shrink
- Use `ReadOnlyMemory<byte>` slice rather than `string` slice
- Add extension of `.ToUnixTimeSeconds` for `DateTime`
- Use singleton for `JsonWebTokenHandler`

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Core/Authentication/JwtAuthentication.cs
+++ b/src/Nethermind/Nethermind.Core/Authentication/JwtAuthentication.cs
@@ -104,21 +104,21 @@ public sealed partial class JwtAuthentication : IRpcAuthentication
         }
     }
 
-    public Task<bool> Authenticate(string? authToken)
+    public Task<bool> Authenticate(string? token)
     {
-        if (string.IsNullOrEmpty(authToken))
+        if (string.IsNullOrEmpty(token))
         {
             if (_logger.IsWarn) WarnTokenNotFound();
             return False;
         }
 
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Core/Extensions/DateTimeExtensions.cs
```diff
@@ -0,0 +1,11 @@
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System;
+
+namespace Nethermind.Core.Extensions;
+public static class DateTimeExtensions
+{
+    public static long ToUnixTimeSeconds(this DateTime date)
+        => (date - DateTime.UnixEpoch).Ticks / TimeSpan.TicksPerSecond;
```
LukaszRozmej, line 10: Maybe use `DateTimeOffset`?
https://learn.microsoft.com/en-us/dotnet/api/system.datetimeoffset.tounixtimeseconds?view=net-9.0

LukaszRozmej (approved):

src/Nethermind/Nethermind.Core/Extensions/DateTimeExtensions.cs
benaadams: Adds additional conversions for the timezone, but we don't have any timezone so is wasted work

PR #8899, opened 2025-07-01
Fix/unencodable header

asdacap:
- The header encoder cannot encode certain header.
  - If base fee is 0, and later field is not zero, it will skip base fee, and encode later field but then decode the later field as basefee without throwing.
- This does not happen in mainnet, but cause long debugging session in test, and make it so that test must activate the right spec all the time.
- This PR change it so that if any later field is populated, earlier field must be set (even null) which make the encoding and decoding consistent.

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- [x] Mainnet can sync

### DIFF
--- a/src/Nethermind/Nethermind.Serialization.Rlp/HeaderDecoder.cs
+++ b/src/Nethermind/Nethermind.Serialization.Rlp/HeaderDecoder.cs
@@ -268,7 +268,7 @@ namespace Nethermind.Serialization.Rlp
             requiredItems[0] = !item.BaseFeePerGas.IsZero;
             requiredItems[1] = (item.WithdrawalsRoot is not null);
             requiredItems[2] = (item.BlobGasUsed is not null);
-            requiredItems[3] = (item.ExcessBlobGas is not null || item.ExcessBlobGas is not null);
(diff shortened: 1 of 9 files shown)


LukaszRozmej:
Based on spec minimal base fee is 7, so 0 would be an invalid block.

src/Nethermind/Nethermind.Serialization.Rlp/HeaderDecoder.cs
```diff
@@ -235,31 +191,25 @@ public void Encode(RlpStream rlpStream, BlockHeader? header, RlpBehaviors rlpBeh
...
-            {
-                rlpStream.Encode(header.WithdrawalsRoot ?? Keccak.Zero);
-            }
-
-            if (header.BlobGasUsed is not null || header.ExcessBlobGas is not null || header.ParentBeaconBlockRoot is not null)
-            {
-                rlpStream.Encode(header.BlobGasUsed.GetValueOrDefault());
-                rlpStream.Encode(header.ExcessBlobGas.GetValueOrDefault());
-            }
+            Span<bool> requiredItems = stackalloc bool[6];
```
flcl42, line 194: Probably can become a reused function

PR #8904, opened 2025-07-01
Just removes the timeout in tests when debugging an rpc call

ak88:
as per title


src/Nethermind/Nethermind.JsonRpc.Test/RpcTest.cs
```diff
@@ -29,7 +31,7 @@ public static async Task<JsonRpcResponse> TestRequest<T>(T module, string method
 
     public static async Task<string> TestSerializedRequest<T>(T module, string method, params object?[]? parameters) where T : class, IRpcModule
     {
-        using AutoCancelTokenSource cts = AutoCancelTokenSource.ThatCancelAfter(10.Seconds());
+        using AutoCancelTokenSource cts = AutoCancelTokenSource.ThatCancelAfter(Debugger.IsAttached ? TimeSpan.FromMilliseconds(-1) : 10.Seconds());
```
LukaszRozmej, line 34: maybe do similar in actual RPC handling?

LukaszRozmej (approved):

PR #8911, opened 2025-07-02
Feature/Address state by block header

asdacap:
- Regardless of paprika, flat, archive or something, just state root is not enough to identify a state. 
  - Two block may (in theory) have the same state root.
  - Some state does not store state root or index the state by block number instead of state root, or maybe blockhash for some reason.
- Because of this, high level state interface should use block header to address a particular state.
  - Technically, we probably just need (block number, state root) tuple, but using the whole block header feels more proper. Plus, like 90% of the time, block header is already available.
- This PR change the interface of `IWorldState` and `IStateReader` to address a particular state via `BlockHeader` and establish the code path to do so.
- Notably `IWorldState.StateRoot` is no longer mutable, replaced with `IWorldState.SetBaseBlock(IBlockHeader)`.
- Implementation wise, its exactly the same as setting the state root from the block header.
- CommitTree still use block number because that function itself calculate state root and you need state root for the block header.

## Changes

- Change `Hash256 stateRoot` to `IBlockHeader baseBlock` where possible.  

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- [x] Mainnet can sync
- [ ] Engine hive test passed
- [ ] Sync github action OK.


src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
```diff
@@ -42,24 +59,21 @@ public Task PreWarmCaches(Block suggestedBlock, Hash256? parentStateRoot, IRelea
                 if (_logger.IsWarn) _logger.Warn($"Caches {result} are not empty. Clearing them.");
             }
 
-            if (!IsGenesisBlock(parentStateRoot) && _concurrencyLevel > 1 && !cancellationToken.IsCancellationRequested)
+            if (parent is not null && _concurrencyLevel > 1 && !cancellationToken.IsCancellationRequested)
```
asdacap, line 64: Change here, parent can be genesis (this block is second block)

src/Nethermind/Nethermind.Shutter/ShutterBlockHandler.cs
```diff
@@ -197,7 +197,7 @@ private void CheckAllValidatorsRegistered(in BlockHeader parent, in ShutterValid
             return;
         }
 
-        using IReadOnlyTxProcessingScope scope = _envFactory.Create().Build(parent.StateRoot!);
+        using IReadOnlyTxProcessingScope scope = _envFactory.Create().Build(parent);
```
asdacap: TODO: Have support for this pattern.

src/Nethermind/Nethermind.Trie.Test/TrieTests.cs
```diff
@@ -798,9 +798,9 @@ private static IEnumerable<TrieStoreConfigurations> CreateTrieStoreConfiguration
         [Repeat(10)]
         public void Fuzz_accounts(
             (TrieStoreConfigurations trieStoreConfig,
-            int accountsCount,
-            int blocksCount,
-            int uniqueValuesCount) test)
+                int accountsCount,
```
asdacap, line 802: Anyone knows the linter rule to make this consistent? Intellij just autoformat this somehow.

src/Nethermind/Nethermind.Consensus/Processing/ReadOnlyTxProcessingScope.cs
```diff
@@ -1,28 +1,23 @@
...
-    IWorldState worldState,
-    Hash256 originalStateRoot
+    IWorldState worldState
 ) : IReadOnlyTxProcessingScope
 {
     public void Dispose()
     {
-        Reset();
+        worldState.Reset();
+        worldState.SetBaseBlock(null);
```
asdacap, line 18: No more original state root. It is always reset to empty state. Consumer should always set base block first.

src/Nethermind/Nethermind.Flashbots/Modules/Rbuilder/RbuilderRpcModule.cs
```diff
@@ -149,7 +149,6 @@ public IOverridableWorldScope Create()
 
         public bool Return(IOverridableWorldScope obj)
         {
-            obj.ResetOverrides();
```
asdacap: Removed to `BeginScope` dispose.

src/Nethermind/Nethermind.Blockchain/SpecificBlockReadOnlyStateProvider.cs
```diff
@@ -34,11 +35,12 @@ public void Accept<TCtx>(ITreeVisitor<TCtx> visitor, Hash256 stateRoot, Visiting
...
         }
 
-        public bool AccountExists(Address address) => _stateReader.TryGetAccount(StateRoot, address, out _);
+        public bool AccountExists(Address address) => _stateReader.TryGetAccount(BaseBlock, address, out _);
 
         [SkipLocalsInit]
         public bool IsEmptyAccount(Address address) => TryGetAccount(address, out AccountStruct account) && account.IsEmpty;
-        public bool HasStateForRoot(Hash256 stateRoot) => _stateReader.HasStateForRoot(stateRoot);
+
+        public bool HasStateForRoot(BlockHeader? header) => _stateReader.HasStateForBlock(header);
```
tanishqjasoria: can also renamed to HasStateForBlock

src/Nethermind/Nethermind.Shutter/ShutterBlockHandler.cs
asdacap: Added in #8920

src/Nethermind/Nethermind.Blockchain/SpecificBlockReadOnlyStateProvider.cs
asdacap: Fixed

src/Nethermind/Nethermind.Flashbots/Modules/Rbuilder/RbuilderRpcModule.cs
LukaszRozmej: Wouldn't that directly clear all the overrides that were just done? Seems off to me.

asdacap: Yes? It was the purpose?

benaadams:
> * Two block may (in theory) have the same state root.

Worried about `SHA3-256` hash collisions when a MPT root? 🤔

asdacap:
More like prevent having a stateroot->blocknumber mapping.

PR #8915, opened 2025-07-02
Only return peers in admin_peers that match the network id

skylenet:
## Changes

The `admin_peers` endpoint was returning peers that didn't belong to the same chain. This change will filter those temporary peers out and only return a list of peers that match the network id. 

Note: All the other clients (geth, reth, besu, erigon ..) return only peers that are on the same network.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Remarks

### DIFF
--- a/src/Nethermind/Nethermind.JsonRpc/Modules/Admin/AdminRpcModule.cs
+++ b/src/Nethermind/Nethermind.JsonRpc/Modules/Admin/AdminRpcModule.cs
@@ -105,8 +105,20 @@ public class AdminRpcModule : IAdminRpcModule
     }
 
     public ResultWrapper<PeerInfo[]> admin_peers(bool includeDetails = false)
-        => ResultWrapper<PeerInfo[]>.Success(
-            _peerPool.ActivePeers.Select(p => new PeerInfo(p.Value, includeDetails)).ToArray());
+    {
+        var validatedPeers = _peerPool.ActivePeers
+            .Where(p => IsValidatedPeer(p.Value))
+            .Select(p => new PeerInfo(p.Value, includeDetails))
+            .ToArray();
+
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.JsonRpc/Modules/Admin/AdminRpcModule.cs
```diff
@@ -156,8 +156,20 @@ public async Task<ResultWrapper<bool>> admin_removeTrustedPeer(string enode)
...
+            .Select(p => new PeerInfo(p.Value, includeDetails))
+            .ToArray();
+
+        return ResultWrapper<PeerInfo[]>.Success(validatedPeers);
+    }
+
+    private static bool IsValidatedPeer(Peer peer)
+    {
+        return peer.InSession?.IsNetworkIdMatched == true ||
+               peer.OutSession?.IsNetworkIdMatched == true;
```
LukaszRozmej, line 170: @asdacap why are even those peers there? Is it because for short period of time we don't know about their network id? Should PeerPool filter them internally?

asdacap: Yes, we don't know them temporarily.

LukaszRozmej:
Test `Test_peers` need to be fixed

LukaszRozmej:
test and code formatting still fails

skylenet:
@LukaszRozmej can you rerun the CI jobs? 🙏  The failed CI steps pass on my machine now.

PR #8916, opened 2025-07-02
Remove "constructor" in chiado chainspec

ak88:
Removes use of deprecated "constructor" in chiado chainspec.


ak88:
> Does that change genesis hash?

no thats why i added the expected hash in config now from 
https://gnosis-chiado.blockscout.com/block/0xada44fd8d2ecab8b08f256af07ad3e777f17fb434f8f8e678b312f576212ba9a

PR #8917, opened 2025-07-02
Reduce txPool allocations for blobs and txs

benaadams:
## Changes

- Pool blobs, proofs and commitments
- Extend tx pooling to return regular tx pool txs and reuse them
- Reduce allocations in PacketSender

Before
image

After
image

- [x] Optimization

#### Requires testing

- [x] No


stdevMac:
Closing: this one-day draft was self-declared unsound in the description ("might be broadcasting messages when remove from pool") and was never revisited; the touched tx-pool surfaces have since been rewritten (#10753, #11639, #12011), so the diff no longer applies.

PR #8918, opened 2025-07-02
Reduce tx requests

benaadams:
## Changes

- Don't request txs that already have in pool; already do this check for Eth68 but not for earlier protocols

- [x] Optimization

#### Requires testing

- [x] No


LukaszRozmej (approved):

src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -825,6 +823,11 @@ public bool RemoveTransaction(Hash256? hash)
             return hasBeenRemoved;
         }
 
+        public bool ContainsTx(Hash256 hash)
+            => _blobTransactions.ContainsKey(hash) ||
```
marcindsobczak, line 914: Eth68 is a requirement for blobs support so we shouldn't land in this part of code with blob tx

marcindsobczak (approved):
Generally looks fine, Eth68 is a requirement for blob txs so it wouldn't we useful for ethereum and most of the supported networks, but we still have support for eth66/eth67 so I'm fine with optimizing for them

src/Nethermind/Nethermind.TxPool/TxPool.cs
benaadams: 😢 then that raises the question of why we are getting so many blob transactions sent to us (that we already have in pool)

marcindsobczak: Good question. We shouldn't request them.
Potential reason - we are not validating received transactions - everyone can send blob tx to us and we will accept it, even if not requested.
Adding such validation is in my backlog, about 1,5 year already

flcl42: @marcindsobczak why does not IsKnown work?

marcindsobczak: It should work, we shouldn't request same tx multiple times.
But if we receive unrequested blob txs, we should probably disconnect peer and we are not doing it.

marcindsobczak: Or if you mean why we need to check both cache (IsKnown) and TxPool, below eth68 we actually don't need, as all txs are in memory.
After eth68 (so with blobs) there is a scenario, when node was restarted and txs were loaded from disc. In this case we don't have them in cache

flcl42: Maybe we should load them on start

benaadams: There seem to be multiple hashes for blob tx?

Does something different if in mempool form etc?

flcl42: Hash is the same for both forms, we expect mempool form to be received from peers. Can be replacement with higher fee?

LukaszRozmej:
@benaadams close?

PR #8920, opened 2025-07-03
Shareable tx processing source

asdacap:
- Some place we have a pattern like this `_envFactory.Create().Build(parent.StateRoot!)`, a build after create. 
- This is because the env (source) is not shareable, so we usually manually track them.
- This PR add a util class that automatically track them and reuse them from a thread pool, which is nice but not required.

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Seems fine...


src/Nethermind/Nethermind.Consensus/Processing/ShareableTxProcessingSource.cs
```diff
@@ -0,0 +1,65 @@
...
+/// <param name="envFactory"></param>
+public class ShareableTxProcessingSource(IReadOnlyTxProcessingEnvFactory envFactory): IShareableTxProcessorSource
+{
+    ObjectPool<IReadOnlyTxProcessorSource> _envPool = new DefaultObjectPool<IReadOnlyTxProcessorSource>(new EnvPoolPolicy(envFactory));
+
+    public IReadOnlyTxProcessingScope Build(BlockHeader? baseBlock)
+    {
+        IReadOnlyTxProcessorSource? source = _envPool.Get();
+        IReadOnlyTxProcessingScope? scope = source.Build(baseBlock?.StateRoot ?? Keccak.EmptyTreeHash);
+        return new ScopeWrapper(source, _envPool, scope);
```
LukaszRozmej, line 28: You could pool scopes too :D

LukaszRozmej (approved):

PR #8921, opened 2025-07-03
Cleanup/reduce use of specific simulate pattern

asdacap:
- The use of certain simulate specific type and hooks make it hard for other chain to automatically have eth simulate as they would have to implement their own variant. 
- These are actually unnecessary and their functionality can be achieve without creating change to core class.
- `TransactionProcessor.ShouldValidate` has no effect and is removed.
- `SimulateTransactionProcessor` simply add `opts |= ExecutionOptions.SkipValidation;` if `validate` is false.
  - The default transaction adapter call `Execute`, meaning by default its `ExecutionOptions.Commit`.
  - Adding `SkipValidation` it make it an `ExecutionOptions.SkipValidationAndCommit`, which is exactly what `Trace` does.
  - This functionality can therefore be implemented using a `ITransactionProcessorAdapter`.
- `SimulateBlockValidationTransactionsExecutor` changed to an adapter instead of a subclass of `BlockValidationTransactionsExecutor`. This is because `BlockValidationTransactionsExecutor` is reimplemented by taiko.
  - `EnhanceBlockExecutionContext` is removed, because its code can simply be added before calling the wrapped `ProcessTransactions`. 
  - `ProcessTransaction` simply change the opts. Which is implemented by simply wrapping the `ProcessTransactions`.
- The simulate specific args is passed via a custom `SimulateRequestState` class which is mutated before calling the block processor.
- This prevent having to recreate the block processor on each `payload.BlockStateCalls`. 
- This also allow using the same simulate env, which is not recycled here.
- Remove optimism specific code. Will add back later.
- Also, the env is not disposed. Which I guess is fine here because the whole thing is reconstructed on each call.

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

#### Notes on testing

- Any idea?


asdacap:
76 failure same as master.

rubo (approved):
> Remove optimism specific code. Will add back later.

@asdacap Could you please elaborate?

PR #8927, opened 2025-07-03
Remove lock in PersistentStorageProvider

benaadams:
## Changes

- Remove lock in PersistentStorageProvider; switch to Dictionary so we can use true/false update without needing synchronisation

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
+++ b/src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
@@ -35,7 +35,7 @@ internal sealed class PersistentStorageProvider : PartialStorageProviderBase
     private readonly ILogManager? _logManager;
     internal readonly IStorageTreeFactory _storageTreeFactory;
     private readonly Dictionary<AddressAsKey, StorageTree> _storages = new();
-    private readonly HashSet<AddressAsKey> _toUpdateRoots = new();
+    private readonly Dictionary<AddressAsKey, bool> _toUpdateRoots = new();
 
     /// <summary>
     /// EIP-1283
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
```diff
@@ -252,7 +252,7 @@ void UpdateRootHashesSingleThread()
         {
             foreach (KeyValuePair<AddressAsKey, StorageTree> kvp in _storages)
             {
-                if (!_toUpdateRoots.Contains(kvp.Key))
+                if (!_toUpdateRoots.ContainsKey(kvp.Key))
```
LukaszRozmej: `TryGetValue` and check value?

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
```diff
@@ -281,7 +281,7 @@ void UpdateRootHashesMultiThread()
                 static (i, state) =>
             {
                 ref var kvp = ref state.storages.GetRef(i);
-                if (!state.toUpdateRoots.Contains(kvp.Key))
+                if (!state.toUpdateRoots.ContainsKey(kvp.Key))
```
LukaszRozmej: `TryGetValue` and check value?

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
benaadams: Can do, but can only be true here

benaadams: Done

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
benaadams: Done

PR #8928, opened 2025-07-03
Add hash->blockNumber cache

benaadams:
## Changes

- Is looked up fairly frequently (most for block just processed, or prior) and currently always goes to db

Before
image

After
image

(Memory lookup is inlined)

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain/Headers/HeaderStore.cs
+++ b/src/Nethermind/Nethermind.Blockchain/Headers/HeaderStore.cs
@@ -99,23 +99,23 @@ public class HeaderStore : IHeaderStore
 
     public long? GetBlockNumber(Hash256 blockHash)
     {
-        long? blockNumber = GetBlockNumberFromBlockNumberDb(blockHash);
+        long? blockNumber = GetBlockNumberThroughCache(blockHash);
         if (blockNumber is not null) return blockNumber.Value;
 
         // Probably still hash based
         return Get(blockHash)?.Number;
     }
 
-    private long? GetBlockNumberFromBlockNumberDb(Hash256 blockHash)
+    private long? GetBlockNumberThroughCache(Hash256 blockHash)
     {
         if (_numberCache.TryGet(blockHash, out long value))
         {
             return value;
         }
-        return GetBlockNumberFromBlockNumberDbSlow(blockHash);
+        return GetBlockNumberFromBlockNumberDb(blockHash);
     }
 
-    private long? GetBlockNumberFromBlockNumberDbSlow(Hash256 blockHash)
+    private long? GetBlockNumberFromBlockNumberDb(Hash256 blockHash)
     {
         Span<byte> numberSpan = _blockNumberDb.GetSpan(blockHash);
         if (numberSpan.IsNullOrEmpty()) return null;
--- a/src/Nethermind/Nethermind.Blockchain/Headers/HeaderStore.cs
+++ b/src/Nethermind/Nethermind.Blockchain/Headers/HeaderStore.cs
@@ -126,7 +126,9 @@ public class HeaderStore : IHeaderStore
                 throw new InvalidDataException($"Unexpected number span length: {numberSpan.Length}");
             }
 
-            return BinaryPrimitives.ReadInt64BigEndian(numberSpan);
+            long number = BinaryPrimitives.ReadInt64BigEndian(numberSpan);
+            _numberCache.Set(blockHash, number);
+            return number;
         }
         finally
         {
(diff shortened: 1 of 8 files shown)


src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
```diff
@@ -480,6 +480,11 @@ private void FireProcessingQueueEmpty()
             Metrics.RecoveryQueueSize = Math.Max(_queueCount - blockQueueCount - (IsProcessingBlock ? 1 : 0), 0);
             Metrics.ProcessingQueueSize = blockQueueCount;
             _stats.UpdateStats(lastProcessed, processingBranch.Root, blockProcessingTimeInMicrosecs);
+
+            foreach (Block block in processedBlocks)
+            {
+                _blockTree.CacheBlockNumber(block.Hash, block.Header.Number);
```
LukaszRozmej: Do we need explicit call? Most of the processed blocks should be recently inserted, and for restarts we can think of something better.

src/Nethermind/Nethermind.Blockchain/Headers/HeaderStore.cs
```diff
@@ -85,18 +93,28 @@ public void InsertBlockNumber(Hash256 blockHash, long blockNumber)
...
         return Get(blockHash)?.Number;
     }
 
+    private long? GetBlockNumberThroughCache(Hash256 blockHash)
+    {
+        if (_numberCache.TryGet(blockHash, out long value))
+        {
+            return value;
+        }
+        return GetBlockNumberFromBlockNumberDb(blockHash);
```
LukaszRozmej, line 115: add to cache here?

benaadams: Added

src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
benaadams: Changed to last block

Do need explicit call for last block as we only insert after forkchoice

benaadams: Oh, get what you are saying. Let me check

benaadams: Yes for some reason (looking into this too) 

image

asdacap: GetTotalDifficulty need to die.

src/Nethermind/Nethermind.Blockchain/Headers/IHeaderStore.cs
```diff
@@ -14,6 +14,7 @@ public interface IHeaderStore
     void BulkInsert(IReadOnlyList<BlockHeader> headers);
     BlockHeader? Get(Hash256 blockHash, bool shouldCache, long? blockNumber = null);
     void Cache(BlockHeader header);
+    void Cache(Hash256 blockHash, long blockNumber);
```
asdacap, line 17: I don't like adding another method. IBlockTree is too fat.

src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
```diff
@@ -480,6 +480,8 @@ private void FireProcessingQueueEmpty()
             Metrics.RecoveryQueueSize = Math.Max(_queueCount - blockQueueCount - (IsProcessingBlock ? 1 : 0), 0);
             Metrics.ProcessingQueueSize = blockQueueCount;
             _stats.UpdateStats(lastProcessed, processingBranch.Root, blockProcessingTimeInMicrosecs);
+
+            _blockTree.CacheBlockNumber(lastProcessed.Hash, lastProcessed.Header.Number);
```
asdacap, line 484: If it goes to the blockchainProcessor, then previously it was already `SuggestBlock`, and therefore can be intercepted in blocktree. Hence, this is not needed.

src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
benaadams: If you want the full path 
image

src/Nethermind/Nethermind.Blockchain/Headers/IHeaderStore.cs
benaadams: Ok, trying to find out why blocks don't cache properly; that might resolve it

src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
benaadams: Was trying to set it after it is confirmed that the hash is valid for the block (i.e post processing); rather than take the jsonrpc's word for it

asdacap: The hash should be consistent with the header, I think. We don't actually store the block hash in RLP. It is recalculated.

benaadams:
Closing in preference of https://github.com/NethermindEth/nethermind/pull/8943

PR #8929, opened 2025-07-03
NewPayloadTimeout in miliseconds

MarekM25:
## Changes

Changing the value of this timeout to miliseconds

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [ ] No


src/Nethermind/Nethermind.Merge.Plugin/IMergeConfig.cs
```diff
@@ -64,8 +64,8 @@ The number of requests to the garbage collector (GC) to release the process memo
             """, DefaultValue = "25")]
     public int CollectionsPerDecommit { get; set; }
 
-    [ConfigItem(Description = "The timeout, in seconds, for the `engine_newPayload` method.", DefaultValue = "7", HiddenFromDocs = true)]
-    public double NewPayloadTimeout { get; }
+    [ConfigItem(Description = "The timeout, in miliseconds, for the `engine_newPayload` method.", DefaultValue = "7000", HiddenFromDocs = true)]
```
rubo: Typo:
```suggestion
    [ConfigItem(Description = "The timeout, in milliseconds, for the `engine_newPayload` method.", DefaultValue = "7000", HiddenFromDocs = true)]
```

src/Nethermind/Nethermind.Merge.Plugin/IMergeConfig.cs
```diff
@@ -64,8 +64,8 @@ The number of requests to the garbage collector (GC) to release the process memo
...
+    public int NewPayloadTimeoutMs { get; }
```
rubo: I'm not sure the "Ms" suffix is needed.

kamilchodola: This is highly unlikely that this flag is commonly used but we need to know the consequences:
1. Without rename if someone forced it to for example "15" - after this fix it will be 15ms  applied automatically
2. With rename node won't start as this flag does not exist.

Honestly preferring to change the name of flag so it will stop the node rather than automatically start time outing all the payloads.

Other option - keep legacy NewPayloadTimeout and add NewPayloadTimeoutMs as a separate flag and handle the case on what if both are set (pick higher for example and throw warning OR crash at initialization when both are set)

asdacap: I added this flag like last week. So very unused.

src/Nethermind/Nethermind.Merge.Plugin.Test/EngineModuleTests.V3.cs
```diff
@@ -519,7 +519,7 @@ static void MarkAsUnprocessed(MergeTestBlockchain chain, int blockNumber)
 
         MergeTestBlockchain chain = await CreateBlockchain(releaseSpec: Cancun.Instance, mergeConfig: new MergeConfig()
         {
-            NewPayloadTimeout = TimeSpan.FromDays(1).TotalSeconds
+            NewPayloadTimeoutMs = 1000
```
asdacap: The original values is in days, not second. Probably something with randomly failing test.

src/Nethermind/Nethermind.Merge.Plugin/IMergeConfig.cs
MarekM25: I think in this case we can assume this flag is not being used

MarekM25: However, the consequences are more confusing with option 1, so it's better not to start the node. I changed the flag name to NewPayloadBlockProcessingTimeout.

src/Nethermind/Nethermind.Merge.Plugin.Test/EngineModuleTests.V3.cs
MarekM25: Thx, fixed

src/Nethermind/Nethermind.Merge.Plugin/IMergeConfig.cs
flcl42: We usually use Ms suffix though

PR #8940, opened 2025-07-04
EIP-7594: Constant maxBlobsPerTx 

flcl42:
## Changes

- Use constant maxBlobsPerTx when PeerDAS is activated

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Core/Eip7594Constants.cs
+++ b/src/Nethermind/Nethermind.Core/Eip7594Constants.cs
@@ -3,7 +3,10 @@
 
 namespace Nethermind.Core;
 
-public class Eip7594Constants
+/// <summary>
+/// See <see href="https://eips.ethereum.org/EIPS/eip-7594">EIP-7594</see>
+/// </summary>
+public static class Eip7594Constants
 {
     public const int MaxBlobsPerTx = 6;
 }
--- a/src/Nethermind/Nethermind.Core/Eip7594Constants.cs
+++ b/src/Nethermind/Nethermind.Core/Eip7594Constants.cs
@@ -1,10 +1,9 @@
-// SPDX-FileCopyrightText: 2022 Demerzel Solutions Limited
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
 // SPDX-License-Identifier: LGPL-3.0-only
 
 namespace Nethermind.Core;
 
 public class Eip7594Constants
 {
-
     public const int MaxBlobsPerTx = 6;
 }
--- a/src/Nethermind/Nethermind.Core/Eip7594Constants.cs
+++ b/src/Nethermind/Nethermind.Core/Eip7594Constants.cs
@@ -1,0 +1,10 @@
+// SPDX-FileCopyrightText: 2022 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+namespace Nethermind.Core;
+
+public class Eip7594Constants
+{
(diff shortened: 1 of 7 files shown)


benaadams (approved):

src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.Blobs.cs
```diff
@@ -1020,59 +1020,38 @@ private Transaction CreateBlobTx(PrivateKey sender, UInt256 nonce = default, int
...
-
-                yield return MakeTestCase("Evicts when per block limit is set",
-                    new BlobScheduleSettings { Max = 6 }, new BlobScheduleSettings { Max = 5 }, [6], 0);
-                yield return MakeTestCase("Evicts when per tx limit only is set",
-                    new BlobScheduleSettings { Max = 6 }, new BlobScheduleSettings { Max = 6, MaxBlobsPerTx = 3 }, [6], 0);
-                yield return MakeTestCase("Evicts when per tx limit only is changed",
-                    new BlobScheduleSettings { Max = 6, MaxBlobsPerTx = 3 }, new BlobScheduleSettings { Max = 6, MaxBlobsPerTx = 2 }, [2, 3, 2], 1);
-                yield return MakeTestCase("Evicts next txs too",
-                    new BlobScheduleSettings { Max = 6 }, new BlobScheduleSettings { Max = 6, MaxBlobsPerTx = 3 }, [6, 3, 3, 3], 0);
-            }
```
emlautarom1, line 1075: What happened with all the previous test cases? No test had `blobTxBlobCounts = []` while the new test only checks this case.

src/Nethermind/Nethermind.Core/Eip7594Constants.cs
```diff
@@ -0,0 +1,9 @@
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+namespace Nethermind.Core;
+
+public class Eip7594Constants
+{
+    public const int MaxBlobsPerTx = 6;
```
emlautarom1, line 11: A link to a spec would be great for future maintainers.

src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.Blobs.cs
flcl42: We are testing now only 9 -> 6 test case, as MaxBlobsPerTx is simply set once by EIP-7594. Added the less_than_6 -> 6 case.

src/Nethermind/Nethermind.Core/Eip7594Constants.cs
flcl42: Good idea, added

flcl42:
> Couple of things i dont understand. Where does the number come from, where is it defined? I do not see it mentioned in 7594. We only use the number in tx pool, but not in consensus?

It's almost in, https://github.com/ethereum/EIPs/pull/9981/files

TxValidator(=consensus) validates this too: https://github.com/NethermindEth/nethermind/blob/const-max-blobs-per-tx/src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs#L208

PR #8943, opened 2025-07-05
Improve Block caching and reduce lookup allocations

benaadams:
## Changes

- Introduce block hash -> block number cache
- Layer header cache, depending if difficulty is available or canonical
- Add ability to only decode tx haches as used by `RemoveOldReceipts` rather than allocating everything; also use for `eth_eth_getBlockBy...` when only hashes
- Delay RemoveOldReceipts by half of the GC delay
- Introduce fast generic constructor, rather than the slower `new T()` vis CreateInstanceT
- Return header directly if found in cache
- Reduce delayed hash from 32kB to 1kB as it burns through the array pool when decoding blocks

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain/BlockTree.cs
+++ b/src/Nethermind/Nethermind.Blockchain/BlockTree.cs
@@ -27,7 +27,6 @@ using Nethermind.Logging;
 using Nethermind.Serialization.Rlp;
 using Nethermind.State.Repositories;
 using Nethermind.Db.Blooms;
-using Nethermind.Serialization.Json;
 
 namespace Nethermind.Blockchain
 {
@@ -563,7 +562,14 @@ namespace Nethermind.Blockchain
                 return null;
             }
 
-            BlockHeader? header = _headerStore.Get(blockHash, shouldCache: false, blockNumber: blockNumber);
+            bool totalDifficultyNeeded = (options & BlockTreeLookupOptions.TotalDifficultyNotNeeded) == BlockTreeLookupOptions.None;
+            bool requiresCanonical = (options & BlockTreeLookupOptions.RequireCanonical) == BlockTreeLookupOptions.RequireCanonical;
+            if (_headerStore.TryGetCache(blockHash, totalDifficultyNeeded, requiresCanonical, out BlockHeader? header))
+            {
+                return header;
+            }
+
+            header = _headerStore.Get(blockHash, blockNumber: blockNumber);
             if (header is null)
             {
                 bool allowInvalid = (options & BlockTreeLookupOptions.AllowInvalid) == BlockTreeLookupOptions.AllowInvalid;
@@ -576,9 +582,7 @@ namespace Nethermind.Blockchain
             }
 
             header.Hash ??= blockHash;
-            bool totalDifficultyNeeded = (options & BlockTreeLookupOptions.TotalDifficultyNotNeeded) == BlockTreeLookupOptions.None;
             bool createLevelIfMissing = (options & BlockTreeLookupOptions.DoNotCreateLevelIfMissing) == BlockTreeLookupOptions.None;
-            bool requiresCanonical = (options & BlockTreeLookupOptions.RequireCanonical) == BlockTreeLookupOptions.RequireCanonical;
 
             if ((totalDifficultyNeeded && header.TotalDifficulty is null) || requiresCanonical)
             {
@@ -615,7 +619,7 @@ namespace Nethermind.Blockchain
 
             if (header is not null && ShouldCache(header.Number))
             {
-                _headerStore.Cache(header);
+                _headerStore.Cache(header, hasDifficulty: totalDifficultyNeeded, requiresCanonical);
             }
 
             return header;
@@ -735,17 +739,6 @@ namespace Nethermind.Blockchain
             } while (current is not null && responseIndex < numberOfBlocks);
 
             return result;
-        }
-
-        private BlockHeader? GetAncestorAtNumber(BlockHeader header, long number)
-        {
-            BlockHeader? result = header;
-            while (result is not null && result.Number < number)
-            {
-                result = this.FindParentHeader(result, BlockTreeLookupOptions.TotalDifficultyNotNeeded);
-            }
-
-            return header;
         }
 
         private Hash256? GetBlockHashOnMainOrBestDifficultyHash(long blockNumber)
@@ -947,7 +940,7 @@ namespace Nethermind.Blockchain
                 if (ShouldCache(block.Number))
                 {
                     _blockStore.Cache(block);
-                    _headerStore.Cache(block.Header);
+                    _headerStore.Cache(block.Header, hasDifficulty: true);
                 }
 
                 ChainLevelInfo? level = LoadLevel(block.Number);
@@ -1013,17 +1006,18 @@ namespace Nethermind.Blockchain
             for (int i = 0; i < blocks.Count; i++)
             {
                 Block block = blocks[i];
-                if (ShouldCache(block.Number))
-                {
-                    _blockStore.Cache(block);
-                    _headerStore.Cache(block.Header);
-                }
 
                 // we only force update head block for last block in processed blocks
                 bool lastProcessedBlock = i == blocks.Count - 1;
 
                 // Where head is set if wereProcessed is true
-                MoveToMain(blocks[i], batch, wereProcessed, forceUpdateHeadBlock && lastProcessedBlock);
+                bool updateHead = forceUpdateHeadBlock && lastProcessedBlock;
+                MoveToMain(blocks[i], batch, wereProcessed, updateHead);
+                if (ShouldCache(block.Number))
+                {
+                    _blockStore.Cache(block);
+                    _headerStore.Cache(block.Header, hasDifficulty: true, isCanonical: updateHead);
+                }
             }
 
             TryUpdateSyncPivot();
@@ -1084,6 +1078,7 @@ namespace Nethermind.Blockchain
                     for (int i = 0; i < level.BlockInfos.Length; ++i)
                     {
                         level.BlockInfos[i].Metadata &= ~BlockMetadata.BeaconMainChain;
+                        _headerStore.ClearCanonicalCache(level.BlockInfos[i].BlockHash);
                     }
 
                     _chainLevelInfoRepository.PersistLevel(j, level, batch);
@@ -1098,13 +1093,15 @@ namespace Nethermind.Blockchain
                 {
                     for (int i = 0; i < level.BlockInfos.Length; ++i)
                     {
-                        if (level.BlockInfos[i].BlockHash == blockInfo.BlockHash)
+                        Hash256 blockHash = level.BlockInfos[i].BlockHash;
+                        if (blockHash == blockInfo.BlockHash)
                         {
                             level.BlockInfos[i].Metadata |= BlockMetadata.BeaconMainChain;
                         }
                         else
                         {
                             level.BlockInfos[i].Metadata &= ~BlockMetadata.BeaconMainChain;
+                            _headerStore.ClearCanonicalCache(blockHash);
                         }
                     }
 
@@ -1499,7 +1496,7 @@ namespace Nethermind.Blockchain
             if (block is not null && ShouldCache(block.Number))
             {
                 _blockStore.Cache(block);
-                _headerStore.Cache(block.Header);
+                _headerStore.Cache(block.Header, hasDifficulty: totalDifficultyNeeded, requiresCanonical);
             }
 
             return block;
--- a/src/Nethermind/Nethermind.Blockchain/BlockTree.cs
+++ b/src/Nethermind/Nethermind.Blockchain/BlockTree.cs
@@ -1078,7 +1078,7 @@ namespace Nethermind.Blockchain
                     for (int i = 0; i < level.BlockInfos.Length; ++i)
                     {
                         level.BlockInfos[i].Metadata &= ~BlockMetadata.BeaconMainChain;
-                        _headerStore.ClearCanonicalCache(level.BlockInfos[i].BlockHash);
+                        _headerStore.DeleteCanonicalCache(level.BlockInfos[i].BlockHash);
                     }
 
                     _chainLevelInfoRepository.PersistLevel(j, level, batch);
@@ -1101,7 +1101,7 @@ namespace Nethermind.Blockchain
                         else
                         {
                             level.BlockInfos[i].Metadata &= ~BlockMetadata.BeaconMainChain;
-                            _headerStore.ClearCanonicalCache(blockHash);
+                            _headerStore.DeleteCanonicalCache(blockHash);
                         }
                     }
 
--- a/src/Nethermind/Nethermind.Blockchain/BlockTree.cs
+++ b/src/Nethermind/Nethermind.Blockchain/BlockTree.cs
@@ -1438,7 +1438,8 @@ namespace Nethermind.Blockchain
                 block = _blockStore.Get(
                     blockNumber.Value,
                     blockHash,
-                    (options & BlockTreeLookupOptions.ExcludeTxHashes) != 0 ? RlpBehaviors.ExcludeHashes : RlpBehaviors.None,
+                    ((options & BlockTreeLookupOptions.ExcludeTxHashes) != 0 ? RlpBehaviors.ExcludeHashes : RlpBehaviors.None) |
+                    ((options & BlockTreeLookupOptions.OnlyTxHashes) != 0 ? RlpBehaviors.OnlyHashes : RlpBehaviors.None),
                     shouldCache: false);
             }
 
@@ -1493,7 +1494,7 @@ namespace Nethermind.Blockchain
                 }
             }
 
-            if (block is not null && ShouldCache(block.Number))
+            if (block is not null && (options & BlockTreeLookupOptions.OnlyTxHashes) == 0 && ShouldCache(block.Number))
             {
                 _blockStore.Cache(block);
                 _headerStore.Cache(block.Header, hasDifficulty: totalDifficultyNeeded, requiresCanonical);
--- a/src/Nethermind/Nethermind.Blockchain/BlockTreeMethodOptions.cs
+++ b/src/Nethermind/Nethermind.Blockchain/BlockTreeMethodOptions.cs
@@ -14,7 +14,8 @@ public enum BlockTreeLookupOptions
     DoNotCreateLevelIfMissing = 4,
     AllowInvalid = 8,
     ExcludeTxHashes = 16,
-    All = 31
+    All = 31,
+    OnlyTxHashes = 32
 }
 
 [Flags]
--- a/src/Nethermind/Nethermind.Blockchain/Headers/IHeaderStore.cs
+++ b/src/Nethermind/Nethermind.Blockchain/Headers/IHeaderStore.cs
@@ -16,7 +16,7 @@ public interface IHeaderStore
     BlockHeader? Get(Hash256 blockHash, long? blockNumber = null);
     bool Cache(BlockHeader header, bool hasDifficulty, bool isCanonical = false);
     bool TryGetCache(Hash256 blockHash, bool needsDifficulty, bool requiresCanonical, [NotNullWhen(true)] out BlockHeader? header);
-    void ClearCanonicalCache(Hash256 blockHash);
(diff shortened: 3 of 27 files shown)


src/Nethermind/Nethermind.Blockchain/BlockTreeMethodOptions.cs
```diff
@@ -14,7 +14,8 @@ public enum BlockTreeLookupOptions
     DoNotCreateLevelIfMissing = 4,
     AllowInvalid = 8,
     ExcludeTxHashes = 16,
-    All = 31
+    All = 31,
```
alexb5dh, line 17: Shouldn't `All` option also be updated? Or maybe renamed somehow to avoid confusion?

src/Nethermind/Nethermind.Blockchain/Headers/IHeaderStore.cs
```diff
@@ -12,8 +13,10 @@ public interface IHeaderStore
 {
     void Insert(BlockHeader header);
     void BulkInsert(IReadOnlyList<BlockHeader> headers);
-    BlockHeader? Get(Hash256 blockHash, bool shouldCache, long? blockNumber = null);
-    void Cache(BlockHeader header);
+    BlockHeader? Get(Hash256 blockHash, long? blockNumber = null);
+    bool Cache(BlockHeader header, bool hasDifficulty, bool isCanonical = false);
+    bool TryGetCache(Hash256 blockHash, bool needsDifficulty, bool requiresCanonical, [NotNullWhen(true)] out BlockHeader? header);
+    void DeleteCanonicalCache(Hash256 blockHash);
```
LukaszRozmej, line 19: a bit weird that caching is exposed

src/Nethermind/Nethermind.Blockchain/Receipts/PersistentReceiptStorage.cs
```diff
@@ -72,24 +73,34 @@ public PersistentReceiptStorage(
...
-                }
-            });
+            long blockNumber = newMain.Number - txLookupLimit;
+            // Don't block main loop
+            Task.Run(() => RemoveOldReceipts(blockNumber));
+        }
+
+        private async Task RemoveOldReceipts(long blockNumber)
+        {
+            await Task.Delay(GCConfig.GCBlockProcessingDelayMs / 2);
```
LukaszRozmej, line 95: why delay?

src/Nethermind/Nethermind.Core/Utils/ObjectCreator.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        NewExpression newExpr = Expression.New(ci);
+        Expression<Func<T>> lambda = Expression.Lambda<Func<T>>(
+            newExpr,
+            name: $"Create{typeof(T).Name}",
+            tailCall: false,
+            parameters: Array.Empty<ParameterExpression>()
+        );
+        return lambda.Compile();
+    }
```
LukaszRozmej, line 29: In newest dotnet Activator isn't similar speed? There were improvements in dontet 8 and 9 to it

LukaszRozmej (approved):

stdevMac:
@benaadams stale-PR check: dormant ~10 months, and the #12011 struct-RLP rewrite means a rebase here is effectively a rewrite. Are you planning to come back to this, or OK to close? If there's no reply in the next few days, we'll close this as part of the stale-PR cleanup — it can always be reopened or resubmitted.

src/Nethermind/Nethermind.Blockchain/BlockTree.cs
```diff
@@ -563,7 +562,14 @@ public AddBlockResult SuggestBlock(Block block, BlockTreeSuggestOptions options
                 return null;
             }
 
-            BlockHeader? header = _headerStore.Get(blockHash, shouldCache: false, blockNumber: blockNumber);
+            bool totalDifficultyNeeded = (options & BlockTreeLookupOptions.TotalDifficultyNotNeeded) == BlockTreeLookupOptions.None;
+            bool requiresCanonical = (options & BlockTreeLookupOptions.RequireCanonical) == BlockTreeLookupOptions.RequireCanonical;
+            if (_headerStore.TryGetCache(blockHash, totalDifficultyNeeded, requiresCanonical, out BlockHeader? header))
+            {
+                return header;
```
flcl42, line 569: **[MEDIUM] Canonical header entries survive execution-chain reorgs**

`UpdateMainChain(..., forceUpdateHeadBlock: true)` now places the selected head in the canonical cache. A later ordinary main-chain reorg swaps or unmarks the relevant chain levels but never calls `DeleteCanonicalCache` for the displaced execution-chain hashes; the only invalidations are in `UpdateBeaconMainChain`. Because this cache hit returns without reloading the level, the former head can continue to satisfy an EIP-1898 `requireCanonical` lookup after it has become noncanonical.

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
```diff
@@ -363,7 +363,7 @@ public ResultWrapper<BlockForRpc> eth_getBlockByNumber(BlockParameter blockParam
 
     protected virtual ResultWrapper<BlockForRpc?> GetBlock(BlockParameter blockParameter, bool returnFullTransactionObjects)
     {
-        SearchResult<Block> searchResult = _blockFinder.SearchForBlock(blockParameter, true);
+        SearchResult<Block> searchResult = _blockFinder.SearchForBlock(blockParameter, allowNulls: true, onlyTxHashes: !returnFullTransactionObjects);
```
flcl42, line 366: **[MEDIUM] Hash-only block decoding underreports the RPC block size**

When an EIP-1898 `requireCanonical` hash lookup reaches this path with `returnFullTransactionObjects` false and the block cache misses, `OnlyHashes` creates transaction objects containing only their hashes. `BlockForRpc` nevertheless derives `size` by re-encoding those default-valued transaction objects through `BlockDecoder.GetLength`, rather than using the decoder-populated `Block.EncodedSize`. A historical block containing transactions is consequently returned with an incorrect, smaller RLP size.

src/Nethermind/Nethermind.JsonRpc/Modules/BlockFinderExtensions.cs
```diff
@@ -58,14 +58,14 @@ public static SearchResult<BlockHeader> SearchForHeader(this IBlockFinder blockF
...
-        public static SearchResult<Block> SearchForBlock(this IBlockFinder blockFinder, BlockParameter? blockParameter, bool allowNulls = false)
+        public static SearchResult<Block> SearchForBlock(this IBlockFinder blockFinder, BlockParameter? blockParameter, bool allowNulls = false, bool onlyTxHashes = false)
         {
             blockParameter ??= BlockParameter.Latest;
 
             Block block;
             if (blockParameter.RequireCanonical)
             {
-                block = blockFinder.FindBlock(blockParameter.BlockHash!, BlockTreeLookupOptions.RequireCanonical);
+                block = blockFinder.FindBlock(blockParameter.BlockHash!, BlockTreeLookupOptions.RequireCanonical | (onlyTxHashes ? BlockTreeLookupOptions.OnlyTxHashes : BlockTreeLookupOptions.None));
```
flcl42, line 68: **[LOW] Ordinary block parameters bypass hash-only decoding**

The new `onlyTxHashes` flag is forwarded only in the `RequireCanonical` branch at this line. Standard `eth_getBlockByHash`, numeric `eth_getBlockByNumber`, transaction-count, and uncle-count requests use parameters with `RequireCanonical` false, so the existing fallback still calls `FindBlock(blockParameter)` without `OnlyTxHashes`. Those requests continue decoding every transaction field and retain the allocations that the new call sites are intended to avoid; only an EIP-1898 canonical hash object reaches hash-only decoding.

src/Nethermind/Nethermind.Merge.Plugin/GC/GCKeeper.cs
```diff
@@ -151,7 +152,7 @@ private async Task ScheduleGCInternal()
             // This should give time to finalize response in Engine API
             // Normally we should get block every 12s (5s on some chains)
             // Lets say we process block in 2s, then delay 125ms, then invoke GC
-            await Task.Delay(125);
+            await Task.Delay(GCConfig.GCBlockProcessingDelayMs);
```
flcl42, line 155: **[LOW] GC scheduling comment still describes the previous 125ms delay**

The comment above this line reasons about the timing budget as "we process block in 2s, then delay 125ms, then invoke GC", but the code now waits GCConfig.GCBlockProcessingDelayMs, which is 250ms. The doubled delay is plausible on its own merits, but the in-place rationale now misdescribes the actual timing, so a future reader tuning this path would reason from an incorrect budget. A comment that reflects the configured value would avoid that ambiguity.

src/Nethermind/Nethermind.Blockchain/Receipts/PersistentReceiptStorage.cs
flcl42, line 95: **[LOW] Receipt tx-index pruning delay derives from the GC tuning constant**

RemoveOldReceipts waits GCConfig.GCBlockProcessingDelayMs / 2 before deleting pruned transaction-index entries. That constant lives in Nethermind.Config and is named and documented for GC scheduling after block processing (it is also used by GCKeeper for forcing collections). Retuning the GC delay for garbage-collection reasons would silently change the receipt-pruning lag, and conversely the pruning path has no knob or named constant of its own, so the rationale for the 125ms lag is not discoverable from the receipts code. A dedicated, separately named delay for the pruning path would keep the two behaviors independently adjustable.

flcl42 (commented):
Found 5 issues (2 medium and 3 low) in `src/Nethermind/Nethermind.Blockchain/BlockTree.cs`, `src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs`, `src/Nethermind/Nethermind.JsonRpc/Modules/BlockFinderExtensions.cs`, `src/Nethermind/Nethermind.Merge.Plugin/GC/GCKeeper.cs`, and 1 other file.

src/Nethermind/Nethermind.Blockchain/BlockTree.cs
```diff
@@ -563,7 +562,14 @@ public AddBlockResult SuggestBlock(Block block, BlockTreeSuggestOptions options
                 return null;
             }
 
-            BlockHeader? header = _headerStore.Get(blockHash, shouldCache: false, blockNumber: blockNumber);
+            bool totalDifficultyNeeded = (options & BlockTreeLookupOptions.TotalDifficultyNotNeeded) == BlockTreeLookupOptions.None;
+            bool requiresCanonical = (options & BlockTreeLookupOptions.RequireCanonical) == BlockTreeLookupOptions.RequireCanonical;
+            if (_headerStore.TryGetCache(blockHash, totalDifficultyNeeded, requiresCanonical, out BlockHeader? header))
```
flcl42, line 567: **[MEDIUM] Canonical header cache entries survive execution-chain reorgs**

Once a header has been looked up with `RequireCanonical`, it remains in `_headerCanonicalCache`, and forced heads are also inserted there. This fast path trusts that entry, while `UpdateMainChain` swaps blocks or clears `HasBlockOnMainChain` without deleting displaced hashes; the only invalidation calls are in `UpdateBeaconMainChain`. After an execution-chain reorg or head rollback that does not run beacon-chain invalidation, `FindHeader(oldHash, RequireCanonical)` can therefore return an orphan until cache eviction, allowing canonicality checks to succeed for a noncanonical header.

src/Nethermind/Nethermind.Facade/Simulate/SimulateDictionaryHeaderStore.cs
```diff
@@ -75,4 +69,12 @@ public void InsertBlockNumber(Hash256 blockHash, long blockNumber)
     {
         return _blockNumberDict.TryGetValue(blockHash, out var blockNumber) ? blockNumber : readonlyBaseHeaderStore.GetBlockNumber(blockHash);
     }
+
+    public bool TryGetCache(Hash256 blockHash, bool needsDifficulty, bool requiresCanonical, [NotNullWhen(true)] out BlockHeader? header)
+    {
+        header = null;
+        return !requiresCanonical && _headerDict.TryGetValue(blockHash, out header);
```
flcl42, line 76: **[MEDIUM] Simulation header cache ignores total-difficulty requirements**

The new cache probe ignores `needsDifficulty`, while `Cache` stores entries regardless of its `hasDifficulty` argument. A recent uncached database header first requested from the simulation tree with `TotalDifficultyNotNeeded` is consequently cached with `TotalDifficulty` unset; a later ordinary `FindHeader` returns that entry from this fast path instead of loading its `BlockInfo`. Total difficulty in simulation lookups now depends on which lookup option ran first.

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
flcl42, line 366: **[MEDIUM] Hash-only block decoding yields an incorrect RPC block size**

When an uncached non-empty block is requested through an accepted `{blockHash, requireCanonical:true}` block parameter and `returnFullTransactionObjects` is false, this call enables `OnlyTxHashes`. The decoder leaves every transaction except `Hash` at default values, but `BlockForRpc` calculates `size` by re-encoding those stubs rather than using the block's recorded `EncodedSize`. The response consequently reports a smaller size than the stored block RLP.

src/Nethermind/Nethermind.JsonRpc/Modules/BlockFinderExtensions.cs
flcl42, line 68: **[LOW] Hash-only lookup option is dropped for ordinary block parameters**

The new parameter is consulted only inside the `RequireCanonical` branch. Ordinary block-hash, block-number, and tag parameters take the other branch and call `FindBlock(blockParameter)` without `OnlyTxHashes`. Uncached historical calls from the block, transaction-count, and uncle-count endpoints therefore still fully decode and allocate every transaction payload field even though those callers now request hash-only lookup.

src/Nethermind/Nethermind.Blockchain/BlockTree.cs
```diff
@@ -615,7 +619,7 @@ public AddBlockResult SuggestBlock(Block block, BlockTreeSuggestOptions options
 
             if (header is not null && ShouldCache(header.Number))
             {
-                _headerStore.Cache(header);
+                _headerStore.Cache(header, hasDifficulty: totalDifficultyNeeded, requiresCanonical);
```
flcl42, line 622: **[MEDIUM] Header cached as difficulty-complete while TotalDifficulty remains unset**

In BlockTree.FindHeader, the header is written to the split cache with hasDifficulty taken from the request's totalDifficultyNeeded flag rather than from whether TotalDifficulty was actually populated. When the level load cannot fill it — for example a lookup with DoNotCreateLevelIfMissing (used by NewPayloadHandler and ForkchoiceUpdatedHandler for parent headers) whose block info is missing because block storage is not transactional, or the beacon-search guard — header.TotalDifficulty stays null yet the header is placed in the with-difficulty partition. The new TryGetCache fast path at the top of FindHeader then returns that entry to every later difficulty-needing lookup without re-checking TotalDifficulty, whereas the previous flow re-ran the (totalDifficultyNeeded && header.TotalDifficulty is null) branch on each call and populated it once the block info became available. Consumers of parentHeader.TotalDifficulty in Engine API validation therefore observe a null difficulty until cache eviction. The BeaconHeadersSyncTests change that now excludes TotalDifficulty from the BestSuggestedHeader equivalency is consistent with headers lacking TD surfacing from this path. Keying the partition on header.TotalDifficulty actually being set, or re-checking null TD on a fast-path hit, would restore the previous self-correcting behavior.

src/Nethermind/Nethermind.Merge.Plugin/GC/GCKeeper.cs
```diff
@@ -151,7 +152,7 @@ private async Task ScheduleGCInternal()
             // This should give time to finalize response in Engine API
             // Normally we should get block every 12s (5s on some chains)
             // Lets say we process block in 2s, then delay 125ms, then invoke GC
```
flcl42, line 154: **[LOW] GC scheduling comment no longer matches the configured delay**

The comment above ScheduleGCInternal's delay still reads 'Lets say we process block in 2s, then delay 125ms, then invoke GC', but the delay was changed from a hard-coded 125ms to GCConfig.GCBlockProcessingDelayMs, which is 250ms. The stated timing in the comment is now half the actual delay, which can mislead future tuning of the post-block GC window. Wording that references the shared constant instead of a literal value would keep the comment accurate.

flcl42 (commented):
Found 6 issues (4 medium and 2 low) in `src/Nethermind/Nethermind.Blockchain/BlockTree.cs`, `src/Nethermind/Nethermind.Facade/Simulate/SimulateDictionaryHeaderStore.cs`, `src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs`, `src/Nethermind/Nethermind.JsonRpc/Modules/BlockFinderExtensions.cs`, and 1 other file.

src/Nethermind/Nethermind.Blockchain/BlockTree.cs
```diff
@@ -1496,10 +1494,10 @@ private bool ShouldCache(long number)
                 }
             }
 
-            if (block is not null && ShouldCache(block.Number))
+            if (block is not null && (options & BlockTreeLookupOptions.OnlyTxHashes) == 0 && ShouldCache(block.Number))
```
wurdum, line 1497: Forwarding `rlpBehaviors` into the tx array makes `ExcludeTxHashes` live for the first time, and transactions decoded that way have `Hash == null`. The cache guard here masks only `OnlyTxHashes`, so `SyncServer.Find` serving a peer's `GetBlockBodies`, or `eth_feeHistory`, publishes such a block into the shared `_blockCache`; later readers get null tx hashes in `eth_getBlockByHash` and an NRE in `ReceiptsRecovery`. Either mask `ExcludeTxHashes` out of the cache guard or forward only `OnlyHashes`.

Repro flow:
```
peer GetBlockBodies -> SyncServer.cs:401  ExcludeTxHashes
  -> BlockTree.cs:1441  RlpBehaviors.ExcludeHashes
  -> BlockTree.cs:1499  _blockStore.Cache(block)
  -> ReceiptsRecovery.cs:136  transaction.Hash!  => NRE
```

Suggested fix:
```csharp
const BlockTreeLookupOptions partialDecode =
    BlockTreeLookupOptions.OnlyTxHashes | BlockTreeLookupOptions.ExcludeTxHashes;
if (block is not null && (options & partialDecode) == 0 && ShouldCache(block.Number))
```

src/Nethermind/Nethermind.Serialization.Rlp/TxDecoder.cs
```diff
@@ -128,7 +128,15 @@ public void Decode(ref Rlp.ValueDecoderContext decoderContext, ref T? transactio
             }
         }
 
-        GetDecoder(txType).Decode(ref Unsafe.As<T, Transaction>(ref transaction), txSequenceStart, transactionSequence, ref decoderContext, rlpBehaviors);
+        if (rlpBehaviors.HasFlag(RlpBehaviors.OnlyHashes) && !rlpBehaviors.HasFlag(RlpBehaviors.InMempoolForm))
+        {
+            BaseTxDecoder<Transaction>.CalculateHash(transaction ??= ObjectCreator<T>.Create(), transactionSequence, forceHashes: true);
```
wurdum, line 133: This branch runs before `GetDecoder(txType)`, so an unknown transaction type in a stored body decodes into a hash instead of throwing, and that path is live via `RemoveOldReceipts`. Under `SkipTypedWrapping` the problem is worse: `transactionSequence` is the whole remaining buffer, so the hash is fabricated and `Position` jumps past the rest of the array. Excluding `SkipTypedWrapping` alongside `InMempoolForm` and keeping the `GetDecoder` lookup covers both.
```csharp
if (rlpBehaviors.HasFlag(RlpBehaviors.OnlyHashes)
    && (rlpBehaviors & (RlpBehaviors.InMempoolForm | RlpBehaviors.SkipTypedWrapping)) == 0)
{
    _ = GetDecoder(txType);
    transaction ??= ObjectCreator<T>.Create();
    transaction.Type = txType;
    ...
}
```

wurdum (approved):

PR #8951, opened 2025-07-07
Align/refactor RPC output with geth

### LINKED ISSUE #8813: eth_call discrepancies with Geth
If we query a Geth Node and a Nethermind node over an eth_call, you may see some inconsistencies in the response. I have attached some examples:

### Test 1

Request:
```
[{"data":"0x6352211e0000000000000000000000000000000000000000000000000000000000001178","to":"0xB83C3CA6e22EF50EC13dC56B6D0729Aef6b4546E"},"latest"]
```

Geth:
```
{
  "error": {
    "code": 3,
    "data": "0xdf2d9b42",
    "message": "execution reverted"
  },
  "id": 1,
  "jsonrpc": "2.0"
}
```

Nethermind:
```
{
  "error": {
    "code": -32015,
    "data": "err: 0xdf2d9b42 (supplied gas 36000000)",
    "message": "VM execution error."
  },
  "id": 1,
  "jsonrpc": "2.0"
}
```

### Test 2

Request:
```
[{"data":"0x70a08231000000000000000000000000fa94a4ef06abb15d7f1953906f953d53faf3a726","to":"0xf61E18A4f2b1A3784054393Fce90848394Fc3799"},"latest"]
```

Geth:
```
{
  "error": {
    "code": 3,
    "data": "0x",
    "message": "execution reverted"
  },
  "id": 1,
  "jsonrpc": "2.0"
}
```

Nethermind:
```
{
  "error": {
    "code": -32015,
    "data": "err: revert (supplied gas 36000000)",
    "message": "VM execution error."
  },
  "id": 1,
  "jsonrpc": "2.0"
}
```

ak88:
Fixes #8813

Aligns Error codes with geth using -32000 for both bad input and any EVM exceptions, **except** for the special case of the top call frame executing revert opcode. This is technically speaking not an exception, but correct execution. The problem is that users still need to be notified that the whole tx reverted, so it is treated still as an error case. 

Geth chooses to return 3 as error code in this case
https://github.com/ethereum/go-ethereum/blob/efbba965b5bf9e192c1b935a800f9ffc5169505c/internal/ethapi/errors.go#L39

This is not following the official spec where error codes have to be negative, but i think their rationale is that this is technically not an error, but they still return an error object with a  positive number. 

In this PR Nethermind still returns -32000 for top frame reverts, but we can discuss if we should change it. 

"Wrong transaction nonce" was fixed by #8897

There are still some scenarios that differs from Geth. 
Geth's eth_estimateGas will return output when top frame reverts, but we do not because `ResultWrapper<>` data is strongly typed and will require refactoring to be able to return data more dynamically.

## Changes

- EvmExceptionType is now part of `TransactionResult` so `BlockchainBridge` can create error messages based on it
- Vm execution error is effectively not used anymore to align with Geth
- refactoring parts for simplification

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_


src/Nethermind/Nethermind.Evm.Test/TransactionProcessorEip7623Tests.cs
```diff
@@ -63,7 +63,7 @@ public void transaction_validation_intrinsic_below_floor(long gasLimit, bool isF
             .WithGasLimit(10000000).TestObject;
 
         TransactionResult result = _transactionProcessor.Execute(tx, new BlockExecutionContext(block.Header, _specProvider.GetSpec(block.Header)), NullTxTracer.Instance);
-        Assert.That(result.Fail, Is.EqualTo(isFail));
+        Assert.That(result.TransactionExecuted, Is.EqualTo(!isFail));
```
emlautarom1: Nitpick: invert the `isFail` in the test cases itself to `executed` (or similar) so that the assertion looks like:


```suggestion
        Assert.That(result.TransactionExecuted, Is.EqualTo(executed));
```

src/Nethermind/Nethermind.Evm/CallResult.cs
```diff
@@ -67,7 +67,7 @@ private CallResult(EvmExceptionType exceptionType)
         public bool ShouldRevert { get; }
         public bool? PrecompileSuccess { get; }
         public bool IsReturn => StateToExecute is null;
-        public bool IsException => ExceptionType != EvmExceptionType.None;
+        public bool IsException => ExceptionType != EvmExceptionType.None && ExceptionType != EvmExceptionType.Revert;
```
emlautarom1, line 71: We should add some comment here to explain the reasoning as to why `Revert` is not treated as an exception for future maintainers.

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
```diff
@@ -389,16 +389,17 @@ public IEnumerable<FilterLog> FindLogs(LogFilter filter, CancellationToken cance
             return logFinder.FindLogs(filter, cancellationToken);
         }
 
-        private static string? ConstructError(TransactionResult txResult, string? tracerError, long gasLimit)
+        private static string? ConstructError(TransactionResult txResult, string? tracerError)
```
emlautarom1, line 396: We no longer include `gasLimit`. Is that intended? Are we breaking anything by removing it?

emlautarom1 (approved):
LGTM, some small comments only

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
ak88: Good question - i do not know what this could break, but i have a feeling people generally rely more on Geth responses. Would be good to check if possible though... @kamilchodola any ideas how to assess this?

It was removed because it's not part of the geth response, and tbh doesn't make sense to return, because the gas limit is supplied by user input.

src/Nethermind/Nethermind.Evm/CallResult.cs
ak88: done

src/Nethermind/Nethermind.Evm.Test/TransactionProcessorEip7623Tests.cs
ak88: done

flcl42 (approved):

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
emlautarom1: Certain clients that use regexes to parse error messages might be affected, but this has been an issue for a long time (see https://github.com/ethereum/execution-apis/issues/463) so we might just not care.

I mention this since we might have our own tools/script that could rely on it though.

emlautarom1:
@ak88 Any reason why we haven't merged this PR yet?

PR #8956, opened 2025-07-08
Add EIP-7910: Add eth_config

flcl42:
## Changes

- Add eth_config

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No


src/Nethermind/Nethermind.Evm/Precompiles/IPrecompile.cs
```diff
@@ -10,6 +10,7 @@ namespace Nethermind.Evm.Precompiles
     public interface IPrecompile
     {
         static virtual Address Address => Address.Zero;
+        static virtual string Name => "";
```
emlautarom1, line 13: Is OK to default to `""` rather than force implementors to implement `Name`?


```suggestion
        static abstract string Name { get; }
```

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
```diff
@@ -772,6 +779,67 @@ private ResultWrapper<TResult> GetStateFailureResult<TResult>(BlockHeader header
...
+    public ResultWrapper<EthConfig> eth_config()
+    {
+        ForkActivationsSummary forks = forkInfo.GetForkActivationsSummary(_blockFinder.Head?.Header);
+
+        ForkConfig current = GetForkConfig(forks.Current, _specProvider);
+        ForkConfig? next = GetForkConfig(forks.Next, _specProvider);
+        ForkConfig? last = GetForkConfig(forks.Last, _specProvider);
+
+        // Fix address as key
+        JsonSerializerOptions defaultOptions = new(EthereumJsonSerializer.JsonOptionsIndented) { DictionaryKeyPolicy = null };
```
emlautarom1: Should we inject the base serializer and override the `DictionaryKeyPolicy` rather than instantiating a new serializer?

src/Nethermind/Nethermind.JsonRpc/Data/EthConfigResult.cs
```diff
@@ -0,0 +1,37 @@
...
+}
+
+public class ForkConfig
+{
+    public int? ActivationTime { get; init; }
+    public int? ActivationBlock { get; init; }
+    public required BlobScheduleSettingsForRpc? BlobSchedule { get; init; }
+    public required ulong ChainId { get; init; }
+    public required Dictionary<Address, string> Precompiles { get; init; }
+    public required Dictionary<string, Address> SystemContracts { get; init; }
```
emlautarom1: When serializing to JSON, can we guarantee the order of keys in the dictionary? Should we use `OrderedDictionary` instead?

src/Nethermind/Nethermind.Evm/Precompiles/IPrecompile.cs
flcl42: There is an interesting C# constraint that disallows this. It's not possible to use IPrecompile as a type for generic parameter, if a static member is abstract, like `IEnumerable<IPrecompile> Precompiles`, says `The interface 'IPrecompile' cannot be used as type argument. Static member 'IPrecompile.Name.get' does not have a most specific implementation in the interface.`
That's why we have Address = Zero I guess

src/Nethermind/Nethermind.JsonRpc/Data/EthConfigResult.cs
flcl42: Yep, let's do this

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
flcl42: Serializer has no interface for changing the options, but let me at least create the options once in ctor

ak88 (approved):

emlautarom1 (approved):

src/Nethermind/Nethermind.Evm.Precompiles/Snarks/Bn254MulPrecompile.cs
```diff
@@ -18,6 +18,8 @@ public class Bn254MulPrecompile : IPrecompile<Bn254MulPrecompile>
 
     public static Address Address { get; } = Address.FromNumber(7);
 
+    public static string Name => "BN256_MUL";
```
rubo, line 21: Why `BN256`?

flcl42: Simply [due to unknown reason](https://hackmd.io/@liangcc/bls-solidity#alt_bn128-bn256-bn254-why-so-many-names-and-so-confusing). This name is taken from https://eips.ethereum.org/EIPS/eip-7910

rubo: So nothing will break if I change to BN254. It's a confusing mess everywhere with this naming.

flcl42: It will break hash function and therefore the entire purpose of eth_config

flcl42: (unless you change both https://eips.ethereum.org/EIPS/eip-7910 and our code)

rubo: Ah, this damn naming is coming from EIP. Missed that. So we check for `Bn128...` to add `BN254...` classes which return `BN256...` 😁 Damn

flcl42: ((we provide hashes of fork configs, so fields order, fields names should be fully the same, and no whitespace))

flcl42: Yeah, as always naming is the hardest problem

rubo: > (unless you change both https://eips.ethereum.org/EIPS/eip-7910 and our code)

I took it seriously: #9023 😄

PR #8965, opened 2025-07-09
Fix Failing Ethereum Tests

SiddharthV1:
- Fix subsegment  metadata for segments starting at jumpdest after call or create. The proceeding subsegments should have entry point set to true but were being overwritten by the jumpdest metdata and being set to false. Hence Ilevm was not able to return to the original call frame if jumpdest followed a call or create

- Fix blob hash : TxExecutionContext is a ref field, that shouldn't be loaded as an address

### DIFF
--- a/src/Nethermind/Nethermind.Evm/CodeAnalysis/IL/IlAnalyzer.cs
+++ b/src/Nethermind/Nethermind.Evm/CodeAnalysis/IL/IlAnalyzer.cs
@@ -346,14 +346,7 @@ public static class IlAnalyzer
             }
             notStart = false;
 
-            if (op.IsCall() || op.IsCreate())
-            {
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Evm/CodeAnalysis/IL/IlAnalyzer.cs
```diff
@@ -340,6 +345,15 @@ internal static void AnalyzeSegment(byte[] fullcode, Range segmentRange, Diction
                     break;
             }
             notStart = false;
+
+            if (op.IsCall() || op.IsCreate())
```
Demuirgos: ```lastOpcodeIsCallOrCreate = op.IsCall() || op.IsCreate();```

PR #8967, opened 2025-07-09
Revise JSON-RPC docs generation

rubo:
## Changes

- Updated JSON-RPC docs generation to include namespaces like `flashbots`, `rbuilder`, `health`. Also, included lots of methods that were missing in the `admin` namespace. Generation results are available [here](https://docs.nethermind.io/next/interacting/json-rpc-ns/admin) (see sidebar namespaces too).
- Revised markdown

- [ ] Optimization
- [ ] Refactoring
- [x] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

### DIFF
--- a/tools/DocGen/JsonRpcGenerator.cs
+++ b/tools/DocGen/JsonRpcGenerator.cs
@@ -1,13 +1,13 @@
 // SPDX-FileCopyrightText: 2023 Demerzel Solutions Limited
 // SPDX-License-Identifier: LGPL-3.0-only
 
-using System.Reflection;
 using Nethermind.JsonRpc.Modules;
 using Nethermind.JsonRpc.Modules.Evm;
 using Nethermind.JsonRpc.Modules.Rpc;
 using Nethermind.JsonRpc.Modules.Subscribe;
 using Newtonsoft.Json;
 using Spectre.Console;
+using System.Reflection;
 
 namespace Nethermind.DocGen;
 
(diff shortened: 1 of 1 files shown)


benaadams (approved):

tools/DocGen/JsonRpcGenerator.cs
```diff
@@ -42,25 +48,53 @@ internal static void Generate(string path)
...
+
+            if (!methodMap.TryAdd(ns, methods))
+                methodMap[ns] = methodMap[ns].Concat(methods);
+        }
+
+        var i = 0;
+
+        foreach (var (ns, methods) in methodMap)
+        {
+            methodMap[ns] = (ns.Equals("eth", StringComparison.Ordinal)
```
alexb5dh: Won't 
```csharp
if(methodMap.ContainsKey("eth"))
  methodMap["eth"] = methodMap["eth"].Concat(...)
```
before `foreach` look cleaner?

alexb5dh (approved):

tools/DocGen/JsonRpcGenerator.cs
rubo: Right, it's not only cleaner, but it makes no sense to have it in the loop — copy-paste remnant. Thanks.

LukaszRozmej (approved):

PR #8968, opened 2025-07-09
Run background tasks at BelowNormal priority

benaadams:
## Changes

- Run background tasks at lower priority than main processing thread etc; requires introduction of a custom TaskScheduler to achieve this.
- Shutdown the scheduler during BlockProcessing
- Order tasks by timeout
- Add new metric for total tasks scheduled

Doesn't cause a regression, happily get processed (1.5M processed, max 8 queued)

image

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
+++ b/src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
@@ -240,6 +240,7 @@ public class BackgroundTaskScheduler : IBackgroundTaskScheduler, IAsyncDisposabl
         private readonly BlockingCollection<Task> _tasks = [];
         private readonly Thread[] workerThreads;
         private readonly ManualResetEventSlim _restartQueueSignal;
+        private readonly int _maxDegreeOfParallelism;
         private readonly ILogger _logger;
         private readonly CancellationToken _cancellationToken;
 
@@ -249,6 +250,7 @@ public class BackgroundTaskScheduler : IBackgroundTaskScheduler, IAsyncDisposabl
 
             _logger = logManager.GetClassLogger();
             _restartQueueSignal = restartQueueSignal;
+            _maxDegreeOfParallelism = maxDegreeOfParallelism;
             _cancellationToken = cancellationToken;
             workerThreads = [.. Enumerable.Range(0, maxDegreeOfParallelism)
                             .Select(i =>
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
```diff
@@ -111,6 +132,15 @@ private async Task StartChannel()
             {
                 if (_logger.IsError) _logger.Error($"Error processing background task {e}.");
             }
+            finally
+            {
+                cts.Dispose();
+            }
```
LukaszRozmej, line 138: why not just `using`?

src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
```diff
@@ -111,6 +132,15 @@ private async Task StartChannel()
...
+
+            continue;
+
+        Throttle:
+            await Task.Delay(millisecondsDelay: 1);
```
LukaszRozmej, line 143: huh?

src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
```diff
@@ -146,36 +178,142 @@ public async ValueTask DisposeAsync()
...
+            catch (OperationCanceledException)
+            {
+            }
+            catch (Exception e)
+            {
+                if (_logger.IsError) _logger.Error($"Error in background task processing {e}.");
+            }
+        }
+
+        public override int MaximumConcurrencyLevel => workerThreads.Length;
```
LukaszRozmej:
```suggestion
        public override int MaximumConcurrencyLevel => maxDegreeOfParallelism;
```

LukaszRozmej (approved):

src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
benaadams: Have just re-added an Action to the channel; so as the next read is going to return the same item; we pause the async briefly

src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
benaadams: goto gets upset; need to use goto to exit the double while (rather than just `break`)

PR #8973, opened 2025-07-10
Only warn about 4444 once

benaadams:
## Changes

- Only warn about 4444 once per run, rather than per request (very noisy)

image

- [x] Other: Logging

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
+++ b/src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
@@ -380,9 +380,10 @@ public class JsonRpcProcessor : IJsonRpcProcessor
         bool isSuccess = localErrorResponse is null;
         if (!isSuccess)
         {
-            if (localErrorResponse?.Error?.SuppressWarning == false)
+            int code = localErrorResponse?.Error?.Code ?? 0;
+            if (localErrorResponse?.Error?.SuppressWarning == false && ShouldWarn(code))
             {
-                if (_logger.IsWarn) _logger.Warn($"Error response handling JsonRpc Id:{request.Id} Method:{request.Method} | Code: {localErrorResponse.Error.Code} Message: {localErrorResponse.Error.Message}");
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
```diff
@@ -399,6 +400,17 @@ static bool HasNonWhitespace(ReadOnlySpan<byte> span)
         return result;
     }
 
+    private bool _hasWarnedAbout4444 = false;
+    private bool ShouldWarn(int code)
+    {
+        if (code == 4444)
+        {
+            return !Interlocked.Exchange(ref _hasWarnedAbout4444, true);
```
LukaszRozmej, line 408: I would drop interlocked. it is fine to do stale read here.

src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
```diff
@@ -399,6 +400,17 @@ static bool HasNonWhitespace(ReadOnlySpan<byte> span)
         return result;
     }
 
+    private bool _hasWarnedAbout4444 = false;
+    private bool ShouldWarn(int code)
```
LukaszRozmej, line 404: not sure, but maybe drop warnings only on engine API endpoint and keep them on standard RPC endpoint?

src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
```diff
@@ -380,9 +380,10 @@ static bool HasNonWhitespace(ReadOnlySpan<byte> span)
         bool isSuccess = localErrorResponse is null;
         if (!isSuccess)
         {
-            if (localErrorResponse?.Error?.SuppressWarning == false)
+            int code = localErrorResponse?.Error?.Code ?? 0;
+            if (localErrorResponse?.Error?.SuppressWarning == false && ShouldWarn(code))
```
LukaszRozmej, line 384: could we use `SupressWarning` for it?

benaadams: Didn't seem particularly easy to plumb that in on a request type/return type

alexb5dh (approved):

PR #8974, opened 2025-07-11
Fix Taiko Engine Api

asdacap:
- The auto collection resolver have a weird ordering behaviour with nested module causing the default engine api to be used instead of taiko's engine api as its registration appear later than taiko's. 
- This PR add some explicit DSL to ensure ordering.

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Logged the registered module order.

### DIFF
--- a/src/Nethermind/Nethermind.Core/Container/OrderedComponents.cs
+++ b/src/Nethermind/Nethermind.Core/Container/OrderedComponents.cs
@@ -5,7 +5,7 @@ using System.Collections.Generic;
 
 namespace Nethermind.Core.Container;
 
-public class OrderedComponents<T>: List<T>
+public class OrderedComponents<T>
 {
     private IList<T> _components = new List<T>();
     public IEnumerable<T> Components => _components;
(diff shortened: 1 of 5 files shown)


src/Nethermind/Nethermind.Core/Container/OrderedComponents.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Collections.Generic;
+
+namespace Nethermind.Core.Container;
+
+public class OrderedComponents<T>
+{
+    private IList<T> _components = new List<T>();
+    public IEnumerable<T> Components => _components;
+
+    public void AddLast(T item)
```
dipkakwani, line 13: Just wondering, this API isn't enforcing which will be the last element? In case of two consecutive `AddLast` calls, the first call becomes effectively invalid and it may get overridden. Would it make sense to somehow enforce calling these functions only once?

asdacap: Ow, I did not expect that use case.

asdacap: I guess that can make sense for other use case, but in this case, there are other rpc module also. We just need taiko to run after merge.

PR #8975, opened 2025-07-11
Fix/OOM when downloading from genesis.

asdacap:
- Block downloader can download block faster than nethermind can process.
- This causes OOM when syncing from genesis as the blocks are buffered in block processing queue.
- This PR stop spawing request when the size of the processing queue exceed certain configured threshold.
- This and previous download buffer size is not configured as MB, although internally it still use TX count.

- [X] Optimization
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- [x] Mainnet can sync normally


benaadams (approved):

benaadams:
sync test keeps failing `Prune_download_requests_map`

PR #8976, opened 2025-07-12
Cache PropertyInfo lookups for JsonRpc

benaadams:
## Changes

- Cache PropertyInfo lookups for JsonRpc rather than searching with reflection each time

Before

image

After

image

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.JsonRpc/JsonRpcService.cs
+++ b/src/Nethermind/Nethermind.JsonRpc/JsonRpcService.cs
@@ -9,8 +9,10 @@ using System.Runtime.CompilerServices;
 using System.Text;
 using System.Text.Json;
 using System.Text.Json.Serialization;
+using System.Threading;
 using System.Threading.Tasks;
 using Nethermind.Core;
+using Nethermind.Core.Collections;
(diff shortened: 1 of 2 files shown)


asdacap (approved):

src/Nethermind/Nethermind.JsonRpc/JsonRpcService.cs
```diff
@@ -244,6 +249,42 @@ private async Task<JsonRpcResponse> ExecuteAsync(JsonRpcRequest request, string
...
+            if (current.TryGetValue(type, out PropertyInfo? value))
+            {
+                return value;
+            }
+
+            // Copy-on-write: create a new dictionary so we don't mutate
+            // the one other threads may be reading without locks.
+            Dictionary<TypeAsKey, PropertyInfo?> propertyInfoCache = new(current);
+            PropertyInfo? propertyInfo = type.GetProperty("Result");
+            propertyInfoCache[type] = propertyInfo;
```
LukaszRozmej, line 278: maybe do frozendictionary then?

PR #8977, opened 2025-07-12
Remove use of Linq.Sum from hot paths

benaadams:
## Changes

image

- Use simple for loop rather than allocating a closure and ienumerable
- Add .Count property for StorageKeys so don't need to enumerate them to work out count

- [x] Optimization

#### Requires testing

- [x] No


src/Nethermind/Nethermind.Consensus.AuRa/Validators/PendingValidatorsDecoder.cs
```diff
@@ -84,6 +83,16 @@ private static (int Total, int Addresses) GetContentLength(PendingValidators ite
...
 
-        private static int GetAddressesLength(Address[] addresses) => addresses.Sum(Rlp.LengthOf);
+        private static int GetAddressesLength(Address[] addresses)
+        {
+            if (addresses.Length == 0) return 0;
+
+            int sum = 0;
+            foreach (Address address in addresses)
+            {
+                sum += Rlp.LengthOf(address);
```
LukaszRozmej: Address length is always constant unless it is null, but we probably can guarantee non-null here? So simple multiplication would suffice.

src/Nethermind/Nethermind.Serialization.Rlp/Eip2930/AccessListDecoder.cs
```diff
@@ -199,9 +198,15 @@ public AccessItemLengths(int indexesCount)
 
         private static int GetContentLength(AccessList accessList)
         {
-            return accessList
-                .Select(static entry => new AccessItemLengths(entry.StorageKeys.Count()))
```
LukaszRozmej, line 203: Should we remove `AccessItemLengths` type alltogether?

benaadams: Is still used by Encoding which uses it in parts

src/Nethermind/Nethermind.Consensus.AuRa/Validators/PendingValidatorsDecoder.cs
benaadams: Done

PR #8981, opened 2025-07-13
Isolate Tracer in Proof Module and remove IVisitingWorldState

asdacap:
- Isolate `ITracer` in `ProofModule` within `IOverridabeEnv`. 
  - This make it consistent with `TraceModule`.
  - This leaves only aura's `StartBlockProducerAura` with dangling `IReadOnlyTxProcessorSource.Build`.
- Move `Tracer.Accept` to use `IStateReader.RunTreeVisitor`.
- Inline `DumpState` to `IWorldState.DumpState`.
- Remove `IVisitingWorldState.Accept`. Its second parameter has a reference to a state. It should not be in `IWorldState` then.
- Remove `IVisitingWorldState` as it is unnecessary.

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

#### Notes on testing

- [x] Mainnet can sync normally.
- [x] Identical result when ccalling proof_call with master.


src/Nethermind/Nethermind.State/WorldState.cs
```diff
@@ -278,6 +278,13 @@ public void ResetTransient()
             _transientStorageProvider.Reset();
         }
 
+        public string DumpState()
+        {
+            TreeDumper dumper = new();
+            Accept(dumper, StateRoot);
+            return dumper.ToString();
+        }
+
```
LukaszRozmej: Does it need to be here? Can it be extension method on `StateProvider`?

src/Nethermind/Nethermind.Evm/State/IWorldState.cs
```diff
@@ -128,4 +128,6 @@ public interface IWorldState : IJournal<Snapshot>, IReadOnlyStateProvider
     void CommitTree(long blockNumber);
     ArrayPoolList<AddressAsKey>? GetAccountChanges();
     void ResetTransient();
+
+    string DumpState();
```
LukaszRozmej: extension method on StateProvider?

asdacap: Moved to state reader

LukaszRozmej: yeah I actually thought state reader when I wrote provider :)

PR #8983, opened 2025-07-13
Only access "warmup" hashtable once per warmup

benaadams:
## Changes

- Rather than doing two phases: check if cold, then warm up; we can just warm up and the hashtable will return if it existed already.

Before

image

After

image

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.cs
+++ b/src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.cs
@@ -366,7 +366,7 @@ internal static unsafe partial class EvmInstructions
             }
 
             // If the account is cold (and not a precompile), charge the cold access cost.
-            if (vmState.AccessTracker.WarmUp(address) && !address.IsPrecompile(spec))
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.cs
```diff
@@ -366,10 +366,9 @@ private static bool ChargeAccountAccessGas(ref long gasAvailable, VirtualMachine
             }
 
             // If the account is cold (and not a precompile), charge the cold access cost.
-            if (vmState.AccessTracker.IsCold(address) && !address.IsPrecompile(spec))
+            if (vmState.AccessTracker.WarmUp(address) && !address.IsPrecompile(spec))
```
LukaszRozmej:
```suggestion
            if (!address.IsPrecompile(spec) && vmState.AccessTracker.WarmUp(address))
```

otherwise you will warm-up precompiles

benaadams: Switched

PR #8986, opened 2025-07-14
Fail fast on Db corruption

### LINKED ISSUE #8220: When DB corruption is detected shut down node
When corruption is detected we just log message which needs manual intervention. The node should shut down.

When Corruption is detected in `DbOnTheRocks.CreateMarkerIfCorrupt` we should close the process in graceful way.

Current behavior needs manual intervention. If node auto-restarts then it can auto-recover.

benaadams:
Resolves #8220

## Changes

- Fail fast on Db corruption rather than continuing to run which may compound the error further and make the db more corrupt

image

Continuing leads to missing trienodes after repair etc

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
+++ b/src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
@@ -292,7 +292,7 @@ public partial class DbOnTheRocks : IDb, ITunableDb, IReadOnlyNativeKeyValueStor
             _fileSystem.File.WriteAllText(CorruptMarkerPath, "marker");
 
             // Don't kill tests checking corruption response
-            if (rocksDbException.Message != "Corruption: test corruption")
+            if (!rocksDbException.Message.Equals("Corruption: test corruption", StringComparison.Ordinal))
             {
                 Environment.FailFast("Fast shutdown due to DB corruption. Please restart.");
             }


src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
```diff
@@ -290,6 +290,12 @@ private void CreateMarkerIfCorrupt(RocksDbSharpException rocksDbException)
         {
             if (_logger.IsWarn) _logger.Warn($"Corrupted DB detected on path {_fullPath}. Please restart Nethermind to attempt repair.");
             _fileSystem.File.WriteAllText(CorruptMarkerPath, "marker");
+
+            // Don't kill tests checking corruption response
+            if (rocksDbException.Message != "Corruption: test corruption")
```
rubo: How reliable is this check?

benaadams: Is the exact phrase the tests raise (to check corrupt marker is written); nunit isn't happy about being failed fast 😅

rubo:
```suggestion
            if (!rocksDbException.Message.Equals("Corruption: test corruption", StringComparison.Ordinal))
```
Maybe we can also add `StringComparison.Ordinal` to the above `if`.

PR #8990, opened 2025-07-15
Don't process gossiped txs during block processing

benaadams:
## Changes

- Remove a source of unpredictability from block processing
- Can both be very heavy on memory allocations (with blob txs) and the head is just about to change; will get them after
- Reduces WorldState access from the txPool during block processing
- Clear the pending requests on each txPool head (to allow re-requesting)

- [x] Optimization

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Init/Steps/InitializeNetwork.cs
+++ b/src/Nethermind/Nethermind.Init/Steps/InitializeNetwork.cs
@@ -121,7 +121,7 @@ public class InitializeNetwork : IStep
         int maxPeersCount = _networkConfig.ActivePeersMaxCount;
         Network.Metrics.PeerLimit = maxPeersCount;
 
-        _api.TxGossipPolicy.Policies.Add(new SyncedTxGossipPolicy(_api.SyncModeSelector));
+        _api.TxGossipPolicy.Policies.Add(new SyncedTxGossipPolicy(_api.SyncModeSelector, () => _api.BlockTree.IsProcessingBlock));
 
         if (cancellationToken.IsCancellationRequested)
         {
--- a/src/Nethermind/Nethermind.Init/Steps/InitializeNetwork.cs
+++ b/src/Nethermind/Nethermind.Init/Steps/InitializeNetwork.cs
@@ -121,7 +121,7 @@ public class InitializeNetwork : IStep
         int maxPeersCount = _networkConfig.ActivePeersMaxCount;
         Network.Metrics.PeerLimit = maxPeersCount;
 
-        _api.TxGossipPolicy.Policies.Add(new SyncedTxGossipPolicy(_api.SyncModeSelector, () => _api.BlockTree.IsProcessingBlock));
+        _api.TxGossipPolicy.Policies.Add(new SyncedTxGossipPolicy(_api.SyncModeSelector));
 
         if (cancellationToken.IsCancellationRequested)
         {
--- a/src/Nethermind/Nethermind.Init/Steps/InitializeNetwork.cs
+++ b/src/Nethermind/Nethermind.Init/Steps/InitializeNetwork.cs
@@ -276,6 +276,7 @@ public class InitializeNetwork : IStep
             _networkConfig,
             _api.LogManager);
         PooledTxsRequestor pooledTxsRequestor = new(_api.TxPool!, _api.Config<ITxPoolConfig>(), _api.SpecProvider);
+        _api.DisposeStack.Push(pooledTxsRequestor);
 
         _api.ProtocolsManager = new ProtocolsManager(
             _api.SyncPeerPool!,
--- a/src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/PooledTxsRequestor.cs
+++ b/src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/PooledTxsRequestor.cs
@@ -15,17 +15,33 @@ using Nethermind.TxPool;
 
 namespace Nethermind.Network.P2P.Subprotocols.Eth
 {
-    public class PooledTxsRequestor(ITxPool txPool, ITxPoolConfig txPoolConfig, ISpecProvider specProvider) : IPooledTxsRequestor
+    public class PooledTxsRequestor : IPooledTxsRequestor, IDisposable
     {
         private const int MaxNumberOfTxsInOneMsg = 256;
-        private readonly bool _blobSupportEnabled = txPoolConfig.BlobsSupport.IsEnabled();
-        private readonly long _configuredMaxTxSize = txPoolConfig.MaxTxSize ?? long.MaxValue;
+        private readonly bool _blobSupportEnabled;
+        private readonly long _configuredMaxTxSize;
 
-        private readonly long _configuredMaxBlobTxSize = txPoolConfig.MaxBlobTxSize is null
+        private readonly long _configuredMaxBlobTxSize;
+
+        private readonly ClockKeyCache<ValueHash256> _pendingHashes = new(MemoryAllowance.RequestedTxHashCacheSize);
+        private readonly ITxPool _txPool;
+
+        public PooledTxsRequestor(ITxPool txPool, ITxPoolConfig txPoolConfig, ISpecProvider specProvider)
+        {
+            _txPool = txPool;
+            _blobSupportEnabled = txPoolConfig.BlobsSupport.IsEnabled();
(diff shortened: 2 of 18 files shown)


src/Nethermind/Nethermind.Init/Steps/InitializeNetwork.cs
```diff
@@ -121,7 +121,7 @@ private async Task Initialize(CancellationToken cancellationToken)
         int maxPeersCount = _networkConfig.ActivePeersMaxCount;
         Network.Metrics.PeerLimit = maxPeersCount;
 
-        _api.TxGossipPolicy.Policies.Add(new SyncedTxGossipPolicy(_api.SyncModeSelector));
+        _api.TxGossipPolicy.Policies.Add(new SyncedTxGossipPolicy(_api.SyncModeSelector, () => _api.BlockTree.IsProcessingBlock));
```
LukaszRozmej: It is weird that `IsProcessingBlock` is in BlockTree, can we remove it from there and only use the one in BlockchainProcessor?

src/Nethermind/Nethermind.Synchronization.Test/ParallelSync/SyncedTxGossipPolicyTests.cs
```diff
@@ -25,5 +25,5 @@ public class SyncedTxGossipPolicyTests
     [TestCase(SyncMode.Disconnected, ExpectedResult = false)]
     [TestCase(SyncMode.UpdatingPivot, ExpectedResult = false)]
     public bool can_gossip(SyncMode mode) =>
-        ((ITxGossipPolicy)new SyncedTxGossipPolicy(new StaticSelector(mode))).ShouldListenToGossipedTransactions;
+        ((ITxGossipPolicy)new SyncedTxGossipPolicy(new StaticSelector(mode), isProcessingBlock: () => false)).ShouldListenToGossipedTransactions;
```
LukaszRozmej:
```suggestion
        ((ITxGossipPolicy)new SyncedTxGossipPolicy(new StaticSelector(mode), isProcessingBlock: static () => false)).ShouldListenToGossipedTransactions;
```

src/Nethermind/Nethermind.Init/Steps/InitializeNetwork.cs
LukaszRozmej: Also maybe `NotProcessingBlocksGossipPolicy` instead of delegate?

src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/PooledTxsRequestor.cs
```diff
@@ -15,17 +15,33 @@
...
         private const int MaxNumberOfTxsInOneMsg = 256;
-        private readonly bool _blobSupportEnabled = txPoolConfig.BlobsSupport.IsEnabled();
-        private readonly long _configuredMaxTxSize = txPoolConfig.MaxTxSize ?? long.MaxValue;
+        private readonly bool _blobSupportEnabled;
+        private readonly long _configuredMaxTxSize;
 
-        private readonly long _configuredMaxBlobTxSize = txPoolConfig.MaxBlobTxSize is null
+        private readonly long _configuredMaxBlobTxSize;
+
+        private readonly ClockKeyCache<ValueHash256> _pendingHashes = new(MemoryAllowance.RequestedTxHashCacheSize);
```
LukaszRozmej, line 26: size reduced from 131k to just 4k?

src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/PooledTxsRequestor.cs
```diff
@@ -15,17 +15,33 @@
...
             ? long.MaxValue
             : txPoolConfig.MaxBlobTxSize.Value + (long)specProvider.GetFinalMaxBlobGasPerBlock();
 
-        private readonly ClockKeyCache<ValueHash256> _pendingHashes = new(MemoryAllowance.TxHashCacheSize);
+            _txPool.TxPoolHeadChanged += TxPoolHeadChanged;
+        }
+
+        private void TxPoolHeadChanged(object? sender, Block e)
+        {
+            _pendingHashes.Clear();
```
LukaszRozmej, line 43: Can't this cause that we don't request some transaction hashes?

LukaszRozmej (commented):
@marcindsobczak can you review?

src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/PooledTxsRequestor.cs
benaadams: Is twice the size of mempool; and we are now clearing each block

On the request side we check if what we are requesting is "known" by tx pool; so this size and not clearing is both over cautious and not under control of txPool to control re-requesting (if it kicks out a tx)

What this currently could mean is we request a tx; peer doesn't give it to us, so we never request it again until we have received 131_072 different txs

The change is so we won't request if it "known" by txpool or (is in last 2048 txs requested and same head block)

src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/PooledTxsRequestor.cs
benaadams: Should mean we request more

src/Nethermind/Nethermind.Init/Steps/InitializeNetwork.cs
benaadams: > It is weird that IsProcessingBlock is in BlockTree,

Plumbing issue; BlockchainProcessor updates BlockTree; everything has reference to blocktree, not much has reference to the processor

benaadams: Moved to own policy attached to BlockchainProcessor

src/Nethermind/Nethermind.Synchronization.Test/ParallelSync/SyncedTxGossipPolicyTests.cs
benaadams: Not delegate now

src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/PooledTxsRequestor.cs
LukaszRozmej: ok what is the point? we want to re-request transactions that are probably less valuable than the ones in pool?

benaadams: Pool kicks out some txs and then marks them not long term blocked, but blocked in this block; so we want to re-request them. Currently this second filter in the requestor will block that

Also these are marked when we request a tx; so as not to repeat requests; however we don't know if we didn't get a response for the request; currently they are blocked just by being requested; so now each head we clear that, so will rerequest ones that never got a response.

One of the filters is if we have in pool (or txpool has marked not interesting); so won't be requesting everything.

> we want to re-request transactions that are probably less valuable than the ones in pool?

Have a gap in pool of 100+ txs after head; so would want the ones we had previously rejected to fill?

LukaszRozmej: Ok can you show metrics of tx request messages before and after?

marcindsobczak (commented):
My intuition suggests this could lead to numerous unnecessary re-requests. We should verify this with a data-driven approach.
Alternatively, we could keep current cache and queue transaction requests, delaying them until the block has been processed.
We need to implement a mapping of requests-responses regardless, to disconnect malicious peers (like those sending unrequested blob transactions). This mapping could be made more universal and facilitate delayed requests.

smartprogrammer93 (commented):
Agree with Marcin on this. When i saw the title, i expected delaying the requests until after the processing is done.
This should also be configurable, as some MEV searchers would like to see transactions as soon as technically possible even if block processing is a tad slower

stdevMac:
@benaadams stale-PR check: reviewers asked for a delayed-queue approach ~8 months ago with no follow-up since. Still wanted in this form, or OK to close (the idea could move to an issue so it isn't lost)? If there's no reply in the next few days, we'll close this as part of the stale-PR cleanup — it can always be reopened or resubmitted.

PR #8993, opened 2025-07-15
Make gas price and premium gas calculation overridable

damian-orzechowski:
Make gas price and premium gas calculation overridable in classed derived from transaction processor.

## Changes
Make the following calculations as virtual methods of `TransactionProcessor` for ease of overriding:
- TryCalculatePremiumPerGas
- CalculateEffectiveGasPrice

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [ ] No


LukaszRozmej (approved):
Should those be virtual methods, or separate strategies?

damian-orzechowski:
> Should those be virtual methods, or separate strategies?

With the current way we override tx processing, I'd keep it virtual methods, but in the future we could think of redesigning it to cover different strategies e.g. to calculate intrinsic gas or calc effective price (if that is what you meant).

PR #8997, opened 2025-07-17
Refactor/Move init DB to DI

asdacap:
- Move db initialization to DI.
- Remove `StandardDbInitializer`. 
- Some `DbProvider` registration. `DbProvider` is now just a proxy to the container.
- Full pruning db is implement by overriding state db's `IDb`. No need for special code elsewhere.

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- [ ] Existing mainnet db run normally.
- [ ] Mainnet sync normally


src/Nethermind/Nethermind.Db/DbTracker.cs
```diff
@@ -0,0 +1,46 @@
...
+
+namespace Nethermind.Db;
+
+public class DbTracker
+{
+    private readonly ConcurrentDictionary<string, IDbMeta> _createdDbs = new ConcurrentDictionary<string, IDbMeta>();
+
+    public void AddDb(string name, IDbMeta dbMeta)
+    {
+        _createdDbs.TryAdd(name, dbMeta);
```
benaadams, line 16: Can this mean the db gets lost; as you add but fail because it already exists; however return one passed in?

Should it be GetOrAdd?

Does the one not added need to be disposed?

benaadams (approved):

src/Nethermind/Nethermind.Db/DbTracker.cs
asdacap: If we have duplicated name then we have a bigger problem. 

Disposal is handled by the container automatically.

LukaszRozmej: Should we do something if adding fail? Dispose? throw error? log error at least?

PR #9000, opened 2025-07-17
Feature/adjust rocksdb config by memory

asdacap:
Require #8997 

- Modified the rocksdb config depending on the pruning config and available memory.
- When available memory is more than 32GB, partitioned index is disabled and write buffer is increased.
- When archive mode is on, level multiplier is set to 10, and write buffer is increased again, to reduce write amplification.

- [X] Optimization

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- [X] Mainnet sync ok
- [X] Index size increased accordingly.
- [X] Available memory respond to cgroup limit.


benaadams (approved):

asdacap:
Rebased

Issue #8936, opened 2025-07-04
Non uniform json-rpc request for admin_peers


barnabasbusa:
The current implementation of admin_peers json-rpc method returns a very different format than every other EL node. 

Somewhat formalized standard return format:
```json
{
      "enr": "enr:-Ke4QBGzMXZeGTe9XW_9PhE5z3oRzB83mow2nu9WCoYLvU_TSnfPiXYVVE_HHkXxckhvfycPDRbV9-OMvpmTrEtL26qGAZerqnftg2V0aMvKhIe8XVCEaGMaIIJpZIJ2NIJpcIRAF5cbiXNlY3AyNTZrMaEDHFFJqZEE-iOQPEh-lOoaHKq3Ww88C9vGLQdUUdfjSzSEc25hcMCDdGNwgnxJg3VkcIJ8SQ",
      "enode": "enode://1c5149a99104fa23903c487e94ea1a1caab75b0f3c0bdbc62d075451d7e34b3405672ef5dc1234a906386cc07b9da1d904ac962cedbb088fc27f4cda2c0fc777@64.23.151.27:31817",
      "id": "ee691f32878517b2aa27b594a5c9508f8ba6124daf562c095847eac21253f3f6",
      "name": "Geth/v1.15.12-unstable-df12e79e-20250623/linux-amd64/go1.24.4",
      "caps": [
        "eth/68",
        "eth/69",
        "snap/1"
      ],
      "network": {
        "localAddress": "172.18.0.7:57830",
        "remoteAddress": "64.23.151.27:31817",
        "inbound": false,
        "trusted": false,
        "static": false
      },
      "protocols": {
        "eth": {
          "version": 69,
          "earliestBlock": 0,
          "latestBlock": 30449,
          "latestBlockHash": "0xfa1bb18f424d950c16f0a5d241c6759ab3f19a921ff1063a256ea97f1158dd61"
        },
        "snap": {
          "version": 1
        }
      }
    },
```

Nethermind response:
```json
{
      "id": "a6be3b7ad539e4968ee0588b1d2cb534565e0c912e0ec9133f14ff8808019f9e",
      "host": "172.16.0.11",
      "port": 30303,
      "address": "[::ffff:172.16.0.11]:30303",
      "isBootnode": false,
      "isTrusted": false,
      "isStatic": false,
[...]

PR #9003, opened 2025-07-17
Admin Get Peers

svlachakis:
Fixes Resolve https://github.com/NethermindEth/nethermind/issues/8936

## Changes

**BREAKING CHANGES** on admin_peers json rpc response in order to make it match with geth's response https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-admin#admin-peers

Before:

```
{
      "id": "a6be3b7ad539e4968ee0588b1d2cb534565e0c912e0ec9133f14ff8808019f9e",
      "host": "172.16.0.11",
      "port": 30303,
      "address": "[::ffff:172.16.0.11]:30303",
      "isBootnode": false,
      "isTrusted": false,
      "isStatic": false,
      "enode": "enode://8e29fa5cc9351c5d6fe939f5a559ddd380e6fbe6c32d5987dd8f4dfb9c8ca1edff0555a74c927fc117d22000d7eab44237880eae0ad3476ce084ff491709d922@172.16.0.11:30303",
      "inbound": true
}
```

After:
Screenshot 2025-07-18 at 12 42 01 AM

- [X] Breaking change (a change that causes existing functionality not to work as expected)
- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

Changes on admin_getpeers are not backwards compatible

- [x] Yes
- [ ] No

https://docs.nethermind.io/interacting/json-rpc-ns/admin/#admin_peers

### DIFF
--- a/tools/Kute/Nethermind.Tools.Kute.Test/MetricsTests.cs
+++ b/tools/Kute/Nethermind.Tools.Kute.Test/MetricsTests.cs
@@ -11,9 +11,9 @@ namespace Nethermind.Tools.Kute.Test;
 
 public class MetricsTests
 {
-    private static JsonRpc.Request.Single Single(int id)
+    private static JsonRpc.Request.Single Single(int id, string method = "test")
     {
-        var json = $$"""{ "id": {{id}}, "method": "test", "params": [] }""";
+        var json = $$"""{ "id": {{id}}, "method": "{{method}}", "params": [] }""";
         return new JsonRpc.Request.Single(JsonNode.Parse(json)!);
     }
 
@@ -31,7 +31,7 @@ public class MetricsTests
         var totalTimer = new Timer();
         using (totalTimer.Time())
         {
-            var single = Single(42);
+            var single = Single(42, "method1");
             var batch = Batch(Single(43), Single(44), Single(45));
 
             var singleTimer = new Timer();
@@ -57,7 +57,9 @@ public class MetricsTests
         report.TotalTime.Should().BeLessThan(TimeSpan.FromMilliseconds(110));
 
         report.Singles.Should().HaveCount(1);
-        report.Singles.Should().ContainKey("42");
+        report.Singles.Should().ContainKey("method1");
+        report.Singles["method1"].Should().HaveCount(1);
+        report.Singles["method1"].Should().ContainKey("42");
 
         report.Batches.Should().HaveCount(1);
         report.Batches.Should().ContainKey("43:45");
--- a/tools/Kute/Nethermind.Tools.Kute/Config.cs
+++ b/tools/Kute/Nethermind.Tools.Kute/Config.cs
@@ -77,8 +77,43 @@ public static class Config
         HelpName = "url",
     };
 
+    public static Option<string?> PrometheusPushGatewayUser { get; } = new("--gateway-user")
+    {
+        Description = "Prometheus Push Gateway username",
+        HelpName = "username",
+    };
+
+    public static Option<string?> PrometheusPushGatewayPassword { get; } = new("--gateway-pass")
+    {
+        Description = "Prometheus Push Gateway password",
+        HelpName = "password",
+    };
+
     public static Option<bool> UnwrapBatch { get; } = new("--unwrapBatch", "-u")
     {
         Description = "Batch requests will be unwraped to single requests"
+    };
+
+    public static Option<Dictionary<string, string>> Labels { get; } = new("--labels", "-l")
+    {
+        DefaultValueFactory = r => new Dictionary<string, string>(),
+        CustomParser = r =>
+        {
+            var labels = new Dictionary<string, string>();
+            foreach (var token in r.Tokens)
+            {
+                foreach (var pair in token.Value.Split(','))
+                {
+                    var parts = pair.Split('=', 2);
+                    if (parts.Length == 2)
+                    {
+                        labels.Add(parts[0], parts[1]);
+                    }
+                }
+            }
+            return labels;
+        },
+        Description = "A comma separated list of tags to be added to the metrics in the format key=value",
+        HelpName = "labels",
     };
 }
--- a/tools/Kute/Nethermind.Tools.Kute/Metrics/MemoryMetricsReporter.cs
+++ b/tools/Kute/Nethermind.Tools.Kute/Metrics/MemoryMetricsReporter.cs
@@ -9,7 +9,7 @@ public sealed class MemoryMetricsReporter
     : IMetricsReporter
     , IMetricsReportProvider
 {
-    private readonly ConcurrentDictionary<string, TimeSpan> _singles = new();
+    private readonly ConcurrentDictionary<string, ConcurrentDictionary<string, TimeSpan>> _singles = new();
     private readonly ConcurrentDictionary<string, TimeSpan> _batches = new();
 
     private TimeSpan _totalRunningTime;
@@ -36,10 +36,19 @@ public sealed class MemoryMetricsReporter
     }
     public Task Single(JsonRpc.Request.Single single, TimeSpan elapsed, CancellationToken token = default)
     {
-        var id = single.Id?.ToString();
-        if (id is not null)
+        var id = single.Id;
+        var methodName = single.MethodName;
+        if (id is not null && methodName is not null)
         {
-            _singles[id] = elapsed;
+            var newMethodDict = new ConcurrentDictionary<string, TimeSpan>();
+            newMethodDict[id] = elapsed;
+            _singles.AddOrUpdate(methodName,
+            newMethodDict,
+            (_, dict) =>
+            {
+                dict[id] = elapsed;
+                return dict;
+            });
         }
 
         return Task.CompletedTask;
@@ -60,7 +69,7 @@ public sealed class MemoryMetricsReporter
             Ignored = _ignored,
             Responses = _responses,
             TotalTime = _totalRunningTime,
-            Singles = _singles.ToDictionary(kvp => kvp.Key, kvp => kvp.Value),
+            Singles = _singles.ToDictionary(kvp => kvp.Key, kvp => (IReadOnlyDictionary<string, TimeSpan>)kvp.Value.ToDictionary(ikvp => ikvp.Key, ikvp => ikvp.Value)),
             Batches = _batches.ToDictionary(kvp => kvp.Key, kvp => kvp.Value),
         };
     }
--- a/tools/Kute/Nethermind.Tools.Kute/Metrics/MetricsReport.cs
+++ b/tools/Kute/Nethermind.Tools.Kute/Metrics/MetricsReport.cs
@@ -54,13 +54,13 @@ public sealed record MetricsReport
     public required long Ignored { get; init; }
     public required long Responses { get; init; }
     public required TimeSpan TotalTime { get; init; }
-    public required IReadOnlyDictionary<string, TimeSpan> Singles { get; init; }
+    public required IReadOnlyDictionary<string, IReadOnlyDictionary<string, TimeSpan>> Singles { get; init; }
     public required IReadOnlyDictionary<string, TimeSpan> Batches { get; init; }
 
     // Computed properties
-    private TimeMetrics? _singlesMetrics;
+    private Dictionary<string, TimeMetrics>? _singlesMetrics;
     private TimeMetrics? _batchesMetrics;
-    public TimeMetrics SinglesMetrics => _singlesMetrics ??= TimeMetrics.From(Singles.Values.ToList());
+    public Dictionary<string, TimeMetrics> SinglesMetrics => _singlesMetrics ??= Singles.ToDictionary(kvp => kvp.Key, kvp => TimeMetrics.From(kvp.Value.Values.ToList()));
     public TimeMetrics BatchesMetrics => _batchesMetrics ??= TimeMetrics.From(Batches.Values.ToList());
 }
 
--- a/tools/Kute/Nethermind.Tools.Kute/Metrics/PrettyMetricsReportFormatter.cs
+++ b/tools/Kute/Nethermind.Tools.Kute/Metrics/PrettyMetricsReportFormatter.cs
@@ -16,11 +16,18 @@ public sealed class PrettyMetricsReportFormatter : IMetricsReportFormatter
         await writer.WriteLineAsync($"Ignored: {report.Ignored}", token);
         await writer.WriteLineAsync($"Responses: {report.Responses}\n", token);
         await writer.WriteLineAsync("Singles:", token);
-        await writer.WriteLineAsync($"  Count: {report.Singles.Count}", token);
-        await writer.WriteLineAsync($"  Max: {report.SinglesMetrics.Max.TotalMilliseconds} ms", token);
-        await writer.WriteLineAsync($"  Average: {report.SinglesMetrics.Average.TotalMilliseconds} ms", token);
-        await writer.WriteLineAsync($"  Min: {report.SinglesMetrics.Min.TotalMilliseconds} ms", token);
-        await writer.WriteLineAsync($"  Stddev: {report.SinglesMetrics.StandardDeviation.TotalMilliseconds} ms", token);
+        foreach (var single in report.SinglesMetrics)
+        {
+            var methodName = single.Key;
+            var metrics = single.Value;
+            await writer.WriteLineAsync($"  {methodName}:", token);
+            await writer.WriteLineAsync($"    Count: {report.Singles[methodName].Count}", token);
+            await writer.WriteLineAsync($"    Max: {metrics.Max.TotalMilliseconds} ms", token);
+            await writer.WriteLineAsync($"    Average: {metrics.Average.TotalMilliseconds} ms", token);
+            await writer.WriteLineAsync($"    Min: {metrics.Min.TotalMilliseconds} ms", token);
+            await writer.WriteLineAsync($"    Stddev: {metrics.StandardDeviation.TotalMilliseconds} ms", token);
+        }
+
         await writer.WriteLineAsync("Batches:", token);
         await writer.WriteLineAsync($"  Count: {report.Batches.Count}", token);
         await writer.WriteLineAsync($"  Max: {report.BatchesMetrics.Max.TotalMilliseconds} ms", token);
--- a/tools/Kute/Nethermind.Tools.Kute/Metrics/PrometheusPushGatewayMetricsReporter.cs
+++ b/tools/Kute/Nethermind.Tools.Kute/Metrics/PrometheusPushGatewayMetricsReporter.cs
@@ -1,6 +1,8 @@
 // SPDX-FileCopyrightText: 2023 Demerzel Solutions Limited
 // SPDX-License-Identifier: LGPL-3.0-only
 
+using System.Net.Http.Headers;
+using System.Text;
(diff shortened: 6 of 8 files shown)


svlachakis:
After discussing with @damian-orzechowski , it's ok to introduce breaking changes and break backwards compatibility, converting the PR to draft.

src/Nethermind/Nethermind.JsonRpc.Test/Modules/AdminModuleTests.cs
```diff
@@ -552,4 +552,396 @@ public void MsgDelivered_after_session_closed()
 
         manualResetEvent.WaitOne(TimeSpan.FromMilliseconds(1000)).Should().Be(false);
     }
+
+    [Test]
+    public async Task Test_peers_new_format()
```
LukaszRozmej: Tests are messy and hard to read. Very long setup and mocks, a lot of detailed assertions.

For assertions - maybe better to serialize and compare the whole thing at once?

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/Models/EthProtocolDetails.cs
```diff
@@ -0,0 +1,25 @@
...
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System.Text.Json.Serialization;
+
+namespace Nethermind.JsonRpc.Modules.Admin.Models
+{
+    public class EthProtocolDetails
+    {
+        [JsonPropertyName("version")]
+        public int Version { get; set; }
```
LukaszRozmej: Default .net json serialization policy is changing from PascalCase to camelCase, so all those `JsonPropertyName` are not really needed.

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/Utils/NetworkConstants.cs
```diff
@@ -0,0 +1,12 @@
+// SPDX-FileCopyrightText: 2022 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+namespace Nethermind.JsonRpc.Modules.Admin.Utils
+{
+    public static class NetworkConstants
```
LukaszRozmej: There is already `Protocol` class with those constants

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/PeerInfo.cs
```diff
@@ -2,60 +2,96 @@
...
+        public string Id { get; set; } = string.Empty;
+
+        [JsonPropertyName("name")]
+        public string? Name { get; set; }
+
+        [JsonPropertyName("caps")]
+        public string[] Caps { get; set; } = Array.Empty<string>();
+
+        [JsonPropertyName("enr")]
+        public string? Enr { get; set; }
```
LukaszRozmej: I would keep as many of those as domain objects and not strings
Then create for them JsonConverters

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/Utils/CapabilityExtractor.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            if (session?.TryGetProtocolHandler(NetworkConstants.P2PPrefix, out IProtocolHandler? handler) == true &&
+                handler is IP2PProtocolHandler p2pHandler)
+            {
+                return p2pHandler.GetCapabilitiesForAdmin() ?? Enumerable.Empty<string>();
+            }
+
+            return Enumerable.Empty<string>();
+        }
+    }
```
LukaszRozmej: JsonConverter would be better, faster and can avoid allocations

LukaszRozmej (changes requested):

src/Nethermind/Nethermind.JsonRpc.Test/Modules/AdminModuleTests.cs
svlachakis: I think it's better and cleaner now, time is 2 times decreased

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/Models/EthProtocolDetails.cs
svlachakis: Makes sense, deleted

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/Utils/NetworkConstants.cs
svlachakis: Ah didn't saw them, used them thanks!

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/PeerInfo.cs
svlachakis: Agreed & done, refactored it.

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/Utils/CapabilityExtractor.cs
svlachakis: Done

src/Nethermind/Nethermind.Network/P2P/ProtocolHandlers/P2PProtocolHandler.cs
```diff
@@ -346,4 +346,13 @@ public override void Dispose()
         ProtocolInitialized = null;
         SubprotocolRequested = null;
     }
+
+    public IEnumerable<Capability> GetCapabilitiesForAdmin()
```
MarekM25: Cosmetic Rename to GetCapabilities

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/PeerInfo.cs
```diff
@@ -1,61 +1,153 @@
...
+                else if (capability.ProtocolCode == Protocol.Snap && snapVersion == 0)
+                {
+                    snapVersion = capability.Version;
+                }
+
+                if (ethVersion > 0 && snapVersion > 0) break;
+            }
+
+            // ETH protocol (always present)
+            protocols[Protocol.Eth] = new EthProtocolDetails { Version = ethVersion };
```
MarekM25: Something is wrong here, Other fields than Version will be always null in EthProtocolDetails

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/Models/EthProtocolDetails.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Text.Json.Serialization;
+
+namespace Nethermind.JsonRpc.Modules.Admin.Models
+{
+    public class EthProtocolDetails
+    {
+        public int Version { get; set; }
+
+        [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+        public string? EarliestBlock { get; set; }
```
MarekM25: are you sure those fileds are needed? it seems you don't use them

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/PeerInfo.cs
```diff
@@ -1,61 +1,153 @@
...
-            Id = peer.Node.Id.Hash.ToString(false);
-            Host = peer.Node.Host is null ? null : IPAddress.Parse(peer.Node.Host).MapToIPv4().ToString();
-            Port = peer.Node.Port;
-            Address = peer.Node.Address.ToString();
-            IsBootnode = peer.Node.IsBootnode;
-            IsStatic = peer.Node.IsStatic;
             Enode = peer.Node.ToString(Node.Format.ENode);
-            Inbound = peer.InSession is not null;
+            Caps = capabilities;
+            Enr = EnrExtractor.GetEnrFromPeer(peer);
```
MarekM25: Why complicate with extra class and return null. Just set null with the proper comment here. 

Btw I doubt it should be always null.

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/PeerInfo.cs
```diff
@@ -1,61 +1,153 @@
...
 
         public PeerInfo()
         {
         }
 
-        public PeerInfo(Peer peer, bool includeDetails)
+        public PeerInfo(Peer peer, bool includeDetails = false)
         {
-            if (peer.Node is null)
+            PeerValidator.ValidatePeer(peer);
```
MarekM25: I think you can just move methods to PeerInfo and remove the extra class PeerValidator. Generally, I see too many classes

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/PeerInfo.cs
MarekM25: Talk with @flcl42 about ENR

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/Utils/NetworkConstants.cs
MarekM25: so please remove them

MarekM25 (changes requested):

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/Utils/NetworkConstants.cs
svlachakis: done removed

src/Nethermind/Nethermind.Network/P2P/ProtocolHandlers/P2PProtocolHandler.cs
svlachakis: done

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/PeerInfo.cs
svlachakis: Fixed only version is needed.

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/Models/EthProtocolDetails.cs
svlachakis: deleted completely and simplyfied it

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/PeerInfo.cs
svlachakis: We need to implement it, as discussed with @LukaszRozmej we need to implement but most probably not now on this PR and add it later. I'll talk with @flcl42.

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/PeerInfo.cs
svlachakis: Makes sense, done

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/Converters/CapabilityConverter.cs
```diff
@@ -0,0 +1,119 @@
...
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Nethermind.Stats.Model;
+
+namespace Nethermind.JsonRpc.Modules.Admin.Converters
+{
+    public class CapabilityConverter : JsonConverter<IReadOnlyList<Capability>>
```
LukaszRozmej: No need to create converters for collections, those will be handled automatically. `JsonConverter<Capability>` is enough.

src/Nethermind/Nethermind.JsonRpc/Modules/Subscribe/PeerAddDropResponse.cs
```diff
@@ -1,27 +1,26 @@
...
+        Local = peerInfo.Network.LocalHost;
+        Remote = peerInfo.Network.RemoteAddress;
+        Error = error;
     }
 
     [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
     public string Type { get; set; }
 
-    public string Peer { get; set; }
+    [JsonConverter(typeof(PublicKeyHashedConverter))]
```
LukaszRozmej, line 22: Maybe that converter should be the default for public keys?

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/Converters/CapabilityConverter.cs
svlachakis: ok, changed it

src/Nethermind/Nethermind.JsonRpc/Modules/Subscribe/PeerAddDropResponse.cs
svlachakis: - PublicKeyConverter -> Writes the raw public key bytes as a hex string (existing)
- PublicKeyHashedConverter -> Writes the hash of the public key as a string (new)

I'm hesitant to change the default PublicKeyConverter in all those places to be honest, most probably it's going to break something.

LukaszRozmej: Can you check where default would be used?

src/Nethermind/Nethermind.Serialization.Json/EthereumJsonSerializer.cs
```diff
@@ -93,6 +93,8 @@ private static JsonSerializerOptions CreateOptions(bool indented, IEnumerable<Js
                     new BigIntegerConverter(),
                     new NullableBigIntegerConverter(),
                     new JavaScriptObjectConverter(),
+                    new PublicKeyConverter(),
+                    new PublicKeyHashedConverter(),
```
benaadams, line 97: Need to also add `new CapabilityConverter()`

benaadams: Or annotate Capability

svlachakis: @benaadams should be fine now, please check

src/Nethermind/Nethermind.JsonRpc/Modules/Subscribe/PeerAddDropResponse.cs
svlachakis: @LukaszRozmej some places that I see existing PublicKeyConverter being most probably used for serialization/deserialization are the following and there are for sure more.

```
Node.cs 
BidTrace.cs
ParityTransaction.cs
```

LukaszRozmej: I don't think `Node` is used in JSON RPC? 
The other two are quite specialised and they would suit better as the edge case with explicitly provided converter. But I am not even sure if they should be converted this way.

src/Nethermind/Nethermind.Network.Stats/Model/CapabilityConverter.cs
```diff
@@ -0,0 +1,90 @@
...
+                return;
+            }
+
+            WriteCapability(writer, value);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static void WriteCapability(Utf8JsonWriter writer, Capability capability)
+        {
+            writer.WriteStringValue($"{capability.ProtocolCode}/{capability.Version}");
```
LukaszRozmej: can you write it without constructing additional strings?

src/Nethermind/Nethermind.JsonRpc/Modules/Subscribe/PeerAddDropResponse.cs
svlachakis: I see, how do you think we should move on?

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/Utils/NetworkInfoBuilder.cs
```diff
@@ -0,0 +1,73 @@
...
+        private static string GetRemoteAddress(ISession? session, Node node)
+        {
+            if (session != null)
+            {
+                return $"{session.RemoteHost}:{session.RemotePort}";
+            }
+
+            return node.Address.ToString();
+        }
+    }
```
LukaszRozmej: Still lot of unnecessary string manipulation?

src/Nethermind/Nethermind.Network.Stats/Model/CapabilityConverter.cs
svlachakis: I've implemented zero-allocation string building using safe managed APIs (stackalloc, ArrayPool, Utf8Formatter). I think this eliminates the string interpolation.

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/Utils/NetworkInfoBuilder.cs
svlachakis: ~~For NetworkInfoBuilder, I'm finding it more challenging to eliminate the string interpolations cleanly while~~ ~~staying within safe code boundaries. The API constraints with IPAddress.TryFormat() and string.Create() not~~~~accepting ref structs make it difficult to avoid some allocations without moving to unsafe code.~~

~~Would you prefer I keep NetworkInfoBuilder readable as-is with the minimal allocations, or would you like~~ ~~me to explore an unsafe implementation for true zero-allocation? I'm happy to go either direction.~~

~~I appreciate the guidance on these optimizations!~~

@LukaszRozmej I think I got what you meant and fixed it.

src/Nethermind/Nethermind.JsonRpc/Modules/Subscribe/PeerAddDropResponse.cs
LukaszRozmej: I think we could reverse the default one, and specify old one in those places

LukaszRozmej (approved):

svlachakis:
@LukaszRozmej should we create a different task for the default public key converter? we need a new one also for ENR implementation.

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/PeerInfo.cs
MarekM25: you haven't applied this comment, please read one more time

svlachakis: @MarekM25 there is a new task added for this here https://github.com/NethermindEth/nethermind/issues/9226

Sorry you're right had not removed the extra class and added the TODO directly here - Done and added the issue link to the TODO comment.

PR #9008, opened 2025-07-18
Optimize ModExp

benaadams:
## Changes

- Do more maths in 64/32 bit space than 256 bit
- Vectorize overflow checks
- Don't allocate input arrays unless completely needed
- Don't throw and catch exception for overflow
- Don't try catch in `CalculateIterationCount`

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Evm.Precompiles/ModExpPrecompile.cs
+++ b/src/Nethermind/Nethermind.Evm.Precompiles/ModExpPrecompile.cs
@@ -2,19 +2,22 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 
 using System;
+using System.Buffers.Binary;
+using System.Diagnostics;
 using System.Numerics;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Runtime.Intrinsics;
 using Nethermind.Core;
 using Nethermind.Core.Extensions;
 using Nethermind.Core.Specs;
-using Nethermind.Evm;
-using Nethermind.Evm.Precompiles;
 using Nethermind.GmpBindings;
 using Nethermind.Int256;
 
 namespace Nethermind.Evm.Precompiles;
 
 /// <summary>
-///     https://github.com/ethereum/EIPs/blob/vbuterin-patch-2/EIPS/bigint_modexp.md
+/// https://github.com/ethereum/EIPs/blob/vbuterin-patch-2/EIPS/bigint_modexp.md
 /// </summary>
 public class ModExpPrecompile : IPrecompile<ModExpPrecompile>
 {
@@ -25,6 +28,11 @@ public class ModExpPrecompile : IPrecompile<ModExpPrecompile>
     /// For more details, see: https://eips.ethereum.org/EIPS/eip-7823
     /// </summary>
     public const int ModExpMaxInputSizeEip7823 = 1024;
+
+    private const int LengthSize = 32;
+    private const int StartBaseLength = 0;
+    private const int StartExpLength = 32;
+    private const int StartModLength = 64;
 
     private ModExpPrecompile()
     {
@@ -57,9 +65,10 @@ public class ModExpPrecompile : IPrecompile<ModExpPrecompile>
 
         try
         {
-            return inputData.Length >= 96
-                ? DataGasCostInternal(inputData.Span.Slice(0, 96), inputData, releaseSpec)
-                : DataGasCostInternal(inputData, releaseSpec);
+            ReadOnlySpan<byte> span = inputData.Span;
+            return span.Length >= 96
+                ? DataGasCostInternal(span, releaseSpec)
+                : DataGasCostShortInternal(span, releaseSpec);
         }
         catch (OverflowException)
         {
@@ -67,32 +76,32 @@ public class ModExpPrecompile : IPrecompile<ModExpPrecompile>
         }
     }
 
-    private static long DataGasCostInternal(ReadOnlyMemory<byte> inputData, IReleaseSpec releaseSpec)
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    private static long DataGasCostShortInternal(ReadOnlySpan<byte> inputData, IReleaseSpec releaseSpec)
     {
-        Span<byte> extendedInput = stackalloc byte[96];
-        inputData[..Math.Min(96, inputData.Length)].Span
-            .CopyTo(extendedInput[..Math.Min(96, inputData.Length)]);
+        Debug.Assert(inputData.Length < 96);
 
-        return DataGasCostInternal(extendedInput, inputData, releaseSpec);
+        Span<byte> extendedInput = stackalloc byte[96];
+        inputData.CopyTo(extendedInput);
+
+        return DataGasCostInternal(extendedInput, releaseSpec);
     }
 
-    private static long DataGasCostInternal(ReadOnlySpan<byte> extendedInput, ReadOnlyMemory<byte> inputData, IReleaseSpec releaseSpec)
+    private static long DataGasCostInternal(ReadOnlySpan<byte> inputData, IReleaseSpec releaseSpec)
     {
-        UInt256 baseLength = new(extendedInput[..32], true);
-        UInt256 expLength = new(extendedInput.Slice(32, 32), true);
-        UInt256 modulusLength = new(extendedInput.Slice(64, 32), true);
-
+        (uint baseLength, uint expLength, uint modulusLength) = GetInputLengths(inputData);
         if (ExceedsMaxInputSize(releaseSpec, baseLength, expLength, modulusLength))
         {
             return long.MaxValue;
         }
 
-        UInt256 complexity = MultComplexity(baseLength, modulusLength, releaseSpec.IsEip7883Enabled);
+        ulong complexity = MultComplexity(baseLength, modulusLength, releaseSpec.IsEip7883Enabled);
 
-        UInt256 expLengthUpTo32 = UInt256.Min(32, expLength);
-        UInt256 startIndex = 96 + baseLength; //+ expLength - expLengthUpTo32; // Geth takes head here, why?
-        UInt256 exp = new(inputData.Span.SliceWithZeroPaddingEmptyOnError((int)startIndex, (int)expLengthUpTo32), true);
+        uint expLengthUpTo32 = Math.Min(32, expLength);
+        uint startIndex = 96 + baseLength; //+ expLength - expLengthUpTo32; // Geth takes head here, why?
+        UInt256 exp = new(inputData.SliceWithZeroPaddingEmptyOnError((int)startIndex, (int)expLengthUpTo32), true);
         UInt256 iterationCount = CalculateIterationCount(expLength, exp, releaseSpec.IsEip7883Enabled);
+
         bool overflow = UInt256.MultiplyOverflow(complexity, iterationCount, out UInt256 result);
         result /= 3;
         return result > long.MaxValue || overflow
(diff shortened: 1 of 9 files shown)


src/Nethermind/Nethermind.Evm.Precompiles/ModExpPrecompile.cs
```diff
@@ -133,9 +240,10 @@ public unsafe (byte[], bool) Run(ReadOnlyMemory<byte> inputData, IReleaseSpec re
 
         using var modulusInt = mpz_t.Create();
 
-        fixed (byte* modulusData = inputData.Span.SliceWithZeroPaddingEmptyOnError(96 + baseLength + expLength, modulusLength))
+        ReadOnlySpan<byte> modulusDataSpan = inputSpan.SliceWithZeroPaddingEmptyOnError(96 + (int)baseLength + (int)expLength, (int)modulusLength);
+        if (modulusDataSpan.Length > 0)
         {
-            if (modulusData is not null)
+            fixed (byte* modulusData = &MemoryMarshal.GetReference(modulusDataSpan))
```
rubo, line 246: What's the advantage of using `&MemoryMarshal` everywhere?

src/Nethermind/Nethermind.Evm.Precompiles/ModExpPrecompile.cs
```diff
@@ -146,23 +254,27 @@ public unsafe (byte[], bool) Run(ReadOnlyMemory<byte> inputData, IReleaseSpec re
...
+            fixed (byte* expData = &MemoryMarshal.GetReference(expDataSpan))
                 Gmp.mpz_import(expInt, (nuint)expLength, 1, 1, 1, nuint.Zero, (nint)expData);
         }
 
         Gmp.mpz_powm(powmResult, baseInt, expInt, modulusInt);
 
-        var powmResultLen = (int)(Gmp.mpz_sizeinbase(powmResult, 2) + 7) / 8;
-        var offset = modulusLength - powmResultLen;
-        byte[] result = new byte[modulusLength];
+        uint powmResultLen = (uint)(Gmp.mpz_sizeinbase(powmResult, 2) + 7) / 8;
```
rubo: Why `uint`?

src/Nethermind/Nethermind.Evm.Precompiles/ModExpPrecompile.cs
benaadams: We already checked lengths so know is not null; so it skips additional checks that `fixed` does to get ref (like returning null instead)

src/Nethermind/Nethermind.Evm.Precompiles/ModExpPrecompile.cs
benaadams: muduliseLeng is uint so less casting

rubo: But it still casts when dealing with `modulusLength` later.

benaadams: `uint -> nuint` (zero extend) is less expensive than `int -> nuint` (sign extend); as its essentially a null op on Intel (just use wider register `ecx => rex`; extending a singed register is more expensive as it needs to fill with `f` when top bit is set

src/Nethermind/Nethermind.Evm.Precompiles/ModExpPrecompile.cs
rubo: I don't see the advantage -- with or without `MemoryMarshal`, the pointer has the same value of 0 and null check passes. I'm genuinely trying to understand the benefit of getting refs with `MemoryMarshal`.

benaadams: Pinning a span goes via [GetPinnableReference](https://source.dot.net/System.Private.CoreLib/R/6cac9a7c32b1b719.html)

```csharp
public ref readonly T GetPinnableReference()
{
    // Ensure that the native code has just one forward branch that is predicted-not-taken.
    ref T ret = ref Unsafe.NullRef<T>();
    if (_length != 0) ret = ref _reference);
    return ref ret;
}
```

Whereas MemoryMarshal.GetReference just returns the ref

```csharp
public ref readonly T GetReference()
{
    return ref _reference;
}
```

src/Nethermind/Nethermind.Evm.Precompiles/ModExpPrecompile.cs
rubo:
```csharp
// You're doing `niunt` to `uint` here:
uint powmResultLen = (uint)(Gmp.mpz_sizeinbase(powmResult, 2) + 7) / 8;
// Then creating another uint subtracting from int
uint offset = modulusLength - powmResultLen;
// Then the uint offset is added to nint ptr. Isn't it more effective to cast once to int and have it that way everywhere?
```

benaadams: mpz_sizeinbase returns a `nuint` not a signed one; will update to make more clear

rubo: Right, so I convert to signed right away, as all others are also signed

benaadams: Lengths are all unsigned?

rubo: GMP prefers unsigned. In this case, we have all signed values to deal with the unsigned one of GMP. So I converted it to signed right away.

benaadams: I don't understand; everything is unsigned now other than the first input to `Gmp.mpz_export` which is `nint` but is also a cast from `byte*` so 🤷‍♂️

benaadams: Also is only changing size where signed to unsigned is messy (e.g. int to ulong or nuint); so to make that less expensive you'd cast via uint e.g. `(ulong)(uint)var` as then doesn't do sign extend

rubo: > everything is unsigned now

How's that? The unsigned value retrieved from `mpz_sizeinbase` was later used to calculate the _signed_ offset to be added to the _signed_ `ptr`. That offset is calculated by subtracting from a _signed_ modulusLength. Hence, why change things to unsigned if in the end we need it signed, and in the middle are signed values. You're changing it to unsigned just to convert it to signed later for `ptr`.

benaadams: modulusLength is unsigned; but ok https://github.com/NethermindEth/nethermind/pull/9008/commits/6fba82b0066c1f1033270c5e96d44e485007301d

rubo: Oh, you changed it?

benaadams: Yeah; and added an "OldLengths" that is still signed (and used for the Old BigInteger compare

benaadams: removed the casts not needed now; might be clearer :)

rubo: That old bigint stuff looks redundant. It's used in a test to compare against the actual one. But given the number of tests we have, we can remove it to keep the code simpler. WDYT?

benaadams: Maybe better to move it to the tests? 🤔

benaadams: Moved it to tests

rubo: Maybe. But I think it's just obsolete. We can do it in another PR, of course.

src/Nethermind/Nethermind.Evm.Precompiles/ModExpPrecompile.cs
```diff
@@ -57,74 +64,173 @@ public long DataGasCost(ReadOnlyMemory<byte> inputData, IReleaseSpec releaseSpec
...
     }
 
-    private static bool ExceedsMaxInputSize(IReleaseSpec releaseSpec, UInt256 baseLength, UInt256 expLength, UInt256 modulusLength)
-        => releaseSpec.IsEip7823Enabled &&
-            (baseLength > ModExpMaxInputSizeEip7823 || expLength > ModExpMaxInputSizeEip7823 || modulusLength > ModExpMaxInputSizeEip7823);
+    private static bool ExceedsMaxInputSize(IReleaseSpec releaseSpec, uint baseLength, uint expLength, uint modulusLength)
+    {
+        return releaseSpec.IsEip7823Enabled
+            ? baseLength > ModExpMaxInputSizeEip7823 || expLength > ModExpMaxInputSizeEip7823 || modulusLength > ModExpMaxInputSizeEip7823
+            : baseLength == int.MaxValue || modulusLength == int.MaxValue;
```
LukaszRozmej: Can we do it branchless?
```suggestion
        return releaseSpec.IsEip7823Enabled && (baseLength > ModExpMaxInputSizeEip7823 || expLength > ModExpMaxInputSizeEip7823 || modulusLength > ModExpMaxInputSizeEip7823)
            || !releaseSpec.IsEip7823Enabled && (baseLength == int.MaxValue || modulusLength == int.MaxValue);

```

src/Nethermind/Nethermind.Evm.Precompiles/ModExpPrecompile.cs
benaadams: Might still be handy as backup since we are changing the precompiles more often now

src/Nethermind/Nethermind.Evm.Precompiles/ModExpPrecompile.cs
```diff
@@ -57,74 +64,173 @@ public long DataGasCost(ReadOnlyMemory<byte> inputData, IReleaseSpec releaseSpec
...
+        return (0, uint.MaxValue, 0);
+    }
+
     public unsafe (byte[], bool) Run(ReadOnlyMemory<byte> inputData, IReleaseSpec releaseSpec)
     {
         Metrics.ModExpPrecompile++;
 
-        (int baseLength, int expLength, int modulusLength) = GetInputLengths(inputData);
+        ReadOnlySpan<byte> inputSpan = inputData.Span;
+        (uint baseLength, uint expLength, uint modulusLength) = inputSpan.Length >= 96
```
LukaszRozmej: 96 to a const?

src/Nethermind/Nethermind.Evm.Precompiles/ModExpPrecompile.cs
```diff
@@ -226,38 +326,30 @@ private static UInt256 MultComplexity(in UInt256 baseLength, in UInt256 modulusL
...
-    private static UInt256 CalculateIterationCount(UInt256 exponentLength, UInt256 exponent, bool isEip7883Enabled)
+    private static UInt256 CalculateIterationCount(uint exponentLength, UInt256 exponent, bool isEip7883Enabled)
     {
-        try
+        ulong iterationCount;
+        if (exponentLength <= 32)
         {
-            UInt256 iterationCount;
-            if (exponentLength <= 32)
+            iterationCount = exponent.IsZero ? 0 : (uint)(exponent.BitLen - 1);
```
LukaszRozmej:
```suggestion
            iterationCount = Math.Min(1, (ulong)exponent.BitLen) - 1
```

src/Nethermind/Nethermind.Evm.Precompiles/ModExpPrecompile.cs
```diff
@@ -226,38 +326,30 @@ private static UInt256 MultComplexity(in UInt256 baseLength, in UInt256 modulusL
...
             }
-
-            return UInt256.Max(iterationCount, UInt256.One);
-        }
-        catch (OverflowException)
-        {
-            return UInt256.MaxValue;
         }
+
+        return iterationCount > 1 ? new UInt256(iterationCount) : UInt256.One;
```
LukaszRozmej: seems unnecessary?

src/Nethermind/Nethermind.Evm.Precompiles/ModExpPrecompile.cs
benaadams: Is still branchy with lazy ors? `||`

src/Nethermind/Nethermind.Evm.Precompiles/ModExpPrecompile.cs
benaadams: Changed

src/Nethermind/Nethermind.Evm.Precompiles/ModExpPrecompile.cs
benaadams: Changed

src/Nethermind/Nethermind.Evm.Precompiles/ModExpPrecompile.cs
benaadams: Added

src/Nethermind/Nethermind.Evm.Precompiles/ModExpPrecompile.cs
benaadams: Reduced the branches a little

LukaszRozmej (approved):

src/Nethermind/Nethermind.Evm.Precompiles/ModExpPrecompile.cs
rubo: This makes perfect sense for `GetReference`. Just checked the source code.

It's more interesting for [`GetArrayDataReference`](https://github.com/dotnet/runtime/blob/main/src/coreclr/System.Private.CoreLib/src/System/Runtime/InteropServices/MemoryMarshal.CoreCLR.cs#L38):

```
.maxstack  1
.locals init (uint8& pinned V_0)
IL_0000:  ldc.i4.s   32
IL_0002:  newarr     [System.Runtime]System.Byte
IL_0007:  call       !!0& [System.Runtime]System.Runtime.InteropServices.MemoryMarshal::GetArrayDataReference<uint8>(!!0[])
IL_000c:  stloc.0
IL_000d:  ldloc.0
IL_000e:  conv.u
IL_000f:  conv.i8
IL_0010:  call       void [System.Console]System.Console::WriteLine(int64)
IL_0015:  ldc.i4.0
IL_0016:  conv.u
IL_0017:  stloc.0
IL_0018:  ret
```
and without it:
```
.maxstack  2
.locals init (uint8* V_0, uint8[] pinned V_1)
IL_0000:  ldc.i4.s   32
IL_0002:  newarr     [System.Runtime]System.Byte
IL_0007:  dup
IL_0008:  stloc.1
IL_0009:  brfalse.s  IL_0010
IL_000b:  ldloc.1
IL_000c:  ldlen
IL_000d:  conv.i4
IL_000e:  brtrue.s   IL_0015
IL_0010:  ldc.i4.0
IL_0011:  conv.u
IL_0012:  stloc.0
IL_0013:  br.s       IL_001e
IL_0015:  ldloc.1
IL_0016:  ldc.i4.0
IL_0017:  ldelema    [System.Runtime]System.Byte
IL_001c:  conv.u
IL_001d:  stloc.0
IL_001e:  ldloc.0
IL_001f:  conv.i8
IL_0020:  call       void [System.Console]System.Console::WriteLine(int64)
IL_0025:  ldnull
IL_0026:  stloc.1
IL_0027:  ret
```
JIT optimizations are so good, it's hard to spot any performance difference.

rubo (approved):

PR #9012, opened 2025-07-19
Let caches clear and code commit in background

benaadams:
## Changes

- As was always the plan let caches clear in background prior to next block
- Wait for caches to be clear at start of next block rather than end of current
- Attach code writing to this for async code commit
- Had to change a lot of tests to async

We currently wait for the prewarm caches/processing to clear before completing the block processing as an over abundance of caution when we introduced the feature. We also wait at start of next block.

Before

image

After

image

The sync-over-async `.GetAwaiter().GetResult()` will occasionally get hit at the end of block and it is quite harsh for performance; so just remove it (and will be taken care of before following block, when is more likely to be complete)

Clear caches:

image

Write code sync:

image

- [x] Optimization

#### Requires testing

- [X] Yes

- [x] Yes - Changed existing tests


benaadams:
Makes DI disposal unhappy :( 

```
Unhandled exception. System.ObjectDisposedException: Instances cannot be resolved and nested lifetimes cannot be created from this LifetimeScope as it (or one of its parent scopes) has already been disposed.
   at Autofac.Core.Lifetime.LifetimeScope.ThrowDisposedException()
   at Autofac.Core.Lifetime.LifetimeScope.InternalBeginLifetimeScope(Object tag, Action`1 configurationAction, Boolean isolatedScope)
   at Autofac.Core.Lifetime.LifetimeScope.BeginLifetimeScope(Object tag, Action`1 configurationAction)
   at Autofac.Core.Lifetime.LifetimeScope.BeginLifetimeScope(Action`1 configurationAction)
   at Nethermind.Consensus.Processing.AutoReadOnlyTxProcessingEnvFactory.CreateForWarmingUp(IWorldState worldStateToWarmUp) in /_/src/Nethermind/Nethermind.Consensus/Processing/AutoReadOnlyTxProcessingEnvFactory.cs:line 33
   at Nethermind.Consensus.Processing.BlockCachePreWarmer.ReadOnlyTxProcessingEnvPooledObjectPolicy.Create() in /_/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs:line 368
   at Microsoft.Extensions.ObjectPool.DefaultObjectPool`1.Get()
   at Nethermind.Consensus.Processing.BlockCachePreWarmer.BlockState..ctor(BlockStateSource src) in /_/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs:line 402
   at Nethermind.Consensus.Processing.BlockCachePreWarmer.BlockStateSource.InitThreadState() in /_/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs:line 383
   at Nethermind.Core.Threading.ParallelUnbalancedWork.InitProcessor`1.Data`1.Init() in /_/src/Nethermind/Nethermind.Core/Threading/ParallelUnbalancedWork.cs:line 337
   at Nethermind.Core.Threading.ParallelUnbalancedWork.InitProcessor`1.Execute() in /_/src/Nethermind/Nethermind.Core/Threading/ParallelUnbalancedWork.cs:line 297
Unhandled exception. System.ObjectDisposedException: Instances cannot be resolved and nested lifetimes cannot be created from this LifetimeScope as it (or one of its parent scopes) has already been disposed.
   at Autofac.Core.Lifetime.LifetimeScope.ThrowDisposedException()
   at Autofac.Core.Lifetime.LifetimeScope.InternalBeginLifetimeScope(Object tag, Action`1 configurationAction, Boolean isolatedScope)
   at Autofac.Core.Lifetime.LifetimeScope.BeginLifetimeScope(Object tag, Action`1 configurationAction)
   at Autofac.Core.Lifetime.LifetimeScope.BeginLifetimeScope(Action`1 configurationAction)
   at Nethermind.Consensus.Processing.AutoReadOnlyTxProcessingEnvFactory.CreateForWarmingUp(IWorldState worldStateToWarmUp) in /_/src/Nethermind/Nethermind.Consensus/Processing/AutoReadOnlyTxProcessingEnvFactory.cs:line 33
   at Nethermind.Consensus.Processing.BlockCachePreWarmer.ReadOnlyTxProcessingEnvPooledObjectPolicy.Create() in /_/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs:line 368
   at Microsoft.Extensions.ObjectPool.DefaultObjectPool`1.Get()
   at Nethermind.Consensus.Processing.BlockCachePreWarmer.BlockState..ctor(BlockStateSource src) in /_/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs:line 402
   at Nethermind.Consensus.Processing.BlockCachePreWarmer.BlockStateSource.InitThreadState() in /_/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs:line 383
   at Nethermind.Core.Threading.ParallelUnbalancedWork.InitProcessor`1.Data`1.Init() in /_/src/Nethermind/Nethermind.Core/Threading/ParallelUnbalancedWork.cs:line 337
   at Nethermind.Core.Threading.ParallelUnbalancedWork.InitProcessor`1.Execute() in /_/src/Nethermind/Nethermind.Core/Threading/ParallelUnbalancedWork.cs:line 297
```

src/Nethermind/Nethermind.Synchronization/ParallelSync/MultiSyncModeSelector.cs
```diff
@@ -108,6 +108,11 @@ private async Task StartAsync(CancellationToken cancellationToken)
                     {
                         Update();
                     }
+                    catch (ObjectDisposedException od) when (od.ObjectName == "Nethermind.Db.Rocks.DbOnTheRocks")
+                    {
```
MarekM25, line 112: Should we log this at least? It might be confusing if happens on production code

src/Nethermind/Nethermind.State/StateProvider.cs
```diff
@@ -506,14 +508,17 @@ public bool AddToBalanceAndCreateIfNotExists(Address address, in UInt256 balance
...
+
         public void Commit(IReleaseSpec releaseSpec, bool commitRoots, bool isGenesis)
             => Commit(releaseSpec, NullStateTracer.Instance, commitRoots, isGenesis);
 
         public void Commit(IReleaseSpec releaseSpec, IWorldStateTracer stateTracer, bool commitRoots, bool isGenesis)
         {
-            Task codeFlushTask = !commitRoots || _codeBatch is null || _codeBatch.Count == 0
-                ? Task.CompletedTask
-                : CommitCodeAsync();
+            if (commitRoots && _codeBatch?.Count > 0)
```
MarekM25, line 518: Quick Question: Can it cause consensus issues if a few commits are made before the previous async write has finished?

We're doing multiple commits during block processing, and we might encounter a situation where the cache hasn't been fully cleared or the previous commit hasn't completed, yet we begin processing a new block.

src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
```diff
@@ -138,6 +138,10 @@ private void WarmupWithdrawals(ParallelOptions parallelOptions, IReleaseSpec spe
                     });
             }
         }
+        catch (ObjectDisposedException)
+        {
+            // Ignore, disposed object pool policy (cancelled in test DI)
```
MarekM25, line 143: worth logging?

MarekM25 (commented):
We should monitor the performance of all background tasks. In some use cases, their performance might be more critical than that of non-background tasks

src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
LukaszRozmej: This should be only in test context? Or maybe better we can expose something that will be awaited in tests, so this doesn't happen

src/Nethermind/Nethermind.Core/Threading/ParallelUnbalancedWork.cs
```diff
@@ -43,19 +45,28 @@ public static void For(int fromInclusive, int toExclusive, ParallelOptions paral
 
         Data data = new(threads, fromInclusive, toExclusive, action, parallelOptions.CancellationToken);
 
-        for (int i = 0; i < threads - 1; i++)
+        // Queue work items to the thread pool for all threads except the current one
+        var tasks = new ParallelUnbalancedWork[threads - 1];
```
LukaszRozmej, line 49: ArrayPool?

src/Nethermind/Nethermind.Core/Threading/ParallelUnbalancedWork.cs
```diff
@@ -266,20 +285,28 @@ public static void For(
             var data = new Data<TLocal>(threads, fromInclusive, toExclusive, action, init, initValue, @finally, parallelOptions.CancellationToken);
 
             // Queue work items to the thread pool for all threads except the current one
-            for (int i = 0; i < threads - 1; i++)
+            var tasks = new InitProcessor<TLocal>[threads - 1];
```
LukaszRozmej, line 288: ArrayPool?

src/Nethermind/Nethermind.Synchronization/ParallelSync/MultiSyncModeSelector.cs
LukaszRozmej: Again, this is not right, either have a test mode, or better expose something test can await before teardown

smartprogrammer93 (approved):
Time to revive this. Looks good

src/Nethermind/Nethermind.State/StateProvider.cs
```diff
@@ -506,14 +508,17 @@ public bool AddToBalanceAndCreateIfNotExists(Address address, in UInt256 balance
...
+            {
+                CommitCodeAsync();
```
flcl42, line 520: **[HIGH] Keep rewritten contract code visible until its flush completes**

At an AuRa `rewriteBytecode` activation, `AuRaBlockProcessor` calls `Commit(commitRoots: true)` immediately after inserting replacement code and then executes the block's transactions. This call now exchanges `_codeBatch` and only schedules the database write, so a transaction calling the rewritten address can observe its new code hash while `GetCode` still finds no corresponding database entry, aborting a valid block nondeterministically. Keep the in-flight batch readable or wait for this flush before transaction execution, and test a rewrite block whose first transaction invokes the rewritten contract with a delayed code DB.

src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.cs
```diff
@@ -212,16 +205,28 @@ public Block[] Process(BlockHeader? baseBlock, IReadOnlyList<Block> suggestedBlo
...
 
+    private async Task ClearCaches(Task? preWarmTask)
+    {
+        if (preWarmTask is not null)
+        {
+            await preWarmTask;
+        }
+        await _stateProvider.WaitForCodeCommit();
+
+        _stateProvider.Reset();
```
flcl42, line 216: **[MEDIUM] Do not reset shared world state after Process returns**

For blocks without a prewarm task, `QueueClearCaches` runs this reset through `Task.Run`, allowing `Process` to return while cleanup still owns the same non-thread-safe world state. `SimulateBridgeHelper` processes payload blocks individually and immediately prepares the next block; if this reset lands after the next block's overrides or account/code changes but before their commit, it clears those changes or races the underlying dictionaries, producing an incorrect `eth_simulateV1` result. Keep the world-state reset synchronous or expose and await per-call cleanup before callers reuse the state, including before `InitBranch` on the next call.

src/Nethermind/Nethermind.Core/Threading/ParallelUnbalancedWork.cs
```diff
@@ -294,21 +321,31 @@ public static void For(
         /// </summary>
         public void Execute()
         {
-            TLocal? value = _data.Init();
             try
             {
-                int i = _data.Index.GetNext();
-                while (i < _data.ToExclusive)
+                TLocal? value = _data.Init();
```
flcl42, line 326: **[MEDIUM] Always signal worker completion when initialization fails**

If `_data.Init()` throws, control reaches the outer catch without executing the inner `finally`, so this worker never calls `MarkThreadCompleted`; a throwing `_data.Finally` has the same problem because the mark follows it. `For` then waits forever on `data.Event`, even with `MaxDegreeOfParallelism = 1`. The cache prewarmer's `InitThreadState` can throw while obtaining or building a read-only scope, which would hang cache cleanup, subsequent block processing, and shutdown. Put `MarkThreadCompleted` in an outer finally guaranteed to run and add initializer/finalizer exception tests.

src/Nethermind/Nethermind.State/StateProvider.cs
```diff
@@ -669,6 +672,10 @@ Task CommitCodeAsync()
                         _codeBatchAlternate = _codeBatch.GetAlternateLookup<ValueHash256>();
                     }
                 });
+
+                _codeFlushTask = _codeFlushTask.IsCompleted ?
```
flcl42, line 676: **[MEDIUM] Preserve failures from every queued code flush**

When two root commits enqueue code batches before the caller waits—for example, an AuRa rewrite followed by contract creation—this continuation runs even if the previous flush faulted and exposes only `flushTask`; if the previous task already faulted, the `IsCompleted` branch replaces it directly. `WaitForCodeCommit` can therefore succeed after the later batch while earlier code was never persisted, leaving committed account hashes pointing to missing code. Combine the tasks with failure propagation, such as `Task.WhenAll`, and test a failing first batch followed by a successful batch.

src/Nethermind/Nethermind.State/StateProvider.cs
```diff
@@ -643,15 +648,13 @@ public void Commit(IReleaseSpec releaseSpec, IWorldStateTracer stateTracer, bool
...
-            Task CommitCodeAsync()
+            void CommitCodeAsync()
             {
                 Dictionary<Hash256AsKey, byte[]> dict = Interlocked.Exchange(ref _codeBatch, null);
-                if (dict is null) return Task.CompletedTask;
+                if (dict is null) return;
                 _codeBatchAlternate = default;
 
-                return Task.Run(() =>
+                Task flushTask = Task.Run(() =>
```
flcl42, line 657: **[HIGH] Committed contract code is temporarily unreadable**

`CommitCodeAsync` exchanges the pending dictionary out of `_codeBatch`, starts this worker, and lets `Commit` return before the database write completes. Until the task runs, `GetCodeCore` can see neither that dictionary nor the database value. The unconditional EVM warm-up inserts a contract, calls `state.Commit(spec)`, and immediately executes it, so a delayed worker makes node initialization throw `InvalidOperationException`; AuRa similarly commits rewritten bytecode immediately before executing the block, allowing the same timing window to reject a valid rewrite block.

src/Nethermind/Nethermind.State/StateProvider.cs
```diff
@@ -669,6 +672,10 @@ Task CommitCodeAsync()
                         _codeBatchAlternate = _codeBatch.GetAlternateLookup<ValueHash256>();
```
flcl42, line 672: **[MEDIUM] Recycled code batches are published before their lookup is initialized**

The flush worker first installs its cleared dictionary into `_codeBatch` and only initializes `_codeBatchAlternate` on this following line. Because `Commit` no longer waits for the worker, a concurrent next `InsertCode` can observe the non-null dictionary in that interval, skip the initialization branch, and index a default alternate lookup. Pre-EIP-658 transaction processing commits roots after each transaction, so consecutive contract deployments during historical sync can hit the window and abort a valid block with a null-reference exception.

src/Nethermind/Nethermind.State/StateProvider.cs
```diff
@@ -669,6 +672,10 @@ Task CommitCodeAsync()
...
                     }
                 });
+
+                _codeFlushTask = _codeFlushTask.IsCompleted ?
+                    flushTask :
+                    _codeFlushTask.ContinueWith((t) => flushTask).Unwrap();
```
flcl42, line 678: **[MEDIUM] A later code flush masks failures from earlier batches**

The continuation ignores the antecedent task parameter, so if an earlier flush faults while a later `flushTask` succeeds, the composed `_codeFlushTask` completes successfully; the `IsCompleted` branch likewise replaces an already faulted task. Pre-EIP-658 blocks can schedule one code flush per transaction before `ClearCaches` calls `WaitForCodeCommit`. An I/O failure in an earlier contract-code batch can therefore be lost, allowing the block's state root to be accepted even though that contract's code is absent from the code database.

src/Nethermind/Nethermind.Core/Threading/ParallelUnbalancedWork.cs
flcl42, line 326: **[MEDIUM] Thread-local initialization failures leave a parallel worker active**

`_data.Init()` is now inside the outer catch, while `_data.MarkThreadCompleted()` remains in an inner `finally` reached only after initialization. If an initializer throws, the worker returns without decrementing `ActiveThreads`; `For` then waits on `data.Event`, which cannot be signaled. The block-cache prewarmer uses this overload with initializers that acquire a pooled environment and build a read-only scope, so an initialization failure there hangs cache clearing and consequently the next block or shutdown instead of reaching its existing exception handler.

src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.cs
flcl42, line 216: **[MEDIUM] Background cache clearing resets world state concurrently with the block-processing thread**

QueueClearCaches now schedules ClearCaches as a background _clearTask, and this added line performs the actual _stateProvider.Reset() on that background thread. The processing thread only waits for _clearTask in WaitForCacheClear at the start of Process (line 84) and of each loop iteration (line 100), but several state-touching paths run before or outside those waits while a previous _clearTask can still be in flight: InitBranch at Process entry (line 80) reads _stateProvider.StateRoot and calls _stateProvider.Reset()/SetBaseBlock before the first WaitForCacheClear; the commit-point InitBranch (line 164, every 64 blocks in a long branch) runs right after QueueClearCaches; RestoreBranch on the DoNotUpdateHead path (line 177) and in the exception handler (line 187) calls _stateProvider.Reset() immediately after QueueClearCaches(preWarmTask) queued the same reset in the background; and BlockProcessed subscribers (line 152) now observe state while the background reset may already be mutating it. StateProvider.Reset and PersistentStorageProvider.Reset clear plain Dictionary/List/HashSet instances (_blockChanges, _changes, _intraTxCache, _storages, _originalValues) with no synchronization, so concurrent execution can throw or leave torn collections, failing block processing or corrupting the change tracking of the next branch. Previously the prewarm task was awaited and _stateProvider.Reset() ran synchronously at the end of each loop iteration, so no such concurrency existed. Sequencing the reset so it cannot overlap InitBranch/RestoreBranch/BlockProcessed on the processing thread would restore the previous single-threaded invariant.

src/Nethermind/Nethermind.Synchronization/ParallelSync/MultiSyncModeSelector.cs
```diff
@@ -108,6 +108,11 @@ private async Task StartAsync(CancellationToken cancellationToken)
...
+                        // Calm down Nethermind.Blockchain.Test output (lots of Db disposed)
+                        break;
```
flcl42, line 114: **[LOW] Sync-mode selector loop exits permanently on a state-database disposal without logging**

The added handler calls break on ObjectDisposedException when ObjectName matches "Nethermind.Db.Rocks.DbOnTheRocks", which exits StartAsync's while (await timer.WaitForNextTickAsync(...)) loop. Every other exception from Update() is logged as "Sync mode update failed" and the loop continues on the next tick, but this path stops the periodic Update() evaluation silently, with no log entry, and it matches a hard-coded RocksDB type name string. If the state database is disposed outside node shutdown (for example during a full-prune database swap), the selector halts permanently instead of resuming once the database is replaced; previously this condition was logged and retried on the following tick. Keeping the loop alive with a log entry, or restarting the periodic evaluation, would retain the prior recovery behavior.

flcl42 (commented):
Found 6 issues (1 high, 4 medium, and 1 low) in `src/Nethermind/Nethermind.State/StateProvider.cs`, `src/Nethermind/Nethermind.Core/Threading/ParallelUnbalancedWork.cs`, `src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.cs`, and `src/Nethermind/Nethermind.Synchronization/ParallelSync/MultiSyncModeSelector.cs`.

PR #8980, opened 2025-07-13
Separate branch and block processor

asdacap:
- Require #8959 
- Split the branch logic out of `IBlockProcessor` into `IBranchProcessor`.
- This is so that the state handling (in `BranchProcessor`) is outside, which resolve some double world scoping.
- This PR only split the class without doing further modification.

#### What types of changes does your code introduce?

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

#### If yes, did you write tests?

- [X] Yes
- [ ] No

#### Notes on testing

- [x] Mainnet can sync normally

PR #9021, opened 2025-07-21
Remove use of blockchain processor in RPC

asdacap:
- Required #8980 
- RPC modules which uses IOverridableEnv have a well defined world scope that it execute within.
- Problem is, it then call `IBlockchainProcessor`, which then call `IBranchProcessor`, which then re-set the state, making it an unclear responsibility who is suppose to handle the state. 
- This PR try to address it by finding the remaining caller of `IBlockchainProcessor ` that already have IOverridableScope to call a new class `BlockchainProcessorFacade` which try to call `IBlockProcessor` directly without re-setting the state.
- Note, this PR try to keep existing behaviour, no matter how weird.

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- [ ] Mainnet can sync as always.
- [ ] I don't know I'll find a hive test somewhere.


src/Nethermind/Nethermind.JsonRpc/Modules/Trace/TraceRpcModule.cs
```diff
@@ -148,12 +148,18 @@ public ResultWrapper<ParityTxTraceFromReplay> trace_replayTransaction(Hash256 tx
             }
 
             Block block = blockSearch.Object!;
-            if (!stateReader.HasStateForBlock(block.Header))
+            SearchResult<BlockHeader> parentSearch = blockFinder.SearchForHeader(new BlockParameter(block.Header.ParentHash));
```
tanishqjasoria, line 151: is this because of the ForceSameBlock change?

tanishqjasoria (approved):

src/Nethermind/Nethermind.JsonRpc/Modules/Trace/TraceRpcModule.cs
asdacap: That one is at https://github.com/NethermindEth/nethermind/pull/9021/files#diff-8ededae48b6899e82e6e8d8d0bfba084ef969d0545d411eb2bb052a2e89c76edR175. 

Previously `BranchProcessor` set the state to the parent, its just not obvious from here.

tanishqjasoria: yeah, was referring to that

PR #9028, opened 2025-07-23
Scoped WorldState

asdacap:
- Require #9021 
- Replace `SetBaseBlock` with `BeginScope(BlockHeader baseBlock)` that returns a disposable.
- Any `IWorldState` can only be called in between starting and closing of the scope. 
- Most change is in test.
- Notable non-test code is in `WorldState` and `BranchProcessor`.
- The rest was isolated in previous PRs.

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

- [ ] Mainnet can sync normally
- [ ] Archive sync from genesis


src/Nethermind/Nethermind.State/OverridableEnv/OverridableEnvFactory.cs
```diff
@@ -57,7 +57,6 @@ public void Dispose()
 
         private void Reset()
         {
-            overridableScope.WorldState.Reset();
             codeInfoRepository.ResetOverrides();
 
             _worldScopeCloser?.Dispose();
```
asdacap, line 62: Closer call reset

src/Nethermind/Nethermind.Consensus/Processing/BranchProcessor.cs
```diff
@@ -51,46 +51,40 @@ public event EventHandler<TxProcessedEventArgs> TransactionProcessed
...
-        /* Please note that we do not reset the state if branch state root is null.
-           That said, I do not remember in what cases we receive null here.*/
-        if (baseBlock is not null && _stateProvider.StateRoot != baseBlock.StateRoot)
-        {
-            /* Discarding the other branch data - chain reorganization.
-               We cannot use cached values any more because they may have been written
-               by blocks that are being reorganized out.*/
-
-            if (incrementReorgMetric)
-                Metrics.Reorganizations++;
```
asdacap, line 65: Reorg metric removed from this codepath as as the worldstate does not store nor allow getting stateroot outsite of it's scope.

Also, doesn't this mean this get incremented on all RPC call before?

src/Nethermind/Nethermind.Consensus/Processing/BranchProcessor.cs
```diff
@@ -188,11 +184,6 @@ public Block[] Process(BlockHeader? baseBlock, IReadOnlyList<Block> suggestedBlo
                 TxHashCalculator.CalculateInBackground(suggestedBlock);
             }
 
-            if (options.ContainsFlag(ProcessingOptions.DoNotUpdateHead))
-            {
-                RestoreBranch(previousBranchStateRoot);
```
asdacap, line 192: Irrelevant, worldstate is always restored on scope exit.

asdacap:
I tried that about 1 year ago, @tanishqjasoria  tried that about half year ago. Result it A LOT of refactor. And we still dont have proper worldstate scope.

src/Nethermind/Nethermind.State/WorldState.cs
```diff
@@ -62,15 +68,23 @@ internal WorldState(
...
 
         public WorldState(ITrieStore trieStore, IKeyValueStoreWithBatching? codeDb, ILogManager? logManager, PreBlockCaches? preBlockCaches, bool populatePreBlockCache = true)
             : this(trieStore, codeDb, logManager, null, preBlockCaches: preBlockCaches, populatePreBlockCache: populatePreBlockCache)
         {
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private void GuardInScope()
+        {
+            if (!_isInScope) throw new InvalidOperationException($"{nameof(IWorldState)} must only be used within scope");
```
benaadams: Should be throw helper call to a `[StackTraceHidden, DoesNotReturn]` method or local function; rather than aggressively inlining a throw and interpolated string constructor

src/Nethermind/Ethereum.Test.Base/BlockchainTestBase.cs
```diff
@@ -140,32 +140,36 @@ protected async Task<EthereumTestResult> RunTest(BlockchainTest test, Stopwatch?
         IBlockchainProcessor blockchainProcessor = mainBlockProcessingContext.BlockchainProcessor;
         IBlockTree blockTree = container.Resolve<IBlockTree>();
         IBlockValidator blockValidator = container.Resolve<IBlockValidator>();
+        blockchainProcessor.Start();
 
-        InitializeTestState(test, stateProvider, specProvider);
+        // Genesis processing
+        using (stateProvider.BeginScope(null))
```
LukaszRozmej, line 146: Can you make some constant called `Genesis` for the parameter?

src/Nethermind/Nethermind.Consensus/Processing/AutoReadOnlyTxProcessingEnvFactory.cs
```diff
@@ -40,12 +40,12 @@ public IReadOnlyTxProcessorSource CreateForWarmingUp(IWorldState worldStateToWar
...
 
-    private class AutoReadOnlyTxProcessingEnv(ITransactionProcessor transactionProcessor, IWorldState worldState, ILifetimeScope lifetimeScope) : IReadOnlyTxProcessorSource, IDisposable
+    public class AutoReadOnlyTxProcessingEnv(ITransactionProcessor transactionProcessor, IWorldState worldState, ILifetimeScope lifetimeScope) : IReadOnlyTxProcessorSource, IDisposable
     {
         public IReadOnlyTxProcessingScope Build(BlockHeader? header)
         {
-            worldState.SetBaseBlock(header);
-            return new ReadOnlyTxProcessingScope(transactionProcessor, worldState);
+            IDisposable closer = worldState.BeginScope(header);
+            return new ReadOnlyTxProcessingScope(transactionProcessor, closer, worldState);
```
LukaszRozmej, line 48: `closer` is terrible name, maybe just call it `currentScope`?

LukaszRozmej (commented):
Would be better if `IWorldState` was the scope and it would have been created from something like `IWoldStateSource/Manager`

LukaszRozmej (approved):

LukaszRozmej:
> I tried that about 1 year ago, @tanishqjasoria tried that about half year ago. Result it A LOT of refactor. And we still dont have proper worldstate scope.

image

src/Nethermind/Nethermind.Init/PruningTrieStateFactory.cs
```diff
@@ -114,9 +114,6 @@ ILogManager logManager
             verifyTrieStarter!
         );
 
-        // Init state if we need system calls before actual processing starts
-        worldState.SetBaseBlock(blockTree.Head?.Header);
-
```
tanishqjasoria, line 119: why?

src/Nethermind/Nethermind.Consensus/Processing/AutoReadOnlyTxProcessingEnvFactory.cs
tanishqjasoria: +1

src/Nethermind/Nethermind.Consensus/Processing/BranchProcessor.cs
```diff
@@ -172,7 +166,9 @@ public Block[] Process(BlockHeader? baseBlock, IReadOnlyList<Block> suggestedBlo
                 {
                     if (_logger.IsInfo) _logger.Info($"Commit part of a long blocks branch {i}/{blocksCount}");
                     previousBranchStateRoot = suggestedBlock.Header;
-                    InitBranch(suggestedBlock.Header, false);
+
+                    worldStateCloser.Dispose();
```
tanishqjasoria, line 170: not relevant right now - but, this is good, we can add another check here, for something like IsTransitionBlock

tanishqjasoria (approved):

src/Nethermind/Nethermind.Init/PruningTrieStateFactory.cs
asdacap: I dont know too.

PR #9030, opened 2025-07-24
L1SLOAD Precompile for Surge

dipkakwani:
L1SLOAD precompile for surge. The precompile takes L1 contract address, storage key and block number as input and outputs the storage value.

## Changes
- Adds new `L1SloadPrecompile` at the address `0x10001`
-  To enable/disable: `IsRip7728Enabled` is added in the release spec and `Rip7728TransitionTimestamp` is added in the chain spec.
   - Example chainspec config to enable L1SLOAD precompile:
  ```
 "params": {
    ...
    "rip7728TransitionTimestamp": "0x0"
  },
```
- Precompile delegates fetching of the L1 storage value through `IL1StorageProvider`. 
   - Current implementation defines `JsonRpcL1StorageProvider` in Surge which uses `eth_getStorageAt` RPC to fetch the value from L1
   - For L1SLOAD in Surge,  `--Surge.L1EthApiEndpoint` should always be specified during NMC startup for `JsonRpcL1StorageProvider`. Note that the same property is also required by `SurgeGasPriceOracle`.

## Types of changes

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing
Also tested locally with live mainnet data:
```
./nethermind
--config none      
--datadir ./data/surge  
--Init.ChainSpecPath surge/chainspec_l1sload.json       
--Init.GenesisHash 0x820481ef44a092878372baee8ef36c481695a2d1e8e6ed3d2704727cb35a7d67       
--Init.DiscoveryEnabled false       
--Metrics.Enabled true       
--Metrics.ExposePort ${L2_METRICS_PORT}       
--JsonRpc.Enabled true       
--JsonRpc.EnabledModules [debug,eth,net,web3,txpool,rpc,subscribe,trace,personal,proof,parity,health]       
--JsonRpc.Host 0.0.0.0       
--JsonRpc.Port ${L2_HTTP_PORT}       
--JsonRpc.WebSocketsPort ${L2_WS_PORT}       
--JsonRpc.JwtSecretFile simple-surge-node/static/jwtsecret       
--JsonRpc.EngineHost 0.0.0.0       
--JsonRpc.EnginePort ${L2_ENGINE_API_PORT}       
--Network.DiscoveryPort ${L2_NETWORK_DISCOVERY_PORT}       
--Network.P2PPort ${L2_NETWORK_DISCOVERY_PORT}       
--Sync.FastSync false       
--Sync.SnapSync false       
--HealthChecks.Enabled true       
--log ${NETHERMIND_LOG_LEVEL}       
--Surge.L1EthApiEndpoint https://ethereum-rpc.publicnode.com 
--Surge.TaikoInboxAddress ${TAIKO_INBOX}


curl -X POST http://localhost:8547   -H "Content-Type: application/json"   -d '{
    "jsonrpc": "2.0",
    "method": "eth_call",
    "params": [{
      "to": "0x0000000000000000000000000000000000010001",
      "data": "0xE592427A0AEce92De3Edee1F18E0157C05861564000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000015F8880"
    }, "latest"],
    "id": 1
  }'
{"jsonrpc":"2.0","result":"0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff","id":1}
```

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### CODE REVIEW

File: src/Nethermind/Nethermind.Evm.Precompiles/L1SloadPrecompile.cs [hunk 1]
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        if (inputData.Length != L1PrecompileConstants.ExpectedInputLength)
+        {
+            return IPrecompile.Failure;
+        }
+
+        var contractAddress = new Address(inputData.Span[..L1PrecompileConstants.AddressBytes]);
+        var storageKey = new UInt256(inputData.Span[L1PrecompileConstants.AddressBytes..(L1PrecompileConstants.AddressBytes + L1PrecompileConstants.StorageKeyBytes)], isBigEndian: true);
+        var blockNumber = new UInt256(inputData.Span[(L1PrecompileConstants.AddressBytes + L1PrecompileConstants.StorageKeyBytes)..], isBigEndian: true);
```
dipkakwani: @smartprogrammer93 Should we support keywords in the block number for L1SLOAD (e.g. `latest`)? I would have to modify the code to use `BlockParameter` similar to `eth_getStorageAt` to support that.

smartprogrammer93: No supporting keywaords no

src/Nethermind/Nethermind.Core/L1PrecompileConstants.cs
```diff
@@ -0,0 +1,40 @@
...
+
+    /// <summary>
+    /// Fixed gas cost for L1SLOAD precompile calls
+    /// </summary>
+    public const long FixedGasCost = 2000L;
+
+    /// <summary>
+    /// Per-load gas cost for L1SLOAD precompile calls
+    /// </summary>
+    public const long PerLoadGasCost = 2000L;
```
smartprogrammer93, line 39: Where did these gas costs come from? meaning, how did you estimate these costs? or were they outlined in the RIP?

src/Nethermind/Nethermind.Evm.Precompiles/L1SloadPrecompile.cs
```diff
@@ -0,0 +1,100 @@
...
+/// </summary>
+public class L1SloadPrecompile : IPrecompile<L1SloadPrecompile>
+{
+    public static readonly L1SloadPrecompile Instance = new();
+
+    private L1SloadPrecompile()
+    {
+    }
+
+    public static Address Address { get; } = Address.FromNumber(0x12);
```
smartprogrammer93: For L2 precompiles, i would like us to use a different range
so lets start with `0x10001` this way, up to 0xffff is reserved for L1

src/Nethermind/Nethermind.Evm/Precompiles/AddressExtensions.cs
```diff
@@ -36,6 +36,7 @@ public static bool IsPrecompile(this Address address, IReleaseSpec releaseSpec)
                     0x0f => releaseSpec.Bls381Enabled,
                     0x10 => releaseSpec.Bls381Enabled,
                     0x11 => releaseSpec.Bls381Enabled,
+                    0x12 => releaseSpec.IsRip7728Enabled,
```
smartprogrammer93: should be `0x10001`

src/Nethermind/Nethermind.Core/L1PrecompileConstants.cs
dipkakwani: These are mentioned in the RIP:

```
FIXED_GAS_COST 	2000 (tentative)
PER_LOAD_GAS_COST 	2000
```
https://github.com/ethereum/RIPs/blob/master/RIPS/rip-7728.md

Is there any other way we typically estimate gas cost? Perhaps running some benchmarks?

smartprogrammer93 (approved):

smartprogrammer93 (approved):

src/Nethermind/Nethermind.Evm/Precompiles/AddressExtensions.cs
```diff
@@ -36,6 +36,8 @@ public static bool IsPrecompile(this Address address, IReleaseSpec releaseSpec)
                     0x0f => releaseSpec.Bls381Enabled,
                     0x10 => releaseSpec.Bls381Enabled,
                     0x11 => releaseSpec.Bls381Enabled,
+                    // L2 precompiles
+                    0x10001 => releaseSpec.IsRip7728Enabled,
```
flcl42: Will it work? Could you confirm 0x10001 can bypass `(data[4] >>> 24)` check?

src/Nethermind/Nethermind.Evm.Precompiles/L1SloadPrecompile.cs
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        return L1PrecompileConstants.PerLoadGasCost;
+    }
+
+    public (byte[], bool) Run(ReadOnlyMemory<byte> inputData, IReleaseSpec releaseSpec)
+    {
+        if (!releaseSpec.IsRip7728Enabled)
+        {
+            return IPrecompile.Failure;
```
flcl42: We don't check it in other precompiles, probably we don't need to do so here

src/Nethermind/Nethermind.Evm.Precompiles/L1SloadPrecompile.cs
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        if (inputData.Length != L1PrecompileConstants.ExpectedInputLength)
+        {
+            return IPrecompile.Failure;
+        }
+
+        var contractAddress = new Address(inputData.Span[..L1PrecompileConstants.AddressBytes]);
```
flcl42: We usually use types not `var`s

flcl42 (changes requested):

src/Nethermind/Nethermind.Taiko/JsonRpcL1StorageProvider.cs
```diff
@@ -0,0 +1,59 @@
...
+            }).GetAwaiter().GetResult();
+
+            if (response == null)
+            {
+                _logger.Warn($"Failed to read L1 storage: contract={contractAddress}, key={storageKey}, block={blockNumber}");
+                return null;
+            }
+
+            // Parse hex response
+            string hexValue = response.StartsWith("0x") ? response[2..] : response;
```
flcl42: It happens in UInt256.Parse already

src/Nethermind/Nethermind.Taiko/TaikoPlugin.cs
```diff
@@ -59,9 +60,33 @@ public Task Init(INethermindApi api)
...
         return Task.CompletedTask;
     }
 
+    private void InitializeL1SloadIfEnabled()
+    {
+        if (_api?.SpecProvider == null) return;
+
+        var taikoSpec = (TaikoReleaseSpec)_api.SpecProvider.GenesisSpec;
+
+        if (!taikoSpec.IsRip7728Enabled)
```
flcl42, line 74: Maybe we can GetFinalSpec and check it instead of genesis, in case it will be initialized later?

src/Nethermind/Nethermind.Taiko/TaikoPlugin.cs
```diff
@@ -59,9 +60,33 @@ public Task Init(INethermindApi api)
...
         _api.BlockPreprocessor.AddFirst(new MergeProcessingRecoveryStep(_api.Context.Resolve<IPoSSwitcher>()));
 
+        InitializeL1SloadIfEnabled();
+
         return Task.CompletedTask;
     }
 
+    private void InitializeL1SloadIfEnabled()
+    {
+        if (_api?.SpecProvider == null) return;
```
flcl42: Throw an exception, it's an abnormal flow

src/Nethermind/Nethermind.Evm/Precompiles/AddressExtensions.cs
dipkakwani: Fixed the implementation to properly handle the new address.
Tested locally with other precompiles as well:
```
curl -X POST http://localhost:8547 \
  -d '{
    "jsonrpc": "2.0",
    "method": "eth_call",
    "params": [{
      "to": "0x0000000000000000000000000000000000000004",
      "data": "0x1234567890abcdef"
    }, "latest"],
    "id": 4
  }'
{"jsonrpc":"2.0","result":"0x1234567890abcdef","id":4}
```

src/Nethermind/Nethermind.Taiko/TaikoPlugin.cs
```diff
@@ -59,9 +60,36 @@ public Task Init(INethermindApi api)
...
+        InitializeL1SloadIfEnabled();
+
         return Task.CompletedTask;
     }
 
+    private void InitializeL1SloadIfEnabled()
+    {
+        if (_api?.SpecProvider == null)
+        {
+            throw new InvalidOperationException("SpecProvider is not initialized");
```
flcl42: Can be `ArgumentNullException.ThrowIfNull(_api.SpecProvider);` to follow the common pattern. See `OptimismPlugin.cs` for example

src/Nethermind/Nethermind.Taiko/TaikoPlugin.cs
```diff
@@ -59,9 +60,36 @@ public Task Init(INethermindApi api)
...
+
+        var taikoSpec = (TaikoReleaseSpec)_api.SpecProvider.GetFinalSpec();
+
+        if (!taikoSpec.IsRip7728Enabled)
+            return;
+
+        ISurgeConfig surgeConfig = _api.Context.Resolve<ISurgeConfig>();
+
+        if (string.IsNullOrEmpty(surgeConfig.L1EthApiEndpoint))
+            throw new ArgumentException("L1EthApiEndpoint must be provided in the Surge configuration to use L1SLOAD precompile");
```
flcl42: $"{nameof(surgeConfig.L1EthApiEndpoint)} must be ...

src/Nethermind/Nethermind.Evm/Precompiles/AddressExtensions.cs
```diff
@@ -13,37 +13,43 @@ public static class AddressExtensions
...
-                {
-                    0x00 => releaseSpec.IsEip7951Enabled || releaseSpec.IsRip7212Enabled,
-                    _ => false
-                },
-                _ => false
-            };
+
+        return (data[4] & 0x000000ff) == 0
+               && data[3] == 0 && data[2] == 0 && data[1] == 0 && data[0] == 0
+               && ((data[4] >>> 12) & 0xfff) switch
```
flcl42: Could you add a test that confirms the address has precompile activated when Rip7728 is active and does not have it for the not active eip?
More like a matter of personal preference ig, but four nibbles here and below might look a lil bit better.

flcl42: Yeah 4 is better, covers all the range. Can cause an issue otherwise
image

dipkakwani: Thanks, added 4 nibbles and a new test. Now all the nibbles should be covered in the cases.
Would be good if you check the logic once again :sweat_smile:

flcl42 (approved):

PR #9041, opened 2025-07-27
Limit deleted block on invalid block

asdacap:
- When an invalid block happen during forward sync from genesis, it will try to delete millions of block which overwrites existing logs making debugging very hard.
- Limit deleted blocks on invalid block to 256 block, which should be a reasonably large size in case of long invalid reorg.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Testing...


LukaszRozmej:
What about all those blocks that are after? They will never be valid.

asdacap:
I mean, these are likely bug in nethermind that assume the block is invalid.

PR #9044, opened 2025-07-28
Hardening/Separate buffer allocator

asdacap:
- Separate buffer allocator for discovery, rlpx and general purpose allocator.
- This prevent leak from one allocator to affect another allocator.
- Move all `PooledByteBufferAllocator.Default` to `NethermindBuffers.Default`.
  - This make it easy to replace for debugging. 
- Also add some metric.

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

#### Notes on testing

- [X] Mainnet sync normally.
- [X] Metric shows up as expected.

### DIFF
--- a/src/Nethermind/Nethermind.Serialization.Rlp/NethermindBuffers.cs
+++ b/src/Nethermind/Nethermind.Serialization.Rlp/NethermindBuffers.cs
@@ -2,14 +2,29 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 
 using DotNetty.Buffers;
+using DotNetty.Common.Internal;
 
 namespace Nethermind.Serialization.Rlp;
 
-public class NethermindBuffers
+public static class NethermindBuffers
 {
     /// <summary>
     /// General <see cref="IByteBufferAllocator"/> used for general purpose deserialization.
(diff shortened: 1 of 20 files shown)


src/Nethermind/Nethermind.Serialization.Rlp/NethermindBuffers.cs
```diff
@@ -0,0 +1,73 @@
...
+
+    /// <summary>
+    /// Allocator used for serializing and deserializing rlpx messages
+    /// </summary>
+    public static IByteBufferAllocator RlpxAllocator = PooledByteBufferAllocator.Default;
+
+    /// <summary>
+    /// Similar to <see cref="RlpxAllocator"/> but for discovery messages.
+    /// </summary>
+    public static IByteBufferAllocator DiscoveryAllocator = PooledByteBufferAllocator.Default;
```
LukaszRozmej, line 27: By default they are initialized to same allocator?

asdacap: Yes, the are changed by `MemoyHintMan` at start.

PR #9048, opened 2025-07-28
Add support for Prometheus

### LINKED ISSUE #9047: Export metrics to Prometheus
Currently, Kute supports two export formats: "pretty" and JSON. This is fine in most scenarios but it would be nice if we could export the data to some service that supports long term storage and facilitates visualization.

We would like for Kute to be able to export its metrics to Prometheus, so they can be later visualizaed in Grafana.

- Export metrics into a relational database like Postgres, which later would be used as a data source in Grafana.
- Generate a standalone HTML report which includes graphs, tables and charts that can be loaded into a browser.

**Additional context**
We should not include infrastructure code. Instead, assume that the required services (Grafana, Prometheus, etc.) is already available.

emlautarom1:
Closes #9047

## Changes

- Add support for reporting to Prometheus
- Improve batch request validation.

- [ ] Optimization
- [ ] Refactoring
- [x] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

The new reporting was tested using a `docker-compose.yml` file with the following versions:
- Prometheus `v3.5.0`
- Prometheus Push Gateway `1.11.1`
- 
Unit tests were added while fixing batch request validation.

- [x] Yes
- [ ] No

The `README.md` file in the `Kute` directory was updated to include instructions on how to use the new exporting mechanism.

- [ ] Yes
- [x] No

## Remarks

Since Kute is a short-lived CLI tool it's not reasonable for Prometheus to scrape metrics during its execution. Instead, we rely on [PushGateway](https://github.com/prometheus/pushgateway) which is a "push acceptor for ephemeral and batch jobs". This service is then scraped by Prometheus, which can later be used as a data source in Grafana.

To identify requests we're leveraging the existing JSON-RPC `id` field. In case of batch requests we make IDs as ranges: `XX:YY` where `XX` is the JSON-RPC `id` from the first request in the batch, and `YY` is the last.


tools/Kute/Nethermind.Tools.Kute/Metrics/PrometheusPushGatewayMetricsReporter.cs
```diff
@@ -0,0 +1,101 @@
...
+        _singleDuration = factory.CreateHistogram("single_duration", "", labelName: "jsonrpc_id");
+        _batchDuration = factory.CreateHistogram("batch_duration", "", labelName: "jsonrpc_id");
+
+        _endpoint = endpoint;
+        _pusher = new MetricPusher(new MetricPusherOptions
+        {
+            CollectorRegistry = registry,
+            Endpoint = _endpoint,
+            Instance = "kute",
+            Job = $"{Guid.NewGuid()}",
```
cbermudez97: Both Instance and Job are later used as labels for the exported metrics. By convention `Job` should be the one with the `kute` value while `Instance` could be either undefined or leave it up to the user. Having the possibility to add custom labels( including the Instance) would be good as well but can be done in a separate PR.

cbermudez97 (approved):

tools/Kute/Nethermind.Tools.Kute/Metrics/PrometheusPushGatewayMetricsReporter.cs
emlautarom1: Swapped `Job`/`Instance` to follow the convention. As for custom labels, I think we cal leave it for a separate PR

PR #9049, opened 2025-07-29
Harden background queue

benaadams:
## Changes

- Use actual count and lock to synchronize queue capacity rather than proxy (better accounts for requeues)
- Update Metrics queue count more often for more accurate value

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
+++ b/src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
@@ -43,6 +43,7 @@ public class BackgroundTaskScheduler : IBackgroundTaskScheduler, IAsyncDisposabl
     private readonly IBlockProcessor _blockProcessor;
     private readonly IChainHeadInfoProvider _headInfo;
     private readonly int _capacity;
+    private long _queueCount;
 
     private CancellationTokenSource _blockProcessorCancellationTokenSource;
     private bool _disposed = false;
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
```diff
@@ -165,17 +168,33 @@ public void ScheduleTask<TReq>(TReq request, Func<TReq, CancellationToken, Task>
...
-        if (!_capacity.Wait(millisecondsTimeout: 0))
+
+        bool success = false;
+        lock (_queueLock)
+        {
+            int currentCount = _taskQueue.Reader.Count;
+            if (currentCount < _capacity)
+            {
+                success = _taskQueue.Writer.TryWrite(activity);
+            }
```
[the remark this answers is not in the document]
benaadams: does

src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
```diff
@@ -165,17 +168,33 @@ public void ScheduleTask<TReq>(TReq request, Func<TReq, CancellationToken, Task>
...
-        _taskQueue.Writer.TryWrite(activity);
 
-        Evm.Metrics.NumberOfBackgroundTasksScheduled = _taskQueue.Reader.Count;
+        [StackTraceHidden, DoesNotReturn]
+        static void UnableToWriteToTaskQueue()
+            => throw new InvalidOperationException("Unable to write to background task queue.");
     }
 
+    private void UpdateQueueCount()
+        => Evm.Metrics.NumberOfBackgroundTasksScheduled = _taskQueue.Reader.Count;
```
LukaszRozmej: wouldn't it be cheaper to keep own counting?

benaadams: Sure, changed

PR #9064, opened 2025-07-31
Incorrect Gas Estimation & Refactoring

### LINKED ISSUE #9061: Incorrect gas estimation when high gas limit is needed to execute the transaction but not used
**Description**

Consider the following contract:

```solidity
contract GasCraver {
    bool public stateChanged = false;

    function foo() external {
        if (gasleft() < 1_000_000) {
            revert("I want more gas");
        }
        stateChanged = true;
    }
}
```

`foo()` can be called without reverting if using a gas limit larger than ~1M, but the executed transaction only consumes ~45k gas.
When calling `eth_estimateGas` on Nethermind I see an estimation of ~45k, which would cause the call to revert.
According to the [specs for `eth_estimateGas`](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_estimategas), I instead expect it to return how much gas is necessary to allow the transaction to complete, which would be a bit more than 1M gas in this case.

**Steps to Reproduce**

There's a deployed version of this contract for testing at address [0xbFcbE82a626ed578B66e80DD72825a9d524357cd](https://gnosisscan.io/address/0xbFcbE82a626ed578B66e80DD72825a9d524357cd#code) on Gnosis.
Calling `eth_estimateGas` on `Nethermind/v1.32.2+d3e7eb98/linux-x64/dotnet9.0.6` returns 43702.
Using another node, like Erigon, returns a more reasonable result of 1021181.

Trying the same on Sepolia ([0xc1ea023E38c3d3472ae688C24131BcBc805C1164](https://sepolia.etherscan.io/address/0xc1ea023E38c3d3472ae688C24131BcBc805C1164#code)) returns 1029663 with Reth and 1031376 with Geth. I don't have access to a Nethermind node on Sepolia to see if it's a Gnosis issue.

You can try out this call with custom node URLs by editing the following script.

<details><summary>Helper script for testing</summary>

```bash
#!/bin/bash

set -o errexit -o nounset -o pipefail

rpc_id=0

rpc() {
  local node=$1
  local method=$2
  local params=$3
  rpc_id=$(("$rpc_id" + 1))
  curl \
    --silent \
    -X POST \
    --max-time 15 \
    "$node" \
    --data "{\"method\":\"$method\",\"params\":$params,\"id\":$rpc_id,\"jsonrpc\":\"2.0\"}"
}

get_node_version() {
  local node=$1
  rpc "$node" web3_clientVersion '[]' | jq --raw-output '.result'
}

get_estimation() {
  local node=$1
  local gas_craver=$2
  local address_with_eth='0x5112D584a1C72Fc250176B57aEba5fFbbB287D8F'
  rpc "$node" "eth_estimateGas" "[{\
\"from\":\"$address_with_eth\",\
\"to\":\"$gas_craver\",\
\"data\":\"0xc2985578\"
}]" \
  | jq --raw-output '.result'
}

gas_craver='0xbFcbE82a626ed578B66e80DD72825a9d524357cd'

echo "Testing contract $gas_craver."
for node in \
'https://rpc.gnosis.gateway.fm' \
'https://gnosis-mainnet.public.blastapi.io' \
; do
  echo "> Using RPC $node..."
  version=$(get_node_version "$node")
  estimation=$(get_estimation "$node" "$gas_craver")
  echo "$((16#${estimation##0x})) with version $version"
done
```

Output:
```
Testing contract 0x6e8E85BC6FF7925533Ad7c19E4b9f0d167A95157.
> Using RPC https://rpc.gnosis.gateway.fm...
43702 with version Nethermind/v1.32.2+d3e7eb98/linux-x64/dotnet9.0.6
> Using RPC https://gnosis-mainnet.public.blastapi.io...
1021181 with version erigon/3.0.11/linux-amd64/go1.23.6
```

Same script but using Sepolia and different nodes:
```
Testing contract 0xc1ea023E38c3d3472ae688C24131BcBc805C1164.
> Using RPC https://eth-sepolia.public.blastapi.io...
1029663 with version reth/v1.6.0-d8451e5/x86_64-unknown-linux-gnu
> Using RPC https://ethereum-sepolia-rpc.publicnode.com...
1031376 with version Geth/v1.15.11-stable-36b2371c/linux-amd64/go1.23.10
```
</details>

**Actual behavior**

Estimate from `eth_estimateGas` causes subsequent call to revert.

 **Expected behavior**
 
 A much higher estimate should be returned by `eth_estimateGas`.
 
 I didn't run this test locally but I tested with all [free RPCs from Chainlist](https://chainlist.org/chain/100/) and used  `web3_clientVersion` to narrow down the issue to Nethermind nodes. Note that sometimes some RPCs on Gnosis use both Nethermind and Erigon in parallel, which could give misleading results in this test.
 Affected versions include:
 
- Nethermind/v1.32.2+d3e7eb98/linux-x64/dotnet9.0.6
- Nethermind/v1.31.11+2be1890e/linux-x64/dotnet9.0.5
- Nethermind/v1.31.8+8fca47e8/linux-x64/dotnet9.0.4

svlachakis:
Fixes Closes Resolves #9061

**Root Cause:** Nethermind's gas estimation binary search only treats OutOfGas errors as "insufficient gas", but ignores contract reverts. This is inconsistent with other Ethereum clients.

See here https://github.com/ethereum/go-ethereum/blob/d4a3bf1b23e3972fb82e085c0e29fe2c4647ed5c/eth/gasestimator/gasestimator.go#L214

**How Gas Estimation Should Work**
Gas estimation uses binary search to find the minimum gas limit where a transaction succeeds:

- Execute transaction with test gas limit
- If execution fails for any reason → increase gas limit
- If execution succeeds → decrease gas limit
- Repeat until convergence

## Changes

- Refactored GasEstimator.cs
- Removed unnecessary tracer and used another existing & added OutOfGas and Revert handling.
- Added Tests
- @emlautarom1 helped to add Test for the specific use case of the issue

Binary search now treats contract reverts as "insufficient gas" and continues searching until it finds a gas limit where the contract doesn't revert.

- [ ] Optimization
- [X] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

I tested also on Sepolia.

Screenshot 2025-08-03 at 9 15 58 PM

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No


src/Nethermind/Nethermind.Evm.Test/Tracing/GasEstimationTests.cs
```diff
@@ -403,5 +403,19 @@ public TestEnvironment()
                 estimator = new(_transactionProcessor, _stateProvider, _specProvider, blocksConfig);
             }
         }
+
+        [Test]
+        public void EstimateGasTracer_should_set_status_code_correctly()
```
emlautarom1: This test is not covering any of the changes in `GasEstimator`. Could we add a test that does?

src/Nethermind/Nethermind.Blockchain/Tracing/GasEstimator.cs
```diff
@@ -125,18 +125,16 @@ private long BinarySearchEstimate(long leftBound, long rightBound, Transaction t
...
 
         _transactionProcessor.SetBlockExecutionContext(new(block, _specProvider.GetSpec(block)));
-        _transactionProcessor.CallAndRestore(txClone, tracer.WithCancellation(token));
+        TransactionResult result = _transactionProcessor.CallAndRestore(txClone, gasTracer.WithCancellation(token));
 
-        return !tracer.OutOfGas;
+        return result.Success && gasTracer.StatusCode == StatusCode.Success;
     }
 
     private class OutOfGasTracer : TxTracer
```
emlautarom1: What about this tracer? It seems like is no longer used (can't find any references to it and code coverage shows no usages).

emlautarom1 (changes requested):
Some things pending before merging.

src/Nethermind/Nethermind.Blockchain/Tracing/GasEstimator.cs
svlachakis: Looks like it's unused, I'll delete.

src/Nethermind/Nethermind.Evm.Test/Tracing/GasEstimationTests.cs
svlachakis: @emlautarom1 The only behavioral change on `GasEstimator.cs` is the return statement `return result.Success && gasTracer.StatusCode == StatusCode.Success;`, all the rest are just parameters.

Since the method is private and we can't directly test it, I wrote a test that validates the return statement for the gasTracer part - specifically testing that StatusCode is set correctly by MarkAsSuccess and MarkAsFailed. This ensures the new condition `gasTracer.StatusCode == StatusCode.Success` will work correctly since we're validating that StatusCode gets set to the right values.

Do you think we should approach it somehow else or maybe change the function to non-private in order to test it directly from the unit tests?

emlautarom1: I think we should trigger the behavior through the `GasEstimator.Estimate` method rather than exposing internals.
Take a look at the other tests in this file for references on how we could achieve this.

svlachakis: Done

src/Nethermind/Nethermind.Evm.Test/Tracing/GasEstimationTests.cs
```diff
@@ -375,6 +375,66 @@ public void Estimate_UseZeroErrorMargin_EstimationResultIsExact()
             }
         }
 
+        [Test]
+        public void Should_return_zero_when_out_of_gas_detected_during_estimation()
```
emlautarom1, line 378: When the estimate is zero, `out string? err` should be not null, right? Can we change the test name/assertions so that this is the expected result? It does not have a lot of sense to have an estimate of zero as a desired output AFAIK.

emlautarom1 (approved):
LGTM.

src/Nethermind/Nethermind.Evm.Test/Tracing/GasEstimationTests.cs
svlachakis: @emlautarom1 In the test, we're simulating an out-of-gas condition:
`testEnvironment.tracer.ReportActionError(EvmExceptionType.OutOfGas);`

This sets `gasTracer.OutOfGas = true`, which makes `TryExecutableTransaction` return `false`. But I think when the binary search exhausts all possibilities and even the maximum gas limit fails, it returns 0 but doesn't set an error message because of this code in `GasEstimator.cs` line 119.

```
        if (rightBound == cap && !TryExecutableTransaction(tx, header, rightBound, token, gasTracer))
        {
            return 0;
        }
```

Do you think we should proceed to further refactoring?

emlautarom1: So a transaction that can never be executed will return zero as an estimation without reporting an error? That does not sound right. What is the behavior of other clients like Geth/Reth?

svlachakis: I agree it doesn't sound right and looks like Geth returns 0 but with error.

https://github.com/ethereum/go-ethereum/blob/792de5d2e3b40837aadb65993e387bf5aa5bce9f/eth/gasestimator/gasestimator.go#L152

svlachakis: @emlautarom1 looks like the issue is that when we try to simulate out of gas scenario with Address.Zero in the test, it falls in here in `GasEstimator.cs` which returns 0 without error.

```
        // Calculate and return additional gas required in case of insufficient funds.
        UInt256 senderBalance = _stateProvider.GetBalance(tx.SenderAddress);
        if (tx.ValueRef != UInt256.Zero && tx.ValueRef > senderBalance && !tx.IsSystem())
        {
            return gasTracer.CalculateAdditionalGasRequired(tx, releaseSpec);
        }
```

If I change this code to return error apart from just returning 0, it breaks multiple tests across different parts in the codebase.

svlachakis:
@emlautarom1 Please take a last look after my last changes on refactoring & error handling and let's merge as long as it looks good to you.

src/Nethermind/Nethermind.Blockchain/Tracing/GasEstimator.cs
```diff
@@ -14,161 +15,313 @@
...
+{
+    EstimationResult Estimate(Transaction tx, BlockHeader header, EstimateGasTracer gasTracer,
+                             EstimationBounds bounds, int errorMargin, CancellationToken token);
+}
+
+internal interface ITransactionExecutor
+{
+    bool TryExecute(Transaction transaction, BlockHeader header, long gasLimit,
+                   CancellationToken token, EstimateGasTracer gasTracer);
+}
```
emlautarom1: Why do we need all these new interfaces?

src/Nethermind/Nethermind.Blockchain/Tracing/GasEstimator.cs
```diff
@@ -14,161 +15,313 @@
...
+        txClone.GasLimit = gasLimit;
+
+        _transactionProcessor.SetBlockExecutionContext(new(header, _specProvider.GetSpec(header)));
+        TransactionResult result = _transactionProcessor.CallAndRestore(txClone, gasTracer.WithCancellation(token));
+
+        return result.Success && gasTracer.StatusCode == StatusCode.Success && !gasTracer.OutOfGas;
+    }
+}
+
+internal class BinarySearchGasEstimationStrategy : IGasEstimationStrategy
```
emlautarom1: Right now we don't have any other strategy than binary search, and most likely will never have a different one (we don't switch at runtime or allow for configuration). Thus, I'd inline this implementation as we did before.

src/Nethermind/Nethermind.Facade.Test/BlockchainBridgeTests.cs
```diff
@@ -586,7 +586,7 @@ public void Call_tx_returns_noError()
     }
 
     [Test]
-    public void EstimateGas_tx_returns_noError()
+    public void EstimateGas_invalid_tx_returns_error()
```
emlautarom1, line 589: Why did we need to change this test?

svlachakis: The test was using an empty transaction that should fail validation. The old code had broken behavior where invalid transactions incorrectly succeeded. When we fixed the transaction processor, this test started failing because it was expecting the wrong behavior. Changing the test name and expectation reflects the correct behavior - invalid transactions should return errors.

src/Nethermind/Nethermind.Blockchain/Tracing/GasEstimator.cs
svlachakis: I introduced them to improve testability, separate concerns for easier understanding and single responsibility, and enable safer modifications by isolating changes to specific components.

LukaszRozmej:
any progres?

emlautarom1:
@LukaszRozmej the fix is implemented and the users report that it has solved their issue, but the PR is blocked due to the additional refactoring: I would like to have an additional approval before merging.

LukaszRozmej:
@emlautarom1 should we just revert the refactoring?

src/Nethermind/Nethermind.Blockchain/Tracing/EstimateGasTracer.cs
```diff
@@ -131,14 +134,21 @@ public override void ReportActionEnd(long gas, Address deploymentAddress, ReadOn
 
     public override void ReportActionError(EvmExceptionType exceptionType)
     {
+        OutOfGas |= exceptionType == EvmExceptionType.OutOfGas || exceptionType == EvmExceptionType.Revert;
```
LukaszRozmej: Can we put that into some method? It is done 3 times

LukaszRozmej (commented):
Yeah a bit overcomplicated

svlachakis:
@LukaszRozmej @emlautarom1 I'm good with any approach (with refactoring or reverting), just let me know how we should proceed 😄

src/Nethermind/Nethermind.Evm.Test/Tracing/GasEstimationTests.cs
svlachakis: this is fixed, I've fixed the other places which were failing

src/Nethermind/Nethermind.Blockchain/Tracing/EstimateGasTracer.cs
svlachakis: done, I reused this method since Action & Operation errors have the same exceptionTypes for this tracer.

PR #9067, opened 2025-08-01
Remove storage cell hash representation

asdacap:
- Remove `StorageCell` has representation. It always represent slot now.

- [X] Optimization
- [X] Refactoring

#### Requires testing

- [ ] Yes
- [X] No

- [ ] Yes
- [X] No


asdacap:
Used by JS tracer. Can't removve

PR #9072, opened 2025-08-01
Check block limit too

flcl42:
In case block blob limit becomes lower, we need to clean up tx pool and disallow inclusion in the very first block in new fork. Before this fix transaction might pass tx validation for old rules with big allowance of blobs per block and then pass lighter validation during block production

## Changes

- Add additional check that is triggered during tx pool new head handling and block production

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No


marcindsobczak (approved):

src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.Blobs.cs
```diff
@@ -1048,7 +1048,41 @@ public async Task should_evict_txs_with_too_many_blobs_after_fork()
...
+            const int regularMaxBlobCount = 9;
+            const int decreasedMaxBlobCount = regularMaxBlobCount - 1;
+
+            TestSpecProvider provider = new(new ReleaseSpec
+            {
+                IsEip4844Enabled = true,
+                MaxBlobCount = regularMaxBlobCount,
+            })
+            {
+                SpecToReturn = new ReleaseSpec
```
Marchhill: maybe add comment to make it more clear that this is the spec after the fork

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
```diff
@@ -237,9 +225,24 @@ private static ValidationResult ValidateBlobFields(Transaction transaction, IRel
         return ValidationResult.Success;
     }
 
-    internal static ValidationResult ValidateBlobPerTxLimit(ulong txBlobGas, IReleaseSpec spec)
+    internal static ValidationResult ValidateBlobGasLimits(int txBlobCount, IReleaseSpec spec)
     {
+        if (txBlobCount < Eip4844Constants.MinBlobsPerTransaction)
```
Marchhill, line 230: could add something to spec instead of using the constant directly here. Also for consistency with the maxBlobGas check we could use min gas instead of min number of blobs, although either is fine

src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.Blobs.cs
flcl42: Renamed to `NextForkSpec`

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
flcl42: Eip4844Constants hold everything that is still constant for every spec. If we would move MinBlobsPerTransaction to the spec, we should do the same for GasPerBlob. Hm we have MinBlobGasPrice in Eip4844Constants, which depends on spec, which is really worth to be taking out of constants, let me check it in a separate pr.

PR #9074, opened 2025-08-01
Add HTML report format

emlautarom1:
## Changes

- Add `html` as a report format
- Report progress in `stderr`
- Fix a bug where Kute would not pick up a single file as input when running in WSL.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

Verified that in WSL the change makes Kute work as intended when using a single file as input.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks

The HTML report is an MVP intended to be used to visualize the data without relying on tools like Prometheus/Grafana which are more heavyweight. The charting library used is [Google Charts](https://developers.google.com/chart).


tools/Kute/Nethermind.Tools.Kute/Metrics/HtmlMetricsReportFormatter.cs
```diff
@@ -0,0 +1,32 @@
+// SPDX-FileCopyrightText: 2023 Demerzel Solutions Limited
```
rubo:
```suggestion
// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
```

rubo (approved):

PR #9078, opened 2025-08-02
feat: Make `GenerateOpCodes` and `ExecuteTransaction` overridable in VM

hudem1:
Arbitrum has its own custom implementation for `GasPrice` opcode.
See [Arbitrum's PR](https://github.com/NethermindEth/nethermind-arbitrum/pull/127).

In my suggested modifications here, the virtual machine warmup `RunOpCodes` will not have the updated GasPrice opcode. To have it, I would need to override the whole `VirtualMachine.WarmUpEvmInstructions();`, which seems a bit overkill. Let me know it's still better to have the update there too.

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks

### DIFF
--- a/src/Nethermind/Nethermind.Evm/VirtualMachine.cs
+++ b/src/Nethermind/Nethermind.Evm/VirtualMachine.cs
@@ -137,7 +137,7 @@ public unsafe partial class VirtualMachineBase(
     /// <exception cref="EvmException">
     /// Thrown when an EVM-specific error occurs during execution.
     /// </exception>
-    public TransactionSubstate ExecuteTransaction<TTracingInst>(
+    public virtual TransactionSubstate ExecuteTransaction<TTracingInst>(
         EvmState evmState,
         IWorldState worldState,
         ITxTracer txTracer)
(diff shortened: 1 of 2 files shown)


wurdum (approved):

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -88,7 +88,7 @@ public unsafe partial class VirtualMachineBase(
 
     private ICodeInfoRepository _codeInfoRepository;
 
-    private OpCode[] _opcodeMethods;
+    protected OpCode[] _opcodeMethods;
```
LukaszRozmej: do you really need to access it directly?

LukaszRozmej (commented):
I think you would actually do better to pack `EvmInstructions.GenerateOpCodes` call from this method into a virtual method.

AnkushinDaniil (approved):

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
hudem1: no need indeed, your idea of putting `EvmInstructions.GenerateOpCodes` in a virtual method sounds much better! Don't know why i haven't thought of this!

PR #9082, opened 2025-08-03
Optimize unused logging from Diagnostics, Activity and Metrics

benaadams:
## Changes

- Disables system-level diagnostics and activity propagation through runtime configuration
- Optimizes logging performance by adding level checks before expensive log message formatting
- Improves DotNetty integration with proper logger factory configuration

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Network.Discovery/Discv5/NethermindLoggerFactory.cs
+++ b/src/Nethermind/Nethermind.Network.Discovery/Discv5/NethermindLoggerFactory.cs
@@ -26,15 +26,7 @@ public sealed class NethermindLoggerFactory(ILogManager logManager, bool lowerLo
         {
             if (lowerLogLevel)
             {
-                if (logLevel <= MsLogLevel.Information)
-                {
-                    // DotNetty outputs Trace level data at Info
-                    logLevel = MsLogLevel.Trace;
-                }
-                else
-                {
-                    logLevel = logLevel - 1;
-                }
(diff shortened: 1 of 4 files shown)


LukaszRozmej (approved):
Looks ok to me, but maybe @flcl42 and @asdacap should have a say as I don't know dotnetty that well

benaadams:
> Looks ok to me, but maybe @flcl42 and @asdacap should have a say as I don't know dotnetty that well

Mostly is changing default logger to be our logger rather than eventsource (which we aren't listening to; but it generates a bunch of junk)

src/Nethermind/Nethermind.Network.Discovery/Discv5/NethermindLoggerFactory.cs
```diff
@@ -43,30 +43,49 @@ public bool IsEnabled(MsLogLevel logLevel)
...
         }
+
+        private static MsLogLevel LowerLogLevel(MsLogLevel logLevel)
+        {
+            // DotNetty outputs Trace level data at Info
+            return logLevel switch
+            {
+                MsLogLevel.Critical => MsLogLevel.Error,
+                MsLogLevel.Error => MsLogLevel.Warning,
+                MsLogLevel.Warning => MsLogLevel.Information,
```
flcl42, line 84: Just imho, critical/error/warning can stay the same

flcl42 (approved):

PR #9083, opened 2025-08-03
Optimize Peer Range Broadcast

benaadams:
## Changes

- _List the changes_

- [x] Optimization

#### Requires testing

- [x] No


src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/V69/Eth69ProtocolHandler.cs
```diff
@@ -158,7 +159,7 @@ public void NotifyOfNewRange(BlockHeader earliest, BlockHeader latest)
         if (earliest.Number > latest.Number)
             throw new ArgumentException($"Earliest block ({earliest.Number}) greater than latest ({latest.Number}) in BlockRangeUpdate.");
 
-        if (latest.Hash is null)
+        if (latest.Hash is null || latest.Hash == Hash256.Zero)
```
LukaszRozmej: Is it any better?

```suggestion
        if (latest.Hash is null || latest.Hash.IsZero())
```

src/Nethermind/Nethermind.Network/Rlpx/RlpxHost.cs
```diff
@@ -128,7 +128,7 @@ public async Task Init()
                     .ChildOption(ChannelOption.Allocator, NethermindBuffers.RlpxAllocator)
                     .ChildOption(ChannelOption.SoBacklog, 100)
                     .ChildOption(ChannelOption.TcpNodelay, true)
-                    .ChildOption(ChannelOption.SoTimeout, (int)_connectTimeout.TotalMilliseconds)
+                    .ChildOption(ChannelOption.SoTimeout, (int)(_connectTimeout.TotalMilliseconds * 1.5))
```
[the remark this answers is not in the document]
LukaszRozmej: yeah why 1.5x?

src/Nethermind/Nethermind.Synchronization.Test/SyncServerTests.cs
```diff
@@ -697,11 +697,16 @@ public void Broadcast_BlockRangeUpdate_when_latest_increased_enough()
...
 
-        const int blocksCount = 100;
+        const int blocksCount = 35;
         var startBlock = (int)localBlockTree.Head!.Number;
         localBlockTree.AddBranch(blocksCount, splitBlockNumber: startBlock, splitVariant: 0);
+        await Task.Delay(10);
+        localBlockTree.AddBranch(blocksCount, splitBlockNumber: startBlock + blocksCount, splitVariant: 0);
+        await Task.Delay(10);
+        localBlockTree.AddBranch(blocksCount, splitBlockNumber: startBlock + blocksCount * 2, splitVariant: 0);
+        await Task.Delay(10);
```
LukaszRozmej: maybe actual task to await instead of delays?

src/Nethermind/Nethermind.Synchronization/Peers/PeerInfo.cs
```diff
@@ -46,6 +46,18 @@ public PeerInfo(ISyncPeer syncPeer)
 
         public Hash256 HeadHash => SyncPeer.HeadHash;
 
+        private long _lastNotifiedHeadNumber;
+
+        [MethodImpl(MethodImplOptions.Synchronized)]
+        public bool ShouldNotifyHeadNumber(long headNumber)
+        {
+            // Also notify if same header as could be reorg with different hash
```
LukaszRozmej, line 54: What about other reorgs?

benaadams: Can a block be reorged without replacement? (Same or higher number as head)

LukaszRozmej: no, but you can reorg to lower level block

src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/V69/Eth69ProtocolHandler.cs
benaadams: Changed

src/Nethermind/Nethermind.Synchronization.Test/SyncServerTests.cs
benaadams: Fixed

src/Nethermind/Nethermind.Synchronization/Peers/PeerInfo.cs
benaadams: Is notifying the low->head range with 32 block delays, so its head notification shouldn't go backwards even from reorg, unless it throws away 32 blocks and number goes backwards more than that?

PR #9088, opened 2025-08-05
Per contract state

asdacap:
- Built in top of #9028 
- Extract some of the per-contract logic into an internal class `PerContractState`.
- This reduces some dictionary re-lookup but mainly to reduce change on another PR.

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

#### Notes on testing

- Testing...

### DIFF
--- a/src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
+++ b/src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
-using System.Linq;
 using System.Runtime.CompilerServices;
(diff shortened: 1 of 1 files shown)


benaadams:
Oooof, merge conflict

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
```diff
@@ -620,4 +490,162 @@ public int GetHashCode([DisallowNull] UInt256 obj)
                 => MemoryMarshal.AsBytes(MemoryMarshal.CreateReadOnlySpan(in obj, 1)).FastHash();
         }
     }
+
+    private class PerContractState
```
benaadams:
```suggestion
    private sealed class PerContractState
```

benaadams (approved):

PR #9089, opened 2025-08-05
Feature/Worldstate backend

asdacap:
- Require #9088.
- `WorldState` does quite a decent amount of logic that is not related to state storage.
- This PR extract the state storage part into an interface called `IWorldStateBackend`.
  - This is implemented as `TrieStoreWorldStateBackend`. 
- Different state layout is expected to only need to reimplement `IWorldStateBackend` instead of the full `IWorldState`. 
- The interface works at address/slot level instead of hash, except for one strange method call.
  - This allow erigon style state.
- Additionally, we can try to remove `IWorldState` to allow devirtualized call to `WorldState`. 

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- [X] Mainnet snap sync
- [X] Mainnet era replay

### DIFF
--- a/src/Nethermind/Nethermind.State/StateProvider.cs
+++ b/src/Nethermind/Nethermind.State/StateProvider.cs
@@ -491,8 +491,6 @@ namespace Nethermind.State
                 if (isTracing) TraceNoChanges();
 
                 codeFlushTask.GetAwaiter().GetResult();
-
-                codeFlushTask.GetAwaiter().GetResult();
                 return;
             }
 
--- a/src/Nethermind/Nethermind.Trie/NodeStorageCache.cs
+++ b/src/Nethermind/Nethermind.Trie/NodeStorageCache.cs
@@ -11,7 +11,7 @@ public sealed class NodeStorageCache
 {
     private ConcurrentDictionary<NodeKey, byte[]> _cache = new();
 
-    private bool _enabled = false;
+    private volatile bool _enabled = false;
 
     public bool Enabled
     {


src/Nethermind/Nethermind.State/StateProvider.cs
```diff
@@ -906,14 +894,12 @@ public void Reset(bool resetBlockChanges = true)
             void Trace() => _logger.Trace("Clearing state provider caches");
         }
 
-        public void CommitTree()
+        public void UpdateStorageRootIfNeeded()
```
tanishqjasoria: UpdateStorageRoot? doesn't seem to update the storage root

src/Nethermind/Nethermind.State/StateProvider.cs
```diff
@@ -100,7 +91,11 @@ public Hash256 StateRoot
                 [DoesNotReturn, StackTraceHidden]
                 static void ThrowStateRootNeedsToBeUpdated() => throw new InvalidOperationException("State root needs to be updated");
             }
-            set => _tree.RootHash = value;
+        }
+
+        public void SetBackendTree(IWorldStateBackend.IStateTree stateTree)
```
tanishqjasoria: maybe add explicit handling when the tree is not set and the state is accessed

src/Nethermind/Nethermind.State/StateProvider.cs
asdacap: ?

src/Nethermind/Nethermind.State/StateProvider.cs
asdacap: I think ben is working on that.

src/Nethermind/Nethermind.State/StateTree.cs
```diff
@@ -73,6 +75,31 @@ public void Set(Address address, Account? account)
...
+                KeccakCache.ComputeTo(key.Bytes, out ValueHash256 keccak);
+
+                Rlp accountRlp = account is null ? null : account.IsTotallyEmpty ? StateTree.EmptyAccountRlp : Rlp.Encode(account);
+
+                _bulkWrite.Add(new BulkSetEntry(keccak, accountRlp?.Bytes));
+            }
+
+            public void Dispose()
+            {
+                tree.BulkSet(_bulkWrite);
```
LukaszRozmej: would sorting `_bulkWrite` before help?

src/Nethermind/Nethermind.State/StorageTree.cs
```diff
@@ -145,6 +168,29 @@ void SetWithKeyGenerate(in UInt256 index, byte[] value)
...
+
+            public void Set(in UInt256 index, byte[] value)
+            {
+                StorageTree.ComputeKeyWithLookup(index, _keyBuff.BytesAsSpan);
+                _bulkWrite.Add(StorageTree.CreateBulkSetEntry(_keyBuff, value));
+            }
+
+            public void Dispose()
+            {
+                storageTree.BulkSet(_bulkWrite);
```
LukaszRozmej: sorting?

LukaszRozmej (commented):
I have mixed feelings, the things I dislike:
1. Hate the name `backend`
2. Not sure if possible but would prefer something even simpler with having Get and Set for Address -> Account and (Address, Index) ->Slot without all the other hassle, but this is closer
3. All those things feel even more complicated

src/Nethermind/Nethermind.State/StateTree.cs
asdacap: Tried, the sorting take more time than doing internal partial sort.

src/Nethermind/Nethermind.State/StateProvider.cs
benaadams: I think the suggestion is 
```diff
-public void UpdateStorageRootIfNeeded()
+public void UpdateStateRootIfNeeded()
```

src/Nethermind/Nethermind.State/WorldStateManager.cs
```diff
@@ -11,12 +11,13 @@
 using Nethermind.State.SnapServer;
 using Nethermind.Trie;
 using Nethermind.Trie.Pruning;
+using Org.BouncyCastle.Pqc.Crypto.Picnic;
```
Marchhill: not needed

src/Nethermind/Nethermind.State/TrieStoreScopeProvider.cs
```diff
@@ -0,0 +1,320 @@
...
+            else
+            {
+                if (_wasSetCalled) throw new InvalidOperationException("Must call clear first in a storage write batch");
+                _hasSelfDestruct = true;
+            }
+        }
+
+        public void Dispose()
+        {
+            bool hasSet = (_wasSetCalled || _hasSelfDestruct);
```
Marchhill:
```suggestion
            bool hasSet = _wasSetCalled || _hasSelfDestruct;
```

src/Nethermind/Nethermind.State/TrieStoreScopeProvider.cs
```diff
@@ -0,0 +1,320 @@
...
+        }
+
+        public void MarkDirty(AddressAsKey address, Hash256 storageTreeRootHash)
+        {
+            _dirtyStorageTree.Enqueue((address, storageTreeRootHash));
+        }
+
+        public void Dispose()
+        {
+            while (_dirtyStorageTree.TryDequeue(out var entry))
```
Marchhill: use explicit types instead of `var`

src/Nethermind/Nethermind.State/StateProvider.cs
```diff
@@ -744,25 +679,18 @@ static void ThrowUnexpectedPosition(int currentPosition, int i, int forAssertion
...
 
-        private void FlushToTree()
+        internal void FlushToTree(IWorldStateScopeProvider.IWorldStateWriteBatch writeBatch)
         {
             int writes = 0;
             int skipped = 0;
 
-            using ArrayPoolListRef<PatriciaTree.BulkSetEntry> bulkWrite = new(_blockChanges.Count);
-            foreach (AddressAsKey key in _blockChanges.Keys)
+            foreach (var key in _blockChanges.Keys)
```
Marchhill: explicit type

src/Nethermind/Nethermind.Init/Steps/InitializeBlockchain.cs
```diff
@@ -18,6 +18,9 @@
 using Nethermind.Consensus.Scheduler;
 using Nethermind.Core;
 using Nethermind.Core.Attributes;
+using Nethermind.Evm;
```
Marchhill: all using needed?

src/Nethermind/Nethermind.State/PartialStorageProviderBase.cs
```diff
@@ -280,6 +262,31 @@ protected readonly struct Change(in StorageCell storageCell, byte[] value, Chang
...
+                Before = StorageTree.ZeroBytes;
+                IsInitialValue = true;
+            }
+
+            public byte[] Before;
+            public byte[] After;
+            public bool IsInitialValue;
+        }
+        */
+
```
benaadams: Clean up

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
```diff
@@ -437,28 +389,51 @@ public override void ClearStorage(Address address)
...
+                       {
+                           storage.Commit();
+                           storage.Return();
+                       }
+                   }
+
+                   Task.WaitAll(commitTask.AsSpan());
+
+           >>>>>>> origin/master
+                        */
```
benaadams: Clean up

src/Nethermind/Nethermind.State/PartialStorageProviderBase.cs
asdacap: Fixed

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
```diff
@@ -437,27 +389,18 @@ public override void ClearStorage(Address address)
     {
         base.ClearStorage(address);
 
-        // here it is important to make sure that we will not reuse the same tree when the contract is revived
-        // by means of CREATE 2 - notice that the cached trie may carry information about items that were not
-        // touched in this block, hence were not zeroed above
-        // TODO: how does it work with pruning?
-        _toUpdateRoots.Remove(address);
+        _toUpdateRoots.TryAdd(address, true);
```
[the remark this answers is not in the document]
asdacap: This fix invalid block in some case. I dont rememberr what case though. But I know it cause invalid block.

src/Nethermind/Nethermind.State/WorldState.cs
```diff
@@ -107,10 +84,34 @@ public Account GetAccount(Address address)
...
         bool IAccountStateProvider.TryGetAccount(Address address, out AccountStruct account)
         {
-            account = _stateProvider.GetAccount(address).ToStruct();
+            // Note: This call is for compatibility with `IAccountStateProvider` and should not be called directly by VM. Because its slower.
+            account = _stateProvider.GetAccount(address)
+                .WithChangedStorageRoot(_persistentStorageProvider.GetStorageRoot(address))
+                .ToStruct();
+
             return !account.IsTotallyEmpty;
         }
```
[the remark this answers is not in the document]
asdacap: It was on one of the goal of the interface where the Account.StorageRoot should not be dependent on. That said, later on I added account updated callback as it was too hard to deal with, which should mean that the Account.StorageRoot should always be in sync.

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
flcl42, line 307: Is it safe to use regular dict here?

asdacap: Apparently it is as the comment said.

PR #9092, opened 2025-08-06
Fix Taiko Anchor Transaction Validation

dipkakwani:
Fix existing NMC issue

## Changes
Fix checks for taiko anchor transaction validation

## Types of changes

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

Tested with Surge devnet locally.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


src/Nethermind/Nethermind.Taiko/TaikoTransactionProcessor.cs
```diff
@@ -22,6 +24,8 @@ public class TaikoTransactionProcessor(
     ILogManager? logManager
     ) : TransactionProcessorBase(specProvider, worldState, virtualMachine, codeInfoRepository, logManager)
 {
+    private readonly ILogger _logger = logManager?.GetClassLogger() ?? throw new ArgumentNullException(nameof(logManager));
```
flcl42: Cannot be null, no need in throw

dipkakwani: I think without throw it would require `ILogger? _logger` and subsequent calls to be `_logger?`
The existing code in many files seem to be following the same pattern : https://github.com/search?q=repo%3ANethermindEth%2Fnethermind%20logManager%3F.GetClassLogger()%20&type=code

flcl42: You can probably use Logger of the base class

src/Nethermind/Nethermind.Taiko/TaikoTransactionProcessor.cs
```diff
@@ -75,4 +79,32 @@ protected override TransactionResult IncrementNonce(Transaction tx, BlockHeader
...
+
+    protected override TransactionResult Execute(Transaction tx, ITxTracer tracer, ExecutionOptions opts)
+    {
+        // Recover sender address if missing, before calling base.Execute
+        // This ensures validation doesn't fail due to missing sender
+        if (tx.SenderAddress is null && tx.Signature is not null)
+        {
+            try
+            {
+                tx.SenderAddress = Ecdsa.RecoverAddress(tx);
```
flcl42: Is it revealed by a test case or during a real run? We most likely should already have it recovered

smartprogrammer93: only real run but this is not the best solution imo

PR #9096, opened 2025-08-07
Perf/Non blocking memory pruning

asdacap:
- Requires #9028 
- One of the blocker for nethermind to run very short slot time is the blocking memory prune where occationally nethermind will need to save the in-memory dirty nodes cache and prune unused nodes along with deleting obsolete nodes. 
  - This process locks the dirty nodes cache and prevent further block processing until it is finished.
  - On mainnet under default config, this happens about every 30 minute or so, and it take around 1 second. 
  - This is a problem for network with under 1 second slot time for weaker client as some block will get delayed.
- This PR adds a `CommitBuffer` where during memory pruning, further read and commit will operate on top of the buffer. 
- Multiple blocks can be added to the buffer up to a configurable limit `--Pruning.MaxBufferedCommit` set to 128 by default to prevent unexpected memory exhaustion.
- After memory pruning is complete, the `CommitBuffer` is flushed to the standard dirty nodes cache.

## Changes

- Move `TrieNode.LastSeen` to dirty nodes cache to not have issue with out of sync last seen number.
- Added `BeginScope` to `ITrieStore` to make it easy to switch between buffered or unbuffered mode.
 
#### What types of changes does your code introduce?,

- [X] Optimization

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Removed code for parallel pruning, resulting in very slow memory pruning. Confirmed via custom logs that a block processing is not blocked during prune cache.

### DIFF
--- a/src/Nethermind/Nethermind.Trie.Test/Pruning/TestPruningStrategy.cs
+++ b/src/Nethermind/Nethermind.Trie.Test/Pruning/TestPruningStrategy.cs
@@ -8,8 +8,8 @@ using Nethermind.Trie.Pruning;
 namespace Nethermind.Trie.Test.Pruning
 {
     public class TestPruningStrategy(
-        bool deleteObsoleteKeys,
         bool shouldPrune = false,
+        bool deleteObsoleteKeys = false,
         int? pruneInterval = null)
         : IPruningStrategy
     {
@@ -18,7 +18,6 @@ namespace Nethermind.Trie.Test.Pruning
         {
             if (pruneInterval is not null && state.LatestCommittedBlock % pruneInterval == 0)
             {
-                Console.Error.WriteLine($"Prune trigger");
                 return true;
             }
             return (ShouldPruneEnabled || WithMemoryLimit is not null && state.DirtyCacheMemory > WithMemoryLimit);
--- a/src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
+++ b/src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
@@ -459,11 +459,14 @@ public sealed class TrieStore : ITrieStore, IPruningTrieStore
 
         // Commit buffer mode would use the
         if (IsInCommitBufferMode)
+        {
             _commitBuffer.EnqueueCommitSet(set);
+        }
         else
+        {
             PushToMainCommitSetQueue(set);
-
-        Prune();
+            Prune();
+        }
     }
 
     private void PushToMainCommitSetQueue(BlockCommitSet set)
--- a/src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
+++ b/src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
@@ -439,13 +439,10 @@ public sealed class TrieStore : ITrieStore, IPruningTrieStore
     {
         if (_currentBlockCommitter is not null) throw new InvalidOperationException("Cannot start a new block commit when an existing one is still not closed");
 
+        if (_logger.IsDebug) _logger.Debug($"Beginning new {nameof(BlockCommitSet)} - {blockNumber}");
         VerifyNewCommitSet(blockNumber);
 
-        BlockCommitSet commitSet = IsInCommitBufferMode
-            ? _commitBuffer.CreateCommitSet(blockNumber)
-            : CreateCommitSet(blockNumber);
-
-        _lastCommitSet = commitSet;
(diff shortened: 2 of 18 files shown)


src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -494,13 +590,6 @@ private void PersistAndPruneDirtyCache()
     {
         try
         {
-            // Flush ahead of time so that memtable is empty which prevent stalling when writing nodes.
-            // Note, the WriteBufferSize * WriteBufferNumber need to be more than about 20% of pruning cache
-            // otherwise, it may not fit the whole dirty cache.
-            // Additionally, if (WriteBufferSize * (WriteBufferNumber - 1)) is already more than 20% of pruning
-            // cache, it is likely that there are enough space for it on most time, except for syncing maybe.
-            _nodeStorage.Flush(onlyWal: false);
```
asdacap, line 497: No longer necessary and make write amp worst if not full.

src/Nethermind/Nethermind.Trie.Test/Pruning/TestPruningStrategy.cs
```diff
@@ -8,8 +8,8 @@
 namespace Nethermind.Trie.Test.Pruning
 {
     public class TestPruningStrategy(
-        bool deleteObsoleteKeys,
```
asdacap, line 11: This was probably a mistake.

src/Nethermind/Nethermind.Blockchain/FullPruning/CopyTreeVisitor.cs
```diff
@@ -59,6 +59,7 @@ public void VisitTree(in TContext nodeContext, in ValueHash256 rootHash)
         [DoesNotReturn, StackTraceHidden]
         public void VisitMissingNode(in TContext ctx, in ValueHash256 nodeHash)
         {
+            Console.Error.WriteLine($"Full Pruning Failed: Missing node {nodeHash} at level {ctx.Storage}:{ctx.Path}.");
```
benaadams: Console rather than Logger?

src/Nethermind/Nethermind.Blockchain/FullPruning/CopyTreeVisitor.cs
```diff
@@ -108,6 +109,7 @@ public void Dispose()
 
         public void Finish()
         {
+            Console.Error.WriteLine("Finish called");
```
benaadams: Console rather than Logger?

src/Nethermind/Nethermind.Blockchain/FullPruning/FullPruner.cs
```diff
@@ -275,18 +270,15 @@ private async Task CopyTrie(IPruningContext pruning, Hash256 stateRoot, Cancella
 
                 if (!cancellationToken.IsCancellationRequested)
                 {
+                    Console.Error.WriteLine("Finishing");
```
benaadams: Console rather than Logger?

src/Nethermind/Nethermind.Blockchain/FullPruning/FullPruner.cs
```diff
@@ -312,6 +306,7 @@ CancellationToken cancellationToken
         {
             CopyTreeVisitor<TContext> copyTreeVisitor = new(targetNodeStorage, writeFlags, _logManager, cancellationToken);
             _stateReader.RunTreeVisitor(copyTreeVisitor, stateRoot, visitingOptions);
+            Console.Error.WriteLine("Visit tree finished");
```
benaadams: Console rather than Logger?

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -760,22 +851,25 @@ private void RecalculateTotalMemoryUsage()
...
+        FlushNonBlockingBuffer();
         PersistOnShutdown();
     }
 
+    private void FlushNonBlockingBuffer()
+    {
+        using var _ = _scopeLock.EnterScope();
+        if (_commitBuffer is null) return;
+
+        using var _2 = _pruningLock.EnterScope();
```
benaadams, line 863: 😅

asdacap:
Dont worry about console log, just debugging.

benaadams (approved):

kamilchodola (approved):

PR #9105, opened 2025-08-10
Lower allocation ByteArrayConverter for large hex strings

benaadams:
## Changes

- Read `ValueSequence` rather than copying to array to parse when input is multiple segments (usually large hex data e.g. blocks)
- ArrayPool normally allocates for very large arrays so skips that allocation path.

- [x] Optimization

#### Requires testing

- [x] Yes

- [x] Yes

### DIFF
--- a/src/Nethermind/Nethermind.Core/JsonConverters/ByteArrayConverter.cs
+++ b/src/Nethermind/Nethermind.Core/JsonConverters/ByteArrayConverter.cs
@@ -31,20 +31,18 @@ public class ByteArrayConverter : JsonConverter<byte[]>
         if (tokenType == JsonTokenType.None || tokenType == JsonTokenType.Null)
             return null;
         else if (tokenType != JsonTokenType.String && tokenType != JsonTokenType.PropertyName)
-        {
             ThrowInvalidOperationException();
-        }
 
         if (reader.HasValueSequence)
         {
             ReadOnlySequence<byte> valueSequence = reader.ValueSequence;
             return checked((int)valueSequence.Length) > 0 ?
-                ConvertValueSequence(in valueSequence) :
-                null;
+                ConvertValueSequence(in valueSequence) : null;
         }
 
         int length = reader.ValueSpan.Length;
         ReadOnlySpan<byte> hex = reader.ValueSpan;
+        if (hex.Length == 0) return null;
         if (length >= 2 && Unsafe.As<byte, ushort>(ref MemoryMarshal.GetReference(hex)) == _hexPrefix)
             hex = hex[2..];
 
@@ -77,7 +75,7 @@ public class ByteArrayConverter : JsonConverter<byte[]>
 
         // Compute total hex digit count (after prefix)
         long totalHexChars = seq.Length - (hadPrefix ? 2 : 0);
-        if (totalHexChars <= 0) return Array.Empty<byte>();
+        if (totalHexChars <= 0) return [];
 
         int odd = (int)(totalHexChars & 1);
         int outLenFinal = (int)(totalHexChars >> 1) + odd;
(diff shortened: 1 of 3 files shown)


src/Nethermind/Nethermind.Core/JsonConverters/ByteArrayConverter.cs
```diff
@@ -31,31 +31,120 @@ public class ByteArrayConverter : JsonConverter<byte[]>
...
+            if (remaining == 0) break;
+
+            // If remaining is even overall, remaining will be >= 2 here; be defensive just in case.
+            if (remaining == 1)
+            {
+                ThrowInvalidOperationException();
+            }
+
+            ushort twoNibbles = 0;
+            Span<byte> temp = MemoryMarshal.AsBytes(MemoryMarshal.CreateSpan(ref twoNibbles, 1));
```
asdacap: qq: Why not stackalloc?

src/Nethermind/Nethermind.Core/JsonConverters/ByteArrayConverter.cs
```diff
@@ -31,31 +31,120 @@ public class ByteArrayConverter : JsonConverter<byte[]>
...
+        int odd = (int)(totalHexChars & 1);
+        int outLenFinal = (int)(totalHexChars >> 1) + odd;
+        if (outLenFinal == 0) return [];
+
+        byte[] result = GC.AllocateUninitializedArray<byte>(outLenFinal);
+        Span<byte> output = result;
+        if (odd == 1)
+        {
+            // If odd, we deal with the extra nibble, so we are left with an even number of nibbles
+            if (!sr.TryRead(out byte lastNibble))
```
asdacap: np: Not sure if a mistake, but `last` or `first`?

src/Nethermind/Nethermind.Core/JsonConverters/ByteArrayConverter.cs
benaadams: inside a loop; so risk of stack overflow. Could move it out and reuse the span though

src/Nethermind/Nethermind.Core/JsonConverters/ByteArrayConverter.cs
```diff
@@ -31,31 +31,120 @@ public class ByteArrayConverter : JsonConverter<byte[]>
...
+
+            // If remaining is even overall, remaining will be >= 2 here; be defensive just in case.
+            if (remaining == 1)
+            {
+                ThrowInvalidOperationException();
+            }
+
+            ushort twoNibbles = 0;
+            Span<byte> temp = MemoryMarshal.AsBytes(MemoryMarshal.CreateSpan(ref twoNibbles, 1));
+            if (!sr.TryCopyTo(temp))
```
asdacap: I'm confused, the line on 114 is suppose to take even number of character from current span. This mean that at this point, there should be exactly 1 character in current span? But you then take two character? Which is still correct since the whole thing in this loop is even. But isnt the point is to operate on current span only?

src/Nethermind/Nethermind.Core/JsonConverters/ByteArrayConverter.cs
benaadams: first, was earlier concept that I forgot to rename

src/Nethermind/Nethermind.Core/JsonConverters/ByteArrayConverter.cs
benaadams: so this is if the chunk is an odd number of hex chars `0x123`, then I have to reach into next chunk to get `34`, using the SequenceReader.TryCopyTo to bridge the spans

benaadams: Clarified

src/Nethermind/Nethermind.Core/JsonConverters/ByteArrayConverter.cs
benaadams: Fixed

src/Nethermind/Nethermind.Core/JsonConverters/ByteArrayConverter.cs
benaadams: Fixed

asdacap (approved):

PR #9131, opened 2025-08-14
Configurable Consensus HealthChecks

svlachakis:
## Changes

Configurable Consensus HealthChecks for Arbitrum Plugin 

Example usage: 

`
--HealthChecks.EnableConsensusLayerHealthChecks=false
`
Default : `true`

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_


src/Nethermind/Nethermind.HealthChecks/IHealthChecksConfig.cs
```diff
@@ -88,6 +88,9 @@ public interface IHealthChecksConfig : IConfig
     [ConfigItem(Description = "The max request interval, in seconds, in which the consensus client is assumed healthy.", DefaultValue = "300")]
     public int MaxIntervalClRequestTime { get; set; }
 
+    [ConfigItem(Description = "Whether to disable consensus layer health checks. Set to true for L2s that don't use consensus clients.", DefaultValue = "false")]
```
rubo:
```suggestion
    [ConfigItem(Description = "Whether to disable consensus layer health checks. Set to `true` for L2s that don't use consensus clients.", DefaultValue = "false")]
```

rubo (approved):

src/Nethermind/Nethermind.HealthChecks/IHealthChecksConfig.cs
svlachakis: done, thanks!

src/Nethermind/Nethermind.HealthChecks/IHealthChecksConfig.cs
```diff
@@ -88,6 +88,9 @@ public interface IHealthChecksConfig : IConfig
     [ConfigItem(Description = "The max request interval, in seconds, in which the consensus client is assumed healthy.", DefaultValue = "300")]
     public int MaxIntervalClRequestTime { get; set; }
 
+    [ConfigItem(Description = "Whether to disable consensus layer health checks. Set to `true` for L2s that don't use consensus clients.", DefaultValue = "false")]
+    public bool DisableConsensusLayerHealthChecks { get; set; }
```
damian-orzechowski: I'm not convinced to the convention here - I think `Enabled` would fit much better to the overall scheme used by Nethermind configuration.

damian-orzechowski: Also, maybe worth considering if not better to use the `MaxIntervalClRequestTime` to control functionality on/off - similar to `LowStorageSpaceWarningThreshold`

src/Nethermind/Nethermind.HealthChecks/ClHealthRequestsTracker.cs
```diff
@@ -25,8 +26,11 @@ public class ClHealthRequestsTracker(ITimestamper timestamper, IHealthChecksConf
 
     public Task StartAsync()
     {
-        _timer = new Timer(ReportClStatus, null, TimeSpan.Zero,
-            TimeSpan.FromSeconds(ClUnavailableReportMessageDelay));
+        if (!_disableConsensusLayerHealthChecks)
```
wurdum: I'd suggest moving `healthChecksConfig.DisableConsensusLayerHealthChecks` check from `StartAsync` to the caller side at `HealthChecksPlugin` (similarly to `_mergeConfig.Enabled`). That way, the config flag will work with any implementation of `IEngineRequestsTracker`.

src/Nethermind/Nethermind.HealthChecks/IHealthChecksConfig.cs
svlachakis: @damian-orzechowski You're absolutely right about following Nethermind's existing conventions. I actually started with `Enabled` initially in this commit https://github.com/NethermindEth/nethermind/pull/9131/commits/4bd160ded4d2b4eb28f72261d884d3c59f94e7e4, then switched to `Disabled` because I was not sure which behaviour is more clear as an override.

Regarding using `MaxIntervalClRequestTime` for on/off control like `LowStorageSpaceWarningThreshold` - that's an interesting approach which I also evaluated. My concern was that setting a timeout value to control enable/disable behaviour might be less explicit than a dedicated boolean flag. With a timeout, it's not immediately clear whether 0 means "disabled" or "no timeout limit" but it has the benefit of less code changes.

I think switching back to `Enabled` is the best option since it follows better Nethermind's existing conventions.

src/Nethermind/Nethermind.HealthChecks/ClHealthRequestsTracker.cs
svlachakis: @wurdum that's a good point, done

src/Nethermind/Nethermind.HealthChecks/ClHealthRequestsTracker.cs
```diff
@@ -54,6 +54,9 @@ private bool IsRequestTooOld(DateTime now, DateTime requestTime)
 
     public bool CheckClAlive()
     {
+        if (!_enableConsensusLayerHealthChecks)
```
wurdum: Sorry, I've missed this one. My bad. I missed that `NodeHealthService` also calls `CheckClAlive`, and there is no easy way to isolate `EnableConsensusLayerHealthChecks` outside of `ClHealthRequestsTracker` without complete refactoring.

It's better to have all of the uses of `EnableConsensusLayerHealthChecks` in one place even if it sits in `StartAsync`, than spreading it across the abstractions. Would you mind rolling back [the change](https://github.com/NethermindEth/nethermind/pull/9131/commits/4bd160ded4d2b4eb28f72261d884d3c59f94e7e4)? 😅

svlachakis: I thought you had seen that :D I can rollback yeah, we still need that there because it's called from other places also.

damian-orzechowski:
> Maybe better would be to swap the whole component for one that would do no-op?

In that case we'd probably have to separate functionalities - although we'll not use CL checks (as not engine api is used), but we could still benefit from other checks e.g. database size

LukaszRozmej:
> > Maybe better would be to swap the whole component for one that would do no-op?
> 
> In that case we'd probably have to separate functionalities - although we'll not use CL checks (as not engine api is used), but we could still benefit from other checks e.g. database size

but you should have interface segragation and maybe have - `IClHealthTracker`?

PR #9146, opened 2025-08-16
feat: add ProgressReporter abstraction

### LINKED ISSUE #8504: More use of ProgressLogger.
- There is a `ProgressLogger` class for logging progress. 
- It is quite nice.
- Some part of the codebase does not use it.
  - Full pruning.
  - Not sure, check it out.
- Make them use it.

berZKerk:
Closes #8504.

## Changes

Introduces `Nethermind.Core.ProgressReporter` — an `IDisposable` that wraps a `ProgressLogger` with a 1-second timer for periodic `LogProgress` and guarantees `MarkEnd` on disposal. Replaces the recurring `Reset` → `new Timer(...)` → `try/finally MarkEnd` pattern with one `using` line.

Applied to four sites that were either ad-hoc or had no progress logging:

- `TotalDifficultyFixMigration` — previously had no progress logging.
- `DbBlocksLoader`, `StartupBlockTreeFixer`, `ReceiptsVerificationVisitor` — three block-tree visitors that emitted ad-hoc modulo-N `Info` strings; now standardised on `ProgressLogger`'s percentage + bar + speed format.

Side effects:

- `DbBlocksLoader` and `StartupBlockTreeFixer` constructor signatures change from `ILogger` to `ILogManager` (required to construct `ProgressLogger`). Callers in `ReviewBlockTree`, `Nethermind.PerfTest`, and tests updated.
- `ReceiptsVerificationVisitor` becomes `IDisposable`; `ReceiptFixMigration` updated to `using` it.
- Tests that previously typed `LimboNoErrorLogger.Instance` as `ILogger` switched to the equivalent `ILogManager` — `NoErrorLimboLogs.Instance` (preserves the throw-on-error fail-fast guard; `LimboLogs.Instance` silently drops errors and would have weakened the invariant).
- Closed pre-existing lifetime leaks: visitor instances in `DbBlocksLoaderTests`, `StartupTreeFixerTests`, and `Nethermind.PerfTest` weren't `using`-disposed despite the classes already being `IDisposable`.

## Scope note

The Full Pruning, Trie Verification, and Analytics work originally proposed in this PR was superseded on master:

- `Nethermind.Analytics` was removed in #9683.
- `CopyTreeVisitor` and `TrieStatsCollector` now use `VisitorProgressTracker`, a path-based estimator that produces a percentage even when total node count is unknown — strictly more useful than wiring `ProgressLogger` directly. `BlockingVerifyTrie` reports progress transitively through `TrieStatsCollector`.

## Types of changes

- [x] Refactoring

## Testing

- [x] No new tests required

Progress logging is observability-only. Behavioural coverage stays with the existing `TotalDifficultyFixMigrationTest`, `DbBlocksLoaderTests`, and `StartupTreeFixerTests`. Full `Nethermind.slnx` builds clean under `EnforceCodeStyleInBuild=true`.


PROGRESS_LOGGER_LOG_SAMPLES.md
```diff
@@ -0,0 +1,109 @@
+# ProgressLogger Log Samples
```
asdacap: I mean, put it in the pr descrpition, not committing it.

src/Nethermind/Nethermind.Analytics/SupplyVerifier.cs
```diff
@@ -46,6 +53,10 @@ public bool ShouldVisit(in OldStyleTrieVisitContext _, in ValueHash256 nextNode)
 
         public void VisitTree(in OldStyleTrieVisitContext _, in ValueHash256 rootHash)
         {
+            if (_progressLogger is not null)
```
asdacap: Dont make it optional then. In test, there is LimboLog.

src/Nethermind/Nethermind.Blockchain/FullPruning/CopyTreeVisitor.cs
```diff
@@ -82,20 +109,59 @@ private void PersistNode(Hash256 storage, in TreePath path, TrieNode node)
...
+        private long EstimateNodesInState(ValueHash256 rootHash)
+        {
+            // This is a simplified estimation - in practice, you might want to:
+            // 1. Do a quick scan of the trie to count nodes
+            // 2. Use historical data from previous pruning operations
+            // 3. Use block number to estimate state size
+
+            // For now, we'll use a conservative estimate based on typical state sizes
+            // This can be improved with more sophisticated estimation logic
+            return 10_000_000; // Conservative estimate for mainnet state
```
asdacap: Do it properly, or don't at all.

asdacap (changes requested):

PROGRESS_LOGGER_LOG_SAMPLES.md
berZKerk: I've committed it for the sake of you to review it and tell what you think, and after I was gonna delete it

berZKerk: But now gonna do as you wish

flcl42:
@asdacap @LukaszRozmej do we still consider merging it?

LukaszRozmej:
@claude review

LukaszRozmej:
@claude  re-review

src/Nethermind/Nethermind.Init/Steps/Migrations/TotalDifficultyFixMigration.cs
```diff
@@ -41,40 +44,55 @@ public Task Run(CancellationToken cancellationToken)
...
 
         if (_logger.IsInfo) _logger.Info($"Starting TotalDifficultyFixMigration. From block {startingBlock} to block {lastBlock}");
 
-        for (long blockNumber = startingBlock; blockNumber <= lastBlock; ++blockNumber)
-        {
-            cancellationToken.ThrowIfCancellationRequested();
+        _progressLogger.Reset(0, lastBlock.Value - startingBlock + 1);
 
-            ChainLevelInfo currentLevel = _chainLevelInfoRepository.LoadLevel(blockNumber)!;
+        using Timer timer = new(1000) { Enabled = true };
```
LukaszRozmej: Nit: match `ReceiptMigration.cs`'s pattern — subscribe to `Elapsed` first, then enable the timer. Avoids the (theoretical) race where a tick could fire before the handler is attached.

```suggestion
        using Timer timer = new(1000);
        timer.Elapsed += (_, _) => _progressLogger.LogProgress();
        timer.Enabled = true;
```

src/Nethermind/Nethermind.Init/Steps/Migrations/TotalDifficultyFixMigration.cs
```diff
@@ -41,40 +44,55 @@ public Task Run(CancellationToken cancellationToken)
...
+                if (shouldPersist)
+                {
+                    _chainLevelInfoRepository.PersistLevel(blockNumber, currentLevel);
+                }
+
+                _progressLogger.Update(blockNumber - startingBlock + 1);
             }
         }
+        finally
+        {
```
LukaszRozmej: Nit: also call `timer.Stop();` here, matching `ReceiptMigration.RunMigration`'s `finally`. `using` will dispose the timer (which stops it), so this is functionally equivalent — but the explicit `Stop()` mirrors the established pattern.

```suggestion
        finally
        {
            _progressLogger.MarkEnd();
            timer.Stop();
        }
```

LukaszRozmej (commented):
Two small nits to align with the existing `ReceiptMigration` pattern in the same folder.

src/Nethermind/Nethermind.Blockchain.Test/Visitors/DbBlocksLoaderTests.cs
```diff
@@ -61,7 +61,7 @@ public async Task Can_load_blocks_from_db()
                 .WithSpecProvider(OlympicSpecProvider.Instance)
                 .TestObject;
 
-            DbBlocksLoader loader = new(blockTree, LimboNoErrorLogger.Instance);
+            DbBlocksLoader loader = new(blockTree, LimboLogs.Instance);
```
flcl42: Not disposed and below

LukaszRozmej (approved):

LukaszRozmej:
@claude review

LukaszRozmej:
@claude re-review

LukaszRozmej:
@claude re-review

svlachakis:
@claude check the validity of the following review

  🟠 WARNING — two real issues, both rooted in the same gap
  
  W1 — Dispose() doesn't drain in-flight timer callbacks; final "done" log can be overwritten by a stale tick

  ProgressReporter.cs:
  _timer.Stop();
  _timer.Dispose();
  _progressLogger.MarkEnd();      // sets UtcEndTime
  _progressLogger.LogProgress();  // emits "done" line
  
  System.Timers.Timer.Stop() is documented as not draining in-flight Elapsed callbacks — a tick already queued on the thread pool can land
  after Stop() returns. That tick calls _progressLogger.LogProgress(), which:
  - reads UtcEndTime (now set by MarkEnd()),
  - resets LastMeasurement / LastValue via SetMeasuringPoint(),
  - writes _lastReportState. 
  
  Two observable bad outcomes:
  1. The final "100% done" line is followed by — and effectively replaced in operator-visible state by — a stale partial-progress line.
  2. The speed counters in the "done" line get recomputed against post-reset state.
  
  Fix: use Timer.Dispose(WaitHandle) and block on it:
  using ManualResetEvent drained = new(false);
  _timer.Dispose(drained);
  drained.WaitOne();
  _progressLogger.MarkEnd();
  _progressLogger.LogProgress();
  Or move to PeriodicTimer which is await-able and drains naturally.
  
  W2 — ProgressLogger field state is unsynchronized; Update() and timer-fired LogProgress() race

  ProgressLogger.Update(long) writes CurrentValue, UtcStartTime, UtcEndTime. LogProgress() reads those and _lastReportState. None of it is
  locked. In ProgressReporter, the work thread calls Update() while the 1 Hz timer callback calls LogProgress() on a thread-pool thread —
  they truly run concurrently.

  Concrete torn-state path: Update() sets UtcEndTime = null first; a tick between that and the CurrentValue = value assignment computes the
  speed line against the old CurrentValue with the new "still running" flag. _lastReportState is a 4-long tuple written non-atomically — a
  half-written tuple either suppresses (dedup fires spuriously) or emits a duplicate line.

  This is technically pre-existing in ProgressLogger, but the old ProgressTracker used Interlocked.Add + AutoReset=false which kept the timer
   single-shot per tick. The new ProgressReporter introduces true concurrency where none existed before — at all 8 call sites.
  
  Fix: a lock in ProgressReporter around both update and log paths is sufficient (uncontended at 1 Hz):
  private readonly object _lock = new();
  public void Update(long value) { lock (_lock) _progressLogger.Update(value); }
  private void OnElapsed(object? _, ElapsedEventArgs __) { lock (_lock) _progressLogger.LogProgress(); }
  Combined with W1's drain, the Dispose path also takes the lock before MarkEnd() + LogProgress().

  ---
  🟡 SUGGESTION
  
  S1 — Functional regression on unknown-size snapshot downloads
  SnapshotDownloader.cs passes totalSize ?? 0 to the reporter. ProgressLogger.DefaultFormatter() does current / (float)Math.Max(total, 1), so
   when Content-Length is absent (chunked transfer is common) the displayed percentage always pegs at 100%. The old ProgressTracker had an
  explicit "unknown total" branch that showed bytes-without-percentage. Outside the race/leak/allocation scope strictly, but it's a real
  operator-facing regression worth flagging since the diff touches this code path.

svlachakis (approved):

LukaszRozmej:
@claude re-review

flcl42 (approved):

asdacap (approved):

PR #9156, opened 2025-08-18
Reduce log noise from bad behaving peers

benaadams:
## Changes

- Reduce log noise from bad behaving peers

- [x] Other: Logging

#### Requires testing

- [x] No

## Remarks

Still need to be improved

### DIFF
--- a/src/Nethermind/Nethermind.Synchronization/Blocks/BlockDownloader.cs
+++ b/src/Nethermind/Nethermind.Synchronization/Blocks/BlockDownloader.cs
@@ -128,7 +128,7 @@ namespace Nethermind.Synchronization.Blocks
             }
             catch (Exception ex)
             {
-                if (_logger.IsError) _logger.Error($"Unhandled exception in {nameof(BlockDownloader)}: {ex}");
+                if (_logger.IsDebug) _logger.Error($"DEBUG/ERROR Unhandled exception in {nameof(BlockDownloader)}: {ex}");
 #if DEBUG
                 throw;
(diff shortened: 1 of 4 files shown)


src/Nethermind/Nethermind.Synchronization/Blocks/BlockDownloader.cs
```diff
@@ -128,7 +128,7 @@ private void BlockTreeOnNewHeadBlock(object? sender, BlockEventArgs e)
             }
             catch (Exception ex)
             {
-                if (_logger.IsError) _logger.Error($"Unhandled exception in {nameof(BlockDownloader)}: {ex}");
+                if (_logger.IsDebug) _logger.Error($"DEBUG/ERROR Unhandled exception in {nameof(BlockDownloader)}: {ex}");
```
alexb5dh, line 131: Is this a typo?

benaadams: What do you mean?

alexb5dh: `if (_logger.IsDebug) _logger.Error`

benaadams: Is so it will still print as errors for us when using debug logging; but not show as errors to users; which is what the "DEBUG/ERROR" at start of string represents so easy to filter

PR #9160, opened 2025-08-18
fix(discovery): add eth forkid to self ENR via IForkInfo

Galoretka:
Reason: The ENR should include the Ethereum fork ID (per devp2p eth enr entry) so peers can detect compatibility and avoid connecting across incompatible forks. This TODO was missing, reducing interoperability.

Goal: Advertise the correct fork hash and next fork block in ENR to improve peer discovery accuracy and protocol compliance.

Changes:
- Injected IForkInfo and IBlockTree into NodeRecordProvider.
- Computed the current fork id from the best header (fallbacks to genesis/known number).
- Added EthEntry with forkHash and nextBlock before signing the ENR.


flcl42 (changes requested):
When node starts head may be at 0, after the sync it becomes the current chain head, and forkid will be different, but it will not be reflected in the enr. For this functionality to really work, it's required the record to be updated, by checking forkid on every new head. Enr sequence should be incremented on forkid change also. We don't change enr seq for now, but if we change we need to preserve the counter between restarts or use something timestamp based.

Could you also check the failing tests?

stdevMac:
Closing: superseded by #12126 ("Handle fork ID and seq in ENRs"), which implements this feature along with ENR rebuild/re-signing on fork ID change and monotonic enrSeq handling. Thanks for the contribution.

PR #9163, opened 2025-08-18
Support OP Isthmus at the Genesis block

### LINKED ISSUE #9145: Optimism Isthmus - Fail to start due to genesis mismatch
**Description**
When the OP fork Isthmus is enabled at genesis, Nethermind fails to launch due to a mismatched genesis hash.

**Steps to Reproduce**
1. Run Nethermind using the Kurtosis [optimism-package](https://github.com/ethpandaops/optimism-package)
2. Ensure both Holocene and Isthmus forks are enabled at genesis
3. Wait until Nethermind starts

**Actual behavior**
Nethermind computes a different genesis hash than the expected one, resulting in a crash.

**Expected behavior**
Nethermind should compute the correct genesis hash.

 - Operating System: Fedora Workstation 42
 - Version: commit 0bd563000c42289ebcead9b374eda80cd758746c
 - Installation Method: from source
 - Consensus Client: `op-node 1.13.4`

**Additional context**
The following yaml file was used for setting up Kurtosis:

```yaml
optimism_package:
  chains:
    chain0:
      participants:
        node0:
          el:
            type: op-geth
        node1:
          el:
            type: op-nethermind
            image: nethermind-local:latest
          cl:
            type: op-node
        node2:
          el:
            type: op-geth
          cl:
            type: op-node
      network_params:
        holocene_time_offset: 0
        isthmus_time_offset: 0
```

**Logs**
```
Processed block 0 (0xa77b69...b655a8) is invalid: 
  15 Aug 15:42:15 | - hash: expected 0x01a7538ba5c3c7ff4497509927cd1814ffa16438f309cbce2a863e87efea8385, got 0xa77b691e5ab9bf64ac6b67cb9f05e47a18aa1e8713427b69e5d1d4cd01b655a8 
  15 Aug 15:42:15 | - block extra data : 00000000fa00000006, UTF8: � 
  15 Aug 15:42:15 | Rejected invalid block 0 (0x01a7538ba5c3c7ff4497509927cd1814ffa16438f309cbce2a863e87efea8385),  ExtraData: 00000000fa00000006, reason: invalid block after processing
```

emlautarom1:
Fixes #9145

## Changes

- Post process the Genesis block when OP Isthmus is enabled at Genesis.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

Added unit tests to ensure that the Genesis post processing is executed when configured.
Tested with an [optimism-package Kurtosis setup](https://github.com/emlautarom1/optimism-package) using the following params:

```yaml
optimism_package:
  chains:
    chain0:
      participants:
        node0:
          el:
            type: op-geth
        node1:
          el:
            type: op-nethermind
            image: nethermind-local:latest
          cl:
            type: op-node
      network_params:
        holocene_time_offset: 0
        isthmus_time_offset: 0
```
where `nethermind-local:latest` is a Docker image built from this PR.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks

This issue only shows up when Isthmus is enabled at the genesis block which is not the case for any of the chains in the Superchain Registry at the time. Nevertheless, in testing environments like Kurtosis this can happen, and it results in Nethermind being unable to start up due to a genesis block hash mismatch when compared against `op-geth`.


src/Nethermind/Nethermind.Core.Test/Modules/AutoMainProcessingContext.cs
```diff
@@ -44,6 +44,7 @@ public AutoMainProcessingContext(
                     StoreReceiptsByDefault = receiptConfig.StoreReceipts,
                     DumpOptions = initConfig.AutoDump
                 })
+                .AddSingleton<IGenesisPostProcessor>(new NullGenesisPostProcessor())
```
asdacap: Should be unnecessary.

src/Nethermind/Nethermind.Core.Test/Blockchain/TestBlockchain.cs
```diff
@@ -277,6 +278,7 @@ protected virtual ContainerBuilder ConfigureContainer(ContainerBuilder builder,
             .AddSingleton<ISealValidator>(Always.Valid)
             .AddSingleton<IUnclesValidator>(Always.Valid)
             .AddSingleton<ISealer>(new NethDevSealEngine(TestItem.AddressD))
+            .AddSingleton<IGenesisPostProcessor>(new NullGenesisPostProcessor())
```
asdacap: Should be in `PseudoNethermindModule`.

asdacap: Or `NethermindModule` if you want to put in prod.

src/Nethermind/Nethermind.Core.Test/Modules/AutoMainProcessingContext.cs
emlautarom1: Unfortunately it is since otherwise tests fail.

Edit: Once the dependency is resolved by the `PseudoNethermindModule` this is no longer required. Removed.

src/Nethermind/Nethermind.Core.Test/Blockchain/TestBlockchain.cs
emlautarom1: Since `GenesisLoader` is constructed manually instead of using DI there is no point in having this dependency in `NethermindModule`, at least for now.

Moved the injection of `IGenesisPostProcessor` to `PseudoNethermindModule` to be used in tests.

src/Nethermind/Nethermind.Api/IApiWithBlockchain.cs
```diff
@@ -26,6 +26,7 @@ public interface IApiWithBlockchain : IApiWithStores, IBlockchainBridgeFactory
         (IApiWithBlockchain GetFromApi, IApiWithBlockchain SetInApi) ForProducer => (this, this);
 
         CompositeBlockPreprocessorStep BlockPreprocessor { get; }
+        IGenesisPostProcessor GenesisPostProcessor { get; set; }
```
LukaszRozmej, line 29: Why not have just custom genesis loader?

emlautarom1: That was actually my first idea: make a `IGenesisLoader`, and have the custom implementation for OP be a decorator to avoid code duplication. 

The "issue" with this approach is in the order of steps:

1. Get genesis from chainspec
2. Preallocate accounts
3. Update state root
4. < here we need to do the OP post processing >
5. Compute the header hash
6. Verify the header hash with an expected hash, if any

To get this we need to refactor the genesis loader into separate classes each with their own responsibilities so we can inject the desired behaviour at the right time. Also, since `GenesisLoader` is constructed directly instead of through DI we would need to make some additional refactors to support DI.

I decided on limiting the scope of this PR to a bugfix and let refactoring for future PRs if considered appropriate.

PR #9167, opened 2025-08-20
Remove hardcoded AuRa release spec decorator to plugin module

asdacap:
- Move hardcoded AuRa release spec decorator to plugin.

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Gnosis running normally.

### DIFF
--- a/src/Nethermind/Nethermind.Core/Specs/SpecProviderDecorator.cs
+++ b/src/Nethermind/Nethermind.Core/Specs/SpecProviderDecorator.cs
@@ -1,0 +1,31 @@
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using Nethermind.Int256;
+
+namespace Nethermind.Core.Specs;
+
+public class SpecProviderDecorator(ISpecProvider baseSpecProvider) : ISpecProvider
+{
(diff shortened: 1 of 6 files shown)


src/Nethermind/Nethermind.Core/Specs/SpecProviderDecorator.cs
```diff
@@ -0,0 +1,31 @@
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using Nethermind.Int256;
+
+namespace Nethermind.Core.Specs;
+
+public class SpecProviderDecorator(ISpecProvider baseSpecProvider) : ISpecProvider
```
LukaszRozmej, line 8: Is there a point for this class existence? We have only one usage and would have to make other things virtual if needed. Seems not needed to me.

LukaszRozmej (approved):

src/Nethermind/Nethermind.Core/Specs/SpecProviderDecorator.cs
asdacap: The thing in arbitrum is better served with this as base class instead of overriding ChainSpecBasedSpecProvider. So I figure its better to have it.

PR #9171, opened 2025-08-20
Extra validation in case `IForwardHeaderProvider` skip header.

asdacap:
- Add extra validation in case `IForwardHeaderProvider` return non consecutive block headers.

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- [ ] Mainnet can sync normally


benaadams (approved):
Test formatting

PR #9180, opened 2025-08-20
Re-create empty account

damian-orzechowski:
## Changes

- Allows to re-create selected empty account directly at `IWorldState` level.

Update:
Tested multiple approach to achieve this functionality, but unfortunately nothing works without modifying directly `StateProvider`. The goal is to be able to re-create an empty account, which normally is not allowed to be created with eip-158 enable.
- Disabling eip-158 doesn't work as it then applies to all accounts, but it only needs to be applied to selected accounts (determined during transaction processing)
- A similar approach is currently done in `SystemTransactionReleaseSpec` where a selected account is exempt from eip-158, but in this case, none of the accounts is known up-front
- Tried implementing a decorator over `IWorldState` which reimplements eip-158 with an option to exempt selected accounts (on per block basis), which partially works. However, disabling eip-158 also modifies `VirtualMachine`, so it would also have to be re-implemented to achieve the goal (see `ClearEmptyAccountWhenTouched`).

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


LukaszRozmej:
Maybe use same pattern we do with SystemUser for AuRa where it is just exempted from Eip158 deletion?

src/Nethermind/Nethermind.State/StateProvider.cs
```diff
@@ -858,6 +892,12 @@ private void PushNew(Address address, Account account)
             stack.Push(_changes.Count);
             _changes.Add(new Change(address, account, ChangeType.New));
         }
+        private void PushRecreateEmpty(Address address, Account account)
+        {
+            Stack<int> stack = SetupCache(address);
```
LukaszRozmej: You can probably skip this line and pass the stack from `CreateEmptyAccountIfDeletedOrNew` or even inline or make this a sub-method

src/Nethermind/Nethermind.State/StateProvider.cs
```diff
@@ -601,6 +624,17 @@ public void Commit(IReleaseSpec releaseSpec, IWorldStateTracer stateTracer, bool
                                 trace?.AddToTrace(change.Address, change.Account);
                             }
 
+                            break;
+                        }
+                    case ChangeType.RecreateEmpty:
+                        {
+                            if (change.Account.IsEmpty)
```
LukaszRozmej: not sure it isn't double checking?

damian-orzechowski: yes, this is double checking

PR #9191, opened 2025-08-23
Optimize Blooms

benaadams:
## Changes

- Replaced byte-level bit operations with 64-bit word operations using masks
- Optimized bloom extract calculation by processing multiple indexes in a single operation
- Shrink BloomExtract to a single long and then removing unnecessary `in` parameters

- [x] Optimization

#### Requires testing

- [x] Yes

- [x] Yes

### DIFF
--- a/src/Nethermind/Nethermind.Core/Bloom.cs
+++ b/src/Nethermind/Nethermind.Core/Bloom.cs
@@ -56,7 +56,7 @@ public class Bloom : IEquatable<Bloom>
     public Span<byte> Bytes => _bloomData.AsSpan();
     private Span<ulong> ULongs => _bloomData.AsULongs();
 
-    private void Set(ReadOnlySpan<byte> sequence, Bloom? masterBloom = null)
+    public void Set(ReadOnlySpan<byte> sequence, Bloom? masterBloom = null)
     {
         if (ReferenceEquals(this, Empty))
         {


src/Nethermind/Nethermind.Core/Bloom.cs
```diff
@@ -2,332 +2,381 @@
...
+        if (w2 == w1) { m1 |= m2; w2 = NotRequired; }
+        if (w3 == w1) { m1 |= m3; w3 = NotRequired; }
+        else if (w3 == w2) { m2 |= m3; w3 = NotRequired; }
 
-        public override string ToString()
-        {
-            return Bytes.ToHexString();
-        }
+        // Write to this bloom using 64-bit words.
+        SetWordMasks(w1, m1, w2, m2, w3, m3);
```
LukaszRozmej: Duplicated code

src/Nethermind/Nethermind.Core/Bloom.cs
```diff
@@ -2,332 +2,381 @@
...
-            if (ReferenceEquals(this, other)) return true;
+        // Merge masks if same word to minimize stores.
+        if (w2 == w1) { m1 |= m2; w2 = NotRequired; }
+        if (w3 == w1) { m1 |= m3; w3 = NotRequired; }
+        else if (w3 == w2) { m2 |= m3; w3 = NotRequired; }
 
-            return Nethermind.Core.Extensions.Bytes.AreEqual(Bytes, other.Bytes);
-        }
+        // Write to this bloom using 64-bit words.
+        SetWordMasks(w1, m1, w2, m2, w3, m3);
```
LukaszRozmej: Duplicated code

src/Nethermind/Nethermind.Core/Bloom.cs
```diff
@@ -2,332 +2,381 @@
...
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    private static (int w, ulong m) WordMask(int bitIndex)
+    {
+        const int LittleEndianBitReverseMask = 7;   // For little-endian, reverse lower 3 bits
+        const int BigEndianBitReverseMask = 63;     // For big-endian, reverse lower 6 bits
+
+        int w = bitIndex >> 6;
+        int bi = bitIndex & 63;
+        // The following bit reversal logic ensures correct mapping of bits in the bloom filter
+        int shift = BitConverter.IsLittleEndian ? (bi ^ LittleEndianBitReverseMask) : (bi ^ BigEndianBitReverseMask);
```
LukaszRozmej: unnecessary bit operation? Wouldn't be better to put the if before the `>>` and `&` so only one is performed?

src/Nethermind/Nethermind.Core/Bloom.cs
```diff
@@ -2,332 +2,381 @@
...
 
-        public bool Matches(ReadOnlySpan<byte> sequence)
-        {
-            BloomExtract indexes = GetExtract(sequence);
-            return Matches(in indexes);
-        }
+        // Merge masks if same word to minimize stores.
+        if (w2 == w1) { m1 |= m2; w2 = NotRequired; }
+        if (w3 == w1) { m1 |= m3; w3 = NotRequired; }
+        else if (w3 == w2) { m2 |= m3; w3 = NotRequired; }
```
LukaszRozmej: So each time we get 3 random 1..32 indexes. There is 0,1% chance that any two of them will be the same (so overall 0,3% chance to get at least one pair) and 0,003% chance that all of them will be the same. Is it worth the branching?

src/Nethermind/Nethermind.Core/Bloom.cs
benaadams: Removed

src/Nethermind/Nethermind.Core/Bloom.cs
benaadams: Removed

src/Nethermind/Nethermind.Core/Bloom.cs
benaadams: Removed

src/Nethermind/Nethermind.Core/Bloom.cs
benaadams: Cleaned it up a little; but is needed; the alternative subtraction which is slower than xor

src/Nethermind/Nethermind.Core/Bloom.cs
LukaszRozmej: hm... my math were wrong the chances are 3,125% (9,375% for any pair) and 0,1%

src/Nethermind/Nethermind.Core/Bloom.cs
```diff
@@ -2,332 +2,350 @@
...
-        public bool Matches(in BloomExtract extract) => Get(extract.Index1) && Get(extract.Index2) && Get(extract.Index3);
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    private static (int wordIndex, ulong mask) WordMask(int bitIndex)
+    {
+        const int LittleEndianReverseMask = 7;   // For little-endian, reverse lower 3 bits
+        const int BigEndianReverseMask = 63;     // For big-endian, reverse lower 6 bits
+
+        int wordIndex = bitIndex >> 6;
+        bitIndex &= 63;
+        int mirrorMask = BitConverter.IsLittleEndian ? LittleEndianReverseMask : BigEndianReverseMask;
```
LukaszRozmej, line 209: this could be saved in static?

benaadams: `BitConverter.IsLittleEndian` is compiler const; so it becomes a constant

PR #9195, opened 2025-08-24
Optimize TrieNode.SeekChild

benaadams:
## Changes

- _List the changes_

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Serialization.Rlp/ValueRlpStream.cs
+++ b/src/Nethermind/Nethermind.Serialization.Rlp/ValueRlpStream.cs
@@ -77,75 +77,173 @@ public ref struct ValueRlpStream(SpanSource data)
 
     public void SkipLength()
     {
-        SkipBytes(PeekPrefixAndContentLength().PrefixLength);
+        SkipBytes(PeekPrefixLength());
     }
 
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public readonly int PeekPrefixLength()
+    {
+        // This is a branchless implementation of Rlp prefix length calculation
+        uint p = PeekByte(); // 0..255
+        uint v = p >> 3; // 0..31
+        uint r = v >> 4; // base: 0 for <0x80, 1 for >=0x80
+        uint t = 1u + (p & 7u); // 1..8
+
+        // 1 for 0xB8..0xBF or 0xF8..0xFF
+        uint longMask = ((v | 8u) == 31u) ? 1u : 0u;   
+
+        // Branchless select: t when longMask == 1 else 0
+        uint add = (0u - longMask) & t;
+        return (int)(r + add);
+    }
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
     public readonly int PeekNextRlpLength()
     {
-        (int a, int b) = PeekPrefixAndContentLength();
-        return a + b;
+        int result;
+        int prefix = PeekByte();
+        if (prefix <= 128)
+        {
+            // Single byte (0x00..0x7f). The byte is its own content. Length = 1.
+            result = 1;
+        }
+        else if (prefix <= 183)
+        {
+            // Short string (0x80..0xb7). Content length = prefix - 0x80 (0..55).
+            // Prefix length = 1.
+            result = 1 + prefix - 128;
+        }
+        else if (prefix < 192)
+        {
+            // Long string (0xb8..0xbf). The next (prefix-0xb7) bytes encode the length.
+            // Content length >= 56.
+            result = PeekLongStringRlpLength(prefix);
+        }
+        else if (prefix <= 247)
+        {
+            // Short list (0xc0..0xf7). Content length = prefix - 0xc0 (0..55).
+            // Prefix length = 1.
+            result = 1 + prefix - 192;
+        }
+        else
+        {
+            // Long list (0xf8..0xff). The next (prefix-0xf7) bytes encode the length.
+            // Content length >= 56.
+            result = PeekLongListRlpLength(prefix);
+        }
+
+        return result;
     }
 
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
     public readonly (int PrefixLength, int ContentLength) PeekPrefixAndContentLength()
     {
         (int prefixLength, int contentLength) result;
         int prefix = PeekByte();
         if (prefix <= 128)
         {
+            // Single byte (0x00..0x7f). The byte is its own content. Length = 1.
             result = (0, 1);
         }
         else if (prefix <= 183)
         {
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Trie/TrieNode.cs
```diff
@@ -1167,10 +1167,11 @@ private void SeekChildNotNull(ref ValueRlpStream rlpStream, int itemToSetOn)
...
             rlpStream.Reset();
             rlpStream.SkipLength();
-            if (IsExtension)
+            if (itemToSetOn == 0 && IsExtension)
             {
-                rlpStream.SkipItem();
-                itemToSetOn--;
+                // Optimization: setting itemToSetOn = 1 achieves the same result as the original skip+decrement pattern,
+                // but is more efficient. This is functionally equivalent to calling SkipItem() and decrementing.
+                itemToSetOn = 1;
```
LukaszRozmej: I'm confused how going from 0->1 is equivalent of decrementing?

src/Nethermind/Nethermind.Serialization.Rlp/ValueRlpStream.cs
```diff
@@ -77,76 +77,185 @@ public int PeekNumberOfItemsRemaining(int? beforePosition = null, int maxSearch
...
+            result = 1 + prefix - 192;
+        }
+        else
+        {
+            // Long list (0xf8..0xff). The next (prefix-0xf7) bytes encode the length.
+            // Content length >= 56.
+            result = PeekLongListRlpLength(prefix);
+        }
+
+        return result;
```
LukaszRozmej: `return prefix switch ...` ?

src/Nethermind/Nethermind.Serialization.Rlp/ValueRlpStream.cs
```diff
@@ -77,76 +77,185 @@ public int PeekNumberOfItemsRemaining(int? beforePosition = null, int maxSearch
...
+            ThrowLengthTooLong(contentLength);
         }
 
-        [DoesNotReturn, StackTraceHidden]
-        static void ThrowLengthTooLong(int length)
+        return lengthOfContentLength + 1 + contentLength;
+    }
+
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    private readonly (int prefixLength, int contentLength) PeekLongString(int prefix)
```
LukaszRozmej: `PeekLongStringRlpLength` and `PeekLongString` have same code?

src/Nethermind/Nethermind.Trie/TrieNode.cs
benaadams: decrementing means it does one less in the loop below; which does exactly same thing as the line above `.SkipItem()`

So not doing anything is the equivalent of decrementing (remove skip and decrement)

However this if does exist, and its for the cornercase of when there will be zero iterations of the loop; but it's also an extension node; so now we add one to the loop item count instead. Also move the typecheck after the zero check

This also means can inline `.SkipItem()` without it double bloating

src/Nethermind/Nethermind.Serialization.Rlp/ValueRlpStream.cs
benaadams: Changed

src/Nethermind/Nethermind.Serialization.Rlp/ValueRlpStream.cs
benaadams: PeekLongStringRlpLength returns a single value via register


PeekLongString returns two values via stack

Normally this wouldn't be too relevant but we currently end up going via `PeekPrefixAndContentLength()` and enormous amount of times; so doing  lot of extra work 

image

src/Nethermind/Nethermind.Trie/TrieNode.cs
benaadams: Hmm maybe that's what the comment should say

benaadams: Improved comment

src/Nethermind/Nethermind.Serialization.Rlp/ValueRlpStream.cs
LukaszRozmej: `PeekLongString` can't utilize 64bit register?

benaadams: Custom struct wrapping a `ulong`? with shifts and adds on properties?

LukaszRozmej: Should be possible? I am surprised it doesn't just work?

src/Nethermind/Nethermind.Trie/TrieNode.cs
```diff
@@ -1163,17 +1163,18 @@ private void SeekChild(ref ValueRlpStream rlpStream, int itemToSetOn)
             SeekChildNotNull(ref rlpStream, itemToSetOn);
```
MarekM25: Cosmetic: rename itemToSetOn to index?

src/Nethermind/Nethermind.Trie/TrieNode.cs
```diff
@@ -1163,17 +1163,18 @@ private void SeekChild(ref ValueRlpStream rlpStream, int itemToSetOn)
...
         }
 
-        private void SeekChildNotNull(ref ValueRlpStream rlpStream, int itemToSetOn)
+        private void SeekChildNotNull(ref ValueRlpStream rlpStream, int index)
         {
             rlpStream.Reset();
             rlpStream.SkipLength();
-            if (IsExtension)
```
MarekM25, line 1170: Is it possible that Extension nodes are being called with the different value than zero?

benaadams: Yes, but currently when not zero they do 
```
  rlpStream.SkipItem();
  itemToSetOn--;
```

Which is same as just not doing the decrement and letting the loop run additionally for 1

MarekM25: Is there any reasoning for calling extension node with index >0? (for future API calls of this method)

benaadams: 🤔 probably not

src/Nethermind/Nethermind.Serialization.Rlp/ValueRlpStream.cs
benaadams: Packing adds extra computation per access and per set; so compiler doesn't optimize that way

benaadams:
Follow up to apply to RlpStream and ValueDecoderContext https://github.com/NethermindEth/nethermind/issues/9207

Issue #9032, opened 2025-07-24
Issues on full sync startup


deffrian:
**Description**
We have two problems on startup during full sync:
1. It's impossible to stop the node with Ctrl^C after restart.
2. If we have unprocessed blocks from last run we wait for full sync before we fully start the client. It's very unfortunate if you want to do something with old block like calling rpc. Note that node will process newly downloaded blocks before fully starting up. 

**Steps to Reproduce**
In order to replicate the behavior, please provide a detailed list of steps:
1. Wait for Download stage and wait for some blocks to get downloaded
2. Restart
3. Try to stop the node

deffrian:
@vortex-hue Hi! We don’t assign issues to non-Nethermind members, but feel free to work on it!

PR #9196, opened 2025-08-25
Await db blocks sync inside ReviewBlockTree

deffrian:
Fixes Closes Resolves #
#9032

## Changes

- It is possible to get blocks from the network wile we spin inside a loop in `RegisterPluginRpcModules`. This pr moves waiting part into ReviewBlockTree, before `InitializeNetwork`
- Similar issue is that we load blocks from db during startup in `StartupBlockTreeFixer`. We can keep it as is probably, but it can cause node to not start for a long time.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


src/Nethermind/Nethermind.Init/Steps/RegisterPluginRpcModules.cs
```diff
@@ -13,7 +13,7 @@
 
 namespace Nethermind.Init.Steps;
 
-[RunnerStepDependencies(typeof(InitializeNetwork), typeof(SetupKeyStore), typeof(InitializeBlockchain), typeof(InitializePlugins), typeof(InitializeBlockProducer), typeof(RegisterRpcModules))]
+[RunnerStepDependencies(typeof(SetupKeyStore), typeof(InitializeBlockchain), typeof(InitializePlugins), typeof(InitializeBlockProducer), typeof(RegisterRpcModules))]
```
LukaszRozmej, line 14: It doesn't still need `Initialize`? As `Start` is separated.

src/Nethermind/Nethermind.Init/Steps/RegisterPluginRpcModules.cs
```diff
@@ -28,9 +28,11 @@ public virtual async Task Execute(CancellationToken cancellationToken)
...
             // Ugly temporary hack to not receive engine API messages before end of processing of all blocks after restart.
             // Then we will wait 5s more to ensure everything is processed
-            while (!blockProcessingQueue!.IsEmpty)
-                await Task.Delay(100);
-            await Task.Delay(5000);
+            while (!blockProcessingQueue.IsEmpty && !cancellationToken.IsCancellationRequested)
+                await Task.Delay(100, cancellationToken);
+
+            await Task.Delay(5000, cancellationToken);
+            cancellationToken.ThrowIfCancellationRequested();
```
LukaszRozmej: Does it need to throw? Instead of just ignoring rest of the method?

src/Nethermind/Nethermind.Init/Steps/RegisterPluginRpcModules.cs
deffrian: `RegisterRpcModules` depends in `Initialize`

src/Nethermind/Nethermind.Init/Steps/ReviewBlockTree.cs
```diff
@@ -68,6 +71,20 @@ await blockTree.Accept(fixer, cancellationToken).ContinueWith(t =>
...
+
+            blockProcessingQueue.ProcessingQueueEmpty += OnProcessingQueueEmpty;
+            if (!blockProcessingQueue.IsEmpty) // Just in case the queue got empty before we subscribed
+            {
+                await _blocksProcessedTaskSource.Task.WaitAsync(cancellationToken);
+            }
+            blockProcessingQueue.ProcessingQueueEmpty -= OnProcessingQueueEmpty;
+        }
+
+        private readonly TaskCompletionSource _blocksProcessedTaskSource = new();
```
LukaszRozmej, line 83: `ManualResetEventSlim` wouldn't be better?

src/Nethermind/Nethermind.Init/Steps/RegisterPluginRpcModules.cs
LukaszRozmej: ok so it is there implicitly, could be left explicitly but whatever.

asdacap (approved):

src/Nethermind/Nethermind.Init/Steps/ReviewBlockTree.cs
deffrian: ManualResetEventSlim is blocking. Better to use async imo

PR #9199, opened 2025-08-25
Improve deposit request fields validation

flcl42:
The PR adds fixed lengths validation. It did not work because we decode as dynamic arrays instead of uints of fixed size https://github.com/NethermindEth/nethermind/blob/70d17f27412c70254e67af10dcdb654f2519342c/src/Nethermind/Nethermind.Consensus/ExecutionRequests/ExecutionRequestProcessor.cs?plain=1#L25 which is kinda OK, because simplifies further copying into deposit request. But we missed length validation.

## Changes

- Add length of fields validation

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/ExecutionRequests/ExecutionRequestProcessor.cs
+++ b/src/Nethermind/Nethermind.Consensus/ExecutionRequests/ExecutionRequestProcessor.cs
@@ -4,6 +4,7 @@
 
 using Nethermind.Abi;
 using Nethermind.Blockchain;
+using Nethermind.Blockchain.Tracing;
 using Nethermind.Consensus.Messages;
 using Nethermind.Core;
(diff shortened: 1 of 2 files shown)


LukaszRozmej (approved):

emlautarom1 (approved):
LGTM.

I think we should add a test to reflect this additional check (and prevent regressions).

How did you notice this issue?

flcl42:
> LGTM.
> 
> I think we should add a test to reflect this additional check (and prevent regressions).
> 
> How did you notice this issue?

Revealed by hive tests, where a big 160 bytes int was passed instead of 8 bytes long one

PR #9212, opened 2025-08-27
Added basic spec classes

ak88:
Renamed project folder because it was named wrong.


smartprogrammer93 (approved):

LukaszRozmej (approved):

src/Nethermind/Nethermind.Xdc/Spec/XdcV2ReleaseSpec.cs
```diff
@@ -0,0 +1,32 @@
...
+using Nethermind.Specs;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Nethermind.Xdc.Spec;
+internal class XdcV2ReleaseSpec : ReleaseSpec, IXdcReleaseSpec
+{
```
Demuirgos, line 14: what is the difference between releaseSpec and a Config ?

ak88: Config is related to node setup usually and specs are settings for the network. 
But i also think sometimes you can find network related stuff in config, but generally it should be divided like that.

PR #9214, opened 2025-08-29
Implement Xdc Header Decoder

### LINKED ISSUE #9185: XDC Header decoder
Implement a RLP decoder/encoder for XDC headers

cicr99:
Fixes #9185 

## Changes

- Introduced `IHeaderRlpCodec` to decouple block header (de)serialization from `BlockDecoder`. Now BlockDecoder accepts an IHeaderRlpCodec via constructor and added a default constructor.
- Added XdcBlockHeader : BlockHeader with XDC-specific fields (`byte[] Validators`, `byte[] Validator`, `byte[] Penalties`)
- Implemented XdcHeaderDecoder : IHeaderRlpCodec
    - AuRa removed for XDC.
    - XDC extras encoded/decoded in order Validators → Validator → Penalties → BaseFee (optional, last).
- Added project with tests

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Serialization.Rlp/BlockDecoder.cs
+++ b/src/Nethermind/Nethermind.Serialization.Rlp/BlockDecoder.cs
@@ -7,17 +7,12 @@ using Nethermind.Core;
 
 namespace Nethermind.Serialization.Rlp
 {
-    public class BlockDecoder : IRlpValueDecoder<Block>, IRlpStreamDecoder<Block>
+    public class BlockDecoder(IHeaderDecoder headerDecoder) : IRlpValueDecoder<Block>, IRlpStreamDecoder<Block>
     {
-        private readonly IHeaderDecoder _headerDecoder;
+        private readonly IHeaderDecoder _headerDecoder = headerDecoder ?? throw new ArgumentNullException(nameof(headerDecoder));
         private readonly BlockBodyDecoder _blockBodyDecoder = BlockBodyDecoder.Instance;
 
         public BlockDecoder() : this(new HeaderDecoder()) { }
-
-        public BlockDecoder(IHeaderDecoder headerCodec)
-        {
-            _headerDecoder = headerCodec ?? throw new ArgumentNullException(nameof(headerCodec));
-        }
 
         public Block? Decode(RlpStream rlpStream, RlpBehaviors rlpBehaviors = RlpBehaviors.None)
         {


src/Nethermind/Nethermind.slnx
```diff
@@ -79,6 +79,7 @@
     <Project Path="Nethermind.Trie.Test/Nethermind.Trie.Test.csproj" />
     <Project Path="Nethermind.TxPool.Test/Nethermind.TxPool.Test.csproj" />
     <Project Path="Nethermind.Wallet.Test/Nethermind.Wallet.Test.csproj" />
+    <Project Path="Nethermind.Xdc.Test\Nethermind.Xdc.Test.csproj" Type="Classic C#" />
```
ak88:
```suggestion
    <Project Path="Nethermind.Xdc.Test\Nethermind.Xdc.Test.csproj"  />
```

src/Nethermind/Nethermind.Xdc.Test/XdcHeaderDecoderTests.cs
```diff
@@ -0,0 +1,137 @@
...
+            Assert.That(original.GasLimit, Is.EqualTo(decoded.GasLimit), "The gas limit should be the same.");
+            Assert.That(original.GasUsed, Is.EqualTo(decoded.GasUsed), "The gas used should be the same.");
+            Assert.That(original.Timestamp, Is.EqualTo(decoded.Timestamp), "The timestamp should be the same.");
+            Assert.That(original.ExtraData, Is.EqualTo(decoded.ExtraData), "The extra data should be the same.");
+            Assert.That(original.MixHash, Is.EqualTo(decoded.MixHash), "The mix hash should be the same.");
+            Assert.That(original.Nonce, Is.EqualTo(decoded.Nonce), "The nonce should be the same.");
+            Assert.That(decoded.Validators, Is.EqualTo(original.Validators), "Validators should match.");
+            Assert.That(original.Validator, Is.EqualTo(decoded.Validator), "Validator should match.");
+            Assert.That(original.Penalties, Is.EqualTo(decoded.Penalties), "Penalties should match.");
+            Assert.That(original.BaseFeePerGas, Is.EqualTo(decoded.BaseFeePerGas), "BaseFeePerGas should be the same.");
```
ak88: You can use https://fluentassertions.com/objectgraphs/ 
We use fluent assertions in a lot of test projects

src/Nethermind/Nethermind.Serialization.Rlp/BlockDecoder.cs
```diff
@@ -9,9 +9,16 @@ namespace Nethermind.Serialization.Rlp
 {
     public class BlockDecoder : IRlpValueDecoder<Block>, IRlpStreamDecoder<Block>
     {
-        private readonly HeaderDecoder _headerDecoder = new();
+        private readonly IHeaderRlpCodec _headerDecoder;
```
ak88: did you check if we can reuse BlockDecoder directly in XDC? 
I think it should be the same.

src/Nethermind/Nethermind.Xdc.Test/XdcHeaderDecoderTests.cs
```diff
@@ -0,0 +1,137 @@
...
+                Penalties = Array.Empty<byte>(),
+            };
+
+            if (includeBaseFee)
+            {
+                header.BaseFeePerGas = (UInt256)1_000_000_000; // 1 gwei
+            }
+
+            return header;
+        }
```
ak88: We have a specialized builder class for setting up objects `BlockBuilder`
You can add `XdcBlockHeader `to it if you want

ak88 (approved):
Good work - only minor stuff.

smartprogrammer93 (commented):
We could potentially use DI to do this in a better way instead. @asdacap

asdacap:
I definitely agree. Tragicaly, the decoder was buried too deep to make it a simple change.

asdacap (approved):

smartprogrammer93 (approved):

src/Nethermind/Nethermind.Serialization.Rlp/BlockDecoder.cs
LukaszRozmej: Not sure why is it here? It seems not used at this level - thus not needed?

src/Nethermind/Nethermind.Serialization.Rlp/HeaderDecoder.cs
```diff
@@ -8,7 +8,9 @@
 
 namespace Nethermind.Serialization.Rlp
 {
-    public class HeaderDecoder : IRlpValueDecoder<BlockHeader>, IRlpStreamDecoder<BlockHeader>
+    public interface IHeaderRlpCodec : IRlpValueDecoder<BlockHeader>, IRlpStreamDecoder<BlockHeader> { }
+
+    public class HeaderDecoder : IHeaderRlpCodec
```
LukaszRozmej: Looking at naming conventions `IHeaderDecoder`?

src/Nethermind/Nethermind.Xdc.Test/XdcHeaderDecoderTests.cs
LukaszRozmej: Potentially we could use the `Build.A.XdcHeader` pattern?

src/Nethermind/Nethermind.Xdc.Test/XdcHeaderDecoderTests.cs
```diff
@@ -0,0 +1,137 @@
...
+
+        [Test]
+        public void No_BaseFee()
+        {
+            var codec = new XdcHeaderDecoder();
+            XdcBlockHeader header = MakeHeader(includeBaseFee: false);
+
+            // Encode without base fee
+            Rlp r = codec.Encode(header);
+            byte[] bytes = r.Bytes;
```
LukaszRozmej: A lot of duplicated code in every test?

src/Nethermind/Nethermind.Xdc/XdcHeaderDecoder.cs
```diff
@@ -0,0 +1,233 @@
+// SPDX-FileCopyrightText: 2023 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System;
+using Nethermind.Core;
+using Nethermind.Core.Crypto;
+using Nethermind.Int256;
+using Nethermind.Serialization.Rlp;
+namespace Nethermind.Xdc;
+public sealed class XdcHeaderDecoder : IHeaderRlpCodec
```
LukaszRozmej: Not sure if we can, but can we reuse the original decoder and extend it by wrapping it, inheriting or creating some helper methods if there is large overlap? (Might not be possible)

src/Nethermind/Nethermind.Xdc.Test/XdcHeaderDecoderTests.cs
cicr99: Done, a lot better now. Thanks!

src/Nethermind/Nethermind.Serialization.Rlp/BlockDecoder.cs
cicr99: I assumed it would be the same tbh, never actually checked. Will take a look.

cicr99: It's being used to get the length of the block in `GetContentLength` method

src/Nethermind/Nethermind.Xdc.Test/XdcHeaderDecoderTests.cs
cicr99: Right, changed it 👍

src/Nethermind/Nethermind.Xdc/XdcHeaderDecoder.cs
cicr99: Inheriting or wrapping the original decoder as it is, won't be a good idea because the XdcHeader contains a few extra fields that go just before BaseFee, so the order wouldn't be correct. What we could do instead is the other possibility you were suggesting , try to extract the shared logic into some helpers and so each decoder can set their own order, just not sure if it'll be worth it: right now the only common factor is the first 13 fields or so, but the rest is pretty different, the current decoder has assumptions and fields that are not present in Xdc.

src/Nethermind/Nethermind.Serialization.Rlp/BlockDecoder.cs
```diff
@@ -9,9 +9,16 @@ namespace Nethermind.Serialization.Rlp
 {
     public class BlockDecoder : IRlpValueDecoder<Block>, IRlpStreamDecoder<Block>
     {
-        private readonly HeaderDecoder _headerDecoder = new();
+        private readonly IHeaderDecoder _headerDecoder;
         private readonly BlockBodyDecoder _blockBodyDecoder = BlockBodyDecoder.Instance;
 
+        public BlockDecoder() : this(new HeaderDecoder()) { }
+
+        public BlockDecoder(IHeaderDecoder headerCodec)
```
LukaszRozmej: `headerCodec` -> `headerDecoder`

Also I would make it primary constructor

cicr99: Sorry about the naming, forgot it was there in the previous change 😅

src/Nethermind/Nethermind.Core.Test/Builders/XdcBlockHeaderBuilder.cs
```diff
@@ -0,0 +1,42 @@
...
+            1_700_000_000,
+            new byte[] { 1, 2, 3 })
+        {
+            StateRoot = Keccak.EmptyTreeHash,
+            TxRoot = Keccak.EmptyTreeHash,
+            ReceiptsRoot = Keccak.EmptyTreeHash,
+            Bloom = Bloom.Empty,
+            GasUsed = 21_000,
+            MixHash = Keccak.Compute("mix_hash"),
+            Nonce = 1000,
```
ak88, line 30:
```suggestion
            Nonce = 0,
```

ak88 (approved):

src/Nethermind/Nethermind.Core.Test/Nethermind.Core.Test.csproj
```diff
@@ -17,6 +17,7 @@
     <ProjectReference Include="..\Nethermind.Facade\Nethermind.Facade.csproj" />
     <ProjectReference Include="..\Nethermind.Merge.Plugin\Nethermind.Merge.Plugin.csproj" />
     <ProjectReference Include="..\Nethermind.Specs.Test\Nethermind.Specs.Test.csproj" />
+    <ProjectReference Include="..\Nethermind.Xdc\Nethermind.Xdc.csproj" />
```
svlachakis, line 20: @ak88 @cicr99 Wouldn't it be better to avoid dependency of Xdc in Core tests? what do you think?

ak88: Yea it shouldnt be necessary

ak88: @svlachakis @cicr99 removed it here https://github.com/NethermindEth/nethermind/pull/9242

PR #9224, opened 2025-09-01
Hive to load blocks before RPC start

flcl42:
Fixes fusaka-devnet-5 consume rlp hive tests

Step are now loading in parallel and RPC may return just genesis instead of the 1st block later loaded by hive

hive blocks, then rpc

## Changes

- Wait hive blocks before start of RPC

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Runner/Ethereum/Steps/StartRpc.cs
+++ b/src/Nethermind/Nethermind.Runner/Ethereum/Steps/StartRpc.cs
@@ -11,6 +11,7 @@ using Nethermind.Api.Extensions;
 using Nethermind.Api.Steps;
 using Nethermind.Core;
 using Nethermind.Core.Authentication;
+using Nethermind.Hive;
 using Nethermind.Init.Steps;
 using Nethermind.JsonRpc;
 using Nethermind.JsonRpc.Modules;
@@ -22,7 +23,7 @@ using Nethermind.Sockets;
 
 namespace Nethermind.Runner.Ethereum.Steps;
 
-[RunnerStepDependencies(typeof(InitializeNetwork), typeof(RegisterRpcModules), typeof(RegisterPluginRpcModules))]
+[RunnerStepDependencies(typeof(InitializeNetwork), typeof(RegisterRpcModules), typeof(RegisterPluginRpcModules), typeof(HiveStep))]
 public class StartRpc(INethermindApi api, IJsonRpcServiceConfigurer[] serviceConfigurers, IWebSocketsManager webSocketsManager) : IStep
 {
     public async Task Execute(CancellationToken cancellationToken)


LukaszRozmej (approved):

src/Nethermind/Nethermind.Api/Steps/RunnerStepAttribute.cs
```diff
@@ -3,16 +3,18 @@
 
 using System;
 
-namespace Nethermind.Api.Steps
+namespace Nethermind.Api.Steps;
+
+[AttributeUsage(AttributeTargets.Class)]
+public class RunnerStepDependenciesAttribute(bool optional, Type[] dependencies, Type[] dependents) : Attribute
```
LukaszRozmej: Maybe not put it in constructor and only have a setter for property?

src/Nethermind/Nethermind.Hive/HiveStep.cs
```diff
@@ -11,7 +11,7 @@
 
 namespace Nethermind.Hive;
 
-[RunnerStepDependencies(typeof(InitializeNetwork))]
+[RunnerStepDependencies(true, typeof(InitializeNetwork))]
```
LukaszRozmej:
```suggestion
[RunnerStepDependencies(typeof(InitializeNetwork), Optional=true)]
```

src/Nethermind/Nethermind.Init/Steps/EthereumStepsManager.cs
```diff
@@ -88,6 +89,26 @@ private List<Task> CreateAndExecuteSteps(CancellationToken cancellationToken)
                 }
             }
 
+            // Remove absent optional steps
```
LukaszRozmej: Can we not remove them, but don't add them in the first place in the above loop?

src/Nethermind/Nethermind.Hive/HiveStep.cs
flcl42: typeof(InitializeNetwork) here is a part of `params` array, that's why optional goes first

src/Nethermind/Nethermind.Api/Steps/RunnerStepAttribute.cs
flcl42: Can't use setter in attributes

src/Nethermind/Nethermind.Init/Steps/EthereumStepsManager.cs
flcl42: Yep, moved to the loop above

LukaszRozmej: removed ToArray

src/Nethermind/Nethermind.Api/Steps/RunnerStepAttribute.cs
LukaszRozmej: sure you can

src/Nethermind/Nethermind.Hive/HiveStep.cs
LukaszRozmej: worked for me

asdacap:
Ah, because you can't use `HiveStep` in `StartRpc`. Just use the `Dependents`, set it to `StartRpc`.

src/Nethermind/Nethermind.Init/Steps/EthereumStepsManager.cs
```diff
@@ -63,44 +63,41 @@ private List<Task> CreateAndExecuteSteps(CancellationToken cancellationToken)
...
-                        throw new StepDependencyException(
-                            $"The dependent step {type.Name} for {stepWrapper.StepInfo.StepBaseType.Name} is missing.");
+                        throw new StepDependencyException($"The dependent step {type.Name} for {stepWrapper.StepInfo.StepBaseType.Name} is missing.");
                     }
                 }
 
-                // Remove absent optional steps
-                foreach (Type dep in stepWrapper.Dependencies.ToArray())
+                // Remove absent optional dependencies
+                for (var i = 0; i < stepWrapper.Dependencies.Count; i++)
```
flcl42, line 87: Can be in reverse order to remove additional i--

flcl42:
> Ah, because you can't use `HiveStep` in `StartRpc`. Just use the `Dependents`, set it to `StartRpc`.

Leads to circular dependency of csprojs. We can try to switch to nameof instead of typeof, it's too much work for this fix though

emlautarom1:
Unrelated to the PR itself, but how did you get the dependency tree printed out like that? Might come in handy in the future.

flcl42:
> Unrelated to the PR itself, but how did you get the dependency tree printed out like that? Might come in handy in the future.

Is a debug log in BuildStepDependencyTree

Issue #9226, opened 2025-09-01
Add ENR support on admin_peers


svlachakis:
We need to add ENR support on admin_peers endpoint after this PR gets merged https://github.com/NethermindEth/nethermind/pull/9003.

Talk to maybe @flcl42 or @asdacap  about how to add it

PR #9228, opened 2025-09-01
ENR Support on admin_peers

svlachakis:
Fixes Closes Resolves https://github.com/NethermindEth/nethermind/issues/9226

## Changes

ENR discv4 & discv5 implementation 

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

## Testing

Screenshot 2025-09-02 at 11 46 30 AM

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

## Remarks

Depends on https://github.com/NethermindEth/nethermind/pull/9003, we need to set ENR in `PeerInfo.cs` when it get's merged


src/Nethermind/Nethermind.Network.Discovery.Test/NodeLifecycleManagerTests.cs
```diff
@@ -364,6 +363,256 @@ public void EvictCandidateStateLostEvictionTest()
         //Assert.IsTrue(closestNodes.Count(x => x.Host == candidateNode.Host) == 1);
     }
 
+    [Test]
+    public async Task ProcessEnrResponseMsg_WhenBondedWithRemoteNode_StoresEnrAndSendsEnrEvent()
```
LukaszRozmej, line 367: A lot of code duplication in the tests, can we reduce it?

src/Nethermind/Nethermind.Network.Discovery/Lifecycle/NodeLifecycleManager.cs
```diff
@@ -370,6 +372,7 @@ private void UpdateState(NodeLifecycleState newState)
 
         State = newState;
         OnStateChanged?.Invoke(this, State);
+        if (newState == NodeLifecycleState.Active && string.IsNullOrEmpty(ManagedNode.Enr)) SendEnrRequest();
```
flcl42: We don't practice one line ifs, because code surrounded by curly braces looks safer

src/Nethermind/Nethermind.Network.Stats/Model/Node.cs
```diff
@@ -73,6 +73,7 @@ public string ClientId
 
         public string EthDetails { get; set; }
         public long CurrentReputation { get; set; }
+        public string Enr { get; set; }
```
flcl42, line 76: Can be marked with ?, as it's not always set in case of discv4. Btw can we request enr even if enrsequence is not available

svlachakis: As discussed in slack:

1. We don't have nullable context here so it's not worth it - leaving it as is.
2. The current implementation in `NodeLifecycleManager` already handles this I think - unless I'm missing something. When a node becomes Active and has no ENR (string.IsNullOrEmpty(ManagedNode.Enr)), it sends an ENR request. The ENR request message doesn't include or require a sequence number, the sequence number is only used to validate responses, ensuring they're more recent than any previously received ENR. - Confirmed

src/Nethermind/Nethermind.Network.Discovery/Lifecycle/NodeLifecycleManager.cs
svlachakis: done

src/Nethermind/Nethermind.Network.Discovery.Test/NodeLifecycleManagerTests.cs
svlachakis: done, what do you think?

PR #9231, opened 2025-09-02
Fix receipts deletion

Marchhill:
## Changes

- Only delete receipts with history pruning, not when disabled

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain.Test/Receipts/PersistentReceiptStorageTests.cs
+++ b/src/Nethermind/Nethermind.Blockchain.Test/Receipts/PersistentReceiptStorageTests.cs
@@ -442,7 +442,7 @@ public class PersistentReceiptStorageTests
     }
 
     [Test]
-    public void When_NewHeadBlock_ClearOldTxIndex()
+    public void When_NewHeadBlock_ClearOldTxIndex_And_KeepsReceipts()
     {
         _receiptConfig.TxLookupLimit = 1000;
         CreateStorage();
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Blockchain.Test/Receipts/PersistentReceiptStorageTests.cs
```diff
@@ -466,6 +466,10 @@ public void When_NewHeadBlock_ClearOldTxIndex()
             () => _receiptsDb.GetColumnDb(ReceiptsColumns.Transactions)[receipts[0].TxHash!.Bytes],
             Is.Null.After(1000, 100)
             );
+
+        Span<byte> blockNumPrefixed = stackalloc byte[40];
+        PersistentReceiptStorage.GetBlockNumPrefixedKey(receipts[0].BlockNumber, receipts[0].BlockHash!, blockNumPrefixed);
```
alexb5dh: What about just doing `Get(blockHash)` without making `GetBlockNumPrefixedKey` internal?

alexb5dh (approved):

src/Nethermind/Nethermind.Blockchain.Test/Receipts/PersistentReceiptStorageTests.cs
Marchhill: yep already changed this to use HasBlock

PR #9237, opened 2025-09-03
Remove unused references for XDC

ak88:


LukaszRozmej (commented):
those are often used in test sooner or later

PR #9239, opened 2025-09-03
Clear storage cell / trie without reading cell's value first fails

damian-orzechowski:
Fixes Closes Resolves #

Fixes an issue when you set empty value (clear) for a storage cell that was not previously read (using `Get`). In such case, new value (empty) would not be set as code assumes non existent values as empty, which in turn fails on value comparison (new vs old) when commiting changes.

## Changes

- `PerContractState` will load value from tree if not found in block cache when saving a new change

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.State.Test/StorageProviderTests.cs
+++ b/src/Nethermind/Nethermind.State.Test/StorageProviderTests.cs
@@ -441,6 +441,66 @@ public class StorageProviderTests
         provider.Get(nonAccessedStorageCell).ToArray().Should().BeEquivalentTo(StorageTree.ZeroBytes);
     }
 
+    [Test]
+    public void Set_empty_value_for_storage_cell_without_read_clears_data()
+    {
+        IWorldState worldState = new WorldState(TestTrieStoreFactory.Build(new MemDb(), LimboLogs.Instance), Substitute.For<IDb>(), LogManager);
+
+        using var disposable = worldState.BeginScope(IWorldState.PreGenesis);
+        worldState.CreateAccount(TestItem.AddressA, 1);
+        worldState.Commit(Prague.Instance);
+        worldState.CommitTree(0);
+        Hash256 emptyHash = worldState.StateRoot;
+
+        worldState.Set(new StorageCell(TestItem.AddressA, 1), _values[11]);
+        worldState.Set(new StorageCell(TestItem.AddressA, 2), _values[12]);
+        worldState.Commit(Prague.Instance);
+        worldState.CommitTree(1);
+
+        var fullHash = worldState.StateRoot;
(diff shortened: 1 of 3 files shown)


src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
```diff
@@ -557,7 +557,11 @@ public void SaveChange(StorageCell storageCell, byte[] value)
             ref ChangeTrace valueChanges = ref BlockChange.GetValueRefOrAddDefault(storageCell.Index, out bool exists);
             if (!exists)
             {
-                valueChanges = new ChangeTrace(value);
+                byte[] treeValue = !_provider._populatePreBlockCache ?
+                    LoadFromTreeReadPreWarmCache(in storageCell) :
+                    LoadFromTreePopulatePrewarmCache(in storageCell);
+
+                valueChanges = new ChangeTrace(treeValue, value);
```
benaadams: Does this just need to be?
```suggestion
                EnsureStorageTree();
                valueChanges = new ChangeTrace(value);
```

src/Nethermind/Nethermind.State.Test/StorageProviderTests.cs
```diff
@@ -441,6 +441,66 @@ public void Selfdestruct_clears_cache()
...
+        using var disposable = worldState.BeginScope(IWorldState.PreGenesis);
+        worldState.CreateAccount(TestItem.AddressA, 1);
+        worldState.Commit(Prague.Instance);
+        worldState.CommitTree(0);
+        Hash256 emptyHash = worldState.StateRoot;
+
+        worldState.Set(new StorageCell(TestItem.AddressA, 1), _values[11]);
+        worldState.Set(new StorageCell(TestItem.AddressA, 2), _values[12]);
+        worldState.Commit(Prague.Instance);
+        worldState.CommitTree(1);
```
asdacap, line 487: Looks like it will also fail previously if `Reset()` is called after the commit. Previously the blockchange and the storage tree is separated, now its the same causing the `CommitTree` to have an implicit reset.

damian-orzechowski: Yes, that's correct. Recreated the same behaviour with calling `Reset()` - though with calling `Reset(false)` works fine.

benaadams (approved):

PR #9243, opened 2025-09-04
Cleanup/Simplify CodeInfoRepository registration.

asdacap:
- ~~Change to singleton. It is probably a mistake to make it a scoped service before.~~ Change to properly scoped at #9262 
- Precompiles is now provided by `IPrecompileProvider`. This make it easy for plugin to add or remove precompile. Also needed for the next change.
- CachedCodeInfoRepository is now a wrapper. This allow a separate ICodeInfoRepository just for the main processing components as the one there uses prewarm caches.

- [ ] Optimization
- [X] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- [X] Mainnet start normally.


LukaszRozmej:
Question - if we have `eth_call` that deploys a contract with new bytecode, that bytecode would end-up in the DB (singleton code repository)?

asdacap:
Is it not suppose to be that? Like the singleton itself dont have a state. Except a static lru.

asdacap:
> @asdacap Great work on refactoring this!
> 
> I noticed there’s still some duplication around precompiles, since they also live in `IReleaseSpec`. You might want to check out my recent PR: #9222, where I moved the old static `AddressExtensions.cs` to the release spec to improve extensibility.
> 
> Since you’re already refactoring, it would be amazing if we could centralize all precompile definitions in one place. I’m a bit concerned that eventually someone might update `EthereumPrecompileProvider.cs` without updating `PrecompiledAddresses.cs`, or vice versa, which could lead to bugs.
> 
> Let me know what you think!

I looks like it is doing what its suppose to do. Not all precompile should be allowed at all block and it need to be enabled/disabled per `IReleaseSpec`. I don't know why `HashSet<AddressAsKey> Precompiles { get; }` exist and exposed though, but `IsPrecompile` should be as is.

asdacap:
You know what, let me make an alternative. The interface highly aggravate me.

PR #9244, opened 2025-09-04
Fix simulate gas

deffrian:
Fixes Closes Resolves #

## Changes

- _List the changes_

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Remarks

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Validators/UnclesValidator.cs
+++ b/src/Nethermind/Nethermind.Consensus/Validators/UnclesValidator.cs
@@ -42,7 +42,8 @@ namespace Nethermind.Consensus.Validators
             for (int i = 0; i < uncles.Length; i++)
             {
                 BlockHeader uncle = uncles[i];
-                if (!_headerValidator.Validate(uncle, parentHeader, true, out string? err))
+                BlockHeader? uncleParent = _blockTree.FindParentHeader(uncle, BlockTreeLookupOptions.TotalDifficultyNotNeeded);
+                if (!_headerValidator.Validate(uncle, uncleParent, true, out string? err))
                 {
                     _logger.Info($"Invalid block ({header.ToString(BlockHeader.Format.Full)}) - uncle's header invalid. Error: {err}");
                     return false;
--- a/src/Nethermind/Nethermind.Consensus/Validators/UnclesValidator.cs
+++ b/src/Nethermind/Nethermind.Consensus/Validators/UnclesValidator.cs
@@ -42,9 +42,9 @@ namespace Nethermind.Consensus.Validators
             for (int i = 0; i < uncles.Length; i++)
             {
                 BlockHeader uncle = uncles[i];
-                if (!_headerValidator.Validate(uncle, parentHeader, true, out _))
+                if (!_headerValidator.Validate(uncle, parentHeader, true, out string? err))
                 {
-                    _logger.Info($"Invalid block ({header.ToString(BlockHeader.Format.Full)}) - uncle's header invalid");
+                    _logger.Info($"Invalid block ({header.ToString(BlockHeader.Format.Full)}) - uncle's header invalid. Error: {err}");
                     return false;
                 }
 
--- a/src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
+++ b/src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
@@ -380,10 +380,10 @@ namespace Nethermind.Evm.TransactionProcessing
 
         protected virtual TransactionResult ValidateGas(Transaction tx, BlockHeader header, long minGasRequired, bool validate)
         {
-            if (tx.GasLimit < minGasRequired)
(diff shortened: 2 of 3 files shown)


src/Nethermind/Nethermind.Consensus/Validators/UnclesValidator.cs
```diff
@@ -42,9 +42,10 @@ public bool Validate(BlockHeader header, BlockHeader? parentHeader, BlockHeader[
             for (int i = 0; i < uncles.Length; i++)
             {
                 BlockHeader uncle = uncles[i];
-                if (!_headerValidator.Validate(uncle, parentHeader, true, out _))
+                BlockHeader? uncleParent = _blockTree.FindParentHeader(uncle, BlockTreeLookupOptions.TotalDifficultyNotNeeded);
```
asdacap, line 45: Can pass the parent explicitly as the `parentHeader`?

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -380,10 +380,10 @@ protected virtual TransactionResult ValidateStatic(
 
         protected virtual TransactionResult ValidateGas(Transaction tx, BlockHeader header, long minGasRequired, bool validate)
         {
-            if (tx.GasLimit < minGasRequired)
+            if (validate && tx.GasLimit < minGasRequired)
```
LukaszRozmej: This flag is confusing me, why should it exist here in the first place?
Is it something used in arbitrum? @damian-orzechowski ?
If not why call this method if it is false at all?

src/Nethermind/Nethermind.Consensus/Validators/UnclesValidator.cs
deffrian: Can we have different parents for uncles? In the past it fetched parent for every uncle seperately

LukaszRozmej, line 45: Uncle parent should be the same as this block parent? Is it possible not to be the same? If yes shouldn't we check it? Maybe just checking hashes would suffice and we could skip this `FindParentHeader` call?

deffrian: In hive uncle.Parent = thisBlock.Parent.Parent

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
deffrian: in eth_simulate we have validate flag. I'm not sure if we need it here, I added this check before I realized the uncles issue. Will double check

src/Nethermind/Nethermind.Consensus/Validators/UnclesValidator.cs
deffrian: I think uncles can be older. At least this is what I'm getting from what happens in `UnclesValidator` @LukaszRozmej

asdacap: Ow... I guess it can.

LukaszRozmej: which hive tests? That seems wrong?

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
LukaszRozmej: I meant removing whole flag entirely

src/Nethermind/Nethermind.Consensus/Validators/UnclesValidator.cs
deffrian: rpc-compat. In blocks from `chain.rlp` . Not sure if it's the same for all tests

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
deffrian: We don't need this one

src/Nethermind/Nethermind.Consensus/Validators/UnclesValidator.cs
LukaszRozmej: Are those eth simulate tests?

deffrian: not only, all rpc tests pretty much

LukaszRozmej (approved):

LukaszRozmej:
Issue introduced in https://github.com/NethermindEth/nethermind/pull/9168

Issue #9186, opened 2025-08-21
XDC genesis spec


ak88:
Create a genesis spec matching the XDC genesis spec producing correct genesis block.

PR #9245, opened 2025-09-04
Xdc genesis spec

cicr99:
Initial implementation of spec for #9186 

## Changes

- Add `chainspec/xdc.json` following the expected format
- Add `configs/xdc.json` (runner config)
- Modify `XdcChainSpecEngineParameters` to include different round-based configurations.
- Implement `XdcChainSpecBasedSpecProvider`
- Implement xdc plugin skeleton
- Remove IXdcConfig in favor of ReleaseSpec

### DIFF
--- a/src/Nethermind/Nethermind.Xdc/Spec/XdcChainSpecEngineParameters.cs
+++ b/src/Nethermind/Nethermind.Xdc/Spec/XdcChainSpecEngineParameters.cs
@@ -23,7 +23,7 @@ public class XdcChainSpecEngineParameters : IChainSpecEngineParameters
 
 
     private List<V2ConfigParams> _v2Configs = new();
-        public List<V2ConfigParams> V2Configs
+    public List<V2ConfigParams> V2Configs
     {
         get => _v2Configs;
         set
--- a/src/Nethermind/Nethermind.Xdc.Test/XdcSpecProviderTests.cs
+++ b/src/Nethermind/Nethermind.Xdc.Test/XdcSpecProviderTests.cs
@@ -76,7 +76,7 @@ public class XdcSpecProviderTests
             new() { SwitchRound = 220000 },
         };
 
-        V2ConfigParams cfg = XdcChainSpecBasedSpecProvider.GetConfigAtRound(v2Configs, round);
+        V2ConfigParams cfg = XdcReleaseSpec.GetConfigAtRound(v2Configs, round);
 
         cfg.SwitchRound.Should().Be(expectedSwitchRound);
     }


src/Nethermind/Nethermind.Runner.Test/EthereumRunnerTests.cs
```diff
@@ -161,6 +161,8 @@ public async Task Smoke((string file, ConfigProvider configProvider) testCase, i
 
         if (testCase.file.Contains("none.json")) Assert.Ignore("engine port missing");
         if (testCase.file.Contains("radius_testnet-sepolia.json")) Assert.Ignore("sequencer url not specified");
+        // TODO: enable when XDC consensus plugin is implemented.
+        if (testCase.file.Contains("xdc.json")) Assert.Ignore("xdc consensus plugin not implemented yet");
```
ak88: Lets add the missing XdcChainSpecEngineParameters

src/Nethermind/Nethermind.Runner.Test/EthereumRunnerTests.cs
```diff
@@ -174,7 +176,8 @@ public async Task Smoke_cancel((string file, ConfigProvider configProvider) test
             // some weird thing, not worth investigating
             return;
         }
-
+        // TODO: enable when XDC consensus plugin is implemented.
+        if (testCase.file.Contains("xdc.json")) Assert.Ignore("xdc consensus plugin not implemented yet");
```
ak88: Lets add the missing XdcChainSpecEngineParameters

src/Nethermind/Nethermind.Xdc/Spec/XdcChainSpecEngineParameters.cs
```diff
@@ -6,19 +6,53 @@
...
+        get => _v2Configs;
+        set
+        {
+            _v2Configs = value ?? new();
+            _v2Configs.Sort((a, b) => a.SwitchRound.CompareTo(b.SwitchRound));
+            CheckConfig(_v2Configs);
+        }
+    }
+
+    private static void CheckConfig(List<V2ConfigParams> list)
```
ak88, line 37: Im fine with doing it like this for now, and then at a later point when we have more of the full picture, we can refactor it if we want.

src/Nethermind/Nethermind.Runner.Test/EthereumRunnerTests.cs
cicr99: After adding the missing XdcChainSpecEngineParameters, we still get an error as the SmokeTest loads the plugins and there is none for Xdc. I was planning to do this in a different PR and enable the test there, wdyt?

cicr99: I've added an initial implementation for xdc plugin and enabled the tests again

ak88 (approved):

PR #9254, opened 2025-09-07
Make plugin ordering culture-invariant and deterministic

sashaodessa:
- **Summary**: Replace culture-sensitive string operations in `PluginLoader.OrderPlugins` with culture-invariant, deterministic comparisons.
- **Changes**:
  - Replace `ToLower()`/`CompareTo` with:
    - `StringComparer.OrdinalIgnoreCase` for the ordering map
    - `string.Compare(..., StringComparison.Ordinal)` for tie-break sorting
  - Build `orderMap` to avoid linear `IndexOf` lookups.
- **Why**: Prevent non-deterministic behavior on non-EN locales (e.g., Turkish “i”) and ensure consistent plugin order across environments.


src/Nethermind/Nethermind.Api/Extensions/PluginLoader.cs
```diff
@@ -108,19 +111,19 @@ public void OrderPlugins(IPluginConfig pluginConfig)
...
-            if (fPos == -1)
+            bool fHas = orderMap.TryGetValue(f.Name, out int fPos);
+            bool sHas = orderMap.TryGetValue(s.Name, out int sPos);
+            if (!fHas)
             {
-                if (sPos == -1)
+                if (!sHas)
                 {
-                    return f.Name.CompareTo(s.Name);
+                    return string.Compare(f.Name, s.Name, StringComparison.Ordinal);
```
LukaszRozmej: To be consistent: 
```suggestion
                    return string.Compare(f.Name, s.Name, StringComparison.OrdinalIgnoreCase);
```

src/Nethermind/Nethermind.Api/Extensions/PluginLoader.cs
```diff
@@ -108,19 +111,19 @@ public void OrderPlugins(IPluginConfig pluginConfig)
                 return 1;
             }
 
-            int fPos = order.IndexOf(f.Name.ToLower());
-            int sPos = order.IndexOf(s.Name.ToLower());
-            if (fPos == -1)
+            bool fHas = orderMap.TryGetValue(f.Name, out int fPos);
+            bool sHas = orderMap.TryGetValue(s.Name, out int sPos);
```
LukaszRozmej: Maybe 

```suggestion
            bool fHasExplicitOrder = orderMap.TryGetValue(f.Name, out int fPos);
            bool sHasExplicitOrder = orderMap.TryGetValue(s.Name, out int sPos);
```

LukaszRozmej (approved):

LukaszRozmej:
Ok I'm now trying to think does it really make that much sense...

1. Plugin's are never ordered by name - only looked up by it, thus locale shouldn't really matter
2. There is only a small handful of ordered plugins (literally a few), so not sure if worth paying for a Dictionary - probably doesn't matter much one way or the other.

src/Nethermind/Nethermind.Api/Extensions/PluginLoader.cs
```diff
@@ -108,19 +111,19 @@ public void OrderPlugins(IPluginConfig pluginConfig)
...
                 {
-                    return f.Name.CompareTo(s.Name);
+                    return string.Compare(f.Name, s.Name, StringComparison.OrdinalIgnoreCase);
                 }
 
                 return 1;
             }
 
-            if (sPos == -1)
+            if (!sHas)
```
LukaszRozmej: btw you didn't rename the usages of the flags.

LukaszRozmej:
This somewhat conflicts with https://github.com/NethermindEth/nethermind/pull/9264, which we need to resolve first

LukaszRozmej:
Closed in favor of https://github.com/NethermindEth/nethermind/pull/9308

PR #9264, opened 2025-09-10
Healthchecks plugin init first

svlachakis:
Initialise Health Checks Plugin first, in order to be able to always override properly by other modules like Arbitrum, regardless of the alphabetical order.

### DIFF
--- a/src/Nethermind/Nethermind.Api/Extensions/PluginLoader.cs
+++ b/src/Nethermind/Nethermind.Api/Extensions/PluginLoader.cs
@@ -94,6 +94,9 @@ public class PluginLoader(string pluginPath, IFileSystem fileSystem, ILogger log
         List<string> order = pluginConfig.PluginOrder.Select(s => s.ToLower() + "plugin").ToList();
         _pluginTypes.Sort((f, s) =>
         {
+            if (f.Name == "HealthChecksPlugin") return -1;
+            if (s.Name == "HealthChecksPlugin") return 1;
+
             bool fIsConsensus = typeof(IConsensusPlugin).IsAssignableFrom(f);
             bool sIsConsensus = typeof(IConsensusPlugin).IsAssignableFrom(s);
 


svlachakis:
@LukaszRozmej @asdacap concerning https://github.com/NethermindEth/nethermind/pull/9131#pullrequestreview-3125846845, We need to initialise HealthCheckPlugin always first, so Autofac can override properly from any other modules with the rule of "last initialised wins". 

Please let me and @damian-orzechowski know your thoughts!

svlachakis:
> What thing do healthcheckplugin need to override?

We don't want CL health checks in arbitrum like that https://github.com/NethermindEth/nethermind-arbitrum/pull/206

svlachakis:
> So... what is the problem with [NethermindEth/nethermind-arbitrum#206](https://github.com/NethermindEth/nethermind-arbitrum/pull/206)?

The problem is that by default Autofac's rule is to keep the latest initialisation and ArbitrumPlugin is initialized before HealthChecksPlugin because of Alphabetical order so override doesn't work.

svlachakis:
> Ahh... okok. Make sense. Can you move the `Priority` property in `IConsensusWrapperPlugin` to `INethermindPlugin` and use that instead? It would make the solution more reusable.

Sure, the question is more architectural whether we prefer this change in priority over this approach https://github.com/NethermindEth/nethermind/pull/9131

src/Nethermind/Nethermind.Api/Extensions/PluginLoader.cs
```diff
@@ -94,6 +94,9 @@ public void OrderPlugins(IPluginConfig pluginConfig)
         List<string> order = pluginConfig.PluginOrder.Select(s => s.ToLower() + "plugin").ToList();
         _pluginTypes.Sort((f, s) =>
         {
+            if (f.Name == "HealthChecksPlugin") return -1;
+            if (s.Name == "HealthChecksPlugin") return 1;
```
LukaszRozmej: Why not put it ahead in PluginOrder?

svlachakis: Since we all feel like that this approach is better, I'll make it work through priority - thanks!

svlachakis: @LukaszRozmej ArbitrumPlugin is IConsensusPlugin so it always goes at front

src/Nethermind/Nethermind.Api/Extensions/PluginLoader.cs
```diff
@@ -147,6 +107,13 @@ public async Task<IList<INethermindPlugin>> LoadPlugins(IConfigProvider configPr
 
         await using IContainer container = builder.Build();
         IList<INethermindPlugin> allPlugins = container.Resolve<IList<INethermindPlugin>>();
+
+        List<string> customOrder = configProvider.GetConfig<IPluginConfig>()
+            .PluginOrder
+            .ToList();
```
LukaszRozmej: no need for ToList

src/Nethermind/Nethermind.Api/Extensions/PluginLoader.cs
```diff
@@ -178,4 +145,34 @@ public async Task<IList<INethermindPlugin>> LoadPlugins(IConfigProvider configPr
...
+                    : name + "Plugin";
+                return (normalizedName, index);
+            })
+            .ToDictionary(x => x.normalizedName, x => x.index, StringComparer.OrdinalIgnoreCase);
+
+        return plugins
+            .OrderBy(p =>
+            {
+                if (p.GetType().Name == "HealthChecksPlugin")
+                    return -2000;
```
LukaszRozmej, line 164: Why still custom code?

svlachakis: Because if we want to keep the rule consensus first as number one, healthchecks need somehow to overrule that 😞 That's why I said maybe we need to simplify things here at standup

PR #9267, opened 2025-09-10
Correctly typed interface for XDC header decoder

ak88:
just gives Xdc header decoder correct return types and args.


LukaszRozmej:
Not sure how you intend to use it, I thought you wanted to reuse BlockDecoder with different HeaderDecoder, now you can't

PR #9270, opened 2025-09-11
Fix transaction processed counter not working

asdacap:
- Fix the counter used to skip prewarming not working.
- Removed the `TransactionProcessed` field and used a prewarmer specific decorator instead.

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

#### Notes on testing

- Verified to work with a custom metric and reversed tx order. No tx is skipped if not reversed.

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
@@ -428,17 +428,3 @@ public sealed class BlockCachePreWarmer(
         }
     }
 }
-
-public class PrewarmerTxAdapter(ITransactionProcessorAdapter baseAdapter, BlockCachePreWarmer preWarmer, IWorldState worldState) : ITransactionProcessorAdapter
-{
-    public TransactionResult Execute(Transaction transaction, ITxTracer txTracer)
-    {
-        if (worldState is IPreBlockCaches preBlockCaches && preBlockCaches.IsToBeWarmedWorldState)
-        {
-            preWarmer.OnBeforeTxExecution(transaction);
-        }
-        return baseAdapter.Execute(transaction, txTracer);
-    }
-
-    public void SetBlockExecutionContext(in BlockExecutionContext blockExecutionContext) => baseAdapter.SetBlockExecutionContext(in blockExecutionContext);
-}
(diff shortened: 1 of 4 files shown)


src/Nethermind/Nethermind.State/WorldState.cs
```diff
@@ -36,6 +36,7 @@ public class WorldState : IWorldState, IPreBlockCaches
         private bool _isInScope = false;
         private readonly ILogger _logger;
         private PreBlockCaches? PreBlockCaches { get; }
+        public bool IsToBeWarmedWorldState { get; }
```
LukaszRozmej:
```suggestion
        public bool WarmWorldState { get; }
```

Or `PopulateCaches`
Or `Enabled`

src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
```diff
@@ -109,6 +111,11 @@ private void PreWarmCachesParallel(Block suggestedBlock, BlockHeader parent, IRe
         }
     }
 
+    public void OnBeforeTxExecution(Transaction transaction)
+    {
+        _currentBlockState?.IncrementTransactionCounter();
```
asdacap, line 116: The `IsToBeWarmedWorldState` make sure that this is the main tx processing one. So it just need to increment the number.

src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
```diff
@@ -412,3 +429,16 @@ public void Dispose()
     }
 }
 
+public class PrewarmerTxAdapter(ITransactionProcessorAdapter baseAdapter, BlockCachePreWarmer preWarmer, IWorldState worldState) : ITransactionProcessorAdapter
```
LukaszRozmej: Move to own file?

src/Nethermind/Nethermind.State/IPreBlockCaches.cs
```diff
@@ -6,5 +6,5 @@ namespace Nethermind.State;
 public interface IPreBlockCaches
 {
     PreBlockCaches Caches { get; }
-    bool IsToBeWarmedWorldState { get; }
+    bool IsWarmWorldState { get; }
```
LukaszRozmej, line 9: Not proper english? `WarmWorldState` would be better? But we are i `IPreBlockCaches` I would like the name not to contain `WorldState` if possible.

PR #9272, opened 2025-09-11
Use extension method in some ITransactionProcessor interface method.

asdacap:
- Slightly simplify `ITransactionProcessor` interface by moving redundant method to extension class.

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

#### Notes on testing

- [x] Mainnet can start fine

### DIFF
--- a/src/Nethermind/Nethermind.AuRa.Test/Contract/ValidatorContractTests.cs
+++ b/src/Nethermind/Nethermind.AuRa.Test/Contract/ValidatorContractTests.cs
@@ -85,8 +85,9 @@ namespace Nethermind.AuRa.Test.Contract
 
             contract.FinalizeChange(_block.Header);
 
+            _transactionProcessor.Received().SetBlockExecutionContext(Arg.Is<BlockHeader>(header => header.Equals(_block.Header)));
             _transactionProcessor.Received().Execute(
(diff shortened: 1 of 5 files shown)


src/Nethermind/Nethermind.Evm/TransactionProcessing/ITransactionProcessor.cs
```diff
@@ -12,56 +12,74 @@ public interface ITransactionProcessor
...
 
     /// <summary>
     /// Call transaction, no validations, don't commit state
     /// Will NOT charge gas from sender account
     /// </summary>
     TransactionResult Warmup(Transaction transaction, ITxTracer txTracer);
-    TransactionResult Warmup(Transaction transaction, in BlockExecutionContext blockExecutionContext, ITxTracer txTracer)
+
+
+    void SetBlockExecutionContext(BlockHeader blockHeader);
```
LukaszRozmej, line 39: Do we really need this one in the interface?

asdacap: Yes, that is what some of the method do internally.

LukaszRozmej: But can we remove it and rely that the caller will give us release spec? It is just about release spec!

PR #9274, opened 2025-09-11
Make expected to fail hive tests list & fix gas accounting in eth_simulate

deffrian:
Fixes Closes Resolves #
Hive tests:
```
eth_simulateV1/ethSimulate-extcodehash-precompile
eth_simulateV1/ethSimulate-override-ecrecover
```
In these tests we override precompiles. Geth warms up all precompile addresses before execution, so they charge them as warm access regardless of whether they are actually precompiles. 

Hive tests pass if failing tests are in the failing list. This make it easier to catch regressions. Now if Hive action fails we actually have a regression, not just same failing tests that we had before

## Changes

- Adds known-failing-hove-test file with list of hove tests that we know to fail.
- Extend verification script
- Warmup precompiles. Not sure if it's good idea. Maybe we should ask geth to change or do something else to cover this case? Some links to geth code: [exclude precompiles from AL](https://github.com/ethereum/go-ethereum/blob/cbf0b5bc92c4ba857d79778f646f0b6b9507d07f/internal/ethapi/api.go#L1266), [warmup precompiles](https://github.com/ethereum/go-ethereum/blob/cbf0b5bc92c4ba857d79778f646f0b6b9507d07f/core/state/statedb.go#L1418)

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/.github/workflows/hive-tests.yml
+++ b/.github/workflows/hive-tests.yml
@@ -51,7 +51,7 @@ jobs:
     strategy:
       matrix:
         test-suite: ${{ fromJson(github.event.inputs.test-suite == '' &&
-          '["devp2p","ethereum/consensus","ethereum/eest","ethereum/engine","ethereum/graphql","ethereum/rpc","ethereum/rpc-compat","ethereum/sync"]' ||
+          '["devp2p","ethereum/engine","ethereum/graphql","ethereum/rpc","ethereum/rpc-compat","ethereum/sync"]' ||
           format('["{0}"]', github.event.inputs.test-suite)
           ) }}
     steps:


src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.cs
```diff
@@ -366,7 +366,7 @@ private static bool ChargeAccountAccessGas(ref long gasAvailable, VirtualMachine
             }
 
             // If the account is cold (and not a precompile), charge the cold access cost.
-            if (!vm.CodeInfoRepository.IsPrecompile(address, spec) && vmState.AccessTracker.WarmUp(address))
```
LukaszRozmej: Can we just make `CodeInfoRepository.IsPrecompile` work correctly with changed precompiles in that call?

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -661,7 +666,9 @@ private int ExecuteEvmCall<TTracingInst>(
 
             if (tracer.IsTracingAccess)
             {
-                tracer.ReportAccess(accessedItems.AccessedAddresses, accessedItems.AccessedStorageCells);
+                tracer.ReportAccess(
+                    accessedItems.AccessedAddresses.Except(spec.Precompiles.Select(addr => addr.Value)).ToArray(),
```
LukaszRozmej: That ToArray is boo

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.cs
deffrian: I'm not sure about that. It is no longer a precompile as we replaced it with code

LukaszRozmej: But it is still warmed up? Can you clarify the behavior?

I think it should be considered precompile for gas accounting as we want to have gas accounting similar to mainnet.

deffrian: I left references to geth code. They warmup precompiles. And charge less gas because they are warmed up, not because replaced precompile is a precompile. So in principle this is either our problem or geth problem, not sure who should change in this case

src/Nethermind/Nethermind.Blockchain/Tracing/AccessTxTracer.cs
```diff
@@ -28,8 +28,7 @@ public override void MarkAsFailed(Address recipient, GasConsumed gasSpent, byte[
         GasSpent += gasSpent.SpentGas;
     }
 
-    public override void ReportAccess(IReadOnlyCollection<Address> accessedAddresses,
-        IReadOnlyCollection<StorageCell> accessedStorageCells)
+    public override void ReportAccess(IEnumerable<Address> accessedAddresses, IEnumerable<StorageCell> accessedStorageCells)
```
LukaszRozmej, line 31: Rollback?

src/Nethermind/Nethermind.Evm/CodeInfoRepository.cs
```diff
@@ -27,13 +27,10 @@ protected CodeInfoRepository(FrozenDictionary<AddressAsKey, PrecompileInfo> prec
...
     }
 
-    public bool IsPrecompile(Address address, IReleaseSpec spec) =>
-        spec.IsPrecompile(address) && _localPrecompiles.ContainsKey(address);
-
     public ICodeInfo GetCachedCodeInfo(IWorldState worldState, Address codeSource, bool followDelegation, IReleaseSpec vmSpec, out Address? delegationAddress)
     {
         delegationAddress = null;
-        if (IsPrecompile(codeSource, vmSpec))
+        if (vmSpec.IsPrecompile(codeSource)) // _localPrecompiles have to have all precompiles
```
LukaszRozmej, line 34: They do so alternatively we can just do:

```suggestion
        if (_localPrecompiles.TryGetValue(codeSource, out PrecompileInfo precompile))
        {
            return precompile;
        }
```

And cut on one lookup

src/Nethermind/Nethermind.Blockchain/Tracing/AccessTxTracer.cs
deffrian: I think it's a good change regardless. No need for collection here

src/Nethermind/Nethermind.Evm/CodeInfoRepository.cs
deffrian: _localPrecompiles have all precompiles, even if they are not activated. So we have to check `spec.IsPrecompile`

benaadams (approved):

PR #9276, opened 2025-09-11
Refactor CalculateValueHash for header

ak88:
This is an attempt to make hash calculation more dynamic, since header hash is done through an extension method and cannot be overridden. 

An alternative could be to have a virtual method returning `ValueHash256` in `BlockHeader` itself, and have it return null in `BlockHeader` since the decoders doesn't live in Core namespace.


src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
```diff
@@ -317,6 +317,18 @@ public static Rlp Encode(Account item, RlpBehaviors behaviors = RlpBehaviors.Non
...
             => LogEntryDecoder.Instance.Encode(item, behaviors);
 
+        public static ValueHash256 EncodeForHash<T>(T item, RlpBehaviors behaviors = RlpBehaviors.None)
+        {
+            KeccakRlpStream stream = new();
+            IRlpStreamDecoder<T>? rlpStreamDecoder = GetStreamDecoder<T>();
+            if (rlpStreamDecoder is null)
+                throw new RlpException($"Decoder not found for {typeof(T).Name}");
+
+            int totalLength = rlpStreamDecoder.GetLength(item, behaviors);
```
LukaszRozmej: not used?

src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
```diff
@@ -317,6 +317,18 @@ public static Rlp Encode(Account item, RlpBehaviors behaviors = RlpBehaviors.Non
         public static Rlp Encode(LogEntry item, RlpBehaviors behaviors = RlpBehaviors.None)
             => LogEntryDecoder.Instance.Encode(item, behaviors);
 
+        public static ValueHash256 EncodeForHash<T>(T item, RlpBehaviors behaviors = RlpBehaviors.None)
+        {
+            KeccakRlpStream stream = new();
+            IRlpStreamDecoder<T>? rlpStreamDecoder = GetStreamDecoder<T>();
+            if (rlpStreamDecoder is null)
+                throw new RlpException($"Decoder not found for {typeof(T).Name}");
```
LukaszRozmej: Move to ThrowXXX method for performance

src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
```diff
@@ -317,6 +317,18 @@ public static Rlp Encode(Account item, RlpBehaviors behaviors = RlpBehaviors.Non
         public static Rlp Encode(LogEntry item, RlpBehaviors behaviors = RlpBehaviors.None)
             => LogEntryDecoder.Instance.Encode(item, behaviors);
 
+        public static ValueHash256 EncodeForHash<T>(T item, RlpBehaviors behaviors = RlpBehaviors.None)
+        {
+            KeccakRlpStream stream = new();
+            IRlpStreamDecoder<T>? rlpStreamDecoder = GetStreamDecoder<T>();
```
LukaszRozmej: bit of cost for every hash calculation, but we probably have to pay it

src/Nethermind/Nethermind.Crypto/BlockHeaderExtensions.cs
```diff
@@ -17,6 +17,9 @@ public static Hash256 CalculateHash(this BlockHeader header, RlpBehaviors behavi
 
         public static ValueHash256 CalculateValueHash(this BlockHeader header, RlpBehaviors behaviors = RlpBehaviors.None)
         {
+            if (header is IHashResolver resolver)
+                return resolver.CalculateHash();
```
LukaszRozmej, line 21: Actually a good idea, but:
1. Drop the interface
2. Put method on class
3. Drop this extension method
4. Use static instance of decoder, not to allocate

Should be good.

PR #9279, opened 2025-09-12
Fix inconsistent test.

asdacap:
- Fix randomly failing test due to genesis loader waiting for `NewHead` event while `AddBlock` wait for `BlockAdded` event.
- Ensure that the head after the context init is of the right block number.
- Ensure TX pool tx removed at end of `AddBlock`.
- Make components lazy to make unit test run faster.
- Do not allow additional tx in the created block unless explicitly allowed.
- Fail test when invalid block is detected in `AddBlock`.

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Waiting....

### DIFF
--- a/src/Nethermind/Nethermind.Core.Test/Blockchain/InvalidBlockDetector.cs
+++ b/src/Nethermind/Nethermind.Core.Test/Blockchain/InvalidBlockDetector.cs
@@ -13,19 +13,14 @@ namespace Nethermind.Core.Test.Blockchain;
 
 public class InvalidBlockDetector
 {
-    public event EventHandler<InvalidBlockEventArgs>? OnInvalidBlock;
+    public event EventHandler<IBlockchainProcessor.InvalidBlockEventArgs>? OnInvalidBlock;
 
     private void TriggerOnInvalidBlock(Block invalidBlock)
     {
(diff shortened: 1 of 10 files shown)


src/Nethermind/Nethermind.Core.Test/Blockchain/InvalidBlockDetector.cs
```diff
@@ -0,0 +1,47 @@
...
+        OnInvalidBlock?.Invoke(this, new InvalidBlockEventArgs()
+        {
+            InvalidBlock = invalidBlock
+        });
+    }
+
+    public class InvalidBlockEventArgs : EventArgs
+    {
+        public Block InvalidBlock { get; init; } = null!;
+    }
```
LukaszRozmej: You can reuse `BlockEventArgs`

LukaszRozmej (approved):
Looks compliecated

src/Nethermind/Nethermind.Core.Test/Blockchain/InvalidBlockDetector.cs
asdacap: Done

PR #9290, opened 2025-09-15
Enforce EIP-7934 block size limit in TxPoolTxSource

fredrik0x:
## Changes

Adds logic to track and enforce the EIP-7934 maximum RLP block size during transaction selection, preventing oversized blocks when building a local block (looking at the code, it appears this is already in place during initial sync or when receiving blocks or validating a mev builder block). Introduces a method to calculate the RLP-encoded size of transactions for accurate block size accounting.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

Can look at creating some tests if you feel this PR makes sense.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Producers/TxPoolTxSource.cs
+++ b/src/Nethermind/Nethermind.Consensus/Producers/TxPoolTxSource.cs
@@ -17,6 +17,7 @@ using Nethermind.Core.Specs;
 using Nethermind.Evm;
 using Nethermind.Int256;
 using Nethermind.Logging;
+using Nethermind.Serialization.Rlp;
 using Nethermind.TxPool;
 using Nethermind.TxPool.Comparison;
 
@@ -60,6 +61,11 @@ namespace Nethermind.Consensus.Producers
             int checkedTransactions = 0;
             int selectedTransactions = 0;
             using ArrayPoolList<Transaction> selectedBlobTxs = new((int)spec.MaxBlobCount);
+            
+            // EIP-7934: Block size tracking to prevent creating oversized blocks
+            const int MAX_HEADER_SIZE = 1024; // Conservative estimate for header, receipts, and other block overhead
+            int cumulativeBlockSize = MAX_HEADER_SIZE;
+            int maxRlpBlockSize = spec.IsEip7934Enabled ? spec.Eip7934MaxRlpBlockSize : int.MaxValue;
 
             SelectBlobTransactions(blobTransactions, parent, spec, baseFee, selectedBlobTxs);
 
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Consensus/Producers/TxPoolTxSource.cs
```diff
@@ -74,6 +80,19 @@ public IEnumerable<Transaction> GetTransactions(BlockHeader parent, long gasLimi
...
 
+                // EIP-7934: Check if adding this transaction would exceed block size limit
+                if (spec.IsEip7934Enabled)
+                {
+                    int txSize = GetTransactionSizeForBlockInclusion(tx);
+                    if (cumulativeBlockSize + txSize > maxRlpBlockSize)
+                    {
+                        if (_logger.IsDebug) 
+                            _logger.Debug($"Stopping transaction selection - would exceed block size limit. Current: {cumulativeBlockSize} bytes, Tx would add: {txSize} bytes, Limit: {maxRlpBlockSize} bytes");
+                        break; // Stop adding transactions to prevent oversized block
```
LukaszRozmej, line 91: We probably could check few more transactions, and stop only if we can't add X of them, as we can have smaller transactions that will fit the limit after.

src/Nethermind/Nethermind.Consensus/Producers/TxPoolTxSource.cs
```diff
@@ -444,6 +463,17 @@ internal static IEnumerable<Transaction> Order(IDictionary<AddressAsKey, Transac
...
+        /// <summary>
+        /// Calculates the exact RLP-encoded size of a transaction for block inclusion.
+        /// This is used for EIP-7934 block size limit enforcement.
+        /// </summary>
+        private static int GetTransactionSizeForBlockInclusion(Transaction tx)
+        {
+            // Use the exact RLP encoding size from TxDecoder
+            // RlpBehaviors.None gives us the standard encoding for block inclusion
+            return TxDecoder.Instance.GetLength(tx, RlpBehaviors.None);
+        }
```
LukaszRozmej, line 476: Is this precise measurment?

src/Nethermind/Nethermind.Consensus/Producers/TxPoolTxSource.cs
```diff
@@ -60,6 +61,11 @@ public IEnumerable<Transaction> GetTransactions(BlockHeader parent, long gasLimi
             int checkedTransactions = 0;
             int selectedTransactions = 0;
             using ArrayPoolList<Transaction> selectedBlobTxs = new((int)spec.MaxBlobCount);
+            
+            // EIP-7934: Block size tracking to prevent creating oversized blocks
+            const int MAX_HEADER_SIZE = 1024; // Conservative estimate for header, receipts, and other block overhead
```
LukaszRozmej, line 66: Does header have estimate? Can't we know the exact size of the header (or exact max size)

LukaszRozmej (commented):
No test provided

src/Nethermind/Nethermind.Consensus/Producers/TxPoolTxSource.cs
marcindsobczak: It should be, but worth to add test and compare calculated value with value of prepared block

src/Nethermind/Nethermind.Consensus/Producers/TxPoolTxSource.cs
marcindsobczak: Or add one more check
if (txSize > minTxSize) continue;
else break;

minTxSize I think will be legacy tx without data

marcindsobczak (commented):
Looks good, but we definitely need tests

flcl42:
Instead of receiving `long gasLimit` in `public IEnumerable<Transaction> GetTransactions(BlockHeader parent, long gasLimit, PayloadAttributes? payloadAttributes = null, bool filterSource = false)` we can pass the entire block header to figure out exact header's size. 
Another thing is that we can pass multiple filters into `private static IEnumerable<(Transaction tx, long resource)> OrderCore(` instead of just gasLimit filter. I mean you acn try to add blockSizeLimit in addition to gasLimit and the method may receive something like `List<(long resourceLimit, Func<Transaction, long> resourceSelector)> resourceLimits` or so

PR #9291, opened 2025-09-16
Add code to `IWorldStateScopeProvider`.

asdacap:
- Add code to `IWorldStateScopeProvider`.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- [ ] Era replay


src/Nethermind/Nethermind.Evm/State/IWorldStateScopeProvider.cs
```diff
@@ -96,4 +108,9 @@ public interface IStateSetter : IDisposable
     {
         void Set(Address key, Account account);
     }
+
+    public interface ICodeSetter : IDisposable
+    {
+        void Set(in ValueHash256 codeHash, byte[] code);
```
LukaszRozmej: Should we keep it ac array or do spans?

src/Nethermind/Nethermind.Evm/State/IWorldStateScopeProvider.cs
```diff
@@ -57,6 +62,13 @@ void Set(Address key, Account account)
         void UpdateRootHash();
     }
 
+    public interface ICodeDb
+    {
+        byte[]? GetCode(in ValueHash256 codeHash);
```
LukaszRozmej, line 67: Should we keep it ac array or do spans?
(Same with `IStorageTree`, should it be `byte[]` or `ReadOnlySpan<byte>` or `Uint256` or `ValueHash256`?)

asdacap: We can change it to `Memory<byte>` as the EVM and `ICodeInfoRepository` take `Memory`. Probably not span though.

src/Nethermind/Nethermind.Evm/State/IWorldStateScopeProvider.cs
asdacap: Yea, we can do span here

src/Nethermind/Nethermind.Evm/State/IWorldStateScopeProvider.cs
LukaszRozmej: do the most perfomant

LukaszRozmej (approved):

PR #9299, opened 2025-09-16
XDC : EpochSwitchInfo Manager

Demuirgos:
Fixes Closes Resolves #

## Changes

- _List the changes_

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Remarks

### DIFF
--- a/src/Nethermind/Nethermind.Xdc/TimeoutCertificateManager.cs
+++ b/src/Nethermind/Nethermind.Xdc/TimeoutCertificateManager.cs
@@ -45,7 +45,7 @@ public class TimeoutCertificateManager(XdcContext context, ISnapshotManager snap
         var collectedTimeouts = _timeouts.GetItems(timeout);
 
         var xdcHeader = _blockTree.Head?.Header as XdcBlockHeader;
-        EpochSwitchInfo epochSwitchInfo = _epochSwitchManager.GetEpochSwitchInfo(xdcHeader, xdcHeader.Hash);
+        EpochSwitchInfo epochSwitchInfo = _epochSwitchManager.GetEpochSwitchInfo(xdcHeader);
         if (epochSwitchInfo is null)
         {
             // Failed to get epoch switch info, cannot process timeout
@@ -201,14 +201,14 @@ public class TimeoutCertificateManager(XdcContext context, ISnapshotManager snap
         var currentHeader = (XdcBlockHeader)_blockTree.Head?.Header;
         if (currentHeader is null) throw new InvalidOperationException("Failed to retrieve current header");
         IXdcReleaseSpec spec = _specProvider.GetXdcSpec(currentHeader, _ctx.CurrentRound);
-        if (_epochSwitchManager.IsEpochSwitchAtRound(_ctx.CurrentRound, currentHeader, out ulong epochNumber))
+        if (_epochSwitchManager.IsEpochSwitchAtRound(_ctx.CurrentRound, currentHeader))
         {
             ulong currentNumber = (ulong)currentHeader.Number + 1;
             gapNumber = Math.Max(0, currentNumber - currentNumber % (ulong)spec.EpochLength - (ulong)spec.Gap);
         }
         else
         {
-            EpochSwitchInfo epochSwitchInfo = _epochSwitchManager.GetEpochSwitchInfo(currentHeader, currentHeader.Hash);
+            EpochSwitchInfo epochSwitchInfo = _epochSwitchManager.GetEpochSwitchInfo(currentHeader);
             if (epochSwitchInfo is null)
                 throw new ConsensusHeaderDataExtractionException(nameof(EpochSwitchInfo));
 
@@ -230,7 +230,7 @@ public class TimeoutCertificateManager(XdcContext context, ISnapshotManager snap
     private bool AllowedToSend()
     {
         var currentHeader = (XdcBlockHeader)_blockTree.Head?.Header;
-        EpochSwitchInfo epochSwitchInfo = _epochSwitchManager.GetEpochSwitchInfo(currentHeader, currentHeader.Hash);
+        EpochSwitchInfo epochSwitchInfo = _epochSwitchManager.GetEpochSwitchInfo(currentHeader);
         if (epochSwitchInfo is null)
             return false;
         return epochSwitchInfo.Masternodes.Contains(_signer.Address);


src/Nethermind/Nethermind.Core/Specs/ISpecProvider.cs
```diff
@@ -2,6 +2,8 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 
 using Nethermind.Int256;
+using System.Runtime.CompilerServices;
+[assembly: InternalsVisibleTo("Nethermind.Xdc.Test")]
```
ak88: We dont want this for XDC only. You can mock the ISpecProvider just like in `QuorumCertificateManagerTest`

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
```diff
@@ -0,0 +1,452 @@
...
+        _xdcSpecProvider = xdcSpecProvider;
+        _tree = tree;
+        _snapshotManager = snapshotManager;
+    }
+
+    private ISpecProvider _xdcSpecProvider { get; }
+    private IBlockTree _tree { get; }
+    private ISnapshotManager _snapshotManager { get; }
+    private ConcurrentDictionary<ulong, BlockInfo> _round2EpochBlockInfo { get; set; } = new();
+    private ConcurrentDictionary<ValueHash256, EpochSwitchInfo> _epochSwitches { get; set; } = new();
```
ak88: Needs to be LRU cache instead

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+        }
+
+        return false;
+    }
+
+    public bool IsEpochSwitchAtRound(ulong currentRound, XdcBlockHeader parent, out ulong epochNumber)
+    {
+        var xdcSpec = _xdcSpecProvider.GetXdcSpec(parent);
+
+        epochNumber = ((ulong)xdcSpec.SwitchEpoch + currentRound) / (ulong)xdcSpec.EpochLength;
```
cicr99: I believe it should be `epochNumber = (ulong)xdcSpec.SwitchEpoch + currentRound / (ulong)xdcSpec.EpochLength;`
You  divide the currentRound by the epochLength to get the amount of epochs and then add that to the switchEpoch number

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+            }
+
+            iteratorHash = epochSwitchInfo.EpochSwitchParentBlockInfo.Hash;
+            iteratorBlockNumber = epochSwitchInfo.EpochSwitchBlockInfo.BlockNumber;
+
+            if (iteratorBlockNumber >= start.Number)
+            {
+                epochSwitchInfos.Add(epochSwitchInfo);
+            }
+        }
```
cicr99: Once you set `iteratorHeader = null`, it's not updated again, so the call to `GetEpochSwitchInfo` above probably won't work after the first iteration. I think you could do it with the hash instead and just forget about the header, as it's not used anywhere else.

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+
+    public EpochSwitchInfo? GetTimeoutCertificateEpochInfo(TimeoutCertificate timeoutCert)
+    {
+        EpochSwitchInfo epochSwitchInfo = GetEpochSwitchInfo((XdcBlockHeader)_tree.Head.Header);
+        if (epochSwitchInfo is null)
+        {
+            return null;
+        }
+
+        var xdcSpec = _xdcSpecProvider.GetXdcSpec((XdcBlockHeader)_tree.Head.Header);
```
cicr99: If you define a variable for the header, you only have to get it once

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+            if ((epochSwitchInfo = GetEpochSwitchInfo(epochSwitchInfo.EpochSwitchParentBlockInfo.Hash)) is null)
+            {
+                return null;
+            }
+        }
+
+        return epochSwitchInfo;
+    }
+
+    internal bool TryGetBlockByEpochNumber(ulong tempTCEpoch, out Types.BlockRoundInfo epochBlockInfo)
```
cicr99:
```suggestion
    internal bool TryGetBlockByEpochNumber(ulong tempTCEpoch, out BlockInfo epochBlockInfo)
```

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+        }
+
+        var epoch = (ulong)xdcSpec.EpochLength;
+        ulong estBlockNumDiff = epoch * (epochRound - tempTCEpoch);
+        long estBlockNum = epochSwitchInfo.EpochSwitchBlockInfo.BlockNumber - (long)estBlockNumDiff;
+
+        if (estBlockNum < xdcSpec.SwitchBlock)
+        {
+            estBlockNum = (long)xdcSpec.SwitchBlock;
+        }
```
cicr99: You can use Math.Max function instead

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+                return null;
+            }
+        }
+
+        return GetEpochSwitchInfo(epochBlockInfo.Hash);
+    }
+
+    public BlockInfo? GetBlockByEpochNumber(ulong epochNumber)
+    {
+        if (TryGetBlockByEpochNumber(epochNumber, out var epochBlockInfo))
```
cicr99: I would say this function is not that helpful, as at the end every time the result is false you set the epochBlockInfo to null and viceversa, and whoever uses GetBlockByEpochNumber will check whether the result is null anyway. I would just move the logic here

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+                    epochBlockInfo = info.EpochSwitchBlockInfo;
+                    return true;
+                }
+            }
+        }
+
+        return TryBinarySearchBlockByEpochNumber(tempTCEpoch, estBlockNum, epochSwitchInfo.EpochSwitchBlockInfo.BlockNumber, (ulong)xdcSpec.SwitchBlock, (ulong)xdcSpec.EpochLength, out epochBlockInfo);
+    }
+
+    private bool TryGetBlockInfoInCache(ulong estRound, ulong epoch, out BlockInfo epochBlockInfo)
```
cicr99: Same reasoning with this one, I would just leave it as `private BlockInfo? GetBlockInfoInCache(ulong estRound, ulong epoch)`

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+            {
+                epochBlockInfo = null;
+                return false;
+            }
+
+            bool isEpochSwitch = IsEpochSwitchAtBlock(header, out ulong epochNum);
+
+            if (epochNum == tempTCEpoch)
+            {
+                if (!Utils.TryGetExtraFields(header, (long)switchBlock, out var extraFieldsV2, out _))
```
cicr99: Why use this one here if you only care about the round? And in similar situations above you'd do `if (header.ExtraConsensusData is null)` instead? Just saying in order to keep it consistent

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+                }
+                else
+                {
+                    end = header.Number;
+                    // trick to shorten the search
+                    ulong estStart = (ulong)end - (round % epoch);
+                    if (start < (long)estStart)
+                    {
+                        start = (long)estStart;
+                    }
```
cicr99: You could use max function here as well

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+                }
+            }
+            else if (epochNum > tempTCEpoch)
+            {
+                end = header.Number;
+            }
+            else if (epochNum < tempTCEpoch)
```
cicr99: The condition can be omitted here, as this is the only possible case left, and just leave it as `else`

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+        _snapshotManager = snapshotManager;
+    }
+
+    private ISpecProvider _xdcSpecProvider { get; }
+    private IBlockTree _tree { get; }
+    private ISnapshotManager _snapshotManager { get; }
+    private ConcurrentDictionary<ulong, BlockInfo> _round2EpochBlockInfo { get; set; } = new();
+    private ConcurrentDictionary<ValueHash256, EpochSwitchInfo> _epochSwitches { get; set; } = new();
+
+    public bool IsEpochSwitchAtBlock(XdcBlockHeader header, out ulong epochNumber)
```
ak88: There is already a method in `XdcHeader` for this

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+        {
+            _round2EpochBlockInfo[round] = new BlockInfo(header.Hash, round, header.Number);
+            return true;
+        }
+
+        return false;
+    }
+
+    public bool IsEpochSwitchAtRound(ulong currentRound, XdcBlockHeader parent, out ulong epochNumber)
+    {
```
ak88: Do we need these out parameters? I dont think they are useful.

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+        if (parentRound < epochStartRound)
+        {
+            _round2EpochBlockInfo[round] = new BlockInfo(header.Hash, round, header.Number);
+            return true;
+        }
+
+        return false;
+    }
+
+    public bool IsEpochSwitchAtRound(ulong currentRound, XdcBlockHeader parent, out ulong epochNumber)
```
ak88: Would be very helpful with a comment here. This checks if a round without an associated header is an epoch switch, which is used when block building eg.

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+        if (currentRound <= parentRound)
+        {
+            return false;
+        }
+
+        ulong epochStartRound = currentRound - (currentRound % (ulong)xdcSpec.EpochLength);
+        return parentRound < epochStartRound;
+    }
+
+    public (ulong currentCheckpointNumber, ulong epochNumber)? GetCurrentEpochNumbers(ulong blockNumber)
```
ak88: Is this useful?

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+
+            if (iteratorBlockNumber >= start.Number)
+            {
+                epochSwitchInfos.Add(epochSwitchInfo);
+            }
+        }
+
+        epochSwitchInfos.Reverse();
+        return epochSwitchInfos.ToArray();
+    }
```
ak88: This is only used in API and not necessary for consensus, so dont spend time fixing this. You can remove it and park it on a separate branch maybe.

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+
+    public EpochSwitchInfo? GetEpochSwitchInfo(XdcBlockHeader header)
+    {
+        var xdcSpec = _xdcSpecProvider.GetXdcSpec(header);
+
```
ak88: Why is the cache not checked here?

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+        while (!IsEpochSwitchAtBlock(header, out _))
+        {
+            header = (XdcBlockHeader)_tree.FindHeader(header.ParentHash);
+        }
+
+        if (header.Number == 0)
+        {
+            // genesis handling
+            return _epochSwitches[header.Hash] = new EpochSwitchInfo(Utils.GetMasternodesFromGenesisHeader(_tree, header), [], [], new BlockRoundInfo(header.Hash, 0, header.Number));
```
ak88: This doesnt make sense. We do not support the actual XDC genesis, so we probably just want to have a predefined epoch switch.

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+        {
+            epochSwitchInfo.EpochSwitchParentBlockInfo = consensusData.QuorumCert.ProposedBlockInfo;
+        }
+
+        return _epochSwitches[header.Hash] = epochSwitchInfo;
+    }
+
+    public EpochSwitchInfo? GetEpochSwitchInfo(Hash256 hash)
+    {
+        if (_epochSwitches.TryGetValue(hash, out var epochSwitchInfo) && epochSwitchInfo is not null)
```
ak88: are there any instance where epochSwitchInfo can be null?

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+            header = (XdcBlockHeader)_tree.FindHeader(header.ParentHash);
+        }
+
+        if (header.Number == 0)
+        {
+            // genesis handling
+            return _epochSwitches[header.Hash] = new EpochSwitchInfo(Utils.GetMasternodesFromGenesisHeader(_tree, header), [], [], new BlockRoundInfo(header.Hash, 0, header.Number));
+        }
+
+        if (!Utils.TryGetExtraFields(header, (long)xdcSpec.SwitchBlock, out ExtraFieldsV2 consensusData, out Address[] masterNodes))
```
ak88: Methods already exists for getting this data

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+        XdcBlockHeader h = (XdcBlockHeader)_tree.FindHeader(hash);
+        if (h is null)
+        {
+            return null;
+        }
+
+        return GetEpochSwitchInfo(h);
+    }
+
+    public EpochSwitchInfo? GetPreviousEpochSwitchInfoByHash(Hash256 blockHash, int limit)
```
ak88: Where is this used and why is it useful?

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
ak88: tempTCEpoch name doesnt make sense. This is the target epoch you want to find a block for.

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
```diff
@@ -0,0 +1,454 @@
...
+        var xdcSpec = _xdcSpecProvider.GetXdcSpec(headHeader);
+
+        EpochSwitchInfo epochSwitchInfo = GetEpochSwitchInfo(headHeader);
+        if (epochSwitchInfo is null)
+        {
+            epochBlockInfo = null;
+            return false;
+        }
+
+        ulong epochRound = (ulong)xdcSpec.SwitchBlock + epochSwitchInfo.EpochSwitchBlockInfo.Round / (ulong)xdcSpec.EpochLength;
```
ak88: This is not correct. You need to calculate the epoch number and not the round. 
V2 start epoch + number of epochs since then.

src/Nethermind/Nethermind.Xdc/Utils.cs
```diff
@@ -0,0 +1,120 @@
...
+
+        if (consensusData is null)
+        {
+            return false;
+        }
+
+        return true;
+    }
+
+    private static Address[] GetMasterNodesFromHeaderExtra(XdcBlockHeader header)
```
ak88: this looks like V1 method which we dont support.

ak88 (changes requested):
Lets get rid of those out epoch number parameters. We can make a small helper method at some point if we feel its useful info to log.

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
Demuirgos: hmm this is used by GetBlockByEpochNumber, which iirc is used somewhere in TC manager

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
Demuirgos: removed

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
Demuirgos: just extra guard not necessary, removed

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
Demuirgos: added cache check

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
Demuirgos: does it makes sense for it to be in the Header or in EpochSwitchManager? imo I think it should be in ESM rather than the header

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
Demuirgos: Done

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
Demuirgos: they are there just to avoid recalculations in case, just for convenience

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
Demuirgos: Fixed

src/Nethermind/Nethermind.Xdc/XdcExtensions.cs
```diff
@@ -68,7 +69,43 @@ public static IXdcReleaseSpec GetXdcSpec(this ISpecProvider specProvider, XdcBlo
...
 
+        Address[] addresses = new Address[data.Length / Address.Size];
+        for (int i = 0; i < addresses.Length; i++)
+        {
+            addresses[i] = new Address(data.Slice(i * Address.Size, Address.Size));
+        }
+        return addresses.ToImmutableArray();
+    }
+
+    public static T[] RemoveItemFromArray<T>(T[] candidates, T[] penalties, int withMaxCap = int.MaxValue)
```
ak88: not an extension

src/Nethermind/Nethermind.Xdc/XdcExtensions.cs
```diff
@@ -68,7 +69,43 @@ public static IXdcReleaseSpec GetXdcSpec(this ISpecProvider specProvider, XdcBlo
 
         return snapshotManager.GetSnapshot(gapBlockHash);
     }
+    public static ImmutableArray<Address>? ExtractAddresses(Span<byte> data)
```
ak88: this is not an extension.

src/Nethermind/Nethermind.Xdc/EpochSwitchManager.cs
```diff
@@ -0,0 +1,410 @@
...
+        while (!IsEpochSwitchAtBlock(header))
+        {
+            header = (XdcBlockHeader)_tree.FindHeader(header.ParentHash);
+        }
+
+        Address[] masterNodes;
+
+        if (header.Number == xdcSpec.SwitchBlock)
+        {
+            masterNodes = XdcExtensions.ExtractAddresses(header.ExtraData[XdcConstants.ExtraVanity..^XdcConstants.ExtraSeal]).Value.ToArray();
```
ak88, line 116: I dont think we want this, but rather just have the starting masternodes defined in spec. We can change that later though.

PR #9301, opened 2025-09-16
evm changes to integrate stylus execution

tanishqjasoria:
Fixes Closes Resolves #

- add EvmState.RentTopLevelFrame, for stylus we need a TopLevel frame to execute *CALL or CREATE. It enables us to use Handle* functions to  terminate the execution and return control back to rust
- make EvmInstructions public to access ChargeAccountAccessGas

## Changes

- _List the changes_

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

#### Notes on testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks


src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.cs
```diff
@@ -352,7 +352,7 @@ private static bool ChargeAccountAccessGasWithDelegation(ref long gasAvailable,
     /// <param name="address">The target account address.</param>
     /// <param name="chargeForWarm">If true, applies the warm read gas cost even if the account is warm.</param>
     /// <returns>True if the gas charge was successful; otherwise false.</returns>
-    private static bool ChargeAccountAccessGas(ref long gasAvailable, VirtualMachine vm, Address address, bool chargeForWarm = true)
+    public static bool ChargeAccountAccessGas(ref long gasAvailable, VirtualMachine vm, Address address, bool chargeForWarm = true)
```
tanishqjasoria: just needed to make EvmInstructions public to use this method

LukaszRozmej: Maybe better to move it to its own class?

src/Nethermind/Nethermind.Evm/EvmState.cs
```diff
@@ -213,6 +213,35 @@ static void ThrowIsInUse()
...
+        in StackAccessTracker stateForAccessLists,
+        in Snapshot snapshot)
+    {
+        EvmState state = Rent();
+        state.Initialize(
+            gasAvailable,
+            outputDestination,
+            outputLength,
+            executionType,
+            isTopLevel: true,
```
LukaszRozmej: So is it top level or not? Comment says "beneath top level"

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.cs
tanishqjasoria: then we should move all these gas calcuation to another class?

src/Nethermind/Nethermind.Evm/EvmState.cs
tanishqjasoria: it is not technically top level, but it acts like top level.
we need to terminate the execution when we are done executing this frame

LukaszRozmej: Maybe just add `isTopLevel` to `RentFrame` with default as false?

benaadams: Agreed

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructionsUtils.cs
```diff
@@ -0,0 +1,229 @@
...
+using Nethermind.Core;
+using Nethermind.Core.Specs;
+using Nethermind.Int256;
+
+namespace Nethermind.Evm;
+
+
+
+
+public static class EvmInstructionsUtils
```
LukaszRozmej: Maybe better naming? Open to other suggestions.

```suggestion
public static class EvmCalculations
```

LukaszRozmej (approved):

benaadams (approved):

PR #9306, opened 2025-09-17
Feature/worldstate scope provider manager

asdacap:
- Require #9089 
- Change `IWorldStateManager` to return `IWorldStateScopeProvider` instead of `IWorldState`.
- Extract prewarmer logic out from `WorldState` to `PrewarmerScopeProvider`.  
- `WorldState` is now a block processing components, like `TransactionProcessor` etc. 

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

#### Notes on testing

- [x] Mainnet can sync
- [x] Mainnet era replay.

### DIFF
--- a/src/Nethermind/Nethermind.State/PrewarmerScopeProvider.cs
+++ b/src/Nethermind/Nethermind.State/PrewarmerScopeProvider.cs
@@ -77,6 +77,7 @@ public class PrewarmerScopeProvider(
             {
                 if (preBlockCache?.TryGetValue(addressAsKey, out Account? account) ?? false)
                 {
+                    baseScope.HintGet(address, account);
                     Metrics.IncrementStateTreeCacheHits();
                 }
                 else
@@ -86,6 +87,8 @@ public class PrewarmerScopeProvider(
                 return account;
             }
         }
+
+        public void HintGet(Address address, Account? account) => baseScope.HintGet(address, account);
 
         private Account? GetFromBaseTree(AddressAsKey address)
         {
@@ -122,6 +125,7 @@ public class PrewarmerScopeProvider(
             {
                 if (preBlockCache?.TryGetValue(storageCell, out byte[] value) ?? false)
                 {
+                    baseStorageTree.HintGet(index, value);
                     Db.Metrics.IncrementStorageTreeCache();
                 }
                 else
@@ -131,6 +135,8 @@ public class PrewarmerScopeProvider(
                 return value;
             }
         }
+
+        public void HintGet(in UInt256 index, byte[]? value) => baseStorageTree.HintGet(in index, value);
 
         private byte[] LoadFromTreeStorage(StorageCell storageCell)
         {


src/Nethermind/Nethermind.Core.Test/TestWorldStateFactory.cs
```diff
@@ -14,13 +14,10 @@ namespace Nethermind.Core.Test;
...
-    public static IWorldState CreateForTest()
-    {
-        return CreateForTest(TestMemDbProvider.Init(), LimboLogs.Instance);
-    }
-
-    public static IWorldState CreateForTest(IDbProvider dbProvider, ILogManager logManager)
+    public static IWorldState CreateForTest(IDbProvider? dbProvider = null, ILogManager? logManager = null)
     {
+        if (dbProvider is null) dbProvider = TestMemDbProvider.Init();
+        if (logManager is null) logManager = LimboLogs.Instance;
```
LukaszRozmej:
```suggestion
        dbProvider ??= TestMemDbProvider.Init();
        logManager ??= LimboLogs.Instance;
```

src/Nethermind/Nethermind.Init/Steps/InitializeBlockchain.cs
```diff
@@ -33,6 +36,7 @@ namespace Nethermind.Init.Steps
     public class InitializeBlockchain(INethermindApi api, IChainHeadInfoProvider chainHeadInfoProvider) : IStep
     {
         private readonly INethermindApi _api = api;
+        private ILogManager _logManager = api.LogManager;
```
LukaszRozmej, line 39: Not used?

src/Nethermind/Nethermind.State/PrewarmerScopeProvider.cs
```diff
@@ -0,0 +1,151 @@
...
+
+                if (Metrics.ThreadLocalStateTreeReads == priorReads)
+                {
+                    Metrics.IncrementStateTreeCacheHits();
+                }
+                return account;
+            }
+            else
+            {
+                if (preBlockCache?.TryGetValue(addressAsKey, out Account? account) ?? false)
```
LukaszRozmej, line 78: My preference, feel free to ignore:
```suggestion
                if (preBlockCache?.TryGetValue(addressAsKey, out Account? account) == true)
```

src/Nethermind/Nethermind.State/PrewarmerScopeProvider.cs
```diff
@@ -0,0 +1,151 @@
...
+        ConcurrentDictionary<StorageCell, byte[]> preBlockCache,
+        Address address,
+        bool populatePreBlockCache
+    ) : IWorldStateScopeProvider.IStorageTree
+    {
+        public Hash256 RootHash => baseStorageTree.RootHash;
+
+        public byte[] Get(in UInt256 index)
+        {
+            StorageCell storageCell = new StorageCell(address, in index); // TODO: Make the dictionary use UInt256 directly
```
LukaszRozmej, line 110: To do that we would have to have cache per contract, which is a lot of dictionaries potentially

src/Nethermind/Nethermind.State/PrewarmerScopeProvider.cs
```diff
@@ -0,0 +1,151 @@
...
+using System.Collections.Concurrent;
+using Nethermind.Core;
+using Nethermind.Core.Crypto;
+using Nethermind.Db;
+using Nethermind.Evm.State;
+using Nethermind.Int256;
+
+namespace Nethermind.State;
+
+public sealed class PrewarmerScopeProvider(
```
LukaszRozmej: Are you sure this works fine? Looks like it might but did you test it well?
Caches need to be in exact same place they were before in terms of between State/Storage Provider and Tree.

LukaszRozmej (commented):
Fix tests?

Should I use this PR as base for my block-STM? As I mess with state and need a stable target.

src/Nethermind/Nethermind.State/PrewarmerScopeProvider.cs
asdacap: Good point

asdacap:
See https://github.com/NethermindEth/nethermind/pull/9402/files#diff-9e552dbd726d749d8e36fe2e31e04926cc2442fd62972cc2d9e1dde24df890b9 for example on how to intercept read and writes.

Issue #9280, opened 2025-09-12
Taiko not syncing properly


stdevMac:
**Description**
We are trying to fetch state before finishing snap sync

**Steps to Reproduce**
In order to replicate the behavior, please provide a detailed list of steps:
1. Run a Taiko Node
2. After some time of snap sync, without finish, we will see errors that we are trying to fetch state.

**Actual behavior**
We are trying to fetch the state before finalizing the snap sync

**Expected behavior**
Sync, and then fetch the state

**Desktop (please complete the following information):**
- Any

**Additional context**
Please provide any additional context about the problem.

**Logs**
```
08 Sep 09:51:19 | Processed             2233673         |       25.1 ms  | slot             25 ms |:fuelpump: Gas gwei: 0.03 .. 0.03 (0.09) .. 1.12
08 Sep 09:51:19 |  Block    0.0000 ETH     1.13 MGas    |       17   txs | calls         85 ( 10) | sload     169 | sstore     46 | create   0
08 Sep 09:51:19 |  Block throughput       45.19 MGas/s  |      677.1 tps |            39.83 Blk/s | exec code cache     194 | new      0 | ops      56,210
08 Sep 09:51:21 | Setting state sync state root to 2233641 0x0d193e2e409f076292dae272a880f898d2dc02559f546fa41a95b82781ed92c9
08 Sep 09:51:21 | Collecting trie stats and verifying that no nodes are missing staring from block Hash: 0x7e18a611dd6b22330beeee95dac3e41178e30e0e28ff741ea2c981f465e709f1
Number: 2233641
Parent: 0x95f2381ae02c734afb335ee3954ec00448199e8efd4db272fa352b47d509b6f9
Beneficiary: 0x5f62d006c10c009ff50c878cd6157ac861c99990
Gas Limit: 241000000
Gas Used: 206270
Timestamp: 1757324885
Extra Data: 000000000000000000000000000000000000000000000000000000000000004b
Difficulty: 0
Mix Hash: 0x8b7c56ab134ffb9c88bf2c41007a9da0fb004daeb0e6bea4e9aa92adddf2492d
Nonce: 0
Uncles Hash: 0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347
Tx Root: 0x898811e6952907606b904673960b869bec2bca1d16ea4d8a5f1099f91431c512
[...]

PR #9314, opened 2025-09-18
Fix Issues in Taiko Fee Handling

dipkakwani:
Fixes https://github.com/NethermindEth/nethermind/issues/9280

This issue was caught in Taiko Hekla (testnet) when the beneficiary account was deleted and the `feeCollector` did not receive base fee, leading to a state root mismatch error.

## Changes
- Remove the required beneficiary existence check for base fee payment
- Add new tests to ensure fee transfer is done correctly

## Types of changes

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

The test repros the issue in the existing code.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


src/Nethermind/Nethermind.Taiko.Test/TransactionProcessorTests.cs
```diff
@@ -119,4 +117,114 @@ static object[] Typed(int basefeeSharingPctg, ulong goesToTreasury, ulong goesTo
...
+        UInt256 finalTreasuryBalance = _stateProvider.GetBalance(_spec.FeeCollector!);
+
+        UInt256 receivedTipFees = finalCoinbaseBalance - initialCoinbaseBalance;
+        UInt256 receivedBaseFees = finalTreasuryBalance - initialTreasuryBalance;
+
+        long gasUsed = 21000;
+        UInt256 expectedTipFees = isAnchorTx ? 0 : (UInt256)gasUsed * tipFee;
+        UInt256 expectedBaseFees = isAnchorTx ? 0 : (UInt256)gasUsed * baseFee;
+
+        receivedTipFees.Should().Be(expectedTipFees, "Transaction did not receive expected tip fees");
```
dipkakwani, line 168: This particular assumption that expectedTipFees = 0 for the anchor transaction needs to be double checked. 

In taiko-geth we do not have any checks for anchor/non-anchor txn, when paying the tip fees: https://github.com/taikoxyz/taiko-geth/blob/0ea6057b4c2a428dad564ea421cb94f17049b5cc/core/state_transition.go#L555 meaning it is always paid by default, unless I missed some part of code which could modify that.

In NMC we specifically do not pay the tip fees only for anchor transaction, since we skip validation of anchor transactions, which actually populates the tip fees: https://github.com/NethermindEth/nethermind/blob/09a864559e9b2eea06633697221df2b83ef285c7/src/Nethermind/Nethermind.Taiko/TaikoTransactionProcessor.cs#L31 and https://github.com/NethermindEth/nethermind/blob/09a864559e9b2eea06633697221df2b83ef285c7/src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs#L479

smartprogrammer93: @flcl42 any reason for this as far as you know?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Taiko/TaikoTransactionProcessor.cs
```diff
@@ -37,35 +37,42 @@ protected override GasConsumed Refund(Transaction tx, BlockHeader header, IRelea
...
+                UInt256 feeTreasury = baseFees - feeCoinbase;
+                WorldState.AddToBalanceAndCreateIfNotExists(spec.FeeCollector, feeTreasury, spec);
+            }
+            else
+            {
+                WorldState.AddToBalanceAndCreateIfNotExists(spec.FeeCollector, baseFees, spec);
+            }
         }
+
+        if (tracer.IsTracingFees)
```
flcl42, line 74: Trace even in case of no balance change?

src/Nethermind/Nethermind.Taiko.Test/TransactionProcessorTests.cs
flcl42: You can run geth and check if gasUsed even set for anchor tx

src/Nethermind/Nethermind.Taiko/TaikoTransactionProcessor.cs
smartprogrammer93: yeah, this is a fee tracer, not a balance tracer

src/Nethermind/Nethermind.Taiko.Test/TransactionProcessorTests.cs
smartprogrammer93: I think premium for anchor is always set to 0 from the driver, so no need to worry about it here

PR #9317, opened 2025-09-18
Kute code quality cleanup

emlautarom1:
## Changes

- Apply several IDE suggestions to the codebase
- Fix some config descriptions, requirements and typos.

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

All Kute tests pass, the changes are completely internal and do not affect users.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks

This is a small cleanup PR to fix some of the issues pointed out by static analyzers like the ones bundled with Jetbrains Rider. There are no logic changes.

### DIFF
--- a/tools/Kute/Nethermind.Tools.Kute/Config.cs
+++ b/tools/Kute/Nethermind.Tools.Kute/Config.cs
@@ -15,14 +15,6 @@ public static class Config
         Required = true,
     };
 
-    public static Option<string> HostAddress { get; } = new("--address", "-a")
-    {
-        DefaultValueFactory = _ => "http://localhost:8551",
-        Description = "Address where to send JSON RPC requests",
-        HelpName = "url",
-        Required = true,
-    };
-
     public static Option<string> JwtSecretFilePath { get; } = new("--secret", "-s")
     {
         Description = "Path to file with hex-encoded secret for JWT authentication",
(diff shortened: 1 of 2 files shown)


tools/Kute/Nethermind.Tools.Kute/JsonRpc.cs
```diff
@@ -28,14 +27,14 @@ public class Single : Request
...
-            public override string? Id { get => _id.Value; }
+            public string? Id { get => _id.Value; }
             public string? MethodName { get => _methodName.Value; }
 
             public Single(JsonNode node) : base(node)
             {
                 _id = new(() =>
                 {
-                    if (_node["id"] is JsonNode id)
+                    if (_node["id"] is { } id)
```
LukaszRozmej, line 37: is this equivalent?

LukaszRozmej (approved):

tools/Kute/Nethermind.Tools.Kute/JsonRpc.cs
emlautarom1: Yes it is since `_node["id"]` is of type `JsonNode?`, so the type check will only fail when the value is `null`.

This suggestion shows up on Rider but not on VSCode (can't check what happens on Visual Studio though).

More details: https://www.jetbrains.com/help/rider/ConvertTypeCheckToNullCheck.html

PR #9318, opened 2025-09-19
XDC : Snapshot Manager

Demuirgos:
Fixes Closes Resolves #

## Changes

- _List the changes_

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Remarks

### DIFF
--- a/src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
+++ b/src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
@@ -79,9 +79,9 @@ internal class SnapshotManager : ISnapshotManager
 
         var candidates = previousSnapshot.NextEpochCandidates;
 
-        if(header.Number == spec.SwitchBlock + 1)
+        if (header.Number == spec.SwitchBlock + 1)
         {
-            if(candidates.Length > maxMasternodes)
+            if (candidates.Length > maxMasternodes)
             {
                 Array.Resize(ref candidates, maxMasternodes);
                 return (candidates, []);
--- a/src/Nethermind/Nethermind.Xdc/XdcExtensions.cs
+++ b/src/Nethermind/Nethermind.Xdc/XdcExtensions.cs
@@ -60,6 +60,6 @@ public static class XdcExtensions
         return snapshotManager.GetSnapshot(gapBlockHash);
     }
 
-    
+
 
 }


src/Nethermind/Nethermind.Xdc/RLP/SnapshotDecoder.cs
```diff
@@ -0,0 +1,77 @@
...
+using Nethermind.Serialization.Rlp;
+using Nethermind.Xdc.Types;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Nethermind.Xdc.RLP;
+internal class SnapshotDecoder : IRlpStreamDecoder<Snapshot>
```
emlautarom1: It would be nice for `SnapshotDecoder` to also implement `IRlpValueDecoder` so we can avoid allocations.

src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
```diff
@@ -0,0 +1,133 @@
...
+        {
+            snapshot = null;
+            return false;
+        }
+        SnapshotDecoder snapshotDecoder = new();
+
+        var stream = new RlpStream(value.ToArray());
+
+        var decoded = snapshotDecoder.Decode(stream, RlpBehaviors.None);
+        snapshot = decoded;
```
emlautarom1: Here we could avoid `.ToArray()` if `SnapshotDecoder` implemented `IRlpValueDecoder`:

```suggestion
        var decoded = snapshotDecoder.Decode(value, RlpBehaviors.None);
        snapshot = decoded;
```

src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
```diff
@@ -0,0 +1,133 @@
...
+        return true;
+    }
+
+    public bool TryGetSnapshot(XdcBlockHeader header, out Snapshot snapshot)
+    {
+        if (header is null)
+        {
+            snapshot = null;
+            return false;
+        }
```
emlautarom1: Do we need this check when `ISnapshotManager` defines `TryGetSnapshot(XdcBlockHeader, out Snapshot)` as both not null? Having a null at this point seems like a bug. If not, we should change the signature on the interface.

src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
```diff
@@ -0,0 +1,133 @@
...
+        return TryGetSnapshot(header.Hash, out snapshot);
+    }
+
+    public bool TryGetSnapshot(ulong number, bool isGapNumber, out Snapshot snap)
```
emlautarom1: I'd suggest separating this into two methods to avoid the `bool` parameter. It's not clear what `number` stands for on.

src/Nethermind/Nethermind.Xdc.Test/SnapshotManagerTests.cs
```diff
@@ -0,0 +1,187 @@
...
+        Assert.That(result, Is.True);
+        Assert.That(_snapshotManager.TryGetSnapshot(TestItem.KeccakH, out var retrievedSnapshot), Is.True);
+        Assert.That(retrievedSnapshot, Is.EqualTo(snapshot));
+    }
+
+    [Test]
+    public void TryStoreSnapshot_ShouldReturnFalseForNullSnapshot()
+    {
+        // Act
+        var result = _snapshotManager.TryStoreSnapshot(null!);
```
emlautarom1: Same as for `TryCacheSnapshot`

src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
```diff
@@ -0,0 +1,133 @@
...
+        }
+        Span<byte> key = snapshot.Hash.Bytes;
+
+
+        if (_snapshotDb.KeyExists(key))
+        {
+            return false;
+        }
+
+        SnapshotDecoder snapshotDecoder = new();
```
emlautarom1: There are several instantiations of `SnapshotDecoder` which we can avoid by using a field.

src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
```diff
@@ -0,0 +1,133 @@
...
+
+        var contentLength = snapshotDecoder.GetLength(snapshot, RlpBehaviors.None);
+        RlpStream stream = new(contentLength);
+        snapshotDecoder.Encode(stream, snapshot, RlpBehaviors.None);
+
+        stream.Reset();
+        Span<byte> value = stream.Read(stream.Length);
+
+        _snapshotDb.Set(key, value.ToArray());
```
emlautarom1: I think it should be fine to use the `Span` overload:

```suggestion
        _snapshotDb.PutSpan(key, value);
```

src/Nethermind/Nethermind.Xdc/Types/Snapshot.cs
```diff
@@ -19,4 +19,27 @@ public object Clone() =>
...
+            {
+                return false;
+            }
+        }
+        return true;
+
+    }
+    public override string ToString() => $"{Number}:{Hash.ToShortString()}:{String.Join<Address>(',', NextEpochCandidates)}";
+
+    public override int GetHashCode() => base.GetHashCode();
```
emlautarom1: Seems redundant

src/Nethermind/Nethermind.Xdc.Test/SnapshotDecoderTests.cs
```diff
@@ -0,0 +1,40 @@
...
+
+using Nethermind.Core.Test.Builders;
+using Nethermind.Core.Extensions;
+using System.Collections;
+using Nethermind.Xdc.Types;
+using Nethermind.Core.Crypto;
+using System.Collections.Generic;
+using Nethermind.Xdc.RLP;
+
+namespace Nethermind.Xdc.Test
```
emlautarom1: Prefer file scoped namespaces.

src/Nethermind/Nethermind.Xdc.Test/SnapshotManagerTests.cs
```diff
@@ -0,0 +1,187 @@
...
+    private ISnapshotManager _snapshotManager;
+    private readonly IDb _snapshotDb = new MemDb();
+
+    [SetUp]
+    public void Setup()
+    {
+        IBlockTree blockTree = NSubstitute.Substitute.For<IBlockTree>();
+        IXdcConfig xdcConfig = NSubstitute.Substitute.For<IXdcConfig>();
+        // Initialize _snapshotManager with a mock or real implementation
+        _snapshotManager = new SnapshotManager(_snapshotDb, blockTree, xdcConfig); // Replace with actual initialization
```
emlautarom1: Are the comments out of date?

src/Nethermind/Nethermind.Xdc.Test/SnapshotManagerTests.cs
```diff
@@ -0,0 +1,187 @@
...
+        // Assert
+        Assert.That(result, Is.False);
+        Assert.That(snapshot, Is.Null);
+    }
+
+    [Test]
+    public void TryCacheSnapshot_ShouldHandleNullSnapshot()
+    {
+        // Act
+        _snapshotManager.TryCacheSnapshot(null!);
```
emlautarom1: If `TryCacheSnapshot` should support null parameters it should be reflected on the `ISnapshotManager` interface. Otherwise, a `null` at this point is most likely a bug.

src/Nethermind/Nethermind.Xdc/ISnapshotManager.cs
```diff
@@ -13,13 +13,8 @@
...
-    bool TryGetSnapshot(long number, Hash256 hash, out Snapshot snapshot);
+    bool TryGetSnapshot(Hash256 hash, out Snapshot snapshot);
     bool TryStoreSnapshot(Snapshot snapshot);
-    Address GetBlockSealer(BlockHeader header);
-    bool IsValidVote(Snapshot snapshot, Address address, bool authorize);
-    bool IsInTurn(Snapshot snapshot, long number, Address signer);
-    bool HasSignedRecently(Snapshot snapshot, long number, Address signer);
     bool TryGetSnapshot(XdcBlockHeader header, out Snapshot snapshot);
     void TryCacheSnapshot(Snapshot snapshot);
     bool TryGetSnapshot(ulong gapNumber, bool isGapNumber, out Snapshot snap);
```
emlautarom1: Maybe a bit more general, but when do we care if we're caching or storing the snapshot? Should it be an implementation detail of the implementation of `ISnapshotManager`?
Currently there are no use cases beyond tests so I might be missing some context.

emlautarom1: A possible design would be the following:
- Remove the `Cache` methods, only supporting `Get` and `Store`
- Client code could have two instances of `ISnapshotManager`, one used as cache and the other as "persistent" (`ISnapshotManager _cache` and `ISnapshotManager _persistent`), and decide which one to use as appropriate.
- Have two implementations, one uses a `ConcurrentDictionary` and the other uses a `IDb`.

emlautarom1 (changes requested):
Left several comments. Do we have any ticket where we're tracking all XDC related stuff? I'd like to have a bit more context on what we're working on here.

ak88 (changes requested):
Its not using the new `ISnapshotManager`

src/Nethermind/Nethermind.Xdc/Types/Snapshot.cs
Demuirgos: if we override equality we have to override GetHashCode too

src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
```diff
@@ -0,0 +1,133 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Nethermind.Xdc;
+internal class SnapshotManager : ISnapshotManager
+{
+
+    private ConcurrentDictionary<Hash256, Snapshot> _snapshotsByHash = new();
+    private IBlockTree _tree { get; }
+    private IXdcConfig _xdcConfig { get; }
```
cicr99: I have removed the IXdcConfig in #9245, could we use ReleaseSpec instead?

src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
```diff
@@ -0,0 +1,133 @@
...
+        if (_snapshotDb.KeyExists(key))
+        {
+            return false;
+        }
+
+        SnapshotDecoder snapshotDecoder = new();
+
+        var contentLength = snapshotDecoder.GetLength(snapshot, RlpBehaviors.None);
+        RlpStream stream = new(contentLength);
+        snapshotDecoder.Encode(stream, snapshot, RlpBehaviors.None);
```
cicr99: you could have this logic inside an `Encode` function in the SnapshotDecoder that only takes 2 parameters: the snapshot and the behavior, same as in XdcHeaderDecoder or ExtraConsensusDataDecoder

src/Nethermind/Nethermind.Xdc/ISnapshotManager.cs
Demuirgos: so far, we are mainly replicating methods from XDC repo, so it is not yet fledged out, but we are considering refactors down the line,

src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
```diff
@@ -1,87 +1,135 @@
...
-    public bool TryGetSnapshot(ulong gapNumber, bool isGapNumber, out Snapshot snap)
+    public bool StoreSnapshot(Snapshot snapshot)
     {
-        throw new NotImplementedException();
+        if (snapshot is null)
+            return false;
+        Span<byte> key = snapshot.Hash.Bytes;
+
+        if (_snapshotDb.KeyExists(key))
+            return false;
```
ak88: Think you meant if not exists, but why not just overwrite the snapshot if there already is one?

src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
```diff
@@ -1,87 +1,135 @@
...
-    public bool IsValidVote(Snapshot snapshot, Address address, bool authorize)
-    {
-        throw new NotImplementedException();
+        Span<byte> key = hash.Bytes;
+        if (!_snapshotDb.KeyExists(key))
+            return null;
+        Span<byte> value = _snapshotDb.Get(key);
+        if (value.IsEmpty)
+            return null;
+        SnapshotDecoder snapshotDecoder = new();
```
emlautarom1: `snapshotDecoder` could be a `static readonly` field.

src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
```diff
@@ -1,87 +1,135 @@
...
+        _snapshotDb = snapshotDb;
+        _tree = tree;
     }
 
-    public bool IsInTurn(Snapshot snapshot, long number, Address signer)
+    public Snapshot? GetSnapshot(Hash256 hash)
     {
-        throw new NotImplementedException();
-    }
+        Snapshot snapshot = _snapshotsByHash.Get(hash);
```
emlautarom1:
```suggestion
        Snapshot? snapshot = _snapshotsByHash.Get(hash);
```

src/Nethermind/Nethermind.Xdc/ISnapshotManager.cs
```diff
@@ -13,15 +13,11 @@
...
-    bool IsInTurn(Snapshot snapshot, long number, Address signer);
-    bool HasSignedRecently(Snapshot snapshot, long number, Address signer);
-    bool TryGetSnapshot(XdcBlockHeader header, out Snapshot snapshot);
-    void TryCacheSnapshot(Snapshot snapshot);
-    bool TryGetSnapshot(ulong gapNumber, bool isGapNumber, out Snapshot snap);
+    Snapshot? GetSnapshot(Hash256 hash);
+    bool StoreSnapshot(Snapshot snapshot);
+    Snapshot? GetSnapshotByGapNumber(ulong gapNumber);
+    Snapshot? GetSnapshotByHeaderNumber(ulong number, ulong xdcEpoch, ulong xdcGap);
+    Snapshot? GetSnapshotByHeader(XdcBlockHeader? header);
```
emlautarom1: These three methods (in the only implementation) delegate the work to `GetSnapshot`. I'd remove them from the interface and make them extensions instead.

src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
```diff
@@ -1,87 +1,135 @@
...
+
+        _snapshotDb.Set(key, rlpEncodedSnapshot.Bytes);
+        return true;
     }
 
-    public bool TryStoreSnapshot(Snapshot snapshot)
+    public Address[] CalculateNextEpochMasternodes(XdcBlockHeader xdcHeader)
     {
-        throw new NotImplementedException();
+        Snapshot snapshot = GetSnapshotByHeader(xdcHeader);
```
emlautarom1:
```suggestion
        Snapshot? snapshot = GetSnapshotByHeader(xdcHeader);
```

src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
```diff
@@ -1,87 +1,135 @@
...
+        if (snapshot is null)
+            throw new InvalidOperationException($"No snapshot found for header {xdcHeader.Number}:{xdcHeader.Hash.ToShortString()}");
```
emlautarom1: Other methods return `null` when not found. Is it OK for us to throw here? (several cases in this class)

src/Nethermind/Nethermind.Xdc/Types/Snapshot.cs
```diff
@@ -9,14 +9,16 @@
 
 namespace Nethermind.Xdc.Types;
 
-public class Snapshot(long number, Hash256 hash, Address[] nextEpochCandidates) : ICloneable
+public class Snapshot(long number, Hash256 hash, Address[] masterNodes, Address[] penalizedNodes) : ICloneable
 {
     public long Number { get; set; } = number;
     public Hash256 Hash { get; set; } = hash;
-    public Address[] NextEpochCandidates { get; set; } = nextEpochCandidates;
+    public Address[] MasterNodes { get; set; } = masterNodes;
```
ak88: Either these are the actual list of master nodes that counts in that epoch, or they are the candidates which was extracted from the smart contract. I think we have to know the list of epoch candidates when verifying a vote and other places as well.

src/Nethermind/Nethermind.Xdc/Types/Snapshot.cs
```diff
@@ -9,14 +9,16 @@
 
 namespace Nethermind.Xdc.Types;
 
-public class Snapshot(long number, Hash256 hash, Address[] nextEpochCandidates) : ICloneable
+public class Snapshot(long number, Hash256 hash, Address[] masterNodes, Address[] penalizedNodes) : ICloneable
 {
     public long Number { get; set; } = number;
```
ak88:
```suggestion
    public long BlockNumber { get; set; } = number;
```

src/Nethermind/Nethermind.Xdc/Types/Snapshot.cs
```diff
@@ -9,14 +9,16 @@
...
     public Hash256 Hash { get; set; } = hash;
```
ak88:
```suggestion
    public Hash256 HeaderHash { get; set; } = hash;
```

src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
Demuirgos: Wouldn't it be redundant?

Demuirgos: I check if key exists, if it does there is no need to store again?

ak88: Im not sure if its possible to somehow have different candidates, but same header hash. Anyway i dont see any harm in saving it again.

src/Nethermind/Nethermind.Xdc/Types/Snapshot.cs
Demuirgos: They are supposed to be MasterNodes and I also have PenalizedAddresses so nextEpochCandidates can be calculated

src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
```diff
@@ -1,87 +1,135 @@
...
+        foreach (var addr in snapshot.MasterNodes)
+        {
+            if (snapshot.PenalizedNodes.Contains(addr))
+            {
+                continue;
+            }
+
+            masternodes[index++] = addr;
+        }
+        return masternodes;
```
emlautarom1: This is a `.Where()` right?

```suggestion
        Address[] masternodes = snapshot.MasterNodes
            .Where(addr => !snapshot.PenalizedNodes.Contains(addr))
            .ToArray();

          return masternodes;
```

src/Nethermind/Nethermind.Xdc.Test/SnapshotDecoderTests.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Nethermind.Xdc.Test;
+
+[TestFixture]
+public class SnapshotDecoderTests
+{
+    private static IEnumerable<Snapshot> Objects => [
```
emlautarom1: Why `Objects`?

```suggestion
    private static IEnumerable<Snapshot> Snapshots => [
```

src/Nethermind/Nethermind.Xdc.Test/SnapshotDecoderTests.cs
```diff
@@ -0,0 +1,58 @@
...
+        new Snapshot(3, Keccak.EmptyTreeHash, [], [Address.FromNumber(1), Address.FromNumber(2)]),
+    ];
+
+    [Test, TestCaseSource(nameof(Objects))]
+    public void RoundTrip_valuedecoder(Snapshot original)
+    {
+        SnapshotDecoder encoder = new();
+        RlpStream rlpStream = new(10000);
```
emlautarom1: Prefer to use the length of the input `Snapshot`. If we add more cases to the `TestCaseSource` then we're sure they won't break:


```suggestion
        SnapshotDecoder encoder = new();
        RlpStream rlpStream = new(encoder.GetLength(original, RlpBehaviors.None));
```

src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
```diff
@@ -1,87 +1,135 @@
...
     {
-        throw new NotImplementedException();
+        ulong gapBlockNum = Math.Max(0, number - number % xdcEpoch - xdcGap);
+
+        return GetSnapshotByGapNumber(gapBlockNum);
     }
 
-    public bool TryGetSnapshot(XdcBlockHeader header, out Snapshot snapshot)
+
+    public Snapshot? GetSnapshotByGapNumber(ulong gapBlockNum)
```
emlautarom1: If we remove this method from the class we can drop the dependency on `IBlockTree` which is quite large.

src/Nethermind/Nethermind.Xdc/Types/Snapshot.cs
```diff
@@ -9,14 +9,16 @@
 
 namespace Nethermind.Xdc.Types;
 
-public class Snapshot(long number, Hash256 hash, Address[] nextEpochCandidates) : ICloneable
+public class Snapshot(long number, Hash256 hash, Address[] masterNodes, Address[] penalizedNodes) : ICloneable
```
emlautarom1: Suggestion: if we make all arrays `IReadOnlyList` then we should be able to convert this class to a `record`, dropping the need for `IClone`.

Is there any need for these properties to be mutable arrays?

src/Nethermind/Nethermind.Xdc.Test/SnapshotManagerTests.cs
```diff
@@ -0,0 +1,164 @@
...
+
+        // Act
+        var result = _snapshotManager.GetSnapshot(TestItem.KeccakE);
+
+        // assert that it was retrieved from cache
+        result.Should().BeEquivalentTo(snapshot);
+    }
+
+    [Test]
+    public void GetSnapshot_ShouldReturnFalseForEmptyDb()
```
emlautarom1: Same test as `GetSnapshot_ShouldReturnFalseForNonExistentSnapshot`

src/Nethermind/Nethermind.Xdc.Test/SnapshotManagerTests.cs
```diff
@@ -0,0 +1,164 @@
...
+using Nethermind.Xdc.Types;
+using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Nethermind.Xdc.Test;
+internal class SnapshotManagerTests
```
emlautarom1, line 20: Several tests have outdated comments or names (ex. `GetSnapshot_ShouldReturnFalseForNonExistentSnapshot`: we return `null` instead of `false` when not found)

src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
```diff
@@ -1,87 +1,135 @@
...
+
+        return GetSnapshot(gapBlockHash);
     }
 
-    public bool TryGetSnapshot(ulong gapNumber, bool isGapNumber, out Snapshot snap)
+    public bool StoreSnapshot(Snapshot snapshot)
     {
-        throw new NotImplementedException();
+        if (snapshot is null)
+            return false;
```
emlautarom1: This should never happen, or we should make the argument nullable (`Snapshot? snapshot`)

src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
```diff
@@ -1,87 +1,135 @@
...
+        Hash256 gapBlockHash = _tree.FindHeader((long)gapBlockNum)?.Hash;
+
+        if (gapBlockHash is null)
+            return null;
+
+        return GetSnapshot(gapBlockHash);
     }
 
-    public bool TryGetSnapshot(ulong gapNumber, bool isGapNumber, out Snapshot snap)
+    public bool StoreSnapshot(Snapshot snapshot)
```
emlautarom1: Not sure why we need to return a boolean here. Do we need to know that the underlying db has been affected?

src/Nethermind/Nethermind.Xdc.Test/SnapshotManagerTests.cs
emlautarom1: Also, there are no tests for:
- `GetSnapshotByHeaderNumber`: Contains tricky logic to compute the gap block number.
- `GetSnapshotByGapNumber`: Uses the `IBlockTree` to find a header.

emlautarom1 (commented):
Added a couple of additional comments.

I think the implementation is pretty much there but there are some API inconsistencies and some missing tests.

src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
Demuirgos: I avoided Linq, because i am not sure of the perf penalty it will have

src/Nethermind/Nethermind.Xdc.Test/SnapshotManagerTests.cs
Demuirgos: Pretty much all methods in interface, depending on Iblocktree except store and get

src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
Demuirgos: What do you suggest? We return void?

emlautarom1: If not needed I guess it's fine to return void, that's why I was asking if something (maybe in a different PR) requires this knowledge.

Demuirgos: I guess for now it can return void, the interface for now is what is needed by components we are working on

ak88: I dont think this knowledge is actually needed.

src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
Demuirgos: @ak88 what do you think in this case? we return null or throw?

src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
Demuirgos: not really, the GetMasterNodes and the CalculateCandidates also use it

src/Nethermind/Nethermind.Xdc/XdcExtensions.cs
```diff
@@ -28,4 +34,57 @@ public static IXdcReleaseSpec GetXdcSpec(this ISpecProvider specProvider, XdcBlo
...
+    {
+        Hash256 gapBlockHash = tree.FindHeader((long)gapBlockNum)?.Hash;
+
+        if (gapBlockHash is null)
+            return null;
+
+        return snapshotManager.GetSnapshot(gapBlockHash);
+    }
+
+    public static T[] RemoveItemFromArray<T>(T[] candidates, T[] penalties, int withMaxCap = int.MaxValue)
```
ak88: Is this used anywhere else? This should probably just be a private method in the class instead.

src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
```diff
@@ -1,87 +1,104 @@
...
-        return GetMasternodes(xdcHeader);
+        var candidates = previousSnapshot.NextEpochCandidates;
+
+        if(header.Number == spec.SwitchBlock + 1)
+        {
+            if(candidates.Length > maxMasternodes)
+            {
+                Array.Resize(ref candidates, maxMasternodes);
+                return (candidates, []) ;
+            }
```
ak88: you can create a local function instead of repeating the code

src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
emlautarom1: I'd go with LINQ here just for clarity.

src/Nethermind/Nethermind.Xdc/RLP/SnapshotDecoder.cs
```diff
@@ -0,0 +1,122 @@
...
+        var contentLength = GetLength(item, rlpBehaviors);
+
+        stream.StartSequence(contentLength);
+        stream.Encode(item.BlockNumber);
+        stream.Encode(item.HeaderHash);
+
+        if (item.NextEpochCandidates is null)
+            stream.EncodeArray<Address>([]);
+        else
+            EncodeAddressSequence(stream, item.NextEpochCandidates);
```
ak88, line 94: Why not use EncodeArray in both cases?

ak88 (approved):

PR #9322, opened 2025-09-19
Fix: ArrayPool buffer leak in PatriciaTree Get methods

gitToki:
## Changes

- Fixed `ArrayPool` buffer leak in `Get()` method - Moved `ArrayPool<byte>.Shared.Return(array)` from try block to finally block to ensure cleanup even when `TrieException` is thrown
- Fixed ArrayPool buffer leak in `GetNodeByKey()` method - Moved `ArrayPool<byte>.Shared.Return(array)` from try block to finally block to ensure cleanup even when `TrieException` is thrown

To give a little bit more detail about the why, the `Get()` and `GetNodeByKey()` methods in PatriciaTree rent buffers from `ArrayPool` but only returned them in the success path.

When `GetNew()` throws a `TrieException`, the rented buffers weren't returned to the pool, it could've cause gradual memory pressure under sustained load (even if the array get GC'ed at some point).

This was only affecting RPC endpoints when handling requests with keys larger than the 64byte stack allocation min.
For the p2p path, the peer would be banned before causing any significant harm.

### DIFF
--- a/src/Nethermind/Nethermind.Trie/PatriciaTree.cs
+++ b/src/Nethermind/Nethermind.Trie/PatriciaTree.cs
@@ -478,10 +478,10 @@ namespace Nethermind.Trie
 
             _writeBeforeCommit++;
 
+            byte[]? array = null;
             try
             {
                 int nibblesCount = 2 * rawKey.Length;
-                byte[] array = null;
                 Span<byte> nibbles = (rawKey.Length <= MaxKeyStackAlloc
                         ? stackalloc byte[MaxKeyStackAlloc] // Fixed size stack allocation
                         : array = ArrayPool<byte>.Shared.Rent(nibblesCount))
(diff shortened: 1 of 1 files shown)


emlautarom1:
The code does not compile as it is:

```
  Nethermind.Trie failed with 2 error(s) (0.2s)
    /home/emlautarom1/Development/Nethermind/nethermind/src/Nethermind/Nethermind.Trie/PatriciaTree.cs(362,21): error CS0103: The name 'array' does not exist in the current context
    /home/emlautarom1/Development/Nethermind/nethermind/src/Nethermind/Nethermind.Trie/PatriciaTree.cs(362,70): error CS0103: The name 'array' does not exist in the current context
```

emlautarom1:
There is a similar case:

https://github.com/NethermindEth/nethermind/blob/ae427a1e7ff62af36dde7775f4c3b47d992b233c/src/Nethermind/Nethermind.Trie/PatriciaTree.cs#L487

If you fix it then we can proceed.

gitToki:
Thanks for the review

PR #9324, opened 2025-09-20
StatelessExecution tool

deffrian:
Fixes Closes Resolves #
This is the thing we want to execution in zisk. There is also part that does witness generation, but it's removed for now.

## Changes

- Add stateless execution

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


src/Nethermind/Nethermind.Consensus/Stateless/StatelessBlockProcessingEnv.cs
```diff
@@ -0,0 +1,106 @@
...
+
+namespace Nethermind.Consensus.Stateless;
+
+public class StatelessBlockProcessingEnv(
+    Witness witness,
+    ISpecProvider specProvider,
+    ISealValidator sealValidator,
+    ILogManager logManager)
+{
+    private readonly ILogger _logger = logManager.GetClassLogger();
```
emlautarom1: The logger is not used anywhere and can be removed.

src/Nethermind/Nethermind.Consensus/Stateless/StatelessBlockProcessingEnv.cs
```diff
@@ -0,0 +1,106 @@
...
+    {
+        IKeyValueStore db = new MemDb();
+        foreach (var stateElement in witness.State)
+        {
+            var hash = Keccak.Compute(stateElement);
+            db.PutSpan(hash.Bytes, stateElement);
+        }
+
+        NodeStorage nodeStorage = new(db, INodeStorage.KeyScheme.Hash);
+        return new TrieStore(nodeStorage, NoPruning.Instance, NoPersistence.Instance, new PruningConfig(), NullLogManager.Instance);
```
emlautarom1: Are we intentionally passing a `NullLogManager` here? If so, why is that the case?

src/Nethermind/Nethermind.Consensus/Stateless/StatelessBlockProcessingEnv.cs
```diff
@@ -0,0 +1,106 @@
...
+            new ExecutionRequestsProcessor(txProcessor)
+        );
+    }
+
+    private ITrieStore CreateTrie()
+    {
+        IKeyValueStore db = new MemDb();
+        foreach (var stateElement in witness.State)
+        {
+            var hash = Keccak.Compute(stateElement);
```
emlautarom1: Can use `ValueKeccak` directly:

```suggestion
            var hash = ValueKeccak.Compute(stateElement);
```

src/Nethermind/Nethermind.Consensus/Stateless/StatelessBlockProcessingEnv.cs
```diff
@@ -0,0 +1,106 @@
...
+using Nethermind.Evm.State;
+using Nethermind.Evm.TransactionProcessing;
+using Nethermind.Logging;
+using Nethermind.State;
+using Nethermind.Trie;
+using Nethermind.Trie.Pruning;
+
+namespace Nethermind.Consensus.Stateless;
+
+public class StatelessBlockProcessingEnv(
```
emlautarom1, line 28: This class is essentially a DI container. Would it make sense to go that route rather than manually constructing the required objects?
This could be beneficial in case we want to override certain components (ex. if we want to support OP Stack we want to use `OptimismHeaderValidator` rather than `HeaderValidator`)

src/Nethermind/Nethermind.Consensus/Stateless/StatelessBlockTree.cs
```diff
@@ -0,0 +1,212 @@
...
+using System.Threading.Tasks;
+using Nethermind.Blockchain;
+using Nethermind.Blockchain.Visitors;
+using Nethermind.Core;
+using Nethermind.Core.Collections;
+using Nethermind.Core.Crypto;
+
+namespace Nethermind.Consensus.Stateless;
+
+public class StatelessBlockTree(BlockHeader[] headers) : IBlockTree
```
emlautarom1: Most methods of `IBlockTree` are not supported by this implementation so we should document it in a docs block (reasoning, use case).
In an ideal world we could split the `IBlockTree` interface into smaller interfaces that can be fully implemented by this class (probably not in this PR)

src/Nethermind/Nethermind.Consensus/Stateless/Witness.cs
```diff
@@ -0,0 +1,40 @@
...
+    public BlockHeader[] DecodedHeaders => _decodedHeaders ??= DecodeHeaders();
+
+    private BlockHeader[] DecodeHeaders()
+    {
+        List<BlockHeader> headers = new(Headers.Length);
+        HeaderDecoder decoder = new();
+        foreach (var encodedHeader in Headers)
+        {
+            RlpStream stream = new(encodedHeader);
+            headers.Add(decoder.Decode(stream) ?? throw new ArgumentException());
```
emlautarom1: Prefer to use the value decoders when possible:


```suggestion
            Rlp.ValueDecoderContext stream = new(encodedHeader);
            headers.Add(decoder.Decode(ref stream) ?? throw new ArgumentException());
```

src/Nethermind/Nethermind.Consensus/Stateless/Witness.cs
```diff
@@ -0,0 +1,40 @@
...
+        }
+        return headers.ToArray();
+    }
```
emlautarom1: If we change the type of `DecodedHeaders` to `List<BlockHeader>` we can avoid calls to `.ToArray()`:


```suggestion
    private List<BlockHeader>? _decodedHeaders = null;
    public List<BlockHeader> DecodedHeaders => _decodedHeaders ??= DecodeHeaders();

    private List<BlockHeader> DecodeHeaders()
    {
        List<BlockHeader> headers = new(Headers.Length);
        HeaderDecoder decoder = new();
        foreach (var encodedHeader in Headers)
        {
            RlpStream stream = new(encodedHeader);
            headers.Add(decoder.Decode(stream) ?? throw new ArgumentException());
        }
        return headers;
    }
```

tools/StatelessExecution/SetupCli.cs
```diff
@@ -0,0 +1,182 @@
...
+            Required = true
+        };
+
+        command.Add(witnessFile);
+        command.Add(blockFile);
+
+        command.SetAction((parseResult, _) =>
+        {
+            string blockFileName = parseResult.GetValue(blockFile)!;
+            string? execWitnessFileName = parseResult.GetValue(witnessFile)!;
```
emlautarom1:
```suggestion
            string execWitnessFileName = parseResult.GetValue(witnessFile)!;
```

tools/StatelessExecution/SetupCli.cs
```diff
@@ -0,0 +1,182 @@
...
+
+            BlockHeader? baseBlock = null;
+            foreach (BlockHeader header in witness.DecodedHeaders)
+            {
+                if (header.Hash == suggestedBlockHeader.ParentHash)
+                {
+                    baseBlock = header;
+                    break;
+                }
+            }
```
emlautarom1:
```suggestion
            BlockHeader? baseBlock = witness
                .DecodedHeaders
                .FirstOrDefault(h => h.Hash == suggestedBlockHeader.ParentHash);
```

tools/StatelessExecution/SetupCli.cs
```diff
@@ -0,0 +1,182 @@
...
+            Console.WriteLine($"✅  Found parent block #{baseBlock.Number}\n");
+
+            Console.WriteLine("⚙️  Initializing block processing environment...");
+            Console.WriteLine("🔄  Processing block...");
+            Console.WriteLine($"📋  Processing {suggestedBlockForRpc.Transactions.Length} transactions...");
+            var transactions = new Transaction[suggestedBlockForRpc.Transactions.Length];
+            for (int j = 0; j < transactions.Length; j++)
+            {
+                var tx = (JsonElement)suggestedBlockForRpc.Transactions[j];
+                transactions[j] = serializer.Deserialize<TransactionForRpc>(tx.GetRawText()).ToTransaction();
```
emlautarom1: If we add an additional overload to `IJsonSerializer` to support `JsonElement` we can avoid the `GetRawText` call:

```
// In `EthereumJsonSerializer.cs`:

public T Deserialize<T>(JsonElement element)
{
    return element.Deserialize<T>(_jsonOptions);
}
```

tools/StatelessExecution/SetupCli.cs
```diff
@@ -0,0 +1,182 @@
...
+            }
+
+            Block suggestedBlock = new Block(suggestedBlockHeader, transactions, [], suggestedBlockForRpc.Withdrawals);
+
+            ISpecProvider specProvider = HoodiSpecProvider.Instance;
```
emlautarom1: Any reason why we pick Hoodi?

tools/StatelessExecution/SetupCli.cs
```diff
@@ -0,0 +1,182 @@
...
+using Nethermind.Specs;
+using Nethermind.Consensus.Stateless;
+using Nethermind.Core.Specs;
+using Nethermind.Trie;
+
+namespace StatelessExecution;
+
+internal static class SetupCli
+{
+    public static void SetupExecute(RootCommand command)
```
emlautarom1, line 23: Nitpick: the CLI setup and main processing logic (here the main command action) should be separated at least in different methods.

src/Nethermind/Nethermind.Consensus/Stateless/StatelessBlockProcessingEnv.cs
deffrian: For this one we don't want to use DI because of riscv5 compilcations

tools/StatelessExecution/SetupCli.cs
deffrian: Not really. It's post-merge and easy to sync/generate test vectors. It's fine to use hoodi only for simplicity imo

tools/StatelessExecution/SetupCli.cs
deffrian: I'm not sure that's a good idea. Can this potentially break logic in some cases? It doesn't seem like this should work. Doesn't JsonSerializer tries to decode JsonElements when it can already?

tools/StatelessExecution/SetupCli.cs
emlautarom1: Understood. We might want to add a comment saying that it does not really matter.

tools/StatelessExecution/SetupCli.cs
tanishqjasoria: we should definitely separate the logic, we would need a different setup for zkvm, different i/o interface.

tools/StatelessExecution/SetupCli.cs
```diff
@@ -0,0 +1,174 @@
...
+            string blockFileName = parseResult.GetValue(blockFile)!;
+            string execWitnessFileName = parseResult.GetValue(witnessFile)!;
+
+            Console.WriteLine("🚀  Starting stateless execution...\n");
+            Console.WriteLine($"📁  Block file: {blockFileName}");
+            Console.WriteLine($"🔍  Witness file: {execWitnessFileName}\n");
+
+            var serializer = new EthereumJsonSerializer();
+
+            Console.WriteLine("📖  Reading input files...");
```
tanishqjasoria: can we add a logger for the logs. also lets remove the emojis
i just added those for the feature branch that we shared in the blog post

tools/StatelessExecution/SetupCli.cs
```diff
@@ -0,0 +1,174 @@
...
+            var transactions = new Transaction[suggestedBlockForRpc.Transactions.Length];
+            for (int j = 0; j < transactions.Length; j++)
+            {
+                var tx = (JsonElement)suggestedBlockForRpc.Transactions[j];
+                transactions[j] = serializer.Deserialize<TransactionForRpc>(tx.GetRawText()).ToTransaction();
+            }
+
+            Block suggestedBlock = new Block(suggestedBlockHeader, transactions, [], suggestedBlockForRpc.Withdrawals);
+
+            ISpecProvider specProvider = HoodiSpecProvider.Instance;
```
tanishqjasoria: can we add it as a config? we are testing with hoodi but this works with any network

tools/StatelessExecution/SetupCli.cs
deffrian: Let's not remove the emojis. I like them:)

src/Nethermind/Nethermind.Logging/SimpleConsoleLogManager.cs
```diff
@@ -6,27 +6,23 @@
...
         {
-            return new(SimpleConsoleLogger.Instance);
+            return new(new SimpleConsoleLogger(logLevel, dateFormat));
         }
 
         public ILogger GetLogger(string loggerName)
         {
-            return new(SimpleConsoleLogger.Instance);
+            return new(new SimpleConsoleLogger(logLevel, dateFormat));
         }
```
tanishqjasoria, line 26: why is this change needed here?

src/Nethermind/Nethermind.Logging/SimpleConsoleLogger.cs
```diff
@@ -8,48 +8,44 @@ namespace Nethermind.Logging
...
-        public bool IsWarn => true;
-        public bool IsDebug => true;
-        public bool IsTrace => true;
-        public bool IsError => true;
+        public bool IsInfo => logLevel >= LogLevel.Info;
+        public bool IsWarn => logLevel >= LogLevel.Warn;
+        public bool IsDebug => logLevel >= LogLevel.Debug;
+        public bool IsTrace => logLevel >= LogLevel.Trace;
+        public bool IsError => logLevel >= LogLevel.Error;
     }
```
tanishqjasoria, line 50: same, why are we changing the logger in this PR?

src/Nethermind/Nethermind.Consensus/Stateless/StatelessBlockProcessingEnv.cs
```diff
@@ -0,0 +1,104 @@
...
+    private IKeyValueStoreWithBatching CreateCodeDb()
+    {
+        IKeyValueStoreWithBatching db = new MemDb();
+        foreach (var code in witness.Codes)
+        {
+            var hash = ValueKeccak.Compute(code).Bytes;
+            db.PutSpan(hash, code);
+        }
+        return db;
+    }
```
tanishqjasoria: maybe we can move these function to witness class?
so we can simply do witness.CreateTrie() and witness.CreateCodeDb()

src/Nethermind/Nethermind.Logging/SimpleConsoleLogManager.cs
deffrian: I thought that using regular logger is an overkill. I decided to use SimpleConsoleLogger, but it doesn't support log level. So this is why I changed it

src/Nethermind/Nethermind.Consensus/Stateless/StatelessBlockProcessingEnv.cs
deffrian: I don't think Witness class should be responsible for that. And I don't see any benefit in doing that

tanishqjasoria: it would be beneficil in the case where we change the witness format, we wont have to change this class.
Also this way we can support multiple witness formats.

tanishqjasoria: but again, no strong opinions

tanishqjasoria (approved):

tools/StatelessExecution/SetupCli.cs
emlautarom1: cc. @benaadams

tools/StatelessExecution/SetupCli.cs
```diff
@@ -0,0 +1,177 @@
...
+        try
+        {
+            (Block processed, TxReceipt[] _) = blockProcessor.ProcessOne(suggestedBlock,
+                ProcessingOptions.ReadOnlyChain, NullBlockTracer.Instance, specProvider.GetSpec(suggestedBlock.Header));
+
+            if (processed.Hash != suggestedBlock.Hash)
+            {
+                logger.Info("❌  Block processing failed - Hash mismatch");
+                logger.Info($"    Expected: {suggestedBlock.Hash}");
+                logger.Info($"    Actual:   {processed.Hash}");
```
LukaszRozmej, line 156: Are those logger calls being executed under zk-evm?
Maybe worth having them removable by JIT?

tools/StatelessExecution/SetupCli.cs
```diff
@@ -0,0 +1,177 @@
...
+        var transactions = new Transaction[suggestedBlockForRpc.Transactions.Length];
+        for (int j = 0; j < transactions.Length; j++)
+        {
+            var tx = (JsonElement)suggestedBlockForRpc.Transactions[j];
+            transactions[j] = serializer.Deserialize<TransactionForRpc>(tx.GetRawText()).ToTransaction();
+        }
+
+        Block suggestedBlock = new Block(suggestedBlockHeader, transactions, [], suggestedBlockForRpc.Withdrawals);
+        return (witness, suggestedBlock, baseBlock);
+    }
```
LukaszRozmej, line 135: All this decoding looks potentially heavy for zkVM context?
I wonder is it something we can improve in the future.

Do zkVM's use json data sources?

LukaszRozmej (approved):

tools/StatelessExecution/SetupCli.cs
deffrian: Probably not. I don't know if zkvms support logging, maybe some do. We can adjust it later, maybe we'll need some type of special logger for that.

emlautarom1 (approved):

PR #9335, opened 2025-09-22
Fix Wait.AnyWhere signature (use params Task<T>[])

viktorking7:
### Description
- Simplifies API and prevents params-flattening issues by accepting tasks directly.
- Replaces `params IEnumerable<Task<T>>` with `params Task<T>[]`.
- Adds a short comment explaining the change.
- No behavior change beyond correct parameter typing.

### Rationale
Using `params IEnumerable<Task<T>>` results in `IEnumerable<Task<T>>[]` at call sites, which is error-prone and can break initialization of task sets.

### DIFF
--- a/src/Nethermind/Nethermind.Core/Tasks/WaitForPassingTasks.cs
+++ b/src/Nethermind/Nethermind.Core/Tasks/WaitForPassingTasks.cs
@@ -46,4 +46,12 @@ public static class Wait
 
         throw new UnreachableException();
     }
-}
+    
+    /// <summary>
+    /// Overload for callers that provide a collection (not an array) of tasks.
+    /// </summary>
+    public static Task<T> AnyWhere<T>(Func<T, bool> cond, IEnumerable<Task<T>> tasks)
+    {
+        // Avoid forcing callers to allocate arrays; convert here if needed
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Core/Tasks/WaitForPassingTasks.cs
```diff
@@ -18,8 +18,9 @@ public static class Wait
     /// <param name="tasks"></param>
     /// <typeparam name="T"></typeparam>
     /// <returns></returns>
-    public static async Task<T> AnyWhere<T>(Func<T, bool> cond, params IEnumerable<Task<T>> tasks)
+    public static async Task<T> AnyWhere<T>(Func<T, bool> cond, params Task<T>[] tasks)
```
benaadams, line 21: We pass a `ArrayPoolList<Task<byte[]?>>` in to this method form https://github.com/NethermindEth/nethermind/blob/c37acedbb5bf8b7449376eb49dfd3ff7ca2d1bb1/src/Nethermind/Nethermind.Synchronization/Trie/NodeDataRecovery.cs#L121-L123

Which isn't an array

PR #9336, opened 2025-09-22
Fix MCL memory leak

rubo:
## Changes

Updated the MCL library to the version that addresses the memory leak on Windows. The stable version will be released once the fix lands in the stable version of MCL.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

Tested manually

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -55,7 +55,7 @@
     <PackageVersion Include="Nethermind.GmpBindings" Version="1.0.3" />
     <PackageVersion Include="Nethermind.Libp2p" Version="1.0.0-preview.45" />
     <PackageVersion Include="Nethermind.Libp2p.Protocols.PubsubPeerDiscovery" Version="1.0.0-preview.45" />
-    <PackageVersion Include="Nethermind.MclBindings" Version="1.0.2" />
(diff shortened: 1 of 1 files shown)


Directory.Packages.props
```diff
@@ -55,7 +55,7 @@
     <PackageVersion Include="Nethermind.GmpBindings" Version="1.0.3" />
     <PackageVersion Include="Nethermind.Libp2p" Version="1.0.0-preview.45" />
     <PackageVersion Include="Nethermind.Libp2p.Protocols.PubsubPeerDiscovery" Version="1.0.0-preview.45" />
-    <PackageVersion Include="Nethermind.MclBindings" Version="1.0.2" />
+    <PackageVersion Include="Nethermind.MclBindings" Version="1.0.3-preview.1" />
```
LukaszRozmej, line 58: Why not final?

rubo: > The stable version will be released once the fix lands in the stable version of MCL.

PR #9338, opened 2025-09-23
Fix hanging sepolia snap sync

asdacap:
- Fix process hang during sepolia snap sync causing random peer drops. 

## Testing

- Reproducable locally.


asdacap:
I dont know. I do know that `UnbalancedParallelFor` will hang reliably when called concurrently, so thats one clue.

PR #9341, opened 2025-09-23
Make eth_getBlobsV1 return UnsupportedFork post Osaka

LukaszRozmej:
## Changes

- Explicitly in spec: https://github.com/ethereum/execution-apis/blob/main/src/engine/osaka.md#cancun-api

#### Requires testing

- [x] Yes

- [x] Yes

### DIFF
--- a/src/Nethermind/Nethermind.Merge.Plugin/MergeErrorMessages.cs
+++ b/src/Nethermind/Nethermind.Merge.Plugin/MergeErrorMessages.cs
@@ -1,4 +1,4 @@
-﻿// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
 // SPDX-License-Identifier: LGPL-3.0-only
 
 namespace Nethermind.Merge.Plugin;


src/Nethermind/Nethermind.Merge.Plugin/Handlers/GetBlobsHandler.cs
```diff
@@ -3,18 +3,24 @@
...
 namespace Nethermind.Merge.Plugin.Handlers;
 
-public class GetBlobsHandler(ITxPool txPool) : IAsyncHandler<byte[][], IEnumerable<BlobAndProofV1?>>
+public class GetBlobsHandler(ITxPool txPool, IChainHeadSpecProvider chainHeadSpecProvider) : IAsyncHandler<byte[][], IEnumerable<BlobAndProofV1?>>
 {
     private const int MaxRequest = 128;
 
     public Task<ResultWrapper<IEnumerable<BlobAndProofV1?>>> HandleAsync(byte[][] request)
     {
+        if (chainHeadSpecProvider.GetCurrentHeadSpec().IsEip7594Enabled)
```
LukaszRozmej, line 19: Is this best way to check for Osaka here?

src/Nethermind/Nethermind.Merge.Plugin/Handlers/GetBlobsHandlerV2.cs
```diff
@@ -55,7 +55,7 @@ public class GetBlobsHandlerV2(ITxPool txPool) : IAsyncHandler<byte[][], IEnumer
             }
 
             Metrics.GetBlobsRequestsSuccessTotal++;
-            return ResultWrapper<IEnumerable<BlobAndProofV2>?>.Success(response.ToList());
+            return ResultWrapper<IEnumerable<BlobAndProofV2>?>.Success(response);
```
LukaszRozmej, line 58: You can put disposables into response and they will be disposed.

Btw this wasn't disposed on success earlier at all!

PR #9343, opened 2025-09-24
Fix selfdestruct missed between transaction

asdacap:
- Fix selfdestruct missed between transaction.
- Also fix the initial value edge case but on multithreaded path.

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- [X] Fix sepolia forward sync
- [X] Reproducable with unit test
- [ ] Mainnet era completed.

### DIFF
--- a/src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
+++ b/src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
@@ -222,7 +222,11 @@ internal sealed class PersistentStorageProvider : PartialStorageProviderBase
             else
             {
                 _toUpdateRoots.Remove(address);
-                _storages.Remove(address);
+                if (_storages.TryGetValue(address, out var storage))
+                {
+                    // BlockChange need to be kept to keep selfdestruct marker (via DefaultableDictionary) working.
+                    storage.RemoveStorageTree();
+                }
             }
         }
         toUpdateRoots.Clear();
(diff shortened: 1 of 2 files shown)


LukaszRozmej (approved):

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
```diff
@@ -658,7 +662,7 @@ private byte[] LoadFromTreeStorage(StorageCell storageCell)
                 foreach (var kvp in BlockChange)
                 {
                     byte[] after = kvp.Value.After;
-                    if (!Bytes.AreEqual(kvp.Value.Before, after))
+                    if (!Bytes.AreEqual(kvp.Value.Before, after) || kvp.Value.IsInitialValue)
```
flcl42, line 665: Maybe IsInitialValue is cheaper to check, so can be the first check

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
```diff
@@ -222,7 +222,11 @@ protected override void CommitCore(IStorageTracer tracer)
             else
             {
                 _toUpdateRoots.Remove(address);
-                _storages.Remove(address);
+                if (_storages.TryGetValue(address, out var storage))
```
flcl42: `var` to explicit type

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
asdacap: Its fine, most of the time `AreEqual` should be false.

benaadams (approved):

PR #9345, opened 2025-09-24
Fix: add nonce EIP2681 check

jsvisa:
Fixes Closes Resolves #

Fixes https://github.com/NethermindEth/nethermind/issues/9344

## Changes

- _List the changes_

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [x] No

- [ ] Yes
- [ ] No

## Remarks

### DIFF
--- a/src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.cs
+++ b/src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.cs
@@ -266,6 +266,39 @@ namespace Nethermind.TxPool.Test
         }
 
         [Test]
+        public void should_ignore_transactions_with_nonce_exceeding_u64_max()
+        {
+            _txPool = CreatePool();
+            UInt256 maxNonce = UInt256.UInt64MaxValue;
+            UInt256 invalidNonce = maxNonce + 1;
+
+            Transaction tx = Build.A.Transaction
+                .WithNonce(invalidNonce)
+                .SignedAndResolved(_ethereumEcdsa, TestItem.PrivateKeyA).TestObject;
+            EnsureSenderBalance(tx);
+
+            AcceptTxResult result = _txPool.SubmitTx(tx, TxHandlingOptions.PersistentBroadcast);
+            _txPool.GetPendingTransactionsCount().Should().Be(0);
+            result.Should().Be(AcceptTxResult.NonceTooHigh);
+        }
+
+        [Test]
+        public void should_accept_transactions_with_nonce_equal_to_u64_max()
+        {
+            _txPool = CreatePool();
+            UInt256 maxNonce = UInt256.UInt64MaxValue;
+
+            Transaction tx = Build.A.Transaction
+                .WithNonce(maxNonce)
+                .SignedAndResolved(_ethereumEcdsa, TestItem.PrivateKeyA).TestObject;
+            EnsureSenderBalance(tx);
+
+            AcceptTxResult result = _txPool.SubmitTx(tx, TxHandlingOptions.PersistentBroadcast);
+            _txPool.GetPendingTransactionsCount().Should().Be(1);
+            result.Should().Be(AcceptTxResult.Accepted);
+        }
+
+        [Test]
         public void get_next_pending_nonce()
         {
             _txPool = CreatePool();
--- a/src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.cs
+++ b/src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.cs
@@ -269,7 +269,7 @@ namespace Nethermind.TxPool.Test
         public void should_ignore_transactions_with_nonce_exceeding_u64_max()
         {
             _txPool = CreatePool();
-            UInt256 maxNonce = UInt256.UInt64MaxValue;
+            UInt256 maxNonce = Transaction.MaxNonce;
             UInt256 invalidNonce = maxNonce + 1;
 
             Transaction tx = Build.A.Transaction
@@ -286,7 +286,7 @@ namespace Nethermind.TxPool.Test
         public void should_accept_transactions_with_nonce_equal_to_u64_max()
         {
             _txPool = CreatePool();
-            UInt256 maxNonce = UInt256.UInt64MaxValue;
+            UInt256 maxNonce = Transaction.MaxNonce;
 
             Transaction tx = Build.A.Transaction
                 .WithNonce(maxNonce)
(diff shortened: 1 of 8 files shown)


src/Nethermind/Nethermind.TxPool/Filters/MaxNonceFilter.cs
```diff
@@ -0,0 +1,38 @@
...
+
+namespace Nethermind.TxPool.Filters
+{
+    /// <summary>
+    /// Filters out transactions where nonce exceeds 2^64-1 (EIP-2681).
+    /// </summary>
+    internal sealed class MaxNonceFilter : IIncomingTxFilter
+    {
+        private readonly ILogger _logger;
+        private static readonly UInt256 MaxNonce = UInt256.UInt64MaxValue;
```
LukaszRozmej: I don't think this value exists? can just use `ulong.MaxValue`, we should make a const under `Transaction.MaxNonce`

Looks AI generated. But to be honest one of better AI generated PR's I've seen.

src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -161,6 +161,7 @@ public TxPool(IEthereumEcdsa ecdsa,
                 new BalanceZeroFilter(thereIsPriorityContract, _logger),
                 new BalanceTooLowFilter(_transactions, _blobTransactions, _logger),
                 new LowNonceFilter(_logger), // has to be after UnknownSenderFilter as it uses sender
+                new MaxNonceFilter(_logger), // check nonce doesn't exceed 2^64-1 (EIP-2681)
```
LukaszRozmej: This is a static check so should go to `_preHashFilters` as it can be done before hash is calculated and without fetching data from db.

BUT...

This check shouldn't be `TxPool` only, it should be a consensus check, thus it should be implemented under `TxValidator`, which will be used in `MalformedTxFilter`, removing the need for explicit filter in `TxPool`

src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.cs
```diff
@@ -265,6 +265,39 @@ public void should_ignore_old_nonce_transactions()
...
+
+            Transaction tx = Build.A.Transaction
+                .WithNonce(maxNonce)
+                .SignedAndResolved(_ethereumEcdsa, TestItem.PrivateKeyA).TestObject;
+            EnsureSenderBalance(tx);
+
+            AcceptTxResult result = _txPool.SubmitTx(tx, TxHandlingOptions.PersistentBroadcast);
+            _txPool.GetPendingTransactionsCount().Should().Be(1);
+            result.Should().Be(AcceptTxResult.Accepted);
+        }
```
LukaszRozmej: This could be one test with 2 TestCases where only nonce is provided as parameter and returns `AcceptTxResult` that is compared with expected return value.

src/Nethermind/Nethermind.TxPool/Filters/MaxNonceFilter.cs
jsvisa: yeah, not familiar with c#, generated by vibe coding

src/Nethermind/Nethermind.TxPool/TxPool.cs
jsvisa: So do you means the consensus layer also not check the nonce >= u64 max? If so, I think this is different to other clients

jsvisa: So do you means the consensus layer also not check the nonce >= u64 max? If so, I think this is different to other clients

LukaszRozmej: yeah looks like it. In practice no chain is even close to this value and consensus checks continuity so nothing urgent, but should be fixed.

src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.cs
```diff
@@ -265,6 +265,39 @@ public void should_ignore_old_nonce_transactions()
...
+            AcceptTxResult result = _txPool.SubmitTx(tx, TxHandlingOptions.PersistentBroadcast);
+            _txPool.GetPendingTransactionsCount().Should().Be(0);
+            result.Should().Be(AcceptTxResult.NonceTooHigh);
+        }
+
+        [Test]
+        public void should_accept_transactions_with_nonce_equal_to_u64_max()
+        {
+            _txPool = CreatePool();
+            UInt256 maxNonce = UInt256.UInt64MaxValue;
```
marcindsobczak: According to EIP-2681, nonce equal 2^64-1 should be invalid.

[Rationale, point 4](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2681.md):
`The reason a transaction with nonce 2^64-1 is invalid, because otherwise after inclusion the sender account's nonce would exceed 2^64-1.`

1033309821:
Thanks for solving the problem. Can you apply for bug bounty and CVE?

src/Nethermind/Nethermind.TxPool/TxPool.cs
marcindsobczak: This limit is only theoretical, but we should implement it anyway. Some math:

- achieving max nonce limit will cost at least: `21000 * (2^64-1) = 387,381,625,547,900,583,915,000` gas
- minimal base fee is equal 7wei. It means that single account would need to spend at least: `21000 * (2^64-1) * 7wei = 2,711,671,378,835,304,087,405,000` wei = 2,7M eth which is currently worth about 11 Billion dollars
- 7 wei base fee is the absolute minimum, realistically, even with base fee equal 0,0001Gwei, price will move to 38 Billion eth which is 320x more than all current supply

jsvisa:
> Actually new implementation of `ITxValidator` should be added to all TxTypes validators, not only legacy one: https://github.com/NethermindEth/nethermind/blob/master/src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs#L22-L77

Hi @marcindsobczak  How about let's add the nonce validtor in `IsWellFormed`, so no need to add for each tx type

marcindsobczak:
`IsWellFormed` is a part of every implementation of `ITxValidator`. So we need both - creating new implementation of `ITxValidator` AND adding there a method `IsWellFormed`. We have separate implementation of `ITxValidator` for each part of transaction which is validated to avoid a mess in the code.

LukaszRozmej:
Closing in favor of #9352

PR #9346, opened 2025-09-24
Better Isolate trie store during pruning

asdacap:
- Further isolate trie store during pruning to prevent random error.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

#### Notes on testing

- Sees to fix issue with trie exception when reading concurrently with read only triestore.


src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -1478,19 +1485,71 @@ public TrieNode FindCachedOrUnknown(TrieStoreDirtyNodesCache.Key key, bool isRea
...
+                    {
+                        bufferShard.GetOrAdd(key, nodeRecord);
+                        if (!isReadOnly) return nodeRecord.Node;
+                    }
+                    else
+                    {
+                        // clone is as if it read only
+                        TrieNode node = nodeRecord.Node.Clone();
+                        if (nodeRecord.Node.IsSealed) node.Seal();
+                        if (nodeRecord.Node.IsPersisted) node.IsPersisted = true;
```
LukaszRozmej, line 1503: Maybe this should be some `CloneWithX` method?

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -1478,19 +1485,71 @@ public TrieNode FindCachedOrUnknown(TrieStoreDirtyNodesCache.Key key, bool isRea
...
+                if (!isReadOnly)
+                {
+                    return bufferNode;
+                }
+                else
+                {
+                    // we returning a copy to avoid multithreaded access
+                    var trieNode = new TrieNode(NodeType.Unknown, key.Keccak, bufferNode.FullRlp);
+                    trieNode.ResolveNode(_trieStore.GetTrieStore(key.Address), key.Path);
+                    trieNode.Keccak = key.Keccak;
```
LukaszRozmej: why?

asdacap: This do the same thing as `bufferShard.FromCachedRlpOrUnknown`, but by the time it get to that call, it may have been removed from commit buffer, causing the race issue.

LukaszRozmej: maybe put is a helper method then?

Issue #9332, opened 2025-09-22
`System.Reflection.TargetInvocationException` of JSON-RPC `eth_call`


Alleysira:
**Description**

Hello developers.

I'm running a separate Pectra PoS testnet chain for API testing. During testing, I observed discrepancies in the JSON-RPC responses from execution clients—geth, besu, nethermind, reth and erigon. Specifically, when calling `eth_call` with fuzzer generated parameters, nethermind will raise `System.Reflection.TargetInvocationException`, which seems to be a `NullReferenceException`.

```
[...]

PR #9349, opened 2025-09-24
Add transaction params checks in rpc

deffrian:
Fixes Closes Resolves #
#9332

## Changes

- All clients make these checks before actually executing txs. For reference [geth](https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/transaction_args.go)

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.TransactionExecutor.cs
+++ b/src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.TransactionExecutor.cs
@@ -25,11 +25,45 @@ namespace Nethermind.JsonRpc.Modules.Eth
         {
             private bool NoBaseFee { get; set; }
 
-            protected override Transaction Prepare(TransactionForRpc call)
+            private ResultWrapper<Transaction> Fail(string error) => ResultWrapper<Transaction>.Fail(error, ErrorCodes.InvalidInput);
+
+            protected override ResultWrapper<Transaction> Prepare(TransactionForRpc call)
             {
+                if (call is LegacyTransactionForRpc legacyTransaction)
+                {
+                    if (legacyTransaction.To is null && legacyTransaction.Input is null or { Length: 0 })
+                        return Fail(ContractCreationWithoutDataError);
+                }
+
+                if (call is EIP1559TransactionForRpc eip1559Transaction)
+                {
+                    if (eip1559Transaction.GasPrice != null)
+                        return Fail(GasPriceInEip1559Error);
+
+                    if (eip1559Transaction.MaxFeePerGas is not null && eip1559Transaction.MaxFeePerGas == 0)
+                        return Fail(ZeroMaxFeePerGasError);
+
+                    if (eip1559Transaction.MaxFeePerGas < eip1559Transaction.MaxPriorityFeePerGas)
+                        return Fail(MaxFeePerGasSmallerThenMaxPriorityFeePerGasError(
+                            eip1559Transaction.MaxFeePerGas,
+                            eip1559Transaction.MaxPriorityFeePerGas));
+                }
+
+                if (call is BlobTransactionForRpc blobTransaction)
+                {
+                    if (blobTransaction.BlobVersionedHashes is null || blobTransaction.BlobVersionedHashes.Length == 0)
+                        return Fail(AtLeastOneBlobInBlobTransactionError);
+
+                    if (blobTransaction.To is null)
+                        return Fail(MissingToInBlobTxError);
+
+                    if (blobTransaction.MaxFeePerBlobGas is not null && blobTransaction.MaxFeePerBlobGas == 0)
+                        return Fail(ZeroMaxFeePerBlobGasError);
+                }
+
                 var tx = call.ToTransaction();
                 tx.ChainId = _blockchainBridge.GetChainId();
-                return tx;
+                return ResultWrapper<Transaction>.Success(tx);
             }
 
             protected override ResultWrapper<TResult> Execute(BlockHeader header, Transaction tx, Dictionary<Address, AccountOverride>? stateOverride, CancellationToken token)
@@ -38,10 +72,6 @@ namespace Nethermind.JsonRpc.Modules.Eth
                 if (NoBaseFee)
                 {
                     clonedHeader.BaseFeePerGas = 0;
-                }
-                if (tx.IsContractCreation && tx.DataLength == 0)
-                {
-                    return ResultWrapper<TResult>.Fail("Contract creation without any data provided.", ErrorCodes.InvalidInput);
                 }
                 return ExecuteTx(clonedHeader, tx, stateOverride, token);
             }
@@ -84,6 +114,18 @@ namespace Nethermind.JsonRpc.Modules.Eth
 
                 return ResultWrapper<TResult>.Success(bodyData);
             }
+
+            private const string GasPriceInEip1559Error = "both gasPrice and (maxFeePerGas or maxPriorityFeePerGas) specified";
+            private const string AtLeastOneBlobInBlobTransactionError = "need at least 1 blob for a blob transaction";
+            private const string MissingToInBlobTxError = "missing \"to\" in blob transaction";
+            private const string ZeroMaxFeePerBlobGasError = "maxFeePerBlobGas, if specified, must be non-zero";
+            private const string ZeroMaxFeePerGasError = "maxFeePerGas must be non-zero";
+            private static string MaxFeePerGasSmallerThenMaxPriorityFeePerGasError(
+                UInt256? maxFeePerGas,
+                UInt256? maxPriorityFeePerGas)
+                => $"maxFeePerGas ({maxFeePerGas}) < maxPriorityFeePerGas ({maxPriorityFeePerGas})";
+
+            private const string ContractCreationWithoutDataError = "contract creation without any data provided";
         }
 
         private class CallTxExecutor(IBlockchainBridge blockchainBridge, IBlockFinder blockFinder, IJsonRpcConfig rpcConfig)
--- a/src/Nethermind/Nethermind.JsonRpc/Modules/Eth/ExecutorBase.cs
+++ b/src/Nethermind/Nethermind.JsonRpc/Modules/Eth/ExecutorBase.cs
@@ -38,11 +38,15 @@ public abstract class ExecutorBase<TResult, TRequest, TProcessing>
                 ErrorCodes.ResourceUnavailable);
 
         using CancellationTokenSource timeout = _rpcConfig.BuildTimeoutCancellationToken();
-        TProcessing? toProcess = Prepare(call);
-        return Execute(header.Clone(), toProcess, stateOverride, timeout.Token);
+        ResultWrapper<TProcessing> toProcess = Prepare(call);
+        if (toProcess.Result != Result.Success)
+        {
+            return ResultWrapper<TResult>.From(toProcess);
+        }
+        return Execute(header.Clone(), toProcess.Data, stateOverride, timeout.Token);
     }
 
-    protected abstract TProcessing Prepare(TRequest call);
+    protected abstract ResultWrapper<TProcessing> Prepare(TRequest call);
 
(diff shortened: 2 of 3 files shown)


src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.TransactionExecutor.cs
```diff
@@ -25,11 +25,45 @@ private abstract class TxExecutor<TResult>(IBlockchainBridge blockchainBridge, I
...
+                    if (blobTransaction.BlobVersionedHashes is null || blobTransaction.BlobVersionedHashes.Length == 0)
+                        return Fail(AtLeastOneBlobInBlobTransactionError);
+
+                    if (blobTransaction.To is null)
+                        return Fail(MissingToInBlobTxError);
+
+                    if (blobTransaction.MaxFeePerBlobGas is not null && blobTransaction.MaxFeePerBlobGas == 0)
+                        return Fail(ZeroMaxFeePerBlobGasError);
+                }
+
```
LukaszRozmej, line 62: this should be done on OO way, so something like: `bool call.Validate(out string? error)`

src/Nethermind/Nethermind.JsonRpc.Test/Modules/Eth/EthRpcModuleTests.EstimateGas.cs
```diff
@@ -373,12 +373,13 @@ public async Task Eth_estimateGas_ignores_invalid_nonce()
         byte[] code = Prepare.EvmCode
          .Op(Instruction.STOP)
          .Done;
-        TransactionForRpc transaction = new EIP1559TransactionForRpc(Build.A.Transaction
+        EIP1559TransactionForRpc transaction = new(Build.A.Transaction
```
emlautarom1, line 376: Why do we change the type here? Does it affect serialization?

src/Nethermind/Nethermind.JsonRpc.Test/Modules/Eth/EthRpcModuleTests.EthCall.cs
```diff
@@ -423,18 +423,159 @@ public async Task Eth_call_ignores_invalid_nonce()
...
+        LegacyTransactionForRpc transaction = new(Build.A.Transaction
+            .WithGasLimit(100000)
+            .SignedAndResolved(TestItem.PrivateKeyA)
+            .TestObject);
+        transaction.To = null;
+        transaction.Data = data;
+        string serialized = await ctx.Test.TestEthRpc("eth_call", transaction);
+
+        Assert.That(
+            serialized, Is.EqualTo("{\"jsonrpc\":\"2.0\",\"error\":{\"code\":-32000,\"message\":\"contract creation without any data provided\"},\"id\":67}"));
```
emlautarom1, line 472: Probably irrelevant, but some test cases have the same error message with different casing (`Contract creation` vs `contract creation...`). Is this intentional?

src/Nethermind/Nethermind.JsonRpc.Test/Modules/Eth/EthRpcModuleTests.EthCall.cs
```diff
@@ -423,18 +423,159 @@ public async Task Eth_call_ignores_invalid_nonce()
...
+
+        Assert.That(
+            serialized, Is.EqualTo("{\"jsonrpc\":\"2.0\",\"error\":{\"code\":-32000,\"message\":\"need at least 1 blob for a blob transaction\"},\"id\":67}"));
+    }
+
+    [Test]
+    public async Task Eth_call_maxFeePerBlobGas_is_zero()
+    {
+        using Context ctx = await Context.Create();
+        BlobTransactionForRpc transaction = new(Build.A.Transaction
```
emlautarom1, line 513: Just for clarity we might want to set the value of `maxFeePerBlobGas` to `0` rather than asume defaults from the builder.

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.TransactionExecutor.cs
```diff
@@ -25,6 +26,43 @@ private abstract class TxExecutor<TResult>(IBlockchainBridge blockchainBridge, I
         {
             private bool NoBaseFee { get; set; }
 
+            protected override string? Validate(TransactionForRpc call)
```
emlautarom1, line 29: As it is, I don't know how we can support plugins that define their own Rpc classes like Optimism (`DepositTransactionForRpc`) and validate them.

I agree with @LukaszRozmej suggestion: a `abstract bool Validate(out string? error)` method on `TransactionForRpc` would allow us to support new classes.

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.TransactionExecutor.cs
```diff
@@ -84,6 +118,18 @@ protected ResultWrapper<TResult> CreateResultWrapper(bool inputError, string? er
 
                 return ResultWrapper<TResult>.Success(bodyData);
             }
+
+            private const string GasPriceInEip1559Error = "both gasPrice and (maxFeePerGas or maxPriorityFeePerGas) specified";
+            private const string AtLeastOneBlobInBlobTransactionError = "need at least 1 blob for a blob transaction";
+            private const string MissingToInBlobTxError = "missing \"to\" in blob transaction";
+            private const string ZeroMaxFeePerBlobGasError = "maxFeePerBlobGas, if specified, must be non-zero";
+            private const string ZeroMaxFeePerGasError = "maxFeePerGas must be non-zero";
+            private static string MaxFeePerGasSmallerThenMaxPriorityFeePerGasError(
```
emlautarom1, line 127: This is the only case where we show the invalid input values. If we remove them then we can make this a constant string (or maybe we want to make all errors functions?)

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.TransactionExecutor.cs
LukaszRozmej: Maybe better to have `ValidationResult Validate()`

LukaszRozmej: btw refactored it a bit in #9354

src/Nethermind/Nethermind.JsonRpc.Test/Modules/Eth/EthRpcModuleTests.EstimateGas.cs
deffrian: No, it's to set GasPrice

src/Nethermind/Nethermind.JsonRpc.Test/Modules/Eth/EthRpcModuleTests.EthCall.cs
deffrian: There is only one test with capital C. And it uses BeEquivalentTo which is case agnostic. Changed to regular equality

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.TransactionExecutor.cs
deffrian: I don't understand tbh

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.TransactionExecutor.cs
deffrian: I don't like TransactionForRpc handling this validation. 
First, this logic belongs to eth rpc module, not others. Other calls can have different checks and different errors.
Second, if plugins want to add checks they can override Validate function in TxExecutor.
So in short I prefer validation to happen locally rather than pushing it into one place
CC: @LukaszRozmej

LukaszRozmej: If we need extensibility we can have `IRpcTransactionValidator<TTransaction>` per type (but not needed for now), but the cascade of `if`s per type is anti-pattern

deffrian: I also thought about having validator class for every tx but these checks are too simple. I think it's unnecessary abstraction. And regarding `if`s. I don't know how we can avoid them without making Validate method on `TxForRpc`

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.TransactionExecutor.cs
emlautarom1: Just a matter of consistency: all error messages do not include the input values except this case. Since there is no error message standard then it's up to us to decide how they should be: we either include them in all errors, or not include them at all.

Just an opinion on error messages, not a blocker.

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.TransactionExecutor.cs
LukaszRozmej: I still prefer Validate method on `TxForRpc`, makes it easier to implement new ones.
Or maybe let's do create validator interface and allow to register it, but then the `TxForRpc` could also implement this interface for itself and provide a default validation if non other registered.

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.TransactionExecutor.cs
deffrian: I still can't understand. These error messages are kinda standard. So they have to be like this

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.TransactionExecutor.cs
deffrian: I can create a `Validate` function on `TxForRpc`, but we will have to refactor it because some checks will need more context like `SpecProvider`. Validator is not tight to only transaction, but to the context in witch this validation is happening. Same problem happens if we make registration. This will map `TxType` to a `validator`. But it should be `TxType` and context to a `validator`. I think we can do something like `TxExecutorEip1559TxValidator`, `SimulateTxExecutorEip1559TxValidator` and so on. But tbh I think it's just to much machinery to solve few ugly ifs. In the end it will be extremely easy to do this refactor if we need it.

emlautarom1: There are several places where we're needing this "register X for transaction Y" (JSON conversion, Rlp encoding, and now execution validation), which makes me think that we need something a bit more general.

Still, for this case, I think having a virtual method with the appropriate parameters is the most straightforward approach.

LukaszRozmej: My complaint would be that having to derive your executor (potentially multiple executors), just to validate your new tx type, when you are already registering that tx type in different place seems very wrong to me. Register everything at same place.

deffrian: I think adding this now is overkill. It won’t be any harder to implement registration later if we actually need it. And tbh I doubt we ever will. I'd rather not spend time on it right now because it adds complexity without real benefit, and if we realize that an implementation is not good enough, rebuilding will be harder than just moving these ifs

LukaszRozmej: I see your point but I disagree. The point that you had to make the method `virtual` shows this.

emlautarom1 (approved):

PR #9354, opened 2025-09-24
Don't pollute ValidationResult with additional stuff

LukaszRozmej:
## Changes

- Remove context stuff from ValidationResult, it should be usable in any validation scenario

- [x] Refactoring
#### Requires testing

- [x] No

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Nethermind.Consensus.csproj
+++ b/src/Nethermind/Nethermind.Consensus/Nethermind.Consensus.csproj
@@ -13,8 +13,4 @@
     <ProjectReference Include="..\Nethermind.TxPool\Nethermind.TxPool.csproj" />
   </ItemGroup>
 
-  <ItemGroup>
-    <Folder Include="Messages\" />
-  </ItemGroup>
-
 </Project>


src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -725,7 +726,7 @@ private void UpdateGasBottleneckAndMarkForEviction(
 
                     if (!valid)
                     {
-                        MarkForEviction(tx, valid.AllowTxPoolReentrance);
+                        MarkForEviction(tx, valid.Error == TxErrorMessages.InvalidProofVersion);
```
flcl42, line 729: It breaks encapsulation, that additional flag did not affect comparison or any other previous logic

src/Nethermind/Nethermind.Crypto/KzgPolynomialCommitments.cs
```diff
@@ -19,7 +20,7 @@ public static class KzgPolynomialCommitments
             System.Globalization.NumberStyles.Integer);
 
     public const byte KzgBlobHashVersionV1 = 1;
-    public const byte BytesPerBlobVersionedHash = 32;
+    public const byte BytesPerBlobVersionedHash = Eip4844Constants.BytesPerBlobVersionedHash;
```
flcl42: If you move a constant, you can have one instead of two

src/Nethermind/Nethermind.Consensus/Nethermind.Consensus.csproj
```diff
@@ -13,4 +13,8 @@
     <ProjectReference Include="..\Nethermind.TxPool\Nethermind.TxPool.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Folder Include="Messages\" />
```
flcl42: Do we need it?

src/Nethermind/Nethermind.Crypto/KzgPolynomialCommitments.cs
LukaszRozmej: Yeah thought about it, IMO not worth additional changes but I can make them if you insist. It is same constant at the moment.

src/Nethermind/Nethermind.Consensus/Nethermind.Consensus.csproj
LukaszRozmej: No

src/Nethermind/Nethermind.TxPool/TxPool.cs
LukaszRozmej: Does it really? I don't see how.
Before: return this special type of error, add a flag, use the flag for special behavior
Now: return this special type of error, use special behavior on this special error

If you really want to have special flags we need `TxValidationResult`, I wanted this struct to be 64bit.

benaadams (approved):

PR #9356, opened 2025-09-25
Rollback - Precompiles Mutability

svlachakis:
Revert Precompiles back to `HashSet` from `FrozenSet` because it blocks our override in Arbitrum through `SpecProviderDecorator`.

The change was introduced on this PR https://github.com/NethermindEth/nethermind/pull/9274 and after discussing with @deffrian there is no issue with reverting this back.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [X] Other: Rollback

#### Requires testing

- [ ] Yes
- [X] No


LukaszRozmej (changes requested):
Looks like wrong way of doing things. I suggest closing

PR #9371, opened 2025-09-28
Fix beneficiary & blob base fee in eth_simulate

deffrian:
Fixes Closes Resolves #

`eth_simulateV1/ethSimulate-blobs (nethermind)`
Set beneficiary as parent beneficiary.

## Changes

- There was no override in blob base fee.
- For some reason we set miner as zero. I don't know how it passes the tests. Should we change that?

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
+++ b/src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
@@ -23,17 +23,17 @@ using static Nethermind.Evm.EvmObjectFormat.EofValidator;
 namespace Nethermind.Evm.TransactionProcessing
 {
     public sealed class TransactionProcessor(
-        ITransactionProcessor.IGasCalculator gasCalculator,
+        ITransactionProcessor.IBlobBaseFeeCalculator blobBaseFeeCalculator,
         ISpecProvider? specProvider,
         IWorldState? worldState,
         IVirtualMachine? virtualMachine,
         ICodeInfoRepository? codeInfoRepository,
         ILogManager? logManager)
-        : TransactionProcessorBase(gasCalculator, specProvider, worldState, virtualMachine, codeInfoRepository, logManager);
+        : TransactionProcessorBase(blobBaseFeeCalculator, specProvider, worldState, virtualMachine, codeInfoRepository, logManager);
 
-    public class GasCalculator : ITransactionProcessor.IGasCalculator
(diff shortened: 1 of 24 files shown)


src/Nethermind/Nethermind.Facade/Simulate/SimulateReadOnlyBlocksProcessingEnvFactory.cs
```diff
@@ -48,6 +48,7 @@ public ISimulateReadOnlyBlocksProcessingEnv Create()
             .AddDecorator<IBlockhashProvider, SimulateBlockhashProvider>()
             .AddDecorator<IVirtualMachine, SimulateVirtualMachine>()
             .AddDecorator<IBlockValidator, SimulateBlockValidatorProxy>()
+            .AddSingleton<ITransactionProcessor, SimulateTransactionProcessor>()
```
asdacap: Got a solution that does not involve a subclass of `SimulateTransactionProcessor`? This will break plugins that override `ITransactionProcessor`. Perhaps some kind gas calculator interface  maybe?

deffrian: I would move all virtual functions into an interface, but it's overkill probably

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -23,12 +23,22 @@
...
+        ITransactionProcessor.IGasCalculator gasCalculator,
         ISpecProvider? specProvider,
         IWorldState? worldState,
         IVirtualMachine? virtualMachine,
         ICodeInfoRepository? codeInfoRepository,
         ILogManager? logManager)
-        : TransactionProcessorBase(specProvider, worldState, virtualMachine, codeInfoRepository, logManager);
+        : TransactionProcessorBase(gasCalculator, specProvider, worldState, virtualMachine, codeInfoRepository, logManager);
+
+    public class GasCalculator : ITransactionProcessor.IGasCalculator
```
LukaszRozmej: Maybe better `BaseFeeCalculator`? It is not calculating gas per se.

PR #9373, opened 2025-09-28
fix eth_config missing checks

nethoxa:
## Changes

- Added check that sets `Last` to `null` if `Next` is unavailable, according to the [specs](https://github.com/ethereum/EIPs/blob/90a2b6fb5c84efdf0079b65a7f1ecab254562ff9/EIPS/eip-7910.md?plain=1#L54):
> "next" and "last" will be `null` if the client is not configured to support a future fork.
- Added a check that ensures `Activation.Timestamp` to be non-null according to the [specs](https://github.com/ethereum/EIPs/blob/90a2b6fb5c84efdf0079b65a7f1ecab254562ff9/EIPS/eip-7910.md?plain=1#L66C88-L66C204)
> If the fork is not scheduled to be activated or its activation time is unknown, it should not be in the RPC results.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
+++ b/src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
@@ -819,7 +819,7 @@ public partial class EthRpcModule(
 
         static ForkConfig? GetForkConfig(Fork? fork, ISpecProvider specProvider)
         {
-            if (fork is null || fork.Value.Activation.Timestamp is null)
+            if (fork is null)
             {
                 return null;
             }


src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
```diff
@@ -806,16 +806,20 @@ public ResultWrapper<JsonNode> eth_config()
     {
         ForkActivationsSummary forks = forkInfo.GetForkActivationsSummary(_blockFinder.Head?.Header);
 
+        ForkConfig? current = GetForkConfig(forks.Current, _specProvider);
+        ForkConfig? next = GetForkConfig(forks.Next, _specProvider);
+        ForkConfig? last = next is not null ? GetForkConfig(forks.Last, _specProvider) : null;
```
LukaszRozmej, line 811: This seems fine, though the spec is weird here as intuitively `current`==`last` should be a valid response, but it prefers `last` to be only future forks so keeps a `null`

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
```diff
@@ -806,16 +806,20 @@ public ResultWrapper<JsonNode> eth_config()
...
-            Last = GetForkConfig(forks.Last, _specProvider)
+            Current = current,
+            Next = next,
+            Last = last
         }), UnchangedDictionaryKeyOptions)));
 
         static ForkConfig? GetForkConfig(Fork? fork, ISpecProvider specProvider)
         {
-            if (fork is null)
+            if (fork is null || fork.Value.Activation.Timestamp is null)
```
LukaszRozmej: While this is probably fine for post-merge networks, it completely ignores any pre-merge network forks. Which I expect were out of scope of the EIP.

This code: https://github.com/NethermindEth/nethermind/blob/6c1143775b45222d2b12883c7758e4d0273abb4d/src/Nethermind/Nethermind.Specs/ChainSpecStyle/SpecProviderBase.cs#L36-L37
Ensures that we can't have a Fork without Timestamp transition activation after we had first one with timestamp. Thus the data structure already guarantees this.

So this check is more about if `eth_config` should even function before the merge - not a problem for mainnet but for other chains if they are not merged yet.

I'm inclined to say yes and removing this check.

@flcl42

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
nethoxa: Yeah, I agree the spec is kinda weird, but just wanna make all client behave the same

PR #9375, opened 2025-09-29
Refactor/Nicer IWorldStateScopeProvider interface

asdacap:
Require #9089 

- Make the interface nicer and more future proof.
- Remove `IStateTree`, storage root and update storage root. 
- Comes at a downside of an additional dictionary and concurrent queue. 
- All set goes through a single write batch. 

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

#### Notes on testing

- [x] Mainnet can sync
- [ ] Sepolia era replay
- [ ] Mainnet era replay.

### DIFF
--- a/src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
+++ b/src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Runtime.Intrinsics;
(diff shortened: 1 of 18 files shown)


src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
```diff
@@ -263,24 +264,25 @@ void UpdateRootHashesSingleThread()
...
 
         void UpdateRootHashesMultiThread()
         {
             // We can recalculate the roots in parallel as they are all independent tries
             using var storages = _storages.ToPooledList();
+            // Cannot be called concurrently so we gather them here.
+            Dictionary<AddressAsKey, IWorldStateScopeProvider.IStorageWriteBatch> storageWriteBatches =
+                _storages.ToDictionary(
+                    kv => kv.Key,
+                    kv => writeBatch.CreateStorageWriteBatch(kv.Key, kv.Value.EstimatedChanges));
```
LukaszRozmej: This seems avoidable. Maybe not nice but can `_storages` have a value of `(PerContractState, IStorageWriteBatch)` - but better as a class not a struct

PR #9379, opened 2025-09-29
Fix estimate gas & add timestamp to logs

deffrian:
Fixes Closes Resolves #

```
eth_estimateGas/estimate-simple-transfer (nethermind)
eth_getBlockReceipts/get-block-receipts-latest (nethermind)
eth_getLogs/contract-addr (nethermind)
eth_getLogs/no-topics (nethermind)
eth_getLogs/topic-exact-match (nethermind)
eth_getLogs/topic-wildcard (nethermind)
eth_getTransactionReceipt/get-access-list (nethermind)
eth_getTransactionReceipt/get-blob-tx (nethermind)
eth_getTransactionReceipt/get-dynamic-fee (nethermind)
```

## Changes

- Add block timestamp to logs.
- Estimate gas is errorMargin problem. Just return 21k for simple transfers.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain/Tracing/GasEstimator.cs
+++ b/src/Nethermind/Nethermind.Blockchain/Tracing/GasEstimator.cs
@@ -77,7 +77,7 @@ public class GasEstimator(
         }
 
         // If transaction is simple transfer return intrinsic gas
-        if (tx.To is not null && tx.Data.IsEmpty) return lowerBound;
+        if (tx.To is not null && tx.Data.IsEmpty && TryExecutableTransaction(tx, header, lowerBound, token, gasTracer)) return lowerBound;
 
         // Execute binary search to find the optimal gas estimation.
         return BinarySearchEstimate(leftBound, rightBound, tx, header, gasTracer, errorMargin, token, out err);


src/Nethermind/Nethermind.Blockchain/Tracing/GasEstimator.cs
```diff
@@ -76,6 +76,9 @@ public long Estimate(Transaction tx, BlockHeader header, EstimateGasTracer gasTr
             return 0;
         }
 
+        // If transaction is simple transfer return intrinsic gas
+        if (tx.To is not null && tx.Data.IsEmpty) return lowerBound;
```
alexb5dh: Does it need to handle the case of sending to a contract with a fallback function?

src/Nethermind/Nethermind.JsonRpc/Converters/TxReceiptConverter.cs
```diff
@@ -23,7 +23,7 @@ public class TxReceiptConverter : JsonConverter<TxReceipt>
     public override void Write(Utf8JsonWriter writer, TxReceipt value, JsonSerializerOptions options)
     {
         writer.WriteStartObject();
-        var receipt = new ReceiptForRpc(value.TxHash!, value, default);
+        var receipt = new ReceiptForRpc(value.TxHash!, value, 0, default);
```
alexb5dh, line 26: It's `0` here just because it's not used anywhere after? Would add a comment about that.

deffrian: I think this function is intended to serialize regular TxReceipt. And both blockTimestamp and gasInfo are not part of TxReceipt. not sure what I should write here

src/Nethermind/Nethermind.Blockchain/Tracing/GasEstimator.cs
deffrian: Didn't know this thing exists🥲

PR #9380, opened 2025-09-29
Disable `Flashbots` module by default

emlautarom1:
## Changes

- Disable the Flashbots module and it's associated RPC endpoints (`flashbots`, `rbuilder`) by default. 

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

Tested a couple of configs to verify that the module is not being loaded or any RPC method is added.

- [ ] Yes
- [x] No

We should not need to update anything since the docs are generated from configuration attributes, which listed the Flashbot module as disabled by default (this was actually not the case though).

- [ ] Yes
- [x] No

## Remarks

Despite the config listing Flashbots as a disabled module by default, our code actually enabled it. This PR disables the plugin by default.

To explicitly enable the plugin include the flag `--Flashbots.Enabled=true`. Also, to enable the RPC endpoints include `--JsonRpc.EnabledModules=flashbots,rbuilder,...` as appropriate (it could be argued that enabling the plugin should automatically include the RPC endpoints though).

### DIFF
--- a/src/Nethermind/Nethermind.JsonRpc/Modules/ModuleType.cs
+++ b/src/Nethermind/Nethermind.JsonRpc/Modules/ModuleType.cs
@@ -31,33 +31,6 @@ namespace Nethermind.JsonRpc.Modules
         public const string Rpc = nameof(Rpc);
         public const string Rbuilder = nameof(Rbuilder);
 
-        public static IEnumerable<string> AllBuiltInModules { get; } = new List<string>()
-        {
-            Admin,
-            Clique,
-            Engine,
-            Db,
-            Debug,
-            Erc20,
-            Eth,
-            Evm,
-            Flashbots,
-            Net,
-            Nft,
-            Parity,
-            Personal,
-            Proof,
-            Subscribe,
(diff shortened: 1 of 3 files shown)


src/Nethermind/Nethermind.JsonRpc/Modules/ModuleType.cs
```diff
@@ -31,33 +31,6 @@ public static class ModuleType
         public const string Rpc = nameof(Rpc);
         public const string Rbuilder = nameof(Rbuilder);
 
-        public static IEnumerable<string> AllBuiltInModules { get; } = new List<string>()
```
LukaszRozmej, line 34: wasn't used?

emlautarom1: No, no usages or references anywhere.

PR #9381, opened 2025-09-29
Xdc TimeoutCertificateManager - Timeout Certificate Validation

cicr99:


src/Nethermind/Nethermind.Xdc/EpochSwitchInfo.cs
ak88, line 1: why is this deleted?

src/Nethermind/Nethermind.Xdc/Types/TimeoutForSign.cs
ak88, line 1: The usual pattern is to use RlpBehavior.ForSealing in the decoder to exclude sealing parameters, instead of using a separate class.

src/Nethermind/Nethermind.Xdc/TimeoutCertificateManager.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        if (snapshot.NextEpochCandidates.Length == 0)
+        {
+            errorMessage = "Empty master node lists from snapshot";
+            return false;
+        }
+
+        var signatures = new HashSet<Signature>(timeoutCert.Signatures);
+        BlockHeader header = _blockTree.Head?.Header;
+        if (header is not XdcBlockHeader xdcHeader)
```
ak88, line 64: If this happens something has screwed up completely

src/Nethermind/Nethermind.Xdc/TimeoutCertificateManager.cs
```diff
@@ -0,0 +1,99 @@
...
+            throw new ArgumentException($"Only type of {nameof(XdcBlockHeader)} is allowed");
```
ak88: would not be an argument exception in this case, but rather something like InvalidOperationException

src/Nethermind/Nethermind.Xdc/TimeoutCertificateManager.cs
```diff
@@ -0,0 +1,99 @@
...
+        IXdcReleaseSpec spec = _specProvider.GetXdcSpec(xdcHeader, timeoutCert.Round);
+        EpochSwitchInfo epochInfo = _epochSwitchManager.GetTimeoutCertificateEpochInfo(timeoutCert);
+        if (epochInfo is null)
+        {
+            errorMessage = "Failed to get epoch switch info for timeout certificate";
+            return false;
+        }
+        if (signatures.Count < epochInfo.Masternodes.Length * spec.CertThreshold)
+        {
+            errorMessage = "Invalid TC Signatures";
```
ak88: A more descriptive message like TC does not meet threshold and then include all the numbers, would be very helpful for debugging.

src/Nethermind/Nethermind.Xdc/TimeoutCertificateManager.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        var signatures = new HashSet<Signature>(timeoutCert.Signatures);
+        BlockHeader header = _blockTree.Head?.Header;
+        if (header is not XdcBlockHeader xdcHeader)
+            throw new ArgumentException($"Only type of {nameof(XdcBlockHeader)} is allowed");
+        IXdcReleaseSpec spec = _specProvider.GetXdcSpec(xdcHeader, timeoutCert.Round);
+        EpochSwitchInfo epochInfo = _epochSwitchManager.GetTimeoutCertificateEpochInfo(timeoutCert);
+        if (epochInfo is null)
+        {
+            errorMessage = "Failed to get epoch switch info for timeout certificate";
```
ak88: Include the round number or whatever is used to get the epoch info

src/Nethermind/Nethermind.Xdc/TimeoutCertificateManager.cs
```diff
@@ -0,0 +1,99 @@
...
+        if (signatures.Count < epochInfo.Masternodes.Length * spec.CertThreshold)
+        {
+            errorMessage = "Invalid TC Signatures";
+            return false;
+        }
+
+        ValueHash256 signedTimeoutObj = new TimeoutForSign(timeoutCert.Round, timeoutCert.GapNumber).SigHash().ValueHash256;
+        bool allValid = true;
+        Parallel.ForEach(signatures,
+            new ParallelOptions { MaxDegreeOfParallelism = Environment.ProcessorCount },
```
ak88: Believe MaxDegreeOfParallelism is set to that by default

src/Nethermind/Nethermind.Crypto/PrivateKeyGenerator.cs
```diff
@@ -22,13 +24,21 @@ public PrivateKeyGenerator(ICryptoRandom cryptoRandom)
         }
 
         public PrivateKey Generate()
+        {
+            return Generate(1).First();
+        }
+        public IEnumerable<PrivateKey> Generate(int number)
```
ak88, line 30: I think this conflicts with my PR, but leave it for now

src/Nethermind/Nethermind.Xdc/EpochSwitchInfo.cs
cicr99: There is a duplicated EpochSwitchInfo inside Types folder. I assumed that's where it should go, so deleted this one

src/Nethermind/Nethermind.Xdc/Types/TimeoutForSign.cs
cicr99: fixed

src/Nethermind/Nethermind.Xdc/TimeoutCertificateManager.cs
cicr99: You are right, fixed

src/Nethermind/Nethermind.Xdc/TimeoutCertificateManager.cs
cicr99: Well `MaxDegreeOfParallelism` is set to -1 by default, I actually went with the same code there was for quorum certificate manager. But after looking in the [docs](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.paralleloptions.maxdegreeofparallelism?view=net-9.0&viewFallbackFrom=net-10.0-pp#remarks) I think it's best to just go with the default one.

src/Nethermind/Nethermind.Xdc/Types/Timeout.cs
```diff
@@ -5,18 +5,19 @@
...
 {
     private Address signer;
 
     public ulong Round { get; set; } = round;
-    public Signature Signature { get; set; } = signature;
+    public Signature? Signature { get; set; } = signature;
     public ulong GapNumber { get; set; } = gapNumber;
 
-    public Hash256 SigHash() => Keccak.Compute(Rlp.Encode(this).Bytes);
+    public ValueHash256 SigHash() => Keccak.Compute(new TimeoutDecoder().Encode(this, RlpBehaviors.ForSealing).Bytes).ValueHash256;
```
ak88: We dont have a hash method for other items. You can create a static decoder in the Manager class and create the hash in a method there.

src/Nethermind/Nethermind.Xdc/TimeoutCertificateManager.cs
```diff
@@ -0,0 +1,99 @@
...
+            errorMessage = $"Failed to get epoch switch info for timeout certificate with round {timeoutCertificate.Round}";
+            return false;
+        }
+        if (signatures.Count < epochInfo.Masternodes.Length * spec.CertThreshold)
+        {
+            errorMessage = $"Number of unique signatures {signatures.Count} does not meet threshold of {epochInfo.Masternodes.Length * spec.CertThreshold}";
+            return false;
+        }
+
+        ValueHash256 timeoutMsgHash = new Timeout(timeoutCertificate.Round, null, timeoutCertificate.GapNumber).SigHash();
```
ak88: Make a static decoder in this class instead.

src/Nethermind/Nethermind.Xdc/TimeoutCertificateManager.cs
```diff
@@ -0,0 +1,99 @@
...
+            return false;
+        }
+        var nextEpochCandidates = new HashSet<Address>(snapshot.NextEpochCandidates);
+
+        var signatures = new HashSet<Signature>(timeoutCertificate.Signatures);
+        BlockHeader header = _blockTree.Head?.Header;
+        if (header is not XdcBlockHeader xdcHeader)
+            throw new InvalidOperationException($"Only type of {nameof(XdcBlockHeader)} is allowed");
+        IXdcReleaseSpec spec = _specProvider.GetXdcSpec(xdcHeader, timeoutCertificate.Round);
+        EpochSwitchInfo epochInfo = _epochSwitchManager.GetTimeoutCertificateEpochInfo(timeoutCertificate);
```
ak88, line 67: this should probably be an extention method or just a private one here, since i doubt it will be used anywhere else.

ak88 (commented):
If you just move the hash method to the manager, we can merge

src/Nethermind/Nethermind.Xdc/TimeoutCertificateManager.cs
cicr99: I was implementing it as a private method in this class at the beginning (I also don't think it's used anywhere else), until I realized it was declared in IEpochSwitchManager, so @Demuirgos would probably have implemented it on his side. Perhaps we could leave this as it is now in order to merge it, and once his PR is up we could move that logic into the TimeoutCert manager instead

src/Nethermind/Nethermind.Xdc/Types/Timeout.cs
```diff
@@ -5,19 +5,18 @@
...
     public ulong Round { get; set; } = round;
-    public Signature Signature { get; set; } = signature;
+    public Signature? Signature { get; set; } = signature;
     public ulong GapNumber { get; set; } = gapNumber;
 
-    public Hash256 SigHash() => Keccak.Compute(Rlp.Encode(this).Bytes);
-
     public override string ToString() => $"{Round}:{GapNumber}";
 
     public Address GetSigner() => signer;
```
ak88:
```suggestion

```

ak88 (approved):

PR #9382, opened 2025-09-30
Make txpool test run faster

asdacap:
- Txpool test complete in 12 minute which is 3 time longer than the longest other test. This PR try to make it shorter down to 5 minute.
- Make txpool test run faster.
- Make parallel some part of the code.
- Reduce repetition or private key count on other.
- Fix randomly failing pyspec test

- [X] Optimization

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes: Its all test
- [ ] No

#### Notes on testing

- Seems to work.

### DIFF
--- a/src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.cs
+++ b/src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.cs
@@ -15,6 +15,7 @@ using Nethermind.Consensus.Comparers;
 using Nethermind.Consensus.Transactions;
 using Nethermind.Consensus.Validators;
 using Nethermind.Core;
+using Nethermind.Core.Collections;
 using Nethermind.Core.Crypto;
 using Nethermind.Core.Events;
 using Nethermind.Core.Extensions;
@@ -563,7 +564,7 @@ namespace Nethermind.TxPool.Test
 
             EnsureSenderBalance(TestItem.AddressA, (UInt256)(oneTxPrice * numberOfTxsPossibleToExecuteBeforeGasExhaustion));
 
-            for (int i = 0; i < 10; i++)
+            Parallel.For(0, 10, i =>
             {
                 transactions[i] = Build.A.Transaction
                     .WithSenderAddress(TestItem.AddressA)
@@ -572,6 +573,10 @@ namespace Nethermind.TxPool.Test
                     .WithGasLimit(_txGasLimit)
                     .WithValue(value)
                     .SignedAndResolved(_ethereumEcdsa, TestItem.PrivateKeyA).TestObject;
+            });
+
+            for (int i = 0; i < 10; i++)
+            {
                 _txPool.SubmitTx(transactions[i], TxHandlingOptions.PersistentBroadcast);
             }
 
@@ -591,16 +596,22 @@ namespace Nethermind.TxPool.Test
 
             EnsureSenderBalance(TestItem.AddressA, (UInt256)(oneTxPrice * numberOfTxsPossibleToExecuteBeforeGasExhaustion));
 
-            for (int i = 0; i < numberOfTxsPossibleToExecuteBeforeGasExhaustion * 2; i++)
+            int count = numberOfTxsPossibleToExecuteBeforeGasExhaustion * 2;
+            using ArrayPoolList<Transaction> transactions = new(count, count);
+            Parallel.For(0, count, i =>
             {
-                Transaction tx = Build.A.Transaction
+                transactions[i] = Build.A.Transaction
                     .WithSenderAddress(TestItem.AddressA)
                     .WithNonce((UInt256)i)
                     .WithGasPrice((UInt256)gasPrice)
                     .WithGasLimit(_txGasLimit)
                     .WithValue(value)
                     .SignedAndResolved(_ethereumEcdsa, TestItem.PrivateKeyA).TestObject;
-                _txPool.SubmitTx(tx, TxHandlingOptions.PersistentBroadcast);
+            });
+
+            for (int i = 0; i < count; i++)
+            {
+                _txPool.SubmitTx(transactions[i], TxHandlingOptions.PersistentBroadcast);
 
                 if (i < numberOfStaleTxsInBucket)
                 {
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.State/StateProvider.cs
```diff
@@ -547,6 +547,9 @@ public void Commit(IReleaseSpec releaseSpec, IWorldStateTracer stateTracer, bool
                 {
                     FlushToTree();
                 }
+
+                codeFlushTask.GetAwaiter().GetResult();
+
```
LukaszRozmej: This is explicit optimization by @benaadams not to do it here, but in the background, then in `Commit` you have wait for the one from previous block.

src/Nethermind/Nethermind.TxPool.Test/TxBroadcasterTests.cs
```diff
@@ -265,7 +265,10 @@ public void should_skip_blob_txs_when_picking_best_persistent_txs_to_broadcast([
                 .WithShardBlobTxTypeAndFieldsIfBlobTx()
                 .SignedAndResolved(_ethereumEcdsa, TestItem.PrivateKeys[i])
                 .TestObject;
+        });
 
+        for (int i = 0; i < addedTxsCount; i++)
+        {
```
LukaszRozmej: Broadcast should be thread safe so you can put it under `Parallel.For`

src/Nethermind/Nethermind.TxPool.Test/TxBroadcasterTests.cs
```diff
@@ -423,14 +426,17 @@ public void should_not_pick_blob_txs_with_MaxFeePerBlobGas_lower_than_CurrentFee
...
         {
             transactions[i] = Build.A.Transaction
                 .WithShardBlobTxTypeAndFields()
                 .WithMaxFeePerBlobGas(i.GWei())
                 .SignedAndResolved(_ethereumEcdsa, TestItem.PrivateKeys[i])
                 .TestObject;
+        });
 
+        for (int i = 0; i < addedTxsCount; i++)
+        {
```
LukaszRozmej, line 436: Same here

src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.Blobs.cs
```diff
@@ -132,17 +132,22 @@ public void should_reject_txs_with_nonce_too_far_in_future(TxType txType, int ma
...
                     .WithShardBlobTxTypeAndFieldsIfBlobTx()
                     .WithMaxFeePerGas(1.GWei())
                     .WithMaxPriorityFeePerGas(1.GWei())
                     .SignedAndResolved(_ethereumEcdsa, TestItem.PrivateKeyA).TestObject;
+            });
 
-                _txPool.SubmitTx(tx, TxHandlingOptions.None).Should().Be(nonce > expectedNumberOfAcceptedTxs
+            for (int nonce = 0; nonce < txPoolConfig.Size; nonce++)
+            {
+                _txPool.SubmitTx(txs[nonce], TxHandlingOptions.None).Should().Be(nonce > expectedNumberOfAcceptedTxs
```
LukaszRozmej, line 150: `TxPool.SubmitTx` should be thread safe, the behavior might diverge though if there is a big gap in nonces though. So potentially it could be parallelized with some batching?

LukaszRozmej (commented):
@marcindsobczak could you take a look?

src/Nethermind/Nethermind.State/StateProvider.cs
asdacap: Really? Then what is with the other wait at the end of the same function?

src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.Blobs.cs
marcindsobczak: You can also reduce TxPool size in first 2 cases - we don't really need 100 if it's slow. 10 should be enough

src/Nethermind/Nethermind.TxPool.Test/TxBroadcasterTests.cs
marcindsobczak: At least it should be - order doesn't matter here so worth to move to parallelized part and verify it

src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.Blobs.cs
marcindsobczak: ~~In this test we are accepting first N (N = TxPool.Size) transactions and returning `AcceptTxResult.NonceTooFarInFuture` for further ones. If we parallelize it, we can accept N+1 and higher nonces and later evict when lower nonces arrive. So we can stick with non-parallelized solution or refactor test to not check acceptance at the moment of adding, but to verify the highest nonce in the pool at the very end~~

marcindsobczak: Ah "big gap in nonces" is configured by `MaxPendingTxsPerSender` and `MaxPendingBlobTxsPerSender`. We are accepting current nonce + `MaxPending...` as the maximal nonce. It is configured at the test level so actually we should reject N+1 nonces and accept only up to N. So for non-blobs should be parallelizable even without refactoring `AcceptTxResult.NonceTooFarInFuture`

marcindsobczak: But for blob txs we are not accepting ANY gaps so adding must be consecutive

src/Nethermind/Nethermind.State/StateProvider.cs
benaadams: I have a PR that moves it to end of block processing; though is horribly merged conflicted atm https://github.com/NethermindEth/nethermind/pull/9012

LukaszRozmej: So @benaadams is this PR acceptable or makes your solution unachievable and you are confident you will revisit it?

asdacap:
I just want faster test 😢.

PR #9385, opened 2025-09-30
Fix ArrayPoolList disposal warning in Block.AccountChanges

bshastry:
# Fix ArrayPoolList disposal warning in Block.AccountChanges

## Changes

- Added `Block.DisposeAccountChanges()` method to properly dispose and null the AccountChanges list
- Updated `TxPool.DisposeBlockAccountChanges()` to use the new disposal method for better encapsulation
- Added disposal calls in `BlockchainTestBase` to clean up AccountChanges after genesis and test blocks
- Added three unit tests in `BlockTests` to verify proper disposal behavior and prevent finalizer warnings

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

Added three comprehensive unit tests in `BlockTests.cs`:
1. `DisposeAccountChanges_should_dispose_and_null_account_changes` - Verifies proper disposal and nulling
2. `DisposeAccountChanges_should_handle_null_account_changes` - Verifies safe handling of null values
3. `DisposeAccountChanges_should_prevent_finalizer_warning` - Verifies prevention of DEBUG-mode finalizer warnings

All existing tests pass (2,302 passed, 0 failed). The fix was validated by running block tests in DEBUG configuration where the finalizer warning would appear.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### Problem
When running large numbers of block tests (~200k) with `nethtest -b -i`, intermittent finalizer warnings appeared in DEBUG builds:

```
Warning: ArrayPoolList was not disposed. Created at:
   at Nethermind.Core.Collections.ArrayPoolList`1..ctor(...)
   at Nethermind.State.StateProvider.ChangedAddresses()
   at Nethermind.State.WorldState.GetAccountChanges()
   at Nethermind.Consensus.Processing.BlockProcessor.ProcessBlock(...)
```

### Root Cause
`Block.AccountChanges` holds an `ArrayPoolList<AddressAsKey>` that was not being disposed, preventing pooled arrays from returning to `ArrayPool<T>.Shared`. This caused memory leaks over many test iterations.

### Solution
Introduced proper disposal mechanism by adding `Block.DisposeAccountChanges()` and ensuring it's called in test cleanup paths. The `TxPool` already had disposal logic, but this change encapsulates it properly in the Block class and extends it to test infrastructure.

### Impact
- Prevents memory leaks from undisposed ArrayPoolList instances
- Eliminates DEBUG-mode finalizer warnings
- Improves resource management in high-volume testing scenarios
- No breaking changes or performance impact

### DIFF
--- a/src/Nethermind/Nethermind.Core.Test/BlockTests.cs
+++ b/src/Nethermind/Nethermind.Core.Test/BlockTests.cs
@@ -27,52 +27,12 @@ internal class BlockTests
     [Test]
     public void DisposeAccountChanges_should_dispose_and_null_account_changes()
     {
-        // Arrange
         Block block = new(new BlockHeader());
         block.AccountChanges = new ArrayPoolList<AddressAsKey>(10);
         block.AccountChanges.Add(TestItem.AddressA);
 
-        // Act
         block.DisposeAccountChanges();
 
-        // Assert
-        block.AccountChanges.Should().BeNull();
-    }
-
-    [Test]
-    public void DisposeAccountChanges_should_handle_null_account_changes()
-    {
-        // Arrange
-        Block block = new(new BlockHeader());
-        block.AccountChanges = null;
-
-        // Act
-        Action act = () => block.DisposeAccountChanges();
-
-        // Assert
-        act.Should().NotThrow();
-        block.AccountChanges.Should().BeNull();
-    }
-
-    [Test]
-    public void DisposeAccountChanges_should_prevent_finalizer_warning()
-    {
-        // This test verifies that disposing AccountChanges prevents the DEBUG-mode finalizer warning
-        // Arrange
-        Block block = new(new BlockHeader());
-        block.AccountChanges = new ArrayPoolList<AddressAsKey>(10);
-        block.AccountChanges.Add(TestItem.AddressA);
-
-        // Act - dispose before GC
-        block.DisposeAccountChanges();
-
-        // Force GC to trigger any finalizers
-        GC.Collect();
-        GC.WaitForPendingFinalizers();
-        GC.Collect();
-
-        // Assert - if the ArrayPoolList wasn't disposed, the finalizer would print a warning
-        // This test passes if no warning is printed (in DEBUG mode)
         block.AccountChanges.Should().BeNull();
     }
 }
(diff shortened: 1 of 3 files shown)


src/Nethermind/Nethermind.Core/Block.cs
```diff
@@ -179,6 +179,16 @@ private string ToFullString()
...
+    /// <summary>
+    /// Disposes the AccountChanges ArrayPoolList if present, returning the pooled array.
+    /// This should be called when the block is no longer needed to prevent memory leaks.
+    /// </summary>
+    public void DisposeAccountChanges()
+    {
+        AccountChanges?.Dispose();
+        AccountChanges = null;
+    }
+
```
LukaszRozmej: I would prefer to move it to extension method, the mere existence of `AccountChanges` is polluting block interface, but we can live with it for performance gains, but I would like to keep the things directly in `Block` to minimum (check `BlockExtensions` classes)

src/Nethermind/Nethermind.Core.Test/BlockTests.cs
```diff
@@ -20,4 +23,56 @@ public void Should_init_withdrawals_in_body_as_expected((BlockHeader Header, int
...
+
+        // Force GC to trigger any finalizers
+        GC.Collect();
+        GC.WaitForPendingFinalizers();
+        GC.Collect();
+
+        // Assert - if the ArrayPoolList wasn't disposed, the finalizer would print a warning
+        // This test passes if no warning is printed (in DEBUG mode)
+        block.AccountChanges.Should().BeNull();
+    }
```
LukaszRozmej, line 37: To be honest looks like an overkill for such a small refactor. 1 simple test would suffice.

bshastry: Thank you. Cut down to a single test

src/Nethermind/Nethermind.Core/Block.cs
bshastry: Done!

benaadams:

PR #9388, opened 2025-09-30
Add support for block producer based on global world state instance.

damian-orzechowski:
Fixes Closes Resolves #

## Changes

Adds support for block producer based on global world state instance. Assumptions are:
- block producing on `GlobalWorldState`
- receipts are stored on default storage (persistent instead on `NullReceiptStorage`)
- Block production doesn't trigger block being suggested to block tree (and subsequently validated) for pre-merge chains

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Remarks

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Producers/BlockProducerBase.cs
+++ b/src/Nethermind/Nethermind.Consensus/Producers/BlockProducerBase.cs
@@ -34,7 +34,7 @@ namespace Nethermind.Consensus.Producers
     /// </summary>
     public abstract class BlockProducerBase : IBlockProducer
     {
-        private IBlockchainProcessor Processor { get; }
+        protected IBlockchainProcessor Processor { get; }
         protected IBlockTree BlockTree { get; }
         private ITimestamper Timestamper { get; }
 
--- a/src/Nethermind/Nethermind.Consensus/Producers/GlobalWorldStateBlockProducerEnvFactory.cs
+++ b/src/Nethermind/Nethermind.Consensus/Producers/GlobalWorldStateBlockProducerEnvFactory.cs
@@ -1,0 +1,48 @@
+// SPDX-FileCopyrightText: 2022 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using Autofac;
+using Nethermind.Consensus.Processing;
+using Nethermind.Consensus.Transactions;
+using Nethermind.Consensus.Withdrawals;
+using Nethermind.Core;
+using Nethermind.Evm.TransactionProcessing;
+using Nethermind.State;
+
+namespace Nethermind.Consensus.Producers
+{
+    /// <summary>
+    /// Block producer environment factory that uses the global world state and stores receipts by default.
+    /// It will not trigger suggesting produced blocks.
+    /// </summary>
+    /// <param name="rootLifetime"></param>
+    /// <param name="worldStateManager"></param>
+    /// <param name="txSourceFactory"></param>
+    public class GlobalWorldStateBlockProducerEnvFactory(
+        ILifetimeScope rootLifetime,
+        IWorldStateManager worldStateManager,
+        IBlockProducerTxSourceFactory txSourceFactory)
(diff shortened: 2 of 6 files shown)


src/Nethermind/Nethermind.Consensus/Producers/BlockProducerEnvFactory.cs
```diff
@@ -27,6 +27,7 @@ protected virtual ContainerBuilder ConfigureBuilder(ContainerBuilder builder) =>
             .AddScoped<IBlockProcessor.IBlockTransactionsExecutor, BlockProcessor.BlockProductionTransactionsExecutor>()
             .AddDecorator<IWithdrawalProcessor, BlockProductionWithdrawalProcessor>()
             .AddDecorator<IBlockchainProcessor, OneTimeChainProcessor>()
+            .AddScoped<IProducedBlockSuggester, ProducedBlockSuggester>()
```
damian-orzechowski: I'd much more prefer to not register any instance by default and have each module register `suggester` depending on their need, but I'm not sure we can separate all pre-merge configurations to have it registered (like. for instance plugins do via a separate module).
This would be much cleaner than overriding with empty `NoOpProducedBlockSuggester`

LukaszRozmej:
Could we have a flag that would enable this feature?

damian-orzechowski:
> Could we have a flag that would enable this feature?

I was thinking that the use case is quite specific, so it would rather require to make further changes on top of `GlobalWorldStateBlockProducerEnvFactory`. What could be the scenarios when it could be useful just with a config flag?

src/Nethermind/Nethermind.Consensus/Producers/BlockProducerEnvFactory.cs
damian-orzechowski: Moved registration to parent scope as it depends on `IBlockProducerRunner` which is registered in parents' scope.

src/Nethermind/Nethermind.Consensus/Producers/BlockProducerBase.cs
```diff
@@ -189,8 +189,13 @@ protected BlockProducerBase(
...
             Sealer.SealBlock(block, token);
 
-        protected virtual Block? ProcessPreparedBlock(Block block, IBlockTracer? blockTracer, CancellationToken token = default) =>
-            Processor.Process(block, ProcessingOptions.ProducingBlock, blockTracer ?? NullBlockTracer.Instance, token);
+        protected virtual Block? ProcessPreparedBlock(Block block, IBlockTracer? blockTracer,
+            CancellationToken token = default)
+        {
+            if (_blocksConfig.BuildBlocksOnMainState)
+                return Processor.Process(block, ProcessingOptions.NoValidation | ProcessingOptions.StoreReceipts | ProcessingOptions.DoNotUpdateHead, blockTracer ?? NullBlockTracer.Instance, token);
+            return Processor.Process(block, ProcessingOptions.ProducingBlock, blockTracer ?? NullBlockTracer.Instance, token);
```
LukaszRozmej:
```suggestion
            ProcessingOptions options = _blocksConfig.BuildBlocksOnMainState 
                ? ProcessingOptions.NoValidation | ProcessingOptions.StoreReceipts | ProcessingOptions.DoNotUpdateHead
                : ProcessingOptions.ProducingBlock;
            return Processor.Process(block, options, blockTracer ?? NullBlockTracer.Instance, token);
```

LukaszRozmej: Maybe even better and have `ProcessingOptions options = GetProcessingOptions()`

damian-orzechowski: Done

PR #9389, opened 2025-09-30
Harden Orphaned Block validations

LukaszRozmej:
We have something like `ValidateOrphanedBlock` used in some occasions in `NewPayload` and `EraSync`.
This validations were severely lacking. Updating them to have as much validations on BlockBody (basically all) and Header (the ones that don't need parent). Keep as much of the code reusable to force future changes to think of both cases.

## Changes

- Make parent in Validations non-optional
- Add `IHeaderValidator.ValidateOrphan` method
- Make Both paths (orphan and not) use same code path in `HeaderValidator` and `BlockValidator` with `OnFlag` for orphan to not miss orphan validation logic when adding new validations and in derived classes.
- Some additional refactors, simplifications, unifications, removal of unused stuff

- [x] Refactoring

#### Requires testing

- [x] Yes

- [x] Yes


src/Nethermind/Nethermind.Consensus/Validators/BlockValidator.cs
```diff
@@ -48,14 +48,8 @@ public bool Validate(BlockHeader header, BlockHeader? parent, bool isUncle, out
...
-    public bool ValidateOrphanedBlock(Block block, out string? error)
-    {
-        if (!ValidateEip4844Fields(block, _specProvider.GetSpec(block.Header), out error))
-            return false;
-
-        error = null;
-        return true;
-    }
+    public bool ValidateOrphanedBlock(Block block, out string? error) =>
+        ValidateBlock<OnFlag>(block, null, out error);
```
Marchhill, line 52: Is there a benefit to doing this through a template, rather than an extra bool argument or something?

src/Nethermind/Nethermind.Consensus/Validators/BlockValidator.cs
```diff
@@ -110,26 +113,17 @@ public bool ValidateSuggestedBlock(Block block, BlockHeader? parent, out string?
...
-            if (_logger.IsDebug) _logger.Debug($"{Invalid(block)} Invalid header");
-            return false;
-        }
+        return true;
+    }
 
-        if (validateHashes)
+    private bool ValidateBlockSize(Block block, IReleaseSpec spec, ref string? errorMessage)
+    {
+        int encodedSize = block.EncodedSize ?? _blockDecoder.GetLength(block, RlpBehaviors.None);
```
Marchhill: small optimisation: could calculate the encoded size only if Eip7934 is enabled

src/Nethermind/Nethermind.Consensus/Validators/Always.cs
```diff
@@ -23,73 +23,28 @@ private Always(bool result)
     // ReSharper disable once NotNullMemberIsNotInitialized
     private static Always _valid;
 
-    public static Always Valid
-        => LazyInitializer.EnsureInitialized(ref _valid, static () => new Always(true));
+    public static Always Valid => LazyInitializer.EnsureInitialized(ref _valid, static () => new Always(true));
```
Marchhill, line 26: nit: file & class name could be a bit clearer. Maybe `AlwaysValidator` or `ConstantValidator`

src/Nethermind/Nethermind.Merge.Plugin/MergeHeaderValidator.cs
```diff
@@ -14,43 +14,34 @@
...
-            ISpecProvider specProvider,
-            ISealValidator sealValidator,
-            ILogManager logManager)
-            : base(blockTree, sealValidator, specProvider, logManager)
-        {
-            _poSSwitcher = poSSwitcher;
-            _preMergeHeaderValidator = preMergeHeaderValidator;
-            _blockTree = blockTree;
-        }
+        private readonly IBlockTree _blockTree = blockTree;
```
Marchhill: unused

src/Nethermind/Nethermind.Merge.Plugin/MergeHeaderValidator.cs
```diff
@@ -14,43 +14,34 @@
...
+        IPoSSwitcher poSSwitcher,
+        IHeaderValidator preMergeHeaderValidator,
+        IBlockTree blockTree,
+        ISpecProvider specProvider,
+        ISealValidator sealValidator,
+        ILogManager logManager)
+        : HeaderValidator(blockTree, sealValidator, specProvider, logManager)
     {
         // https://eips.ethereum.org/EIPS/eip-3675#constants
         private const int MaxExtraDataBytes = 32;
```
Marchhill, line 27: while we're changing this file, may as well move this to Eip3675 constants file

src/Nethermind/Nethermind.Optimism/OptimismBlockValidator.cs
```diff
@@ -97,7 +97,7 @@ private bool ValidateWithdrawals(BlockHeader header, BlockBody body, out string?
         }
         else
         {
-            if (header.WithdrawalsRoot != null)
+            if (header.WithdrawalsRoot is not null)
```
Marchhill: nit: use `else if`

src/Nethermind/Nethermind.Taiko/TaikoHeaderValidator.cs
```diff
@@ -46,7 +46,7 @@ protected override bool ValidateTotalDifficulty(BlockHeader parent, BlockHeader
         return true;
     }
 
-    protected override bool ValidateBlobGasFields(BlockHeader header, BlockHeader parentHeader, IReleaseSpec spec, ref string? error)
+    protected override bool ValidateBlobGasFields(BlockHeader header, BlockHeader parent, IReleaseSpec spec, ref string? error)
```
Marchhill: nit: use `=>` syntax in this file

src/Nethermind/Nethermind.Consensus.Clique/CliqueSealValidator.cs
```diff
@@ -25,7 +25,7 @@ public bool ValidateParams(BlockHeader parent, BlockHeader header, bool isUncle
         {
             long number = header.Number;
             // Retrieve the snapshot needed to validate this header and cache it
-            Snapshot snapshot = _snapshotManager.GetOrCreateSnapshot(number - 1, header.ParentHash);
+            Snapshot snapshot = _snapshotManager.GetOrCreateSnapshot(number - 1, header.ParentHash!);
```
Marchhill, line 25: could change to primary constructor

src/Nethermind/Nethermind.Xdc/XdcHeaderValidator.cs
```diff
@@ -20,13 +20,12 @@ namespace Nethermind.Xdc;
 
 public class XdcHeaderValidator(IBlockTree blockTree, ISealValidator sealValidator, ISpecProvider specProvider, ILogManager? logManager = null) : HeaderValidator(blockTree, sealValidator, specProvider, logManager)
 {
-
-    public override bool Validate(BlockHeader header, BlockHeader? parent, bool isUncle, out string? error)
+    protected override bool Validate<TOrphaned>(BlockHeader header, BlockHeader? parent, bool isUncle, out string? error)
```
Marchhill, line 19: unneeded `using` in this file

src/Nethermind/Nethermind.Xdc.Test/XdcHeaderValidatorTests.cs
```diff
@@ -24,76 +24,48 @@ public class Tests
     [Test]
     public void Validate_NotAnXdcHeader_ThrowsArgumentException()
     {
-        var header = Build.A.BlockHeader.TestObject;
+        BlockHeader parent = Build.A.BlockHeader.TestObject;
```
Marchhill, line 24: unneeded `using` in this file

Marchhill (approved):
Looks good, only thing is whether to use this template approach to specify an orphaned blocks. It's fine but seems a bit less clear to me than having another argument and overloading

src/Nethermind/Nethermind.Consensus/Validators/BlockValidator.cs
LukaszRozmej: It is, compiler can resolve that if in compile time and cut the code that resolves to `false`, thus removing unneeded checks for orphan blocks completely.

src/Nethermind/Nethermind.Consensus/Validators/Always.cs
LukaszRozmej: It reads very well in usages `Always.Valid`

src/Nethermind/Nethermind.Merge.Plugin/MergeHeaderValidator.cs
LukaszRozmej: Overkill for one constant with one local use

flcl42 (approved):

tanishqjasoria (approved):

src/Nethermind/Nethermind.Consensus/Validators/HeaderValidator.cs
```diff
@@ -16,28 +16,20 @@
...
-            _blockTree = blockTree ?? throw new ArgumentNullException(nameof(blockTree));
-            _sealValidator = sealValidator ?? throw new ArgumentNullException(nameof(sealValidator));
-            _specProvider = specProvider ?? throw new ArgumentNullException(nameof(specProvider));
-            _daoBlockNumber = specProvider.DaoBlockNumber;
-        }
+        protected readonly ISealValidator _sealValidator = sealValidator ?? throw new ArgumentNullException(nameof(sealValidator));
+        protected readonly ISpecProvider _specProvider = specProvider ?? throw new ArgumentNullException(nameof(specProvider));
+        private readonly long? _daoBlockNumber = specProvider.DaoBlockNumber;
+        protected readonly ILogger _logger = logManager?.GetClassLogger() ?? throw new ArgumentNullException(nameof(logManager));
+        private readonly IBlockTree _blockTree = blockTree ?? throw new ArgumentNullException(nameof(blockTree));
```
ak88, line 32: Since we dont get parent blocks anymore, maybe we can just pass genesis difficulty instead, and remove this dependency

LukaszRozmej: We could, not necessarily in this PR. @asdacap could you help with how to set it up in DI?

ak88 (approved):

src/Nethermind/Nethermind.Optimism/OptimismBlockValidator.cs
```diff
@@ -38,31 +38,31 @@ public class OptimismBlockValidator(
...
 
-        errorMessage = null;
+        error = null;
         return true;
     }
 
-    protected override bool ValidateWithdrawals(Block block, IReleaseSpec spec, out string? error) =>
-        ValidateWithdrawals(block.Header, block.Body, out error);
+    protected override bool ValidateWithdrawals(Block block, IReleaseSpec spec, bool validateHashes, ref string? error) =>
+        !validateHashes || ValidateWithdrawals(block.Header, block.Body, out error);
```
emlautarom1: Not sure why we have this `validateHashes` here: in OP Isthmus, you always need to verify that the withdrawals hash is set: https://specs.optimism.io/protocol/isthmus/exec-engine.html#header-validity-rules

emlautarom1: Note that this "hash" is not actually the `Withdrawals` hash, but rather something different (OP reuses the field).

emlautarom1 (changes requested):
Just one comment on OP stack.

src/Nethermind/Nethermind.Optimism/OptimismBlockValidator.cs
LukaszRozmej: This was introduced in #6952, the logic behind it is that in NP we just calculated those hashes from Payload itself, so we don't need to validate them there. Before this check was outside of this method and method wasn't called at all, now I pushed it inside the method.

LukaszRozmej: fixed

LukaszRozmej:
ok this is enough approvals for me

PR #9396, opened 2025-10-02
Make ExecutePrecompile overridable instead of RunPrecompile

hudem1:
For Arbitrum, we actually want to override `ExecutePrecompile()` because precompiles can both revert and fail (without reverting) impacting any returned output as well as any gas refund (to sender if top-level, or to calling contract if nested call).

See [Arbitrum PR](https://github.com/NethermindEth/nethermind-arbitrum/pull/279)

For example, one case where the base implementation of `ExecutePrecompile()` does not work for Arbitrum: anytime a precompile reverts (not an ungraceful failure). The base implem will either set the result as a failure (not a revert) or will set the gas left as 0 (no refund).

I guess it's normal behaviour for Ethereum as it's been working so far, but i thought reverting ethereum precompiles would refund caller..?

Fixes Closes Resolves #

## Changes

- _List the changes_

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks

### DIFF
--- a/src/Nethermind/Nethermind.Evm/VirtualMachine.cs
+++ b/src/Nethermind/Nethermind.Evm/VirtualMachine.cs
@@ -763,7 +763,7 @@ public unsafe partial class VirtualMachine(
     /// A <see cref="CallResult"/> containing the results of the precompile execution. In case of a failure,
     /// returns the default value of <see cref="CallResult"/>.
     /// </returns>
-    protected CallResult ExecutePrecompile(EvmState currentState, bool isTracingActions, out Exception? failure)
+    protected virtual CallResult ExecutePrecompile(EvmState currentState, bool isTracingActions, out Exception? failure)
     {
         // Report the precompile action if tracing is enabled.
         if (isTracingActions)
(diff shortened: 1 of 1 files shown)


LukaszRozmej (approved):

hudem1:
Hello @marcindsobczak,
As described in the PR description, it seems like according to base NM implementation, reverting ethereum precompiles do not refund, is this normal ?
And if they're top-level, they're usually considered as failed (not reverted) it seems. I guess that's the wanted behaviour ?

marcindsobczak (approved):

hudem1:
> Yeah, that's all correct. Gas doesn't get refunded on a revert because it pays for the computation that already happened. And a top-level call results in a failed transaction, not a revert, because there's no calling contract to revert back to.

yes but i mean, the precompile call will receive some gas but it might not spend everything, so it should return the gas left to the caller. But if the call doesn't revert but fails instead, the gas left is not returned to the caller if nested call, or sender if top-level

PR #9398, opened 2025-10-02
Safer types for a one of case

flcl42:
Fixes a bug revealed by devnet-3. 
Either it's time for some monads?

## Changes

- Fix input type

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Core/Either.cs
+++ b/src/Nethermind/Nethermind.Core/Either.cs
@@ -31,13 +31,13 @@ public readonly struct Either<T1, T2>
     public bool Is(out T1? val)
     {
         val = value1;
-        return value1 != null;
+        return value1 is not null;
     }
 
     public bool Is(out T2? val)
     {
         val = value2;
-        return value2 != null;
+        return value2 is not null;
     }
 
     public void To(out T1 val) => val = value1 ?? throw new ArgumentException(nameof(T1));
--- a/src/Nethermind/Nethermind.Core/Either.cs
+++ b/src/Nethermind/Nethermind.Core/Either.cs
@@ -40,7 +40,7 @@ public readonly struct Either<T1, T2>
         return Value2 != null;
     }
 
-    public void As(out T1 val) => val = Value1 ?? throw new ArgumentException(nameof(T1));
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Core/Either.cs
```diff
@@ -0,0 +1,46 @@
...
+
+namespace Nethermind.Core;
+
+public readonly struct Either<T1, T2>
+{
+    private readonly T1? value1;
+    private readonly T2? value2;
+
+    [Obsolete("Do not use", error: true)]
+    public Either() { }
```
LukaszRozmej, line 14: make it private?

src/Nethermind/Nethermind.Core/Either.cs
```diff
@@ -0,0 +1,46 @@
...
+        value1 = default;
+    }
+
+    public static implicit operator Either<T1, T2>(T1 val) => new(val);
+    public static implicit operator Either<T1, T2>(T2 val) => new(val);
+
+    public bool Is(out T1? val)
+    {
+        val = value1;
+        return value1 != null;
```
LukaszRozmej:
```suggestion
        return value1 is not null;
```

src/Nethermind/Nethermind.Core/Either.cs
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    public bool Is(out T2? val)
+    {
+        val = value2;
+        return value2 != null;
```
LukaszRozmej:
```suggestion
        return value2 is not null;
```

LukaszRozmej (approved):

src/Nethermind/Nethermind.Core/Either.cs
flcl42: struct ctor can't be private. It exists by default, defined here just to make the attribute to raise an error if someone decides to created Either without a value

PR #9404, opened 2025-10-03
Add BlockTimestamp to ReceiptForRpc

LukaszRozmej:

#### Requires testing

- [x] Yes

- [x] Yes


LukaszRozmej:
> Why do we add this? Looks like hive tests expect to not have timestamp https://github.com/NethermindEth/nethermind/actions/runs/18220910861/job/51880440283

@MarekM25 said to add it... @MarekM25 ?

PR #9406, opened 2025-10-03
Add timestamp to simulate logs & blob gas & remove total difficulty

deffrian:
Fixes Closes Resolves #

```
eth_simulateV1/ethSimulate-simple-send-from-contract (nethermind)
eth_simulateV1/ethSimulate-override-address-twice-in-separate-BlockStateCalls (nethermind)
eth_simulateV1/ethSimulate-eth-send-should-produce-logs (nethermind)
eth_simulateV1/ethSimulate-eth-send-should-produce-more-logs-on-forward (nethermind)
eth_simulateV1/ethSimulate-empty-with-block-num-set-firstblock (nethermind)
eth_getBlockByNumber/get-genesis (nethermind)
eth_getBlockByHash/get-block-by-hash (nethermind)
eth_getBlockByNumber/get-block-london-fork (nethermind)
```
There is a bug in geth that sets timstamps to zero in eth simulate
For context: https://github.com/ethereum/go-ethereum/pull/32831

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_


src/Nethermind/Nethermind.Facade/Eth/BlockForRpc.cs
```diff
@@ -120,9 +119,6 @@ public BlockForRpc(Block block, bool includeFullTransactionData, ISpecProvider s
     public Hash256 StateRoot { get; set; }
     [JsonConverter(typeof(NullableRawLongConverter))]
     public long? Step { get; set; }
-
-    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
-    public UInt256? TotalDifficulty { get; set; }
```
LukaszRozmej, line 125: I say leave it for AuRa and Clique
If we really want we can remove (by setting `null`) in past blocks if we are post-merge. But that probably will still fail the test.

But the test is shit as TD is valid property for pre-merge blocks and Geth just yolo removed it.

src/Nethermind/Nethermind.Facade/Simulate/SimulateBridgeHelper.cs
```diff
@@ -253,10 +255,20 @@ private Transaction CreateTransaction(
...
+        if ((ForkActivation)result.Number >= specProvider.MergeBlockNumber)
+        {
+            result.Difficulty = UInt256.Zero;
+            result.IsPostMerge = true;
+        }
+        else
+        {
+            result.Difficulty = parent.Difficulty;
+            result.IsPostMerge = false;
+        }
```
LukaszRozmej, line 270: In theory this is not the best way to check during the transition, we have `IPoSSwitcher` for that. But we could probably ignore that/

deffrian: I tried, but PosSwitcher is not expecting this type of usage

src/Nethermind/Nethermind.Facade/Eth/BlockForRpc.cs
deffrian: I would remove it as well. Rn now we store it in the db for no apparent reason
Do we use clique anywhere? I think at this point we can remove it with ethash. And we probably should

LukaszRozmej: 1. Japan Open Chain still uses Clique and is supported.
2. We still have some Clique and AuRa uses, but they are not officially supported networks.
3. We still need AuRa and Ethash for full syncing networks that were merged later (mainnet, gnosis). We could probably rip out large part from them, but can't remove them completely. But it is an effort.

Thus lets not do sudden moves without proper plan and scope.

src/Nethermind/Nethermind.Facade/Simulate/SimulateBridgeHelper.cs
LukaszRozmej: Yeah probably not worth the effort, this is fine.

PR #9417, opened 2025-10-06
Add recovery of missing code

LukaszRozmej:
Allows to get missing code from network via snap protocol

#### Requires testing

- [x] Yes

- [x] Yes


benaadams (approved):
Let's go!

flcl42 (approved):

damian-orzechowski (approved):

asdacap (approved):

stdevMac:
@LukaszRozmej is this PR still valid? Otherwise, perhaps we can close it? Anyway, we will need to rebase/solve bunch of conflicts

LukaszRozmej:
@stdevMac need to rebase/merge and test properly. Let's keep it

src/Nethermind/Nethermind.Synchronization/Trie/CodeRecovery.cs
```diff
@@ -0,0 +1,88 @@
...
+        if (!peer.TryGetSatelliteProtocol(Protocol.Snap, out ISnapSyncPeer? snapProtocol)) return null;
+
+        using ArrayPoolList<ValueHash256> hashes = new(1);
+        hashes.Add(new ValueHash256(key));
+
+        using IByteArrayList? result = await snapProtocol.GetByteCodes(hashes, token);
+
+        if (_logger.IsTrace) _logger.Trace($"Fetched code {key} from {peer}");
+
+        return result is { Count: 1 } ? result[0].ToArray() : null;
```
marcindsobczak: Would be good to validate result to be safe against malicious peer
```suggestion
        return result is { Count: 1 } && ValueKeccak.Compute(result[0]) == hashes[0] ? result[0].ToArray() : null;
```

marcindsobczak (approved):

benaadams:
@claude review

LukaszRozmej:
@claude review

LukaszRozmej:
@claude re-review

LukaszRozmej:
@claude re-review

src/Nethermind/Nethermind.Synchronization.Test/Trie/HealingTreeTests.cs
```diff
@@ -225,6 +273,13 @@ void AssertStorage(IContainer client)
             {
                 Assert.That(mainWorldState.Get(new StorageCell(storageAddress, (UInt256)i)).ToArray(), Is.EqualTo(i.ToBigEndianByteArray()));
             }
+
+            // The code db was never copied, so this can only be served by recovering the code over snap.
+            // A hash-keyed state db serves no snap, and snap is the only protocol carrying bytecode.
```
LukaszRozmej:
```suggestion
```

src/Nethermind/Nethermind.Synchronization/Trie/CodeRecovery.cs
```diff
@@ -0,0 +1,114 @@
...
+
+    private const int ConcurrentAttempt = 3;
+    private readonly ILogger _logger = logManager.GetClassLogger<CodeRecovery>();
+
+    public async Task<byte[]?> Recover(ValueHash256 codeHash, CancellationToken cancellationToken = default)
+    {
+        using AutoCancelTokenSource cts = cancellationToken.CreateChildTokenSource(RecoveryTimeout);
+        // Read once. A losing attempt still between Allocate and its request would otherwise read
+        // the token after the winner returned and disposed the source, and be blamed for the
+        // ObjectDisposedException that follows.
```
LukaszRozmej:
```suggestion
```

LukaszRozmej:
@claude re-review

LukaszRozmej:
@claude re-review

LukaszRozmej:
@claude re-review

src/Nethermind/Nethermind.Synchronization.Test/Trie/HealingTreeTests.cs
```diff
@@ -79,6 +82,98 @@ HealingStorageTree CreateHealingStorageTree(ITrieStore trieStore, INodeStorage n
...
+
+        HealingCodeDb codeDb = new(db, new Lazy<ICodeRecovery>(recovery));
+
+        // Dedicated threads, not the pool: both readers park on the recovery, and a saturated pool
+        // could otherwise leave the second one queued behind the first.
+        Task<byte[]?> first = ReadOnOwnThread(codeDb);
+        Assert.That(() => Volatile.Read(ref recoveries), Is.EqualTo(1).After(10000, 10));
+
+        Task<byte[]?> second = ReadOnOwnThread(codeDb);
+        Assert.That(() => Volatile.Read(ref reads), Is.EqualTo(2).After(10000, 10));
```
benaadams, line 129: [P2] Synchronize on joining the shared recovery before releasing the gate. Reposting because the prior thread was resolved but this race remains: `reads == 2` only observes the second DB read, before that caller reaches `_inFlight.GetOrAdd`. If it is descheduled there, the first reader can finish and remove the entry before the second joins, producing two recoveries and intermittently failing the assertion. I reproduced this scheduling with the actual `HealingCodeDb` in an isolated harness. Use a deterministic join handshake; a settling delay would still depend on scheduling.

src/Nethermind/Nethermind.Synchronization/Trie/NodeDataRecovery.cs
```diff
@@ -49,50 +49,76 @@ public class NodeDataRecovery(ISyncPeerPool peerPool, INodeStorage nodeStorage,
...
-        return recoveredNodes;
+            return recoveredNodes;
+        }
+        catch (OperationCanceledException)
+        {
+            return null;
+        }
+        catch (Exception ex)
+        {
+            // Never fault the sibling recovery racing this one in PathNodeRecovery.
```
benaadams, line 100: [P3] Dispose `recoveredNodes` in both exception handlers before returning null. The repository GC-reclaimable-array exemption explicitly excludes `ArrayPoolList<T>` and `SafeArrayPool<T>`, so the comment below applies the exemption too broadly. Preserve ownership transfer on the successful return. This loses managed pool reuse on handled failures; it is not a persistent native-memory leak.

src/Nethermind/Nethermind.Synchronization.Test/Trie/RecoveryTests.cs
```diff
@@ -172,14 +201,122 @@ public async Task cannot_recover_eth67_hash_mismatch()
...
+        INodeStorage throwingNodeStorage = Substitute.For<INodeStorage>();
+        throwingNodeStorage.Get(Arg.Any<Hash256?>(), Arg.Any<TreePath>(), Arg.Any<ValueHash256>(), Arg.Any<ReadFlags>())
+            .Returns<byte[]?>(_ => throw new InvalidOperationException("node storage unavailable"));
+
+        PathNodeRecovery recovery = new(
+            new NodeDataRecovery(_syncPeerPool, throwingNodeStorage, LimboLogs.Instance),
+            _snapRecovery,
+            LimboLogs.Instance);
+
+        IOwnedReadOnlyList<(TreePath, byte[])>? response = await Recover(recovery, _peerEth67);
```
benaadams: [P3] Use a `using` declaration for this owned recovery response. The new test keeps an `IOwnedReadOnlyList` without disposing it, contrary to the resource-ownership rule. The four new `TestMemDb` locals in `HealingTreeTests.cs` (lines 88, 114, 158 and 169) also need `using` declarations under that rule.

src/Nethermind/Nethermind.Synchronization.Test/Trie/HealingTreeTests.cs
```diff
@@ -79,6 +82,98 @@ HealingStorageTree CreateHealingStorageTree(ITrieStore trieStore, INodeStorage n
         recovery_works(successfullyRecovered, addressPath, path, fullPath, CreateHealingStorageTree);
     }
 
+    [Test]
+    public void code_recovery_works([Values(true, false)] bool successfullyRecovered)
```
benaadams: [P3] Use bare `[Values]` for this boolean parameter, as required by the test guidelines. The two peer-count parameters in `RecoveryTests.cs` (lines 140 and 155) should likewise use `[Range(1, 2)]` for their contiguous integer cases.

src/Nethermind/Nethermind.Synchronization.Test/Trie/RecoveryTests.cs
```diff
@@ -172,14 +201,122 @@ public async Task cannot_recover_eth67_hash_mismatch()
...
+        Assert.That(response![0].Item1, Is.EqualTo(_path));
```
benaadams: [P3] Wrap these two independent assertions on the recovered response in `using (Assert.EnterMultipleScope())` so a single run reports both mismatches. Apply the same test-guideline rule to the final independent result/count assertions in `HealingTreeTests.cs`, lines 132–134, after the completion assertion.

src/Nethermind/Nethermind.Synchronization.Test/Trie/RecoveryTests.cs
```diff
@@ -92,6 +97,7 @@ ISyncPeer MakeEth67Peer()
         _syncPeerPool = Substitute.For<ISyncPeerPool>();
         _snapRecovery = new SnapRangeRecovery(_syncPeerPool, LimboLogs.Instance);
         _nodeDataDataRecovery = new NodeDataRecovery(_syncPeerPool, new NodeStorage(new MemDb()), LimboLogs.Instance);
+        _codeRecovery = new CodeRecovery(_syncPeerPool, LimboLogs.Instance);
```
benaadams: [P3] Resolve `ICodeRecovery` through the production module with a peer-pool override instead of manually constructing `CodeRecovery`. This PR registers it in `PruningTrieStoreModule`, and the repository DI/test rules require using production registrations for components already wired there.

src/Nethermind/Nethermind.Synchronization/Trie/CodeRecovery.cs
```diff
@@ -0,0 +1,111 @@
...
+            Protocol.Snap);
+
+    // The caller blocks a processing thread on this, and peer allocation waits indefinitely on its own,
+    // so the wait must be bounded here. Matches PathNodeRecovery.
+    private static readonly TimeSpan RecoveryTimeout = TimeSpan.FromSeconds(3);
+
+    private const int ConcurrentAttempt = 3;
+    private readonly ILogger _logger = logManager.GetClassLogger<CodeRecovery>();
+
+    public async Task<byte[]?> Recover(ValueHash256 codeHash, CancellationToken cancellationToken = default)
```
benaadams, line 45: [P3] Add `<inheritdoc/>` to this public interface implementation and a summary to the new public `CodeRecovery` class. The new public `HealingCodeDb.Set` and `StartWriteBatch` implementations also need `<inheritdoc/>` under the public-API documentation rule.

src/Nethermind/Nethermind.Synchronization.Test/Trie/RecoveryTests.cs
LukaszRozmej: Leaving this as-is, with reasoning.

`CodeRecovery` is a two-dependency leaf (`ISyncPeerPool`, `ILogManager`), and a substituted peer pool is the whole point of this fixture. Resolving it through `PruningTrieStoreModule` would drag in the pruning/db/config graph to obtain a two-arg object, and the fixture already constructs its three siblings — `SnapRangeRecovery`, `NodeDataRecovery`, `PathNodeRecovery` — the same way, so this would mean restructuring the class rather than changing one line.

As above, the DI rule is aimed at infrastructure graphs, not leaf services.

The production wiring is covered elsewhere: `EthereumRunnerTests` builds the full container for every shipped config, and `HealingTreeTests.HealingTreeTest` exercises code healing end-to-end through a real `TestNethermindModule` container. I did check the registration uses the preferred DSL form — `.AddSingleton<ICodeRecovery, CodeRecovery>()`, not the `ctx => new ...` shape the neighbouring `IPathRecovery` uses.

src/Nethermind/Nethermind.Synchronization.Test/Trie/HealingTreeTests.cs
LukaszRozmej: Applied the `[Values]` half — the boolean is now bare `[Values]`.

Not taking the `[Range(1, 2)]` half for the peer counts. There's no repo rule preferring `[Range]`, and usage runs the other way: 536 `[Values(...)]` and 286 bare `[Values]` against 19 `[Range(`. `[Range]` reads well for a contiguous span; for two discrete peer counts `[Values(1, 2)]` says what's meant more directly.

src/Nethermind/Nethermind.Synchronization/Trie/CodeRecovery.cs
LukaszRozmej: Applied in part.

Added the `CodeRecovery` class summary/remarks and `<inheritdoc/>` on `Recover` — `ICodeRecovery` is fully documented, so that inherits real content.

Skipped the `HealingCodeDb` half: `IWriteOnlyKeyValueStore.Set` (`IKeyValueStore.cs:104`) and `IKeyValueStoreWithBatching.StartWriteBatch` (`IKeyValueStoreWithBatching.cs:8`) carry no XML docs at all, so `<inheritdoc/>` there would inherit nothing and just add two noise comments to one-line delegating members.

LukaszRozmej:
@claude re-review

LukaszRozmej:
Worked all five. One fixed here, two already fixed in an open PR, two left as awareness with reasoning.

### Fixed here — the narrowed read surface (`fe8f01b2`)

Sharper than the finding states, and worth recording: those members were not merely slow, they were **not on the class at all**. `HealingCodeDb` declared only `Get`/`Set`/`StartWriteBatch`, and default interface members are not inherited by the implementing type — so `codeDb.KeyExists(key)` did not compile against the concrete type. I found this by deleting the new override to mutation-test it and getting CS1503 instead of a failing assertion. They were reachable only through an `IReadOnlyKeyValueStore` reference, which is where the allocating `byte[] Get` fallback took over.

`GetSpan`, `Get(key, output)`, `GetOwnedMemory` and `KeyExists` now each try the wrapped store first and heal only on a miss, so recovery still reaches every member rather than being lost to a plain forward.

The one part that needed care is span ownership. `DangerousReleaseMemory` forwards to the wrapped store, and `DbOnTheRocks` decrements `_allocatedSpan` and calls `_db.DangerousReleaseMemory(span)` — handing it a managed array would corrupt that accounting and free memory RocksDB does not own. So `GetSpan` persists the recovered bytes and re-reads through `codeDb.GetSpan`, which makes every span it hands out owned by the wrapped store by construction. The two `return` sites are the only sources, both `codeDb.GetSpan`. Documented on the member, including that a store which does not serve its own write back reads as a miss rather than healing.

Two tests, each mutation-verified:
- `code_recovery_reaches_every_read_member([Values] CodeRead read)` — all six members (including the indexer) heal on a miss and are served locally afterwards, with exactly one recovery. Forwarding any of the three to the wrapped store without the fallback fails 3 of the 6 cases.
- `code_key_exists_that_hits_skips_the_allocating_get` — restoring the old `Get(key) is not null` shape flips the recorded-read count from 0 to 1.

`KeyExists` heals rather than answering locally, so existence keeps agreeing with `Get`. That matches the previous behaviour (the default chained `KeyExists → GetSpan → Get`), so this change is allocation-only, not semantic.

### Already fixed, in #13266 rather than here

Both rental findings are closed on `fix/recovery-pooled-list-leaks`, which is green with no open threads:

- **`SnapRangeRecovery.AssembleResponse`** — `result.Dispose()` before the `result.Count == 0` return, exactly the one-liner you describe. I verified it bites: with the `Dispose` removed, `cannot_recover_eth67_unassemblable_proofs` fails.
- **`Wait.AnyWhere` dropping a losing racer's result** — no longer awareness-only there. The loop is wrapped in `try`/`finally` with `DiscardRemaining(taskSet)`, and the resolved task stays in the set until it is forwarded, so a faulted or cancelled `await`, a throwing `cond`, and the abandoned-after-return case all dispose what they drop. `DiscardRemaining` delegates to the same runtime `is IDisposable` check as the rejected path, so the two cannot disagree.

Doing them here would duplicate the same lines on a second branch, so I would rather they land once. Happy to cherry-pick if #13266 stalls.

### Left as awareness

- **No memory of a failed recovery.** Agreed and not taken: the first failure throws `Code … is missing from the database` and aborts the block, so the repeat cost needs a caller that survives the throw. Matches `HealingStateTree`, and a negative cache wants its own eviction policy rather than a dictionary bolted on here.
- **Unbounded `Warn`.** Consistent with `PathNodeRecovery`; rate-limiting one of the two and not the other would be the worse outcome. Belongs in a change that covers both.

src/Nethermind/Nethermind.Synchronization.Test/Trie/RecoveryTests.cs
benaadams: **Low — production DI in test setup.** Reposting because the earlier thread was resolved. Resolve `ICodeRecovery` through production modules with a substituted `ISyncPeerPool` instead of constructing `CodeRecovery` here; this PR registers it in `PruningTrieStoreModule`. I read the earlier leaf-service rationale, but the current `di-patterns.md` and `test-infrastructure.md` explicitly apply to components already registered in production, without a leaf-service exception. This is a rule mismatch, not a functional defect.

src/Nethermind/Nethermind.Synchronization.Test/Trie/HealingTreeTests.cs
```diff
@@ -79,6 +83,181 @@ HealingStorageTree CreateHealingStorageTree(ITrieStore trieStore, INodeStorage n
...
+            Assert.That(ReadThrough(codeDb, db, read), Is.EqualTo(_rlp), "served locally afterwards");
+        }
+
+        recovery.Received(1).Recover(_key.ValueHash256, Arg.Any<CancellationToken>());
+    }
+
+    [Test]
+    public void code_key_exists_that_hits_skips_the_allocating_get()
+    {
+        using TestMemDb db = new() { [_key.Bytes] = _rlp };
```
benaadams, line 184: **Low — repeated test setup.** Extract the shared database/recovery/wrapper setup into a helper: this arrangement repeats in `code_recovery_skips_present_code` and `code_recovery_skips_keys_that_cannot_be_a_code_hash`, while `code_recovery_works` and `code_recovery_reaches_every_read_member` also repeat successful-recovery configuration. Keep the database's `using` ownership and each scenario's distinct behavior visible in the tests, as required by `test-infrastructure.md` for shared arrange steps.

src/Nethermind/Nethermind.Synchronization.Test/Trie/RecoveryTests.cs
```diff
@@ -131,25 +136,47 @@ public async Task cannot_recover_eth66_invalid_rlp()
     }
 
     [Test]
-    public async Task can_recover_eth67()
+    public async Task can_recover_eth67([Values(1, 2)] int peerCount)
```
benaadams: **Low — contiguous integer test cases.** Reposting the remaining part of the resolved attribute thread: use `[Range(1, 2)]` here and on `can_recover_code_eth67` at line 153. The earlier reply says there is no such rule, but the current `test-infrastructure.md` explicitly says "a contiguous integer run takes `[Range(from, to)]`". This preserves the same two cases.

src/Nethermind/Nethermind.State/Healing/HealingCodeDb.cs
```diff
@@ -0,0 +1,113 @@
...
+        }
+        finally
+        {
+            // Match on the entry, not just the hash: a caller that resumes late would otherwise evict a
+            // newer recovery that other callers are still sharing.
+            _inFlight.TryRemove(new KeyValuePair<ValueHash256, Lazy<Task<byte[]?>>>(codeHash, attempt));
+        }
+    }
+
+    public void Set(ReadOnlySpan<byte> key, byte[]? value, WriteFlags flags = WriteFlags.None) =>
```
benaadams, line 111: **Low — public API documentation.** Reposting the remaining part of the resolved documentation thread: `Set` and `StartWriteBatch` are still undocumented public implementations under `coding-style.md`. The earlier reply is correct that their interface declarations also lack XML docs, so `<inheritdoc/>` alone would inherit nothing. Add meaningful contract documentation to the interfaces and inherit it, or document these members directly.

benaadams (commented):
Re-review at fe8f01b294: four Low rule findings below; no production blockers found. The earlier correctness findings are addressed. Focused Release validation: 41 healing/recovery tests passed, 0 failed, 0 skipped. Fresh threads replace resolved discussions where the rule mismatch remains.

src/Nethermind/Nethermind.State/Healing/HealingCodeDb.cs
LukaszRozmej: Applied at ebab0ef.

`Set` and `StartWriteBatch` now carry docs of their own rather than an `<inheritdoc/>` that would inherit nothing, and they say the thing a reader of this class actually needs to know: healing is a read-side concern, so neither the direct write nor the batch handed back is intercepted. That is the non-obvious half, given every read member above them recovers.

Not documenting `IWriteOnlyKeyValueStore.Set` / `IKeyValueStoreWithBatching.StartWriteBatch` themselves — writing the contract for two interfaces the whole db layer implements is a much wider change than this PR, and it would land on unrelated code.

src/Nethermind/Nethermind.Synchronization.Test/Trie/HealingTreeTests.cs
LukaszRozmej: Applied at ebab0ef.

Added `HealingCodeDbOver(db)`, which returns the wrapper and the substituted recovery as a tuple. Each of the six code tests now reads `using TestMemDb db = new() { ... };` followed by `(HealingCodeDb codeDb, ICodeRecovery recovery) = HealingCodeDbOver(db);`, so the db's `using` ownership and the db's own arrangement stay in the test, and so does every stub — `code_recovery_works` still sets its own return, `code_recovery_collapses_concurrent_misses` still installs its gate, and the two `skips` tests still leave the recovery unstubbed on purpose. Only the `Substitute.For` + `new HealingCodeDb(db, new Lazy<...>(...))` pair moved.

43 healing/recovery tests pass in Release.

src/Nethermind/Nethermind.Synchronization.Test/Trie/RecoveryTests.cs
LukaszRozmej: Still not taking this, and the citation does not hold up.

`[Range(` appears nowhere in the repository's rules: `grep -rn "Range(" .agents/rules/ AGENTS.md CONTRIBUTING.md` returns nothing, and so does `grep -rn "contiguous" .agents/ AGENTS.md`. `test-infrastructure.md` "Test guidelines" covers `[TestCase]` / `[TestCaseSource]` for parameterisation and is silent on `[Range]` versus `[Values]`. The quoted sentence is not in the file.

Usage points the same way. Across `src/`: 169 `[Values(...)]`, 44 bare `[Values]`, and 3 `[Range(` — and all three `[Range]` sites are long runs (`[Range(0, 31)]`, `[Range(0, 30)]`, `[Range(0, 2 * DepositEventFieldCount - 1)]`), never a pair. `[Values(1, 2)]` states the two peer counts; `[Range(1, 2)]` states an interval that happens to contain them.

src/Nethermind/Nethermind.Synchronization.Test/Trie/RecoveryTests.cs
LukaszRozmej: Still leaving this as-is; the repost does not change the picture, and one detail is worth correcting.

`PruningTrieStoreModule` does not register the sibling recoveries at all — `IPathRecovery` is a single registration whose factory news up `NodeDataRecovery` and `SnapRangeRecovery` inline:

```csharp
.AddSingleton<IPathRecovery, ISyncPeerPool, INodeStorage, ILogManager>((peerPool, nodeStorage, logManager) => new PathNodeRecovery(
    new NodeDataRecovery(peerPool!, nodeStorage, logManager),
    new SnapRangeRecovery(peerPool!, logManager),
    ...
```

So "already registered in production" applies to exactly one of the four types this fixture builds, and there is no registration to resolve the other three through. Routing `ICodeRecovery` alone through a container would make the fixture inconsistent with itself while dragging the pruning/db/config graph in to obtain a two-dependency object (`ISyncPeerPool`, `ILogManager`) whose whole point here is a substituted peer pool.

The rule that snippet illustrates is about infrastructure graphs — `WorldState` over `TrieStore` over `MemDb`, `BlockProcessor` over `TransactionProcessor` — where manual construction encodes wiring that production owns. There is no wiring to get wrong here.

Production wiring stays covered elsewhere: `EthereumRunnerTests` builds the full container for every shipped config, and `HealingTreeTests.HealingTreeTest` exercises code healing end-to-end through a real `TestNethermindModule` container. The registration itself uses the preferred DSL form, `.AddSingleton<ICodeRecovery, CodeRecovery>()`.

LukaszRozmej: Taken in `a06ef31d` — you were right that the rules have no leaf-service exception, and my earlier rationale was arguing against the rule rather than from it.

`ICodeRecovery` now resolves from `TestNethermindModule` with the substituted `ISyncPeerPool` overridden in, following the recipe `HealingTreeTest` already uses in the sibling fixture (`IFlatDbConfig.Enabled = false` and `PruningMode.Full` so the patricia store's wiring is the active one). That turns out to be worth more than compliance: the hand-built version never checked that `PruningTrieStoreModule`'s registration actually composes, and now the code-recovery tests exercise it.

Its two neighbours stay hand-built, and I added a comment saying why rather than leaving it looking like an oversight: production hand-builds `NodeDataRecovery` and `SnapRangeRecovery` inside the `IPathRecovery` factory lambda (`PruningTrieStoreModule.cs:109-113`), so there is no registration to resolve either of them from. `ICodeRecovery` is genuinely the only one of the three the rule reaches.

Cost is about four seconds across the fixture's 23 tests, and the assembly's 2,216 still pass.

src/Nethermind/Nethermind.Synchronization.Test/Trie/RecoveryTests.cs
LukaszRozmej: Taken in `a06ef31d`. You are right and my earlier reply was wrong — `test-infrastructure.md:56` does say "a contiguous integer run takes `[Range(from, to)]`". Both peer-count parameters now use `[Range(1, 2)]`, same two cases.

src/Nethermind/Nethermind.Synchronization.Test/Trie/HealingTreeTests.cs
LukaszRozmej: Done in `ebab0ef3` — `HealingCodeDbOver(db)` now folds the shared database/recovery/wrapper arrangement, with each test keeping its own `using` on the db and its own stubbing so the scenario stays visible.

src/Nethermind/Nethermind.State/Healing/HealingCodeDb.cs
LukaszRozmej: Done in `ebab0ef3` — documented directly rather than via `<inheritdoc/>`, since as you note the interface declarations carry nothing to inherit. `Set` and `StartWriteBatch` now state that healing is a read-side concern and that writes, including those through the returned batch, are not intercepted. That is the non-obvious half given the class documents its read surface at length.

benaadams:
@claude review

benaadams (approved):

PR #9423, opened 2025-10-07
Use proper blob proofs version

flcl42:
## Changes

- Use proper blob proofs version

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Flashbots/Handlers/ValidateBuilderSubmissionHandler.cs
+++ b/src/Nethermind/Nethermind.Flashbots/Handlers/ValidateBuilderSubmissionHandler.cs
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: 2024 Demerzel Solutions Limited
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
 // SPDX-License-Identifier: LGPL-3.0-only
 
 using System;
@@ -93,7 +93,7 @@ public class ValidateSubmissionHandler
             return FlashbotsResult.Invalid(error ?? "Block validation failed");
         }
 
-        if (!ValidateBlobsBundle(block.Transactions, blobsBundle, out string? blobsError))
+        if (!ValidateBlobsBundle(block.Transactions, blobsBundle, _specProvider.GetSpec(block.Header), out string? blobsError))
         {
             if (_logger.IsWarn) _logger.Warn($"Invalid blobs bundle. Result of {payloadStr}. Error: {blobsError}");
             return FlashbotsResult.Invalid(blobsError ?? "Blobs bundle validation failed");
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Flashbots/Handlers/ValidateBuilderSubmissionHandler.cs
```diff
@@ -174,7 +174,7 @@ private bool ValidateBlobsBundle(Transaction[] transactions, BlobsBundleV1 blobs
             return false;
         }
 
-        if (!IBlobProofsManager.For(ProofVersion.V1).ValidateProofs(new ShardBlobNetworkWrapper(blobsBundle.Blobs, blobsBundle.Commitments, blobsBundle.Proofs, ProofVersion.V1)))
+        if (!IBlobProofsManager.For(releaseSpec.BlobProofVersion).ValidateProofs(new ShardBlobNetworkWrapper(blobsBundle.Blobs, blobsBundle.Commitments, blobsBundle.Proofs, ProofVersion.V1)))
```
marcindsobczak: Shouldn't it also be `releaseSpec.BlobProofVersion` instead of `ProofVersion.V1` as an argument for `new ShardBlobNetworkWrapper`?

In current version it will be always invalid before Fusaka, as far as I understand

flcl42: Indeed, nice catch!

src/Nethermind/Nethermind.Flashbots/Handlers/ValidateBuilderSubmissionHandler.cs
```diff
@@ -93,7 +94,7 @@ public Task<ResultWrapper<FlashbotsResult>> ValidateSubmission(BuilderBlockValid
             return FlashbotsResult.Invalid(error ?? "Block validation failed");
         }
 
-        if (!ValidateBlobsBundle(block.Transactions, blobsBundle, out string? blobsError))
+        if (!ValidateBlobsBundle(block.Transactions, blobsBundle, _specProvider.GetSpec(block.Header), out string? blobsError))
```
LukaszRozmej: overall in this class `_specProvider.GetSpec` is now called 3x per request, let's make it 1

src/Nethermind/Nethermind.Flashbots/Handlers/ValidateBuilderSubmissionHandler.cs
```diff
@@ -143,47 +144,49 @@ private bool ValidateBlock(Block block, BidTrace message, long registeredGasLimi
...
+                    }
+                }
             }
         }
 
-        if (totalBlobsLength != blobsBundle.Blobs.Length)
+        if (!verifier.ValidateHashes(blobs, hashes.AsSpan()))
         {
-            error = $"Total blobs length mismatch. Expected {totalBlobsLength} but got {blobsBundle.Blobs.Length}";
+            error = "Invalid blob lengths";
```
LukaszRozmej:
```suggestion
            error = "Invalid blob hashes";
```

src/Nethermind/Nethermind.Flashbots/Handlers/ValidateBuilderSubmissionHandler.cs
flcl42: 👍

PR #9424, opened 2025-10-07
Add interval for history pruning

Marchhill:
## Changes

- Prune every x epochs, configurable by the yser

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

Explain new config option in docs.

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.History/HistoryPruner.cs
+++ b/src/Nethermind/Nethermind.History/HistoryPruner.cs
@@ -45,6 +45,7 @@ public class HistoryPruner : IHistoryPruner
     private readonly IHistoryConfig _historyConfig;
     private readonly bool _enabled;
     private readonly long _epochLength;
+    private readonly long _pruningInterval;
     private readonly long _minHistoryRetentionEpochs;
     private readonly int _deletionProgressLoggingInterval;
     private readonly long _ancientBarrier;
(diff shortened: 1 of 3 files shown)


src/Nethermind/Nethermind.History/HistoryPruner.cs
```diff
@@ -336,7 +338,7 @@ private bool ShouldPruneHistory(out ulong? cutoffTimestamp)
     {
         cutoffTimestamp = null;
 
-        if (!_enabled)
+        if (!_enabled || (_pruningInterval != 0 && _blockTree.Head!.Number % _pruningInterval != 0))
```
LukaszRozmej: Can we put it into some named condition? 
Maybe you have better name?
```suggestion
        if (!_enabled || IsNotPruningIntervalSlotBoundry())
```

src/Nethermind/Nethermind.History/HistoryPruner.cs
```diff
@@ -84,6 +85,7 @@ public HistoryPruner(
         _historyConfig = historyConfig;
         _enabled = historyConfig.Enabled();
         _epochLength = (long)blocksConfig.SecondsPerSlot * 32; // must be changed if slot length changes
+        _pruningInterval = historyConfig.PruningInterval * 32;
```
LukaszRozmej: 32 should be constant?

src/Nethermind/Nethermind.History/HistoryPruner.cs
```diff
@@ -29,6 +29,7 @@ public class HistoryPruner : IHistoryPruner
 {
     private const int MaxOptimisticSearchAttempts = 3;
     private const int LockWaitTimeoutMs = 100;
+    private const int SlotsPerEpoch = 32;
```
flcl42, line 32: Might make more sense to have it configurable as it may differ across the chains. `IMergeConfig.SlotsPerEpoch` or so. With some mentions what it may affect.

PR #9428, opened 2025-10-08
Adjusting execution reverted code. Fixes #9329

### LINKED ISSUE #9329: Adjusting error codes (start with execution reverted) of eth_calls to geth
At least the following three tests are not passing in **rpc-combat** because our error code differs from Geth.  
Geth returns **`3`**, while we return **`-32000`**:

- `eth_call/call-revert-abi-error`
- `eth_call/call-revert-abi-panic`
- `eth_estimateGas/estimate-failed-call`

Fixing these should improve how Nethermind works with tooling that is built for Geth.  
Ideally, we should try to align all `EvmExceptionType` values. The current definitions can be found here:  
👉 https://github.com/NethermindEth/nethermind/blob/master/src/Nethermind/Nethermind.Evm/EvmExceptionExtensions.cs#L8

As a low-hanging fruit, we could start by aligning **`execution reverted`**.  
We can also add `"message": "execution reverted"` to our JSON response.

```
{
  "error": {
    "code": -- -32000 ++ 3,
    ++ "data": "0x77726f6e672d63616c6c6461746173697a65"
  },
  "id": 1,
  "jsonrpc": "2.0"
}
```

Hive links:
https://hive.ethpandaops.io/ 
https://hive.ethpandaops.io/#/test/generic/1758336306-af5b1ba61693a3d0ebb56416e3491b5e?testnumber=25

mrzeszutko:
Fixes #9329

## Changes

Adjusted jsonRPC responses to match the responses expected in the following hive tests:
- eth_call/call-revert-abi-error
- eth_call/call-revert-abi-panic
- eth_estimateGas/estimate-failed-call

Added a new ErrorCode ExecutionReverted = 3
Added `data` property for errors returned by eth_estimateGas - had to introduce additional version of ResultWrapper that allows to specify a different data type for error responses.
Passed the error from EVM when creating a TransactionResult - hive tests expect the error message to be added as `data` in the response

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

I adjusted the existing tests and run hive tests locally to check if the failing test cases pass.

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

For eth_call and eth_estimateGas the api returns error code = 3 when execution is reverted.


src/Nethermind/Nethermind.Evm/ByteCodeBuilder.cs
```diff
@@ -243,6 +244,50 @@ public Prepare Revert(int size, int position)
             return this;
         }
 
+        public Prepare RevertWithSolidityErrorEncoding(string errorMessage)
```
deffrian: Can we add offset and length as params here?

MarekM25: this should be in the test

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.TransactionExecutor.cs
```diff
@@ -109,12 +110,23 @@ public ResultWrapper<TResult> ExecuteTx(TransactionForRpc transactionCall, Block
...
+                if (inputError || errorMessage is not null)
+                {
+                    if (executionReverted)
+                    {
+                        if (bodyData is not null)
+                        {
+                            return ResultWrapper<TResult>.Fail("execution reverted: " + errorMessage, ErrorCodes.ExecutionReverted, bodyData);
+                        }
+
+                        var errorData = errorMessage is not null ? Encoding.UTF8.GetBytes(errorMessage).ToHexString(true) : null;
```
deffrian, line 124: So Revert returns byte array. And we do this converting into string and back. Maybe we should pass this data through `CallOutput.OutputData`. I think it's already there

deffrian (commented):
Also we need to adjust `scripts/known-failing-hive-tests.txt`

src/Nethermind/Nethermind.Evm/ByteCodeBuilder.cs
LukaszRozmej: I am not sure to be honest. We could make it extension method and put it in test project if really needed.

src/Nethermind/Nethermind.JsonRpc/ResultWrapper.cs
```diff
@@ -65,4 +65,34 @@ public void Dispose()
             }
         }
     }
+
+    public class ResultWrapper<T, TErrorData> : ResultWrapper<T>, IResultWrapper, IDisposable
```
LukaszRozmej, line 69: Is this needed? It looks like we could achieve everything with normal ResultWrapper?

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.TransactionExecutor.cs
LukaszRozmej: @benaadams WDYT?

src/Nethermind/Nethermind.Evm/ByteCodeBuilder.cs
mrzeszutko: I moved this to be an extension method inside the tests project

mrzeszutko: I adjusted scripts/known-failing-hive-tests.txt

I moved this method to the tests project. I don't really want to add parameters that will be unused right now - let's add those as soon as we need them in other test cases

src/Nethermind/Nethermind.JsonRpc/ResultWrapper.cs
mrzeszutko: I moved `RevertWithSolidityErrorEncoding` to be an extension method in the test project.

We need this additional `ResultWrapper` because `eth_estimateGas` returns `ResultWrapper<UInt256?>` in a normal successful case but when the execution was reverted we expect the data field to contain a string error message. If you have another idea how to achieve this I'm happy to change.

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.TransactionExecutor.cs
mrzeszutko: We don't have access to this error string as a byte array to pass it via CallResult - when calling BlockchainBridge.estimateGas the error message is available as a string at estimateGasTracer.Error, on the other hand estimateGasTracer.ReturnValue contains a byte array but the content of it is different - it contains the whole Error(string) ABI encoded result, not just the error message that we need in this case (we use it for errors when eth_call is executed but the expected error content differs for eth_call and eth_estimateGas)

benaadams: Maybe could be plumbed though differently so we have the bytes (if coming from a common message); however seems fine for now

PR #9434, opened 2025-10-09
Feature/sorted key value store

asdacap:
- Expose rocksdb iterator through `ISortedKeyValueStore`
- Fix block cache metric not working.

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Tested on another branch.

### DIFF
--- a/src/Nethermind/Nethermind.Db.Rocks/ColumnDb.cs
+++ b/src/Nethermind/Nethermind.Db.Rocks/ColumnDb.cs
@@ -176,11 +176,6 @@ public class ColumnDb : IDb, ISortedKeyValueStore
 
     public ISortedView GetViewBetween(ReadOnlySpan<byte> firstKey, ReadOnlySpan<byte> lastKey)
     {
-        ReadOptions readOptions = new ReadOptions();
-        readOptions.SetIterateLowerBound(firstKey.ToArray());
-        readOptions.SetIterateUpperBound(lastKey.ToArray());
-
-        Iterator iterator = _mainDb.CreateIterator(readOptions, ch: _columnFamily);
-        return new RocksdbSortedView(iterator);
+        return _mainDb.GetViewBetween(firstKey, lastKey, _columnFamily);
     }
 }
--- a/src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
+++ b/src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
@@ -1832,11 +1832,25 @@ public partial class DbOnTheRocks : IDb, ITunableDb, IReadOnlyNativeKeyValueStor
 
     public ISortedView GetViewBetween(ReadOnlySpan<byte> firstKey, ReadOnlySpan<byte> lastKey)
     {
-        ReadOptions readOptions = new ReadOptions();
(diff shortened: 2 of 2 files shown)


src/Nethermind/Nethermind.Core/IKeyValueStore.cs
```diff
@@ -69,6 +69,24 @@ public interface IWriteOnlyKeyValueStore
         void Remove(ReadOnlySpan<byte> key) => Set(key, null);
     }
 
+    public interface ISortedKeyValueStore : IKeyValueStore
+    {
+        byte[]? FirstKey { get; }
+        byte[]? LastKey { get; }
```
LukaszRozmej, line 75: ReadOnlySpan?

src/Nethermind/Nethermind.Db.Rocks/ColumnDb.cs
```diff
@@ -151,4 +151,36 @@ public void DangerousReleaseMemory(in ReadOnlySpan<byte> span)
...
+            iterator.SeekToLast();
+            return iterator.Valid() ? iterator.GetKeySpan().ToArray() : null;
+        }
+    }
+
+    public ISortedView GetViewBetween(ReadOnlySpan<byte> firstKey, ReadOnlySpan<byte> lastKey)
+    {
+        ReadOptions readOptions = new ReadOptions();
+        readOptions.SetIterateLowerBound(firstKey.ToArray());
+        readOptions.SetIterateUpperBound(lastKey.ToArray());
```
LukaszRozmej: ToArray cannot be avoided?

src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
```diff
@@ -1805,4 +1809,34 @@ public void Dispose()
...
+    }
+
+    public ISortedView GetViewBetween(ReadOnlySpan<byte> firstKey, ReadOnlySpan<byte> lastKey)
+    {
+        ReadOptions readOptions = new ReadOptions();
+        readOptions.SetIterateLowerBound(firstKey.ToArray());
+        readOptions.SetIterateUpperBound(lastKey.ToArray());
+
+        Iterator iterator = CreateIterator(readOptions);
+        return new RocksdbSortedView(iterator);
```
LukaszRozmej, line 1854: Can we avoid ToArrays?

src/Nethermind/Nethermind.Core/IKeyValueStore.cs
asdacap: I think its fine. The span need to live somewhere so making it work with span here is hard to get right.

src/Nethermind/Nethermind.Db.Rocks/ColumnDb.cs
asdacap: Changed to unsafe code

src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
asdacap: Removed

PR #9400, opened 2025-10-02
Refactor/Slight cleanup to world state write batch API.

asdacap:
Fixes Closes Resolves #

- The write batch api have this nasty `Action<Address, Account> onAccountUpdated` callback due to the fact that the `Account` have storage root so it need to be updated. 
- This is very awkward to decorate. 
- This PR remove that by adding back `IStorageTree.RootHash` and moving the storage root related operation to `PersistentStorageProvider`, eliminating the need to update the `Account` on `StateProvider`.

#### What types of changes does your code introduce?

- [X] Refactoring

#### Requires testing

- [X] Yes
- [ ] No

#### If yes, did you write tests?

- [ ] Yes
- [X] No

#### Notes on testing

- [X] Mainnet start and runs normally.

PR #9436, opened 2025-10-09
Feature/add back `OnAccountUpdated`

asdacap:
- #9400 was a mistake. It is slower and make it even harder to keep track of the account. This will likely remain so as long as StorageRoot is a part of `Account`.
- Add `HintGet` to allow making commit faster in case of higher level acceleration layer causes commit to get slower as the trie node was not loaded. 

- [X] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

### Notes on testing

- [X] Mainnet runs normally.

### DIFF
--- a/src/Nethermind/Nethermind.Evm/State/IWorldStateScopeProvider.cs
+++ b/src/Nethermind/Nethermind.Evm/State/IWorldStateScopeProvider.cs
@@ -100,7 +100,17 @@ public interface IWorldStateScopeProvider
         IStorageWriteBatch CreateStorageWriteBatch(Address key, int estimatedEntries);
     }
 
-    public record AccountUpdated(Address Address, Account? Account) : EventArgs;
+    public class AccountUpdated(Address Address, Account? Account) : EventArgs
+    {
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Evm/State/IWorldStateScopeProvider.cs
```diff
@@ -84,10 +93,15 @@ public interface IStorageTree
...
     public interface IWorldStateWriteBatch : IDisposable
     {
+        public event EventHandler<AccountUpdated> OnAccountUpdated;
+
         void Set(Address key, Account? account);
+
         IStorageWriteBatch CreateStorageWriteBatch(Address key, int estimatedEntries);
     }
 
+    public record AccountUpdated(Address Address, Account? Account);
```
LukaszRozmej:
```suggestion
    public record AccountUpdated(Address Address, Account? Account) : EventArgs;
```

PR #9437, opened 2025-10-10
Fix logs in eth_simulate

deffrian:
Fixes Closes Resolves #
```
eth_simulateV1/ethSimulate-logs (nethermind)
eth_simulateV1/ethSimulate-self-destructive-contract-produces-logs
```

## Changes

- We should not change logs when traceActions is set to true. Just add them into the list

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

### DIFF
--- a/src/Nethermind/Nethermind.Facade/Simulate/SimulateTxTracer.cs
+++ b/src/Nethermind/Nethermind.Facade/Simulate/SimulateTxTracer.cs
@@ -18,8 +18,10 @@ namespace Nethermind.Facade.Simulate;
 
 public sealed class SimulateTxTracer : TxTracer
 {
-    private static readonly Hash256 transferSignature =
+    private static readonly Hash256 TransferSignature =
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Facade/Simulate/SimulateTxTracer.cs
```diff
@@ -40,34 +39,49 @@ public SimulateTxMutatorTracer(bool isTracingTransfers, Transaction tx, ulong cu
...
         }
     }
 
+    public override void ReportSelfDestruct(Address address, UInt256 balance, Address refundAddress)
+    {
+        base.ReportSelfDestruct(address, balance, refundAddress);
+        if (balance > UInt256.Zero)
+        {
+            var data = AbiEncoder.Instance.Encode(AbiEncodingStyle.Packed, new AbiSignature("", AbiType.UInt256),
+                balance);
```
LukaszRozmej: new AbiSignature("", AbiType.UInt256) - could be saved in static field

PR #9445, opened 2025-10-11
Update Dockerfiles automatically

rubo:
This is a follow-up to #9365. Since our Dockerfiles now have a hardcoded full version of .NET and an image digest for reproducible builds, we need a way to automate their update. Doing it manually is error-prone and prone to forgetting, which may cause unpredictable issues later.

## Changes

- Added a new workflow to update Dockerfiles with the latest digests and versions available. It creates a PR with changes, if any.
- The workflow runs automatically every midnight. Such a frequency is needed to be on par with that of GitHub Actions. If required, it can be triggered manually as well.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [x] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

Tested manually

### DIFF
--- a/.github/workflows/update-dockerfiles.yml
+++ b/.github/workflows/update-dockerfiles.yml
@@ -31,24 +31,24 @@ jobs:
           sdk_version=$(echo "$manifest" | jq -r '."latest-sdk"')
 
           # Get the digests for the specified tags
-          runtime_chiseled_tag=$runtime_version-noble-chiseled
-          runtime_chiseled_digest=$(skopeo inspect docker://mcr.microsoft.com/dotnet/aspnet:$runtime_chiseled_tag --no-tags | jq -r '.Digest')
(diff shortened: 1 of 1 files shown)


.github/workflows/update-dockerfiles.yml
```diff
@@ -0,0 +1,78 @@
+name: Update Dockerfiles
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
+  workflow_dispatch:
```
LukaszRozmej, line 6: Maybe this should run on every PR merged to master?
Do a quick run and create a PR only when actually needed?

rubo: It does create a PR only when needed as mentioned in the description ;) Running once a day is enough than running on each master commit.

PR #9448, opened 2025-10-13
Should correctly broadcast light transactions

LukaszRozmej:
## Changes

- Correctly use *GetProofVersion* for light transactions to determine broadcasting.

#### Requires testing

- [x] Yes

- [x] Yes

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
+++ b/src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
@@ -311,7 +311,7 @@ public sealed class MempoolBlobTxProofVersionValidator : ITxValidator
 
         ProofVersion? version = transaction.GetProofVersion();
         return version is null
-            ? TxErrorMessages.InvalidTransactionForm
+            ? transaction.NetworkWrapper is not null ? TxErrorMessages.InvalidTransactionForm : ValidationResult.Success
             : ValidateProofVersion(version.Value, releaseSpec);
 
         static ValidationResult ValidateProofVersion(ProofVersion txProofVersion, IReleaseSpec spec) =>


src/Nethermind/Nethermind.TxPool.Test/TxBroadcasterTests.cs
```diff
@@ -729,6 +729,29 @@ public void calculation_of_baseFeeThreshold_should_handle_overflow_correctly([Va
...
+
+        var blobTransaction = Build.A.Transaction
+            .WithShardBlobTxTypeAndFields()
+            .SignedAndResolved(_ethereumEcdsa, TestItem.PrivateKeyA)
+            .TestObject;
+
+        var lightTransaction = new LightTransaction(blobTransaction);
+
+        // Act
+        bool result = gossipPolicy.ShouldGossipTransaction(lightTransaction);
```
marcindsobczak, line 752: In production code, `gossipPolicy.ShouldGossipTransaction()` is called only once, with full blob tx as an argument

src/Nethermind/Nethermind.TxPool.Test/TxBroadcasterTests.cs
```diff
@@ -729,6 +729,29 @@ public void calculation_of_baseFeeThreshold_should_handle_overflow_correctly([Va
...
+
+        // Assert
+        result.Should().BeTrue("LightTransaction from blob transaction should be gossiped when proof version matches.");
```
marcindsobczak: And it is, as it is added to `_persistentBroadcast` after full blob tx passing the conditions from `SpecDrivenTxGossipPolicy.ShouldGossipTransaction()`

src/Nethermind/Nethermind.Core/Transaction.cs
```diff
@@ -332,12 +332,18 @@ public void CopyTo(Transaction tx)
             tx._size = _size;
             tx.AuthorizationList = AuthorizationList;
         }
+
+        public virtual ProofVersion? GetProofVersion() => SupportsBlobs
+            ? this is { NetworkWrapper: ShardBlobNetworkWrapper { Version: var version } }
+                ? version
+                : ProofVersion.V0
```
LukaszRozmej: should this return null if no wrapper?

flcl42: or throw an exception if no wrapper but proof version is requested?

PR #9449, opened 2025-10-13
Restrict IPC socket to owner-only mode if set

LukaszRozmej:
## Changes

- Adds owner-only mode restriction for IPC socket for better safety. Makes it configurable so other users process can access IPC if needed.

- [x] Other: security hardening

#### Requires testing

- [x] Yes

- [x] No

- [x] Yes

- [x] Yes

We are changing default behavior so only owner of Nethermind process has IPC access. This can be changed with `JsonRpc.RestrictIpcSocketPermissions false` flag to bring back previous behavior where any process on the machine can interact with Nethermind process through IPC.

## Remarks

If multiple users share the same machine (e.g., in shared hosting, CI/CD, or containerized environments), another user could:
- Open the socket directly,
- Call JSON-RPC methods like personal_listAccounts or admin_*,
- Potentially control the node or extract sensitive data.

Restricting the socket’s permissions ensures that only the node’s owner process (and root) can connect.

### DIFF
--- a/src/Nethermind/Nethermind.JsonRpc/IJsonRpcConfig.cs
+++ b/src/Nethermind/Nethermind.JsonRpc/IJsonRpcConfig.cs
@@ -53,7 +53,7 @@ public interface IJsonRpcConfig : IConfig
     [ConfigItem(Description = "The path to connect a UNIX domain socket over.")]
     string IpcUnixDomainSocketPath { get; set; }
 
-    [ConfigItem(Description = "Should Owner-only access be applied to IPC socket.", DefaultValue = "true")]
+    [ConfigItem(Description = "Should Owner-only 600 (rw-------) access be applied to IPC socket file path.", DefaultValue = "true")]
     bool RestrictIpcSocketPermissions { get; set; }
 
     [ConfigItem(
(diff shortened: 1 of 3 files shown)


src/Nethermind/Nethermind.JsonRpc/IJsonRpcConfig.cs
```diff
@@ -53,6 +53,9 @@ public interface IJsonRpcConfig : IConfig
     [ConfigItem(Description = "The path to connect a UNIX domain socket over.")]
     string IpcUnixDomainSocketPath { get; set; }
 
+    [ConfigItem(Description = "Should Owner-only 600 (rw-------) access be applied to IPC socket file path.", DefaultValue = "true")]
```
rubo:
```suggestion
    [ConfigItem(Description = "Whether to set the IPC socket UNIX file permissions to owner-only (600).", DefaultValue = "true")]
```

benaadams (approved):

flcl42 (approved):

asdacap (approved):

flcl42 (approved):
Might be cleaner to create with such permissions in the first place

svlachakis (approved):

LukaszRozmej:
@claude review

svlachakis (approved):

LukaszRozmej:
@claude re-review

Issue #9408, opened 2025-10-04
eth_getBalance responds with TrieException


MarekM25:
Curl command:
`   curl -s -X POST -H "Content-Type: application/json"     --data '{"jsonrpc":"2.0","method":"eth_getBalance","params":["0xcC4e00A72d871D6c328BcFE9025AD93d0a26dF51", "0x4a8411"],"id":1}'     X`
   Response:
[...]

PR #9457, opened 2025-10-14
Updated blockchainBridge hasStateForBlock to use config instead of st…

mrzeszutko:
…ate. Relates to #9408

## Changes

- Updated BlockChainBridge. HasStateForBlock to check the configuration and not call the state at all. The result is calculated based on the node type: archive node vs. full node, the information if we finish the initial sync and pruning boundary.
- To cater for the race condition, we return false at the pruning boundary (So we return that we don't have state when we are at the pruning boundary) - this is a safety margin

- [x] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

## Remarks

There were no changes done to the snap serving. 128 is hardcoded there so there should be no impact


src/Nethermind/Nethermind.Facade/IBlockchainBridge.cs
```diff
@@ -49,6 +49,16 @@ public interface IBlockchainBridge : ILogFinder
 
         bool TryGetLogs(int filterId, out IEnumerable<FilterLog> filterLogs, CancellationToken cancellationToken = default);
         void RunTreeVisitor<TCtx>(ITreeVisitor<TCtx> treeVisitor, Hash256 stateRoot) where TCtx : struct, INodeContext<TCtx>;
+
+        /// <summary>
+        /// Checks if state is available for the given block header using configuration-based logic to avoid race conditions.
+        /// For archive nodes (no pruning), returns true only when fully synced.
+        /// For pruning nodes, checks if the block is within the pruning window minus 1 block as a safety margin.
+        /// This conservative approach sacrifices 1 block of history to eliminate race conditions with pruning.
+        /// This method is optimized for RPC calls and may return false negatives to prevent exceptions.
```
LukaszRozmej: IMO everything past `Checks if state is available for the given block header` should be in <remarks>

src/Nethermind/Nethermind.Init/PruningTrieStateFactory.cs
```diff
@@ -202,6 +202,11 @@ ILogManager logManager
             pruningConfig.PruningBoundary = 64;
         }
 
+        // Add 1 to the configured pruning boundary to compensate for the -1 safety margin
+        // applied in BlockchainBridge when checking if state is available. This ensures users
+        // get exactly the number of blocks they configure.
+        pruningConfig.PruningBoundary += 1;
```
LukaszRozmej: Very weird we are modifying the config. Looks wrong to me.

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
```diff
@@ -382,9 +385,42 @@ public void RunTreeVisitor<TCtx>(ITreeVisitor<TCtx> treeVisitor, Hash256 stateRo
...
+            if (baseBlock is null)
+            {
+                return false;
+            }
+
+            // For archive nodes (no pruning), return true only when fully synced
+            if (pruningConfig.Mode == PruningMode.None)
+            {
+                // Archive node should have all state, but only after initial sync is complete
+                return !ethSyncingInfo.IsSyncing();
```
LukaszRozmej: Archive node has all state up to current head

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
```diff
@@ -382,9 +385,42 @@ public void RunTreeVisitor<TCtx>(ITreeVisitor<TCtx> treeVisitor, Hash256 stateRo
...
+            }
+
+            // For nodes with pruning enabled, check if we're still in initial sync
+            if (ethSyncingInfo.IsSyncing())
+            {
+                // Conservative: don't claim we have state during initial sync
+                // This prevents race conditions during sync process
+                return false;
+            }
```
LukaszRozmej: If we are past state sync (in old bodies or receipts sync) we do already have all state from synced block, up to pruning boundary

LukaszRozmej (changes requested):

src/Nethermind/Nethermind.Facade/IBlockchainBridge.cs
mrzeszutko: moved into remarks tag

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
mrzeszutko: changed to use `ethSyncingInfo.SyncMode.HaveNotSyncedStateYet()`

src/Nethermind/Nethermind.Init/PruningTrieStateFactory.cs
mrzeszutko: @MarekM25 what do you think about this? If we don't add 1 to the configured value rpc endpoints will always fail to return data at the pruning boundary - e.g. someone configures pruning boundary to be 250 and we will return data only from the last 249 blocks.

asdacap: Then its a bug in the TrieStore. Can you change it in there instead?

tanishqjasoria: Maybe define the safety margin constant in TrieStore and let it handle the logic. If user sets the pruning boundary to be 200 block then trieStore would automatically set it to 200 + SafetyMargin internally, and not modify the config value anywhere. 
This will make sure that TrieStore is responsible for keeping the state for PruningBoundary and we wont have to use this safety margin anywhere 

Remove it from here and BlockchainBridge.HasStateForBlock

tanishqjasoria (changes requested):

src/Nethermind/Nethermind.Init/PruningTrieStateFactory.cs
mrzeszutko: > Then its a bug in the TrieStore. Can you change it in there instead?

There is no bug in the TrieStore - what I mentioned is related to my changes in BlockchainBridge used by rpc endpoints to check if we have state

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
mrzeszutko: but it has all the state only after the initial sync is finished? Is there any other check I should do here instead?

src/Nethermind/Nethermind.Init/PruningTrieStateFactory.cs
mrzeszutko: > Maybe define the safety margin constant in TrieStore and let it handle the logic. If user sets the pruning boundary to be 200 block then trieStore would automatically set it to 200 + SafetyMargin internally, and not modify the config value anywhere. This will make sure that TrieStore is responsible for keeping the state for PruningBoundary and we wont have to use this safety margin anywhere
> 
> Remove it from here and BlockchainBridge.HasStateForBlock

That seems like a good way of implementing it - I see no potential issues with this approach and it will be clearer without magic change to the config value. @LukaszRozmej @MarekM25 what do you think about this?

LukaszRozmej: I still don't get it, are you depending on this flag in BlockchainBridge? Maybe you should be dependent on this flag -1 there?

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
LukaszRozmej: No, it always have state up until head.

src/Nethermind/Nethermind.Init/PruningTrieStateFactory.cs
mrzeszutko: as discussed I completely removed the automatic change of PruningBoundary

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
mrzeszutko: as discussed I changed the logic to `requestedNumber <= headNumber` for archive nodes

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
```diff
@@ -383,9 +389,44 @@ public void RunTreeVisitor<TCtx>(ITreeVisitor<TCtx> treeVisitor, Hash256 stateRo
...
         }
 
-        public bool HasStateForBlock(BlockHeader baseBlock)
+        public bool HasStateForBlock(BlockHeader? baseBlock)
         {
-            return stateReader.HasStateForBlock(baseBlock);
+            if (baseBlock is null)
+            {
+                return false;
+            }
```
LukaszRozmej: Why do we need to support nullability here?

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
```diff
@@ -383,9 +389,44 @@ public void RunTreeVisitor<TCtx>(ITreeVisitor<TCtx> treeVisitor, Hash256 stateRo
...
+
+            // For nodes with pruning enabled, check if we're still syncing state
+            // Once state sync is complete (even if still downloading old bodies/receipts),
+            // we have all state up to the pruning boundary
+            if (ethSyncingInfo.SyncMode.HaveNotSyncedStateYet())
+            {
+                // Conservative: don't claim we have state during state sync
+                // This prevents race conditions during sync process
+                return false;
+            }
```
LukaszRozmej: To be honest we are still skipping some scenarios here, not sure if worth to cover:
1. pruning off but still snap synced - archive will be from sync point only and no state before.
2. FullPruning on - archive will be available only from last full pruned block.

Those are kind of edge cases, but we should still probably support them?

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
mrzeszutko: we don't - I removed this

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
mrzeszutko: @LukaszRozmej I added handling of those edge cases, please re-review

src/Nethermind/Nethermind.Trie/Pruning/ITrieStore.cs
```diff
@@ -52,6 +52,12 @@ public interface IPruningTrieStore : ITrieStore
         // Acquire lock, then persist and flush cache.
         // Used for full pruning operation that change underlying node storage.
         TrieStore.StableLockScope PrepareStableState(CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Gets the block number of the last persisted block.
+        /// Used to track the oldest available state after full pruning operations.
+        /// </summary>
+        long LastPersistedBlockNumber { get; }
```
LukaszRozmej: Are you sure this is the OLDEST persisted block number or just LAST (newest) persisted block number?
Not entirely sure if it solves what we expect it does.
@asdacap can you advise?

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
```diff
@@ -385,7 +393,53 @@ public void RunTreeVisitor<TCtx>(ITreeVisitor<TCtx> treeVisitor, Hash256 stateRo
...
+            {
+                return false;
+            }
+
+            // For archive nodes (no pruning), check if requested block is within available range
+            if (pruningConfig.Mode == PruningMode.None)
+            {
+                // Archive mode: but state only available from sync pivot onwards
+                // If node was snap/fast synced, state before pivot was never downloaded
+                long syncPivotBlock = blockTree.SyncPivot.BlockNumber;
```
LukaszRozmej: Are we sure `BlockTree.SyncPivot` is reliable data here?
Is it 0 for full archive? Is it 1st block with state for archive with snap sync?
@asdacap can you advise?

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
```diff
@@ -385,7 +393,53 @@ public void RunTreeVisitor<TCtx>(ITreeVisitor<TCtx> treeVisitor, Hash256 stateRo
...
+
+            if (pruningConfig.Mode.IsFull())
+            {
+                // Full pruning enabled: track oldest state from multiple sources
+                // 1. Last persisted block (from trie store persistence/full pruning operations)
+                // 2. Sync pivot (state before this was never downloaded)
+                // 3. Current pruning window (head - boundary)
+                long lastPersistedBlock = pruningTrieStore.LastPersistedBlockNumber;
+                long syncPivotBlock = blockTree.SyncPivot.BlockNumber;
+                long pruningWindowStart = headNumber - pruningConfig.PruningBoundary + 1; // +1 for safety margin
```
LukaszRozmej: this calculation is used in both if and else path, let's name it as local method? Or do it before if and name correctly.

src/Nethermind/Nethermind.Trie/Pruning/ITrieStore.cs
asdacap: Its newest

asdacap:
I think the fundamental issue is that 'HasStateForBlock` check in `IWorldState` and `IStateReader` is not correct. I was kinda not correct as the existance of the root node does not necessarily mean that the whole trie is available. Its also weird that the root node is not pruned randomly.

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
asdacap: SyncPivot moves nowadays. So no.

asdacap:
Its children might get removed during in memory pruning. Thats why. Some of the nodes may not get removed due to memory limitation for tracking.

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
mrzeszutko: I removed SyncPivot dependency

src/Nethermind/Nethermind.Trie/Pruning/ITrieStore.cs
mrzeszutko: I removed this

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
mrzeszutko: not relevant anymore after my last changes

mrzeszutko:
> I think the fundamental issue is that 'HasStateForBlock`check in`IWorldState`and`IStateReader` is not correct. I was kinda not correct as the existance of the root node does not necessarily mean that the whole trie is available. Its also weird that the root node is not pruned randomly.

@asdacap is this relevant for this PR? For RPC calls we don't want to reference state, just base the response on config (e.g. pruning boundary etc)

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
```diff
@@ -403,41 +401,32 @@ public bool HasStateForBlock(BlockHeader baseBlock)
...
-                long syncPivotBlock = blockTree.SyncPivot.BlockNumber;
-                return requestedNumber >= syncPivotBlock && requestedNumber <= headNumber;
-            }
-
-            // For nodes with pruning enabled, determine the oldest available state block
-            long oldestAvailableState;
+                // For archive nodes that were snap synced, check the lowest inserted header
+                // We use LowestInsertedHeader (not SyncPivot) because SyncPivot moves forward over time
+                // as finalized blocks progress, while LowestInsertedHeader remains at the original starting point
+                long? lowestInsertedBlockNumber = blockTree.LowestInsertedHeader?.Number;
```
LukaszRozmej: This is false, we always have all headers.
We will probably always have all block bodies too.

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
```diff
@@ -403,41 +401,32 @@ public bool HasStateForBlock(BlockHeader baseBlock)
...
-                int pruningBoundary = pruningConfig.PruningBoundary;
-                oldestAvailableState = headNumber - pruningBoundary + 1;
+                // For archive nodes, state is available for all blocks from snap sync point to head
+                return requestedNumber <= headNumber;
             }
 
+            // Pruning nodes: calculate oldest available state based on pruning boundary
+            // This is conservative but correct for all pruning node configurations:
+            // - Memory pruning: respects the configured boundary
+            // - Full pruning: also uses pruning boundary (persisted state within this window)
```
LukaszRozmej: I think it doesn't work for full pruning only like this

LukaszRozmej (changes requested):

src/Nethermind/Nethermind.Init/Modules/WorldStateModule.cs
```diff
@@ -76,6 +77,7 @@ dbFactory is not MemDbFactory
 
             .Map<IWorldStateManager, PruningTrieStateFactoryOutput>((o) => o.WorldStateManager)
             .Map<IStateReader, IWorldStateManager>((m) => m.GlobalStateReader)
+            .Map<IPruningTrieStore, MainPruningTrieStoreFactory>((f) => f.PruningTrieStore)
```
asdacap: I think you dont need this anymore. Please dont use `IPruningTrieStore`. It will make migrating to other world state layout harder.

asdacap:
I think you can simply check for `IBlockTree.BestPersistedState`, which is kinda the same as `TrieStore.LastPersistedBlockNumber`. It should already have pruning boundary considered. Except for archive mode, then just dont check.

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
```diff
@@ -385,7 +391,44 @@ public void RunTreeVisitor<TCtx>(ITreeVisitor<TCtx> treeVisitor, Hash256 stateRo
 
         public bool HasStateForBlock(BlockHeader baseBlock)
         {
-            return stateReader.HasStateForBlock(baseBlock);
+            long headNumber = blockTree.Head?.Number ?? 0;
+            long requestedNumber = baseBlock.Number;
+
+            // Conservative: don't claim we have state during state sync
+            // This prevents race conditions during the sync process for all node types
+            if (ethSyncingInfo.SyncMode.HaveNotSyncedStateYet())
```
MarekM25: This can be incorrect in edge case - archive node, starting from scratch, archive node should always have state

stdevMac:
Closing: the target issue #9408 was fixed by #10534, which added a `BestPersistedState` check to `BlockchainBridge.HasStateForBlock` — the approach suggested in the review thread here.

PR #9473, opened 2025-10-15
Changes for Arbitrum mainnet

svlachakis:
Extensibility changes on `BlockTree.cs` and `BlockHeader.cs ` for Arbitrum Mainnet.


src/Nethermind/Nethermind.Core/BlockHeader.cs
```diff
@@ -44,7 +44,7 @@ public BlockHeader(
         ExcessBlobGas = excessBlobGas;
     }
 
-    public bool IsGenesis => Number == 0L;
+    public virtual bool IsGenesis => Number == 0L;
```
LukaszRozmej: Maybe something better?

```suggestion
    public virtual long GenesisBlockNumber => 0; // this would be better as static virtual 
    public bool IsGenesis => Number == GenesisBlockNumber; // this could be extension method.
```

svlachakis: true that's better - done!

src/Nethermind/Nethermind.Core/BlockHeader.cs
```diff
@@ -44,7 +44,8 @@ public BlockHeader(
         ExcessBlobGas = excessBlobGas;
     }
 
-    public bool IsGenesis => Number == 0L;
+    public virtual long GenesisBlockNumber => 0;
```
hudem1: can make it static

PR #9474, opened 2025-10-15
Fix/more flaky tests

LukaszRozmej:

- [x] Other: tests

### DIFF
--- a/src/Nethermind/Nethermind.Era1.Test/EraImporterTest.cs
+++ b/src/Nethermind/Nethermind.Era1.Test/EraImporterTest.cs
@@ -136,6 +136,7 @@ public class EraImporterTest
     }
 
     [CancelAfter(4000)]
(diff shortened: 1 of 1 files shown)


benaadams (approved):

src/Nethermind/Nethermind.Init/Modules/MainProcessingContext.cs
```diff
@@ -60,7 +60,8 @@ public MainProcessingContext(
                 {
                     IsMainProcessor = true // Manual construction because of this flag
                 })
-
+                .AddScoped<IBlockchainProcessor>(ctx => ctx.Resolve<BlockchainProcessor>())
+                .AddScoped<IBlockProcessingQueue>(ctx => ctx.Resolve<BlockchainProcessor>())
```
benaadams, line 64: Are these both same?

PR #9475, opened 2025-10-15
Xdc timeout handler

cicr99:


src/Nethermind/Nethermind.Xdc/TimeoutCertificateManager.cs
```diff
@@ -3,41 +3,87 @@
...
+    private void OnTimeoutPoolThresholdReached(Timeout timeout)
     {
-        throw new NotImplementedException();
+        Signature[] signatures = _timeouts.GetItems(timeout).Select(t => t.Signature).ToArray();
+
+        var timeoutCertificate = new TimeoutCertificate(timeout.Round, signatures, timeout.GapNumber);
+
+        ProcessTimeoutCertificate(timeoutCertificate);
+
+        SyncInfo syncInfo = _syncInfoManager.GetSyncInfo();
```
Demuirgos, line 72: hmmm this may be problematic, ISyncInfoManager will end up depending on ITimeoutCertificateManager, so having ITimeoutCertificateManager depend on ISyncInfoManager , creates a circular loop

PR #9476, opened 2025-10-15
Add `NetworkId` flag to `InitConfig`

marcindsobczak:
## Changes

- Add `NetworkId` flag to `InitConfig` which is needed for bloatnet

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Api/IInitConfig.cs
+++ b/src/Nethermind/Nethermind.Api/IInitConfig.cs
@@ -39,7 +39,7 @@ public interface IInitConfig : IConfig
     [ConfigItem(Description = "The hash of the genesis block. If not specified, the genesis block validity is not checked which is useful in the case of ad hoc test/private networks.", DefaultValue = "null")]
     string? GenesisHash { get; set; }
 
-    [ConfigItem(Description = "The network id. If not specified, taken from the chain spec file.", DefaultValue = "null")]
+    [ConfigItem(Description = "The network id. If not specified, taken from the chain spec file.", DefaultValue = "null", HiddenFromDocs = true)]
     ulong? NetworkId { get; set; }
 
     [ConfigItem(Description = "The path to the static nodes file.", DefaultValue = "static-nodes.json")]


src/Nethermind/Nethermind.Api/IInitConfig.cs
```diff
@@ -39,6 +39,9 @@ public interface IInitConfig : IConfig
     [ConfigItem(Description = "The hash of the genesis block. If not specified, the genesis block validity is not checked which is useful in the case of ad hoc test/private networks.", DefaultValue = "null")]
     string? GenesisHash { get; set; }
 
+    [ConfigItem(Description = "The Network Id. If not specified, the NetworkId is taken from the chain spec file.", DefaultValue = "null")]
```
rubo:
```suggestion
Comment view+
    [ConfigItem(Description = "The network identifier. If not specified, taken from the chain spec file.", DefaultValue = "null")]
```

flcl42:
Maybe just hide it from docs. Network id is actually not a consensus part, while genesis is about consensus, so not so bad. Can be in network config?

marcindsobczak:
For now it's needed only for bloatnet, so I chery-picked it to performance branch and converting PR to the draft. I would like to avoid bigger refactors just for the purpose of this simple adjustment - hiding it from docs and merging or keeping in a branch outside of master seems reasonable

MarekM25 (commented):
Cover this behaviour with tests and should be okay?

src/Nethermind/Nethermind.Api/IInitConfig.cs
```diff
@@ -39,6 +39,9 @@ public interface IInitConfig : IConfig
     [ConfigItem(Description = "The hash of the genesis block. If not specified, the genesis block validity is not checked which is useful in the case of ad hoc test/private networks.", DefaultValue = "null")]
     string? GenesisHash { get; set; }
 
+    [ConfigItem(Description = "The network id. If not specified, taken from the chain spec file.", DefaultValue = "null")]
```
benaadams:
```suggestion
    [ConfigItem(Description = "The network id. If not specified, taken from the chain spec file.", DefaultValue = "null", HiddenFromDocs = true)]
```

PR #7891, opened 2024-12-11
Implement admin_addTrustedPeer rpc endpoint

richardgreg:
Fixes #7811

## Changes

1. **Added ITrustedNodesManager Interface:**
Created a new ITrustedNodesManager interface to handle trusted nodes. This interface is similar to IStaticNodesManager but for trusted nodes. It provides methods InitAsync(), AddAsync(), RemoveAsync(), IsTrusted(), and exposes Nodes.

1. **Implemented TrustedNodesManager Class:**
Implemented TrustedNodesManager, which reads/writes a trusted nodes file and manages a list of trusted nodes. It also implements INodeSource (or inherits from ITrustedNodesManager : INodeSource) so it can be used by CompositeNodeSource.

1. **Added IsTrusted Property to Node:**
Modified the Node class used by the networking code to include a public bool IsTrusted { get; set; } property. This allows peers to reflect if they originate from trusted nodes.

1. **Integrated TrustedNodesManager into PeerPool:**
Passed an ITrustedNodesManager instance into the PeerPool constructor and updated the peer creation logic (CreateNew methods) to mark nodes as trusted if they come from the trusted nodes manager.

1. **Updated InitializeNetwork to Include TrustedNodesManager:**
Created and initialized TrustedNodesManager inside InitializeNetwork.cs, added it to CompositeNodeSource so that trusted nodes are considered in the node discovery pipeline, and passed it to PeerPool.

1. **Updated IApiWithNetwork and IInitConfig:**

Extended IApiWithNetwork interface to have a ITrustedNodesManager? TrustedNodesManager { get; set; } property (if needed).
Added a TrustedNodesPath property to IInitConfig and InitConfig to define the file path for trusted nodes.
Modified AdminRpcModule Constructor:
Updated the AdminRpcModule constructor to require an ITrustedNodesManager and implemented the admin_addTrustedPeer method using the trusted nodes manager and peer pool.

1. **Updated Tests:**
Adjusted existing tests to provide an ITrustedNodesManager mock where needed. Added a new test (Test_admin_addTrustedPeer) to ensure admin_addTrustedPeer works correctly.

#### What types of changes does your code introduce?

- [ ] Bugfix (a non-breaking change that fixes an issue)
- [x] New feature (a non-breaking change that adds functionality)
- [ ] Breaking change (a change that causes existing functionality not to work as expected)
- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

[...]

PR #9477, opened 2025-10-15
Move static/trusted nodes file to the data dir

### LINKED ISSUE #9181: static-nodes.json should live within the data dir
**Description**

When I run a node with my own datadir, and then do an `admin_addTrustedPeer` RPC call, I get the following error:

> data": "System.UnauthorizedAccessException: Access to the path '/nethermind/Data/trusted-nodes.json' is denied.\n ---> System.IO.IOException: Permission denied\n   --- End of inner exception stack trace ---\n   at Interop.ThrowExceptionForIoErrno(ErrorInfo errorInfo, String path, Boolean isDirError)\n   at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String fullPath, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize, UnixFileMode openPermissions, Int64& fileLength, UnixFileMode& filePermissions, Boolean failForSymlink, Boolean& wasSymlink, Func`4 createOpenException)\n   at System.IO.Strategies.OSFileStreamStrategy..ctor(String path, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocationSize, Nullable`1 unixCreateMode)\n   at System.IO.File.Create(String path)\n   at Nethermind.Network.TrustedNodesManager.SaveFileAsync() in /src/Nethermind/Nethermind.Network/TrustedNodes/TrustedNodesManager.cs:line 186\n   at Nethermind.Network.TrustedNodesManager.AddAsync(Enode enode, Boolean updateFile) in /src/Nethermind/Nethermind.Network/TrustedNodes/TrustedNodesManager.cs:line 125\n   at Nethermind.JsonRpc.Modules.Admin.AdminRpcModule.admin_addTrustedPeer(String enode) in /src/Nethermind/Nethermind.JsonRpc/Modules/Admin/AdminRpcModule.cs:line 132\n   at Nethermind.JsonRpc.JsonRpcService.ExecuteAsync(JsonRpcRequest request, String methodName, ResolvedMethodInfo method, JsonRpcContext context) in /src/Nethermind/Nethermind.JsonRpc/JsonRpcService.cs:line 203

The problem is that I'm running Nethermind within a container, and I'm setting `--datadir=/data`. The container has write access to the `/data` directory only. Somehow the `static-nodes.json` file lives in a different directory within `/nethermind/Data`. This sounds wrong and the `static-nodes.json` file should ideally also live in the directory defined via the datadir flag.

rubo:
Fixes #9181 

## Changes

- Moved `static-nodes.json` to the Nethermind data directory. The original `static-nodes.json` is embedded into the assembly and copied to the data directory if it doesn't already exist there.
- Moved the `trusted-nodes.json` default location to the Nethermind data directory. It's now created if doesn't exist like the static nodes file.
- Refactored and fixed `TrustedNodesManager` initialization to handle `trusted-nodes.json` correctly. See the [comment](https://github.com/NethermindEth/nethermind/pull/7891#discussion_r2432964311) in #7891
- Moved shared code from `StaticNodesManager` and `TrustedNodesManager` to the new `NodesManager` class
- Added `IInitConfig.DataDir` to store the `--data-dir` value. It's needed later elsewhere. Otherwise, the `Program.cs` would be cluttered if we handled all data dir-related stuff there.
- Removed `Init.StaticNodesPath` from Sepolia configs as the file is long gone.

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

Tested manually as well

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

> Until now, Nethermind has been using the `Data` directory to store static and trusted nodes. This caused issues with setups when write permissions are restricted to specific paths only, and the [`Init.StaticNodesPath`](https://docs.nethermind.io/fundamentals/configuration#init-staticnodespath) is not set or has the default value. To address these issues, static and trusted nodes are now expected to be in the Nethermind data directory (see [`--data-dir`](https://docs.nethermind.io/fundamentals/configuration#data-dir)) by default. If you have existing nodes files in the default location, move them to the data directory.

### DIFF
--- a/src/Nethermind/Nethermind.Network/NodesManager.cs
+++ b/src/Nethermind/Nethermind.Network/NodesManager.cs
@@ -53,7 +53,7 @@ public abstract class NodesManager(string path, ILogger logger)
         {
             var separator = $"{Environment.NewLine}  ";
 
-            if (_logger.IsDebug) _logger.Debug($"{title}:{separator}{string.Join(separator, nodes.Values.Select(n => n.ToString()))}");
+            _logger.Debug($"{title}:{separator}{string.Join(separator, nodes.Values.Select(n => n.ToString()))}");
         }
     }
 


src/Nethermind/Nethermind.Network/StaticNodes/static-nodes.json
rubo, line 1: @LukaszRozmej This file is currently empty. What's the point of having it distributed with Nethermind? Maybe it's worth doing the way trusted nodes are done -- no file is there until needed.

LukaszRozmej: I think it was there so users could edit it manually when needed - so they can discover it.
`trusted-nodes.json` are not discoverable unless you create it - kind of bad.

src/Nethermind/Nethermind.Network/StaticNodes/StaticNodesManager.cs
```diff
@@ -18,158 +18,150 @@
...
+            if (_logger.IsTrace) _logger.Trace($"Static nodes file was not found, creating one at {Path.GetFullPath(staticNodesPath)}");
 
-            _nodes = new ConcurrentDictionary<PublicKey, NetworkNode>(networkNodes.ToDictionary(static n => n.NodeId, static n => n));
+            await embeddedNodes.CopyToAsync(actualNodes);
         }
 
-        private static string[] GetNodes(string data)
+        string data = await File.ReadAllTextAsync(staticNodesPath);
+        IEnumerable<string> nodes = INodeSource.ParseNodes(data);
+        var nodeCount = nodes.Count();
```
LukaszRozmej: If we need Count, maybe better to return something else than IEnumerable?

src/Nethermind/Nethermind.Network/INodeSource.cs
```diff
@@ -1,15 +1,36 @@
...
+        try
+        {
+            nodes = JsonSerializer.Deserialize<string[]>(data) ?? [];
+        }
+        catch (JsonException)
+        {
+            nodes = data.Split(separator, StringSplitOptions.RemoveEmptyEntries);
+        }
+
+        return nodes.Distinct();
```
LukaszRozmej: This looks like we should deserialize to `HashSet`

src/Nethermind/Nethermind.Network/StaticNodes/static-nodes.json
rubo: Hmm... What do you think about mentioning that in the docs instead of providing empty files?

src/Nethermind/Nethermind.Network/StaticNodes/StaticNodesManager.cs
```diff
@@ -49,33 +49,37 @@ public async Task InitAsync()
...
+            _logger.Info($"Loaded {nodeSet.Count} static nodes from {Path.GetFullPath(staticNodesPath)}");
 
-        if (nodeCount != 0)
+        if (nodeSet.Count != 0)
         {
             if (_logger.IsDebug) _logger.Debug($"Static nodes: {Environment.NewLine}{data}");
         }
 
-        List<NetworkNode> networkNodes = [];
+        List<NetworkNode> nodes = [];
```
LukaszRozmej: Why do we need this List, then we do ToDictionary, then we create ConcurrentDictionary

Can we just have ConcurrentDictionary in the first place?

rubo: I'll check. Trying not to make too many changes to the original logic, so as not to introduce new issues.

src/Nethermind/Nethermind.Network/StaticNodes/static-nodes.json
LukaszRozmej: Docs should be obvious yes, but empty files don't hurt and if someone peeks into dir it would be very helpful for them to see those. I am for both :)

I see your point that from purist perspective it is a waste, but the cost is practically none and even when someone reads the docs he won't need to figure out where exactly to put the file in - it is there.

src/Nethermind/Nethermind.Network/StaticNodes/StaticNodesManager.cs
LukaszRozmej: I think it is doing exactly this in the 2nd place (maybe we can unify that code?). And it isn't that big of a change.

src/Nethermind/Nethermind.Network/TrustedNodes/TrustedNodesManager.cs
```diff
@@ -17,176 +16,160 @@
...
-            _trustedNodesPath = trustedNodesPath.GetApplicationResourcePath();
-            _logger = logManager.GetClassLogger();
+            if (_logger.IsDebug) _logger.Debug($"Trusted nodes file not found at {Path.GetFullPath(trustedNodesPath)}");
+            return;
         }
 
-        public IEnumerable<NetworkNode> Nodes => _nodes.Values;
+        string data = await File.ReadAllTextAsync(trustedNodesPath);
+        ISet<string> nodeSet = INodeSource.ParseNodes(data);
+        ConcurrentDictionary<PublicKey, NetworkNode> nodes = [];
```
LukaszRozmej: Unify with this code?

LukaszRozmej: By unify I meant just have a helper method or something that is called for both :)

src/Nethermind/Nethermind.Network/StaticNodes/static-nodes.json
rubo: I see. Now it creates an empty `trusted-nodes.json` too if it doesn't exist.

PR #9481, opened 2025-10-15
Log index (preparation)

alexb5dh:
Helper methods/features/fixes needed for #8464 and extracted from that PR.

## Changes

- RocksDB:
  - Support for [merging](https://github.com/facebook/rocksdb/wiki/merge-operator) (via `IMergeOperator`,  `MergeOperatorAdapter`, and `IMergeableKeyValueStore`)
  - Support for clearing/cancelling RocksDB batch (via `IWriteBatch.Clear`)
  - ~~Expose RocksDB iterator API (via `IIterator` and `RocksDbIteratorWrapper`)~~
  - `StartBefore` method for `ISortedView` - to try start just before the specified key
  - Fix for `ColumnsDb` flushing whole DB instead of a related column family
  - Fix for rare case of `NullReferenceException` when clearing iterators

- New `IReceiptStorage.ReceiptsInserted` event, which will be invoked for any receipts insertion. Old `ReceiptsInserted`, which was invoked only for the new main blocks, renamed to `NewCanonicalReceipts`.
- Fix for `ProgressLogger` cancelling completion no `LogProgress` invocation
- Support for byte array as a JSON property in RPC (serialized as hex string)
- Some testing helper classes/extensions

#### Requires testing

- [x] No

- [x] No

- [x] No

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Core/Collections/DictionaryExtensions.cs
+++ b/src/Nethermind/Nethermind.Core/Collections/DictionaryExtensions.cs
@@ -15,14 +15,21 @@ public static class DictionaryExtensions
         res++;
     }
 
-    public static TValue GetOrAdd<TKey, TValue>(this Dictionary<TKey, TValue> dictionary, TKey key, Func<TKey, TValue> factory)
+    public static TValue GetOrAdd<TKey, TValue>(this Dictionary<TKey, TValue> dictionary,
+        TKey key, Func<TKey, TValue> factory,
+        out bool exists)
         where TKey : notnull
     {
-        ref TValue? existing = ref CollectionsMarshal.GetValueRefOrAddDefault(dictionary, key, out bool exists);
+        ref TValue? existing = ref CollectionsMarshal.GetValueRefOrAddDefault(dictionary, key, out exists);
 
         if (!exists)
             existing = factory(key);
 
         return existing!;
     }
+
+    public static TValue GetOrAdd<TKey, TValue>(this Dictionary<TKey, TValue> dictionary,
+        TKey key, Func<TKey, TValue> factory)
+        where TKey : notnull =>
+        GetOrAdd(dictionary, key, factory, out _);
 }
--- a/src/Nethermind/Nethermind.Core/JsonConverters/ByteArrayConverter.cs
+++ b/src/Nethermind/Nethermind.Core/JsonConverters/ByteArrayConverter.cs
@@ -177,7 +177,7 @@ public class ByteArrayConverter : JsonConverter<byte[]>
     private static void ThrowFormatException() => throw new FormatException();
 
     [DoesNotReturn, StackTraceHidden]
-    private static Exception ThrowInvalidOperationException() => throw new InvalidOperationException();
+    private static void ThrowInvalidOperationException() => throw new InvalidOperationException();
 
     public override void Write(
         Utf8JsonWriter writer,
@@ -253,7 +253,12 @@ public class ByteArrayConverter : JsonConverter<byte[]>
 
     public override byte[] ReadAsPropertyName(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
     {
-        return Convert(ref reader) ?? throw ThrowInvalidOperationException();
+        byte[]? result = Convert(ref reader);
+
+        if (result is null)
+            ThrowInvalidOperationException();
+
+        return result;
     }
 
     public override void WriteAsPropertyName(Utf8JsonWriter writer, byte[] value, JsonSerializerOptions options)


src/Nethermind/Nethermind.Core/IWriteBatch.cs
```diff
@@ -5,5 +5,8 @@
 
 namespace Nethermind.Core
 {
-    public interface IWriteBatch : IDisposable, IWriteOnlyKeyValueStore;
+    public interface IWriteBatch : IDisposable, IWriteOnlyKeyValueStore
+    {
+        void Clear();
```
alexb5dh, line 10: `Clear` is needed to reset data in case of an error in the middle of the transaction, when no data should actually be written. Current version with write on `Dispose` will submit it regardless.

I would prefer to make the batch follow explicit-commit approach, but that would affect a lot of code.

src/Nethermind/Nethermind.Db.Rocks/ColumnDb.cs
```diff
@@ -130,7 +145,7 @@ public bool KeyExists(ReadOnlySpan<byte> key)
 
     public void Flush(bool onlyWal)
     {
-        _mainDb.Flush(onlyWal);
+        _mainDb.FlushWithColumnFamily(_columnFamily);
```
alexb5dh, line 148: `onlyWal` will be ignored here but looks liks it's not supported by CF flush.

src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
```diff
@@ -66,6 +66,11 @@ public partial class DbOnTheRocks : IDb, ITunableDb, IReadOnlyNativeKeyValueStor
 
     private readonly DbSettings _settings;
 
+    // ReSharper disable once CollectionNeverQueried.Local
+    // Need to keep options from GC in case of merge operator applied, as they are used in callback
+    // TODO: find better way?
+    private readonly ConcurrentBag<OptionsHandle> _doNotGcOptions = [];
```
alexb5dh: Open for discussion.

Need to preserve created `OptionsHandle` from GC as they are used for native merge callbacks.

src/Nethermind/Nethermind.Blockchain/Receipts/IReceiptStorage.cs
```diff
@@ -19,8 +19,13 @@ public interface IReceiptStorage : IReceiptFinder
...
         /// <summary>
-        /// Receipts for a block are inserted
+        /// Receipts for a new main block are inserted
         /// </summary>
         event EventHandler<BlockReplacementEventArgs> ReceiptsInserted;
+
+        /// <summary>
+        /// Any receipts are inserted
+        /// </summary>
+        event EventHandler<ReceiptsEventArgs> AnyReceiptsInserted;
```
alexb5dh: Naming is open for discussion.

Problem is that method above (`ReceiptsInserted`) is only invoked for the new head block (regular sync or reorg), while this one is actually triggered for any new receipts, including backward sync. Also, args are different - as needed by index.

src/Nethermind/Nethermind.Db/RocksDbSettings.cs
```diff
@@ -17,6 +20,9 @@ public DbSettings(string name, string path)
         public bool DeleteOnStart { get; set; }
         public bool CanDeleteFolder { get; set; } = true;
 
+        public IMergeOperator? MergeOperator { get; set; }
+        public Dictionary<string, IMergeOperator>? MergeOperatorByColumnFamily { get; set; }
```
alexb5dh: Option to provide custom merge operator per column family. Naming and property location is open for discussion.

src/Nethermind/Nethermind.Core/Collections/CollectionExtensions.cs
```diff
@@ -87,6 +87,21 @@ public static bool NoResizeClear<TKey, TValue>(this ConcurrentDictionary<TKey, T
             return true;
         }
 
+        public static TValue GetOrAdd<TKey, TValue>(this Dictionary<TKey, TValue> dictionary, TKey key, Func<TKey, TValue> factory)
+            where TKey : notnull
+        {
+            if (dictionary.TryGetValue(key, out TValue? value))
```
asdacap: There is `CollectionMarshall.GetValueRefOrAddDefault`

src/Nethermind/Nethermind.Core/IKeyValueStore.cs
```diff
@@ -59,6 +59,7 @@ public interface IWriteOnlyKeyValueStore
         }
 
         void Set(ReadOnlySpan<byte> key, byte[]? value, WriteFlags flags = WriteFlags.None);
+        void Merge(ReadOnlySpan<byte> key, ReadOnlySpan<byte> value, WriteFlags flags = WriteFlags.None);
```
asdacap: Consider having a dedicated interface for this.

src/Nethermind/Nethermind.Core/JsonConverters/ByteArrayConverter.cs
```diff
@@ -173,7 +173,7 @@ public static void Convert(ref Utf8JsonReader reader, scoped Span<byte> span)
     private static void ThrowFormatException() => throw new FormatException();
 
     [DoesNotReturn, StackTraceHidden]
-    private static void ThrowInvalidOperationException() => throw new InvalidOperationException();
+    private static Exception ThrowInvalidOperationException() => throw new InvalidOperationException();
```
asdacap: Why?

src/Nethermind/Nethermind.Db.Rocks/Config/PerTableDbConfig.cs
```diff
@@ -38,7 +38,7 @@ private void EnsureConfigIsAvailable(string propertyName)
         foreach (var prefix in _prefixes)
         {
             string prefixed = string.Concat(prefix, propertyName);
-            if (type.GetProperty(prefixed, BindingFlags.Public | BindingFlags.Instance) is null)
+            if (GetProperty(type, prefixed, caseSensitive: true) is null)
```
asdacap: Why change?

asdacap: Nevermind

src/Nethermind/Nethermind.Db.Rocks/RocksDbIteratorWrapper.cs
```diff
@@ -0,0 +1,21 @@
+// SPDX-FileCopyrightText: 2024 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using RocksDbSharp;
+using System;
+
+namespace Nethermind.Db.Rocks;
+
+public sealed class RocksDbIteratorWrapper(Iterator iterator) : IIterator
```
asdacap: Is this necessary? Can you modify `ISortedKeyValueStore` instead?

src/Nethermind/Nethermind.Blockchain/Receipts/IReceiptStorage.cs
LukaszRozmej: I have a mixed feeling about this and would prefer one or the other:

1. Reuse `ReceiptsInserted` and have some additional arguments saying if they are canonical and from block processing. The only usage is in `ReceiptCanonicalityMonitor` so we can work that out.
2. Renaming this to `ReceiptsInserted` and `ReceiptsInserted` to `NewCanonicalReceipts` or something like that?

src/Nethermind/Nethermind.Core.Test/Json/ConverterTestBase.cs
```diff
@@ -10,7 +11,10 @@ namespace Nethermind.Core.Test.Json;
 
 public class ConverterTestBase<T>
 {
-    protected void TestConverter(T? item, Func<T, T, bool> equalityComparer, JsonConverter<T> converter)
+    protected void TestConverter(T? item, Func<T, T, bool> equalityComparer, JsonConverter<T> converter) =>
+        TestConverter(item, converter, equalityComparer);
+
+    protected void TestConverter<R>(R? item, JsonConverter<T> converter, Func<R, R, bool>? equalityComparer = null)
```
LukaszRozmej: We follow MS guidelines so rename `R`->`TValue`

src/Nethermind/Nethermind.Core.Test/Json/ConverterTestBase.cs
```diff
@@ -22,10 +26,15 @@ protected void TestConverter(T? item, Func<T, T, bool> equalityComparer, JsonCon
 
         string result = JsonSerializer.Serialize(item, options);
 
-        T? deserialized = JsonSerializer.Deserialize<T>(result, options);
+        R? deserialized = JsonSerializer.Deserialize<R>(result, options);
 
 #pragma warning disable CS8604
-        Assert.That(equalityComparer(item, deserialized), Is.True);
+        if (equalityComparer != null)
```
LukaszRozmej:
```suggestion
        if (equalityComparer is not null)
```

src/Nethermind/Nethermind.Core/Collections/CollectionExtensions.cs
LukaszRozmej: This allows to avoid double lookup in ditionary (one on `TryGetValue` and second on `Add`)

src/Nethermind/Nethermind.Core/Extensions/Bytes.cs
```diff
@@ -276,6 +276,15 @@ public static byte[] PadRight(this byte[] bytes, int length)
 
             byte[] result = new byte[length];
             Buffer.BlockCopy(bytes, 0, result, 0, bytes.Length);
+
+            if (padding != 0)
+            {
+                for (int i = bytes.Length; i < length; i++)
+                {
+                    result[i] = padding;
+                }
```
LukaszRozmej:
```suggestion
                result.AsSpan(bytes.Length, length - bytes.Length).Fill(padding);
```

src/Nethermind/Nethermind.Core/Utils/ConcurrentWriteBatcher.cs
```diff
@@ -40,6 +40,24 @@ public void PutSpan(ReadOnlySpan<byte> key, ReadOnlySpan<byte> value, WriteFlags
...
+    {
+        IWriteBatch currentBatch = RentWriteBatch();
+        currentBatch.Merge(key, value, flags);
+        ReturnWriteBatch(currentBatch);
+    }
+
+    public void Clear()
+    {
+        if (_disposing)
+            throw new InvalidOperationException("Batch is already disposed.");
```
LukaszRozmej: Put it into `ThrowXXX` method (avoids JITing until needed and decreases hot-path code size)

src/Nethermind/Nethermind.Core/FakeWriteBatch.cs
```diff
@@ -31,5 +31,15 @@ public void Set(ReadOnlySpan<byte> key, byte[]? value, WriteFlags flags = WriteF
...
         }
+
+        public void Merge(ReadOnlySpan<byte> key, ReadOnlySpan<byte> value, WriteFlags flags = WriteFlags.None)
+        {
+            throw new NotSupportedException("Merging is not supported by this implementation.");
+        }
+
+        public void Clear()
+        {
+            throw new NotSupportedException("Clearing is not supported by this implementation.");
```
LukaszRozmej: this probably shouldn't throw

src/Nethermind/Nethermind.Db.Rocks/ColumnsDb.cs
```diff
@@ -61,9 +61,9 @@ private static IReadOnlyList<T> GetEnumKeys(IReadOnlyList<T> keys)
         return keys;
     }
 
-    protected override void BuildOptions<O>(IRocksDbConfig dbConfig, Options<O> options, IntPtr? sharedCache)
+    protected override void BuildOptions<O>(IRocksDbConfig dbConfig, Options<O> options, IntPtr? sharedCache, IMergeOperator? mergeOperator)
```
LukaszRozmej: O -> T

src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
LukaszRozmej: ConcurrentBag is very peculiar collection and something you probably don't want. I think simple List with locks is better here. Otherwise ConcurrentSet.

src/Nethermind/Nethermind.Db.Rocks/MergeOperatorAdapter.cs
```diff
@@ -0,0 +1,80 @@
...
+
+            resultLength = result.Length;
+
+            // Fixing RocksDbSharp invalid callback signature, TODO: submit an issue/PR
+            Unsafe.SkipInit(out success);
+            Unsafe.As<IntPtr, byte>(ref success) = 1;
+
+            return (IntPtr)resultPtr;
+        }
+    }
```
LukaszRozmej, line 41: Yeah this pattern seems wrong to me, what I would do:

1. Add those ArrayPoolLists into a Dictionary by `IntPtr`
2. fix their internal array in memory and return their pointer of internal array
3. On Delete go through the dictionary based on IntPtr unfix the the array and remove the ArrayPoolList from Dictionary.

You can use `GCHandle` for pinning.

src/Nethermind/Nethermind.Db.Rocks/RocksDbIteratorWrapper.cs
LukaszRozmej: I think we didn't have that interface when this was written, so it might be good to reevaluate and it would be also mostly mod of `ISortedView`

src/Nethermind/Nethermind.Db/CompressingDb.cs
```diff
@@ -52,6 +54,11 @@ public void PutSpan(ReadOnlySpan<byte> key, ReadOnlySpan<byte> value, WriteFlags
                     _wrapped.PutSpan(key, Compress(value, stackalloc byte[value.Length]), flags);
                 }
 
+                public void Merge(ReadOnlySpan<byte> key, ReadOnlySpan<byte> value, WriteFlags flags = WriteFlags.None)
+                {
+                    throw new InvalidOperationException("EOA compressing DB does not support merging");
```
LukaszRozmej, line 59: I think we can support it here and do simple pass-through?

src/Nethermind/Nethermind.Db/RocksDbSettings.cs
LukaszRozmej: I suggest renaming to `ColumnMergeOperators` or `ColumnsMergeOperators`

src/Nethermind/Nethermind.Db/SimpleFilePublicKeyDb.cs
```diff
@@ -325,5 +325,20 @@ private byte[] Add(byte[] value)
...
+
+        public IIterator GetIterator(ref IteratorOptions options)
+        {
+            throw new NotSupportedException("Iteration is not supported by this implementation.");
+        }
+
+        public void Merge(ReadOnlySpan<byte> key, ReadOnlySpan<byte> value, WriteFlags flags = WriteFlags.None)
+        {
+            throw new NotSupportedException("Merging is not supported by this implementation.");
+        }
```
LukaszRozmej: I think we have too many Db stuff with `NotSupportedException`. This makes me think we need interface segregation for those features.

src/Nethermind/Nethermind.Blockchain/Receipts/IReceiptStorage.cs
alexb5dh: Event args are different, so option 2 sound better.

src/Nethermind/Nethermind.Core/JsonConverters/ByteArrayConverter.cs
alexb5dh: So I can do [this](https://github.com/NethermindEth/nethermind/pull/9481/files#diff-88e2214a1cbcfccd95965ddc5b455f53ed30be43c6d9856e4b2e6b90733a5b1fR252):
```
public override byte[] ReadAsPropertyName(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
{
    return Convert(ref reader) ?? throw ThrowInvalidOperationException();
}
```

src/Nethermind/Nethermind.Db.Rocks/MergeOperatorAdapter.cs
alexb5dh: Looks like  `ConcurrentDictionary` is needed, so will loose a bit of performance for synchronization.

src/Nethermind/Nethermind.Db/CompressingDb.cs
alexb5dh: Don't know if direct pass-through will make "compressed" parts valid.
It probably depends on how storage will be used (i.e. will we merge to empty key) and which merge operator is attached.

So would prefer to leave implementation in the hands of whoever will need `Merge` here.

src/Nethermind/Nethermind.Db.Rocks/MergeOperatorAdapter.cs
alexb5dh: I suspect that `ConcurrentDictionary` add/delete (including locking & hashing) + `Pin`/`Free` + probably allocating new object to store both handle and list will make it worse in terms of peformance (not memory usage) than current one with memory allocation & copying, but this needs benchmarks.

Copying here will operate with 4-512 bytes on average.

src/Nethermind/Nethermind.Core/FakeWriteBatch.cs
```diff
@@ -31,5 +31,12 @@ public void Set(ReadOnlySpan<byte> key, byte[]? value, WriteFlags flags = WriteF
         {
             _storePretendingToSupportBatches.Set(key, value, flags);
         }
+
+        public void Merge(ReadOnlySpan<byte> key, ReadOnlySpan<byte> value, WriteFlags flags = WriteFlags.None)
```
asdacap, line 35: I think you dont need this anymore right?

alexb5dh: It implements `IWriteBatch` and I need `Merge` there

src/Nethermind/Nethermind.Db.Rocks/MergeOperatorAdapter.cs
LukaszRozmej: `ConcurrentDictionary` is actually good and doesn't really lock until there is actual concurrent access to same key.
But with such small bytes you are right it is hard to say.

src/Nethermind/Nethermind.Db.Rocks/Config/PerTableDbConfig.cs
```diff
@@ -38,7 +38,7 @@ private void EnsureConfigIsAvailable(string propertyName)
         foreach (var prefix in _prefixes)
         {
             string prefixed = string.Concat(prefix, propertyName);
-            if (type.GetProperty(prefixed, BindingFlags.Public | BindingFlags.Instance) is null)
+            if (GetProperty(type, prefixed) is null)
```
alexb5dh: Changed code will now add `BindingFlags.IgnoreCase` here.
Before, config validation wasn't using this flag. while actual config reading was.

src/Nethermind/Nethermind.Db.Rocks/Config/PerTableDbConfig.cs
alexb5dh: Can `IDbConfig` property-reading be also made case-insensitive?

Somehow ended with log-index DB being named "logIndexStorage" and current checks now require `IDbConfig` to have properties prefixed with "logIndexStorage" case-sensitive. Would prefer to not break already synced nodes by renaming DB to "LogIndexStorage".

src/Nethermind/Nethermind.Db.Rocks/RocksDbIteratorWrapper.cs
alexb5dh: Replaced iterator with `ISortedView`, but needed to add `StartBefore` method doing `iterator.SeekForPrev`.

Also, upper bound (`lastKey`) being exclusive should probably be reflected in parameter naming or method documentation somehow:
```
ISortedView GetViewBetween(ReadOnlySpan<byte> firstKey, ReadOnlySpan<byte> lastKey)
```

src/Nethermind/Nethermind.Blockchain/Receipts/IReceiptStorage.cs
```diff
@@ -19,8 +19,13 @@ public interface IReceiptStorage : IReceiptFinder
         void RemoveReceipts(Block block);
 
         /// <summary>
-        /// Receipts for a block are inserted
+        /// Receipts for a new main block are inserted
```
LukaszRozmej:
```suggestion
        /// Receipts for canonical chain changed.
```

src/Nethermind/Nethermind.Blockchain/Receipts/IReceiptStorage.cs
```diff
@@ -19,8 +19,13 @@ public interface IReceiptStorage : IReceiptFinder
...
         /// </summary>
-        event EventHandler<BlockReplacementEventArgs> ReceiptsInserted;
+        event EventHandler<BlockReplacementEventArgs> NewCanonicalReceipts;
+
+        /// <summary>
+        /// Any receipts are inserted
+        /// </summary>
```
LukaszRozmej, line 28:
```suggestion
        /// <summary>
        /// Receipts for any block are inserted.
        /// </summary>
        /// <remarks>
        /// This is invoked for both canonical and non-canonical blocks.
        /// </remarks>
```

src/Nethermind/Nethermind.Blockchain/Receipts/NullReceiptStorage.cs
```diff
@@ -13,7 +13,8 @@ public class NullReceiptStorage : IReceiptStorage
         public static NullReceiptStorage Instance { get; } = new();
 
 #pragma warning disable CS0067
-        public event EventHandler<BlockReplacementEventArgs> ReceiptsInserted;
+        public event EventHandler<BlockReplacementEventArgs> NewCanonicalReceipts;
```
LukaszRozmej:
```suggestion
        public event EventHandler<BlockReplacementEventArgs>? NewCanonicalReceipts;
```

src/Nethermind/Nethermind.Blockchain/Receipts/PersistentReceiptStorage.cs
```diff
@@ -38,7 +38,8 @@ public class PersistentReceiptStorage : IReceiptStorage
         private const int CacheSize = 64;
         private readonly LruCache<ValueHash256, TxReceipt[]> _receiptsCache = new(CacheSize, CacheSize, "receipts");
 
-        public event EventHandler<BlockReplacementEventArgs> ReceiptsInserted;
+        public event EventHandler<BlockReplacementEventArgs> NewCanonicalReceipts;
```
LukaszRozmej:
```suggestion
        public event EventHandler<BlockReplacementEventArgs>? NewCanonicalReceipts;
```

src/Nethermind/Nethermind.Db.Rocks/Config/PerTableDbConfig.cs
LukaszRozmej: I think yes - @asdacap ?

src/Nethermind/Nethermind.Core/IKeyValueStore.cs
```diff
@@ -69,6 +69,11 @@ public interface IWriteOnlyKeyValueStore
         void Remove(ReadOnlySpan<byte> key) => Set(key, null);
     }
 
+    public interface IMergeableKeyValueStore
```
LukaszRozmej: Shouldn't it be?
```suggestion
    public interface IMergeableKeyValueStore : IKeyValueStore
```

or `ISortedKeyValueStore`?

src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
```diff
@@ -577,6 +583,13 @@ protected virtual void BuildOptions<T>(IRocksDbConfig dbConfig, Options<T> optio
             }
         }
 
+        if (mergeOperator is not null)
+        {
+            options.SetMergeOperator(new MergeOperatorAdapter(mergeOperator));
+            lock (_doNotGcOptionsLocker)
+                _doNotGcOptions.Add(options);
```
LukaszRozmej: I think `_doNotGcOptionsLocker` is redundant, you can just do:


```suggestion
            lock (_doNotGcOptions)
                _doNotGcOptions.Add(options);
```

src/Nethermind/Nethermind.Runner.Test/Ethereum/Steps/Migrations/ReceiptMigrationTests.cs
```diff
@@ -160,7 +160,8 @@ public void RemoveReceipts(Block block)
             }
 
 #pragma warning disable CS0067
-            public event EventHandler<BlockReplacementEventArgs> ReceiptsInserted;
+            public event EventHandler<BlockReplacementEventArgs> NewCanonicalReceipts;
+            public event EventHandler<ReceiptsEventArgs> ReceiptsInserted;
```
LukaszRozmej, line 164:
```suggestion
            public event EventHandler<BlockReplacementEventArgs>? NewCanonicalReceipts;
            public event EventHandler<ReceiptsEventArgs>? ReceiptsInserted;
```

src/Nethermind/Nethermind.Core/IKeyValueStore.cs
alexb5dh: Added `IWriteOnlyKeyValueStore`

src/Nethermind/Nethermind.Runner.Test/Ethereum/Steps/Migrations/ReceiptMigrationTests.cs
alexb5dh: This one complains that nullability is not enabled for the project, should I toggle it on?

LukaszRozmej: Ah we still don't have it for some test projects...
I'm fine with either.

alexb5dh: Nah, enabling it generates too many errors for this PR to resolve :sweat_smile:

LukaszRozmej (approved):

src/Nethermind/Nethermind.Core/IKeyValueStore.cs
```diff
@@ -82,6 +87,7 @@ public interface ISortedKeyValueStore : IKeyValueStore
     /// </summary>
     public interface ISortedView : IDisposable
     {
+        public bool StartBefore(ReadOnlySpan<byte> value);
```
asdacap, line 90: `GetViewBetween` does not work? Or you need the key exactly before `value`?

src/Nethermind/Nethermind.Core/Utils/ConcurrentWriteBatcher.cs
```diff
@@ -40,6 +41,23 @@ public void PutSpan(ReadOnlySpan<byte> key, ReadOnlySpan<byte> value, WriteFlags
...
+    public void Merge(ReadOnlySpan<byte> key, ReadOnlySpan<byte> value, WriteFlags flags = WriteFlags.None)
+    {
+        IWriteBatch currentBatch = RentWriteBatch();
+        currentBatch.Merge(key, value, flags);
+        ReturnWriteBatch(currentBatch);
+    }
+
+    public void Clear()
+    {
+        ThrowIfDisposed();
```
asdacap: Might wanna throw on all case here instead. ConcurrentWriteBatcher straight up cannot be cancelled.

src/Nethermind/Nethermind.Db.Rocks/Config/PerTableDbConfig.cs
asdacap: Fair enough.

src/Nethermind/Nethermind.Db/CompressingDb.cs
asdacap: I'd prefer to not have it be implemented in the first place.

asdacap (approved):
Consider ways to not have to implement Merge on place where that is unnecessary.

src/Nethermind/Nethermind.Db.Rocks/ColumnsDb.cs
```diff
@@ -130,9 +130,19 @@ public void Dispose()
...
+        }
+
         public void Set(ReadOnlySpan<byte> key, byte[]? value, WriteFlags flags = WriteFlags.None)
         {
             _writeBatch._writeBatch.Set(key, value, _column._columnFamily, flags);
         }
+
+        public void Merge(ReadOnlySpan<byte> key, ReadOnlySpan<byte> value, WriteFlags flags = WriteFlags.None)
+        {
+            _writeBatch._writeBatch.Merge(key, value, flags);
```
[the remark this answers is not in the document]
benaadams: @alexb5dh @asdacap valid issue?

src/Nethermind/Nethermind.Db/CompressingDb.cs
alexb5dh: `StartWriteBatch` returns `IWriteBatch`, so the only option I see is to move interface inheritance to specific classes and convert result to `IMergeableKeyValueStore` (or throw an error otherwise). This is how I'm using `ISortedKeyValueStore` as of now.

src/Nethermind/Nethermind.Db.Rocks/ColumnsDb.cs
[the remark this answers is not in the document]
alexb5dh: Yes. Didn't notice this as was using `ColumnsDbWriteBatch` instance.

src/Nethermind/Nethermind.Core/IKeyValueStore.cs
alexb5dh: Yes, need to start exactly before so lower bound may not work as a starting point. Still prefer to have lower bound present - for example, for correct check on iterator validity. Plus, if key before is out of bounds - need to start at the lower bound.

src/Nethermind/Nethermind.Db/CompressingDb.cs
alexb5dh: If approach [like this](https://github.com/NethermindEth/nethermind/pull/8464/files#diff-8ab45f3e34fe34d810e6db3df6b31c545706a8b0b914f2405b840c2e659a7ebbR546-R548) is fine, will move `IMergeableKeyValueStore` interface inheritance to concrete instances:
```
IDb db = GetDb(topicIndex);
ISortedKeyValueStore? sortedDb = db as ISortedKeyValueStore
  ?? throw new NotSupportedException($"{db.GetType().Name} DB does not support sorted lookups.");
```

src/Nethermind/Nethermind.Core/Collections/DictionaryExtensions.cs
```diff
@@ -13,4 +14,15 @@ public static void Increment<TKey>(this Dictionary<TKey, int> dictionary, TKey k
         ref int res = ref CollectionsMarshal.GetValueRefOrAddDefault(dictionary, key, out bool _);
         res++;
     }
+
+    public static TValue GetOrAdd<TKey, TValue>(this Dictionary<TKey, TValue> dictionary, TKey key, Func<TKey, TValue> factory)
```
benaadams: I'd `out` param whether is get or add (e.g. return `out bool exists`)

src/Nethermind/Nethermind.Core/JsonConverters/ByteArrayConverter.cs
```diff
@@ -250,4 +250,14 @@ public static void Convert(
         if (array is not null)
             ArrayPool<byte>.Shared.Return(array);
     }
+
+    public override byte[] ReadAsPropertyName(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+    {
+        return Convert(ref reader) ?? throw ThrowInvalidOperationException();
```
benaadams: Change to `if`; as already would have thrown at this point from the helper; so this isn't throwing anything; so doesn't make much implementation sense other than the syntax sugar

src/Nethermind/Nethermind.Core/JsonConverters/ByteArrayConverter.cs
```diff
@@ -177,7 +177,7 @@ public static void Convert(ref Utf8JsonReader reader, scoped Span<byte> span)
     private static void ThrowFormatException() => throw new FormatException();
 
     [DoesNotReturn, StackTraceHidden]
-    private static void ThrowInvalidOperationException() => throw new InvalidOperationException();
+    private static Exception ThrowInvalidOperationException() => throw new InvalidOperationException();
```
benaadams: Revert

src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
```diff
@@ -1328,6 +1403,18 @@ private void InnerFlush(bool onlyWal)
...
 
+    private void InnerFlush(ColumnFamilyHandle columnFamilyHandle)
+    {
+        try
+        {
+            _rocksDbNative.rocksdb_flush_cf(_db.Handle, FlushOptions.DefaultFlushOptions.Handle, columnFamilyHandle.Handle);
+        }
+        catch (RocksDbSharpException e)
+        {
+            CreateMarkerIfCorrupt(e);
```
benaadams, line 1414: Should rethrow? Otherwise continuing with a corrupt db

src/Nethermind/Nethermind.Core/JsonConverters/ByteArrayConverter.cs
alexb5dh: Yeah, syntax sugar was the idea here - 1 line instead of 3-4, but ok.

src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
alexb5dh: This is almost identical to old `InnerFlush` above, which doesn't throw, not sure if it was intended or not.

@asdacap, as it [were your changes](https://github.com/NethermindEth/nethermind/commit/664a3453e0edf5f6ecd96a9ed017a0cdd8e679c2#diff-37e7eb8b529265d0cd0f25ad1f5eed170ca123883a8dc9ba13d1cef5a2409549R657), should `throw` be added here and above?

benaadams: Ah, ok so we terminate the process when calling `CreateMarkerIfCorrupt` with `Environment.Exit` so technically doesn't matter

PR #9395, opened 2025-10-02
Enable Osaka on Chiado

Marchhill:
## Changes

- Enable Osaka fork on Chiado testnet
- Test refactoring

#### What types of changes does your code introduce?

- [ ] Bugfix (a non-breaking change that fixes an issue)
- [x] New feature (a non-breaking change that adds functionality)
- [ ] Breaking change (a change that causes existing functionality not to work as expected)
- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

#### If yes, did you write tests?

- [x] Yes
- [ ] No

#### Requires documentation update

- [ ] Yes
- [x] No

#### Requires explanation in Release Notes

- [x] Yes
- [ ] No

PR #9484, opened 2025-10-16
Revert "Enable Osaka on Chiado"

stdevMac:
Reverts NethermindEth/nethermind#9395


Marchhill:
Instead of reverting everything could we just remove from chainspec and comment out tests? There are some other refactoring changes reverted and stuff like `OsakaGnosis` removed which doesn't need to be

stdevMac:
> Instead of reverting everything could we just remove from chainspec and comment out tests?

Fine for me, I see it cleaner as a revert, but I am ok with that. In the end we will need to add it back eventually. Could you please take care of it?

PR #9488, opened 2025-10-17
Fix genesis post processor have no allocations

asdacap:
- Fix genesis post processor have no access to allocations due to the mis-ordering of chain spec allocations clearing.

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- reproducable locally

### DIFF
--- a/src/Nethermind/Nethermind.Era1.Test/Era1ModuleTests.cs
+++ b/src/Nethermind/Nethermind.Era1.Test/Era1ModuleTests.cs
@@ -221,7 +221,7 @@ public class Era1ModuleTests
         TestBlockchain testBlockchain = await BasicTestBlockchain.Create(configurer: builder =>
         {
             builder.AddScoped<IGenesisPostProcessor, IWorldState, ISpecProvider>((worldState, specProvider) =>
-                new FunctionalGenesisPostProcessor(worldState, (block, state) =>
+                new FunctionalGenesisPostProcessor((block) =>
                 {
                     worldState.AddToBalance(TestItem.AddressA, 10.Ether(), specProvider.GenesisSpec);
                     worldState.RecalculateStateRoot();


src/Nethermind/Nethermind.Core.Test/Container/FunctionalGenesisPostProcessor.cs
```diff
@@ -3,14 +3,13 @@
...
 
 namespace Nethermind.Core.Test.Container;
 
-public class FunctionalGenesisPostProcessor(IWorldState state, Action<Block, IWorldState> postProcessor) : IGenesisPostProcessor
+public class FunctionalGenesisPostProcessor(Action<Block> postProcessor) : IGenesisPostProcessor
 {
     public void PostProcess(Block genesis)
     {
-        postProcessor.Invoke(genesis, state);
+        postProcessor.Invoke(genesis);
```
LukaszRozmej: I always wonder why you use `Invoke` rather than: 

```suggestion
        postProcessor(genesis);
```

asdacap: Me too

src/Nethermind/Nethermind.Blockchain.Test/GenesisBuilderTests.cs
```diff
@@ -2,16 +2,17 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 
 using System.IO;
+using FluentAssertions;
```
rubo, line 5: Generally, I'd recommend avoiding a new code with FluentAssertions, as we're stuck with its old version.

src/Nethermind/Nethermind.Core.Test/Container/FunctionalGenesisPostProcessor.cs
LukaszRozmej: And yet you still do...

PR #9489, opened 2025-10-17
Improve Hive RPC compat.

smartprogrammer93:
should fix 11 RPC Compat hive tests:
```
debug_getRawBlock/get-invalid-number (nethermind)
debug_getRawHeader/get-invalid-number (nethermind)
debug_getRawReceipts/get-invalid-number (nethermind)
debug_getRawTransaction/get-invalid-hash (nethermind)
debug_getRawTransaction/get-tx (nethermind)
eth_getBlockReceipts/get-block-receipts-by-hash (nethermind)
eth_getBlockReceipts/get-block-receipts-n (nethermind)
eth_getProof/get-account-proof-with-storage (nethermind)
eth_getTransactionReceipt/get-legacy-contract (nethermind)
eth_getTransactionReceipt/get-legacy-input (nethermind)
eth_getTransactionReceipt/get-legacy-receipt (nethermind)
```

hive test:

image

## Changes

- added a new flag that when set makes parsing more restrictive. default value set to true
- when flag is set
  - stop reading numbers not in strings
  - stop reading numbers not in hex format and starting with `0x`
  - stop reading hashes not starting with `0x`
- in Receipts, stop returning status when root is not null
- remove leading zeros from some places
- adds 0x to start of raw tx
- update `known-failing-hive-tests.txt` to remove the passing tests

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

### DIFF
--- a/scripts/known-failing-hive-tests.txt
+++ b/scripts/known-failing-hive-tests.txt
@@ -1,18 +1,7 @@
 # rpc-compat
 
-debug_getRawBlock/get-invalid-number (nethermind)
-debug_getRawHeader/get-invalid-number (nethermind)
-debug_getRawReceipts/get-invalid-number (nethermind)
-debug_getRawTransaction/get-invalid-hash (nethermind)
-debug_getRawTransaction/get-tx (nethermind)
-eth_getBlockReceipts/get-block-receipts-by-hash (nethermind)
-eth_getBlockReceipts/get-block-receipts-n (nethermind)
-eth_getProof/get-account-proof-with-storage (nethermind)
 eth_getStorageAt/get-storage-invalid-key (nethermind)
 eth_getStorageAt/get-storage-invalid-key-too-large (nethermind)
-eth_getTransactionReceipt/get-legacy-contract (nethermind)
-eth_getTransactionReceipt/get-legacy-input (nethermind)
-eth_getTransactionReceipt/get-legacy-receipt (nethermind)
 eth_sendRawTransaction/send-blob-tx (nethermind)
 eth_simulateV1/ethSimulate-empty-with-block-num-set-plus1 (nethermind)
 eth_simulateV1/ethSimulate-eth-send-should-not-produce-logs-on-revert (nethermind)


flcl42:
Can it be by default?

MarekM25 (changes requested):
IMO, this is weird. Receipts definitely need to be fixed, we shouldn’t have flags or different behaviors just for tests.

LukaszRozmej (changes requested):
Receipt.Status is done wrong. It should just be nullable and if status is not present in receipt just skip it. (there is a proper json attribute for null handling). BTW same with Root as it is one or the other!
As for parsing:
I would rename it to: `StrictHexFormat` (so flip the logic)
Make true by default
Add description in release notes.
If someone wants higher compatibility RPC he can flip the `StrictHexFormat` off. If people won't be doing that we can deprecate it and remove in few versions. 

Also modify the failing hive test list

src/Nethermind/Nethermind.JsonRpc/IJsonRpcConfig.cs
```diff
@@ -187,4 +187,9 @@ public interface IJsonRpcConfig : IConfig
 
     [ConfigItem(Description = "Preload rpc modules. Useful in rpc provider to reduce latency on first request.", DefaultValue = "false")]
     bool PreloadRpcModules { get; set; }
+
+    [ConfigItem(
+        Description = "Follow the exact standardization rules for hash and block param. this will decrease compatibility with some requests",
```
LukaszRozmej: @rubo on description

src/Nethermind/Nethermind.Serialization.Json/EthereumJsonSerializer.cs
```diff
@@ -113,6 +115,20 @@ public static void AddConverter(JsonConverter converter)
             JsonOptionsIndented = CreateOptions(indented: true);
         }
 
+        private static bool _strictHexFormat;
+        public static bool StrictHexFormat
```
LukaszRozmej, line 119: I wonder can we make it not static?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Serialization.Json/EthereumJsonSerializer.cs
smartprogrammer93: tried. would be a lot of changes because the options generators are used statically in a lot of places and the creation of the option depends on this flag

src/Nethermind/Nethermind.JsonRpc/IJsonRpcConfig.cs
smartprogrammer93: could be better maybe. I can say

```suggestion
        Description = "Enable strict parsing rules for Block Params and Hashas in RPC requests. this will decrease compatibility but increase compliance with the spec.",
```

PR #9501, opened 2025-10-19
Add back partition by key

asdacap:
- Add back partitioning of triestore by statetree subtree.
- This ensure that when a partition was pruned due to memory limitation, whole subtree is removed which prevent cases where top level node was not deleted but inner node was, which would cause random trie exception.
- Fix arb.
- Unfortunately this will make pruning slower due to whole contract in a shard and some contract is big. This can be mitigated a little bit by increasing the number of shard, like `--Pruning.DirtyNodeShardBit 12`. But not much. Alternatively, prune persisted node can be made to priority large shard. But the effect is still limited.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- [x] Arbitrum can run overnight
- [x] Mainnet can sync
- [ ] Mainnet era replay
- [x] Mainnet full prune

### DIFF
--- a/src/Nethermind/Nethermind.Trie.Test/Pruning/TreeStoreTests.cs
+++ b/src/Nethermind/Nethermind.Trie.Test/Pruning/TreeStoreTests.cs
@@ -1206,11 +1206,11 @@ namespace Nethermind.Trie.Test.Pruning
             fullTrieStore.WaitForPruning();
 
             fullTrieStore.PrunePersistedNodes();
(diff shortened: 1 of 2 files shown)


asdacap:
Testing #9555 as this really make pruning very long.

PR #9511, opened 2025-10-20
fix: CompositeTxTracer aggregates IsTracingLogs and gates ReportLog correctly

Forostovec:
This change aligns CompositeTxTracer with the ITxTracer contract and VM call sites by aggregating IsTracingLogs in the constructor and gating ReportLog by IsTracingLogs instead of IsTracingInstructions, which previously caused logs to be suppressed even when inner tracers enabled logging; also removed the unused public ReportOperationLogs method that is not part of the interface.


deffrian (approved):
I have no idea why it's that way

PR #9512, opened 2025-10-20
XDC block producer

ak88:
Block builder for XDC

### DIFF
--- a/src/Nethermind/Nethermind.Xdc/XdcBlockProducer.cs
+++ b/src/Nethermind/Nethermind.Xdc/XdcBlockProducer.cs
@@ -83,7 +83,7 @@ internal class XdcBlockProducer : BlockProducerBase
 
         xdcBlockHeader.MixHash = Hash256.Zero;
 
-        if (epochSwitchManager.IsEpochSwitch(xdcBlockHeader, out _))
+        if (epochSwitchManager.IsEpochSwitchAtBlock(xdcBlockHeader))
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Xdc/XdcBlockProducer.cs
```diff
@@ -0,0 +1,102 @@
...
+
+            xdcBlockHeader.Penalties = new byte[penalties.Length * Address.Size];
+
+            for (int i = 0; i < penalties.Length; i++)
+            {
+                Array.Copy(penalties[i].Bytes, 0, xdcBlockHeader.Penalties, i * Address.Size, Address.Size);
+            }
+        }
+        return xdcBlockHeader;
+    }
```
LukaszRozmej: Would it make sense to:
Have some base or helper method taking in generic for actual header type.
Reusing base implementation to fill the basic stuff.
Only ammend custom stuff in this method?

ak88: In this case there is very little work saved, since the constructor will set most of the fields and other fields has to be overridden in the XDC version.

LukaszRozmej: To be honest I'm more concern about future changes that could be auto-applied for XDC and here will have to be manually ported

PR #9513, opened 2025-10-20
fix/crypto-random-secure-rng

sashaodessa:
## Changes

- Replace `System.Random` with `RandomNumberGenerator.GetInt32()` in `CryptoRandom.NextInt()`
- Remove unused `Random` field from `CryptoRandom` class
- Add compatibility handling for non-positive max values

### DIFF
--- a/src/Nethermind/Nethermind.Crypto/CryptoRandom.cs
+++ b/src/Nethermind/Nethermind.Crypto/CryptoRandom.cs
@@ -24,7 +24,6 @@ namespace Nethermind.Crypto
             return bytes;
         }
 
-        [RequiresSecurityReview("There should be no unsecured method in a class that suggests security")]
         public int NextInt(int max)
         {
             // Use cryptographically secure RNG; preserve Random.Next behavior for non-positive max


src/Nethermind/Nethermind.Crypto/CryptoRandom.cs
```diff
@@ -28,7 +27,14 @@ public byte[] GenerateRandomBytes(int length)
...
         {
-            return _random.Next(max);
+            // Use cryptographically secure RNG; preserve Random.Next behavior for non-positive max
+            // (Random.Next(0) returns 0; negatives throw). This keeps compatibility.
+            if (max <= 0)
+            {
+                return 0;
+            }
+
+            return RandomNumberGenerator.GetInt32(max);
```
LukaszRozmej: Can be written simpler:
```suggestion
            return max <= 0 ? 0 : RandomNumberGenerator.GetInt32(max);
```

sashaodessa: @LukaszRozmej updated

src/Nethermind/Nethermind.Crypto/CryptoRandom.cs
```diff
@@ -28,7 +27,9 @@ public byte[] GenerateRandomBytes(int length)
         [RequiresSecurityReview("There should be no unsecured method in a class that suggests security")]
```
LukaszRozmej: this needs removing
```suggestion
```

PR #9520, opened 2025-10-21
Slight logging change

asdacap:
- Slight logging change in block downloader.
- Increase the sync report log frequency. 

#### Requires testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

#### Notes on testing

- Seems to work.

### DIFF
--- a/src/Nethermind/Nethermind.Synchronization/ParallelSync/MultiSyncModeSelector.cs
+++ b/src/Nethermind/Nethermind.Synchronization/ParallelSync/MultiSyncModeSelector.cs
@@ -420,8 +420,7 @@ namespace Nethermind.Synchronization.ParallelSync
                     (nameof(postPivotPeerAvailable), postPivotPeerAvailable),
                     (nameof(notReachedFullSyncTransition), notReachedFullSyncTransition),
                     (nameof(notInAStickyFullSync), notInAStickyFullSync),
-                    (nameof(stateNotDownloadedYet), stateNotDownloadedYet),
-                    (nameof(longRangeCatchUp), longRangeCatchUp),
+                    ($"{nameof(stateNotDownloadedYet)}||${longRangeCatchUp}", stateNotDownloadedYet || longRangeCatchUp),
                     (nameof(notNeedToWaitForHeaders), notNeedToWaitForHeaders));
             }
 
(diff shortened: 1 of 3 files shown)


src/Nethermind/Nethermind.Synchronization/ParallelSync/MultiSyncModeSelector.cs
```diff
@@ -420,8 +420,7 @@ private bool ShouldBeInFastSyncMode(Snapshot best)
                     (nameof(postPivotPeerAvailable), postPivotPeerAvailable),
                     (nameof(notReachedFullSyncTransition), notReachedFullSyncTransition),
                     (nameof(notInAStickyFullSync), notInAStickyFullSync),
-                    (nameof(stateNotDownloadedYet), stateNotDownloadedYet),
-                    (nameof(longRangeCatchUp), longRangeCatchUp),
+                    ($"{nameof(stateNotDownloadedYet)}||${longRangeCatchUp}", stateNotDownloadedYet || longRangeCatchUp),
```
LukaszRozmej, line 423:
```suggestion
                    ($"{nameof(stateNotDownloadedYet)} || ${longRangeCatchUp}", stateNotDownloadedYet || longRangeCatchUp),
```

src/Nethermind/Nethermind.Synchronization/ParallelSync/MultiSyncModeSelector.cs
```diff
@@ -608,8 +607,7 @@ private bool ShouldBeInStateSyncMode(Snapshot best)
                     (nameof(hasFastSyncBeenActive), hasFastSyncBeenActive),
                     (nameof(hasAnyPostPivotPeer), hasAnyPostPivotPeer),
                     ($"{nameof(notInFastSync)}||{nameof(stickyStateNodes)}", notInFastSync || stickyStateNodes),
-                    (nameof(stateNotDownloadedYet), stateNotDownloadedYet),
-                    (nameof(longRangeCatchUp), longRangeCatchUp),
+                    ($"{nameof(stateNotDownloadedYet)}||{nameof(longRangeCatchUp)}", stateNotDownloadedYet || longRangeCatchUp),
```
LukaszRozmej, line 610:
```suggestion
                    ($"{nameof(stateNotDownloadedYet)} || {nameof(longRangeCatchUp)}", stateNotDownloadedYet || longRangeCatchUp),
```

change the one above too?

src/Nethermind/Nethermind.Synchronization/Reporting/SyncReport.cs
```diff
@@ -28,7 +28,7 @@ public class SyncReport : ISyncReport
         private const int NoProgressStateSyncReportFrequency = 30;
         private const int SyncAllocatedPeersReportFrequency = 30;
         private const int SyncFullPeersReportFrequency = 120;
-        private static readonly TimeSpan _defaultReportingIntervals = TimeSpan.FromSeconds(10);
+        private static readonly TimeSpan _defaultReportingIntervals = TimeSpan.FromSeconds(1);
```
LukaszRozmej, line 31: Do we want to spam the users that much?

LukaszRozmej: Maybe 1 in DEBUG compilation and 10 in RELEASE?

asdacap: The log that shows up every second only show up on debug. So the only log that show up here is the one run every 15 cycle which is the peer count.

src/Nethermind/Nethermind.Synchronization/ParallelSync/MultiSyncModeSelector.cs
asdacap: I'm just keeping in consistent with the original format, see the line before.

LukaszRozmej: Yeah I know, I don't like the original format :D

PR #9521, opened 2025-10-21
Xdc Pool for timeouts and votes

cicr99:


src/Nethermind/Nethermind.Xdc/XdcPool.cs
```diff
@@ -0,0 +1,81 @@
...
+                list.Add(item);
+            return list.Count;
+        }
+    }
+
+    public void EndRound(ulong round)
+    {
+        using var lockRelease = _lock.Acquire();
+        {
+            foreach (var key in _items.Keys.ToArray())
```
ak88: you dont need to allocate a new array

cicr99: Since I was removing elements while iterating through the dictionary, I thought it would be safest this way. But you are right, I looked it up, and `Remove` method can be safely called without invalidating active enumerators of the dictionary instance. Nice catch!

src/Nethermind/Nethermind.Xdc/XdcPool.cs
```diff
@@ -0,0 +1,81 @@
...
+using System.Linq;
+using Nethermind.Core.Collections;
+using Nethermind.Core.Crypto;
+using Nethermind.Core.Threading;
+
+namespace Nethermind.Xdc;
+
+public class XdcPool<T> where T : IXdcPoolItem
+{
+    private readonly Dictionary<(ulong Round, Hash256 Hash), ArrayPoolList<T>> _items = new();
```
LukaszRozmej, line 16: If used ConcurrentDictionary you could go lockless if needed and get better performance on multithread access

cicr99: Yes, there are some optimizations we could do here. We've discussed it and for now, we'll keep it like this. Even with a ConcurrentDictionary, operations inside the ArrayPoolList still need coordination. We would like to prioritize having the consensus tests in place, before making any modifications

PR #9523, opened 2025-10-21
fix: Remove duplicate HexConvert utility in SendBlobs tool

anim001k:
## Description

The `tools/SendBlobs/HexConvert.cs` utility duplicated existing functionality in the codebase:
- `ToUInt64()` used `Convert.ToUInt64()` with basic hex/decimal detection
- `ToUInt256()` was implemented via `ToUInt64()` cast, causing potential overflow for values > 64-bit

### Solution
- Replaced all `HexConvert.ToUInt64()` calls with `Bytes.FromHexString().AsSpan().ReadEthUInt64()` for hex parsing or `Convert.ToUInt64()` for decimal strings
- Replaced all `HexConvert.ToUInt256()` calls with `new UInt256(Bytes.FromHexString())` using the standard parsing path
- Added necessary `using Nethermind.Core.Extensions;` imports
- Removed the duplicate `HexConvert.cs` file entirely

### DIFF
--- a/tools/SendBlobs/SetupCli.cs
+++ b/tools/SendBlobs/SetupCli.cs
@@ -10,6 +10,7 @@ using Nethermind.Specs.Forks;
 using Nethermind.JsonRpc.Client;
 using Nethermind.Serialization.Json;
 using Nethermind.Specs;
+using Nethermind.Core.Extensions;
 
 namespace SendBlobs;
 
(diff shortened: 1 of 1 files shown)


LukaszRozmej (changes requested):
Before: one simple call
After: complex calls and conditional expressions

Doesn't seem an upgrade for me

PR #9529, opened 2025-10-21
Fix eth_sumulate defaults

deffrian:
Fixes Closes Resolves #

```
eth_simulateV1/ethSimulate-simple-more-params-validate (nethermind)
eth_estimateGas/estimate-with-eip4844 (nethermind)
```

## Changes

- Adjust tx type to geth

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

### DIFF
--- a/src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
+++ b/src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
@@ -142,8 +142,11 @@ public abstract class TransactionForRpc
             if (untyped.TryGetPropertyValue(typeFieldKey, out JsonNode? node))
             {
                 TxType? setType = node.Deserialize<TxType?>(options);
-                return _txTypes.FirstOrDefault(p => p.TxType == setType)?.Type ??
-                       throw new JsonException("Unknown transaction type");
+                if (setType is not null)
+                {
+                    return _txTypes.FirstOrDefault(p => p.TxType == setType)?.Type ??
+                           throw new JsonException("Unknown transaction type");
+                }
             }
 
             if (untyped.ContainsKey(gasPriceFieldKey))
--- a/src/Nethermind/Nethermind.JsonRpc.Test/Data/Eip2930Tests.cs
+++ b/src/Nethermind/Nethermind.JsonRpc.Test/Data/Eip2930Tests.cs
@@ -98,7 +98,7 @@ public class Eip2930Tests
     [Test]
     public void can_deserialize_null_accessList()
     {
-        string json = """{"nonce":"0x0","blockHash":null,"blockNumber":null,"transactionIndex":null,"to":null,"value":"0x0","gasPrice":"0x0","gas":"0x0","input":null,"accessList":null}""";
+        string json = """{"nonce":"0x0","blockHash":null,"blockNumber":null,"transactionIndex":null,"to":null,"value":"0x0","gas":"0x0","input":null,"accessList":null}""";
 
         var transactionForRpc = _serializer.Deserialize<TransactionForRpc>(json);
 
@@ -138,7 +138,7 @@ public class Eip2930Tests
     [Test]
     public void can_deserialize_empty_accessList()
     {
-        string json = """{"nonce":"0x0","blockHash":null,"blockNumber":null,"transactionIndex":null,"to":null,"value":"0x0","gasPrice":"0x0","gas":"0x0","input":null,"accessList":[]}""";
+        string json = """{"nonce":"0x0","blockHash":null,"blockNumber":null,"transactionIndex":null,"to":null,"value":"0x0","gas":"0x0","input":null,"accessList":[]}""";
         TransactionForRpc transactionForRpc = _serializer.Deserialize<TransactionForRpc>(json);
 
         transactionForRpc.Type.Should().Be(TxType.AccessList);
@@ -172,7 +172,7 @@ public class Eip2930Tests
     [Test]
     public void can_deserialize_accessList_with_empty_storageKeys()
     {
-        string json = """{"nonce":"0x0","blockHash":null,"blockNumber":null,"transactionIndex":null,"to":null,"value":"0x0","gasPrice":"0x0","gas":"0x0","input":null,"accessList":[{"address":"0xb7705ae4c6f81b66cdb323c65f4e8133690fc099","storageKeys":[]}]}""";
+        string json = """{"nonce":"0x0","blockHash":null,"blockNumber":null,"transactionIndex":null,"to":null,"value":"0x0","gas":"0x0","input":null,"accessList":[{"address":"0xb7705ae4c6f81b66cdb323c65f4e8133690fc099","storageKeys":[]}]}""";
         TransactionForRpc transactionForRpc = _serializer.Deserialize<TransactionForRpc>(json);
 
         transactionForRpc.Type.Should().Be(TxType.AccessList);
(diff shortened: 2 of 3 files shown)


src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/BlobTransactionForRpc.cs
```diff
@@ -15,7 +15,6 @@ public class BlobTransactionForRpc : EIP1559TransactionForRpc, IFromTransaction<
     public override TxType? Type => TxType;
 
     [JsonIgnore(Condition = JsonIgnoreCondition.Never)]
-    [JsonDiscriminator]
```
LukaszRozmej, line 18: We want to keep that in for `eth_call` for example

deffrian: Do we? If we want compatibility with geth these rules should apply for all rpc calls

deffrian: Actually this UpdateTxType should be applied to every transaction that comes from rpc

LukaszRozmej: Really? @emlautarom1 @MarekM25 WDYT?

deffrian: They use different `defaultType` for different methods
```
func (args *TransactionArgs) ToTransaction(defaultType int) *types.Transaction {
	usedType := types.LegacyTxType
	switch {
	case args.AuthorizationList != nil || defaultType == types.SetCodeTxType:
		usedType = types.SetCodeTxType
	case args.BlobHashes != nil || defaultType == types.BlobTxType:
		usedType = types.BlobTxType
	case args.MaxFeePerGas != nil || defaultType == types.DynamicFeeTxType:
		usedType = types.DynamicFeeTxType
	case args.AccessList != nil || defaultType == types.AccessListTxType:
		usedType = types.AccessListTxType
	}
	// Make it possible to default to newer tx, but use legacy if gasprice is provided
	if args.GasPrice != nil {
		usedType = types.LegacyTxType
	}
```

emlautarom1: Not sure here, I stand by my comment from back then: https://github.com/NethermindEth/nethermind/pull/8081#pullrequestreview-2567277129

Defaulting will always introduce problems. It's not clear to me what we need here: do we want defaulting on all RPC methods or do we need to specialize some methods?

LukaszRozmej: Let's discuss in json-rpc channel in discord with other teams

deffrian: I'm not sure what we should discuss. Removing default type?

LukaszRozmej: how default tx types should be handled in RPC

LukaszRozmej (approved):

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
```diff
@@ -123,27 +123,29 @@ internal static void RegisterTransactionType<T>() where T : TransactionForRpc, I
...
             JsonObject untyped = JsonSerializer.Deserialize<JsonObject>(ref txTypeReader, options);
 
-            TxType? txType = null;
-            if (untyped.TryGetPropertyValue(TypeFieldKey, out JsonNode? node))
+            Type concreteTxType = DeriveTxType(untyped, options);
+
+            return (TransactionForRpc?)JsonSerializer.Deserialize(ref reader, concreteTxType, options);
+        }
+
+        private Type DeriveTxType(JsonObject untyped, JsonSerializerOptions options)
```
LukaszRozmej, line 136:
```suggestion
        private Type DeriveTxType(JsonObject untyped)
```

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
```diff
@@ -123,27 +123,29 @@ internal static void RegisterTransactionType<T>() where T : TransactionForRpc, I
...
-                    txType is not null
-                        ? _txTypes.FirstOrDefault(p => p.TxType == txType)
-                        : _txTypes.FirstOrDefault(p => p.DiscriminatorProperties.Any(name => untyped.ContainsKey(name)), _txTypes[^1])
-                )?.Type
-                ?? throw new JsonException("Unknown transaction type");
-
-            return (TransactionForRpc?)JsonSerializer.Deserialize(ref reader, concreteTxType, options);
+            return _txTypes
+                .FirstOrDefault(p => p.TxType == defaultTxType || p.DiscriminatorProperties.Any(untyped.ContainsKey),
+                    _txTypes[^1]).Type;
```
LukaszRozmej: I don't get it...

Won't this often default to `defaultTxType` and it is based on order?

I would expect something like below:

```suggestion
            return _txTypes.FirstOrDefault(p => p.DiscriminatorProperties.Any(untyped.ContainsKey))?.Type ?? defaultTxType;
```

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
```diff
@@ -123,27 +123,36 @@ internal static void RegisterTransactionType<T>() where T : TransactionForRpc, I
...
-                (
-                    txType is not null
-                        ? _txTypes.FirstOrDefault(p => p.TxType == txType)
-                        : _txTypes.FirstOrDefault(p => p.DiscriminatorProperties.Any(name => untyped.ContainsKey(name)), _txTypes[^1])
-                )?.Type
-                ?? throw new JsonException("Unknown transaction type");
+            if (untyped.ContainsKey(gasPriceFieldKey))
+            {
+                return typeof(LegacyTransactionForRpc);
+            }
```
LukaszRozmej, line 155: If Legacy is first item in `_txTypes` we can probably drop this `if`?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/TransactionForRpc.cs
deffrian: I think it's last. So we should return Legacy if gasPrice is set. Even if we have blobHashes

LukaszRozmej: Ok we have:

`// Adding in reverse order so newer tx types are in priority`

Interesting, maybe we should do that ignoring index?

So:

`int indexOfPreviousTxType = _txTypes.FindIndex(t => t.TxType != TxType.Legacy && t.TxType < typeInfo.TxType);` ?

deffrian: I don't understand what's `typeInfo.TxType`. I think now it's very straight-forward

PR #9538, opened 2025-10-22
Fix Leading Zero Issue for taiko

dipkakwani:
Fix:
Taiko client driver started hitting error with the latest master:
```
Error while processing BatchProposed events, keep retrying error="failed to check whether L1 chain has been reorged: json: cannot unmarshal hex string of odd length into Go struct field L1Origin.l2BlockHash of type common.Hash"
```

This is due to `l2BlockHash` returned without leading 0:
```
curl -X POST \
  -d '{"jsonrpc":"2.0","method":"taiko_l1OriginByID","params":["0x6"],"id":1}' \
  http://localhost:8547/
{"jsonrpc":"2.0","result":{"blockId":"0x6","l2BlockHash":"0x35a48c5b3ee5b1b2a365fcd1aa68c738d1c06474578087a78fa79dd45de6214","l1BlockHeight":"0x14d7ac","l1BlockHash":"0x12daf7e4b06ca2d3a82c775d9e9ad0c973545a608684146cda0df5f7d71188a5","buildPayloadArgsId":[2,253,63,135,169,239,238,214],"isPreconfBlock":false},"id":1}
```
`eth_getBlockByNumber` correctly returned the value with leading 0, but `taiko_l1OriginByID` was returning without leading 0.

## Changes
- Use `Hash256` instead of `ValueHash256`

## Types of changes

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

Tested locally with taiko client driver.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

#### Remarks:
Might need more testing as the original changes to make it true was introduced here: https://github.com/NethermindEth/nethermind/pull/9489/files#diff-b60f4d8c256b3344279ec90e115f210c3a9724d23bcd151cd01d2f3cd19a9c75

### DIFF
--- a/src/Nethermind/Nethermind.Consensus.Test/GenesisLoaderTests.cs
+++ b/src/Nethermind/Nethermind.Consensus.Test/GenesisLoaderTests.cs
@@ -1,0 +1,217 @@
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System;
+using System.Threading;
+using FluentAssertions;
+using Nethermind.Blockchain;
+using Nethermind.Consensus.Processing;
+using Nethermind.Core;
+using Nethermind.Core.Test.Builders;
+using Nethermind.Evm.State;
+using Nethermind.Logging;
+using Nethermind.State;
+using NSubstitute;
+using NUnit.Framework;
+
+namespace Nethermind.Consensus.Test;
+
+[TestFixture]
+public class GenesisLoaderTests
+{
+    [Test]
+    public void Load_ShouldFlushCacheAfterSuccessfulGenesisProcessing()
+    {
+        // Arrange
+        Block genesisBlock = Build.A.Block.Genesis.TestObject;
+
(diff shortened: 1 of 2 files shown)


flcl42:
Looks like ValueHash256 serialization bug? Maybe we should fix it instead

LukaszRozmej:
yes lets unify please!

dipkakwani:
Taiko driver is using go-ethereum's hexutil package, which does not allow odd length strings: https://github.com/ethereum/go-ethereum/blob/master/common/hexutil/json.go#L390

Though at the same time, Geth allows odd length hashes for the standard Ethereum APIs: https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/api.go#L446

This is conflicting code within Geth, so it has be changed in either taiko-geth or geth.

@smartprogrammer93 @LukaszRozmej

PR #9544, opened 2025-10-22
Fix: Include exception details in SimpleConsoleLogger.Error()

prestoalvarez:
## Description

The `Error` method in `SimpleConsoleLogger` was ignoring the `ex` parameter, causing exception details to be lost in error logs.
This improves debugging capabilities by ensuring exception stack traces and messages are properly logged when available.

### DIFF
--- a/src/Nethermind/Nethermind.Logging/SimpleConsoleLogger.cs
+++ b/src/Nethermind/Nethermind.Logging/SimpleConsoleLogger.cs
@@ -34,7 +34,7 @@ namespace Nethermind.Logging
 
         public void Error(string text, Exception ex = null)
         {
-            if (IsError) Console.Error.WriteLine(DateTime.Now.ToString(dateFormat) + text + (ex != null ? " " + ex : string.Empty));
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Logging/SimpleConsoleLogger.cs
```diff
@@ -34,7 +34,7 @@ public void Trace(string text)
 
         public void Error(string text, Exception ex = null)
         {
-            if (IsError) Console.Error.WriteLine(DateTime.Now.ToString(dateFormat) + text);
+            if (IsError) Console.Error.WriteLine(DateTime.Now.ToString(dateFormat) + text + (ex != null ? " " + ex : string.Empty));
```
LukaszRozmej:
```suggestion
            if (IsError) Console.Error.WriteLine(DateTime.Now.ToString(dateFormat) + text + (ex is not null ? " " + ex : string.Empty));
```

LukaszRozmej (approved):

PR #9553, opened 2025-10-23
Remove redundant Interlocked.MemoryBarrier calls

GarmashAlex:
Remove two Interlocked.MemoryBarrier() calls in McsLock
- After setting State = Acquired in  WaitForUnlock() 
- After setting next.State = ReadyToAcquire in  Disposable.Dispose()

Rely on .NET volatile semantics (acquire/release) for ThreadNode.State and monitor semantics (Monitor.Wait/Pulse) to ensure correct memory ordering and visibility.


LukaszRozmej:
I doubt they are not needed, but @benaadams ?

PR #9554, opened 2025-10-23
fix: use UTC timestamps in SimpleConsoleLogger for consistent log correlation

leopardracer:
Replace `DateTime.Now` with `DateTime.UtcNow` in `SimpleConsoleLogger` to ensure consistent log timestamps across different time zones and environments.


LukaszRozmej:
Thank you, but `SimpleConsoleLogger` is really a helper class and not used for main login. And local time is there by design.

PR #9555, opened 2025-10-24
Fix Arb TrieException

asdacap:
- Fix Arb trie exception.
- Fix long pruning due to commit buffer immediately filling up cache which did not give chance to prune persisted nodes completly which increase pruning time which increase commit buffer size.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- [X] Arb can run for a few days
- [x] Mainnet can sync
- [ ] Mainnet era 
- [x] Mainnet full pruning

### DIFF
--- a/src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
+++ b/src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
@@ -323,6 +323,7 @@ public sealed class TrieStore : ITrieStore, IPruningTrieStore
     {
         _scopeLock.Enter();
 
+        SpinWait spinWait = new SpinWait();
         while (true)
         {
             if (_pruningLock.TryEnter())
@@ -345,7 +346,7 @@ public sealed class TrieStore : ITrieStore, IPruningTrieStore
                 {
                     // This can happen in the tiny time in between pruningLock was acquired but the exact block to
                     // persist was not determined yet.
-                    Thread.Sleep(1);
+                    spinWait.SpinOnce();
                     continue;
                 }
 
--- a/src/Nethermind/Nethermind.Trie/Pruning/TrieStoreDirtyNodesCache.cs
+++ b/src/Nethermind/Nethermind.Trie/Pruning/TrieStoreDirtyNodesCache.cs
@@ -44,7 +44,12 @@ internal class TrieStoreDirtyNodesCache
         // If the nodestore indicated that path is not required,
         // we will use a map with hash as its key instead of the full Key to reduce memory usage.
         _storeByHash = storeByHash;
+
+        // Keep root causes persisted root nodes to not get pruned out of the cache. This ensure that it will
+        // be deleted when another canonical state is persisted which prevent having incomplete state which can happen
+        // when inner nodes get deleted but the root does not.
         _keepRoot = keepRoot;
+
         // NOTE: DirtyNodesCache is already sharded.
         int concurrencyLevel = Math.Min(Environment.ProcessorCount * 4, 32);
         int initialBuckets = TrieStore.HashHelpers.GetPrime(Math.Max(31, concurrencyLevel));
@@ -493,6 +498,11 @@ internal class TrieStoreDirtyNodesCache
         public override string ToString()
         {
             return $"A:{Address} P:{Path} K:{Keccak}";
+        }
+
+        public bool IsRoot()
+        {
+            return Path.Length == 0;
         }
     }
 


src/Nethermind/Nethermind.Trie/Pruning/TrieStoreDirtyNodesCache.cs
```diff
@@ -325,7 +320,7 @@ public void PruneCache(
                         continue;
                     }
 
-                    if (_trieStore.IsNoLongerNeeded(lastCommit))
+                    if (_trieStore.IsNoLongerNeeded(lastCommit) && !(_keepRoot && key.Path.Length == 0))
```
LukaszRozmej: Add method/extension method?

```suggestion
                    if (_trieStore.IsNoLongerNeeded(lastCommit) && !(_keepRoot && key.IsRoot()))
```

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -321,22 +322,51 @@ long blockNumber
...
             {
-                _pruningLock.Exit();
-                _scopeLock.Exit();
-            });
+                long persistedBoundary = Interlocked.Read(ref _toBePersistedBlockNumber);
+                if (persistedBoundary == -1)
+                {
+                    // This can happen in the tiny time in between pruningLock was acquired but the exact block to
+                    // persist was not determined yet.
+                    Thread.Sleep(1);
```
LukaszRozmej: Maybe `Thread.Yield()`?

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -1483,47 +1553,55 @@ public TrieNode FindCachedOrUnknown(TrieStoreDirtyNodesCache.Key key, bool isRea
...
                 }
             }
 
-            return isReadOnly ? bufferShard.FromCachedRlpOrUnknown(key) : bufferShard.FindCachedOrUnknown(key);
+            if (hasInBuffer)
+            {
+                return bufferNode;
+            }
+            else
+                return bufferShard.FindCachedOrUnknown(key);
```
LukaszRozmej:
```suggestion
            return hasInBuffer ? bufferNode : bufferShard.FindCachedOrUnknown(key);
```

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
benaadams: Would use a `SpinWait` and `.SpinOnce()` instead https://learn.microsoft.com/en-us/dotnet/api/system.threading.spinwait?view=net-9.0

It first does cpu Yield, then Thread.Yield and then Sleep as it gets called more times

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -84,18 +85,18 @@ public TrieStore(
...
+        _deleteOldNodes = _pruningStrategy.DeleteObsoleteKeys;
         _shardBit = pruningConfig.DirtyNodeShardBit;
         _shardedDirtyNodeCount = 1 << _shardBit;
         _dirtyNodes = new TrieStoreDirtyNodesCache[_shardedDirtyNodeCount];
         _dirtyNodesTasks = new Task[_shardedDirtyNodeCount];
         _persistedHashes = new ConcurrentDictionary<HashAndTinyPath, Hash256?>[_shardedDirtyNodeCount];
         for (int i = 0; i < _shardedDirtyNodeCount; i++)
         {
-            _dirtyNodes[i] = new TrieStoreDirtyNodesCache(this, !_nodeStorage.RequirePath, _logger);
+            _dirtyNodes[i] = new TrieStoreDirtyNodesCache(this, !_nodeStorage.RequirePath, keepRoot: _deleteOldNodes, _logger);
```
damian-orzechowski, line 96: To me, `keepRoot` assignement is a bit confusing here

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
asdacap: Added comment in `TrieStoreDirtyNodesCache`

PR #9560, opened 2025-10-24
Add OP Jovian fork

### LINKED ISSUE #9390: OP Stack hardfork support - Jovian
Jovian is the upcoming OP-stack hardfork which introduces breaking changes to how the Execution Layer component should behave. 

The required changes are documented in the official OP Stack specification: https://specs.optimism.io/protocol/jovian/exec-engine.html. Also, public documents from OP Labs are available with summaries of the fork: https://www.notion.so/oplabs/EXTERNAL-Upgrade-17-Jovian-Hub-263f153ee1628081bbacc5e78142a45a and https://oplabs.notion.site/Jovian-U17-Scope-22df153ee162800595a3dfe1f53c525a

Nethermind should make the required changes to its Optimism plugin in order to be properly compliant with the latest hardfork.

The only option besides actually implementing the hardfork is dropping support for OP-stack chains. We won't go down this route.

**Additional context**
Jovian includes the following features:

## Minimum base fee:
- Specification: https://specs.optimism.io/protocol/jovian/exec-engine.html#minimum-base-fee
- OP-geth: https://github.com/ethereum-optimism/op-geth/pull/666

## DA Footprint Block Limit
- Specification: https://specs.optimism.io/protocol/jovian/exec-engine.html#da-footprint-block-limit
- OP-geth: https://github.com/ethereum-optimism/op-geth/pull/675

## Operator fee fix
- Specification: https://github.com/ethereum-optimism/design-docs/blob/main/protocol/operator-fee-scalar-fix.md
- OP-geth: https://github.com/ethereum-optimism/op-geth/pull/676

**Timeline**
As of the time of writing, there are no defined deadlines for Jovian - mostly due to the development of the Fusaka fork on Ethereum mainnet. Nevertheless, Jovian is expected to be ready late October to mid November.

alexb5dh:
Resolves #9390

## Changes

- Support OP Jovian

#### Requires testing

- [x] Yes

- [x] Yes

#### Notes on testing

Manual testing is still required.

- [x] No

- [x] Yes

We should communicate that Nethermind is ready to support the new OP fork.

### DIFF
--- a/src/Nethermind/Nethermind.Optimism/OptimismCostHelper.cs
+++ b/src/Nethermind/Nethermind.Optimism/OptimismCostHelper.cs
@@ -164,7 +164,7 @@ public class OptimismCostHelper(IOptimismSpecHelper opSpecHelper, Address l1Bloc
             UInt256 flzLen = L1CostFastlzCoef * ComputeFlzCompressLen(tx);
             UInt256 daUsageEstimate = UInt256.Max(
                 MinTransactionSizeScaled,
-                flzLen > L1CostInterceptNeg ? flzLen - L1CostInterceptNeg : 0 // avoid uint overflow
+                flzLen > L1CostInterceptNeg ? flzLen - L1CostInterceptNeg : 0 // avoid uint underflow
             ) / DaFootprintScale;
 
             footprint += daUsageEstimate * daFootprintScalar;
--- a/src/Nethermind/Nethermind.Optimism/OptimismHeaderValidator.cs
+++ b/src/Nethermind/Nethermind.Optimism/OptimismHeaderValidator.cs
@@ -99,7 +99,7 @@ public class OptimismHeaderValidator(
                 return false;
             }
 
-            if ((long)blobGasUsed > header.GasLimit)
+            if (blobGasUsed > (ulong)header.GasLimit)
             {
                 if (_logger.IsWarn) _logger.Warn($"Invalid block header ({header.Hash}) - gas used above gas limit");
                 error = ErrorMessages.DaFootprintExceededGasLimit;
@@ -115,7 +115,7 @@ public class OptimismHeaderValidator(
     private static class ErrorMessages
     {
         public static readonly string RequestHashShouldBeOfShaOfEmpty = $"{nameof(BlockHeader.RequestsHash)} should be {OptimismPostMergeBlockProducer.PostIsthmusRequestHash} for post-Isthmus blocks";
-        public const string DaFootprintMissing = $"DA footprint missing from block header {nameof(BlockHeader.BlobGasUsed)}.";
+        public const string DaFootprintMissing = $"InvalidBlobGasUsed: DA footprint is missing from the block header {nameof(BlockHeader.BlobGasUsed)}.";
         public const string DaFootprintExceededGasLimit = "ExceededGasLimit: DA footprint exceeds gas limit.";
     }
 }
--- a/src/Nethermind/Nethermind.Core/Specs/IReleaseSpec.cs
+++ b/src/Nethermind/Nethermind.Core/Specs/IReleaseSpec.cs
@@ -354,7 +354,7 @@ namespace Nethermind.Core.Specs
         /// OP Holocene
         bool IsOpHoloceneEnabled { get; }
 
-        /// OP Holocene
+        /// OP Jovian
         bool IsOpJovianEnabled { get; }
 
         // OP Isthmus
--- a/src/Nethermind/Nethermind.Optimism.Test/OptimismBaseFeeCalculatorTests.cs
+++ b/src/Nethermind/Nethermind.Optimism.Test/OptimismBaseFeeCalculatorTests.cs
@@ -51,7 +51,7 @@ public class OptimismBaseFeeCalculatorTests
         actualBaseFee.Should().Be((UInt256)expectedBaseFee);
     }
 
-    private class JovianTest
+    private static class JovianTest
     {
         public const long GasLimit = 30_000_000;
         public const uint Denominator = 50;


src/Nethermind/Nethermind.Optimism/OptimismBaseFeeCalculator.cs
```diff
@@ -35,6 +36,14 @@ public UInt256 Calculate(BlockHeader parent, IEip1559Spec specFor1559)
             };
         }
 
-        return baseFeeCalculator.Calculate(parent, spec);
+        UInt256 baseFee = baseFeeCalculator.Calculate(parent, spec);
+
+        // TODO: use jovianTimestamp check instead?
```
emlautarom1: This is valid too since the header should be valid at this point, thus if `MinBaseFee` is not null then we're on Jovian

src/Nethermind/Nethermind.Optimism/OptimismBlockProcessor.cs
```diff
@@ -51,13 +55,29 @@ public OptimismBlockProcessor(
...
 
     protected override TxReceipt[] ProcessBlock(Block block, IBlockTracer blockTracer, ProcessingOptions options, IReleaseSpec spec, CancellationToken token)
     {
         _contractRewriter?.RewriteContract(block.Header, _stateProvider);
-        return base.ProcessBlock(block, blockTracer, options, spec, token);
+        TxReceipt[] receipts = base.ProcessBlock(block, blockTracer, options, spec, token);
+
+        if (_opSpecHelper.IsJovian(block.Header))
+        {
+            // TODO: handle overflow differently or calculate footprint as long?
```
emlautarom1: To avoid overflow do all computations in `UInt256`, and then cast to `long` and check for overflow

src/Nethermind/Nethermind.Optimism/OptimismCostHelper.cs
```diff
@@ -141,6 +146,35 @@ public UInt256 ComputeOperatorCost(long gas, BlockHeader header, IWorldState wor
...
+            return 0;
+
+        var daFootprintGasScalar = ReadUInt16BigEndian(span[scalarPosition..]);
+
+        UInt256 footprint = UInt256.Zero;
+
+        foreach (Transaction tx in block.Transactions)
+        {
+            if (tx.Type == TxType.DepositTx)
+                return 0;
```
emlautarom1: We don't abort with 0 but instead skip Deposit transactions.

```suggestion
                continue;
```

src/Nethermind/Nethermind.Optimism/OptimismEIP1559Parameters.cs
```diff
@@ -25,69 +29,102 @@ public EIP1559Parameters(byte version, UInt32 denominator, UInt32 elasticity)
...
         }
 
-        parameters = new EIP1559Parameters(version, denominator, elasticity);
+        parameters = new EIP1559Parameters(1, denominator, elasticity, minBaseFee);
+
         return true;
     }
 
-    public bool IsZero() => Denominator == 0 && Elasticity == 0;
+    public bool IsZero() => Version == 0 && Denominator == 0 && Elasticity == 0 && MinBaseFee == null;
```
emlautarom1: `IsZero` is used to default the EIP1559 parameters in Holocene. For Jovian, the same defaulting mechanism should be used and note that `MinBaseFee` is never `null` according to the spec (otherwise it's invalid).

With this code, after Jovian, a valid instance of `EIP1559Parameters` struct will never return `false` for this method since `MinBaseFee == null` is always false, which is incorrect.

src/Nethermind/Nethermind.Optimism.Test/OptimismPayloadAttributesTests.cs
```diff
@@ -19,12 +19,21 @@ public class OptimismPayloadAttributesTests
...
     {
+        // V0
         yield return ("0x000000000100000000", "0x00dea77451f10b20");
         yield return ("0x0000000001000001bc", "0xf2975f6725d5f2e5");
         yield return ("0x0000000001ffffffff", "0x6b09fc2a90d6c067");
         yield return ("0x00ffffffff00000000", "0x9787e23f29594f18");
         yield return ("0x00ffffffff000001bc", "0x2cb414f72aac7824");
         yield return ("0x00ffffffffffffffff", "0xe411646692277df5");
+        // V1
+        //TODO: source of expected results?
```
alexb5dh: Were existing cases taken from somewhere or just calculated manully?

src/Nethermind/Nethermind.Optimism/OptimismBaseFeeCalculator.cs
```diff
@@ -35,6 +37,27 @@ public UInt256 Calculate(BlockHeader parent, IEip1559Spec specFor1559)
...
+        if (parent.Timestamp >= jovianTimestamp)
+        {
+            if (parent.BlobGasUsed is null)
+                throw new InvalidOperationException($"{nameof(parent.BlobGasUsed)} does not store DA footprint in post-Jovian block.");
+
+            // Override gas used for calculation if the DA footprint is larger
+            var blobGasUsed = (long)parent.BlobGasUsed;
+            if (blobGasUsed > parent.GasUsed)
+            {
+                parent = parent.Clone();
```
alexb5dh: Don't like additional allocation and time wasted on cloning, but simply modifying original entity seems unsafe.

Alternative can be something like adding `virtual protected Calculate` method with additional `long gasOverride` parameter to `DefaultBaseFeeCalculator`, inheriting the class instead of wrapping it, and calling the new method instead of regular `Calculate`.

src/Nethermind/Nethermind.Optimism/OptimismCostHelper.cs
```diff
@@ -290,4 +324,31 @@ uint setNextHash(uint ip, ref Span<uint> ht)
...
+        var firstTx = block.Transactions.FirstOrDefault();
+        if (firstTx?.Type is not TxType.DepositTx)
+            return DaFootprintScalarDefault;
+
+        if (firstTx.Data.Length < 178)
+            return DaFootprintScalarDefault;
+
+        var scalar = ReadUInt16BigEndian(firstTx.Data.Span[176..178]);
+        return scalar == 0 ? DaFootprintScalarDefault : scalar;
+    }
```
alexb5dh, line 339: Added 2 options to calculate `daFootprintGasScalar` as [per spec](https://specs.optimism.io/protocol/jovian/exec-engine.html#scalar-loading).

Not sure which one is better, but version taking scalar from the block deposit tx doesn't require state, so has more diverse usage options. Can also try to fetch from both sources before falling back to the default value.

LukaszRozmej (approved):

src/Nethermind/Nethermind.Optimism/OptimismCostHelper.cs
emlautarom1: > version taking scalar from the block deposit tx doesn't require state, so has more diverse usage options

Let's go with this approach since, as you said, it's more flexible and we don't need to reach out or state. If this fails let's directly fall back to the default value.

src/Nethermind/Nethermind.Optimism/OptimismBaseFeeCalculator.cs
emlautarom1: > but simply modifying original entity seems unsafe.

Block is not being used concurrently AFAIK so it's safe to modify and then restore to the previous state (with proper use of `try...finally`), so no need to clone.

> Alternative can be something like...

This is extremely heavyweight for this kind of change. I'd rather have a clone than this kind of setup.

src/Nethermind/Nethermind.Optimism.Test/OptimismPayloadAttributesTests.cs
emlautarom1: If I recall correctly, I created some tests in op-geth with these EIP 1559 params and printed the computed values, and then ported those results to Nethermind.

src/Nethermind/Nethermind.Optimism/OptimismCostHelper.cs
emlautarom1: Also, what does op-geth do in this scenario? Maybe the better approach is just to copy what they're doing.

src/Nethermind/Nethermind.Optimism/OptimismBaseFeeCalculator.cs
```diff
@@ -35,6 +37,27 @@ public UInt256 Calculate(BlockHeader parent, IEip1559Spec specFor1559)
...
+                parent.GasUsed = blobGasUsed;
+            }
```
emlautarom1: Here there are a couple of options:

- Clone the block and make the overrides. It's the safest approach but a bit heavyweight since we only need to override one field.
- Mutate the block directly in a `try...finally` block where we restore the original values. I'm pretty sure this is safe since no other threads should be manipulating the same block. If you check the call tree for this method I'm pretty sure that you'll find other situations where we mutate the block.
- Lastly, we could copy the body of `DefaultBaseFeeCalculator.Calculate` into this implementation and make the required adjustments here. We'll have a bit of duplication but we drop the dependency on other components and get full control of this implementation.

src/Nethermind/Nethermind.Optimism/OptimismChainSpecEngineParameters.cs
```diff
@@ -52,7 +54,7 @@ public void ApplyToReleaseSpec(ReleaseSpec spec, long startBlock, ulong? startTi
 
         if (HoloceneTimestamp is not null)
         {
-            spec.BaseFeeCalculator = new OptimismBaseFeeCalculator(HoloceneTimestamp.Value, new DefaultBaseFeeCalculator());
+            spec.BaseFeeCalculator = new OptimismBaseFeeCalculator(HoloceneTimestamp.Value, JovianTimestamp, new DefaultBaseFeeCalculator());
```
emlautarom1: I'd rather have this a bit more consistent and pass both Holocene and Jovian timestamps as optionals (most likely the `HolocenteTimestamp is not null` check should be removed).

src/Nethermind/Nethermind.Optimism/OptimismHeaderValidator.cs
```diff
@@ -73,8 +84,37 @@ protected override bool ValidateRequestsHash(BlockHeader header, IReleaseSpec sp
...
+                if (_logger.IsWarn) _logger.Warn($"Invalid block header ({header.Hash}) - gas used above gas limit");
+                error = ErrorMessages.DaFootprintExceededGasLimit;
+                return false;
+            }
+        }
+
+        return true;
+    }
+
+    protected override ulong? CalculateExcessBlobGas(BlockHeader parent, IReleaseSpec spec) => 0;
```
emlautarom1, line 113: Why is this here? There should be no calculations performed in validation.

src/Nethermind/Nethermind.Optimism.Test/EIP1559ParametersTests.cs
```diff
@@ -7,6 +7,9 @@
 using NUnit.Framework;
 using FluentAssertions;
 using System;
+using Nethermind.Core.Specs;
+using Nethermind.Specs.Test;
+using NSubstitute;
```
emlautarom1: Unused imports.

src/Nethermind/Nethermind.Optimism.Test/OptimismCostHelperTests.cs
```diff
@@ -0,0 +1,113 @@
...
+        var worldState = new WorldStateStab(new() {
+            // https://specs.optimism.io/protocol/jovian/exec-engine.html#scalar-loading
+            { new(l1BlockAddr, new UInt256(8)), DefaultDaFootprintGasScalar.ToBigEndianByteArray().PadLeft(12 + sizeof(UInt16)) }
+        });
+
+        var helper = new OptimismCostHelper(specHelper, l1BlockAddr);
+
+        Block block = Build.A.Block.WithTransactions(transactions).TestObject;
+        return (long)helper.ComputeDaFootprint(block, worldState);
+    }
```
emlautarom1: If we go the route of using the block's deposit tx to fetch the scalar then we don't need this test.

src/Nethermind/Nethermind.Optimism.Test/OptimismHeaderValidatorTests.cs
```diff
@@ -24,36 +22,53 @@ public class OptimismHeaderValidatorTests
     private static readonly Hash256 PostCanyonWithdrawalsRoot = Keccak.OfAnEmptySequenceRlp;
 
     [TestCaseSource(nameof(EIP1559ParametersExtraData))]
-    public void Validates_EIP1559Parameters_InExtraData_AfterHolocene((string HexString, bool IsValid) testCase)
+    public void Validates_EIP1559Parameters_InExtraData_AfterHolocene_OrJovian((string HexString, bool validHolocene, bool validJovian) testCase)
```
emlautarom1: I'd rather have separate tests for Holocene and Jovian. Currently it's a bit hard to understand what's being tested on each case.

src/Nethermind/Nethermind.Optimism.Test/OptimismPayloadAttributesTests.cs
```diff
@@ -44,15 +53,20 @@ public void Compute_PayloadID_with_EIP1559Params((string HexStringEIP1559Params,
...
+        yield return (new byte[15], PayloadAttributesValidationResult.InvalidPayloadAttributes, PayloadAttributesValidationResult.InvalidPayloadAttributes, PayloadAttributesValidationResult.InvalidPayloadAttributes);
+        yield return (new byte[16], PayloadAttributesValidationResult.InvalidPayloadAttributes, PayloadAttributesValidationResult.InvalidPayloadAttributes, PayloadAttributesValidationResult.Success);
+        yield return (new byte[17], PayloadAttributesValidationResult.InvalidPayloadAttributes, PayloadAttributesValidationResult.InvalidPayloadAttributes, PayloadAttributesValidationResult.InvalidPayloadAttributes);
     }
-    [TestCaseSource(nameof(Validate_EIP1559Params_Holocene_TestCases))]
-    public void Validate_EIP1559Params_Holocene((byte[]? Eip1559Params, PayloadAttributesValidationResult BeforeHolocene, PayloadAttributesValidationResult AfterHolocene) testCase)
+    [TestCaseSource(nameof(Validate_EIP1559Params_Holocene_Jovian_TestCases))]
+    public void Validate_EIP1559Params_Holocene_Jovian((byte[]? Eip1559Params, PayloadAttributesValidationResult BeforeHolocene,
```
emlautarom1: Same here, maybe it's better to have a separate test with "before/after Jovian" than mixing it with Holocene.

Note how for Holocene we had an implicit "after Granite" in "before Holocene": this case is similar, where we have an implicit "after Holocene" in "before Jovian"

src/Nethermind/Nethermind.Core.Test/Builders/TransactionBuilder.cs
```diff
@@ -52,12 +53,14 @@ public TransactionBuilder<T> To(Address? address)
...
         }
 
-        public TransactionBuilder<T> WithData(byte[] data)
+        public TransactionBuilder<T> WithData(byte[]? data)
         {
             TestObjectInternal.Data = data;
             return this;
         }
 
+        public TransactionBuilder<T> WithDataHex(string dataHex) => WithData(Bytes.FromHexString(dataHex));
```
emlautarom1, line 62: You can leave the `string -> byte[]` conversion to the caller instead of adding an additional overload.

src/Nethermind/Nethermind.Consensus/Validators/HeaderValidator.cs
```diff
@@ -376,5 +376,10 @@ protected virtual bool ValidateBlobGasFields(BlockHeader header, BlockHeader par
 
             return true;
         }
+
+        protected virtual ulong? CalculateExcessBlobGas(BlockHeader parent, IReleaseSpec spec)
+        {
+            return BlobGasCalculator.CalculateExcessBlobGas(parent, spec);
+        }
```
emlautarom1, line 383: This should not be part of `HeaderValidator`. Instead, pass the calculator as a constructor argument (like `BlockTree`).

src/Nethermind/Nethermind.Optimism/OptimismHeaderValidator.cs
alexb5dh: There are no calculations, it return's 0. This way I'm reusing `base.ValidateBlobGasFields` to validate that `ExcessBlobGas` is 0. There are other useful (and required) checks performed in this method, and I dind't want to copy it.

src/Nethermind/Nethermind.Optimism.Test/OptimismCostHelperTests.cs
alexb5dh: We still need to validate that computed values matches expected. Scalar is just one of multipliers.

src/Nethermind/Nethermind.Core.Test/Builders/TransactionBuilder.cs
alexb5dh: I would prefer this approach instead of adding `Bytes.FromHexString` in every call in places like this. Is this a problem?
image

src/Nethermind/Nethermind.Consensus/Validators/HeaderValidator.cs
alexb5dh: Static `BlobGasCalculator` call is currently used in `master`, should I change it?

src/Nethermind/Nethermind.Optimism/OptimismHeaderValidator.cs
alexb5dh: And base `HeaderValidator` does some calculations - it uses `BlobGasCalculator.CalculateExcessBlobGas` instead as seen below.

src/Nethermind/Nethermind.Optimism.Test/OptimismCostHelperTests.cs
emlautarom1: But do we need the whole `WorldStateStab` setup?

src/Nethermind/Nethermind.Consensus/Validators/HeaderValidator.cs
LukaszRozmej: feel free to wrap it

src/Nethermind/Nethermind.Core.Test/Builders/TransactionBuilder.cs
LukaszRozmej: I also like this

src/Nethermind/Nethermind.Optimism/OptimismCostHelper.cs
alexb5dh: They also extract it from block tx data:
https://github.com/ethereum-optimism/op-geth/blob/aa940538653c59df97b432bc24a2a4e4b17a06a8/core/types/rollup_cost.go#L547C6-L547C33

src/Nethermind/Nethermind.Optimism.Test/OptimismHeaderValidatorTests.cs
```diff
@@ -130,23 +150,87 @@ public void ValidateRequestHash(ulong timestamp, Hash256? requestHash, bool isVa
         valid.Should().Be(isValid);
     }
 
-    private static IEnumerable<(string, bool)> EIP1559ParametersExtraData()
+    private static IEnumerable<(string, bool, bool)> EIP1559ParametersExtraData()
```
flcl42: Could you name the tuple items

src/Nethermind/Nethermind.Optimism/OptimismHeaderValidator.cs
```diff
@@ -73,8 +84,37 @@ protected override bool ValidateRequestsHash(BlockHeader header, IReleaseSpec sp
...
     protected override bool ValidateGasLimitRange(BlockHeader header, BlockHeader parent, IReleaseSpec spec, ref string? error) => true;
 
+    protected override bool ValidateGasUsed(BlockHeader header, ref string? error)
+    {
+        if (!base.ValidateGasUsed(header, ref error))
+            return false;
+
+        if (specHelper.IsJovian(header))
+        {
+            if (header.BlobGasUsed is not { } blobGasUsed)
```
flcl42, line 95: Is there a reason blob check are not in ValidateBlobGasFields override?

src/Nethermind/Nethermind.Optimism/OptimismEIP1559Parameters.cs
```diff
@@ -25,69 +28,128 @@ public EIP1559Parameters(byte version, UInt32 denominator, UInt32 elasticity)
...
-        var denominator = BinaryPrimitives.ReadUInt32BigEndian(extraData.TakeAndMove(4));
-        var elasticity = BinaryPrimitives.ReadUInt32BigEndian(extraData.TakeAndMove(4));
+        var denominator = BinaryPrimitives.ReadUInt32BigEndian(data.TakeAndMove(sizeof(UInt32)));
+        var elasticity = BinaryPrimitives.ReadUInt32BigEndian(data.TakeAndMove(sizeof(UInt32)));
+
+        if (version == 0)
+            return EIP1559Parameters.TryCreateV0(denominator, elasticity, out parameters, out error);
+
+        var minBaseFee = BinaryPrimitives.ReadUInt64BigEndian(data.TakeAndMove(sizeof(UInt64)));
+        return EIP1559Parameters.TryCreateV1(denominator, elasticity, minBaseFee, out parameters, out error);
```
flcl42, line 125: Whitespace below is redundant

src/Nethermind/Nethermind.Optimism/OptimismEIP1559Parameters.cs
```diff
@@ -25,69 +28,128 @@ public EIP1559Parameters(byte version, UInt32 denominator, UInt32 elasticity)
...
+        {
+            error = $"{nameof(attributes.EIP1559Params)} has invalid length";
+            return false;
+        }
+
+        var denominator = BinaryPrimitives.ReadUInt32BigEndian(data.TakeAndMove(sizeof(UInt32)));
+        var elasticity = BinaryPrimitives.ReadUInt32BigEndian(data.TakeAndMove(sizeof(UInt32)));
+
+        if (version == 0)
+            return EIP1559Parameters.TryCreateV0(denominator, elasticity, out parameters, out error);
```
flcl42, line 152: It's better to surround the statement with  {} and use specific types instead of vars

src/Nethermind/Nethermind.Optimism/OptimismHeaderValidator.cs
```diff
@@ -52,6 +51,18 @@ protected override bool Validate<TOrphaned>(BlockHeader header, BlockHeader? par
...
                 return false;
             }
+
+            if (!specHelper.IsJovian(header) && parameters.Version != 0)
+            {
+                error = $"{nameof(EIP1559Parameters)} version should be 0 before Jovian";
+                return false;
+            }
+
+            if (specHelper.IsJovian(header) && parameters.Version != 1)
```
flcl42: Can it so happen that with the next fork it will start failing? I mean the check probably can be not `Version != 1` but `Version < 1`

alexb5dh: But if version is increased related fork should explicitly check for the exact version value, so this code will be updated accordingly.

src/Nethermind/Nethermind.Optimism/OptimismEIP1559Parameters.cs
alexb5dh: Which exact vars are we talking about?
Rider usually suggest this if resulting type is unclear, but everything seems straightforward here.

src/Nethermind/Nethermind.Optimism/OptimismHeaderValidator.cs
flcl42: IsJovian will be still true during the next post-Jovian fork, right? Seems like it will be true and version will be 2 it will lead to a false positive?

src/Nethermind/Nethermind.Optimism/OptimismEIP1559Parameters.cs
flcl42: For example `var elasticity` and actually everywhere in the pr. It's much easier to read when type is mentioned. @rubo is it possible to make IDEs suggest/enforce using explicit types?

src/Nethermind/Nethermind.Optimism/OptimismHeaderValidator.cs
alexb5dh: Idea is that each fork should check for exact expected version. Other checks should not even be entered. Updated code to hopefully make it more explicit.

src/Nethermind/Nethermind.Optimism.Test/OptimismCostHelperTests.cs
alexb5dh: Replaced test cases with exact blocks from the alpha devnet.

src/Nethermind/Nethermind.Optimism.Test/OptimismHeaderValidatorTests.cs
alexb5dh: Didn't like amount of duplicated code already present and introduced by tests separation so [refactored](https://github.com/NethermindEth/nethermind/pull/9560/commits/239dfad303b2d25f7c5d0cbf3a67881b730958f3) this fixture a bit.

<details>
  <summary>Made each test set to be explicitly run for every fork timestamp:</summary>
  image
  
</details>

<details>
  <summary>Added helper class to describe timestamp validity ranges for every value(s):</summary>
  image
  image
  
</details>

Not sure if this is an overengineering, but at it increases test coverage and reduces code duplication. Can apply to other modified classes if looks fine.

src/Nethermind/Nethermind.Consensus/Validators/HeaderValidator.cs
alexb5dh: Is it worth it?

`BlobGasCalculator` has 7 static methods used accross ~20 files, so modification should affect only `HeaderValidator` to not overgrow the PR. And since `BlobGasCalculator` name is taken, instance alternative should probably have a different one - `BlobGasCalculatorInstance`?

emlautarom1: I guess for now it's fine as it is but we should revisit this later (refactor PR at some point?).

src/Nethermind/Nethermind.Optimism.Test/Fork.cs
```diff
@@ -0,0 +1,25 @@
...
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
+namespace Nethermind.Optimism.Test;
+
+public record Fork(string Name, ulong Timestamp)
```
emlautarom1, line 12: Although I appreciate the resulting code I think we can make do with less machinery here considering that this class is only used in `Valid`, which uses it for its `.ToString` method.

I suggest removing it entirely and making the `Valid` class more like a mathematical range (ex. `[1..]`, `[..3]` and `[1..3]`). If we do that we could eventually reuse the class outside the Optimism module

src/Nethermind/Nethermind.Optimism.Test/OptimismHeaderValidatorTests.cs
```diff
@@ -1,152 +1,165 @@
...
 namespace Nethermind.Optimism.Test;
 
 [Parallelizable(ParallelScope.All)]
-public class OptimismHeaderValidatorTests
+[TestFixture(GenesisTimestamp + 1, TestName = "Genesis")]
+[TestFixture(CanyonTimestamp, TestName = "Canyon")]
+[TestFixture(EcotoneTimestamp, TestName = "Ecotone")]
+[TestFixture(HoloceneTimeStamp, TestName = "Holocene")]
+[TestFixture(IsthmusTimeStamp, TestName = "Isthmus")]
+[TestFixture(JovianTimeStamp, TestName = "Jovian")]
```
emlautarom1: Nitpick: we should prefix this with `Spec` rather than rely on `using static` since it can be confusing for new devs: it looks like these timestamps are related to production and constants across all OP chains.

src/Nethermind/Nethermind.Optimism.Test/OptimismHeaderValidatorTests.cs
```diff
@@ -1,152 +1,165 @@
...
+        new("0x0", Valid.Never),
+        new("0xffffaaaa", Valid.Never),
+        new("0x01ffffffff00000000", Valid.Never),
+        new("0xff0000000100000001", Valid.Never),
+        new("0x000000000000000000", Valid.Never),
+        new("0x000000000000000001", Valid.Never),
+        new("0x01ffffffff000001bc00000000000000", Valid.Never),
+        new("0x01ffffffff000001bc000000000000000000", Valid.Never),
+    }.Select(x =>
+        (x.data, Valid.Before(HoloceneTimeStamp) | x.valid) // No validation before Holocene
```
emlautarom1: This is slightly incorrect: pre Holocene, the `extraData` must always be empty, except for Genesis which might contain arbitrary data: https://specs.optimism.io/protocol/exec-engine.html?highlight=extraData#extra-data

src/Nethermind/Nethermind.Optimism.Test/OptimismHeaderValidatorTests.cs
```diff
@@ -1,152 +1,165 @@
...
-        var valid = validator.Validate(header, genesis);
+        Assert.That(() => validator.Validate(header, genesis), Is.EqualTo(testCase.valid.On(timestamp)));
+    }
 
-        valid.Should().Be(testCase.IsValid);
+    private static IEnumerable<(Hash256? requestHash, Valid valid)> WithdrawalsRequestHashTestCases()
+    {
+        yield return (null, Valid.Before(IsthmusTimeStamp));
+        yield return (TestItem.KeccakA, Valid.Before(IsthmusTimeStamp));
+        yield return (OptimismPostMergeBlockProducer.PostIsthmusRequestHash, Valid.Always);
```
emlautarom1: We're reducing the coverage from previous tests (ex. we are not checking that post Isthmus a `null` request hash is invalid)

src/Nethermind/Nethermind.Optimism.Test/OptimismHeaderValidatorTests.cs
```diff
@@ -1,152 +1,165 @@
...
 
-        var valid = validator.Validate(header, genesis);
-
-        valid.Should().BeTrue();
+        string? error = null;
+        Assert.That(() => validator.Validate(header, genesis, false, out error), Is.EqualTo(testCase.valid.On(timestamp)), () => error!);
     }
 
-    private static IEnumerable<TestCaseData> WithdrawalsRequestHashTestCases()
+    private static IEnumerable<(long gasLimit, long gasUsed, long? blobGasUsed, Valid valid)> GasLimitTestCases()
```
emlautarom1: I suggest using `TestCaseData` where you can use `.SetName` to describe each case instead of relying on comments.

src/Nethermind/Nethermind.Optimism.Test/OptimismHeaderValidatorTests.cs
alexb5dh: Hmm, this was "migrated" from the existing [Ignores_ExtraData_BeforeHolocene test](https://github.com/NethermindEth/nethermind/blob/master/src/Nethermind/Nethermind.Optimism.Test/OptimismHeaderValidatorTests.cs#L56C17-L56C49).

src/Nethermind/Nethermind.Optimism.Test/OptimismHeaderValidatorTests.cs
alexb5dh:
```
yield return (null, Valid.Before(IsthmusTimeStamp));
```
is covering this.

Every test case is run for each fork timestamp. `Valid.Before(IsthmusTimeStamp)` parameter will check that `null` is valid for `timestamp < IsthmusTimeStamp` and invalid for  `timestamp >= IsthmusTimeStamp`.

src/Nethermind/Nethermind.Optimism.Test/OptimismHeaderValidatorTests.cs
emlautarom1: Oh, interesting. So I guess we where not checking this previously. Then let's leave it as it is, no validation pre Holocene.

src/Nethermind/Nethermind.Optimism.Test/Fork.cs
alexb5dh: Needed to return `Fork` to also use it as `TestCaseSource`, representing all fork timestamps.

Updated `Valid` to use ranges.

src/Nethermind/Nethermind.Optimism/OptimismEIP1559Parameters.cs
```diff
@@ -25,69 +28,131 @@ public EIP1559Parameters(byte version, UInt32 denominator, UInt32 elasticity)
...
+            error = $"{nameof(header.ExtraData)} must not be empty";
             return false;
         }
 
-        ReadOnlySpan<byte> extraData = header.ExtraData.AsSpan();
-        var version = extraData.TakeAndMove(1)[0];
-        var denominator = BinaryPrimitives.ReadUInt32BigEndian(extraData.TakeAndMove(4));
-        var elasticity = BinaryPrimitives.ReadUInt32BigEndian(extraData.TakeAndMove(4));
+        var maxVersion = EIP1559Parameters.ByteLengthByVersion.Length - 1;
+        var version = data.TakeAndMove(1)[0];
```
LukaszRozmej: I know it was already there but avoid vars - hard to read outside IDE

PR #9565, opened 2025-10-25
Fix simulate errors

deffrian:
Fixes Closes Resolves #

```
eth_simulateV1/ethSimulate-gas-fees-and-value-error-38014 (nethermind)
eth_simulateV1/ethSimulate-move-to-address-itself-reference-38022 (nethermind)
eth_simulateV1/ethSimulate-move-two-non-precompiles-accounts-to-same (nethermind)
eth_simulateV1/ethSimulate-overflow-nonce-validation (nethermind)
eth_simulateV1/ethSimulate-override-address-twice (nethermind)
eth_simulateV1/ethSimulate-simple-no-funds (nethermind)
eth_simulateV1/ethSimulate-simple-no-funds-with-balance-querying (nethermind)
eth_simulateV1/ethSimulate-simple-send-from-contract-no-balance (nethermind)
eth_simulateV1/ethSimulate-try-to-move-non-precompile (nethermind)
```

## Changes

- Throw InvalidTransactionException instead of InvalidBlockException
- Map TransactionResult to SimulateErrorCode

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_


src/Nethermind/Nethermind.Facade/Simulate/SimulateErrorCodes.cs
```diff
@@ -0,0 +1,24 @@
...
+    BlockTimestampInvalid = -38021,
+    SenderIsNotEOA = -38024,
+    MaxInitCodeSizeExceeded = -38025,
+    ClientLimitExceeded = -38026,
+    InternalError = -32603,
+    InvalidParams = -32602,
+    Reverted = -32000,
+    VMError = -32015,
+    TxSyncTimeout = 4,
+}
```
LukaszRozmej: Maybe put those with standard `ErrorCodes`? Lot of them are duplicates.
Also Revert should be now 3

deffrian: These are geth error codes. I think some of them are wrong on there side, some on ours. I will need to investigate each one witch is a lot of work. For now I thought it's good idea to finish with the tests, so we pass all of them. And right after we can negotiate/fix the exact numbers.

LukaszRozmej: Sure but we have some of them already here, can we not duplicate?

https://github.com/NethermindEth/nethermind/blob/master/src/Nethermind/Nethermind.JsonRpc/ErrorCodes.cs

LukaszRozmej: At minimum put those error codes there (which are missing) and reuse const values for your enum.

src/Nethermind/Nethermind.JsonRpc/ErrorCodes.cs
```diff
@@ -126,5 +126,67 @@ public static class ErrorCodes
...
+
+        /// <summary>
+        /// EIP-3860. Code size is to big
+        /// </summary>
+        public const int MaxInitCodeSizeExceeded = -38025;
+
+        /// <summary>
+        /// Transaction reverted. Geth sets it to -32000 in simulate but suppose to 3. We kepp it as geth for now
+        /// </summary>
+        public const int RevertedSimulate = -32000;
```
LukaszRozmej, line 183: Maybe Geth already returns 3, but tests weren't re-generated?

deffrian: No, I checked. It's a bug in their code

PR #9566, opened 2025-10-25
Remove redundant length==1 special-case in SliceWithZeroPadding

GarmashAlex:
- Removed the length == 1 special-case and related comments from 
ReadOnlySpan.SliceWithZeroPadding(int, int, PadDirection) in ByteArrayExtensions.cs.
- The special-case was redundant because the general logic (copiedLength = Math.Min(span.Length - startIndex, length)) produces identical results when length == 1.
- The inner span.Length == 0 branch was unreachable due to the prior guard if (startIndex >= span.Length) return ....
- This change eliminates dead code and misleading comments, improving clarity without altering behavior.


benaadams (approved):

LukaszRozmej (approved):

PR #9569, opened 2025-10-26
Enable ProofRpcModuleTests fixture with non-zero gas price and system account

MozirDmitriy:
This PR enables the previously disabled TestFixture(true, true) in ProofRpcModuleTests to cover the scenario with a pre-created system account and non-zero gas price. The existing tests already account for the miner/system account inclusion in the collected accounts, so no additional logic changes were required. Enabling this fixture increases test coverage for proof_call behavior under paid gas conditions while keeping expectations aligned with current tracer and proof collection behavior.


LukaszRozmej:
doesn't work, feel free to reopen if fixed

PR #9571, opened 2025-10-27
Improve Block related caching

benaadams:
## Changes

- More reduced version of https://github.com/NethermindEth/nethermind/pull/8943
- Introduce block hash -> block number cache
- Don't update ClockCache concurrent dictionary if value the same
- Don't update cache in Block/Header finder if it came from cache in first place
- Use smaller keys for Block and Header cache

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain/BlockTree.cs
+++ b/src/Nethermind/Nethermind.Blockchain/BlockTree.cs
@@ -579,7 +579,17 @@ namespace Nethermind.Blockchain
                 return null;
             }
 
-            BlockHeader? header = _headerStore.Get(blockHash, out bool fromCache, shouldCache: false, blockNumber: blockNumber);
+            BlockHeader? header = null;
+            if ((options & BlockTreeLookupOptions.RequireCanonical) == 0)
+            {
+                header = _headerStore.GetFromCache(blockHash);
+                if (header is not null && (!blockNumber.HasValue || header.Number == blockNumber.Value))
+                {
+                    return header;
+                }
+            }
+
+            header = _headerStore.Get(blockHash, out bool fromCache, shouldCache: false, blockNumber: blockNumber);
             if (header is null)
             {
                 bool allowInvalid = (options & BlockTreeLookupOptions.AllowInvalid) == BlockTreeLookupOptions.AllowInvalid;
@@ -1435,8 +1445,17 @@ namespace Nethermind.Blockchain
                 return null;
             }
 
-            bool fromCache = false;
             Block? block = null;
+            if ((options & BlockTreeLookupOptions.RequireCanonical) == 0)
+            {
+                block = _blockStore.GetFromCache(blockHash);
+                if (block is not null && block.TotalDifficulty.HasValue && (!blockNumber.HasValue || block.Number == blockNumber.Value))
+                {
+                    return block;
+                }
+            }
+
+            bool fromCache = false;
             blockNumber ??= _headerStore.GetBlockNumber(blockHash);
             if (blockNumber is not null)
             {
--- a/src/Nethermind/Nethermind.Blockchain/Blocks/BlockStore.cs
+++ b/src/Nethermind/Nethermind.Blockchain/Blocks/BlockStore.cs
@@ -102,4 +102,7 @@ public class BlockStore([KeyFilter(DbNames.Blocks)] IDb blockDb, IHeaderDecoder 
     {
         _blockCache.Set(block.Hash, block);
     }
+
+    public Block? GetFromCache(Hash256 blockHash)
+        => _blockCache.Get(blockHash);
 }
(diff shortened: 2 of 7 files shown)


src/Nethermind/Nethermind.Core/Caching/ClockCache.cs
```diff
@@ -58,8 +58,18 @@ public bool Set(TKey key, TValue val)
...
-            if (_cacheMap.TryUpdate(key, new(val, ov.Offset), comparisonValue: ov))
+            bool needsUpdate;
+            if (typeof(TValue).IsValueType)
+            {
+                needsUpdate = !EqualityComparer<TValue>.Default.Equals(ov.Value, val);
+            }
+            else
+            {
+                needsUpdate = !ReferenceEquals(ov.Value, val);
+            }
```
LukaszRozmej: collapse?
```suggestion
            bool needsUpdate = typeof(TValue).IsValueType && !EqualityComparer<TValue>.Default.Equals(ov.Value, val) || !typeof(TValue).IsValueType && !ReferenceEquals(ov.Value, val);
```

benaadams: Changed

src/Nethermind/Nethermind.Blockchain/BlockhashProvider.cs
```diff
@@ -46,26 +46,7 @@ public BlockhashProvider(IBlockFinder blockTree, ISpecProvider specProvider, IWo
...
-                header = _blockTree.FindParentHeader(header, BlockTreeLookupOptions.TotalDifficultyNotNeeded);
-                if (header is null)
-                {
-                    throw new InvalidDataException("Parent header cannot be found when executing BLOCKHASH operation");
-                }
-            }
-
-            if (_logger.IsTrace) _logger.Trace($"BLOCKHASH opcode returning null for {currentBlock.Number} -> {number}");
-            return null;
+            return _blockTree.FindBlockHash(number);
```
asdacap: There was a bug before where the canonical block was not correct. Hence, why it iterate one by one.

asdacap: Can add a SaveMode flag or something that would use the previous implementation. Just replace it in BlockProcessingModule.

LukaszRozmej:
Could we add:
- ParentHash to the cache where we have blocks and numbers.
- On Head change actively scan through cache, double check and ensure we have all the items for Blockhash opcode? (unless a long reorg you won't actually hit DB!)

src/Nethermind/Nethermind.Blockchain/BlockhashProvider.cs
benaadams: Changed the caching to only update number when on mainchain (so block inserts alone don't update number cache)

Changed test to more explicitly check around this; also included the pivot forked block in when updating mainchain in test (check this if correct @asdacap); as it was updating mainchain post the changing block but not including the changing block as part of branch update

benaadams:
Made new bugs :(

src/Nethermind/Nethermind.Blockchain/BlockhashProvider.cs
asdacap: Think about it, what if this block is in a long branch? What branch is `_blockTree.FindBlockHash` running on?

benaadams: `_blockTree.FindBlockHash` should be the one the evm is running on

benaadams: You mean if no UpdateMainChain?

benaadams: Fixed the legacy blockhashes by adding it back as switch; but this test is cursed

benaadams: @asdacap better?

benaadams:
> * On Head change actively scan through cache, double check and ensure we have all the items for Blockhash opcode? (unless a long reorg you won't actually hit DB!)

Not sure how this would help? (adding 256 blockhash refreshes to every block)

src/Nethermind/Nethermind.Blockchain/BlockhashProvider.cs
asdacap: Do you really want to rely on a completely correct concurrent canonical block info updates for this use case?

benaadams: You think there is a situation where the main chain is being changed while validating a mainchain block in the evm?

asdacap: Executing downloaded blocks while FCU, yes.

PR #9574, opened 2025-10-27
Fix flaky test in SimpleRetryCacheTests by pooling instead of delay

LukaszRozmej:

- [x] Refactoring
- [x] Other: tests

## Testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/V68/Eth68ProtocolHandler.cs
+++ b/src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/V68/Eth68ProtocolHandler.cs
@@ -253,6 +253,7 @@ public class Eth68ProtocolHandler(ISession session,
         {
             if (!ValidateSizeAndType(request.txs[i]))
             {
+                request.txs.Dispose();
                 throw new SubprotocolException("invalid pooled tx type or size");
             }
         }
--- a/src/Nethermind/Nethermind.TxPool.Test/RetryCacheTests.cs
+++ b/src/Nethermind/Nethermind.TxPool.Test/RetryCacheTests.cs
@@ -13,7 +13,7 @@ using Nethermind.Core.Test;
 namespace Nethermind.TxPool.Test;
 
 [TestFixture]
-public class SimpleRetryCacheTests
+public class RetryCacheTests
 {
     public readonly struct ResourceRequestMessage : INew<int, ResourceRequestMessage>
     {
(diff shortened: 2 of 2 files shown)


src/Nethermind/Nethermind.Core.Test/NSubstituteExtensions.cs
```diff
@@ -0,0 +1,36 @@
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System;
+using NSubstitute;
+
+namespace Nethermind.Core.Test;
+
+public static class NSubstituteExtensions
```
flcl42, line 9: Will it raise like 25 exceptions per assert for these particular tests?

src/Nethermind/Nethermind.TxPool.Test/SimpleRetryCacheTests.cs
```diff
@@ -22,11 +23,10 @@ public class SimpleRetryCacheTests
 
     public interface ITestHandler : IMessageHandler<ResourceRequestMessage>;
 
-
     private CancellationTokenSource _cancellationTokenSource;
     private RetryCache<ResourceRequestMessage, int> cache;
 
-    private readonly int Timeout = 3000;
+    private readonly int Timeout = 5000;
```
flcl42: Simpler would be to just have timeout = 1000; and timeoutMs passed to RetryCache ctor to be half of that

flcl42: It's flaky most likely because retries happen 5 times RetryCache's timeout and there was not enough time for the first retry after that internal timeout

src/Nethermind/Nethermind.Core.Test/NSubstituteExtensions.cs
LukaszRozmej: depends when it ends, yes in worst case. We can reduce pooling rate to reduce that

src/Nethermind/Nethermind.TxPool.Test/SimpleRetryCacheTests.cs
LukaszRozmej: feel free to fix - it is your code

flcl42: Added a couple of fixes

PR #9580, opened 2025-10-27
Optimise CREATE(2) by removing redundant check

Marchhill:
## Changes

- Remove redundant check for dead account to clear storage. This is safe to do as we implement [EIP-7610](https://eips.ethereum.org/EIPS/eip-7610), so the contract deployment would have already failed if the address had storage
- Microoptimisations

- [x] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Evm/AddressExtensions.cs
+++ b/src/Nethermind/Nethermind.Evm/AddressExtensions.cs
@@ -44,9 +44,8 @@ namespace Nethermind.Evm
         // See https://eips.ethereum.org/EIPS/eip-7610
         public static bool IsNonZeroAccount(this Address contractAddress, IReleaseSpec spec, ICodeInfoRepository codeInfoRepository, IWorldState state)
         {
-            return state.GetNonce(contractAddress) != 0 ||
-                   state.GetBalance(contractAddress) != 0 ||
-                   codeInfoRepository.GetCachedCodeInfo(contractAddress, spec).CodeSpan.Length != 0 ||
+            return codeInfoRepository.GetCachedCodeInfo(contractAddress, spec).CodeSpan.Length != 0 ||
+                   state.GetNonce(contractAddress) != 0 ||
                    !state.IsStorageEmpty(contractAddress);
         }
     }


LukaszRozmej (approved):

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Create.cs
```diff
@@ -215,12 +215,6 @@ public static EvmExceptionType InstructionCreate<TOpCreate, TTracingInst>(
             goto None;
         }
 
-        // If the contract address refers to a dead account, clear its storage before creation.
-        if (state.IsDeadAccount(contractAddress))
-        {
-            state.ClearStorage(contractAddress);
-        }
-
```
tanishqjasoria, line 223: what about archive sync? when the eip was not enabled?

LukaszRozmej: Add spec check for eip then

marcindsobczak: If I understand it correctly, we will not reach this code if storage is non-empty, because it will fail earlier (one check before, line 211, on `contractAddress.IsNonZeroAccount()` and `goto None`.

So if archive sync is working right now, it should still work after this change, as `state.ClearStorage(contractAddress)` is reachable only if storage was already empty

marcindsobczak (approved):
From what I understand, `state.ClearStorage()` is already unreachable if state is non-empty, so this PR is removing indeed redundant check.

The better question is, why our archive sync is working fine when we already have this kind of bug? Clearing non-empty storage never happened in history?

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Create.cs
Marchhill: The EIP was never enabled since it just clarifies some undefined behaviour

Marchhill:
> Any way to test it?

Not really as the code is unreachable

PR #9581, opened 2025-10-27
Optimise CALL by throwing stack underflow earlier

Marchhill:
## Changes

- Pop from stack at start of instruction to fail faster without having to access state etc. Also better matches spec

- [x] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

No regressions in hive consensus tests.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Evm.Test/CallTests.cs
+++ b/src/Nethermind/Nethermind.Evm.Test/CallTests.cs
@@ -1,0 +1,55 @@
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using Nethermind.Specs;
+using NUnit.Framework;
+
+namespace Nethermind.Evm.Test
+{
+    public class CallTests : VirtualMachineTestsBase
+    {
+        protected override long BlockNumber => MainnetSpecProvider.ParisBlockNumber;
+        protected override ulong Timestamp => MainnetSpecProvider.OsakaBlockTimestamp;
+
+        [Test]
+        [TestCase(Instruction.CALL)]
+        [TestCase(Instruction.CALLCODE)]
+        [TestCase(Instruction.DELEGATECALL)]
+        [TestCase(Instruction.STATICCALL)]
+        public void Stack_underflow_on_call(Instruction instruction)
+        {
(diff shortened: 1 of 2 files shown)


Marchhill:
> i think this would result in different access lists and then cause gas mismatch in nested calls.

Are you talking block access lists here? Because yes that's the main point of the change, the BAL would be incorrect without it. In the execution specs popping from the stack always happens at the start of the CALL: https://github.com/ethereum/execution-specs/blob/forks/osaka/src/ethereum/forks/osaka/vm/instructions/system.py#L365

Could you expand on why it could cause a gas mismatch?

tanishqjasoria (approved):
we revert access tracker when the call fails.

src/Nethermind/Nethermind.Evm.Test/CallTests.cs
```diff
@@ -0,0 +1,55 @@
...
+    {
+        protected override long BlockNumber => MainnetSpecProvider.ParisBlockNumber;
+        protected override ulong Timestamp => MainnetSpecProvider.OsakaBlockTimestamp;
+
+        [Test]
+        [TestCase(Instruction.CALL)]
+        [TestCase(Instruction.CALLCODE)]
+        [TestCase(Instruction.DELEGATECALL)]
+        [TestCase(Instruction.STATICCALL)]
+        public void Stack_underflow_on_call(Instruction instruction)
```
flcl42, line 19: This test fails on master. We need to repeat it in hive.
To be more clear: looks like consensus on execution is changed in this pr, and it's not revealed by tests common for ELs

Marchhill: yeah I added as a regression test. Do you mean we should add to hive suite? It's not critical as it just means one error is thrown instead of another

benaadams (approved):

src/Nethermind/Nethermind.Evm.Test/CallTests.cs
benaadams: Consensus would be the same with outofgas? Doesn't differentiate why outofgas?

PR #9587, opened 2025-10-28
Add field support to ExpressionExtensions.GetSetter

radik878:
Added support for fields in ExpressionExtensions.GetSetter by handling FieldInfo: it rejects static, readonly, and const fields, builds an assignment expression using Expression.Assign, and applies a type conversion when TProperty differs from the field type; the existing property-based setter path remains unchanged and exception messages follow the file’s current style


LukaszRozmej (approved):

LukaszRozmej:
Doesn't build

radik878:
> Doesn't build

fixed

PR #9588, opened 2025-10-29
Feature/triestore respect finalized block marker

asdacap:
- Causes TrieStore to check for finalized block before persisting state, causing it to persist canonical state only by default.
- If the finalized block is behind the in memory state, it will persist all state up until pruning boundary.
- This allow very long reorg that follows the FCU.
- To determine the finalized state, an interface `IFinalizedStateProvider` is created which return the last finalized state number and a method to find the state root for any state number.
   - For PoW network, it uses the reorg depth which is `IBlockTree.BestKnownNumber - ReorgDepth`. 
   - For PoS network, it uses either:
       - IBlockTree.FinalizedHash
       - or IBlockCacheService.FinalizedHash
       - This is because the `IBlockTree.FinalizedHash` is only updated on processed block. Which would cause forward sync to become archive mode. So the `IBlockCacheService.FinalizedHash` is used as a backup. Ironing out the 4 different place of finalization header source is outside of this PR's scope.

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- [x] Mainnet can sync
- [x] Mainnet era replay
- [x] Mainnet full pruning

### DIFF
--- a/src/Nethermind/Nethermind.Merge.Plugin/MergeFinalizedStateProvider.cs
+++ b/src/Nethermind/Nethermind.Merge.Plugin/MergeFinalizedStateProvider.cs
@@ -30,7 +30,7 @@ public class MergeFinalizedStateProvider(IPoSSwitcher poSSwitcher, IBlockCacheSe
                 if (blockCacheService.FinalizedHash is { } blockCacheFinalizedHash)
                 {
                     BlockHeader? fromBlockCache = blockTree.FindHeader(blockCacheFinalizedHash);
-                    if (fromBlockCache != null)
+                    if (fromBlockCache is not null)
                     {
                         if (currentFinalized is null || fromBlockCache.Number > currentFinalized.Number)
                         {


asdacap:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/NethermindEth/nethermind/pull/9588#pullrequestreview-3392124425)

src/Nethermind/Nethermind.Trie.Test/Pruning/TreeStoreTests.cs
```diff
@@ -1206,11 +1214,11 @@ void WriteRandomData(int seed)
             fullTrieStore.WaitForPruning();
 
             fullTrieStore.PrunePersistedNodes();
-            fullTrieStore.CachedNodesCount.Should().Be(52);
+            fullTrieStore.CachedNodesCount.Should().Be(65);
```
asdacap, line 1217: Pruning boundary previously acted like PruningBoundary + 1

src/Nethermind/Nethermind.Merge.Plugin/MergeFinalizedStateProvider.cs
```diff
@@ -0,0 +1,54 @@
...
+                {
+                    currentFinalized = blockTree.FindHeader(blockTreeFinalizedHash, BlockTreeLookupOptions.None);
+                }
+
+                // Finalized hash from blocktree is not updated until it is processed, which is a problem for long
+                // catchup. So we use from blockCacheService as a backup.
+                if (blockCacheService.FinalizedHash is { } blockCacheFinalizedHash)
+                {
+                    BlockHeader? fromBlockCache = blockTree.FindHeader(blockCacheFinalizedHash);
+                    if (fromBlockCache != null)
```
rubo:
```suggestion
                    if (fromBlockCache is not null)
```

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -691,41 +704,92 @@ private void SaveSnapshot()
...
         try
         {
+            if (_commitSetQueue.IsEmpty)
+            {
+                if (_logger.IsDebug) _logger.Debug("Unable to persist commit set due to empty queue");
+                return (candidateSets, null);
+            }
+
+            long finalizedBlockNumber = _finalizedStateProvider.FinalizedBlockNumber;
+            long pruningBoundaryBlockNumber = _commitSetQueue.MaxBlockNumber.Value - _maxDepth;
```
flcl42, line 725: Can it happen so finalized is older than pruning depth and some commit sets are not saved?

src/Nethermind/Nethermind.Merge.Plugin/MergeFinalizedStateProvider.cs
```diff
@@ -0,0 +1,54 @@
...
+            {
+                BlockHeader? currentFinalized = null;
+                if (blockTree.FinalizedHash is { } blockTreeFinalizedHash)
+                {
+                    currentFinalized = blockTree.FindHeader(blockTreeFinalizedHash, BlockTreeLookupOptions.None);
+                }
+
+                // Finalized hash from blocktree is not updated until it is processed, which is a problem for long
+                // catchup. So we use from blockCacheService as a backup.
+                if (blockCacheService.FinalizedHash is { } blockCacheFinalizedHash)
```
flcl42, line 30: It seems like blockCacheService should be hidden in blocktree

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
asdacap: That is the case in line 733.

src/Nethermind/Nethermind.Trie/Pruning/IFinalizedStateProvider.cs
```diff
@@ -0,0 +1,13 @@
...
+using Nethermind.Core;
+using Nethermind.Core.Crypto;
+
+namespace Nethermind.Trie.Pruning;
+
+public interface IFinalizedStateProvider
+{
+    long FinalizedBlockNumber { get; }
+    Hash256? GetFinalizedStateRootAt(long blockNumber);
+}
```
LukaszRozmej, line 13: We have `IBlockFinalizationManager` already, so this seems redundant?

src/Nethermind/Nethermind.Merge.Plugin/MergeFinalizedStateProvider.cs
```diff
@@ -0,0 +1,54 @@
...
+using Nethermind.Blockchain.Find;
+using Nethermind.Consensus;
+using Nethermind.Core;
+using Nethermind.Core.Crypto;
+using Nethermind.Merge.Plugin.Handlers;
+using Nethermind.Trie.Pruning;
+
+namespace Nethermind.Merge.Plugin;
+
+public class MergeFinalizedStateProvider(IPoSSwitcher poSSwitcher, IBlockCacheService blockCacheService, IBlockTree blockTree, IFinalizedStateProvider baseFinalizedStateProvider) : IFinalizedStateProvider
```
LukaszRozmej, line 14: We already have `MergeFinalizationManager`, shouldn't be needed

src/Nethermind/Nethermind.Trie/Pruning/CommitSetQueue.cs
```diff
@@ -0,0 +1,116 @@
...
+
+public class CommitSetQueue
+{
+    private SortedSet<BlockCommitSet> _queue = new();
+
+    public int Count
+    {
+        get
+        {
+            lock (_queue)
```
LukaszRozmej, line 20: Does the commit set queue even has multi-threaded access?
We have block processing queue so the write access should be single threaded?
Alternatively could use `ReaderWriterLockSlim` for multiple readers.

src/Nethermind/Nethermind.Trie/Pruning/CommitSetQueue.cs
```diff
@@ -0,0 +1,116 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using Nethermind.Core.Collections;
+using Nethermind.Core.Crypto;
+
+namespace Nethermind.Trie.Pruning;
+
+public class CommitSetQueue
+{
+    private SortedSet<BlockCommitSet> _queue = new();
```
LukaszRozmej, line 14: Won't `PriorityQueue` be faster?
Especially for Min/Max.

https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.priorityqueue-2?view=net-10.0

src/Nethermind/Nethermind.Trie/Pruning/IFinalizedStateProvider.cs
asdacap: I dont know, the `IBlockFinalizationManager` is redudant with `IBlockTree`.

src/Nethermind/Nethermind.Trie/Pruning/CommitSetQueue.cs
asdacap: It can write and prune at the same time.

src/Nethermind/Nethermind.Trie/Pruning/CommitSetQueue.cs
asdacap: Dont seems to have max. Need both min and max.

src/Nethermind/Nethermind.Trie/Pruning/IFinalizedStateProvider.cs
LukaszRozmej: BlockFinalizationManager:
- works on different consensus engines (maybe we need to adjust it for PoW)
- is already there

This is being redundant with it and doesn't have other consensus support.
So I insist on it.

src/Nethermind/Nethermind.Trie/Pruning/CommitSetQueue.cs
LukaszRozmej: ok can we go with `ReaderWriterLockSlim` or wouldn't it change anything?

flcl42 (approved):

src/Nethermind/Nethermind.Trie/Pruning/IFinalizedStateProvider.cs
damian-orzechowski: Think I agree with @LukaszRozmej here - extending `IBlockFinalizationManager` should be relatively easy

src/Nethermind/Nethermind.Trie/Pruning/CommitSetQueue.cs
asdacap: I dont know. But its not used a lot so its probably fine.

PR #9596, opened 2025-10-30
Xdc : Add XdcTestBlockchain for Module testing

Demuirgos:
Adds a lot of fixes and refactor for XDC. 

Some Changes to NMC modules, for inherited behavior:
src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.cs
src/Nethermind/Nethermind.Consensus/Producers/BlockProducerBase.cs

Also fixes an issue with BasicTestBlockchain unrelated to XDC.

### DIFF
--- a/src/Nethermind/Nethermind.Xdc.Test/XdcHeaderValidatorTests.cs
+++ b/src/Nethermind/Nethermind.Xdc.Test/XdcHeaderValidatorTests.cs
@@ -72,8 +72,9 @@ public class Tests
 
         //Invalid total difficulty
         blockHeaderBuilder = CreateValidHeader();
-        blockHeaderBuilder.WithDifficulty(2);
-        blockHeaderBuilder.WithTotalDifficulty(1);
+        blockHeaderBuilder
+            .WithDifficulty(1)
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Xdc.Test/XdcHeaderValidatorTests.cs
```diff
@@ -65,6 +65,17 @@ public static IEnumerable<object[]> HeaderTestCases()
...
 
+        //Invalid difficulty
+        blockHeaderBuilder = CreateValidHeader();
+        blockHeaderBuilder.WithDifficulty(2);
+        yield return [blockHeaderBuilder, false];
+
+        //Invalid total difficulty
+        blockHeaderBuilder = CreateValidHeader();
+        blockHeaderBuilder.WithDifficulty(2);
+        blockHeaderBuilder.WithTotalDifficulty(1);
```
cicr99: Difficulty is already invalid here, it'll return false no matter the total difficulty set.
Also these two calls to the builder can be made fluently 🙂

cicr99 (approved):

PR #9600, opened 2025-10-30
Refactor JsonRpcService.ExecuteAsync

benaadams:
## Changes

- Split JsonRpcService.ExecuteAsync into multiple methods as is of a confusing size and hard to workout where the action is
- More the non-async Parameter perpetration to own method
- Single catch and then handle the exceptions via pattern matching in own method
- Optimize DeserializeParameters

- [x] Optimization
- [x] Refactoring

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.JsonRpc/JsonRpcService.cs
+++ b/src/Nethermind/Nethermind.JsonRpc/JsonRpcService.cs
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 
 using System;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
@@ -398,7 +397,7 @@ public class JsonRpcService : IJsonRpcService
             {
                 if (!_reparseReflectionCache.TryGetValue(paramType, out bool reparseString))
                 {
-                    reparseString = CreateNewCacheEntry(paramType); 
+                    reparseString = CreateNewCacheEntry(paramType);
                 }
 
                 executionParam = reparseString


flcl42 (approved):

src/Nethermind/Nethermind.JsonRpc/JsonRpcService.cs
```diff
@@ -377,72 +414,36 @@ private void LogRequest(string methodName, JsonElement providedParameters, Expec
...
-                        state.hasMissing = true;
-                    }
-
-                    return state;
-                });
-            }
+        bool hasMissing = missingParamsCount != 0;
+        int i = 0;
+        var enumerator = providedParameters.EnumerateArray();
+        while (enumerator.MoveNext())
```
LukaszRozmej, line 455: no parallel?

src/Nethermind/Nethermind.JsonRpc/JsonRpcService.cs
```diff
@@ -363,8 +395,13 @@ private void LogRequest(string methodName, JsonElement providedParameters, Expec
         {
             if (providedParameter.ValueKind == JsonValueKind.String)
             {
-                JsonConverter converter = EthereumJsonSerializer.JsonOptions.GetConverter(paramType);
-                executionParam = converter.GetType().Namespace.StartsWith("System.", StringComparison.Ordinal)
+                bool reparseString = _reparseReflectionCache.GetOrAdd(paramType, static type =>
```
LukaszRozmej: Adds here will be sparse, maybe better normal Dictionary with GetValueRefOrNullRef and ReaderWriterLock?


ChatGpt says not to do `GetOrAdd`:

https://chatgpt.com/s/t_690346ed55188191beaf495ce5f621fd

LukaszRozmej: Claude 4.5 doesn't agree:

```
Yes, even with many, many reads, **ConcurrentDictionary can become competitive or even better once writes exceed ~5-10%** of operations. Here's why this is counterintuitive:

## The Problem with ReaderWriterLockSlim Under Write Pressure

When a write lock is requested:
1. **All new readers are blocked immediately** - they queue up waiting
2. The writer waits for existing readers to finish
3. Writer executes (exclusive access)
4. All queued readers can finally proceed

So even if you have 1000 reads and 10 writes (99% reads), those 10 writes will cause **reader pile-ups** where hundreds of readers get blocked waiting for the write lock.

## ConcurrentDictionary's Advantage

ConcurrentDictionary uses **fine-grained locking** (lock striping):
- Only the specific bucket being written to is locked
- Readers accessing other buckets proceed unimpeded
- Modern implementations use lock-free techniques for reads in many cases
- No global "stop the world" moment for writes

## The Crossover Point

The exact threshold depends on:
- **Contention level**: How many threads?
- **Lock hold time**: How long do writes take?
- **Hardware**: Cache coherency costs, core count

In my experience:
- **<5% writes, low contention**: Dictionary + RWLockSlim wins
- **5-10% writes, moderate contention**: Roughly equal, depends on specifics
- **>10% writes, high contention**: ConcurrentDictionary wins

## Still Read-Heavy Champions

If you truly have <1% writes and many readers, Dictionary + ReaderWriterLockSlim with `GetValueRefOrNullRef` is still king. The performance gap for pure reads is significant - often 2-3x faster.

**Bottom line**: "Writes are rare" matters less than "writes are rare *and* there's not much read contention when writes happen." If you have 100 threads hammering reads and occasional writes, the write-induced pile-ups hurt more than you'd expect.
```

src/Nethermind/Nethermind.JsonRpc/JsonRpcService.cs
```diff
@@ -106,6 +108,75 @@ private Task<JsonRpcResponse> ExecuteRequestAsync(JsonRpcRequest rpcRequest, Jso
...
+        }
+        bool returnImmediately = methodName != GetLogsMethodName;
+        Action? returnAction = returnImmediately ? null : () => _rpcModuleProvider.Return(methodName, rpcModule);
+        IResultWrapper? resultWrapper = null;
+        try
+        {
+            // Execute method
+            object invocationResult = hasMissing ?
+                method.MethodInfo.Invoke(rpcModule, parameters) :
+                method.Invoker.Invoke(rpcModule, new Span<object?>(parameters));
```
LukaszRozmej, line 132: Can we make it faster in any way? Compile to delegates?

src/Nethermind/Nethermind.JsonRpc/JsonRpcService.cs
benaadams: "In my experience" Claude  is such a bs merchant

src/Nethermind/Nethermind.JsonRpc/JsonRpcService.cs
benaadams: Path doesn't even get hit for new payload

src/Nethermind/Nethermind.JsonRpc/JsonRpcService.cs
benaadams: Will look into it the `method.Invoker` path is pretty fast (MethodInfo one is slow)
image

src/Nethermind/Nethermind.JsonRpc/JsonRpcService.cs
benaadams: Switched to dictionary swap

PR #9603, opened 2025-10-30
Catch index exceptions in RlpDecoders and pack it in RlpException

LukaszRozmej:
## Changes

- Creates base RlpDecoder classes.
- Makes them catch `IndexOutOfRandeException` and `ArgumentOutOfRangeException` and pack it in `RlpException`
- Makes it easier to catch them

- [x] Refactoring

#### Requires testing

- [x] Yes

- [x] No

#### Notes on testing

Should be covered by existing tests

- [x] No

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Optimism.Test/ReceiptDecoderTests.cs
+++ b/src/Nethermind/Nethermind.Optimism.Test/ReceiptDecoderTests.cs
@@ -16,7 +16,7 @@ public class ReceiptDecoderTests
         static OptimismTxReceipt TestNetworkEncodingRoundTrip(byte[] rlp, bool includesNonce, bool includesVersion)
         {
             OptimismReceiptMessageDecoder decoder = new();
-            OptimismTxReceipt decodedReceipt = decoder.Decode(new RlpStream(rlp), RlpBehaviors.SkipTypedWrapping);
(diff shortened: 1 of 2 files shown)


LukaszRozmej:
@copilot can you fix other Rlp decoders?
This needs changes like I did in `ReceiptMessageDecoder69` inheriting from `RlpValueDecoder` or `RlpStreamDecoder`

src/Nethermind/Nethermind.Serialization.Rlp/IRlpDecoder.cs
```diff
@@ -38,4 +40,51 @@ public static T Decode<T>(this IRlpValueDecoder<T> decoder, ReadOnlySpan<byte> b
...
+            }
+            catch (Exception e) when (e is IndexOutOfRangeException or ArgumentOutOfRangeException)
+            {
+                ThrowRlpException(e);
+                return default;
+            }
+        }
+
+        protected abstract T DecodeInternal(ref Rlp.ValueDecoderContext decoderContext, RlpBehaviors rlpBehaviors = RlpBehaviors.None);
+    }
```
LukaszRozmej, line 89: Main changes here

alexb5dh (approved):
Only don't like that this makes it impossible for a single class to implement multiple decoders.

LukaszRozmej:
> Only don't like that this makes it impossible for a single class to implement multiple decoders.

It is still possible, but you need to implement the catch stuff yourself

PR #9605, opened 2025-10-30
fix: validate LogIndex when deriving deposit sourceHash

Forostovec:
## Changes

- Add non-null and non-negative validation for log.LogIndex when computing sourceHash in DecodeDepositTransactionFromLogEvent().
- Replace (ulong)(log.LogIndex ?? 0) with (ulong)log.LogIndex.Value.
- Extend InvalidLogFormatTestCases() with cases where LogIndex is null and -1, both expected to throw ArgumentException.

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Optimism.Test/CL/DepositTransactionBuilderTest.cs
+++ b/src/Nethermind/Nethermind.Optimism.Test/CL/DepositTransactionBuilderTest.cs
@@ -145,6 +145,50 @@ public class DepositTransactionBuilderTest
             LogIndex = 0,
             BlockHash = SomeHash,
         };
+
+        yield return new LogEntryForRpc
+        {
+            Address = DepositAddress,
+            Topics =
+            [
+                DepositEvent.ABIHash,
+                new Hash256(SomeAddressA.Bytes.PadLeft(32)),
+                new Hash256(SomeAddressB.Bytes.PadLeft(32)),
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Optimism/CL/Derivation/DepositTransactionBuilder.cs
```diff
@@ -141,7 +141,9 @@ static Hash256 ComputeSourceHash(Hash256 l1BlockHash, ulong logIndex)
         if (version == DepositEvent.Version0)
         {
             var depositLogEventV0 = DepositLogEventV0.FromBytes(log.Data);
-            var sourceHash = ComputeSourceHash(log.BlockHash, (ulong)(log.LogIndex ?? 0)); // TODO: Unsafe cast with possible null;
+            if (log.LogIndex is null) throw new ArgumentException("Expected non-null log index");
+            if (log.LogIndex < 0) throw new ArgumentException($"Expected non-negative log index, got {log.LogIndex}");
```
deffrian, line 145: It can't be negative or null. I think we can just do (ulong)log.LogIndex!.Value and remove todo

PR #9607, opened 2025-10-30
Perf/StateSync/Cache node between state root change

asdacap:
- With statesync, when the sync take long enough, it need to change the state root that it is downloading. When that happen, it essentially loses all its progress, notably pending nodes that is waiting for its dependencies to be downloaded.
- This PR add a cache so that when the state root change, the pending nodes is cached so that if it is re-needed on later state root, it is fetched from the cache instead of being re-downloaded. 
- When simulating frequent state root change `--Sync.StateMinDistanceFromHead 16 --Sync.StateMaxDistanceFromHead 20`, cache hit amount to 30% of written nodes.

- [X] Optimization

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Testing

- [x] Mainnet snap sync

### DIFF
--- a/src/Nethermind/Nethermind.Synchronization/FastSync/TreeSync.cs
+++ b/src/Nethermind/Nethermind.Synchronization/FastSync/TreeSync.cs
@@ -10,6 +10,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Autofac.Features.AttributeFilters;
 using Nethermind.Blockchain;
+using Nethermind.Blockchain.Synchronization;
 using Nethermind.Core;
 using Nethermind.Core.Caching;
 using Nethermind.Core.Crypto;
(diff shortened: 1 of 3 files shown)


src/Nethermind/Nethermind.Synchronization/FastSync/TreeSync.cs
```diff
@@ -70,6 +71,8 @@ public class TreeSync : ITreeSync
         private Dictionary<StateSyncItem.NodeKey, HashSet<DependentItem>> _dependencies = new();
         private readonly LruKeyCache<StateSyncItem.NodeKey> _alreadySavedNode = new(AlreadySavedCapacity, "saved nodes");
         private readonly LruKeyCache<ValueHash256> _alreadySavedCode = new(AlreadySavedCapacity, "saved nodes");
+        private ConcurrentDictionary<StateSyncItem.NodeKey, byte[]> _previouslyPendingItems = new();
```
LukaszRozmej, line 74: This could be potentially normal `Dictionary` (or even `Frozen`):
- pros: faster reads
- cons: slower to create

asdacap: Mmm... need to be recreated from concurrent dictionary then. Must be concurrent for the recorder.

LukaszRozmej: Yes exactly this is the con, so will let you decide what is better.
Maybe make it `IDictionary<>` set it to `ConcurrentDictionary` first and create `Dictionary` or `Frozen` in the background?

Not sure if worth it at all.

PR #9610, opened 2025-10-30
Arbitrum - Refund on contract creation failure

svlachakis:
## Changes

Held gas is not being refunded if contact deployment fails on Arbitrum Sepolia sync and we need to extend the functionality in order to override it on Arbitrum Plugin side.

### DIFF
--- a/src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Create.cs
+++ b/src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Create.cs
@@ -88,12 +88,6 @@ internal static partial class EvmInstructions
         ref readonly ExecutionEnvironment env = ref vm.EvmState.Env;
         IWorldState state = vm.WorldState;
 
-        // Ensure the executing account exists in the world state. If not, create it with a zero balance.
-        if (!state.AccountExists(env.ExecutingAccount))
-        {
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -734,6 +734,7 @@ private int ExecuteEvmCall<TTracingInst>(
         FailContractCreate:
             if (Logger.IsTrace) Logger.Trace("Restoring state from before transaction");
             WorldState.Restore(snapshot);
+            gasConsumed = RefundOnFailContractCreation(tx, spec, opts, in VirtualMachine.TxExecutionContext.GasPrice);
```
LukaszRozmej, line 737: where was it set before?

svlachakis: Here if I'm not mistaken, because it fails contract creation and never goes to `Refund` https://github.com/NethermindEth/nethermind/blob/d483c1a7139378ffd8cdc621e9999e250472f6dc/src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs#L648

PR #9611, opened 2025-10-30
feat: Add configurable EIP-2935 ring buffer size

AnkushinDaniil:
Associated with [#356](https://github.com/NethermindEth/nethermind-arbitrum/pull/356/files#diff-f6a63299dd18461331066c30053ca3fd567b0d349b6b24cf15a80af5598bd578)

## Changes

  - Add `Eip2935RingBufferSize` property to `IReleaseSpec` interface with default implementation returning `Eip2935Constants.RingBufferSize` (8,191 blocks)
  - Add settable `Eip2935RingBufferSize` property to `ReleaseSpec` class to allow chains to override the default
  - Update `BlockhashStore.ApplyBlockhashStateChanges()` to use `spec.Eip2935RingBufferSize` instead of hardcoded `Eip2935Constants.RingBufferSize`
  - Update `BlockhashStore.GetBlockHashFromState()` to use spec-provided ring buffer size with fallback to default
  - Add `Eip2935RingBufferSize` to chainspec configuration pipeline:
    - `ChainSpecParamsJson` - JSON deserialization (nullable long)
    - `ChainParameters` - in-memory storage
    - `ChainSpecLoader` - loading with default fallback
    - `ChainSpecBasedSpecProvider` - assignment to spec instances

  ## Types of changes

  - [ ] Optimization
  - [ ] Refactoring
  - [ ] Documentation update
  - [ ] Build-related changes
  - [ ] Other: _Description_

  ## Testing

  #### Requires testing

  - [x] Yes
  - [ ] No

  - [ ] Yes
  - [x] No

  #### Notes on testing

  Existing EIP-2935 tests will verify backward compatibility (Ethereum mainnet chains continue using 8,191 blocks). Chain-specific testing will be done in downstream implementations (e.g., Arbitrum plugin)
  that utilize this configuration option.

  ## Documentation

  - [ ] Yes
  - [x] No

  - [x] Yes
  - [ ] No

  **Release Notes:**

  EIP-2935 ring buffer size is now configurable via chainspec. Chains can override the default 8,191 block ring buffer by adding `"eip2935RingBufferSize": <value>` to their chainspec parameters. This enables
   chains with different block times to maintain equivalent historical block hash storage windows.

  ## Remarks

  This change maintains full backward compatibility - all existing chainspecs without the `eip2935RingBufferSize` parameter will continue using Ethereum mainnet's default of 8,191 blocks via the interface
  default implementation and loader fallback logic.

  The primary use case is Arbitrum, which requires a ring buffer of 393,168 blocks to provide ~1 day of history at its 0.22 second block time (vs Ethereum's ~1 day at 12 second block time with 8,191 blocks).

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain/Blocks/BlockhashStore.cs
+++ b/src/Nethermind/Nethermind.Blockchain/Blocks/BlockhashStore.cs
@@ -30,7 +30,7 @@ public class BlockhashStore(ISpecProvider specProvider, IWorldState worldState)
         if (!worldState.IsContract(eip2935Account)) return;
 
         Hash256 parentBlockHash = blockHeader.ParentHash;
-        var parentBlockIndex = new UInt256((ulong)((blockHeader.Number - 1) % spec.Eip2935RingBufferSize));
+        UInt256 parentBlockIndex = new UInt256((ulong)((blockHeader.Number - 1) % spec.Eip2935RingBufferSize));
         StorageCell blockHashStoreCell = new(eip2935Account, parentBlockIndex);
         worldState.Set(blockHashStoreCell, parentBlockHash!.Bytes.WithoutLeadingZeros().ToArray());
     }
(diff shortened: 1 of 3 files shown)


src/Nethermind/Nethermind.Blockchain/Blocks/BlockhashStore.cs
```diff
@@ -41,11 +41,11 @@ public void ApplyBlockhashStateChanges(BlockHeader blockHeader, IReleaseSpec spe
     public Hash256? GetBlockHashFromState(BlockHeader currentHeader, long requiredBlockNumber, IReleaseSpec? spec)
```
flcl42: Can we make spec not nullable; use then spec.Eip2935RingBufferSize without ?? alternative?

src/Nethermind/Nethermind.Blockchain/Blocks/BlockhashStore.cs
```diff
@@ -41,11 +41,11 @@ public void ApplyBlockhashStateChanges(BlockHeader blockHeader, IReleaseSpec spe
...
     {
         if (requiredBlockNumber >= currentHeader.Number ||
-            requiredBlockNumber + Eip2935Constants.RingBufferSize < currentHeader.Number)
+            requiredBlockNumber + (spec?.Eip2935RingBufferSize ?? Eip2935Constants.RingBufferSize) < currentHeader.Number)
         {
             return null;
         }
-        var blockIndex = new UInt256((ulong)(requiredBlockNumber % Eip2935Constants.RingBufferSize));
+        var blockIndex = new UInt256((ulong)(requiredBlockNumber % (spec?.Eip2935RingBufferSize ?? Eip2935Constants.RingBufferSize)));
```
flcl42: Could you use explicit type instead of var

flcl42 (changes requested):
Might be cool to have a test that tests non-standard setting is applied and used by blockhash store

flcl42 (approved):

PR #9613, opened 2025-10-30
Fix StatelessExecution tool build

rubo:
## Changes

- Removed target framework moniker from the project file. It must be taken from the Directory.Builds.props. Otherwise, build fails on framework upgrade.
- Reordered projects in workflow to be alphabetical

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [x] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

### DIFF
--- a/tools/StatelessExecution/StatelessExecution.csproj
+++ b/tools/StatelessExecution/StatelessExecution.csproj
@@ -2,9 +2,6 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net9.0</TargetFramework>
-    <ImplicitUsings>enable</ImplicitUsings>
-    <Nullable>enable</Nullable>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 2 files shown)


tools/StatelessExecution/StatelessExecution.csproj
```diff
@@ -2,9 +2,6 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net9.0</TargetFramework>
-    <ImplicitUsings>enable</ImplicitUsings>
-    <Nullable>enable</Nullable>
```
deffrian, line 7: Not sure why nullable removed here

deffrian: and implicit usings also

rubo: [This](https://github.com/NethermindEth/nethermind/blob/fix/statelessexec-tool/tools/Directory.Build.props#L6-L7) is why

PR #9614, opened 2025-10-30
Fix/pow forward header provider cache ownership

radik878:
## Changes

- In PowForwardHeaderProvider.AssembleResponseFromLastResponseBatch() return a pooled copy via ToPooledList(newResponse.Count) while keeping LastResponseBatch as the internal cache to avoid use-after-dispose.
- Added Cache_block_headers_with_disposal to ForwardHeaderProviderTests.cs to call GetBlockHeaders(...) twice with disposal and assert only a single underlying ISyncPeer.GetBlockHeaders(...) call occurs (cache path).

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

ForwardHeaderProviderTests.Cache_block_headers_with_disposal validates that two consecutive calls with using do not trigger a second peer header request when the cache is valid/large enough.


asdacap (approved):
Thank you for your contribution

PR #9618, opened 2025-10-31
Remove redundant WithAttributeFiltering() in PseudoNetworkModule

phrwlk:
The AddAdvance() helper already applies CommonNethermindConfig(), which includes .WithAttributeFiltering(). The extra .WithAttributeFiltering() on ProtocolsManager  registration in PseudoNetworkModule.cs was redundant and removed.


phrwlk:
> ## Pull Request Overview
> This PR removes a redundant `.WithAttributeFiltering()` call from the `ProtocolsManager` registration in the test module. The call is unnecessary because the `AddAdvance<T>` method internally applies `CommonNethermindConfig()`, which already includes `.WithAttributeFiltering()`.
> 
> * Simplified DI registration by removing duplicate attribute filtering configuration
> 
> Comments suppressed due to low confidence (1)

ok, copilot

PR #9621, opened 2025-11-01
Optimize BN254Pairing call

benaadams:
## Changes

- Remove double copy
- Use intrinsic endian swap
- Using `mclBn_millerLoop` inside the loop and one `mclBn_finalExp` at the end removes N‑1 final exponentiations, which are the expensive part. On BN254 that is a meaningful win at even modest N.

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Evm.Precompiles/BN254.cs
+++ b/src/Nethermind/Nethermind.Evm.Precompiles/BN254.cs
@@ -61,11 +61,8 @@ internal static unsafe class BN254
             if (!DeserializeG1(data, out mclBnG1 x))
                 return false;
 
-            byte* tmp = stackalloc byte[32];
-            CopyReverse32((data + chunkSize), tmp); // To little-endian
-
             Unsafe.SkipInit(out mclBnFr y);
-            if (mclBnFr_setLittleEndianMod(ref y, (nint)tmp, 32) == -1 || mclBnFr_isValid(y) == 0)
+            if (mclBnFr_setBigEndianMod(ref y, (nint)data + chunkSize, 32) == -1 || mclBnFr_isValid(y) == 0)
                 return false;
 
             mclBnG1_mul(ref x, x, y);  // x *= y
@@ -164,7 +161,7 @@ internal static unsafe class BN254
         if (mclBnFp_deserialize(ref point.x, (nint)tmp, chunkSize) == nuint.Zero)
             return false;
         // y
-        CopyReverse32((data + chunkSize), tmp);
+        CopyReverse32(data + chunkSize, tmp);
         if (mclBnFp_deserialize(ref point.y, (nint)tmp, chunkSize) == nuint.Zero)
             return false;
 
@@ -194,17 +191,17 @@ internal static unsafe class BN254
             return false;
 
         // x.re
-        CopyReverse32((data + chunkSize), tmp);
+        CopyReverse32(data + chunkSize, tmp);
         if (mclBnFp_deserialize(ref point.x.d0, (nint)tmp, chunkSize) == nuint.Zero)
             return false;
 
         // y.im
-        CopyReverse32((data + chunkSize * 2), tmp);
+        CopyReverse32(data + chunkSize * 2, tmp);
         if (mclBnFp_deserialize(ref point.y.d1, (nint)tmp, chunkSize) == nuint.Zero)
             return false;
 
         // y.re
-        CopyReverse32((data + chunkSize * 3), tmp);
+        CopyReverse32(data + chunkSize * 3, tmp);
         if (mclBnFp_deserialize(ref point.y.d0, (nint)tmp, chunkSize) == nuint.Zero)
             return false;
 


src/Nethermind/Nethermind.Evm.Precompiles/BN254PairingPrecompile.cs
```diff
@@ -36,15 +36,11 @@ public long DataGasCost(ReadOnlyMemory<byte> inputData, IReleaseSpec releaseSpec
             return IPrecompile.Failure;
         }
 
-        var input = ArrayPool<byte>.Shared.Rent(inputData.Length);
-        Span<byte> output = stackalloc byte[32];
+        byte[] output = new byte[32];
+        Memory<byte> input = Unsafe.As<ReadOnlyMemory<byte>, Memory<byte>>(ref inputData);
```
LukaszRozmej: It is mutated

LukaszRozmej (changes requested):
can't mutate input

benaadams:
> can't mutate input

Fixed

src/Nethermind/Nethermind.Evm.Precompiles/BN254.cs
```diff
@@ -74,109 +82,126 @@ internal static bool CheckPairing(Span<byte> input, Span<byte> output)
...
         {
             output[31] = 1;
             return true;
         }
 
-        return mclBnGT_isValid(gt) == 1;
+        // Single final exponentiation for the product
+        mclBn_finalExp(ref acc, acc);
+
+        // True iff the product of pairings equals 1 in GT
```
LukaszRozmej:
```suggestion
        // True if the product of pairings equals 1 in GT
```

src/Nethermind/Nethermind.Evm.Precompiles/BN254.cs
```diff
@@ -203,9 +228,97 @@ private static bool SerializeG1(in mclBnG1 point, Span<byte> output)
                 return false;
         }
 
-        x.Reverse(); // To big-endian
-        y.Reverse(); // To big-endian
+        CopyReverse32(x, x); // To big-endian
```
rubo: The `Reverse` method is already hardware-accelerated. Why is this?

benaadams: So isn't modifying the input

rubo: It didn't before too ;)

benaadams: Before it rented and returned from arraypool; copied the data in and out; all of which is removed

benaadams:
> Apart from using Miller loop/final exp combo, other changes don't bring any performance benefits according to my benchmarks, just made the code less readable:

Better?

Though it was more about BN254Pairing than Add or Mul

src/Nethermind/Nethermind.Evm.Precompiles/BN254.cs
benaadams: Better now?

src/Nethermind/Nethermind.Evm.Precompiles/BN254.cs
```diff
@@ -187,25 +201,182 @@ private static bool DeserializeG2(Span<byte> data, out mclBnG2 point)
...
+            (byte)19, (byte)18, (byte)17, (byte)16,
+            (byte)15, (byte)14, (byte)13, (byte)12,
+            (byte)11, (byte)10, (byte)9, (byte)8,
+            (byte)7, (byte)6, (byte)5, (byte)4,
+            (byte)3, (byte)2, (byte)1, (byte)0);
+
+        Vector256<byte> revInLane = Avx2.Shuffle(vec, mask);
+        Vector256<byte> fullRev = Avx2.Permute2x128(revInLane, revInLane, 0x01);
+        Unsafe.WriteUnaligned(dstRef, fullRev);
+    }
```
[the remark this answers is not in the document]
benaadams: Jit does that

src/Nethermind/Nethermind.Evm.Precompiles/BN254.cs
```diff
@@ -74,109 +89,116 @@ internal static bool CheckPairing(Span<byte> input, Span<byte> output)
...
-    private static bool DeserializeG1(Span<byte> data, out mclBnG1 point)
+    [SkipLocalsInit]
+    private static bool DeserializeG1(byte* data, out mclBnG1 point)
     {
+        const int chunkSize = 32;
+
         point = default;
 
-        // Check for all-zero data
-        if (data.IndexOfAnyExcept((byte)0) == -1)
```
rubo, line 119: Not an overkill? Replacing this already hardware-accelerated one-liner doesn't seem to buy us any performance gains.

src/Nethermind/Nethermind.Evm.Precompiles/BN254.cs
```diff
@@ -74,109 +89,116 @@ internal static bool CheckPairing(Span<byte> input, Span<byte> output)
...
+        if (IsZero64(data))
             return true;
 
-        Span<byte> x = data[0..32];
-        x.Reverse(); // To little-endian
-
-        fixed (byte* ptr = &MemoryMarshal.GetReference(x))
+        // Input is big-endian; MCL call below expects little-endian byte order for Fp
+        Span<byte> tmp = stackalloc byte[chunkSize];
+        fixed (byte* p = &MemoryMarshal.GetReference(tmp))
```
rubo: The `fixed` statement seems redundant: `byte* p = stackalloc byte[chunkSize];`

src/Nethermind/Nethermind.Evm.Precompiles/BN254.cs
```diff
@@ -74,109 +89,116 @@ internal static bool CheckPairing(Span<byte> input, Span<byte> output)
...
-        Span<byte> x1 = data[0..32];
-        x0.Reverse(); // To little-endian
-        x1.Reverse(); // To little-endian
-
-        fixed (byte* ptr0 = &MemoryMarshal.GetReference(x0))
-        fixed (byte* ptr1 = &MemoryMarshal.GetReference(x1))
+        // Input layout: x_im, x_re, y_im, y_re (each 32 bytes, big-endian)
+        // MCL Fp2 layout: d0 = re, d1 = im
+        Span<byte> tmp = stackalloc byte[chunkSize];
+        fixed (byte* p = &MemoryMarshal.GetReference(tmp))
```
rubo: Same here: `byte* p = stackalloc byte[chunkSize];`

src/Nethermind/Nethermind.Crypto/Ripemd.cs
rubo, line 1: I'd recommend moving RIPEMD changes to another PR, so they are easily revertable/trackable.

rubo:
> Better?

Doesn't look like they made any difference

> Though it was more about BN254Pairing than Add or Mul

Then why are the changes given no performance gain but a negligible loss (at least on my machine)?

src/Nethermind/Nethermind.Crypto/Ripemd.cs
benaadams: https://github.com/NethermindEth/nethermind/pull/9627

LukaszRozmej (approved):

src/Nethermind/Nethermind.Evm.Precompiles/BN254.cs
benaadams: Done

src/Nethermind/Nethermind.Evm.Precompiles/BN254.cs
benaadams: DOne

src/Nethermind/Nethermind.Evm.Precompiles/BN254.cs
benaadams: Doesn't work over pointers

benaadams:
Added change to annotate some lightweight methods with `SuppressGCTransition ` https://github.com/NethermindEth/mcl-bindings/pull/8

benaadams:
> Well, there are lots of changes bringing no performance gain but a negligible loss (at least on my machine)?

Better now? Removes one stackalloc entirely as well as .ToArray and moves other stackalloc out of common path

image

benaadams:
Though is more about BN254Pairing than Mul and Add

LukaszRozmej (approved):

rubo:
Overall, the situation is the following:

Pairing check:

```
| Method      | Mean     | Error     | StdDev    |
|------------ |---------:|----------:|----------:|
| 'Before PR' | 3.272 ms | 0.0349 ms | 0.0326 ms |
| 'After PR'  | 2.210 ms | 0.0297 ms | 0.0278 ms |
```
where almost all the improvement comes from the Miller loop/final exp combo:

```
| Method      | Mean     | Error     | StdDev    |
|------------ |---------:|----------:|----------:|
| 'Before PR' | 2.356 ms | 0.0362 ms | 0.0339 ms |
| 'After PR'  | 2.232 ms | 0.0400 ms | 0.0521 ms |
```

Thus, all other bells and whistles add just ~5% improvement.

For add:

```
| Method      | Mean     | Error   | StdDev  |
|------------ |---------:|--------:|--------:|
| 'Before PR' | 857.3 ns | 4.05 ns | 3.59 ns |
| 'After PR'  | 861.6 ns | 6.65 ns | 5.89 ns |
```

On each run, the changes are slower, although slightly.

The same for the mul:

```
| Method      | Mean     | Error    | StdDev   |
|------------ |---------:|---------:|---------:|
| 'Before PR' | 21.43 us | 0.235 us | 0.220 us |
| 'After PR'  | 22.06 us | 0.346 us | 0.324 us |
```

rubo:
> Though is more about BN254Pairing than Mul and Add

Then maybe better to leave them intact? They're clearly worse now.

> Added change to annotate some lightweight methods with `SuppressGCTransition`

That's interesting, but that doesn't seem to make any difference in this particular case.

benaadams:
> > Though is more about BN254Pairing than Mul and Add
> 
> Then maybe better to leave them intact? They're clearly worse now.

Are you using 96 byte inputs for Mul and 128 byte inputs for Add in the benchmarks?

benaadams:
> that doesn't seem to make any difference in this particular case.

Raised issue https://github.com/dotnet/runtime/issues/121288

src/Nethermind/Nethermind.Evm.Precompiles/BN254.cs
```diff
@@ -21,50 +25,62 @@ static BN254()
...
+        fixed (byte* data = &MemoryMarshal.GetReference(input))
         {
-            if (mclBnFr_setLittleEndianMod(ref y, (nint)ptr, 32) == -1 || mclBnFr_isValid(y) == 0)
+            if (!DeserializeG1(data, out mclBnG1 x))
                 return false;
-        }
 
-        mclBnG1_mul(ref x, x, y);  // x *= y
-        mclBnG1_normalize(ref x, x);
+            byte* tmp = stackalloc byte[32];
```
Marchhill: more specific name eg. `yData`. Also for other `tmp` where applicable

src/Nethermind/Nethermind.Evm.Precompiles/BN254.cs
```diff
@@ -21,50 +25,62 @@ static BN254()
...
+            CopyReverse32((data + chunkSize), tmp); // To little-endian
```
Marchhill: extra parens

src/Nethermind/Nethermind.Evm.Precompiles/BN254.cs
```diff
@@ -74,138 +90,307 @@ internal static bool CheckPairing(Span<byte> input, Span<byte> output)
...
+
+            for (int i = 0; i < input.Length; i += PairSize)
+            {
+                if (!DeserializeG1(data + i, out mclBnG1 g1))
+                    return false;
+
+                if (!DeserializeG2(data + i + 64, out mclBnG2 g2))
+                    return false;
+
+                // Skip explicit neutral pairs
```
Marchhill:
```suggestion
                // Skip if g1 or g2 are zero
```

src/Nethermind/Nethermind.Evm.Precompiles/BN254.cs
```diff
@@ -74,138 +90,307 @@ internal static bool CheckPairing(Span<byte> input, Span<byte> output)
...
+                {
+                    mclBnGT_mul(ref acc, acc, ml);
+                }
+                else
+                {
+                    hasMl = true;
+                }
+            }
+
+            // No effective pairs -> valid
```
Marchhill:
```suggestion
            // All pairs had zero element -> valid
```

src/Nethermind/Nethermind.Evm.Precompiles/BN254.cs
```diff
@@ -74,138 +90,307 @@ internal static bool CheckPairing(Span<byte> input, Span<byte> output)
...
+            {
+                output[31] = 1;
+                return true;
+            }
+
+            // Single final exponentiation for the product
+            mclBn_finalExp(ref acc, acc);
+
+            // True if the product of pairings equals 1 in GT
+            output[31] = (byte)(mclBnGT_isOne(acc) == 1 ? 1 : 0);
```
Marchhill:
```suggestion
            output[31] = Convert.ToByte(mclBnGT_isOne(acc) == 1);
```

src/Nethermind/Nethermind.Evm.Precompiles/BN254.cs
```diff
@@ -74,138 +90,307 @@ internal static bool CheckPairing(Span<byte> input, Span<byte> output)
...
-        return mclBnGT_isValid(gt) == 1;
+        // Input is big-endian; MCL call below expects little-endian byte order for Fp
+        byte* tmp = stackalloc byte[chunkSize];
+
+        // x
+        CopyReverse32(data, tmp);
+        if (mclBnFp_deserialize(ref point.x, (nint)tmp, chunkSize) == nuint.Zero)
+            return false;
+        // y
+        CopyReverse32((data + chunkSize), tmp);
```
Marchhill: parens

src/Nethermind/Nethermind.Evm.Precompiles/BN254.cs
```diff
@@ -74,138 +90,307 @@ internal static bool CheckPairing(Span<byte> input, Span<byte> output)
...
         }
 
-        Span<byte> y0 = data[96..128];
-        Span<byte> y1 = data[64..96];
-        y0.Reverse(); // To little-endian
-        y1.Reverse(); // To little-endian
+        return true;
+    }
+
+    private static unsafe bool IsZero64(byte* ptr)
```
Marchhill, line 232: Could these methods be moved to somewhere they can be reused, like byte array extensions or something

benaadams: Not sure want to encourage pointers  😅

Will move if crops up again

PR #9651, opened 2025-11-05
State related suggestions

flcl42:
Potential state related fixes

## Changes

- Fix `Will_NotRemove_ReCommittedNode` that previously saved random nodes on the edge of pruning only due to no persistence and it could happen that not all nodes were in the db in the end
- Safer `BeginScope`, no issue found, but why not
- Wider try catch, that potentially cleans up after early PreWarmTransactions

- [x] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No, edited tests


LukaszRozmej (approved):

src/Nethermind/Nethermind.Trie.Test/Pruning/TreeStoreTests.cs
```diff
@@ -1003,6 +1003,8 @@ public async Task Will_NotRemove_ReCommittedNode()
                 await Task.Delay(TimeSpan.FromMilliseconds(10));
             }
 
+            fullTrieStore.PersistCache(default);
```
asdacap: This defeat the purpose. `PersistCache` will always persist everything. Try making sure that pruning always happen on line 1002.

asdacap: Like.... remove this.

flcl42: Done

PR #9655, opened 2025-11-05
Feature/xdc prevent finalized reorg

ak88:
Make BlockTree `BestSuggestedImprovementRequirementsSatisfied `virtual and prevent `SuggestBlock `processing a block that is not descendant of finalized block.

### DIFF
--- a/src/Nethermind/Nethermind.Xdc/XdcBlockTree.cs
+++ b/src/Nethermind/Nethermind.Xdc/XdcBlockTree.cs
@@ -68,7 +68,7 @@ internal class XdcBlockTree : BlockTree
                 return base.Suggest(block, header, options);
 
             current = FindHeader(current.ParentHash, BlockTreeLookupOptions.TotalDifficultyNotNeeded | BlockTreeLookupOptions.DoNotCreateLevelIfMissing);
-            if (current == null)
+            if (current is null)
                 return AddBlockResult.UnknownParent;
         }
         //This is not possible to reach


src/Nethermind/Nethermind.Xdc/XdcBlockTree.cs
```diff
@@ -0,0 +1,77 @@
...
+        [KeyFilter("metadata")] IDb? metadataDb,
+        IBadBlockStore? badBlockStore,
+        IChainLevelInfoRepository? chainLevelInfoRepository,
+        ISpecProvider? specProvider,
+        IBloomStorage? bloomStorage,
+        ISyncConfig? syncConfig,
+        ILogManager? logManager,
+        long genesisBlockNumber = 0) : base(blockStore, headerDb, blockInfoDb, metadataDb, badBlockStore, chainLevelInfoRepository, specProvider, bloomStorage, syncConfig, logManager, genesisBlockNumber)
+    {
+        this.xdcConsensus = xdcConsensus;
```
flcl42: @rubo is it possible to have some simple stuff being auto suggested in prs like:
- start private field names with underscore
- place spaces between methods/ifs
- wrap statements in blocks in if-else
- don't use vars

src/Nethermind/Nethermind.Xdc/XdcBlockTree.cs
```diff
@@ -0,0 +1,73 @@
...
+                return AddBlockResult.InvalidBlock;
+
+            if (finalizedBlockInfo.Hash == current.ParentHash)
+                return base.Suggest(block, header, options);
+
+            current = FindHeader(current.ParentHash, BlockTreeLookupOptions.TotalDifficultyNotNeeded | BlockTreeLookupOptions.DoNotCreateLevelIfMissing);
+            if (current == null)
+                return AddBlockResult.UnknownParent;
+            depth++;
+            if (depth == MaxSearchDepth)
```
LukaszRozmej: This could be potentially checked upfront rather than in a loop? Based on `finalizedBlockInfo.BlockNumber` and `header.Number`. Then loop can be `for (int i = header.Number; i >= finalizedBlockInfo.BlockNumber; i--)`

src/Nethermind/Nethermind.Xdc/XdcBlockTree.cs
```diff
@@ -0,0 +1,73 @@
...
+        BlockHeader current = header;
+        while (true)
+        {
+            if (finalizedBlockInfo.BlockNumber >= current.Number)
+                return AddBlockResult.InvalidBlock;
+
+            if (finalizedBlockInfo.Hash == current.ParentHash)
+                return base.Suggest(block, header, options);
+
+            current = FindHeader(current.ParentHash, BlockTreeLookupOptions.TotalDifficultyNotNeeded | BlockTreeLookupOptions.DoNotCreateLevelIfMissing);
```
LukaszRozmej, line 70: In theory we could also load multiple levels at once with `LoadLevel` and check that way more efficiently.
But we don't have a good caching strategy for levels, so they would probably always hit DB.
While we do have a good caching strategy for block headers.

So hard to say what is better. (Maybe we should cache levels better?)

ak88: I want to leave it as is for now, since this only affects XDC, and should only be a problem under extreme circumstances.

src/Nethermind/Nethermind.Xdc/XdcBlockTree.cs
```diff
@@ -0,0 +1,78 @@
...
+        for (long i = header.Number; i >= finalizedBlockInfo.BlockNumber; i--)
+        {
+            if (finalizedBlockInfo.BlockNumber >= current.Number)
+                return AddBlockResult.InvalidBlock;
+
+            if (finalizedBlockInfo.Hash == current.ParentHash)
+                return base.Suggest(block, header, options);
+
+            current = FindHeader(current.ParentHash, BlockTreeLookupOptions.TotalDifficultyNotNeeded | BlockTreeLookupOptions.DoNotCreateLevelIfMissing);
+            if (current == null)
```
LukaszRozmej:
```suggestion
            if (current is null)
```

LukaszRozmej (approved):

PR #9657, opened 2025-11-05
fix: serialize transaction before publishing in AnalyticsSteps

CreeptoGengar:
Convert Transaction to TransactionForRpc before publishing to ensure proper serialization for IPublisher implementations (LogPublisher, AnalyticsWebSocketsModule).

Resolves TODO comment by adding proper transaction serialization.


LukaszRozmej:
@NethermindEth/core Analytics module seems obsolete, should we remove it?
@benaadams @rubo @asdacap

rubo:
> @LukaszRozmej @rubo It's used as an example for plugin implementers

@flcl42 Where? We don't have that in our [docs](https://docs.nethermind.io/developers/plugins/). I doubt it's publicly linked anywhere.

PR #9664, opened 2025-11-07
Reduce storage tracking allocations

benaadams:
## Changes

- Pool the various data structures used

Before (830secs)

image

After (830secs)

image

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Core/Caching/StaticPool.cs
+++ b/src/Nethermind/Nethermind.Core/Caching/StaticPool.cs
@@ -26,7 +26,7 @@ public static class StaticPool<T> where T : class, IResettable, new()
     /// Global pool shared between threads.
     /// </summary>
     private static readonly ConcurrentQueue<T> _pool = [];
-    
+
     /// <summary>
     /// Manual count of items in the queue.
     /// We maintain this separately because ConcurrentQueue.Count


src/Nethermind/Nethermind.Core/Caching/StaticPoolcs.cs
```diff
@@ -0,0 +1,110 @@
...
+            throw new InvalidOperationException($"{typeof(T).Name} has no parameterless ctor.");
+
+        NewExpression newExpr = Expression.New(ci);
+        Expression<Func<T>> lambda = Expression.Lambda<Func<T>>(
+            newExpr,
+            name: $"Create{typeof(T).Name}",
+            tailCall: false,
+            parameters: Array.Empty<ParameterExpression>()
+        );
+        return lambda.Compile();
```
LukaszRozmej: `where T : new()`

and just use `new T()`

I think Activator is now fast? (was optimized in some new dotnet versions?)

benaadams: Done

src/Nethermind/Nethermind.Core/Caching/StaticPoolcs.cs
```diff
@@ -0,0 +1,87 @@
...
+        // Try to pop from the global pool — this is only hit when a thread
+        // has exhausted its own fast slot or is cross-thread renting.
+        if (_pool.TryDequeue(out item))
+        {
+            // We track count manually with Interlocked ops instead of using queue.Count.
+            Interlocked.Decrement(ref _poolCount);
+            return item;
+        }
+
+        // 3. Nothing available, allocate new instance via cached delegate.
```
LukaszRozmej:
```suggestion
        // 3. Nothing available, allocate new instance
```

src/Nethermind/Nethermind.State/Storage/DefaultableDictionary.cs
```diff
@@ -0,0 +1,68 @@
...
+        return ref value;
+    }
+
+    public ref StorageChangeTrace GetValueRefOrNullRef(UInt256 storageCellIndex)
+        => ref CollectionsMarshal.GetValueRefOrNullRef(_dictionary, storageCellIndex);
+
+    public StorageChangeTrace this[UInt256 key]
+    {
+        set => _dictionary[key] = value;
+    }
```
LukaszRozmej: Looks odd with others being methods, is it needed? Should we make it a method?

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
```diff
@@ -376,10 +373,19 @@ public void CommitTrees(IBlockCommitter blockCommitter)
...
     }
 
+    private void ClearAndPoolStorages()
+    {
+        foreach (PerContractState storage in _storages.Values)
+        {
+            storage.Reset();
+        }
+        _storages.Clear();
+    }
```
LukaszRozmej: This pattern could be an instance method (we would have to inherit Dictionary), or at least extension method

src/Nethermind/Nethermind.State/StateProvider.cs
```diff
@@ -953,6 +956,15 @@ public void Reset(bool resetBlockChanges = true)
...
         }
 
+        private void ClearIntraTxCache()
+        {
+            foreach (StackList<int> stack in _intraTxCache.Values)
+            {
+                StackList<int>.Return(stack);
+            }
+            _intraTxCache.Clear();
+        }
```
LukaszRozmej: Also used here

src/Nethermind/Nethermind.State/PartialStorageProviderBase.cs
```diff
@@ -202,21 +179,23 @@ protected virtual void CommitStorageRoots()
...
+        private void Reset()
         {
             if (_logger.IsTrace) _logger.Trace("Resetting storage");
 
             _changes.Clear();
+
+            foreach (StackList<int> stack in _intraBlockCache.Values)
+            {
+                StackList<int>.Return(stack);
+            }
```
LukaszRozmej: Also here

LukaszRozmej (commented):
Can we revert making things `internal`? They were better static and together.
If you want to have them in separate file use `partial` on the main class, but it makes it harder to review.

BTW 30MB in 15 mins is not much compared to total allocations we do.

LukaszRozmej:
Why remove that thread static?

LukaszRozmej:
Do we need a concurrent queue? EVM is single threaded, maybe it should be ThreadStatic Queue?

benaadams:
> Do we need a concurrent queue? EVM is single threaded, maybe it should be ThreadStatic Queue?

Isn't single threaded with prewarming (and later BALs)

benaadams:
> ObjectPool

Is abstract class thats then overengineered with a Provider and Policy

Also default implementation does have Interlocked on static as fast path; which as mentioned doesn't match usage

https://source.dot.net/#Microsoft.Extensions.ObjectPool/DefaultObjectPool.cs,48

src/Nethermind/Nethermind.State/PartialStorageProviderBase.cs
LukaszRozmej: This should be under method/extension method and used in all places (like at least 3)
```
            foreach (StackList<int> stack in _intraBlockCache.Values)
            {
                StackList<int>.Return(stack);
            }
            _intraBlockCache.Clear();
```

benaadams: Added extension

tanishqjasoria (approved):

src/Nethermind/Nethermind.Core/Caching/StaticPoolcs.cs
```diff
@@ -0,0 +1,55 @@
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System;
+using System.Collections.Concurrent;
+using System.Threading;
+
+namespace Nethermind.Core.Caching;
+
+public static class StaticPool<T> where T : class, new()
```
flcl42: It does not guarantee it will return clean items, seems a bit risky. It can additionally reset items

benaadams: Changed to reset in pool

PR #9667, opened 2025-11-07
Fix/rlp receipt limits

LukaszRozmej:
## Changes

- Revert multi-block limits to previous values
- Remove receipt limit (I confused multi-block receipt limit with single-block receipts limit). Default limit of 1M receipts should be fine.
- Fix default limit to be 1M instead of 256M

#### Requires testing

- [x] Yes

#### Notes on testing

Node sync and operations.

### DIFF
--- a/src/Nethermind/Nethermind.Network.Test/P2P/Subprotocols/Eth/V63/Eth63ProtocolHandlerTests.cs
+++ b/src/Nethermind/Nethermind.Network.Test/P2P/Subprotocols/Eth/V63/Eth63ProtocolHandlerTests.cs
@@ -128,7 +128,7 @@ namespace Nethermind.Network.Test.P2P.Subprotocols.Eth.V63
 
         private ArrayPoolList<T> RepeatPooled<T>(T txReceipts, int count) =>
             Enumerable.Repeat(txReceipts, count).ToPooledList(count).AddTo(_disposables);
-        
+
         [Test]
         public void Will_not_send_messages_larger_than_2MB()
         {


src/Nethermind/Nethermind.Network.Test/P2P/Subprotocols/Eth/V63/Eth63ProtocolHandlerTests.cs
```diff
@@ -128,18 +128,7 @@ public async Task Limit_receipt_request()
...
-        public void Will_not_serve_receipts_requests_above_512()
-        {
-            using GetReceiptsMessage getReceiptsMessage = new(RepeatPooled(Keccak.Zero, NethermindSyncLimits.MaxReceiptFetch + 1));
-            Packet getReceiptsPacket =
-                new("eth", Eth63MessageCode.GetReceipts, _ctx._getReceiptMessageSerializer.Serialize(getReceiptsMessage));
-
-            Assert.Throws<RlpLimitException>(() => _ctx.ProtocolHandler.HandleMessage(getReceiptsPacket));
-            _ctx.Session.Received().InitiateDisconnect(Arg.Any<DisconnectReason>(), Arg.Any<string>());
-        }
-
```
LukaszRozmej, line 142: IMO this is not needed, we do limit response to 2MB, so if someone asks for more receipts we just won't return more data.

marcindsobczak (approved):
Looks fine, I will test syncing locally to confirm if it is fixing snap sync issue

marcindsobczak:
Snap synced locally 2 times in a row without any issues, seems like it is a proper fix

PR #9669, opened 2025-11-07
Fix resource leak in JsonRpcService module disposal

kilavvy:
## Changes

- Fixed resource leak where RPC modules were not returned to the pool when exceptions occurred during context initialization
- Moved context assignment inside try block to ensure proper exception handling
- Added `responseCreated` flag to track response lifecycle and prevent double-return of modules
- Ensured modules are always returned in `finally` block when response is not created for `eth_getLogs` method


LukaszRozmej:
Returning of the module is handled through `returnAction`, which is correctly pass when exception happens.

PR #9671, opened 2025-11-08
Revise hive test scripts

rubo:
## Changes

- Set test parallelism to 4. Higher values fail tests.
- Improved color-coded representation. Removed duplicates.
- Prevented fails when JSON files not found

- [x] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [x] Other: Tests

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

Tested manually

### DIFF
--- a/scripts/hive-results.sh
+++ b/scripts/hive-results.sh
@@ -46,4 +46,6 @@ done
 echo -e "\n  Unexpected passes: ${#should_not_pass[@]}"
 echo "  Unexpected fails: ${#should_pass[@]}"
 
-(( ${#should_pass[@]} + ${#should_not_pass[@]} > 0 )) && exit 1 || exit 0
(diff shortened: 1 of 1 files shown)


.github/workflows/hive-tests.yml
```diff
@@ -29,6 +29,15 @@ on:
         default: '3'
         type: choice
         options: ['0', '1', '2', '3', '4', '5']
+      parallelism:
+        description: Parallelism
+        required: true
+        default: 4
```
LukaszRozmej, line 35: PR description says it is set to 6 but I see 4?

rubo: Sadly, turned out 6 is also error-prone for some tests. Updated the description.

PR #9672, opened 2025-11-08
fix: DynamicCallWithInput to skip value for DELEGATECALL/STATICCALL

Snezhkko:
DynamicCallWithInput in ByteCodeBuilder incorrectly pushed a zero “value” argument for DELEGATECALL and STATICCALL, which only accept six arguments (gas, to, in_offset, in_size, out_offset, out_size). This led to misaligned call parameters where in_offset and in_size were shifted and an extra zero remained on the stack after the call, effectively discarding input data and polluting the stack. Per EIP-7 and EIP-214, only CALL takes a value; DELEGATECALL and STATICCALL must not. This change pushes the value argument only when the callType is CALL, aligning with the EVM spec and the rest of the builder methods and preventing subtle bugs in dynamic calls.

### DIFF
--- a/src/Nethermind/Nethermind.Evm.Test/Eip1153Tests.cs
+++ b/src/Nethermind/Nethermind.Evm.Test/Eip1153Tests.cs
@@ -535,7 +535,7 @@ internal class Eip1153Tests : VirtualMachineTestsBase
             .Op(Instruction.CALLER)
             .PushData(TestItem.AddressD)
             .Op(Instruction.EQ)
-            .PushData(113)
+            .PushData(callType == Instruction.CALL ? 113 : 111)
             .Op(Instruction.JUMPI)
 
             // Non-reentrant, call self after TSTORE 8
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Evm/ByteCodeBuilder.cs
```diff
@@ -209,7 +209,10 @@ public Prepare DynamicCallWithInput(Instruction callType, Address address, long
             PushData(0);
             PushData(input is not null ? input.Length : 32);
             PushData(0);
-            PushData(0);
+            if (callType == Instruction.CALL)
```
benaadams, line 212: CALLCODE also?

benaadams: Ah not supported

benaadams (approved):

LukaszRozmej:
@Snezhkko Looks like test need updating

PR #9681, opened 2025-11-10
perf: avoid unnecessary allocations in EOACompressingDb Set/PutSpan

Forostovec:
This change eliminates unnecessary allocations in EOACompressingDb when compression does not apply. Previously Compress(byte[]?) always allocated by calling ToArray() even if the value did not end with the EOA suffix, causing a full copy and defeating PreferWriteByArray guidance. Now we short-circuit and return the original array when no compression is needed. Additionally, PutSpan previously did a stackalloc of value.Length unconditionally, which is both wasteful and risky for large values. We now only allocate a temporary buffer when compression is actually needed: using stackalloc for small payloads and ArrayPool for larger ones. This aligns with the IWriteOnlyKeyValueStore contract that stores may keep the input array directly, reduces CPU and GC pressure on common non-compressing paths, and keeps compression efficient and safe when it is required.
- Changed Compress(byte[]?) to avoid allocation when not compressing.
- Rewrote PutSpan in the wrapper and its batch to bypass buffers when not compressing, and to use stackalloc/ArrayPool only when compressing.
- Added using System.Buffers and a small stackalloc threshold.
- deleted the now-redundant Compress(ReadOnlySpan<byte>, Span<byte>) helper

### DIFF
--- a/src/Nethermind/Nethermind.Db/CompressingDb.cs
+++ b/src/Nethermind/Nethermind.Db/CompressingDb.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Buffers;
 using System.Runtime.CompilerServices;
 using Nethermind.Core;
 using Nethermind.Core.Extensions;
@@ -23,7 +22,6 @@ namespace Nethermind.Db
         private class EOACompressingDb : IDb, ITunableDb
         {
             private readonly IDb _wrapped;
-            private const int StackAllocThreshold = 512;
 
             public EOACompressingDb(IDb wrapped)
             {
@@ -100,27 +98,10 @@ namespace Nethermind.Db
 
                 int storedLength = value.Length - EmptyCodeHashStorageRoot.Length;
                 int compressedLength = storedLength + PreambleLength;
-
-                byte[]? rented = null;
-                scoped Span<byte> tmp;
-                if (compressedLength <= StackAllocThreshold)
-                {
-                    tmp = (stackalloc byte[StackAllocThreshold])[..compressedLength];
-                }
-                else
-                {
-                    rented = ArrayPool<byte>.Shared.Rent(compressedLength);
-                    tmp = rented.AsSpan(0, compressedLength);
-                }
-
+                scoped Span<byte> tmp = stackalloc byte[compressedLength];
                 tmp[PreambleIndex] = PreambleValue;
                 value[..storedLength].CopyTo(tmp[PreambleLength..]);
                 store.PutSpan(key, tmp, flags);
-
-                if (rented is not null)
-                {
-                    ArrayPool<byte>.Shared.Return(rented);
-                }
             }
 
             private static byte[]? Compress(byte[]? bytes)


src/Nethermind/Nethermind.Db/CompressingDb.cs
```diff
@@ -51,7 +53,37 @@ public void Set(ReadOnlySpan<byte> key, byte[]? value, WriteFlags flags = WriteF
...
+                            tmp[EOACompressingDb.PreambleIndex] = EOACompressingDb.PreambleValue;
+                            value[..storedLength].CopyTo(tmp[EOACompressingDb.PreambleLength..]);
+                            _wrapped.PutSpan(key, tmp, flags);
+                        }
+                        finally
+                        {
+                            ArrayPool<byte>.Shared.Return(rented);
+                        }
+                    }
                 }
```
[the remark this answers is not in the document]
LukaszRozmej: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/NethermindEth/nethermind/pull/9681#discussion_r2512501285)

src/Nethermind/Nethermind.Db/CompressingDb.cs
```diff
@@ -51,7 +53,37 @@ public void Set(ReadOnlySpan<byte> key, byte[]? value, WriteFlags flags = WriteF
...
+                    {
+                        byte[] rented = ArrayPool<byte>.Shared.Rent(compressedLength);
+                        try
+                        {
+                            Span<byte> tmp = rented.AsSpan(0, compressedLength);
+                            tmp[EOACompressingDb.PreambleIndex] = EOACompressingDb.PreambleValue;
+                            value[..storedLength].CopyTo(tmp[EOACompressingDb.PreambleLength..]);
+                            _wrapped.PutSpan(key, tmp, flags);
+                        }
+                        finally
```
benaadams: Drop the try/finally; doesn't really matter too much if don't return an array to the pool as they aren't tracked and it will just create a new one. Should also happen virtually 0% of times

So doesn't make sense to add the extra overhead of the try/finally block which will happen every time

src/Nethermind/Nethermind.Db/CompressingDb.cs
```diff
@@ -159,7 +184,37 @@ public void Set(ReadOnlySpan<byte> key, byte[]? value, WriteFlags flags = WriteF
...
+                {
+                    Span<byte> tmp = stackalloc byte[compressedLength];
+                    tmp[PreambleIndex] = PreambleValue;
+                    value[..storedLength].CopyTo(tmp[PreambleLength..]);
+                    _wrapped.PutSpan(key, tmp, flags);
+                }
+                else
+                {
+                    byte[] rented = ArrayPool<byte>.Shared.Rent(compressedLength);
+                    try
```
benaadams: Drop try/finally

src/Nethermind/Nethermind.Db/CompressingDb.cs
```diff
@@ -51,7 +53,37 @@ public void Set(ReadOnlySpan<byte> key, byte[]? value, WriteFlags flags = WriteF
...
+                        _wrapped.PutSpan(key, value, flags);
+                        return;
+                    }
+
+                    int storedLength = value.Length - EmptyCodeHashStorageRoot.Length;
+                    int compressedLength = storedLength + PreambleLength;
+
+                    if (compressedLength <= EOACompressingDb.StackAllocThreshold)
+                    {
+                        Span<byte> tmp = stackalloc byte[compressedLength];
```
benaadams: Stackalloc to fixed const `StackAllocThreshold` and slice to `compressedLength`; then add `[SkipLocalsInit]` to method

e.g.
```csharp
Span<byte> tmp = stackalloc byte[EOACompressingDb.StackAllocThreshold][..compressedLength];
```

src/Nethermind/Nethermind.Db/CompressingDb.cs
```diff
@@ -159,7 +184,37 @@ public void Set(ReadOnlySpan<byte> key, byte[]? value, WriteFlags flags = WriteF
...
+                    _wrapped.PutSpan(key, value, flags);
+                    return;
+                }
+
+                int storedLength = value.Length - EmptyCodeHashStorageRoot.Length;
+                int compressedLength = storedLength + PreambleLength;
+
+                if (compressedLength <= StackAllocThreshold)
+                {
+                    Span<byte> tmp = stackalloc byte[compressedLength];
```
benaadams: Also here

benaadams (approved):

benaadams:

src/Nethermind/Nethermind.Db/CompressingDb.cs
```diff
@@ -49,9 +52,38 @@ private class WriteBatch : IWriteBatch
...
+                    }
+
+                    tmp[PreambleIndex] = PreambleValue;
+                    value[..storedLength].CopyTo(tmp[PreambleLength..]);
+                    _wrapped.PutSpan(key, tmp, flags);
+
+                    if (rented is not null)
+                    {
+                        ArrayPool<byte>.Shared.Return(rented);
+                    }
```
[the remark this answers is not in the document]
LukaszRozmej: Exactly, please unify the code.

LukaszRozmej (changes requested):
Please unify the code

src/Nethermind/Nethermind.Db/CompressingDb.cs
```diff
@@ -114,6 +86,46 @@ public byte[]? this[ReadOnlySpan<byte> key]
...
             private const byte PreambleValue = 0;
 
+            private delegate void PutSpanWriter(ReadOnlySpan<byte> key, ReadOnlySpan<byte> value, WriteFlags flags);
+
+            [SkipLocalsInit]
+            private static void PutSpanCompressedIfNeeded(
+                ReadOnlySpan<byte> key,
+                ReadOnlySpan<byte> value,
+                WriteFlags flags,
+                PutSpanWriter put)
```
LukaszRozmej: 1. Avoid passing delegate.
2. Make method instance not static
3. Just call `_wrapped.PutSpan()` in the method

or just pass _wrapped

src/Nethermind/Nethermind.Db/CompressingDb.cs
```diff
@@ -82,29 +86,62 @@ public byte[]? this[ReadOnlySpan<byte> key]
...
 
-            private static byte[]? Compress(byte[]? bytes)
+            private delegate void PutSpanWriter(ReadOnlySpan<byte> key, ReadOnlySpan<byte> value, WriteFlags flags);
+
+            [SkipLocalsInit]
+            private static void PutSpanCompressedIfNeeded(
+                ReadOnlySpan<byte> key,
+                ReadOnlySpan<byte> value,
+                WriteFlags flags,
+                PutSpanWriter put)
```
benaadams: Can take `IWriteOnlyKeyValueStore` which both types are rather than creating a new delegate type

Forostovec: > Can take `IWriteOnlyKeyValueStore` which both types are rather than creating a new delegate type

Did it

src/Nethermind/Nethermind.Db/CompressingDb.cs
```diff
@@ -49,9 +52,10 @@ private class WriteBatch : IWriteBatch
                 public void Set(ReadOnlySpan<byte> key, byte[]? value, WriteFlags flags = WriteFlags.None)
                     => _wrapped.Set(key, Compress(value), flags);
 
+                [SkipLocalsInit]
```
LukaszRozmej:
```suggestion
```

src/Nethermind/Nethermind.Db/CompressingDb.cs
```diff
@@ -157,9 +192,10 @@ public void Set(ReadOnlySpan<byte> key, byte[]? value, WriteFlags flags = WriteF
                 => Decompress(_wrapped.Get(key, flags));
 
 
+            [SkipLocalsInit]
```
LukaszRozmej:
```suggestion
```

PR #9686, opened 2025-11-11
Feature/block production blob limit

### LINKED ISSUE #7502: Add max blob inclusion config for block production
Non-commercial Internet connections are often asymmetric with faster downloads than upload speeds

This means while a solo staker might be able to validate with a high number of blobs; they might not be able to produce blocks and upload the blobs in time (upload)

Add a configurable option to cap the number of blob transactions to include when producing blocks

LukaszRozmej:
## Changes

- Adds blon limit when producing blocks
- Replaces #7505
- Resolves #7502

#### Requires testing

- [x] Yes

- [x] Yes

### DIFF
--- a/src/Nethermind/Nethermind.Config/IBlocksConfig.cs
+++ b/src/Nethermind/Nethermind.Config/IBlocksConfig.cs
@@ -63,6 +63,6 @@ public interface IBlocksConfig : IConfig
 
     byte[] GetExtraDataBytes();
 
-    [ConfigItem(Description = "The blob limit after which the block producer should stop adding blobs. Minimum value is 0.", DefaultValue = "null")]
+    [ConfigItem(Description = "The max blob count after which the block producer should stop adding blobs. Minimum value is `0`.", DefaultValue = "null")]
     int? BlockProductionBlobLimit { get; set; }
 }


src/Nethermind/Nethermind.Evm/ReleaseSpecExtensions.cs
```diff
@@ -93,5 +93,10 @@ public static ulong GetMaxBlobGasPerTx(this IReleaseSpec spec) =>
 
         public static ulong GetTargetBlobGasPerBlock(this IReleaseSpec spec) =>
             spec.TargetBlobCount * Eip4844Constants.GasPerBlob;
+
+        public static int MaxBlobCount(this IReleaseSpec spec, int? blockProductionBlobLimit) =>
```
flcl42: May be named as MaxProductionBlobCount or so, to not confuse with the spec.MaxBlobCount property.

src/Nethermind/Nethermind.Config/IBlocksConfig.cs
```diff
@@ -62,4 +62,7 @@ public interface IBlocksConfig : IConfig
     bool BuildBlocksOnMainState { get; set; }
 
     byte[] GetExtraDataBytes();
+
+    [ConfigItem(Description = "The blob limit after which the block producer should stop adding blobs. Minimum value is 0.", DefaultValue = "null")]
+    int? BlockProductionBlobLimit { get; set; }
```
LukaszRozmej: @rubo ?

rubo: You got this. It may be worth emphasizing which limit we're referring to here -- the blob number limit or the blob size limit.
```suggestion
    [ConfigItem(Description = "The max blob number after which the block producer should stop adding blobs. Minimum value is `0`.", DefaultValue = "null")]
    int? BlockProductionBlobLimit { get; set; }
```

flcl42 (approved):

src/Nethermind/Nethermind.Config/IBlocksConfig.cs
smartprogrammer93:
```suggestion
    [ConfigItem(Description = "The max blob count after which the block producer should stop adding blobs. Minimum value is `0`.", DefaultValue = "null")]
    int? BlockProductionBlobLimit { get; set; }
```

PR #9695, opened 2025-11-13
XDC - Snapshot Manager refactor

cicr99:


Demuirgos (approved):

Demuirgos (commented):
does this build on the testBlockchain PR from Anders? it seems quiet different

src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
```diff
@@ -4,82 +4,78 @@
...
 {
 
     private LruCache<Hash256, Snapshot> _snapshotCache = new(128, 128, "XDC Snapshot cache");
-    private IDb _snapshotDb { get; }
-    private IPenaltyHandler _penaltyHandler { get; }
 
     private readonly SnapshotDecoder _snapshotDecoder = new();
 
-    public SnapshotManager(IDb snapshotDb, IPenaltyHandler penaltyHandler)
+    public Snapshot? GetSnapshotByGapNumber(ulong gapNumber)
```
ak88, line 24: why is it ulong?

cicr99: Took it directly from the old code that had this method tbh, but I think the reason is that gap number in both vote and timeout classes is defined as ulong, and those managers are the ones that call this method with that value

PR #9701, opened 2025-11-13
fix: EnrTreeParser tests to cover enr: leaves and enrtree:// links

MozirDmitriy:
The leaf and linked-tree tests were copy-pasted from the branch test and never exercised enr: or enrtree:// parsing. This masked real issues in the parser, notably that ParseEnrLeaf stores the full enr: string including the prefix, which leads to an incorrect ToString() of enr:enr:..., and kept enrtree:// formatting unverified. The tests now parse enr: and enrtree:// entries through EnrTreeParser.ParseNode(), assert ToString(), and verify Links/Refs/Records and types. This exposes the existing ParseEnrLeaf bug for correction and prevents regressions in linked-tree handling.


flcl42:
Hello @MozirDmitriy! Do you plan to fix the found issue? We for sure can't merge with failing tests

flcl42 (approved):

src/Nethermind/Nethermind.Network.Dns/EnrLeaf.cs
```diff
@@ -17,5 +17,5 @@ public class EnrLeaf : EnrTreeNode
 
     public override string[] Refs => [];
 
-    public override string[] Records => new[] { NodeRecord };
+    public override string[] Records => new[] { ToString() };
```
LukaszRozmej: Should this allocate on each access or be cached in a field?

LukaszRozmej (approved):

svlachakis:
@claude review

src/Nethermind/Nethermind.Network.Dns/EnrLeaf.cs
flcl42: Does not seem to be hot path, still cached because it's easy

LukaszRozmej (approved):

svlachakis (approved):

PR #9703, opened 2025-11-13
Refactor SnapServer to remove unused state reader dependency

soniseth0:
remove the unused IStateReader field and constructor parameter from SnapServer, update WorldStateManager and snap-sync tests to use the slimmer constructor and drop redundant substitutes/usings, inline VisitingOptions creation to avoid a throwaway local allocation


marcindsobczak:
You need to also remove `IStateReader` argument from all usages of modified class. There are 8 usages.

PR #9709, opened 2025-11-14
Taiko Shasta Changes

dipkakwani:
Resolves https://github.com/NethermindEth/Surge/issues/216

## Changes

- New APIs introduced as part of Shasta
- Extra data stored in L1OriginStore
- Taiko anchor validation updated
- ....

## Types of changes

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing
[Tested with Catalyst]
:heavy_check_mark: 

[Testing with Taiko Devnet]
It was able to sync a few blocks before taiko driver started complaining about : `proposal ID not found in cache`, most likely due to `RING_BUFFER_SIZE` which can only store some recent proposals and the L2 devnet chain has already moved 3k+ blocks.

NOTE: There is also a compat issue in `SetBatchToLastBlock` in taiko-mono and taiko-geth, which has to be fixed: https://github.com/taikoxyz/taiko-mono/blob/edda65255c595a7aa0913ab040e9a0fe569615ac/packages/taiko-client/pkg/rpc/engine.go#L187 (Similar to https://github.com/taikoxyz/taiko-mono/pull/20234)

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Taiko/L1OriginDecoder.cs
+++ b/src/Nethermind/Nethermind.Taiko/L1OriginDecoder.cs
@@ -11,7 +11,7 @@ namespace Nethermind.Taiko;
 public sealed class L1OriginDecoder : RlpStreamDecoder<L1Origin>
 {
     const int BuildPayloadArgsIdLength = 8;
-    const int SignatureLength = 65;
+    internal const int SignatureLength = 65;
 
     protected override L1Origin DecodeInternal(RlpStream rlpStream, RlpBehaviors rlpBehaviors = RlpBehaviors.None)
     {
--- a/src/Nethermind/Nethermind.Taiko/TaikoHeaderHelper.cs
+++ b/src/Nethermind/Nethermind.Taiko/TaikoHeaderHelper.cs
@@ -10,15 +10,9 @@ internal static class TaikoHeaderHelper
 {
     public static byte? DecodeOntakeExtraData(this BlockHeader header) => header.ExtraData is { Length: >= 32 } ? Math.Min(header.ExtraData[31], (byte)100) : null;
 
-    public static byte? DecodeShastaExtraData(this BlockHeader header)
-    {
-        if (header.ExtraData is not { Length: >= 2 })
-        {
-            return null;
-        }
-
-        // First byte: basefeeSharingPctg
-        // Second byte: isLowBondProposal (lowest bit)
-        return header.ExtraData[0];
-    }
+    // Two bytes encoded in the extra data for Shasta
+    // - First byte: basefeeSharingPctg
+    // - Second byte: isLowBondProposal (lowest bit)
+    // Returns only the basefeeSharingPctg
+    public static byte? DecodeShastaExtraData(this BlockHeader header) => header.ExtraData is not { Length: >= 2 } ? null : header.ExtraData[0];
 }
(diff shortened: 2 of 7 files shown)


src/Nethermind/Nethermind.Taiko/L1OriginDecoder.cs
```diff
@@ -42,27 +53,68 @@ public override void Encode(RlpStream stream, L1Origin item, RlpBehaviors rlpBeh
 
         stream.Encode(item.BlockId);
         stream.Encode(item.L2BlockHash);
-        stream.Encode(item.L1BlockHeight);
+        stream.Encode(item.L1BlockHeight ?? 0);
```
flcl42, line 56: If it's always some value, do we need L1BlockHeight to be nullable?

src/Nethermind/Nethermind.Taiko/TaikoHeaderHelper.cs
```diff
@@ -9,4 +9,16 @@ namespace Nethermind.Taiko;
 internal static class TaikoHeaderHelper
 {
     public static byte? DecodeOntakeExtraData(this BlockHeader header) => header.ExtraData is { Length: >= 32 } ? Math.Min(header.ExtraData[31], (byte)100) : null;
+
+    public static byte? DecodeShastaExtraData(this BlockHeader header)
```
flcl42: Can be one-liner as above

src/Nethermind/Nethermind.Taiko/TaikoSpec/TaikoCancunReleaseSpec.cs
flcl42, line 1: You can activate them thing by thing in every new spec, instead of having all 3 forks mentioned you can just have IsOntakeEnabled = true; in TaikoOntakeReleaseSpec, IsPacayaEnabled = true; in TaikoPacayaReleaseSpec and so one

src/Nethermind/Nethermind.Taiko/L1OriginDecoder.cs
dipkakwani: It was made as optional to match with Taiko geth (https://github.com/taikoxyz/taiko-geth/blob/ea79c83137e3e1ab5c90dc8f8ddf8efefd621d53/core/rawdb/taiko_l1_origin.go#L42) since I encountered an issue due to it during testing:

```
20 Nov 12:45:00 | Incorrect JSON RPC parameters when calling engine_forkchoiceUpdatedV2 with params 
"l1BlockHeight":null,"l1BlockHash":"0x0000000000000000000000000000000000000000000000000000000000000000","buildPayloadArgsID":[2,222,108,143,174,157,33,137],"isForcedInclusion":false,"signature":[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]}}]] 
System.Text.Json.JsonException: The JSON value could not be converted to Nethermind.Taiko.L1Origin. Path: $.l1Origin.l1BlockHeight | LineNumber: 0 | BytePositionInLine: 1576.
```

PR #9714, opened 2025-11-14
Remove unused System.Collections import from XdcHotStuff.cs

soniseth0:
Removes unused `using System.Collections;` import from `XdcHotStuff.cs` to clean up unnecessary dependencies.


ak88:
not impactful

PR #9715, opened 2025-11-15
fix: prevent loop index modification in ExtractStatsPerLevel

Lil-Duckling-22:
Fixed loop index modification bug where ++i was used inside switch expression, which could cause incorrect group processing. Now size_scale group is explicitly skipped and multiplier calculation uses direct group access.


LukaszRozmej:
Looks the same to me, feel free to reopen if you have a test demonstrating the issue.

PR #9725, opened 2025-11-17
Feature/blockhash cache

LukaszRozmej:
## Changes

- Introduces thread safe multi-fork `BlockhashCashe` to improve performance of Blockhash opcode.
- Prefetches Blockhashes in `BranchProcessor`
- Wires it up correctly in `eth_simulate` with proper caches that also include simulated blocks
- Introduces `InterlockEx` for safe `Max` and `Min`
- Introduces `IHeaderFinder` on `HeaderStore`
- Moved `ArrayPoolListCore` from generic methods to generic type.
- And a lot of minor unnecessary refactors - sorry got distracted.

Potential further optimization in #9731 

- [x] Optimization
- [x] Refactoring

#### Requires testing

- [x] Yes

- [x] Yes

#### Notes on testing

Heavy forking devnet?

- [x] No

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Processing/BranchProcessor.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/BranchProcessor.cs
@@ -82,8 +82,8 @@ public class BranchProcessor(
             // Start prewarming as early as possible
             WaitForCacheClear();
             IReleaseSpec spec = specProvider.GetSpec(suggestedBlock.Header);
-    	    preWarmTask = PreWarmTransactions(suggestedBlock, baseBlock!, spec, backgroundCancellation.Token);
-        	Task? prefetchBlockhash = blockhashProvider.Prefetch(suggestedBlock.Header, backgroundCancellation.Token);
+            preWarmTask = PreWarmTransactions(suggestedBlock, baseBlock!, spec, backgroundCancellation.Token);
+            Task? prefetchBlockhash = blockhashProvider.Prefetch(suggestedBlock.Header, backgroundCancellation.Token);
 
             BlocksProcessing?.Invoke(this, new BlocksProcessingEventArgs(suggestedBlocks));
 


src/Nethermind/Nethermind.Core.Test/Builders/BlockTreeBuilder.cs
```diff
@@ -252,31 +238,70 @@ public BlockTreeBuilder OfChainLength(out Block headBlock, int chainLength, int
             return this;
         }
 
+        public BlockTreeBuilder OfChainLengthWithSharedSplits(out Block headBlock, int chainLength, int splitVariant = 0, int splitFrom = 0, bool withWithdrawals = false, params Address[] blockBeneficiaries)
```
LukaszRozmej, line 241: `OfChainLenght` handles `splitFrom` in weird way - it always starts at genesis. I tried to make a version that would work with both correctly, but ended up breaking a lot of tests, so just added my new better version, maybe @asdacap wants to revisit his tests using old version?

src/Nethermind/Nethermind.Blockchain/BlockhashCache.cs
```diff
@@ -0,0 +1,238 @@
...
+    private CacheNode? Load(BlockHeader blockHeader, int depth, CancellationToken cancellationToken = default)
+    {
+        if (depth > MaxDepth) return null;
+        bool alwaysAdd = depth == MaxDepth;
+        using ArrayPoolListRef<(CacheNode Node, bool NeedToAdd)> blocks = new(depth + 1);
+        Hash256 currentHash = blockHeader.Hash!;
+        CacheNode currentNode = null;
+        bool needToAddAny = false;
+        int skipped = 0;
+        for (int i = 0; i <= depth && !cancellationToken.IsCancellationRequested; i++)
```
deffrian, line 47: As I understand we cache up to depth. So if someone requests headers in order from head down to head - 256. We will re-cache on every read. Can it be a problem?

src/Nethermind/Nethermind.Consensus/Stateless/StatelessBlockTree.cs
```diff
@@ -17,15 +17,8 @@ namespace Nethermind.Consensus.Stateless;
 /// <summary>
 /// This class is part of the StatelessExecution tool. It's intended to be used only inside the processing pipeline.
 /// </summary>
-/// <param name="headers"></param>
-public class StatelessBlockTree(IReadOnlyCollection<BlockHeader> headers) : IBlockTree
+public class StatelessBlockTree(Dictionary<Hash256, BlockHeader> hashToHeader, Dictionary<long, BlockHeader> numberToHeader) : IBlockTree
```
deffrian: Maybe make this class both `IBlockTree` and `IHeaderFinder`?

src/Nethermind/Nethermind.Blockchain/BlockhashCache.cs
LukaszRozmej: Depends which cache:
- linked list one - we cache up to depth, next time we will add more to the linked list
- flat - we cache only when full 256 are there - when we prefetch we prefetch full 256.

src/Nethermind/Nethermind.Consensus/Stateless/StatelessBlockTree.cs
LukaszRozmej: Actually made something custom to make the stateless code simpler

src/Nethermind/Nethermind.Blockchain/BlockhashCache.cs
```diff
@@ -0,0 +1,248 @@
...
+        }
+
+        if (alwaysAdd)
+        {
+            hashes = new Hash256[blocks.Count];
+            for (int i = 0; i < blocks.Count; i++)
+            {
+                hashes[i] = blocks[i].Node.Hash;
+            }
+            _flatCache.Set(blockHeader.Hash, hashes);
```
flcl42, line 111: Do we need to set it if cancellationToken.IsCancellationRequested? It may be not all blocks are in blocks? Could you confirm cancellation will still allow to fetch complete list of hashes(probably can be done by adding a delay to block finder or so)?

LukaszRozmej: ah good point, fixing

LukaszRozmej: fixed

src/Nethermind/Nethermind.Consensus/Stateless/StatelessBlockProcessingEnv.cs
```diff
@@ -47,8 +51,11 @@ public IWorldState WorldState
 
     private IBlockProcessor GetProcessor()
     {
-        IBlockTree statelessBlockTree = new StatelessBlockTree(witness.DecodedHeaders);
-        ITransactionProcessor txProcessor = CreateTransactionProcessor(WorldState, statelessBlockTree);
+        Dictionary<Hash256, BlockHeader> headersByHash = witness.DecodedHeaders.ToDictionary(header => header.Hash ?? throw new ArgumentNullException(), header => header);
+        Dictionary<long, BlockHeader> headersByNumber = witness.DecodedHeaders.ToDictionary(header => header.Number, header => header);
+
+        IBlockTree statelessBlockTree = new StatelessBlockTree(headersByHash, headersByNumber);
```
deffrian: I don't like that we create these dictionaries outside `StatelessBlockTree`. `StatelessBlockTree` user can put whatever `headersByHash` and `headersByNumber` he wants. This is why I suggested to put two interfaces into one class. I think we can implement `IBlockhashCache` for `StatelessBlockTree`.

src/Nethermind/Nethermind.Consensus/Processing/BranchProcessor.cs
```diff
@@ -98,10 +101,9 @@ public Block[] Process(BlockHeader? baseBlock, IReadOnlyList<Block> suggestedBlo
                 }
                 // If prewarmCancellation is not null it means we are in first iteration of loop
                 // and started prewarming at method entry, so don't start it again
-                if (prewarmCancellation is null)
-                {
-                    (prewarmCancellation, preWarmTask) = PreWarmTransactions(suggestedBlock, preBlockBaseBlock, spec);
-                }
+                backgroundCancellation ??= new CancellationTokenSource();
+                preWarmTask ??= PreWarmTransactions(suggestedBlock, preBlockBaseBlock, spec, backgroundCancellation.Token);
+                prefetchBlockhash ??= blockhashProvider.Prefetch(suggestedBlock.Header, backgroundCancellation.Token);
```
asdacap, line 109: Probably need to make sure this finished to prevent prefetch of early blockhash to get applied when executing later blockc.

LukaszRozmej: Good catch, decided not to await the task to keep the BranchProcessor as fast as possible, but added synchronization in the Prefetch itself.

LukaszRozmej: Alternatively I can just wait for that Task, but I really don't want to :D

src/Nethermind/Nethermind.Blockchain/BlockhashCache.cs
```diff
@@ -0,0 +1,249 @@
...
+    private readonly LruCache<Hash256AsKey, Hash256[]> _flatCache = new(32, nameof(BlockhashCache));
+    public const int MaxDepth = BlockhashProvider.MaxDepth;
+    private long _minBlock = int.MaxValue;
+    private Task _pruningTask = Task.CompletedTask;
+
+    public Hash256? GetHash(BlockHeader headBlock, int depth) =>
+        depth == 0 ? headBlock.Hash
+        : depth == 1 ? headBlock.ParentHash
+        : depth > MaxDepth ? null
+        : _flatCache.TryGet(headBlock.Hash!, out Hash256[] array) ? array[depth]
```
asdacap: You can query based on ParentHash instead of Hash, since it was populated from the last block, there is no need for the timing of the current Prefetch to be finished in this block.

src/Nethermind/Nethermind.Blockchain/BlockhashCache.cs
```diff
@@ -0,0 +1,249 @@
...
+
+            return hashes;
+        });
+    }
+
+    private void PruneInBackground(BlockHeader blockHeader)
+    {
+        if (ShouldPrune())
+        {
+            lock (_pruningTask)
```
asdacap: Probably should be a dedicated object, since it is replaced in line 15.

src/Nethermind/Nethermind.Consensus/Stateless/StatelessBlockTree.cs
```diff
@@ -17,11 +17,11 @@ namespace Nethermind.Consensus.Stateless;
 /// <summary>
 /// This class is part of the StatelessExecution tool. It's intended to be used only inside the processing pipeline.
 /// </summary>
-/// <param name="headers"></param>
-public class StatelessBlockTree(IReadOnlyCollection<BlockHeader> headers) : IBlockTree
+public class StatelessBlockTree(IReadOnlyCollection<BlockHeader> headers)
+    : IBlockTree, IBlockhashCache
```
asdacap, line 21: Why?

asdacap:
Please move the refactor into its own PR. It makes it hard to focus on important VM details.

src/Nethermind/Nethermind.Consensus/Stateless/StatelessBlockTree.cs
LukaszRozmej: for stateless simplicity and performance, it is for zkVM context

src/Nethermind/Nethermind.Blockchain/BlockhashCache.cs
flcl42: Or at least query _flatCache[parent] in Load and if it's found go quick path by returning [..._flatCacheParentItem[1..].concat(parentHash)]

PR #9737, opened 2025-11-19
Call Result - Should Revert

svlachakis:
Found it during some Arbitrum related bug investigation.

Looks like the change is correct - but would appreciate feedback from people.

### DIFF
--- a/.github/workflows/hive-tests.yml
+++ b/.github/workflows/hive-tests.yml
@@ -29,6 +29,15 @@ on:
         default: '3'
         type: choice
         options: ['0', '1', '2', '3', '4', '5']
+      parallelism:
+        description: Parallelism
+        required: true
+        default: 4
+        type: number
+      buildarg:
+        description: Docker build arguments
+        required: false
+        type: string
       hive-repo:
         description: Hive repo
         required: true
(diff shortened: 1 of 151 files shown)


benaadams (approved):

src/Nethermind/Nethermind.Evm/CallResult.cs
```diff
@@ -57,7 +57,7 @@ public CallResult(EvmExceptionType exceptionType)
             StateToExecute = null;
             Output = (null, StatusCode.FailureBytes);
             PrecompileSuccess = null;
-            ShouldRevert = false;
+            ShouldRevert = exceptionType == EvmExceptionType.Revert;
```
flcl42: Its probably a .ctor that should not be used outside

svlachakis: Good point! I've already worked around it by using the full constructor with explicit parameters. 
The workaround works but adds a bit more verbosity.

If that's the case - should I make this constructor internal/private to prevent future misuse?

flcl42: Less of them accessible seems to be better

LukaszRozmej: yeah lets make it private

svlachakis: rolled back the change and made it private, will use the full contructor on Arbitrum.

PR #9740, opened 2025-11-19
Add minimal fee per blob gas requirement

marcindsobczak:
## Changes

- Add minimal fee per blob gas requirement

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.Blobs.cs
+++ b/src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.Blobs.cs
@@ -300,9 +300,10 @@ namespace Nethermind.TxPool.Test
                 .WithMaxFeePerGas(1.GWei())
                 .SignedAndResolved(_ethereumEcdsa, TestItem.PrivateKeyA).TestObject;
 
-            _txPool.SubmitTx(tx, TxHandlingOptions.None).Should().Be(isRequirementEnabled && !isMaxFeePerBlobGasHighEnough
-                ? AcceptTxResult.FeeTooLow
-                : AcceptTxResult.Accepted);
+            Assert.That(_txPool.SubmitTx(tx, TxHandlingOptions.None),
+                Is.EqualTo(isRequirementEnabled && !isMaxFeePerBlobGasHighEnough
+                    ? AcceptTxResult.FeeTooLow
+                    : AcceptTxResult.Accepted));
         }
 
         [Test]


src/Nethermind/Nethermind.TxPool/Filters/PriorityFeeTooLowFilter.cs
```diff
@@ -12,22 +12,47 @@ namespace Nethermind.TxPool.Filters;
...
+        // quick path - allowing txs with max fee per blob gas higher than current fee per blob gas
+        if (tx.MaxFeePerBlobGas >= _chainHeadInfoProvider.CurrentFeePerBlobGas)
+        {
+            return AcceptTxResult.Accepted;
+        }
+
+        // and slow one
+        bool overflow = UInt256.MultiplyOverflow(_chainHeadInfoProvider.CurrentFeePerBlobGas, (UInt256)_minBlobBaseFeePercent, out UInt256 minFeePerBlobGas);
+        UInt256.Divide(minFeePerBlobGas, 100, out minFeePerBlobGas);
+        if (!overflow && tx.MaxFeePerBlobGas < minFeePerBlobGas)
```
flcl42: Maybe `if (overflow || tx.MaxFeePerBlobGas < minFeePerBlobGas)...`

flcl42 (approved):

src/Nethermind/Nethermind.TxPool/Filters/PriorityFeeTooLowFilter.cs
marcindsobczak: Should we fail if it is an overflow? It means that we are close (by default 50x blob base fee means overflow) to max UInt256 values

flcl42: I guess it's just not significant, all inputs are trusted after all

src/Nethermind/Nethermind.TxPool/Filters/PriorityFeeTooLowFilter.cs
```diff
@@ -8,20 +8,41 @@
 
 namespace Nethermind.TxPool.Filters;
 
-public class PriorityFeeTooLowFilter(ILogger logger, ITxPoolConfig txPoolConfig) : IIncomingTxFilter
+public class PriorityFeeTooLowFilter(IChainHeadInfoProvider chainHeadInfoProvider, ITxPoolConfig txPoolConfig, ILogger logger) : IIncomingTxFilter
 {
-    private readonly ILogger _logger = logger;
     private readonly UInt256 _minBlobsPriorityFee = txPoolConfig.MinBlobTxPriorityFee;
+    private readonly int _minBlobBaseFeePercent = txPoolConfig.MinFeePerBlobGasPercentRequirement;
```
LukaszRozmej: Make it UInt256 not to do conversion on every call!

```suggestion
    private readonly UInt256 _minBlobBaseFeePercent = txPoolConfig.MinFeePerBlobGasPercentRequirement;
```

src/Nethermind/Nethermind.TxPool/Filters/PriorityFeeTooLowFilter.cs
```diff
@@ -8,20 +8,41 @@
...
 
+        // quick path - allowing txs with max fee per blob gas higher than current fee per blob gas
+        if (tx.MaxFeePerBlobGas >= chainHeadInfoProvider.CurrentFeePerBlobGas)
+        {
+            return AcceptTxResult.Accepted;
+        }
+
+        // and slow one
+        bool overflow = UInt256.MultiplyOverflow(chainHeadInfoProvider.CurrentFeePerBlobGas, (UInt256)_minBlobBaseFeePercent, out UInt256 minFeePerBlobGas);
+        UInt256.Divide(minFeePerBlobGas, 100, out minFeePerBlobGas);
```
LukaszRozmej: Make 100 a UINt256 private static, so we don't convert on every call

src/Nethermind/Nethermind.TxPool/TxPoolConfig.cs
```diff
@@ -27,5 +27,6 @@ public class TxPoolConfig : ITxPoolConfig
     public int? ReportMinutes { get; set; } = null;
     public bool AcceptTxWhenNotSynced { get; set; } = false;
     public bool PersistentBroadcastEnabled { get; set; } = true;
+    public int MinFeePerBlobGasPercentRequirement { get; set; } = 50;
```
LukaszRozmej: I would start with at least 75-80, not 50

src/Nethermind/Nethermind.TxPool/Filters/PriorityFeeTooLowFilter.cs
```diff
@@ -8,20 +8,41 @@
...
+        if (!overflow && tx.MaxFeePerBlobGas < minFeePerBlobGas)
+        {
+            Metrics.PendingTransactionsTooLowFeePerBlobGas++;
+            if (logger.IsTrace) logger.Trace($"Skipped adding transaction {tx.ToString("  ")}, too low blob fee per gas with options {handlingOptions} from {new StackTrace()}");
+            return AcceptTxResult.FeeTooLow.WithMessage($"MaxFeePerBlobGas needs to be at least {_minBlobBaseFeePercent} percent of CurrentFeePerBlobGas ({chainHeadInfoProvider.CurrentFeePerBlobGas}), is {tx.MaxFeePerBlobGas}.");
+        }
```
LukaszRozmej: Maybe all we need is fast path? Should we even add blobs that are below `CurrentFeePerBlobGas`? Is it worth the complexity?

LukaszRozmej (commented):
Add test please

src/Nethermind/Nethermind.TxPool/Filters/PriorityFeeTooLowFilter.cs
marcindsobczak: Good point. We can just add bool flag if we should require full current fee per blob gas - way less complexity and we still have blob inclusion rate unaffected in current block

src/Nethermind/Nethermind.TxPool/ITxPoolConfig.cs
```diff
@@ -52,6 +52,9 @@ public interface ITxPoolConfig : IConfig
         Description = "The max blob transaction size allowed, excluding blobs, in bytes.")]
     long? MaxBlobTxSize { get; set; }
 
+    [ConfigItem(DefaultValue = "true", Description = "If enabled, max fee per blob gas at least equal to current blob base fee is required to add blob transaction to the pool.")]
```
LukaszRozmej: I think this description isn't up to our standards @rubo

rubo: Assuming I understood it correctly, I rephrased as follows:

```suggestion
    [ConfigItem(DefaultValue = "true", Description = "Whether to require the max fee per blob to be greater than or equal to the blob base fee when adding a blob transaction to the pool.")]
```

src/Nethermind/Nethermind.TxPool/ITxPoolConfig.cs
```diff
@@ -52,6 +52,9 @@ public interface ITxPoolConfig : IConfig
...
+    bool MaxFeePerBlobGasRequirementEnabled { get; set; }
```
rubo: Given this will become a command line option, maybe it's worth to shorten it a bit to `MaxFeePerBlobGasRequired` ("requirement enabled" changed to "required").

OurOptionNamesAreLAreadyTooLongLikeThis ;)

marcindsobczak: Changed to `MinBlobBaseFeeRequired` - it will be more precise about what we actually require

marcindsobczak: Hmm will be even better with `CurrentBlobBaseFeeRequired`

src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.Blobs.cs
```diff
@@ -272,6 +273,38 @@ public void should_allow_to_add_blob_tx_with_MaxPriorityFeePerGas_lower_than_1GW
...
+                chainHeadInfoProvider: chainHeadInfoProvider);
+            EnsureSenderBalance(TestItem.AddressA, UInt256.MaxValue);
+
+            Transaction tx = Build.A.Transaction
+                .WithShardBlobTxTypeAndFields()
+                .WithMaxFeePerBlobGas(isMaxFeePerBlobGasHighEnough ? currentFeePerBlobGas : currentFeePerBlobGas - 1)
+                .WithMaxFeePerGas(1.GWei())
+                .SignedAndResolved(_ethereumEcdsa, TestItem.PrivateKeyA).TestObject;
+
+            _txPool.SubmitTx(tx, TxHandlingOptions.None).Should().Be(isRequirementEnabled && !isMaxFeePerBlobGasHighEnough
```
alexb5dh: Haven't we decided to get rid of Fluent Assertions, at least for the new tests?

src/Nethermind/Nethermind.TxPool/Filters/PriorityFeeTooLowFilter.cs
```diff
@@ -8,20 +8,32 @@
...
+        {
+            return AcceptTxResult.Accepted;
+        }
+
+        if (tx.MaxPriorityFeePerGas < _minBlobsPriorityFee)
         {
             Metrics.PendingTransactionsTooLowPriorityFee++;
-            if (_logger.IsTrace) _logger.Trace($"Skipped adding transaction {tx.ToString("  ")}, too low priority fee with options {handlingOptions} from {new StackTrace()}");
+            if (logger.IsTrace) logger.Trace($"Skipped adding transaction {tx.ToString("  ")}, too low priority fee with options {handlingOptions} from {new StackTrace()}");
             return AcceptTxResult.FeeTooLow.WithMessage($"MaxPriorityFeePerGas for blob transaction needs to be at least {_minBlobsPriorityFee} (1 GWei), is {tx.MaxPriorityFeePerGas}.");
```
alexb5dh: Where does `"(1 GWei)"` come from? Not sure it still makes sense here.

alexb5dh (approved):

src/Nethermind/Nethermind.TxPool/Filters/PriorityFeeTooLowFilter.cs
marcindsobczak: It doesn't, good catch! 1Gwei was old default value, changed recently

LukaszRozmej (approved):

PR #9743, opened 2025-11-20
Fix `RequestPooledTransactions` generic parameter name

alexb5dh:
Keeps being flagged as error in Rider and just seems wrong.

## Changes

- `RequestPooledTransactions` generic parameter name.

- [x] Other: code analysis

#### Requires testing

- [x] No

- [x] No

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/V65/Eth65ProtocolHandler.cs
+++ b/src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/V65/Eth65ProtocolHandler.cs
@@ -176,7 +176,7 @@ namespace Nethermind.Network.P2P.Subprotocols.Eth.V65
             }
         }
 
-        protected virtual void RequestPooledTransactions<TMessage>(IOwnedReadOnlyList<Hash256> hashes)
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/V65/Eth65ProtocolHandler.cs
```diff
@@ -176,8 +176,8 @@ protected override void SendNewTransactionsCore(IEnumerable<Transaction> txs, bo
             }
         }
 
-        protected virtual void RequestPooledTransactions<GetPooledTransactionsMessage>(IOwnedReadOnlyList<Hash256> hashes)
-            where GetPooledTransactionsMessage : P2PMessage, INew<IOwnedReadOnlyList<Hash256>, GetPooledTransactionsMessage>
+        protected virtual void RequestPooledTransactions<TMessage>(IOwnedReadOnlyList<Hash256> hashes)
```
LukaszRozmej: does it need to be virtual?

alexb5dh: no, removed

PR #9746, opened 2025-11-20
XDC: stream Signature as spans to eliminate per-call 65B allocations

radik878:
## Changes

Replaced all uses of Signature.BytesWithRecovery in XDC RLP encoders with direct streaming via StartByteArray, Write of R/S spans and WriteByte for recovery id. This preserves exact RLP output while removing a byte[65] allocation from each encode call on hot serialization paths such as quorum and timeout certificates. RlpStream already exposes span-based APIs, so no API changes to Signature are required and we retain its Vector512-based storage. Other areas where a byte[] is truly required (e.g., header fields and tests) are left intact to maintain existing data model contracts. This change reduces GC pressure and improves throughput

- [x] Optimization


stdevMac:
Closing: superseded by #10295 ("Perf/xdc"), which introduced `Signature.WriteBytesWithRecoveryTo(Span<byte>)` and applied it in the same four decoders, eliminating the same per-call byte[65] allocation.

PR #9747, opened 2025-11-20
Propagate cancellation through DNS lookups in ENR discovery

Forostovec:
Previously, cancellation requested by consumers of EnrDiscovery could not cancel in-flight DNS TXT queries because IDnsClient.Lookup did not accept a CancellationToken and DnsClient.Lookup hardcoded CancellationToken.None, while EnrTreeCrawler’s async iterators did not accept a token to forward. This change threads CancellationToken end-to-end by adding an optional token to IDnsClient.Lookup, passing it to LookupClient.QueryAsync, and updating EnrTreeCrawler’s async iterators to accept a token with [EnumeratorCancellation], check it, and propagate it to nested async enumerations. This aligns with the repository’s existing use of [EnumeratorCancellation] and WithCancellation in EnrDiscovery, enables cooperative cancellation for responsiveness and graceful shutdown, reduces wasted work during cancellation, and preserves source compatibility via default parameters.


LukaszRozmej:
What's the point if the parameter is never used in any invocations?

src/Nethermind/Nethermind.Network.Dns/IDnsClient.cs
```diff
@@ -22,7 +22,7 @@ public DnsClient(string domain)
         _client = new();
     }
 
-    public async Task<IEnumerable<string>> Lookup(string query)
+    public async Task<IEnumerable<string>> Lookup(string query, CancellationToken cancellationToken = default)
```
LukaszRozmej, line 25: Should any usages also be updated?

src/Nethermind/Nethermind.Network.Dns/IDnsClient.cs
```diff
@@ -8,7 +8,7 @@ namespace Nethermind.Network.Dns;
 
 public interface IDnsClient
 {
-    Task<IEnumerable<string>> Lookup(string query);
+    Task<IEnumerable<string>> Lookup(string query, CancellationToken cancellationToken = default);
```
LukaszRozmej, line 11: Should any usages also be updated?

src/Nethermind/Nethermind.Network.Dns/IDnsClient.cs
Forostovec: > Should any usages also be updated?

already updated?

LukaszRozmej (approved):

LukaszRozmej:
Doesn't build

benaadams (approved):

PR #9752, opened 2025-11-22
Fix block number format in `WaitForBlobInclusion`

maradini77:
## Changes

- Fixed incorrect block number format passed to `eth_getBlockByNumber` RPC call in `WaitForBlobInclusion` method
- Block numbers are now correctly converted to hexadecimal format with `0x` prefix as required by JSON-RPC specification
- Previously, decimal string format caused RPC calls to fail silently, preventing blob transaction inclusion verification

### DIFF
--- a/tools/SendBlobs/BlobSender.cs
+++ b/tools/SendBlobs/BlobSender.cs
@@ -356,11 +356,7 @@ internal class BlobSender
 
         while (true)
         {
-            string blockNumberHex = Hex.ToHexString(lastBlockNumber.ToBigEndian()).TrimStart('0');
-            blockNumberHex = blockNumberHex.Length == 0 ? "0" : blockNumberHex;
-            string blockTag = $"0x{blockNumberHex}";
-
-            var blockResult = await rpcClient.Post<BlockModel<Hash256>>("eth_getBlockByNumber", blockTag, false);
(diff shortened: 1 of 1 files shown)


tools/SendBlobs/BlobSender.cs
```diff
@@ -356,7 +356,11 @@ private async static Task WaitForBlobInclusion(IJsonRpcClient rpcClient, Hash256
 
         while (true)
         {
-            var blockResult = await rpcClient.Post<BlockModel<Hash256>>("eth_getBlockByNumber", lastBlockNumber.ToString() ?? "latest", false);
+            string blockNumberHex = Hex.ToHexString(lastBlockNumber.ToBigEndian()).TrimStart('0');
+            blockNumberHex = blockNumberHex.Length == 0 ? "0" : blockNumberHex;
+            string blockTag = $"0x{blockNumberHex}";
+
+            var blockResult = await rpcClient.Post<BlockModel<Hash256>>("eth_getBlockByNumber", blockTag, false);
```
flcl42: Instead of blockTag could you just pass `lastBlockNumber` without any transformations? An internal converter will convert it to hex correctly. lastBlockNumber is always not null and it was a mistake to convert it to string and return latest if it's null

PR #9779, opened 2025-11-25
Devirtualize filter{store | manager}

benaadams:
## Changes

This PR optimizes the filter system by removing interface abstractions (`IFilterStore` and `IFilterManager`) and making the concrete classes (`FilterStore` and `FilterManager`) sealed, enabling better JIT compiler optimizations.

Tests were adjusted to use concrete implementations instead of mocks.

**Key changes:**
- Return `Array.Empty` when no filters rather than enumerating the ConcurrentDictionary
- Removed `IFilterStore` and `IFilterManager` interfaces and their null implementations
- Made `FilterStore` and `FilterManager` sealed classes
- Improved thread-safety in `SaveFilter` by using `TryAdd` instead of separate check-then-add
- Added `SaveFilters` batch method for tests

Before

image

Allocations
image

After

_No longer shows up even in an 8 minute trace_

- [x] Optimization

#### Requires testing

- [x] Yes

- [x] Yes - Adjusted to not use moq

### DIFF
--- a/src/Nethermind/Nethermind.Facade/Filters/FilterStore.cs
+++ b/src/Nethermind/Nethermind.Facade/Filters/FilterStore.cs
@@ -152,15 +152,6 @@ namespace Nethermind.Blockchain.Filters
             }
         }
 
-        // Used by tests
-        public void SaveFilters(IEnumerable<FilterBase> filters)
-        {
-            foreach (FilterBase filter in filters)
-            {
-                SaveFilter(filter);
-            }
-        }
-
         private int GetFilterId(bool generateId)
         {
             if (generateId)
--- a/src/Nethermind/Nethermind.Blockchain.Test/Filters/FilterManagerTests.cs
+++ b/src/Nethermind/Nethermind.Blockchain.Test/Filters/FilterManagerTests.cs
@@ -369,3 +369,15 @@ public class FilterManagerTests
         return builderInstance.TestObject;
     }
 }
+
+file static class FilterExtensions
+{
+    public static void SaveFilters<T>(this FilterStore store, IEnumerable<T> filters)
+        where T : FilterBase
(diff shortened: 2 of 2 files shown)


src/Nethermind/Nethermind.Facade/Filters/FilterStore.cs
```diff
@@ -140,8 +147,22 @@ public void SaveFilter(FilterBase filter)
...
+                {
+                    if (!_filters.TryAdd(filter.Id, filter))
+                    {
+                        throw new InvalidOperationException($"Filter with ID {filter.Id} already exists");
+                    }
+
+                    _currentFilterId = Math.Max(filter.Id, _currentFilterId);
+                }
+            }
         }
```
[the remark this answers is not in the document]
benaadams: Only used by tests

asdacap:
Is it used enough to warrant this?

benaadams:
> Is it used enough to warrant this?

Is called every block and should do no work; so doing work is too much

src/Nethermind/Nethermind.Facade/Filters/FilterStore.cs
```diff
@@ -140,8 +150,15 @@ public void SaveFilter(FilterBase filter)
...
+        }
 
-            _filters[filter.Id] = filter;
+        // Used by tests
+        public void SaveFilters(IEnumerable<FilterBase> filters)
+        {
+            foreach (FilterBase filter in filters)
+            {
+                SaveFilter(filter);
+            }
```
LukaszRozmej: Should be extension method on tests side

benaadams: Moved to extension in tests

PR #9790, opened 2025-11-26
chore: remove dead IsSyncing field and its assignment

ANtutov:
Removed the IsSyncing property from SubscriptionSyncingResult and the corresponding assignment in the object initializer. The field was marked with JsonIgnore and never consumed, resulting in a dead write and unnecessary allocation during message creation. Tests and the eth_subscribe syncing contract expect either a boolean false or an object with startingBlock, currentBlock, and highestBlock only, with no isSyncing field included. The SyncingResult converter governs this shape, so carrying IsSyncing in the wrapper type was redundant

### DIFF
--- a/src/Nethermind/Nethermind.JsonRpc/Modules/Subscribe/SyncingSubscription.cs
+++ b/src/Nethermind/Nethermind.JsonRpc/Modules/Subscribe/SyncingSubscription.cs
@@ -61,7 +61,6 @@ namespace Nethermind.JsonRpc.Modules.Subscribe
                 if (_logger.IsTrace) _logger.Trace($"Syncing subscription {Id} changed syncing status from {_lastIsSyncing} to {isSyncing}");
 
                 _lastIsSyncing = isSyncing;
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.JsonRpc/Modules/Subscribe/SyncingSubscription.cs
LukaszRozmej:
```suggestion
                using (JsonRpcResult result = !isSyncing
                           ? CreateSubscriptionMessage(false)
                           : CreateSubscriptionMessage(new SubscriptionSyncingResult()
                           {
                               StartingBlock = syncingResult.StartingBlock,
                               CurrentBlock = syncingResult.CurrentBlock,
                               HighestBlock = syncingResult.HighestBlock
                           }))
                {
                    await JsonRpcDuplexClient.SendJsonRpcResult(result);
                }

```

LukaszRozmej (approved):

PR #9797, opened 2025-11-26
chore: drop unused sequential path in bls msm precompiles

Doryu001:
remove the unused DisableConcurrency flag from `BlsConst`, delete the unreachable sequential decode branches in both MSM precompiles, keep the existing parallel logic as the single code path


LukaszRozmej:
Those are there for a reason

PR #9803, opened 2025-11-27
[WIP] Block-STM integration

LukaszRozmej:
This PR replaces and is built upon #8550
This PR is based on world-state refactoring of #9089

## Changes

- This PR introduces block-STM. Based on paper: https://arxiv.org/abs/2203.06871 : Paper has spin-loop and 2 multi-threading bugs though(!).
- Current status: abstract algorithm is implemented and tested. Integration with Nethermind block processing is in progress.

- [x] Optimization
- [x] Refactoring

#### Requires testing

- [x] Yes

- [x] Yes

- [x] Yes

- [x] Yes


LukaszRozmej:
@claude review

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/BlockStm/BlockStmTransactionsExecutor.cs
```diff
@@ -0,0 +1,678 @@
...
+            // ReturnUnusedReadSet below.
+            ConcurrentQueue<List<Read>> pool = _readSetPool;
+            multiVersionMemory.DrainReadSets(list =>
+            {
+                list.Clear();
+                pool.Enqueue(list);
+            });
+            while (envPool.TryTake(out ParallelEnvFactory.ParallelAutoReadOnlyTxProcessingEnv? env))
+            {
+                env.WorldStateScopeProvider.ReturnUnusedReadSet();
```
benaadams, line 187: **[C1] HIGH: successful read sets have two owners after return**

`Record` retains the provider's read list. At block cleanup, `DrainReadSets` returns it to the shared pool, then `ReturnUnusedReadSet` returns that same list again because the provider never relinquished ownership. Two subsequent rentals can alias; beginning one transaction clears another transaction's live conflict dependencies. This can suppress required reexecution.

**Evidence:** a one-worker, one-transaction production integration probe found two pool entries referencing the same object; clearing the second rental cleared a dependency added to the first. Fix ownership transfer explicitly: a published read set belongs to MVMM until retired; the provider must detach it. Return unpublished and superseded sets exactly once, after readers finish.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/BlockStm/ParallelScheduler.cs
```diff
@@ -0,0 +1,405 @@
...
+
+        while (true)
+        {
+            long current = Volatile.Read(ref stateInt);
+            TxState currentState = Unsafe.As<long, TxState>(ref current);
+            // Only wake from Aborting; a tx parked on multiple blockers gets a SetReady per
+            // blocker resume, but only the first should transition / bump the incarnation.
+            // A second wake while already Ready / Executing / Executed would otherwise re-bump
+            // and leave an in-flight worker writing under a stale incarnation.
+            if (currentState.Status != TxStatus.Aborting) return;
```
benaadams, line 283: **[C2] HIGH: stale dependency wakes can resume a validation abort**

Initial nonce scheduling adds a transaction to multiple predecessors' wait sets. Finishing the first predecessor wakes it without removing its other memberships. Later, a remaining predecessor can finish while that transaction is in `Aborting` for read-set validation. `SetReady` checks only the status, so the stale wake starts a new incarnation before the old validator finishes converting writes to estimates. That old validator then marks the replacement's writes as estimates.

**Evidence:** a deterministic probe using the actual scheduler/MVMM APIs finishes with `Done=True; sender estimate=True; fee estimate=True`. Final replay skips estimates, losing the accepted transaction's writes. This is a forced legal interleaving, not a probabilistic stress-test result. Distinguish dependency suspension from validation invalidation, bind wakeups to blocker/incarnation identities, and prevent a new incarnation from publishing until invalidation of the old one completes. Also reduce and correctly retire initial nonce dependencies.

src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.cs
```diff
@@ -132,7 +132,8 @@ protected virtual TxReceipt[] ProcessBlock(
 
         _balManager.Setup(block);
 
-        _systemContractHandler.StoreBeaconRoot(block, spec, NullTxTracer.Instance);
+        ITxTracer systemTracer = _blockTransactionsExecutor.PreTxStateCommitTracer ?? NullTxTracer.Instance;
+        _systemContractHandler.StoreBeaconRoot(block, spec, systemTracer);
```
benaadams, line 136: **[C3] HIGH: the EIP-2935 pre-block write is absent from worker state**

The new tracer captures beacon-root changes, but `ApplyBlockhashStateChanges` writes the history contract without this tracer. Workers read the parent state plus the incomplete overlay, so a transaction reading the latest stored parent hash sees stale data. Capture every pre-transaction system-state mutation, or provide a shared immutable snapshot after all those mutations.

**Evidence:** a Prague differential test calling the history contract and storing its result produced different state roots, with serial gas 110,916 versus STM 91,016. The pre-block write is required by [EIP-2935](https://eips.ethereum.org/EIPS/eip-2935).

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/BlockStm/MultiVersionMemoryScopeProvider.cs
```diff
@@ -0,0 +1,464 @@
...
+        ConcurrentDictionary<ValueHash256, byte[]> blockCodeWrites,
+        BlockBaseReadCache baseReadCache) : IWorldStateScopeProvider.IScope
+    {
+        private readonly TrackingCodeDb _codeDb = new(baseScope.CodeDb, blockCodeWrites);
+
+        public void Dispose() => baseScope.Dispose();
+
+        public Hash256 RootHash => baseScope.RootHash;
+
+        public void UpdateRootHash() => baseScope.UpdateRootHash();
```
benaadams, line 155: **[C4] HIGH: pre-Byzantium receipt state roots describe the wrong state**

Pre-Byzantium receipts require each transaction's resulting state root. Parallel execution calculates receipts against a base scope while writes and fees remain deferred, so these intermediate roots do not represent serial prefix state. Fall back for these forks unless ordered prefix-root construction is implemented.

**Evidence:** a Spurious Dragon single-transfer test had equal final state and gas but unequal receipt roots: serial `537c95bff90b3d65f188e4b93fbefadbe1971ff840c8ec5895a7e68b31b502f3`, STM `8c08a7b8ba06fcb56b4e06cacff355aeaa19e9c47a835332d7f13b8279176005`.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/BlockStm/BlockStmTransactionsExecutor.cs
```diff
@@ -0,0 +1,678 @@
...
+
+    public TxReceipt[] ProcessTransactions(Block block, ProcessingOptions processingOptions, BlockReceiptsTracer receiptsTracer, CancellationToken token = default)
+    {
+        // Genesis, system txs, and Optimism deposit txs need strict ordering; fall back
+        // to the wrapped sequential executor. This is a full-block fallback: a single deposit
+        // tx surrenders the parallelism for every regular tx in the block. Splitting the run
+        // (sequential prefix/suffix + parallel middle) is on the backlog — the current
+        // executor doesn't have a "skip this tx" entry point so the all-or-nothing decision
+        // lives here.
+        if (block.IsGenesis || ContainsSequentialOnlyTxs(block.Transactions))
```
benaadams, line 102: **[C5] HIGH, BAL-enabled specs: STM bypasses BAL execution/collection**

STM is the outer main-processing decorator. Its fallback checks genesis/system/deposit transactions, not BAL eligibility, so normal BAL blocks do not reach the existing BAL transaction executor. The module's documentation says the opposite. Preserve the BAL dispatch contract or integrate equivalent recording and validation into STM before selecting it.

**Evidence:** under `Amsterdam.NoEip8037Instance`, a production differential probe generated different BAL hashes for the same single transfer despite equal gas and state. Serial hash `bc17c3eab62a991c286cf4d0c382f13bf3cf37fd8fb82a390fd29fcb080b8a49`; STM `d47b3f6db644e79dbad400e0eb8c0000d9ce0db2538fc3e6c22f1786f8fab80b`.

benaadams (commented):
Correctness and performance review of `4a376b1e`: new line-specific findings below. Additional evidence for overlapping findings is being added to existing unresolved threads; measurements, newer research, redesign priorities and reproduction sources are provided in the accompanying general comment.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/BlockStm/BlockStmTransactionsExecutor.cs
```diff
@@ -0,0 +1,678 @@
...
+            if (transaction.GasLimit > remainingGas)
+            {
+                BlockProcessor.BlockValidationTransactionsExecutor.ThrowInvalidTransactionException(
+                    TransactionResult.BlockGasLimitExceeded,
+                    block.Header,
+                    transaction,
+                    i);
+            }
+
+            gasUsed += receipts[i].GasUsed;
```
benaadams, line 453: **[C6] HIGH, affected gas-accounting specs: receipt gas is not block gas**

`FinalizeGasUsed` sums post-refund receipt gas and uses that sum for block admission and `header.GasUsed`. Forks separating receipt gas from pre-refund or state-creation block accounting require distinct quantities. Carry the same per-transaction block-accounting result as the serial processor and fold it in order; do not reconstruct it from receipts.

**Evidence:** with Prague plus EIP-7778 enabled, a storage-write/refund differential test gave serial block gas 98,766 versus STM 90,124. The spec explicitly separates these quantities: [EIP-7778](https://eips.ethereum.org/EIPS/eip-7778). The EIP-8037 accounting path also needs to be preserved, not reduced to this receipt sum.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/BlockStm/ParallelEnvFactory.cs
```diff
@@ -0,0 +1,76 @@
...
+            readSetPool
+        );
+
+        ILifetimeScope childScope = parentLifetime.BeginLifetimeScope(builder =>
+        {
+            builder
+                .AddSingleton<IWorldStateScopeProvider>(worldState)
+                .AddSingleton<MultiVersionMemoryScopeProvider>(worldState)
+                // Parallel-mode TxProcessor: no header.GasUsed mutation, lets us share block.Header across workers.
+                .AddScoped<ITransactionProcessor, EthereumParallelTransactionProcessor>()
```
benaadams, line 49: **[C8] HIGH, plugin configurations: worker DI replaces chain semantics**

Each worker unconditionally registers `EthereumParallelTransactionProcessor`, replacing the transaction processor chosen by the chain plugin. Adding optional fee recorders to Optimism/Taiko processors does not help when those processors are never instantiated in the worker. Deposit fallback covers only the tested transaction condition, not every plugin-specific rule. Preserve the chain-selected processor through composition/capabilities, or explicitly restrict eligibility to supported processors.

**Evidence:** production registration and caller inspection; no L2 differential execution was run for this finding.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/BlockStm/BlockStmTransactionsExecutor.cs
```diff
@@ -0,0 +1,678 @@
...
+        Address? sender = block.Transactions[txIndex].SenderAddress;
+        if (sender is null) return;
+
+        // Park this tx on EVERY prior tx that bumps the sender's nonce (same sender, or any
+        // EIP-7702 authorization with this sender as authority). The scheduler tolerates
+        // multiple AbortExecution calls for the same tx — SetReady is idempotent on the
+        // Aborting status — and ResumeDependencies wakes the tx exactly once even if it's
+        // queued on multiple blockers. Chaining everything avoids cascade gaps in
+        // multi-auth scenarios where the auth chain isn't co-parked by the same-sender rule.
+        for (int i = txIndex - 1; i >= 0; i--)
```
benaadams, line 208: **[P1] Performance: Build nonce dependencies in one pass**

`BlockStmTransactionsExecutor.cs:197`: each transaction scans every earlier transaction and authorization list for nonce dependencies. Same-sender blocks also materialize all predecessor edges. There are n(n−1)/2 predecessor visits: 523,776 at 1,024 transactions, 8,386,560 at 4,096. Early wakes add unnecessary retries.

**Suggested direction:** Build last-nonce-writer information in one ordered pass, including authorization authorities: O(n + total authorizations), with only necessary edges. Use generation-aware dependency completion; retaining only a sender map without authorization semantics would be incorrect.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/BlockStm/ParallelScheduler.cs
```diff
@@ -0,0 +1,405 @@
...
+        // Mark Executed BEFORE claiming dependencies — AbortExecution's in-lock re-check
+        // then short-circuits instead of racing us for set ownership.
+        Interlocked.Exchange(ref state.Status, TxStatus.Executed);
+        ResumeDependencies(txIndex);
+
+        if (Volatile.Read(ref _validationIndex) > txIndex)
+        {
+            if (writeSetChanged)
+            {
+                DecreaseIndex(ref _validationIndex, txIndex);
```
benaadams, line 310: **[P2] Performance: Invalidate actual readers instead of rescanning suffixes**

`ParallelScheduler.cs:301`: new writes rewind suffix validation, including transactions that never read those keys. Repeated rewinds cause avoidable read-set scans and scheduler traffic.

**Suggested direction:** Track readers of changed versions and push invalidations to actual dependents. Retain discovered dependencies across retries and prioritize the oldest blockers. Bound speculation with a committed-prefix window.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/BlockStm/MultiVersionMemory.cs
```diff
@@ -0,0 +1,297 @@
...
+    // SLOAD warm-cache already short-circuit duplicates before they reach MVMM), and the
+    // enumerator over a span beats a HashSet's bucket walk by ~3x per entry.
+    private readonly List<Read>?[] _lastReads;
+
+    public MultiVersionMemory(int txCount)
+    {
+        _data = new ConcurrentDictionary<ParallelStateKey, Value>[txCount];
+        for (int i = 0; i < txCount; i++)
+        {
+            _data[i] = new ConcurrentDictionary<ParallelStateKey, Value>();
```
benaadams, line 52: **[P4] Performance: Avoid eager per-transaction concurrent dictionaries**

`MultiVersionMemory.cs:43`: eagerly constructs one concurrent dictionary per transaction, before any writes. Later metadata duplicates keys across dictionaries, sets and writer lists.

**Suggested direction:** Lazily allocate cold transaction metadata, or use compact immutable per-incarnation write sets plus a per-key version index. Preserve safe publication to concurrent readers. Measured empty MVMM allocation at 1,024 transactions is 1,656,504 bytes on this 32-logical-processor machine.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/BlockStm/MultiVersionMemoryScopeProvider.cs
```diff
@@ -0,0 +1,464 @@
...
+                Status valueStatus = TryRead(location, out TxVersion valueVersion, out object? value);
+                Status clearStatus = TryRead(clearKey, out TxVersion clearVersion, out object? clearValue);
+
+                readSet.Add(new Read(location, valueVersion));
+
+                bool hasClear = clearStatus == Status.Ok && ReferenceEquals(clearValue, MultiVersionMemory.SelfDestructMonit);
+
+                // Always record the clearKey dependency so a later concurrent SELFDESTRUCT
+                // re-triggers validation, even on base-zero / not-found paths.
+                readSet.Add(new Read(clearKey, clearVersion));
```
benaadams, line 337: **[P5] Performance: Deduplicate storage-clear dependencies**

`MultiVersionMemoryScopeProvider.cs:326`: each distinct storage read records the same contract-wide storage-clear dependency again. `ParallelStateKey` is 64 bytes and `Read` 72 bytes on the measured x64 runtime.

**Suggested direction:** Record storage-clear version once per contract/incarnation: S slots need S + 1 dependencies rather than 2S before other reads. Consider compact key IDs and typed metadata. Account for lookup cost when choosing deduplication strategy.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/BlockStm/BlockStmTransactionsExecutor.cs
```diff
@@ -0,0 +1,678 @@
...
+        ParallelUnbalancedWork.For(1, txCount, i => FindNonceDependencies(i, block, scheduler));
+        BlockHeader parent = blockFinder.FindParentHeader(block.Header, BlockTreeLookupOptions.TotalDifficultyNotNeeded);
+        IReleaseSpec spec = _blockExecutionContext.Spec;
+        // One env per worker, reused across every tx that worker handles.
+        ConcurrentBag<ParallelEnvFactory.ParallelAutoReadOnlyTxProcessingEnv> envPool = [];
+        ParallelEnvFactory.ParallelAutoReadOnlyTxProcessingEnv CreateEnv() =>
+            parallelEnvFactory.Create(multiVersionMemory, feeAccumulator, blockCodeWrites, baseReadCache, _readSetPool, spec);
+        for (int i = 0; i < _concurrencyLevel; i++)
+        {
+            envPool.Add(CreateEnv());
```
benaadams, line 150: **[P7] Performance: Reuse worker environments across blocks**

`BlockStmTransactionsExecutor.cs:145`: constructs and disposes all worker world states and Autofac scopes each block; tracer pools are also block-local.

**Suggested direction:** Keep bounded worker environments across blocks and reset block-specific context. Prove that no prior block's MVMM, code dictionary, caches or receipts remain referenced.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/BlockStm/BlockStmTransactionsExecutor.cs
```diff
@@ -0,0 +1,678 @@
...
+
+    /// <summary>
+    /// Most recent per-block metrics from this executor instance. Tests can assert against
+    /// this without contending on the static <see cref="Metrics"/> counters.
+    /// </summary>
+    public ParallelBlockMetrics LastBlockSnapshot { get; private set; } = ParallelBlockMetrics.Empty;
+    // 0 (default) => 75% of logical processors, min 1 — leaves headroom for CL traffic, RPC, sync.
+    private readonly int _concurrencyLevel = blocksConfig.BlockStmConcurrency > 0
+        ? blocksConfig.BlockStmConcurrency
+        : Math.Max(1, (Environment.ProcessorCount * 3) / 4);
```
benaadams, line 92: **[P8] Performance: Use a serial threshold and clamp worker count**

Executor `:90`, `:149`; `ParallelRunner.cs:27`: worker count is based on CPU count, not transaction count or estimated work. Empty, single-transaction and short blocks pay the parallel infrastructure cost.

**Suggested direction:** Clamp to usable work; use a measured serial threshold and contention-aware fallback. On 32 logical processors the default requests 24 workers even for a one-transaction block.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/BlockStm/MultiVersionMemoryScopeProvider.cs
```diff
@@ -0,0 +1,464 @@
...
+    /// <summary>
+    /// Try to enter (or remain in) speculative-prewarm mode. Only incarnation 0 is eligible:
+    /// for incarnation &gt;= 1 the cache should already be warm from the prior speculative
+    /// run, so we want to fail fast on the real blocker rather than burn CPU twice.
+    /// </summary>
+    /// <returns>
+    /// <c>true</c> if speculation is active (caller substitutes sentinel and continues);
+    /// <c>false</c> if the caller should throw the real abort.
+    /// </returns>
+    internal bool TryEnterSpeculative(in TxVersion blocker, int incarnation)
```
benaadams, line 107: **[P9] Performance: Bound speculative warming after blocked reads**

Scope provider `:100`; executor `:607`: a blocked first attempt can continue EVM execution using substituted/base values as speculative warming. That can follow the wrong branches and consume a large gas budget without producing a usable result.

**Suggested direction:** Bound speculative work and measure useful cache hits. Prefer targeted prefetch from observed/hinted reads, or abort and wait for a stable blocker. Give retries a work budget and switch hot regions to serial execution.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/BlockStm/BlockStmTransactionsExecutor.cs
```diff
@@ -0,0 +1,678 @@
...
+        // is not thread-safe, so application stays sequential below.
+        TxWriteAggregate[] aggregates = new TxWriteAggregate[txCount];
+        ParallelUnbalancedWork.For(0, txCount, i =>
+        {
+            CategorizeTxWrites(ref aggregates[i], multiVersionMemory.GetFinalWriteSet(i));
+        });
+
+        try
+        {
+            for (int txIndex = 0; txIndex < txCount; txIndex++)
```
benaadams, line 249: **[P11] Performance: Reduce the serial replay tail**

Executor `:235`: after parallel execution and parallel categorization, all intermediate transaction writes are replayed serially into the main world state. This duplicates work and leaves a serial tail.

**Suggested direction:** Reduce to final writes per key and pipeline stable-prefix processing/root work where safe. Preserve storage-clear/deletion epochs, account/code changes, fee ordering, receipts and BAL requirements; simply taking the last dictionary entry is not sufficient.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/BlockStm/BlockStmTransactionsExecutor.cs
```diff
@@ -0,0 +1,678 @@
...
+            {
+                // Failed: publish zero-fee keys + MarkCommitted so later coinbase readers don't livelock; readSet still published so prior-tx funding triggers re-validation.
+                receipts[txIndex] = null;
+                env.WorldStateScopeProvider.WriteSet.Clear();
+                EnsureFeeKeys(env.WorldStateScopeProvider, txIndex);
+                feeAccumulator.MarkCommitted(txIndex);
+                multiVersionMemory.Record(version, env.WorldStateScopeProvider.ReadSet, env.WorldStateScopeProvider.WriteSet);
+                return Status.Ok;
+            }
+            scope.WorldState.Commit(_blockExecutionContext.Spec, txTracer, commitRoots: true);
```
benaadams, line 625: **[P12] Performance: Avoid nested worker commit machinery**

Executor `:625`; scope provider storage batches `:442`: worker `Commit(commitRoots: true)` goes through normal storage commit machinery, including nested parallelism and batch merging back into the MVMM write dictionary. Code persistence can also schedule work.

**Suggested direction:** Capture a transaction journal/delta directly, bulk-publish it once, and keep one bounded parallel level. Avoid building per-worker trie results that final replay recomputes. Check which normal commit side effects remain required.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/BlockStm/MultiVersionMemory.cs
```diff
@@ -0,0 +1,297 @@
...
+    /// already read this tx may need re-validation (any added, removed, or re-written
+    /// key — the stored TxVersion advances even when the value is unchanged).
+    /// </summary>
+    public bool Record(TxVersion version, List<Read> readSet, Dictionary<ParallelStateKey, object> writeSet)
+    {
+        bool writeSetChanged = ApplyWriteSet(version, writeSet);
+        // Ownership transfers to MVMM here: future validations of higher txs iterate this list
+        // until a newer Record for the same txIndex overwrites the slot. The scope provider's
+        // pool of working lists never touches a list after it lands in _lastReads.
+        _lastReads[version.TxIndex] = readSet;
```
benaadams, line 75: **[P13] Performance: Recover pooling and bound retained capacities**

MVMM `:70`; provider `:74`, `:413`; executor aggregate pools `:54`: retry/superseded read lists miss pool returns, aborted code writes are copied and retained for the block, and pooled collections retain peak backing capacities.

**Suggested direction:** Repair ownership after C1, retire old incarnations safely, avoid redundant immutable-code copies and cap retained capacities. These are allocation/retention costs, not proof of an unbounded native leak.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/BlockStm/MultiVersionMemoryScopeProvider.cs
```diff
@@ -0,0 +1,464 @@
...
+                if (valueStatus == Status.Ok)
+                {
+                    if (hasClear && IsLater(clearVersion, valueVersion))
+                    {
+                        return VirtualMachineStatics.BytesZero;
+                    }
+                    return (byte[])value!;
+                }
+
+                byte[] baseValue = ReadBaseSlot(in cell, getFromStorage);
```
benaadams, line 348: **[P14] Performance: Avoid obsolete and duplicate base-state reads**

`BlockBaseReadCache.cs`; scope provider `:355`, `:367`: concurrent cold misses can issue duplicate base-state reads; a storage-clear miss path still fetches base storage before discarding it.

**Suggested direction:** Skip known-obsolete base reads. Evaluate batching/prefetch or single-flight loading for expensive cold misses without serializing cheap cache hits. Measure cold and warm runs separately.

src/Nethermind/Nethermind.Init/Modules/PrewarmerModule.cs
```diff
@@ -19,17 +19,22 @@ public class PrewarmerModule(IBlocksConfig blocksConfig) : Module
...
+        builder.AddSingleton<NodeStorageCache>();
 
-                // Note: Need a small modification to have this work on all branch processor due to the shared
-                // NodeStorageCache and the FrozenDictionary and the fact that some processing does not have
-                // branch processor, and use block processor instead.
-                .AddSingleton<IMainProcessingModule, PrewarmerMainProcessingModule>();
+        // The full speculative-execution prewarmer (PrewarmerScopeProvider + PreBlockCaches +
+        // PrecompileCachedCodeInfoRepository) duplicates work Block-STM already does via MVMM,
+        // so only wire it on the sequential path.
+        if (!blocksConfig.BlockStmEnabled)
```
benaadams, line 35: **[P15] Performance: Preserve safe warming and serial fallback performance**

`Nethermind.Init/Modules/PrewarmerModule.cs`: enabling STM disables the existing prewarmer, including useful precompile caching; serial system/deposit fallback also loses that baseline assistance.

**Suggested direction:** Preserve safe immutable/precompile warming and the normal serial fallback configuration. Split cache responsibilities instead of disabling the complete module.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/BlockStm/ParallelScheduler.cs
```diff
@@ -0,0 +1,405 @@
...
+/// </summary>
+/// <remarks>
+/// The algorithm tracks <c>_executionIndex</c>, <c>_validationIndex</c>, and
+/// <c>_activeTasks</c>; priority is to always schedule the lowest indexed transaction that
+/// needs work. When transactions finish out-of-order or a dependency is detected, the
+/// indexes are decreased and transactions are re-validated or re-executed.
+/// </remarks>
+public sealed class ParallelScheduler(int txCount, ObjectPool<HashSet<int>> setPool) : IDisposable
+{
+    private int _executionIndex;
```
benaadams, line 26: **[P17] Performance: Measure and reduce scheduler coherence/wakeup overhead**

`ParallelScheduler.cs:24`, `:44`; `ParallelRunner.cs:38`; collector counters: shared cursors, adjacent transaction states, interlocked metrics, event wakeups and highest-priority pool workers introduce coherence and scheduling overhead.

**Suggested direction:** Profile before changing: padded hot counters, worker-local metrics, bounded persistent workers, dependency-ready queues and fewer broad wakeups are candidates. False-sharing impact and lock/wait shares were not measured here.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/BlockStm/ParallelBlockMetricsCollector.cs
```diff
@@ -0,0 +1,68 @@
...
+            // is what gives parallelizationPercent its honest meaning — one badly-contended tx
+            // re-executing 20 times shouldn't shrink the reported parallelism the way summing
+            // raw events would.
+            if (attempts > 1) dependentTxs++;
+            if (attempts > maxIncarnations) maxIncarnations = attempts;
+        }
+
+        reexecutions -= txCount;
+
+        long parallelizationPercent = Metrics.CalculateParallelizationPercent(txCount, dependentTxs);
```
benaadams, line 65: **[P18] Performance: Report actual work and concurrency**

`ParallelBlockMetricsCollector.cs:38`; executor logging; `ParallelRunner.cs`: reported parallelization is essentially the fraction that avoided reexecution, so it can report 100% with one worker. Validation failures omit successful repeat validations. Worker execution also does not propagate the normal block-processing thread marker.

**Suggested direction:** Measure useful EVM time versus total CPU work, all validation attempts, worker occupancy, setup/replay/root wall time, allocations/GC and blocked time. Restore the processing-thread accounting contract. Do not use the existing percentage as evidence of speedup.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/BlockStm/ParallelEnvFactory.cs
```diff
@@ -0,0 +1,76 @@
...
+                .AddSingleton<ParallelFeeRecorder>(ctx => new ParallelFeeRecorder(feeAccumulator, worldState, ctx.Resolve<IWorldState>(), spec))
```
benaadams, line 50: **[R1] DI, `ParallelEnvFactory.cs:50`:** the manual `ctx => new ParallelFeeRecorder(...ctx.Resolve...)` factory duplicates constructor wiring, contrary to `di-patterns.md`. Use normal registration with explicit supplied block parameters. This is separate from C8's semantic processor replacement.

src/Nethermind/Nethermind.Consensus.Test/Processing/ParallelProcessing/BlockStm/BlockStmIntegrationTests.cs
```diff
@@ -0,0 +1,534 @@
...
+        }
+    }
+
+    private sealed class DelayedTransactionProcessor(ITransactionProcessor inner, TransactionDelayPolicy policy)
+        : ITransactionProcessor
+    {
+        public TransactionResult Process(Transaction tx, ITxTracer t, ExecutionOptions opts) { Apply(tx); return inner.Process(tx, t, opts); }
+        public void SetBlockExecutionContext(BlockHeader h) => inner.SetBlockExecutionContext(h);
+        public void SetBlockExecutionContext(in BlockExecutionContext c) => inner.SetBlockExecutionContext(in c);
+        private void Apply(Transaction tx) { if (policy.TryGetDelay(tx, out int ms) && ms > 0) Thread.Sleep(ms); }
```
benaadams, line 79: **[R2] tests, `BlockStmIntegrationTests.cs:79`:** `Thread.Sleep` creates timing-dependent interleavings. Use barriers or controlled scheduler hooks; add deterministic cases like C2. The existing tests also lack reproducible allocation/throughput coverage for the new hot path.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/BlockStm/README.md
```diff
@@ -0,0 +1,179 @@
+# Block-STM Parallel Transaction Execution
+
+This directory contains a Block-STM (Software Transactional Memory) parallel transaction
+executor for Nethermind, derived from the work in PR #9803 and follow-up branch
+`feature/block-stm-integration-codex`.
+
+## Status: WORK IN PROGRESS — compiles, not yet wired
```
benaadams, line 7: **[R3] documentation, `README.md:7`, `:62`; `BlockStmModule.cs:14`; `AbortParallelExecutionException.cs:10`:** the README says integration is not wired and lists a removed `ParallelTrace.cs`; the module claims BAL delegation that its predicate does not implement; the abort helper claims zero allocation contrary to measurement. Update these behavioral claims. Clearly redundant added member summaries in `ParallelBlockMetricsCollector.cs` should also be removed under the repository comment rule.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/BlockStm/README.md
```diff
@@ -0,0 +1,179 @@
...
+  `HintBal` (forwarded to base).
+- IDE0028 (`new()` → `[]`) and IDE0290 (primary ctor) style errors fixed.
+
+## Remaining known issues from the deep review
+
+Ordered by severity. The four critical bugs flagged at the top of this README and the
+selfdestruct / code-propagation issues have been fixed; what's left:
+
+- **EIP-8037 + SELFDESTRUCT-of-coinbase** (critical when EIP-8037 activates on mainnet).
+  Under EIP-8037 the fee is always paid to the coinbase then burned in the destroy-list
```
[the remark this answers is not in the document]
benaadams: Follow-up on reviewed head `4a376b1e` with additional evidence/qualification:

**[C7] HIGH, EIP-8037 behavior: deferred own fees survive beneficiary destruction**

The serial processor applies a transaction's fee before its final destruction/burn handling. STM defers that fee until after state replay; a beneficiary destroyed by that transaction is missing its own fee during burn accounting, and final fee application can recreate it. Deferred fees must participate in transaction-visible balance, burn and account-existence semantics at the correct point.

**Evidence:** Prague plus the EIP-8037/EIP-7708 switches, with the block beneficiary set to a newly created contract that self-destructs, produced different final state and receipt roots. This limitation was already acknowledged in the PR README; the probe confirms it remains. See the fork rules in [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037).



More precisely than the earlier double-credit description: the reproduced failure is that deferred own fees are absent during destruction/burn handling and applied afterward. The regression uses the relevant EIP switches explicitly, not ordinary Prague/Cancun alone.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/BlockStm/MultiVersionMemory.cs
```diff
@@ -0,0 +1,262 @@
...
+        public void Remove(int txIndex)
+        {
+            lock (_indicesDesc)
+            {
+                int idx = _indicesDesc.BinarySearch(txIndex, DescendingComparer.Instance);
+                if (idx >= 0) _indicesDesc.RemoveAt(idx);
+            }
+        }
+
+        public int FindHighestBelow(int txIndex)
```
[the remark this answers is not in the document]
benaadams: Follow-up on reviewed head `4a376b1e` with additional evidence/qualification:

**[P3] Performance: Replace linear hot-key predecessor search**

`MultiVersionMemory.cs:274`: predecessor lookup linearly walks a descending writer list under an exclusive lock. Hot accounts therefore cause both O(m) search and serialization; insertion/removal also shift list contents.

**Suggested direction:** First use binary predecessor search under the existing lock. Then evaluate immutable/chunked indexes or sparse/dense adaptive indexes, with atomic version publication. A different lock alone does not remove the search work.



A warmed Release/.NET 10 x64 microprobe of 100,000 uniform queries over 1,024 writers measured median 18.818 ms for the current search versus 6.447 ms for binary search (~2.92x), both retaining a lock and returning identical results. Five pairs, single thread, original first in each pair: this is a local lookup result, not whole-block speedup. Changing to ReaderWriterLockSlim alone leaves the O(m) scan.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/BlockStm/MultiVersionMemoryScopeProvider.cs
```diff
@@ -0,0 +1,328 @@
...
+
+            readSet.Add(new Read(location, readVersion));
+
+            FeeRecipientKind? feeKind = feeAccumulator.GetFeeKind(address);
+            if (feeKind == FeeRecipientKind.None)
+            {
+                return result;
+            }
+
+            int startTxIndex = readVersion.IsEmpty ? 0 : readVersion.TxIndex;
```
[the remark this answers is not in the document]
benaadams: Follow-up on reviewed head `4a376b1e` with additional evidence/qualification:

**[P6] Performance: Avoid repeated fee-prefix scans**

`FeeAccumulator.cs:130`; scope provider `:263`: fee-account reads enumerate earlier fee dependencies and scan accumulated fee prefixes. Repeated beneficiary-balance reads can create quadratic work and large read sets.

**Suggested direction:** Use versioned prefix aggregates or a segment tree with range-version validation, or stable committed prefixes plus a short speculative tail. A plain mutable prefix sum/Fenwick tree without version validation is insufficient during rollback. Keep fee zero/existence/destruction semantics.



This concern also applies to actual transaction reads of the beneficiary, independently of the prewarmer condition discussed above. Do not suppress semantically required fee dependencies for an EVM read.

src/Nethermind/Nethermind.Consensus/Processing/ParallelProcessing/BlockStm/BlockStmTransactionsExecutor.cs
```diff
@@ -0,0 +1,605 @@
...
+            parallelEnvFactory.Create(multiVersionMemory, feeAccumulator, blockCodeWrites, spec);
+        for (int i = 0; i < _concurrencyLevel; i++)
+        {
+            envPool.Add(CreateEnv());
+        }
+        ParallelTransactionProcessor parallelTransactionProcessor = new(block, parent, envPool, CreateEnv, multiVersionMemory, feeAccumulator, receipts, results, in _blockExecutionContext);
+        try
+        {
+            using ParallelRunner parallelRunner = new(scheduler, multiVersionMemory, parallelTransactionProcessor, blockMetrics, _concurrencyLevel);
+            parallelRunner.Run().GetAwaiter().GetResult();
```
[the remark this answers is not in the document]
benaadams: Follow-up on reviewed head `4a376b1e` with additional evidence/qualification:

**[P10] Performance: Cached aborts still allocate**

`AbortParallelExecutionException.cs:54`: cached `ExceptionDispatchInfo.Throw()` still allocates and unwinds; its allocation-free documentation is false on the tested runtime.

**Suggested direction:** A specialized cooperative blocked-result path through EVM execution can avoid exceptions. Measure the effect on the ordinary VM path before adding checks globally. Actual project probe: 34,400,000 bytes for 100,000 warmed throws, or 344 bytes/throw at this call depth.

src/Nethermind/Nethermind.State/EmptyStorageTreeShortCircuit.cs
```diff
@@ -0,0 +1,17 @@
...
+/// <summary>
+/// Process-wide kill switch for <c>PersistentStorageProvider.PerContractState.CreateStorageTree</c>'s
+/// "empty trie → short-circuit reads to zero" optimization. Block-STM sets
+/// <see cref="Disabled"/> at DI bootstrap: its <c>MultiVersionMemory</c> overlay can
+/// carry pre-block system-contract writes (EIP-4788 BeaconRoots, EIP-2935 blockhash) for
+/// slots of an otherwise-empty trie, and the short-circuit would mask them, sending
+/// every per-tx SLOAD on the affected contract to <c>ZeroBytes</c>.
+/// </summary>
+public static class EmptyStorageTreeShortCircuit
+{
```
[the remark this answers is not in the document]
benaadams: Follow-up on reviewed head `4a376b1e` with additional evidence/qualification:

**[P16] Performance: Scope the empty-tree workaround to STM**

`Nethermind.Init/Modules/BlockStmModule.cs:33`: enabling one STM context permanently changes static `EmptyStorageTreeShortCircuit.Disabled`, affecting unrelated serial/RPC contexts too.

**Suggested direction:** Scope the capability to the MVMM provider/environment. The workaround should not remove an empty-tree optimization from every world state in the process.

benaadams:
## Block-STM correctness and performance review

Reviewed head `4a376b1efe41e8f28aa1bad5ee8b03e67893b34a` against merge base `a68028f954667633e1f3ef0044829a7d66de969a`; all 40 changed files, none skipped. Recommendation: **request changes**. Findings C1–C8, P1–P18 and R1–R3 are posted inline or as evidence updates in existing unresolved threads.

Seven correctness problems have failing focused probes; an eighth follows from worker DI registration. The original 33 targeted tests pass. All defects require STM enabled, and protocol-specific findings depend on the corresponding release-spec switches; these are not claims about live-network activation.

The implementation decorates main block processing, but does not register a parallel block-production executor: `GlobalWorldStateBlockProducerEnvFactory` still uses `BlockProductionTransactionsExecutor`. Parallel validation of an agreed order and parallel-friendly candidate construction need separate evaluation.


Release build, .NET 10, Windows x64, 32 logical processors. These are diagnostic microprobes, not BenchmarkDotNet production throughput results.

| Probe | Result | Interpretation |
|---|---|---|
| Empty MVMM with 1,024 transaction slots | 1,656,504 allocated bytes (~1.58 MiB) | Fixed metadata overhead before transaction execution. |
| Metadata size | key 64 B, read 72 B | Repeated dependencies consume material memory and bandwidth. |
| Cached abort, 100,000 throws after warmup | 34,400,000 B, 344 B/throw | Caching the exception does not make this path allocation-free. Allocation depends on stack/runtime. |
| 100,000 predecessor queries over 1,024 writers | Median linear 18.818 ms; binary 6.447 ms, ~2.92x faster locally | Both alternatives held a lock and returned equivalent answers. Single-thread, uniform query workload; original ran first in each pair. Not a block-speedup estimate. |

## Newer papers and implementations

These are the most useful updates to the [original Block-STM paper](https://arxiv.org/abs/2203.06871). Their published speedups use different engines, transaction models, workloads and baselines; none predicts Nethermind's gain.

| Work | What transfers to this PR | Limitation |
|---|---|---|
| [Aptos Block-STM v2 scheduler, pinned source](https://github.com/aptos-labs/aptos-core/blob/0f0ce68d6ba73fd369d43dd346c744f710696b11/aptos-move/block-executor/src/scheduler_v2.rs) | Push invalidation of dependent readers, propagation of stalled dependencies, delayed first retries until earlier first writes exist, ready queues, ordered commit and parallel post-commit work. Directly addresses P2/P9/P11/P17. | This is a public implementation, not a separately verified v2 research paper. Its Move-specific machinery cannot be copied wholesale. |
| [NEMO, BRAINS 2025](https://arxiv.org/html/2510.15122v1) | Keep dependencies from successful reads, wait for stronger blocker progress before retrying, use incomplete hints and prioritize transactions that unblock others. | Evaluation uses simulated execution; reported up to 42% over Block-STM with 16 workers. Its owned-object greedy-commit rule does not directly apply to arbitrary Ethereum state. |
| [Efficient Parallel Execution of Blockchain Transactions with Conflict Specifications, AFT 2025](https://drops.dagstuhl.de/entities/document/10.4230/LIPIcs.AFT.2025.29) | Combine sound conflict information with optimistic execution and adapt under contention. Useful for sender/authorization hints and known access information without requiring perfect static analysis. | General EVM conflicts are dynamic. The evaluated restricted contract analysis does not establish a sound cheap analyzer for arbitrary calls, proxies and delegation. |
| [ParallelEVM, EuroSys 2025](https://yajin.org/papers/EuroSys_2025_ParallelEVM.pdf) | Operation-level dependency tracking and selective redo can reuse most of a costly transaction instead of rerunning the full EVM. | Higher implementation complexity: control-flow/gas changes need fallback, and stack/memory/revert semantics must survive. Headline results exclude an initial prefetch pass; do not compare them to cold end-to-end validation. |
| [RapidLane / Deferred Objects, 2024](https://arxiv.org/html/2405.06117v1) | Defer suitable commutative updates and validate their constraints. Most natural here for fee and narrowly defined balance deltas. | Requires preserved observation and conditional-update semantics. Arbitrary ERC-20 storage writes are not automatically commutative EVM operations. |
| [Anthemius, 2025](https://arxiv.org/abs/2502.10074) | Workload-aware block assembly avoids choosing a candidate dominated by a long dependency chain. Relevant to the user's block-construction objective. | A block importer cannot change the already agreed transaction order. This PR does not yet wire the production path this approach would improve. |
| [Exploiting Multi-Core Parallelism in Blockchain Validation and Construction, SEA 2026](https://drops.dagstuhl.de/storage/00lipics/lipics-vol371-sea2026/html/LIPIcs.SEA.2026.23/LIPIcs.SEA.2026.23.html) | Separates fixed-block scheduling from reward-constrained construction, using dependency structure and critical paths. Supports measuring the dependency-limited upper bound before adding cores. | Trace/DAG-based evaluation is not live Nethermind performance; building an all-pairs graph would reproduce P1's problem. |

[RISE pevm](https://github.com/risechain/pevm) is also useful implementation evidence for lazy fees/raw transfers. Its README distinguishes implemented work from proposed scheduling, state-root and other optimizations; do not treat all listed ideas as completed features.

## Recommended redesign order

1. **Make serial equivalence the gate.** Fix C1/C2 first, then capture complete pre-block state and preserve fork/plugin/BAL behavior. Add differential state, receipts, gas, logs and access-list checks for supported configurations.
2. **Remove avoidable work before attempting fine-grained EVM changes.** Linear-time nonce hints, binary predecessor search, deduplicated clear dependencies, compact metadata, reused workers and measured small-block fallback are relatively contained changes.
3. **Replace broad rescanning/retrying with dependency-directed scheduling.** Borrow v2/NEMO concepts, with explicit incarnation ownership and an ordered committed prefix. A simpler competing baseline is a bounded OCC window: one initial speculative attempt, then retry failed transactions only once all predecessors are committed. Under that discipline, stable predecessor state bounds full attempts to two, at the cost of less overlap in hot regions. Benchmark it rather than assuming more aggressive speculation wins.
4. **Reduce the serial tail.** Publish journals rather than per-worker trie commits; coalesce final updates under deletion/clear epochs; overlap only stable-prefix work with later execution using an immutable worker base. State-root calculation and final application must be included in speedup measurements.
5. **Then evaluate semantic deltas or selective EVM redo.** Raw transfer/fee fast paths require exact existence, overflow, balance-check, logs and fork behavior. Recognizing an ERC-20 selector is insufficient with arbitrary code, proxies and EIP-7702. Selective redo additionally needs PC, stack, memory, gas/refund, warmness, call frames, logs and revert checkpoints.
6. **Treat block construction as a separate feature.** Use Anthemius/critical-path ideas while respecting nonce chains, bundles, gas budgets and revenue. For validation, maintain consensus order and consider serial system-transaction prefixes followed by safe parallel regions instead of whole-block fallback.

The useful performance model is setup + (useful execution + redo + validation)/effective parallelism + serial application/root work + GC/coherence/waits. Effective parallelism is constrained by the dependency critical path, not the machine's logical CPU count. This PR increases almost every overhead term while leaving a substantial serial tail.


## Validation and scope limits

The seven focused correctness probes demonstrate read-set pool aliasing, a stale-dependency scheduler race, missing EIP-2935 state, legacy receipt roots, BAL generation, block-gas accounting and beneficiary-destruction fee handling. The allocation assertion also fails; the lookup equivalence/measurement probe passes. The plugin finding is based on DI/caller inspection, without an L2 execution probe.

Local tests used `-p:NuGetAudit=false` because the branch's existing package audit finding otherwise stopped restore; no package versions or production code were changed. No full-mainnet replay, Docker A/B, CPU profile or GC/lock timeline was run, so whole-block speedups and attribution of the historical slowdown remain unmeasured.

I did not retain non-atomic UInt256 reads as an independent accepted-wrong-state finding: fee-version validation can invalidate such observations. Likewise, I did not establish an empty-code-reset defect; `ClearCaches` does disable node caching; cancellation is not a new regression against the serial validation executor. Superseded read lists are GC-reclaimable, not an unbounded native leak.

Compare the serial baseline, current STM, and each redesign on the same machine/state/fork settings. Include empty/tiny blocks, independent expensive transactions, sender chains, hot contract storage, beneficiary reads, authorization lists, creations/destructions and real mainnet blocks. Sweep worker counts and cold/warm state independently. Require byte-identical outputs first, then compare full block mean/p50/p95, CPU work, allocations, peak working set, retained heap, GC pauses, contention, retries and all validation attempts. Sampling explains CPU; EventPipe/timeline explains GC/waits. A narrower executor microbenchmark must not substitute for final state-root and block-processing measurements.




<details>
<summary>Local reproduction source: ReviewSchedulerRaceTests.cs</summary>

These temporary review probes were run locally, not committed. For the partial integration files, change the existing `BlockStmIntegrationTests` declaration to `partial` to reuse its existing fixture.

```csharp
// SPDX-FileCopyrightText: 2026 Demerzel Solutions Limited
#nullable enable

using Microsoft.Extensions.ObjectPool;
using Nethermind.Consensus.Processing.ParallelProcessing.BlockStm;
using Nethermind.Core.Test.Builders;
using NUnit.Framework;

namespace Nethermind.Consensus.Test.Processing.ParallelProcessing.BlockStm;

public class ReviewSchedulerRaceTests
{
    [Test]
    public void Stale_nonce_dependency_must_not_resume_a_validation_abort()
    {
        using ParallelScheduler scheduler = new(3, new DefaultObjectPool<HashSet<int>>(new DefaultPooledObjectPolicy<HashSet<int>>()));
        MultiVersionMemory memory = new(3);
        ParallelStateKey sender = ParallelStateKey.ForAccount(TestItem.AddressA);
        ParallelStateKey fee = ParallelStateKey.ForFee(FeeRecipientKind.GasBeneficiary, 2);

        Assert.That(scheduler.AbortExecution(1, 0, fromActiveTask: false), Is.True);
        Assert.That(scheduler.AbortExecution(2, 0, fromActiveTask: false), Is.True);
        Assert.That(scheduler.AbortExecution(2, 1, fromActiveTask: false), Is.True);

        TxTask tx0 = FetchExecution(scheduler, memory);
        Assert.That(tx0.TxVersion.TxIndex, Is.EqualTo(0));
        scheduler.FinishExecution(tx0.TxVersion, memory.Record(tx0.TxVersion, [], new() { [sender] = 0 }));

        TxTask tx1 = FetchExecution(scheduler, memory);
        TxTask earlyTx2 = FetchExecution(scheduler, memory);
        Assert.That(tx1.TxVersion.TxIndex, Is.EqualTo(1));
        Assert.That(earlyTx2.TxVersion.TxIndex, Is.EqualTo(2));

        // Tx2 ran before tx1's nonce increment: the executor publishes its reads and zero fee even on invalid nonce.
        memory.Record(earlyTx2.TxVersion, [new Read(sender, tx0.TxVersion)], new() { [fee] = 0 });
        scheduler.FinishExecution(earlyTx2.TxVersion, writeSetChanged: false);
        TxTask validation = scheduler.NextTask();
        Assert.That(validation.Validating && validation.TxVersion == earlyTx2.TxVersion, Is.True);

        bool tx1Changed = memory.Record(tx1.TxVersion, [], new() { [sender] = 1 });
        Assert.That(memory.ValidateReadSet(2), Is.False);
        Assert.That(scheduler.TryValidationAbort(validation.TxVersion), Is.True);

        // Pause validator before ConvertWritesToEstimates. Finishing tx1 drains its stale initial dependency on tx2.
        scheduler.FinishExecution(tx1.TxVersion, tx1Changed);
        TxTask replacement = FetchExecution(scheduler, memory);
        Assert.That(replacement.TxVersion.TxIndex, Is.EqualTo(2));
        Assert.That(replacement.TxVersion.Incarnation, Is.GreaterThan(earlyTx2.TxVersion.Incarnation));
        bool replacementChanged = memory.Record(replacement.TxVersion,
            [new Read(sender, tx1.TxVersion)], new() { [sender] = 2, [fee] = 1 });
        scheduler.FinishExecution(replacement.TxVersion, replacementChanged);

        // Resume old validator: this must not turn the replacement's already published writes into estimates.
        memory.ConvertWritesToEstimates(2);
        scheduler.FinishValidation(2, aborted: true);
        for (int i = 0; i < 32 && !scheduler.Done; i++)
        {
            TxTask task = scheduler.NextTask();
            if (task.IsEmpty) continue;
            Assert.That(task.Validating, Is.True, "No remaining execution should have been necessary");
            Assert.That(memory.ValidateReadSet(task.TxVersion.TxIndex), Is.True);
            scheduler.FinishValidation(task.TxVersion.TxIndex, aborted: false);
        }

        Assert.That(scheduler.Done, Is.True, "The scheduler accepts the block despite stale estimates");
        TestContext.Out.WriteLine($"Done={scheduler.Done}; sender estimate={memory.GetFinalWriteSet(2)[sender].IsEstimate}; fee estimate={memory.GetFinalWriteSet(2)[fee].IsEstimate}");
        Assert.That(memory.GetFinalWriteSet(2)[sender].IsEstimate, Is.False,
            "An old validator must not overwrite the final incarnation, which PushChanges would silently skip");
    }

    private static TxTask FetchExecution(ParallelScheduler scheduler, MultiVersionMemory memory)
    {
        for (int i = 0; i < 32; i++)
        {
            TxTask task = scheduler.NextTask();
            if (task.IsEmpty) continue;
            if (!task.Validating) return task;
            Assert.That(memory.ValidateReadSet(task.TxVersion.TxIndex), Is.True);
            scheduler.FinishValidation(task.TxVersion.TxIndex, aborted: false);
        }
        Assert.Fail("Expected an executable transaction");
        return TxTask.Empty;
    }
}

```

</details>

<details>
<summary>Local reproduction source: BlockStmIntegrationTests.ReviewResource.cs</summary>

These temporary review probes were run locally, not committed. For the partial integration files, change the existing `BlockStmIntegrationTests` declaration to `partial` to reuse its existing fixture.

```csharp
// SPDX-FileCopyrightText: 2026 Demerzel Solutions Limited
#nullable enable

using System;
using System.Collections.Concurrent;
using System.Reflection;
using System.Threading;
using System.Threading.Tasks;
using Nethermind.Config;
using Nethermind.Blockchain.Tracing;
using Nethermind.Core.Specs;
using Nethermind.Consensus.Processing;
using Nethermind.Consensus.Processing.ParallelProcessing.BlockStm;
using Nethermind.Core;
using Nethermind.Core.Test.Builders;
using Nethermind.Evm.Tracing;
using NUnit.Framework;

namespace Nethermind.Consensus.Test.Processing.ParallelProcessing.BlockStm;

public partial class BlockStmIntegrationTests
{
    [Test]
    public async Task Review_returned_readsets_have_exclusive_ownership()
    {
        using StmTestBlockchain chain = await StmTestBlockchain.Create(new BlocksConfig
        {
            MinGasPrice = 0,
            BlockStmEnabled = true,
            BlockStmConcurrency = 1,
            PreWarmStateOnBlockProcessing = false
        });
        BlockHeader parent = chain.BlockTree.Head!.Header;
        Block block = Build.A.Block.WithParent(parent).WithGasLimit(TestBlockGasLimit)
            .WithTransactions(Tx(TestItem.PrivateKeyA, TestItem.AddressD, nonce: 0)).TestObject;
        using (IDisposable scope = chain.MainProcessingContext.WorldState.BeginScope(parent))
        {
            chain.MainProcessingContext.BlockProcessor.ProcessOne(block, ProcessingOptions.NoValidation,
                NullBlockTracer.Instance, chain.SpecProvider.GetSpec(block.Header), CancellationToken.None);
        }

        object processor = chain.MainProcessingContext.BlockProcessor;
        while (processor is not BlockProcessor)
        {
            FieldInfo inner = System.Array.Find(processor.GetType().GetFields(BindingFlags.Instance | BindingFlags.NonPublic),
                field => typeof(IBlockProcessor).IsAssignableFrom(field.FieldType))!;
            processor = inner.GetValue(processor)!;
        }
        object executor = typeof(BlockProcessor).GetField("_blockTransactionsExecutor", BindingFlags.Instance | BindingFlags.NonPublic)!
            .GetValue(processor)!;
        ConcurrentQueue<List<Read>> pool = (ConcurrentQueue<List<Read>>)typeof(BlockStmTransactionsExecutor)
            .GetField("_readSetPool", BindingFlags.Instance | BindingFlags.NonPublic)!.GetValue(executor)!;
        List<Read>[] rentals = pool.ToArray();
        TestContext.Out.WriteLine($"After one successful tx: pool entries={rentals.Length}; same object={rentals.Length > 1 && ReferenceEquals(rentals[0], rentals[1])}");
        if (rentals.Length > 1)
        {
            rentals[0].Add(new Read(ParallelStateKey.ForAccount(TestItem.AddressA), TxVersion.Empty));
            rentals[1].Clear();
            Assert.That(rentals[0], Has.Count.EqualTo(1), "Beginning a second rental must not erase the first transaction's dependency");
        }
        Assert.That(rentals, Has.Length.EqualTo(1), "The published read set must return exactly once");
    }

    [Test]
    public void Review_original_cached_abort_exception_allocations()
    {
        const int count = 100_000;
        for (int i = 0; i < 1000; i++) ThrowAndCatch();
        long before = GC.GetAllocatedBytesForCurrentThread();
        for (int i = 0; i < count; i++) ThrowAndCatch();
        long allocated = GC.GetAllocatedBytesForCurrentThread() - before;
        TestContext.Out.WriteLine($"Original AbortParallelExecutionException: {allocated / (double)count:F2} bytes per throw ({count} throws)");
        Assert.That(allocated, Is.Zero, "The cached EDI throw is documented as allocation-free");

        static void ThrowAndCatch()
        {
            try { AbortParallelExecutionException.Throw(new TxVersion(0, 0)); }
            catch (AbortParallelExecutionException) { }
        }
    }
}


```

</details>

<details>
<summary>Local reproduction source: BlockStmIntegrationTests.ReviewProtocol.cs</summary>

These temporary review probes were run locally, not committed. For the partial integration files, change the existing `BlockStmIntegrationTests` declaration to `partial` to reuse its existing fixture.

```csharp
using System;
using System.Threading;
using System.Threading.Tasks;
using Nethermind.Blockchain.Tracing;
using Nethermind.Consensus.Processing;
using Nethermind.Core;
using Nethermind.Core.Extensions;
using Nethermind.Core.Specs;
using Nethermind.Core.Test.Builders;
using Nethermind.Evm;
using Nethermind.Specs.Forks;
using NUnit.Framework;

namespace Nethermind.Consensus.Test.Processing.ParallelProcessing.BlockStm;

public partial class BlockStmIntegrationTests
{
    [Test]
    public async Task Review_destroyed_coinbase_burns_own_fee()
    {
        IReleaseSpec spec = new Prague { IsEip8037Enabled = true, IsEip7708Enabled = true };
        using StmTestBlockchain parallel = await StmTestBlockchain.Create(BuildConfig(true), spec);
        using StmTestBlockchain serial = await StmTestBlockchain.Create(BuildConfig(false), spec);
        static Block Process(StmTestBlockchain chain, IReleaseSpec spec)
        {
            byte[] init = Prepare.EvmCode.Op(Instruction.ADDRESS).Op(Instruction.SELFDESTRUCT).Done;
            Transaction tx = Build.A.Transaction.WithNonce(0).WithGasLimit(1_000_000)
                .WithGasPrice(1.GWei).WithCode(init).WithValue(1.Ether)
                .SignedAndResolved(TestItem.PrivateKeyB, false).TestObject;
            BlockHeader parent = chain.BlockTree.Head!.Header;
            Block block = Build.A.Block.WithParent(parent).WithGasLimit(TestBlockGasLimit)
                .WithBeneficiary(ContractAddress.From(TestItem.AddressB, 0)).WithTransactions(tx).TestObject;
            using IDisposable scope = chain.MainProcessingContext.WorldState.BeginScope(parent);
            return chain.MainProcessingContext.BlockProcessor.ProcessOne(block, ProcessingOptions.NoValidation,
                NullBlockTracer.Instance, spec, CancellationToken.None).Block;
        }
        Block p = Process(parallel, spec);
        Block s = Process(serial, spec);
        using (Assert.EnterMultipleScope())
        {
            Assert.That(p.Header.StateRoot, Is.EqualTo(s.Header.StateRoot));
            Assert.That(p.Header.ReceiptsRoot, Is.EqualTo(s.Header.ReceiptsRoot));
        }
    }

    [Test]
    public async Task Review_refund_block_gas_matches_serial()
    {
        await using DualBlockchain chains = await DualBlockchain.Create(new Prague { IsEip7778Enabled = true });
        byte[] code = Prepare.EvmCode.PushData(1).PushData(0).Op(Instruction.SSTORE)
            .PushData(0).PushData(0).Op(Instruction.SSTORE).STOP().Done;
        byte[] init = Prepare.EvmCode.ForInitOf(code).Done;
        Address target = ContractAddress.From(TestItem.AddressB, 0);
        BlockPair blocks = chains.ProcessBlock(TxCreateContract(TestItem.PrivateKeyB, 0, init),
            TxToContract(TestItem.PrivateKeyC, target, 0, []));
        blocks.AssertFullMatch(2);
    }

    [Test]
    public async Task Review_BAL_dispatch_preserves_generated_access_list()
    {
        await using DualBlockchain chains = await DualBlockchain.Create(Amsterdam.NoEip8037Instance);
        BlockPair blocks = chains.ProcessBlock(TxToContract(TestItem.PrivateKeyB, TestItem.AddressD, 0, []));
        blocks.AssertFullMatch(1);
        Assert.That(blocks.Parallel.Header.BlockAccessListHash, Is.EqualTo(blocks.Single.Header.BlockAccessListHash));
    }

    [Test]
    public async Task Review_pre_Byzantium_receipts_match_serial()
    {
        await using DualBlockchain chains = await DualBlockchain.Create(SpuriousDragon.Instance);
        BlockPair blocks = chains.ProcessBlock(TxToContract(TestItem.PrivateKeyB, TestItem.AddressD, 0, []));
        blocks.AssertFullMatch(1);
        Assert.That(blocks.Parallel.Header.ReceiptsRoot, Is.EqualTo(blocks.Single.Header.ReceiptsRoot));
    }

    [Test]
    public async Task Review_history_contract_reads_current_parent_hash()
    {
        await using DualBlockchain chains = await DualBlockchain.Create(Prague.Instance);
        byte[] code = Prepare.EvmCode
            .PushData(1).Op(Instruction.NUMBER).Op(Instruction.SUB)
            .PushData(0).Op(Instruction.MSTORE)
            .PushData(32).PushData(32).PushData(32).PushData(0)
            .PushData(Eip2935Constants.BlockHashHistoryAddress).Op(Instruction.GAS).Op(Instruction.STATICCALL)
            .Op(Instruction.POP)
            .PushData(32).Op(Instruction.MLOAD)
            .Op(Instruction.PUSH0).Op(Instruction.SSTORE)
            .STOP().Done;
        byte[] init = Prepare.EvmCode.ForInitOf(code).Done;
        Address target = ContractAddress.From(TestItem.AddressB, 0);
        BlockPair blocks = chains.ProcessBlock(
            TxCreateContract(TestItem.PrivateKeyB, 0, init),
            TxToContract(TestItem.PrivateKeyC, target, 0, []));
        blocks.AssertFullMatch(2);
    }
}


```

</details>

<details>
<summary>Local reproduction source: ReviewPerformanceTests.cs</summary>

These temporary review probes were run locally, not committed. For the partial integration files, change the existing `BlockStmIntegrationTests` declaration to `partial` to reuse its existing fixture.

```csharp
using System;
using System.Collections;
using System.Diagnostics;
using System.Reflection;
using System.Runtime.CompilerServices;
using Nethermind.Consensus.Processing.ParallelProcessing.BlockStm;
using Nethermind.Core.Test.Builders;
using NUnit.Framework;

namespace Nethermind.Consensus.Test.Processing.ParallelProcessing.BlockStm;

public class ReviewPerformanceTests
{
    [Test]
    public void Review_memory_and_writer_lookup_cost()
    {
        _ = new MultiVersionMemory(1);
        long before = GC.GetAllocatedBytesForCurrentThread();
        MultiVersionMemory memory = new(1024);
        long allocated = GC.GetAllocatedBytesForCurrentThread() - before;
        TestContext.Out.WriteLine($"Logical processors={Environment.ProcessorCount}; empty MVMM(1024)={allocated} bytes; key={Unsafe.SizeOf<ParallelStateKey>()} bytes; read={Unsafe.SizeOf<Read>()} bytes");
        ParallelStateKey key = ParallelStateKey.ForAccount(TestItem.AddressA);
        for (int i = 0; i < 1024; i++) memory.Record(new TxVersion(i, 0), [], new() { [key] = "v" });
        IDictionary map = (IDictionary)typeof(MultiVersionMemory).GetField("_keyWriters", BindingFlags.Instance | BindingFlags.NonPublic)!.GetValue(memory)!;
        object writer = map[key]!;
        List<int> indices = (List<int>)writer.GetType().GetField("_indicesDesc", BindingFlags.Instance | BindingFlags.NonPublic)!.GetValue(writer)!;
        Func<int, int> original = writer.GetType().GetMethod("FindHighestBelow")!.CreateDelegate<Func<int, int>>(writer);
        Func<int, int> binary = upper =>
        {
            lock (indices)
            {
                int lo = 0, hi = indices.Count;
                while (lo < hi)
                {
                    int mid = lo + ((hi - lo) >> 1);
                    if (indices[mid] >= upper) lo = mid + 1;
                    else hi = mid;
                }
                return lo < indices.Count ? indices[lo] : -1;
            }
        };
        for (int upper = 0; upper <= 1024; upper++) Assert.That(binary(upper), Is.EqualTo(original(upper)));
        const int count = 100_000;
        static long Run(Func<int, int> lookup, int count)
        {
            long sum = 0;
            for (int i = 0; i < count; i++) sum += lookup(i & 1023);
            return sum;
        }
        Run(original, count); Run(binary, count);
        for (int run = 0; run < 5; run++)
        {
            Stopwatch timer = Stopwatch.StartNew();
            long first = Run(original, count);
            double linearMs = timer.Elapsed.TotalMilliseconds;
            timer.Restart();
            long second = Run(binary, count);
            double binaryMs = timer.Elapsed.TotalMilliseconds;
            Assert.That(second, Is.EqualTo(first));
            TestContext.Out.WriteLine($"{count} hot-key predecessor queries, 1024 writers, run {run}: linear={linearMs:F3} ms binary={binaryMs:F3} ms");
        }
    }
}

```

</details>

PR #9809, opened 2025-11-27
Make IProtocolsManager customizable

ak88:


asdacap:
Can you describe what do you want customized?

PR #9818, opened 2025-11-28
Fix ColumnsDb constructor recursion and reuse resolved keys

soniseth0:
redirect ColumnsDb’s public ctor to a private overload that accepts both the resolved keys and their string column names, compute enum fallbacks and ToString() names exactly once via ResolveKeys, keep the existing fallback to enumerate T when no keys are provided. І also reopened the pull request with these updated changes.


LukaszRozmej:
please fix whitespace

PR #9819, opened 2025-11-28
fix: avoid allocation in PutSpan by throwing immediately

Fibonacci747:
Changed RpcDb.PutSpan to throw immediately instead of allocating value.ToArray() before calling Set, which always throws with “RPC DB does not support writes”. This removes needless allocation and GC pressure, aligns with the read-only intent already expressed by Set/Remove/StartWriteBatch, and mirrors the early-throw pattern used by ReadOnlyDb.PutSpan when writes are not allowed.

### DIFF
--- a/src/Nethermind/Nethermind.Db.Rpc/RpcDb.cs
+++ b/src/Nethermind/Nethermind.Db.Rpc/RpcDb.cs
@@ -51,7 +51,7 @@ namespace Nethermind.Db.Rpc
 
         public void Set(ReadOnlySpan<byte> key, byte[] value, WriteFlags flags = WriteFlags.None)
         {
-            throw new InvalidOperationException("RPC DB does not support writes");
+            ThrowWritesNotSupported();
         }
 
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Db.Rpc/RpcDb.cs
```diff
@@ -113,7 +113,7 @@ public Span<byte> GetSpan(ReadOnlySpan<byte> key)
 
         public void PutSpan(ReadOnlySpan<byte> key, ReadOnlySpan<byte> value, WriteFlags writeFlags)
         {
-            Set(key, value.ToArray(), writeFlags);
+            throw new InvalidOperationException("RPC DB does not support writes");
```
LukaszRozmej: Unify both throws into ThrowXXX method

Fibonacci747: > Unify both throws into ThrowXXX method

ok

PR #9820, opened 2025-11-28
chore: remove unused ShouldSerialize hooks and fix nullable properties

sashass1315:
Replace Newtonsoft-style ShouldSerialize hooks with System.Text.Json-compatible null-based omission by removing ShouldSerializeMixHash, ShouldSerializeNonce, ShouldSerializeSignature, and ShouldSerializeStep from BlockForRpc. System.Text.Json does not call these methods and the project’s EthereumJsonSerializer uses DefaultIgnoreCondition = WhenWritingNull, so conditional serialization must be driven by nullability. Adjust MixHash, Nonce, and Signature to nullable types to reflect actual object state for AuRa vs non-AuRa blocks, ensuring properties are omitted when unset and eliminating misleading non-null annotations. This aligns with existing patterns like SimulateBlockResult and prevents confusion while keeping the JSON output unchanged.

### DIFF
--- a/src/Nethermind/Nethermind.Optimism/CL/L2Api.cs
+++ b/src/Nethermind/Nethermind.Optimism/CL/L2Api.cs
@@ -52,7 +52,7 @@ public class L2Api(
             EIP1559Params = block.ExtraData.Length == 0 ? null : block.ExtraData[1..],
             GasLimit = block.GasLimit,
             ParentBeaconBlockRoot = block.ParentBeaconBlockRoot,
-            PrevRandao = block.MixHash,
+            PrevRandao = block.MixHash!,
             SuggestedFeeRecipient = block.Miner,
             Timestamp = block.Timestamp.ToUInt64(null),
             Withdrawals = block.Withdrawals?.ToArray()


src/Nethermind/Nethermind.Facade/Eth/BlockForRpc.cs
```diff
@@ -106,29 +106,23 @@ public BlockForRpc(Block block, bool includeFullTransactionData, ISpecProvider s
     [JsonIgnore(Condition = JsonIgnoreCondition.Never)]
     public Bloom LogsBloom { get; set; }
     public Address Miner { get; set; }
-    public Hash256 MixHash { get; set; }
-
-    public bool ShouldSerializeMixHash() => !_isAuRaBlock && MixHash is not null;
```
LukaszRozmej, line 111: Does it makes sense to keep `_isAuRaBlock` as field then?

sashass1315: > Does it makes sense to keep `_isAuRaBlock` as field then?

I think no

LukaszRozmej (approved):

sashass1315:
> please fix build

Hi Boss! I'm sorry for failling CI and delay, i've fixed the problem

image

PR #9821, opened 2025-11-28
Remove redundant branch hash lookup in EngineModuleTests V4

soniseth0:
drop the unused `lastHash` assignment in `NewPayloadV4_reject_payload_with_bad_authorization_list_rlp`, stop computing the branch hash just to discard it while keeping the branch setup side effect intact, keep test semantics unchanged while reducing needless work


LukaszRozmej:
Not really meaningful change

PR #9825, opened 2025-11-28
Replace Enumerable.Range().Select() with direct for loops to reduce allocations

0xLogicalx:
Replaces LINQ-based task collection initialization with direct `for` loops in performance-critical code paths, eliminating unnecessary intermediate allocations.

### DIFF
--- a/src/Nethermind/Nethermind.Network/PeerManager.cs
+++ b/src/Nethermind/Nethermind.Network/PeerManager.cs
@@ -221,9 +221,30 @@ namespace Nethermind.Network
             Channel<Peer> taskChannel = Channel.CreateBounded<Peer>(1);
             using ArrayPoolList<Task> tasks = new(_outgoingConnectParallelism);
             for (int idx = 0; idx < _outgoingConnectParallelism; idx++)
+        {
+            tasks.Add(RunWorker(idx));
+        }
+        
+        async Task RunWorker(int workerIdx)
+        {
+            await foreach (Peer peer in taskChannel.Reader.ReadAllAsync(_cancellationTokenSource.Token))
             {
-                int workerIdx = idx;
-                tasks.Add(Task.Run(async () =>
+                try
+                {
+                    await SetupOutgoingPeerConnection(peer);
+                }
+                catch (TaskCanceledException)
+                {
+                    if (_logger.IsDebug) _logger.Debug($"Connect worker {workerIdx} cancelled");
+                    break;
+                }
+                catch (Exception e)
+                {
+                    if (_logger.IsError) _logger.Error($"Error setting up connection to {peer}, {e}");
+                }
+            }
+            if (_logger.IsDebug) _logger.Debug($"Connect worker {workerIdx} completed");
+        }
                 {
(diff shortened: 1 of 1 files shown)


benaadams (approved):

src/Nethermind/Nethermind.Network/PeerManager.cs
```diff
@@ -219,27 +219,32 @@ private class CandidateSelection
         private async Task RunPeerUpdateLoop()
         {
             Channel<Peer> taskChannel = Channel.CreateBounded<Peer>(1);
-            using ArrayPoolList<Task> tasks = Enumerable.Range(0, _outgoingConnectParallelism).Select(async idx =>
+            using ArrayPoolList<Task> tasks = new(_outgoingConnectParallelism);
+            for (int idx = 0; idx < _outgoingConnectParallelism; idx++)
             {
-                await foreach (Peer peer in taskChannel.Reader.ReadAllAsync(_cancellationTokenSource.Token))
+                int workerIdx = idx;
```
LukaszRozmej: we can probably avoid closure with something like:

```
for (int idx = 0; idx < _outgoingConnectParallelism; idx++)
{
    tasks.Add(RunWorker(idx));
}

async Task RunWorker(int workerIdx)
{
    await foreach (Peer peer in taskChannel.Reader.ReadAllAsync(_cancellationTokenSource.Token))
    {
        try
        {
            await SetupOutgoingPeerConnection(peer);
        }
        catch (TaskCanceledException)
        {
            if (_logger.IsDebug) _logger.Debug($"Connect worker {workerIdx} cancelled");
            break;
        }
        catch (Exception e)
        {
            if (_logger.IsError) _logger.Error($"Error setting up connection to {peer}, {e}");
        }
    }
    if (_logger.IsDebug) _logger.Debug($"Connect worker {workerIdx} completed");
}
```

LukaszRozmej (changes requested):

src/Nethermind/Nethermind.Network/PeerManager.cs
0xLogicalx: Thanks for the suggestion! I switched the loop to call a local RunWorker(int workerIdx) helper so each worker receives its own index instead of capturing the loop variable. This removes the closure while keeping the behavior the same.

LukaszRozmej (approved):

src/Nethermind/Nethermind.Network/PeerManager.cs
```diff
@@ -219,27 +219,53 @@ private class CandidateSelection
...
+                            // This is strictly speaking not related to the connection, but something outside of it.
+                            if (_logger.IsError) _logger.Error($"Error setting up connection to {peer}, {e}");
+                        }
                     }
-                }
-                if (_logger.IsDebug) _logger.Debug($"Connect worker {idx} completed");
-            }).ToPooledList(_outgoingConnectParallelism);
+                    if (_logger.IsDebug) _logger.Debug($"Connect worker {workerIdx} completed");
+                }));
+            }
```
LukaszRozmej, line 268:
```suggestion
```

PR #9826, opened 2025-11-28
fix: add ReadOnlySpan<byte> overload for long BE decode

Galoretka:
Introduced a ReadOnlySpan overload for ToLongFromBigEndianByteArrayWithoutLeadingZeros in Int64Extensions and refactored the existing byte[] overload to delegate to it, unifying the decoding semantics with UInt64Extensions by delegating to the ulong span-based implementation and casting to long. The update removes a subtle correctness risk in the previous reverse-index add/shift algorithm where shift counts could wrap for inputs longer than eight bytes, and provides a zero-allocation path that matches the project’s preference for span-based APIs. Existing call sites continue working without modification, behavior for typical ≤8-byte inputs remains consistent, and decoding of oversized inputs now matches the well-defined forward accumulation semantics already used for ulong.


src/Nethermind/Nethermind.Core/Extensions/Int64Extensions.cs
```diff
@@ -91,21 +91,24 @@ public static string ToHexString(this in UInt256 value, bool skipLeadingZeros)
         return bytes.ToHexString(true, skipLeadingZeros, false);
     }
 
-    public static long ToLongFromBigEndianByteArrayWithoutLeadingZeros(this byte[]? bytes)
+    public static long ToLongFromBigEndianByteArrayWithoutLeadingZeros(this ReadOnlySpan<byte> bytes)
```
flcl42, line 91: Where is it used?

flcl42:
@claude review

flcl42:
@claude review again

src/Nethermind/Nethermind.Db/Blooms/BloomStorage.cs
```diff
@@ -65,7 +65,13 @@ private set
 
         public BloomStorage(IBloomConfig config, [KeyFilter(DbNames.Bloom)] IDb bloomDb, [KeyFilter(nameof(BloomStorage))] IFileStoreFactory fileStoreFactory)
         {
-            long Get(Hash256 key, long defaultValue) => bloomDb.Get(key)?.ToLongFromBigEndianByteArrayWithoutLeadingZeros() ?? defaultValue;
+            long Get(Hash256 key, long defaultValue)
+            {
+                Span<byte> bytes = bloomDb.GetSpan(key);
+                long value = bytes.IsNull() ? defaultValue : bytes.ToLongFromBigEndianByteArrayWithoutLeadingZeros();
+                bloomDb.DangerousReleaseMemory(bytes);
```
LukaszRozmej: this pattern is in so many places, maybe extract to a helper method?

LukaszRozmej (approved):

smartprogrammer93 (approved):

PR #9833, opened 2025-11-29
Parallel Bloom Storage

benaadams:
## Changes

- Takes a significant time of ForkChoice and we put it into background in other places

image

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain/BlockTree.cs
+++ b/src/Nethermind/Nethermind.Blockchain/BlockTree.cs
@@ -219,7 +219,11 @@ namespace Nethermind.Blockchain
                 SetTotalDifficulty(header);
             }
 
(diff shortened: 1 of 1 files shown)


LukaszRozmej (commented):
1. Consistency - bloom storage needs to be updated before we update the head/main chain so when query comes it is already updated.
2. With @alexb5dh log index we will be retiring and removing bloom storage altogether after a while.

flcl42 (approved):

flcl42:
Do we still need this change?

LukaszRozmej:
Closing in favor of removing `BloomStorage` altogether — tracked in #11880. With @alexb5dh's log index (#8464) we no longer need to invest in the bloom index path. Thanks for the patch!

PR #9835, opened 2025-11-29
Optimize BlockParameter parsing

benaadams:
## Changes

- Single case-insentive pass
- Jump table on length for common cases

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain/Find/BlockParameter.cs
+++ b/src/Nethermind/Nethermind.Blockchain/Find/BlockParameter.cs
@@ -8,9 +8,9 @@ using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
-using System.Text.Json.Serialization;
-
+using System.Text;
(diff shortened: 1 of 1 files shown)


flcl42 (approved):

src/Nethermind/Nethermind.Blockchain/Find/BlockParameter.cs
```diff
@@ -181,28 +184,62 @@ private BlockParameter ReadObjectFormat(ref Utf8JsonReader reader, Type typeToCo
...
+                        return BlockParameter.Earliest;
+                    break;
+                case 9:
+                    if (Ascii.EqualsIgnoreCase(span, "finalized"u8))
+                        return BlockParameter.Finalized;
+                    break;
+            }
+
+            // Unknown tag or 0x quantity etc
+            return ReadStringFormatOther(span);
```
LukaszRozmej, line 219: keep it switch expression?

benaadams: Doesn't generate as good asm

PR #9838, opened 2025-11-29
fix: clear _accountChanges in ResetBlockAccessIndex to prevent cross-block leaks

qu0b:


qu0b:
I just tried this out to see if it might be the cause for why we are seeing 0x03 pre-compile in blocks, while besu and reth are not including this pre-compile. @Marchhill

PR #9842, opened 2025-11-30
perf: avoid extra allocation by iterating snapshot

Fibonacci747:
Use the already created snapshot array in IntersectWith instead of calling this.ToArray(). This removes dead code and prevents an extra allocation without changing semantics. The approach now mirrors SymmetricExceptWith, ensuring safe iteration while mutating the set.

### DIFF
--- a/src/Nethermind/Nethermind.Core/Collections/LinkedHashSet.cs
+++ b/src/Nethermind/Nethermind.Core/Collections/LinkedHashSet.cs
@@ -72,7 +72,7 @@ namespace Nethermind.Core.Collections
             T[] ts = new T[Count];
             CopyTo(ts, 0);
             HashSet<T> set = other.ToHashSet();
-            foreach (T t in this.ToArray())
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Core/Collections/LinkedHashSet.cs
```diff
@@ -72,7 +72,7 @@ public void IntersectWith(IEnumerable<T>? other)
             T[] ts = new T[Count];
             CopyTo(ts, 0);
             HashSet<T> set = other.ToHashSet();
-            foreach (T t in this.ToArray())
+            foreach (T t in ts)
```
LukaszRozmej, line 75: Won't `Remove` break enumerator?

LukaszRozmej: Ah it is already a copy

PR #9846, opened 2025-11-30
feat(evm): generify VirtualMachine with policy-based gas tracking

AnkushinDaniil:
This PR introduces a generic gas policy system (`IGasPolicy<TSelf>`) that enables chains to implement multi-dimensional gas accounting (Arbitrum's MultiGas) while maintaining zero-overhead for standard Ethereum execution via `SimpleGasPolicy`.
 

-  Related PR: NethermindEth/nethermind-arbitrum#448

  
## Changes

  - Add `IGasPolicy<TSelf>` interface with static abstract methods for gas tracking extensibility
  - Add `GasState` struct that wraps `RemainingGas` + `PolicyData` for complex gas accounting
  - Add `SimpleGasPolicy` struct implementing `IGasPolicy` for standard Ethereum
  - Rename `VirtualMachine` to `VirtualMachine<TGasPolicy>` with generic gas policy type parameter
  - Add `EthereumVirtualMachine` as non-generic alias for `VirtualMachine<SimpleGasPolicy>` for backward compatibility
  - Update all EVM instruction signatures from `ref long gasAvailable` to `ref GasState gasState`
  - Update `EvmCalculations` methods (`UpdateGas`, `UpdateMemoryCost`, `ChargeAccountAccessGas`, etc.) to use generic gas policy

  ## Types of changes

  - [ ] Optimization
  - [x] Refactoring
  - [ ] Documentation update
  - [ ] Build-related changes
  - [ ] Other: _Description_

  ## Testing

  #### Requires testing

  - [x] Yes
  - [ ] No

  - [x] Yes
  - [ ] No

  #### Notes on testing

  - Existing EVM tests validate backward compatibility through `EthereumVirtualMachine` alias

  ## Documentation

  - [ ] Yes
  - [x] No

  - [ ] Yes
  - [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Evm.Benchmark/EvmBenchmarks.cs
+++ b/src/Nethermind/Nethermind.Evm.Benchmark/EvmBenchmarks.cs
@@ -11,7 +11,7 @@ using Nethermind.Core.Test;
 using Nethermind.Db;
 using Nethermind.Evm.CodeAnalysis;
 using Nethermind.Specs;
-using Nethermind.Evm.Gas;
+using Nethermind.Evm.GasPolicy;
 using Nethermind.Evm.State;
 using Nethermind.Evm.Tracing;
 using Nethermind.Int256;
--- a/src/Nethermind/Nethermind.Evm.Benchmark/MultipleUnsignedOperations.cs
+++ b/src/Nethermind/Nethermind.Evm.Benchmark/MultipleUnsignedOperations.cs
@@ -11,7 +11,7 @@ using Nethermind.Core.Specs;
 using Nethermind.Core.Test;
 using Nethermind.Db;
 using Nethermind.Evm.CodeAnalysis;
-using Nethermind.Evm.Gas;
+using Nethermind.Evm.GasPolicy;
 using Nethermind.Evm.State;
 using Nethermind.Evm.Tracing;
 using Nethermind.Int256;
--- a/src/Nethermind/Nethermind.Evm.Benchmark/StaticCallBenchmarks.cs
+++ b/src/Nethermind/Nethermind.Evm.Benchmark/StaticCallBenchmarks.cs
@@ -12,7 +12,7 @@ using Nethermind.Core.Test;
 using Nethermind.Db;
 using Nethermind.Evm.CodeAnalysis;
 using Nethermind.Specs;
-using Nethermind.Evm.Gas;
+using Nethermind.Evm.GasPolicy;
 using Nethermind.Evm.State;
 using Nethermind.Evm.Tracing;
 using Nethermind.Int256;


src/Nethermind/Nethermind.Evm/Gas/GasState.cs
```diff
@@ -0,0 +1,42 @@
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+namespace Nethermind.Evm.Gas;
+
+/// <summary>
+/// Generic gas state container that can hold both simple and complex gas accounting.
+/// </summary>
+public struct GasState
```
LukaszRozmej: Make it a ref struct

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -87,12 +92,13 @@ public unsafe partial class VirtualMachine(
...
     private ICodeInfoRepository _codeInfoRepository;
 
-    private OpCode[] _opcodeMethods;
+    private Array _opcodeMethods;
     private static long _txCount;
 
     protected EvmState _currentState;
     protected ReadOnlyMemory<byte>? _previousCallResult;
     protected UInt256 _previousCallOutputDestination;
+    private GasState _transactionGasState = default;
```
LukaszRozmej: Why do we need it as a field? Prevents `GasState` to be `ref struct`

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -867,18 +875,22 @@ private void PrepareOpcodes<TTracingInst>(IReleaseSpec spec)
                 spec.EvmInstructionsNoTrace = GenerateOpCodes<TTracingInst>(spec);
             }
             // Ensure the non-traced opcode set is generated and assign it to the _opcodeMethods field.
-            _opcodeMethods = (OpCode[])(spec.EvmInstructionsNoTrace ??= GenerateOpCodes<TTracingInst>(spec));
+            _opcodeMethods = spec.EvmInstructionsNoTrace ??=
+                GenerateOpCodes<TTracingInst>(spec);
```
LukaszRozmej: avoid pointless whitespace changes! Makes the PR twice as hard to read and some of those whitespace also decrease readability.

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -867,18 +875,22 @@ private void PrepareOpcodes<TTracingInst>(IReleaseSpec spec)
...
         }
     }
 
-    protected virtual OpCode[] GenerateOpCodes<TTracingInst>(IReleaseSpec spec)
+    protected virtual Array GenerateOpCodes<TTracingInst>(IReleaseSpec spec)
         where TTracingInst : struct, IFlag
-        => EvmInstructions.GenerateOpCodes<TTracingInst>(spec);
+    {
+        return EvmInstructions.GenerateOpCodes<TGasPolicy, TTracingInst>(spec);
+    }
```
LukaszRozmej: revert to shorter syntax

src/Nethermind/Nethermind.Core/TransactionReceipt.cs
```diff
@@ -65,6 +66,12 @@ public TxReceipt(TxReceipt other)
         public LogEntry[]? Logs { get; set; }
         public string? Error { get; set; }
 
+        /// <summary>
+        /// Policy-specific gas data.
+        /// Null for Ethereum and chains using simple gas.
+        /// </summary>
+        public object? PolicyData { get; set; }
```
LukaszRozmej: Not sure if this should live in base class or some inheritor?

src/Nethermind/Nethermind.Blockchain/Tracing/BlockReceiptsTracer.cs
```diff
@@ -35,7 +35,7 @@ public class BlockReceiptsTracer : IBlockTracer, ITxTracer, IJournal<int>, ITxTr
 
     public void MarkAsSuccess(Address recipient, GasConsumed gasSpent, byte[] output, LogEntry[] logs, Hash256? stateRoot = null)
     {
-        _txReceipts.Add(BuildReceipt(recipient, gasSpent.SpentGas, StatusCode.Success, logs, stateRoot));
+        _txReceipts.Add(BuildReceipt(recipient, gasSpent.SpentGas, StatusCode.Success, logs, gasSpent.PolicyData, stateRoot));
```
LukaszRozmej: just pass `gasSpent` to `BuildReceipt` and `BuildFailedReceipt`, so we can ignore `PolicyData`

src/Nethermind/Nethermind.Evm/TransactionProcessing/GasConsumed.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Nethermind.Evm.TransactionProcessing;
 
-public readonly record struct GasConsumed(long SpentGas, long OperationGas)
+public readonly record struct GasConsumed(long SpentGas, long OperationGas, object? PolicyData = null)
```
LukaszRozmej: I would love to remove `PolicyData` from here if possible
But probably would have to make a lot of things generic and might not be worth it

src/Nethermind/Nethermind.Evm/TransactionSubstate.cs
```diff
@@ -58,6 +58,12 @@ public readonly ref struct TransactionSubstate
     public IToArrayCollection<LogEntry> Logs => _logs ?? _emptyLogs;
     public IHashSetEnumerableCollection<Address> DestroyList => _destroyList ?? _emptyDestroyList;
 
+    /// <summary>
+    /// Policy-specific data for this transaction (e.g., multigas for Arbitrum).
+    /// Null for standard Ethereum transactions using SimpleGasPolicy.
+    /// </summary>
+    public object? PolicyData { get; set; }
```
LukaszRozmej: Does this needs to be settable? Can't it come only on init in constructor?

We want to keep this struct `readonly` for performance reasons.

src/Nethermind/Nethermind.Evm/TransactionSubstate.cs
```diff
@@ -16,7 +16,7 @@
 
 namespace Nethermind.Evm;
 
-public readonly ref struct TransactionSubstate
+public ref struct TransactionSubstate
```
LukaszRozmej: keep it readonly

src/Nethermind/Nethermind.Evm/Gas/GasState.cs
```diff
@@ -0,0 +1,42 @@
...
+        PolicyData = null;
+    }
+
+    /// <summary>
+    /// Returns a string representation of the gas state for debugging.
+    /// </summary>
+    public readonly override string ToString()
+    {
+        return PolicyData is null ? $"Gas: {RemainingGas}" : $"Gas: {RemainingGas}, PolicyData: {PolicyData}";
+    }
```
LukaszRozmej:
```suggestion
    public readonly override string ToString() =>
        PolicyData is null ? $"Gas: {RemainingGas}" : $"Gas: {RemainingGas}, PolicyData: {PolicyData}";
```

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Call.cs
```diff
@@ -135,7 +135,16 @@ public static EvmExceptionType InstructionCall<TOpCall, TTracingInst>(
...
         // Charge gas for accessing the account's code (including delegation logic if applicable).
-        if (!EvmCalculations.ChargeAccountAccessGasWithDelegation(ref gasAvailable, vm, codeSource)) goto OutOfGas;
+        Instruction instruction = TOpCall.ExecutionType switch
+        {
+            ExecutionType.CALL => Instruction.CALL,
+            ExecutionType.CALLCODE => Instruction.CALLCODE,
+            ExecutionType.DELEGATECALL => Instruction.DELEGATECALL,
+            ExecutionType.STATICCALL => Instruction.STATICCALL,
+            _ => Instruction.CALL
+        };
```
LukaszRozmej: There is `ToInstruction` extension method already

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Create.cs
```diff
@@ -109,6 +110,9 @@ public static EvmExceptionType InstructionCreate<TOpCreate, TTracingInst>(
                 goto OutOfGas;
         }
 
+        // Determine the instruction type based on the create operation type.
+        Instruction instruction = typeof(TOpCreate) == typeof(OpCreate) ? Instruction.CREATE : Instruction.CREATE2;
```
LukaszRozmej:
```suggestion
        Instruction instruction = TOpCreate.OpCode;
```

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.cs
```diff
@@ -1,30 +1,34 @@
...
-    public static OpCode[] GenerateOpCodes<TTracingInst>(IReleaseSpec spec)
+    public static Array GenerateOpCodes<TGasPolicy, TTracingInst>(IReleaseSpec spec)
+        where TGasPolicy : struct, IGasPolicy<TGasPolicy>
         where TTracingInst : struct, IFlag
     {
         // Allocate lookup table for all possible opcodes.
-        OpCode[] lookup = new OpCode[byte.MaxValue + 1];
+        var lookup =
+            new delegate*<VirtualMachine<TGasPolicy>, ref EvmStack, ref GasState, ref int, EvmExceptionType>[byte
+                .MaxValue + 1];
```
LukaszRozmej: why the complication and lack of alias?

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Stack.cs
```diff
@@ -498,19 +505,23 @@ public static EvmExceptionType InstructionPush0<TTracingInst>(VirtualMachine vm,
...
-    public static EvmExceptionType InstructionPush<TOpCount, TTracingInst>(VirtualMachine vm, ref EvmStack stack, ref long gasAvailable, ref int programCounter)
+    public static EvmExceptionType InstructionPush<TGasPolicy, TOpCount, TTracingInst>(VirtualMachine<TGasPolicy> vm,
+        ref EvmStack stack, ref GasState gasState, ref int programCounter)
+        where TGasPolicy : struct, IGasPolicy<TGasPolicy>
         where TOpCount : IOpCount
         where TTracingInst : struct, IFlag
     {
         // Deduct a very low gas cost for the push operation.
-        gasAvailable -= GasCostOf.VeryLow;
+        TGasPolicy.ConsumeGas(ref gasState, GasCostOf.VeryLow, Instruction.PUSH1 + (byte)(TOpCount.Count - 1));
```
LukaszRozmej: `Instruction.PUSH1 + (byte)(TOpCount.Count - 1)` 

should be generalized to some helper method (used in multiple places here) like `TOpCount.GetInstruction(Instruction baseInstruction)`

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -29,21 +30,25 @@
...
     IBlockhashProvider? blockHashProvider,
     ISpecProvider? specProvider,
     ILogManager? logManager
-) : VirtualMachine(blockHashProvider, specProvider, logManager)
+) : VirtualMachine<SimpleGasPolicy>(blockHashProvider, specProvider, logManager)
 {
 }
 
-public unsafe partial class VirtualMachine(
+public unsafe partial class VirtualMachine<TGasPolicy>(
```
LukaszRozmej: Try moving static fields out of generic type

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -1204,7 +1216,7 @@ protected CallResult ExecuteCall<TTracingInst>(
     [SkipLocalsInit]
     protected virtual unsafe CallResult RunByteCode<TTracingInst, TCancelable>(
         scoped ref EvmStack stack,
-        long gasAvailable)
+        GasState gasState)
```
LukaszRozmej:
```suggestion
        in GasState gasState)
```

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -1238,7 +1254,8 @@ protected virtual unsafe CallResult RunByteCode<TTracingInst, TCancelable>(
             {
 #if DEBUG
                 // Allow the debugger to inspect and possibly pause execution for debugging purposes.
-                debugger?.TryWait(ref _currentState, ref programCounter, ref gasAvailable, ref stack.Head);
+                var currentGasForDebugger = TGasPolicy.GetRemainingGas(in gasState);
```
LukaszRozmej: avoid var

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -1258,19 +1276,20 @@ protected virtual unsafe CallResult RunByteCode<TTracingInst, TCancelable>(
...
                 if (Instruction.POP == instruction)
                 {
-                    exceptionType = EvmInstructions.InstructionPop(this, ref stack, ref gasAvailable, ref programCounter);
+                    exceptionType = EvmInstructions.InstructionPop(this, ref stack, ref gasState, ref programCounter);
                 }
                 else
                 {
                     // Retrieve the opcode function pointer corresponding to the current instruction.
-                    OpCode opcodeMethod = opcodeMethods[(int)instruction];
+                    delegate*<VirtualMachine<TGasPolicy>, ref EvmStack, ref GasState, ref int, EvmExceptionType>
```
LukaszRozmej: keep OpCode alias

AnkushinDaniil: C# doesn't support generic type aliases

src/Nethermind/Nethermind.Evm/Gas/SimpleGasPolicy.cs
```diff
@@ -0,0 +1,58 @@
...
+public readonly struct SimpleGasPolicy : IGasPolicy<SimpleGasPolicy>
+{
+    /// <summary>
+    /// Initialize gas state for a new transaction.
+    /// </summary>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static GasState<SimpleGasPolicy> InitializeForTransaction(long gasLimit, long intrinsicGas)
+    {
+        return new GasState<SimpleGasPolicy>(gasLimit);
+    }
```
LukaszRozmej:
```suggestion
    public static GasState<SimpleGasPolicy> InitializeForTransaction(long gasLimit, long intrinsicGas) =>
        new GasState<SimpleGasPolicy>(gasLimit); 
```

src/Nethermind/Nethermind.Evm/Gas/SimpleGasPolicy.cs
```diff
@@ -0,0 +1,58 @@
...
+
+    /// <summary>
+    /// Get remaining gas for OOG checks.
+    /// </summary>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static long GetRemainingGas(in GasState<SimpleGasPolicy> gasState)
+    {
+        return gasState.RemainingGas;
+    }
```
LukaszRozmej:
```suggestion
    public static long GetRemainingGas(in GasState<SimpleGasPolicy> gasState) => gasState.RemainingGas;
```

src/Nethermind/Nethermind.Evm/Gas/SimpleGasPolicy.cs
```diff
@@ -0,0 +1,58 @@
...
+
+    /// <summary>
+    /// Consume gas for an operation.
+    /// </summary>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static void ConsumeGas(ref GasState<SimpleGasPolicy> gasState, long gasCost, Instruction instruction)
+    {
+        gasState.RemainingGas -= gasCost;
+    }
```
LukaszRozmej:
```suggestion
    public static void ConsumeGas(ref GasState<SimpleGasPolicy> gasState, long gasCost, Instruction instruction) =>
        gasState.RemainingGas -= gasCost;
```

src/Nethermind/Nethermind.Evm/Gas/SimpleGasPolicy.cs
```diff
@@ -0,0 +1,58 @@
...
+
+    /// <summary>
+    /// Refund unused gas (e.g., from failed CALL/CREATE).
+    /// </summary>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static void RefundGas(ref GasState<SimpleGasPolicy> gasState, long gasAmount)
+    {
+        gasState.RemainingGas += gasAmount;
+    }
```
LukaszRozmej:
```suggestion
    public static void RefundGas(ref GasState<SimpleGasPolicy> gasState, long gasAmount) =>
        gasState.RemainingGas += gasAmount;
```

src/Nethermind/Nethermind.Evm/Gas/SimpleGasPolicy.cs
```diff
@@ -0,0 +1,58 @@
...
+
+    /// <summary>
+    /// Mark the gas state as out of gas.
+    /// </summary>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static void SetOutOfGas(ref GasState<SimpleGasPolicy> gasState)
+    {
+        gasState.RemainingGas = 0;
+    }
```
LukaszRozmej:
```suggestion
    public static void SetOutOfGas(ref GasState<SimpleGasPolicy> gasState) =>
        gasState.RemainingGas = 0;
```

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Bitwise.cs
```diff
@@ -34,18 +35,21 @@ public interface IOpBitwise
...
     [SkipLocalsInit]
-    public static EvmExceptionType InstructionBitwise<TOpBitwise>(VirtualMachine _, ref EvmStack stack, ref long gasAvailable, ref int programCounter)
+    public static EvmExceptionType InstructionBitwise<TGasPolicy, TOpBitwise>(VirtualMachine<TGasPolicy> _,
+        ref EvmStack stack, ref GasState<TGasPolicy> gasState, ref int programCounter)
+        where TGasPolicy : struct, IGasPolicy<TGasPolicy>
         where TOpBitwise : struct, IOpBitwise
     {
         // Deduct the operation's gas cost.
-        gasAvailable -= TOpBitwise.GasCost;
+        TGasPolicy.ConsumeGas(ref gasState, TOpBitwise.GasCost, Instruction.AND);
```
LukaszRozmej: Shouldn't it be:

```suggestion
        TGasPolicy.ConsumeGas(ref gasState, TOpBitwise.GasCost, TOpBitwise.OpCode);
```

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Call.cs
```diff
@@ -135,7 +135,9 @@ public static EvmExceptionType InstructionCall<TOpCall, TTracingInst>(
             goto StackUnderflow;
 
         // Charge gas for accessing the account's code (including delegation logic if applicable).
-        if (!EvmCalculations.ChargeAccountAccessGasWithDelegation(ref gasAvailable, vm, codeSource)) goto OutOfGas;
+        var instruction = TOpCall.ExecutionType.ToInstruction();
```
LukaszRozmej: avoid var

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Call.cs
```diff
@@ -170,10 +172,12 @@ public static EvmExceptionType InstructionCall<TOpCall, TTracingInst>(
...
-        if (!EvmCalculations.UpdateGas(spec.GetCallCost(), ref gasAvailable) ||
-            !EvmCalculations.UpdateMemoryCost(vm.EvmState, ref gasAvailable, in dataOffset, dataLength) ||
-            !EvmCalculations.UpdateMemoryCost(vm.EvmState, ref gasAvailable, in outputOffset, outputLength) ||
-            !EvmCalculations.UpdateGas(gasExtra, ref gasAvailable))
+        if (!EvmCalculations.UpdateGas<TGasPolicy>(ref gasState, spec.GetCallCost(), instruction) ||
+            !EvmCalculations.UpdateMemoryCost<TGasPolicy>(vm.EvmState, ref gasState, in dataOffset, dataLength,
+                instruction) ||
+            !EvmCalculations.UpdateMemoryCost<TGasPolicy>(vm.EvmState, ref gasState, in outputOffset, outputLength,
+                instruction) ||
+            !EvmCalculations.UpdateGas<TGasPolicy>(ref gasState, gasExtra, instruction))
```
LukaszRozmej: keep old whitespace - easier to read if per line

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Call.cs
```diff
@@ -183,9 +187,14 @@ public static EvmExceptionType InstructionCall<TOpCall, TTracingInst>(
         if (spec.IsEip7907Enabled)
         {
             uint excessContractSize = (uint)Math.Max(0, codeInfo.CodeSpan.Length - CodeSizeConstants.MaxCodeSizeEip170);
-            if (excessContractSize > 0 && !ChargeForLargeContractAccess(excessContractSize, codeSource, in vm.EvmState.AccessTracker, ref gasAvailable))
+            if (excessContractSize > 0 && !ChargeForLargeContractAccess<TGasPolicy>(excessContractSize, codeSource,
+                    in vm.EvmState.AccessTracker, ref gasState, instruction))
```
LukaszRozmej: revert whitespace

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Call.cs
```diff
@@ -183,9 +187,14 @@ public static EvmExceptionType InstructionCall<TOpCall, TTracingInst>(
...
                 goto OutOfGas;
         }
+
+        // Get remaining gas for 63/64 calculation
+        var gasAvailable = TGasPolicy.GetRemainingGas(in gasState);
```
LukaszRozmej: avoid var

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.CodeCopy.cs
```diff
@@ -261,7 +279,7 @@ public static EvmExceptionType InstructionExtCodeSize<TTracingInst>(
                 vm.OpCodeCount++;
                 programCounter++;
                 // Deduct very-low gas cost for the next operation (ISZERO, GT, or EQ).
-                gasAvailable -= GasCostOf.VeryLow;
+                TGasPolicy.ConsumeGas(ref gasState, GasCostOf.VeryLow, Instruction.ISZERO);
```
LukaszRozmej:
```suggestion
                TGasPolicy.ConsumeGas(ref gasState, GasCostOf.VeryLow, nextInstruction);
```

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.cs
```diff
@@ -1,30 +1,34 @@
...
+using System;
 using System.Runtime.CompilerServices;
 using Nethermind.Core;
 using Nethermind.Core.Specs;
+using Nethermind.Evm.Gas;
 
 [assembly: InternalsVisibleTo("Nethermind.Evm.Precompiles")]
 namespace Nethermind.Evm;
 
-using unsafe OpCode = delegate*<VirtualMachine, ref EvmStack, ref long, ref int, EvmExceptionType>;
```
LukaszRozmej, line 11: keep alias

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Environment.cs
```diff
@@ -307,18 +350,21 @@ public static EvmExceptionType InstructionBlkUInt64<TOpEnv, TTracingInst>(Virtua
...
-    public static EvmExceptionType InstructionEnv32Bytes<TOpEnv, TTracingInst>(VirtualMachine vm, ref EvmStack stack, ref long gasAvailable, ref int programCounter)
-        where TOpEnv : struct, IOpEnv32Bytes
+    public static EvmExceptionType InstructionEnv32Bytes<TGasPolicy, TOpEnv, TTracingInst>(
+        VirtualMachine<TGasPolicy> vm, ref EvmStack stack, ref GasState<TGasPolicy> gasState, ref int programCounter)
+        where TGasPolicy : struct, IGasPolicy<TGasPolicy>
+        where TOpEnv : struct, IOpEnv32Bytes<TGasPolicy>
         where TTracingInst : struct, IFlag
     {
-        gasAvailable -= TOpEnv.GasCost;
+        TGasPolicy.ConsumeGas(ref gasState, TOpEnv.GasCost, Instruction.ORIGIN);
```
LukaszRozmej:
```suggestion
        TGasPolicy.ConsumeGas(ref gasState, TOpEnv.GasCost, TOpEnv.OpCode);
```

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Shifts.cs
```diff
@@ -38,22 +39,25 @@ public interface IOpShift
...
-    public static EvmExceptionType InstructionShift<TOpShift, TTracingInst>(VirtualMachine vm, ref EvmStack stack, ref long gasAvailable, ref int programCounter)
+    public static EvmExceptionType InstructionShift<TGasPolicy, TOpShift, TTracingInst>(VirtualMachine<TGasPolicy> vm,
+        ref EvmStack stack, ref GasState<TGasPolicy> gasState, ref int programCounter)
+        where TGasPolicy : struct, IGasPolicy<TGasPolicy>
         where TOpShift : struct, IOpShift
         where TTracingInst : struct, IFlag
     {
         // Deduct gas cost specific to the shift operation.
-        gasAvailable -= TOpShift.GasCost;
+        TGasPolicy.ConsumeGas(ref gasState, TOpShift.GasCost, Instruction.SHL);
```
LukaszRozmej:
```suggestion
        TGasPolicy.ConsumeGas(ref gasState, TOpShift.GasCost, TOpShift.OpCode);
```

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Stack.cs
```diff
@@ -498,19 +513,23 @@ public static EvmExceptionType InstructionPush0<TTracingInst>(VirtualMachine vm,
...
-    public static EvmExceptionType InstructionPush<TOpCount, TTracingInst>(VirtualMachine vm, ref EvmStack stack, ref long gasAvailable, ref int programCounter)
+    public static EvmExceptionType InstructionPush<TGasPolicy, TOpCount, TTracingInst>(VirtualMachine<TGasPolicy> vm,
+        ref EvmStack stack, ref GasState<TGasPolicy> gasState, ref int programCounter)
+        where TGasPolicy : struct, IGasPolicy<TGasPolicy>
         where TOpCount : IOpCount
         where TTracingInst : struct, IFlag
     {
         // Deduct a very low gas cost for the push operation.
-        gasAvailable -= GasCostOf.VeryLow;
+        TGasPolicy.ConsumeGas(ref gasState, GasCostOf.VeryLow, GetInstruction<TOpCount>(Instruction.PUSH1));
```
LukaszRozmej:
```suggestion
        TGasPolicy.ConsumeGas(ref gasState, GasCostOf.VeryLow, TOpCount.GetInstruction(Instruction.PUSH1));
```

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Storage.cs
```diff
@@ -282,13 +301,17 @@ public static EvmExceptionType InstructionMCopy<TTracingInst>(VirtualMachine vm,
         if (!stack.PopUInt256(out UInt256 a) || !stack.PopUInt256(out UInt256 b) || !stack.PopUInt256(out UInt256 c)) goto StackUnderflow;
 
         // Calculate additional gas cost based on the length (using a division rounding-up method) and deduct the total cost.
-        gasAvailable -= GasCostOf.VeryLow + GasCostOf.VeryLow * EvmCalculations.Div32Ceiling(c, out bool outOfGas);
+        var copyGasCost = GasCostOf.VeryLow + GasCostOf.VeryLow * EvmCalculations.Div32Ceiling(c, out var outOfGas);
```
LukaszRozmej: avoid var

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.cs
AnkushinDaniil: C# doesn't support generic type aliases

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
AnkushinDaniil: use `scoped ref` instead

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -1230,7 +1242,11 @@ protected virtual unsafe CallResult RunByteCode<TTracingInst, TCancelable>(
         // Pin the opcode methods array to obtain a fixed pointer, avoiding repeated bounds checks and casts.
         // If we don't use a pointer we have to cast for each call (delegate*<...> can't be used as a generic arg)
         // Or have bounds checks (however only 256 opcodes and opcode is a byte so know always in bounds).
-        fixed (OpCode* opcodeMethods = &_opcodeMethods[0])
+        var opcodeArray =
+            (delegate*<VirtualMachine<TGasPolicy>, ref EvmStack, ref GasState<TGasPolicy>, ref int, EvmExceptionType>[])
+            _opcodeMethods;
+        fixed (delegate*<VirtualMachine<TGasPolicy>, ref EvmStack, ref GasState<TGasPolicy>, ref int, EvmExceptionType>*
+               opcodeMethods = &opcodeArray[0])
```
AnkushinDaniil, line 1235: C# doesn't support generic type aliases

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -29,50 +31,62 @@
...
-    internal static readonly OutOfGasException PrecompileOutOfGasException = new();
+    public static readonly PrecompileExecutionFailureException PrecompileExecutionFailureException = new();
+    public static readonly OutOfGasException PrecompileOutOfGasException = new();
+}
 
+/// <summary>
+/// Backward-compatible alias for VirtualMachine with simple single-dimensional gas.
+/// Allows existing code to extend VirtualMachine without specifying a gas policy.
+/// </summary>
+public class VirtualMachine(
```
LukaszRozmej: Keep the name `EthereumVirtualMachine`
Keep it sealed

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.cs
AnkushinDaniil: The alias isn't used

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Call.cs
```diff
@@ -149,31 +150,28 @@ public static EvmExceptionType InstructionCall<TOpCall, TTracingInst>(
...
         {
-            gasExtra += GasCostOf.NewAccount;
+            if (!TGasPolicy.ChargeCallExtra(ref gasState, GasCostOf.NewAccount, isNewAccount: true)) goto OutOfGas;
         }
 
-        // Update gas: call cost, memory expansion for input and output, and extra gas.
-        if (!EvmCalculations.UpdateGas(spec.GetCallCost(), ref gasAvailable) ||
-            !EvmCalculations.UpdateMemoryCost(vm.EvmState, ref gasAvailable, in dataOffset, dataLength) ||
-            !EvmCalculations.UpdateMemoryCost(vm.EvmState, ref gasAvailable, in outputOffset, outputLength) ||
-            !EvmCalculations.UpdateGas(gasExtra, ref gasAvailable))
```
LukaszRozmej, line 176: this changes the order of charging, not sure if it can break something?
Do you need `ChargeCallExtra` and can't just use here `TGasPolicy.UpdateGas` with `gasExtra`?

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.ControlFlow.cs
```diff
@@ -173,7 +180,7 @@ public static EvmExceptionType InstructionRevert(VirtualMachine vm, ref EvmStack
         }
 
         // Ensure sufficient gas for any required memory expansion.
-        if (!EvmCalculations.UpdateMemoryCost(vm.EvmState, ref gasAvailable, in position, in length) ||
+        if (!TGasPolicy.UpdateMemoryCost(vm.EvmState, ref gasState, in position, in length) ||
```
LukaszRozmej: I know it was like that in original, but maybe worth re-ordering to:
```suggestion
        if (!TGasPolicy.UpdateMemoryCost(ref gasState, in position, in length, vm.EvmState) ||
```

src/Nethermind/Nethermind.Evm/Gas/EthereumGasPolicy.cs
```diff
@@ -0,0 +1,173 @@
...
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static long GetRemainingGas(in GasState<EthereumGasPolicy> gasState) => gasState.RemainingGas;
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static void ConsumeGas(ref GasState<EthereumGasPolicy> gasState, long gasCost) =>
+        gasState.RemainingGas -= gasCost;
+
+    public static void ChargeSelfDestructGas(ref GasState<EthereumGasPolicy> gasState)
+        => gasState.RemainingGas -= GasCostOf.SelfDestructEip150;
```
LukaszRozmej:
```suggestion
        => ConsumeGas(ref gasState, GasCostOf.SelfDestructEip150);
```

src/Nethermind/Nethermind.Evm/Gas/IGasPolicy.cs
```diff
@@ -0,0 +1,166 @@
...
+public interface IGasPolicy<TSelf> where TSelf : struct, IGasPolicy<TSelf>
+{
+    /// <summary>
+    /// Initialize gas state for a new transaction with intrinsic gas already deducted.
+    /// Called by TransactionProcessor before ExecuteTransaction.
+    /// </summary>
+    /// <param name="gasLimit">The gas limit provided for the transaction.</param>
+    /// <param name="intrinsicGas">Intrinsic gas already calculated.</param>
+    /// <returns>Initialized gas state with remaining gas</returns>
+    static abstract GasState<TSelf> InitializeForTransaction(long gasLimit, long intrinsicGas);
```
LukaszRozmej: This seems to be only called by Warmup, not in `TransactionProcessor`?

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.cs
```diff
@@ -1,30 +1,34 @@
...
-    public static OpCode[] GenerateOpCodes<TTracingInst>(IReleaseSpec spec)
+    public static Array GenerateOpCodes<TGasPolicy, TTracingInst>(IReleaseSpec spec)
+        where TGasPolicy : struct, IGasPolicy<TGasPolicy>
         where TTracingInst : struct, IFlag
     {
         // Allocate lookup table for all possible opcodes.
-        OpCode[] lookup = new OpCode[byte.MaxValue + 1];
+        var lookup =
+            new delegate*<VirtualMachine<TGasPolicy>, ref EvmStack, ref GasState<TGasPolicy>, ref int, EvmExceptionType>[byte
+                .MaxValue + 1];
```
LukaszRozmej: keep it as one line

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.cs
```diff
@@ -1,30 +1,34 @@
...
     /// Generates the opcode lookup table for the Ethereum Virtual Machine.
     /// Each of the 256 entries in the returned array corresponds to an EVM instruction,
     /// with unassigned opcodes defaulting to a bad instruction handler.
     /// </summary>
+    /// <typeparam name="TGasPolicy">The gas policy type used for gas accounting.</typeparam>
     /// <typeparam name="TTracingInst">A struct implementing IFlag used for tracing purposes.</typeparam>
     /// <param name="spec">The release specification containing enabled features and opcode flags.</param>
     /// <returns>An array of function pointers (opcode handlers) indexed by opcode value.</returns>
-    public static OpCode[] GenerateOpCodes<TTracingInst>(IReleaseSpec spec)
+    public static Array GenerateOpCodes<TGasPolicy, TTracingInst>(IReleaseSpec spec)
```
LukaszRozmej: why `Array`? Shouldn't we have concreate type?

src/Nethermind/Nethermind.Evm/Gas/EthereumGasPolicy.cs
```diff
@@ -0,0 +1,173 @@
...
+        ref GasState<EthereumGasPolicy> gasState,
+        long topicCount,
+        long dataSize,
+        long totalCost)
+    {
+        if (gasState.RemainingGas < totalCost)
+            return false;
+        gasState.RemainingGas -= totalCost;
+        return true;
+    }
```
LukaszRozmej:
```suggestion
        long totalCost) =>
        UpdateGas(ref gasState, totalCost);
```

src/Nethermind/Nethermind.Evm/Gas/EthereumGasPolicy.cs
```diff
@@ -0,0 +1,173 @@
...
+    public static bool ChargeCallExtra(
+        ref GasState<EthereumGasPolicy> gasState,
+        long cost,
+        bool isNewAccount)
+    {
+        if (gasState.RemainingGas < cost)
+            return false;
+        gasState.RemainingGas -= cost;
+        return true;
+    }
```
LukaszRozmej:
```suggestion
        bool isNewAccount) =>
        UpdateGas(ref gasState, totalCost);
```

src/Nethermind/Nethermind.Evm/Gas/EthereumGasPolicy.cs
```diff
@@ -0,0 +1,173 @@
...
+    public static bool ChargeStorageWrite(
+        ref GasState<EthereumGasPolicy> gasState,
+        long cost,
+        bool isSlotCreation)
+    {
+        if (gasState.RemainingGas < cost)
+            return false;
+        gasState.RemainingGas -= cost;
+        return true;
+    }
```
LukaszRozmej:
```suggestion
        bool isSlotCreation) =>
        UpdateGas(ref gasState, totalCost);
```

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Storage.cs
```diff
@@ -197,8 +202,8 @@ public static EvmExceptionType InstructionMStore8<TTracingInst>(VirtualMachine v
         EvmState vmState = vm.EvmState;
 
         // Update the memory cost for a single-byte extension; if insufficient, signal out-of-gas.
-        if (!EvmCalculations.UpdateMemoryCost(vmState, ref gasAvailable, in result, in UInt256.One) ||
-            !vmState.Memory.TrySaveByte(in result, data))
+        if (!TGasPolicy.UpdateMemoryCost(vmState, ref gasState, in result, in UInt256.One) ||
+        !vmState.Memory.TrySaveByte(in result, data))
```
LukaszRozmej: make it one-liner

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -87,7 +97,7 @@ public unsafe partial class VirtualMachine(
 
     private ICodeInfoRepository _codeInfoRepository;
 
-    private OpCode[] _opcodeMethods;
+    private Array _opcodeMethods;
```
LukaszRozmej: why not concreate type?

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -964,22 +974,22 @@ private void RevertParityTouchBugAccount()
         }
     }
 
-    protected static bool UpdateGas(long gasCost, ref long gasAvailable)
+    protected static bool UpdateGas(long gasCost, ref GasState<TGasPolicy> gasState)
```
LukaszRozmej: this whole method can be replaced by `TGasPolicy.UpdateGas`

LukaszRozmej:
Also merge master in

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Call.cs
AnkushinDaniil: The order change is intentional. In the original code, all extra gas (call value + new account) was combined into a single `gasExtra` variable and charged once at the end.

  The order shouldn't affect consensus - gas is fungible, and if any charge fails, execution stops. Could you please mention someone who can verify it? However, `ChargeCallExtra` is needed for Arbitrum multigas because we need to distinguish:
  - Value transfer - Computation
  - New account - StorageGrowth

  Using plain `UpdateGas` would lose the resource kind categorisation. For `EthereumGasPolicy`, both implementations are identical, but `ArbitrumGasPolicy` will categorise differently.

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.cs
AnkushinDaniil: It will look like

```suggestion
    public static delegate*<VirtualMachine<TGasPolicy>, ref EvmStack, ref GasState<TGasPolicy>, ref int, EvmExceptionType>[] GenerateOpCodes<TGasPolicy, TTracingInst>(IReleaseSpec spec)
```
I think `Array` is better. However, if you prefer, I can change it to a concreate type.

LukaszRozmej: Does it cause any future unnecessary casts?

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
AnkushinDaniil: The same comment regarding the concrete type is applicable here

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.cs
AnkushinDaniil: Yes, fixed

smartprogrammer93 (changes requested):
The IGasPolicy depends on GasState, but if i have multi-deminsional gas, then i need to be able to choose a different struct, so it should be a generic. 
Also, the update gas and consume gas in gasPolicy does not take opcode as a param. if i have a different pricing policy for opcodes and precompiles for my chain, i didnt gain anything from this gas policy class.

Am i missing something here?

src/Nethermind/Nethermind.Consensus/Stateless/StatelessBlockProcessingEnv.cs
```diff
@@ -81,7 +82,7 @@ private IBlockProcessor GetProcessor()
     private ITransactionProcessor CreateTransactionProcessor(IWorldState state, IBlockhashCache blockhashCache)
     {
         BlockhashProvider blockhashProvider = new(blockhashCache, state, logManager);
-        VirtualMachine vm = new(blockhashProvider, specProvider, logManager);
-        return new TransactionProcessor(BlobBaseFeeCalculator.Instance, specProvider, state, vm, new EthereumCodeInfoRepository(state), logManager);
+        EthereumVirtualMachine vm = new(blockhashProvider, specProvider, logManager);
+        return new TransactionProcessor<EthereumGas>(BlobBaseFeeCalculator.Instance, specProvider, state, vm, new EthereumCodeInfoRepository(state), logManager);
```
LukaszRozmej: Should it be?

```suggestion
        return new EthereumTransactionProcessor(BlobBaseFeeCalculator.Instance, specProvider, state, vm, new EthereumCodeInfoRepository(state), logManager);
```

src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs
```diff
@@ -124,8 +124,8 @@ private IntrinsicGasTxValidator() { }
     public ValidationResult IsWellFormed(Transaction transaction, IReleaseSpec releaseSpec)
     {
         // This is unnecessarily calculated twice - at validation and execution times.
-        IntrinsicGas intrinsicGas = IntrinsicGasCalculator.Calculate(transaction, releaseSpec);
-        return transaction.GasLimit < intrinsicGas.MinimalGas
+        EthereumIntrinsicGas ethereumIntrinsicGas = IntrinsicGasCalculator.Calculate(transaction, releaseSpec);
+        return transaction.GasLimit < ethereumIntrinsicGas.MinimalGas
```
LukaszRozmej: maybe don't rename variables?

src/Nethermind/Nethermind.Evm.Benchmark/EvmBenchmarks.cs
```diff
@@ -32,7 +33,7 @@ public class EvmBenchmarks
         private IVirtualMachine _virtualMachine;
         private BlockHeader _header = new BlockHeader(Keccak.Zero, Keccak.Zero, Address.Zero, UInt256.One, MainnetSpecProvider.IstanbulBlockNumber, Int64.MaxValue, 1UL, Bytes.Empty);
         private IBlockhashProvider _blockhashProvider = new TestBlockhashProvider();
-        private EvmState _evmState;
+        private VmState<EthereumGas> _evmState;
```
LukaszRozmej: name `EvmState`?

```suggestion
        private EvmState _evmState;
```

src/Nethermind/Nethermind.Evm/Gas/EthereumGas.cs
```diff
@@ -0,0 +1,207 @@
...
+using Nethermind.Core.Extensions;
+using Nethermind.Core.Specs;
+using Nethermind.Int256;
+
+namespace Nethermind.Evm.Gas;
+
+/// <summary>
+/// Standard Ethereum single-dimensional gas.Value policy.
+/// </summary>
+public struct EthereumGas : IGas<EthereumGas>
```
LukaszRozmej: Can it be a `ref struct`?

src/Nethermind/Nethermind.Evm/Gas/EthereumGas.cs
```diff
@@ -0,0 +1,207 @@
...
+        return totalZeros + (data.Length - totalZeros) * txDataNonZeroMultiplier;
+    }
+
+    private static long AuthorizationListCost(Transaction tx, IReleaseSpec spec)
+    {
+        AuthorizationTuple[]? authList = tx.AuthorizationList;
+        if (authList is not null)
+        {
+            if (!spec.IsAuthorizationListEnabled)
+                throw new InvalidDataException($"Transaction with an authorization list received within the context of {spec.Name}. EIP-7702 is not enabled.");
```
LukaszRozmej: minor: move to separate ThrowXXX method

src/Nethermind/Nethermind.Evm/GasPolicy/IGasPolicy.cs
```diff
@@ -0,0 +1,190 @@
...
+    /// <returns>A new gas instance</returns>
+    static abstract TSelf FromLong(long value);
+
+    /// <summary>
+    /// Get the remaining single-dimensional gas available for execution.
+    /// This is what's checked against zero to detect out-of-gas conditions.
+    /// </summary>
+    /// <param name="gas"></param>
+    /// <returns>Remaining gas (negative values indicate out-of-gas)</returns>
+    static abstract long GetRemainingGas(in TSelf gas);
```
LukaszRozmej, line 31: Should this still return `long`?

src/Nethermind/Nethermind.Evm/GasPolicy/IGasPolicy.cs
```diff
@@ -0,0 +1,190 @@
...
+
+    /// <summary>
+    /// Consume gas for an EVM operation.
+    /// </summary>
+    /// <param name="gas"></param>
+    /// <param name="cost"></param>
+    static abstract void Consume(ref TSelf gas, long cost);
```
LukaszRozmej, line 38: should this take `long` as cost?

src/Nethermind/Nethermind.Evm/GasPolicy/IGasPolicy.cs
```diff
@@ -0,0 +1,190 @@
...
+    /// <param name="gasCost">The gas cost to deduct.</param>
+    /// <returns><c>true</c> if there was enough gas; otherwise, <c>false</c>.</returns>
+    static abstract bool UpdateGas(ref TSelf gas, long gasCost);
+
+    /// <summary>
+    /// Refunds gas by adding the specified amount back to the available gas.
+    /// </summary>
+    /// <param name="gas"></param>
+    /// <param name="refund">The gas amount to refund.</param>
+    static abstract void UpdateGasUp(ref TSelf gas, long refund);
```
LukaszRozmej, line 145: same here

src/Nethermind/Nethermind.Evm/Gas/IGas.cs
```diff
@@ -0,0 +1,190 @@
...
+    static abstract bool ConsumeStorageWrite(ref TSelf gas, long cost, bool isSlotCreation);
+
+    /// <summary>
+    /// Charges gas for CALL value transfer or new account creation.
+    /// </summary>
+    /// <param name="gas"></param>
+    /// <param name="cost">The gas cost to charge.</param>
+    /// <param name="isNewAccount">True if creating a new account (address was empty).</param>
+    /// <returns>True if sufficient gas available</returns>
+    static abstract bool ConsumeCallExtra(ref TSelf gas, long cost, bool isNewAccount);
```
LukaszRozmej: Shouldn't all those `long cost` be `TSelf cost`?

src/Nethermind/Nethermind.Evm.Benchmark/EvmBenchmarks.cs
AnkushinDaniil: If rename, we need to introduce virtual method for `_statePool.Enqueue(this);` that overrides pooling behavior. Without it, when disposing `EvmState`, it would be returned to the base class pool instead of its own type-specific pool. This virtualisation can affect performance.

src/Nethermind/Nethermind.Evm/GasPolicy/IGasPolicy.cs
AnkushinDaniil: I've thought about it, and this way we'll need to make every gas generic, including the opcode cost. This can be done in the future and should benefit @smartprogrammer93. Currently, this method is only used for one resource type. If the cost is `ArbitrumGas`, we need to iterate through all dimensions that affect performance. This is related to the similar comments.

src/Nethermind/Nethermind.Evm/Gas/EthereumGas.cs
AnkushinDaniil: No. `ref struct` can't be used as a generic type argument. Regular struct with `ref` parameters should give the same performance

LukaszRozmej: It can be used with dotnet 10! Just needs additional constraint `allows ref struct`
But might not play well with making gas multi-dimensional

src/Nethermind/Nethermind.Evm/GasPolicy/IGasPolicy.cs
LukaszRozmej: Yes I think it could look like:

```suggestion
    static abstract void Consume(ref TSelf gas, in TSelf cost);
```

We could have this working with current gas scheme probably without performance penalty.

LukaszRozmej: Maybe let's explore it in follow-up PR!

AnkushinDaniil: Yes, I think it's the best option and I'm still not sure how vector sum affects performance

src/Nethermind/Nethermind.Evm/Gas/EthereumGas.cs
AnkushinDaniil: VmState has TGas field:
```csharp
  public class VmState<TGas>
  {
      public TGas Gas;
```
A ref struct can only be declared as a field of another ref struct , not in classes or non-ref struct types. We could try cascading `ref struct` through, but the problem is that `VmState` is a pooled class using `ConcurrentQueue`. `ref struct`s cannot be stored in collections, pooled or returned from async methods. There are 2 options (I can see) that can work:
Pass through `TGas` as a `ref` parameter;
Keep it a regular `struct`, as I mentioned earlier, it shouldn't affect performance

smartprogrammer93 (approved):
I dont like it. why this complexity. Now if you want to introduce the other dimension, you need to make changes in the EthereumVirtual machine or the instructions themselves. and then there, you would need to have two calls, one consumes one gas dimension and another consumes another gas dimension. imo, all of this logic can just be moved to the gasPolicy. vm just calls once, and depending on gasPolicy implementation, gas dimensions are consumed. and you can have one EthereumGasPolicy, and then one ArbitrumGasPolicy, and then a composite GasPolicy that combines both and calls both, all implementing the same interface.

making the virtual machine and instructions aware of the GasDimensions does not look like the right approach to me.

src/Nethermind/Nethermind.Evm/GasPolicy/EthereumGasPolicy.cs
```diff
@@ -0,0 +1,235 @@
...
+                // Otherwise, if warm access should be charged, apply the warm read cost.
+                result = UpdateGas(ref gas, GasCostOf.WarmStateRead);
+            }
+        }
+
+        return result;
+    }
+
+    public static bool ConsumeStorageAccessGas(ref EthereumGasPolicy gas,
+        VirtualMachine<EthereumGasPolicy> vm,
```
smartprogrammer93: again, cant just pass the vm into such a call, that doesnt look right

PR #9847, opened 2025-12-01
fix: dispose automatic pruning triggers to stop background timers

MozirDmitriy:
Problem: Lifecycle leak. PathSizePruningTrigger instances created in PruningTrieStateFactory.InitializeFullPruning() are not disposed; timers keep running.
Fix: Lifecycle leak fixed by disposing automatic pruning triggers via disposeStack.

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain/FullPruning/CompositePruningTrigger.cs
+++ b/src/Nethermind/Nethermind.Blockchain/FullPruning/CompositePruningTrigger.cs
@@ -11,10 +11,10 @@ namespace Nethermind.Blockchain.FullPruning;
 /// </summary>
 public class CompositePruningTrigger : IPruningTrigger, IDisposable
 {
-    private readonly List<IPruningTrigger> _triggers = new List<IPruningTrigger>();
+    private readonly List<IPruningTrigger> _triggers = [];
 
     /// <summary>
-    /// Adds new <see cref="IPruningTrigger"/> to the be watched."/>
+    /// Adds new <see cref="IPruningTrigger"/> to be watched.
     /// </summary>
     /// <param name="trigger">trigger to be watched</param>
     public void Add(IPruningTrigger trigger)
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Init/PruningTrieStateFactory.cs
```diff
@@ -151,6 +151,10 @@ private void InitializeFullPruning(IDb stateDb,
             if (pruningTrigger is not null)
             {
                 compositePruningTrigger.Add(pruningTrigger);
+                if (pruningTrigger is IDisposable d)
+                {
+                    disposeStack.Push(d);
+                }
```
LukaszRozmej: Hm... I think those should be disposed through disposing compositePruningTrigger, so it should implement IDisposable and be added to stack and then it should internally iterate and check if it's members are disposable

LukaszRozmej (changes requested):

src/Nethermind/Nethermind.Init/PruningTrieStateFactory.cs
MozirDmitriy: > Hm... I think those should be disposed through disposing compositePruningTrigger, so it should implement IDisposable and be added to stack and then it should internally iterate and check if it's members are disposable

Alright, updated

src/Nethermind/Nethermind.Blockchain/FullPruning/CompositePruningTrigger.cs
```diff
@@ -2,20 +2,24 @@
...
 
 namespace Nethermind.Blockchain.FullPruning;
 
 /// <summary>
 /// Allows to have multiple <see cref="IPruningTrigger"/>s.
 /// </summary>
-public class CompositePruningTrigger : IPruningTrigger
+public class CompositePruningTrigger : IPruningTrigger, IDisposable
 {
+    private readonly List<IPruningTrigger> _triggers = new List<IPruningTrigger>();
```
LukaszRozmej:
```suggestion
    private readonly List<IPruningTrigger> _triggers = [];
```

PR #9848, opened 2025-12-01
Dispose plugins when supported

flcl42:
## Changes

- Dispose plugins when supported

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Runner.Test/PluginDisposalTests.cs
+++ b/src/Nethermind/Nethermind.Runner.Test/PluginDisposalTests.cs
@@ -7,13 +7,13 @@ using Autofac;
 using Nethermind.Api;
 using Nethermind.Api.Extensions;
 using Nethermind.Config;
-using Nethermind.Consensus;
 using Nethermind.Core;
 using Nethermind.Logging;
 using Nethermind.Runner.Ethereum.Modules;
 using Nethermind.Serialization.Json;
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Init/Steps/InitializePlugins.cs
```diff
@@ -36,5 +40,18 @@ public async Task Execute(CancellationToken cancellationToken)
                 }
             }
         }
+
+        private void RegisterForShutdown(INethermindPlugin plugin)
+        {
+            switch (plugin)
+            {
+                case IAsyncDisposable asyncDisposable:
+                    (_disposeStack ??= api.DisposeStack).Push(asyncDisposable);
```
LukaszRozmej: IMO: Drop the field

```suggestion
                    api.DisposeStack.Push(asyncDisposable);
```

asdacap:
In `NethermindRunnerModule`, line 85, the plugin is added as a singleton instance, which implicitly imply that it is owned externally. Just add another DLS like, `AddAndTakeOwnershipOfInstance` and modify it there, and there should not need to be other change.

PR #9849, opened 2025-12-01
Return all forks if needed

flcl42:
## Changes

- Add Nethermind-specific `debug_config` output for the complete ordered fork schedule.
- Return every configured fork, including historical entries, under `forks`; only the active entry includes `active: true`.
- Remove `current`, `next`, `last`, and application-version fields from the debug response while keeping standard `eth_config` limited to its existing three fields.

The full schedule is Nethermind-specific diagnostic data, so it belongs under `debug_` rather than the standard `eth_` namespace. This keeps `eth_config` aligned with [EIP-7910](https://eips.ethereum.org/EIPS/eip-7910) while exposing client-specific diagnostics through [Geth's debug namespace](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug). Application version information remains available through the standard `web3_clientVersion` and other existing client metadata surfaces.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

- Release builds pass with warnings treated as errors.
- Focused regression tests pass for the fork-only debug response and standard `eth_config` compatibility.
- The serialized debug response contains `forks` only, with one visible `active: true` marker and no app-version payload.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


src/Nethermind/Nethermind.Network/IForkInfo.cs
```diff
@@ -19,6 +20,8 @@ public interface IForkInfo
     ValidationResult ValidateForkId(ForkId peerId, BlockHeader? head);
 
     ForkActivationsSummary GetForkActivationsSummary(BlockHeader? head);
+
+    IReadOnlyList<Fork> GetAllForks();
```
LukaszRozmej: minor: instead make `IForkInfo : IEnumerable<Fork>` ?

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/IEthRpcModule.cs
```diff
@@ -289,7 +289,7 @@ ResultWrapper<TransactionForRpc> eth_getTransactionByBlockNumberAndIndex(
         [JsonRpcMethod(IsImplemented = true, Description = "Retrieves Account with code and no storageRoot via Address and Blocknumber", IsSharable = true)]
         ResultWrapper<AccountInfoForRpc?> eth_getAccountInfo([JsonRpcParameter(ExampleValue = "[\"0xaa00000000000000000000000000000000000000\", \"latest\"]")] Address accountAddress, BlockParameter? blockParameter = null);
 
-        [JsonRpcMethod(IsImplemented = true, Description = "Provides configuration data for the current and next fork", IsSharable = true)]
-        ResultWrapper<JsonNode> eth_config();
+        [JsonRpcMethod(IsImplemented = true, Description = "Provides configuration data for the current and next fork or the full fork schedule", IsSharable = true)]
+        ResultWrapper<JsonNode> eth_config([JsonRpcParameter(Description = "Returns every known fork when true", ExampleValue = "[true]")] bool? showAllForks = null);
```
LukaszRozmej: Why use nullable when we can just default to false?

```suggestion
        ResultWrapper<JsonNode> eth_config([JsonRpcParameter(Description = "Returns every known fork when true", ExampleValue = "[true]")] bool showAllForks = false);
```

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
```diff
@@ -802,38 +802,51 @@ private ResultWrapper<TResult> GetStateFailureResult<TResult>(BlockHeader header
...
+            Current = BuildForkConfig(forks.Current, _specProvider),
+            Next = GetForkConfigOrNull(forks.Next, _specProvider),
+            Last = GetForkConfigOrNull(forks.Last, _specProvider),
+            All = allForkConfigs
+        }, UnchangedDictionaryKeyOptions)));
+
+        static ForkConfig? GetForkConfigOrNull(Fork? fork, ISpecProvider specProvider) =>
+            fork is null ? null : BuildForkConfig(fork.Value, specProvider);
+
+        static ForkConfig BuildForkConfig(Fork fork, ISpecProvider specProvider)
```
LukaszRozmej: Should we cache `ForkConfig`s so we don't have to build them on every call? They won't change for the running time of the application (maybe on Arbitrum or Optimism they do?)

LukaszRozmej (approved):

src/Nethermind/Nethermind.Network/IForkInfo.cs
flcl42: Seems like IReadOnlyList and IEnumerable will lead to spawning an enumerator, maybe just Fork[]?

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/IEthRpcModule.cs
flcl42: Good idea

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
flcl42: Caching/performance is good, but on other side a bit more complexity and memory consumed. While method should be called rarely

flcl42: I doubt spec may change without restart

src/Nethermind/Nethermind.Network/IForkInfo.cs
LukaszRozmej: `IReadOnlySpan<Fork> GetAllForks();`

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
flcl42: Added simple cache

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
```diff
@@ -802,38 +805,54 @@ private ResultWrapper<TResult> GetStateFailureResult<TResult>(BlockHeader header
         };
     }
 
-    public ResultWrapper<JsonNode> eth_config()
+    public ResultWrapper<JsonNode> eth_config(bool showAllForks = false)
```
LukaszRozmej: why does it return `JsonNode` isn't it inefficient?

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
```diff
@@ -67,6 +68,8 @@ public partial class EthRpcModule(
     IForkInfo forkInfo,
     ulong? secondsPerSlot) : IEthRpcModule
 {
+    private static readonly ConcurrentDictionary<ForkId, ForkConfig> forkConfigCache = [];
```
LukaszRozmej: Could be `FrozenDictionary` if we built it once with all forks

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
flcl42: It's due to [custom rules of json serialization](https://github.com/NethermindEth/nethermind/blob/a326b0fca440a4808436f5b86e4e32a907d3b510/src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs?plain=1#L86C1-L86C147), that simplifies json comparison.

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
```diff
@@ -808,6 +808,20 @@ private ResultWrapper<TResult> GetStateFailureResult<TResult>(BlockHeader header
...
+            ReadOnlySpan<Fork> forkSchedule = forkInfo.GetAllForks();
+            Dictionary<ForkId, ForkConfig> allForks = new(forkSchedule.Length);
+
+            foreach (Fork scheduledFork in forkSchedule)
+            {
+                allForks.Add(scheduledFork.Id, BuildForkConfig(scheduledFork, _specProvider));
+            }
+
+            forkConfigCache = allForks.ToFrozenDictionary();
+        }
```
LukaszRozmej: run it in the background?

LukaszRozmej (dismissed):

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
flcl42: It's too much for such a small peace of code

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
```diff
@@ -802,38 +805,60 @@ private ResultWrapper<TResult> GetStateFailureResult<TResult>(BlockHeader header
...
-            if (fork is null)
+            ReadOnlySpan<Fork> forkSchedule = forkInfo.GetAllForks();
+            allForkConfigs = new(forkSchedule.Length);
+
+            foreach (Fork scheduledFork in forkSchedule)
             {
-                return null;
+                allForkConfigs.Add(forkConfigCache[scheduledFork.Id]);
             }
+        }
```
LukaszRozmej: Reuse FrozenDictionary.Keys

flcl42: I doubt the will be in the same order

LukaszRozmej: Frozen dictionary should keep order of the enumatation it is built upon

flcl42: It does not really keep (it returns them sorted but not in the initial order)
```
using System.Collections.Frozen;
using System.Collections.Immutable;

ImmutableArray<KeyValuePair<int, string>>.Builder allForks = ImmutableArray.CreateBuilder<KeyValuePair<int, string>>();

allForks.Add(new KeyValuePair<int, string>(2, "two"));
allForks.Add(new KeyValuePair<int, string>(1, "one"));
allForks.Add(new KeyValuePair<int, string>(3, "three"));

foreach (var item in allForks.ToFrozenDictionary().Values)
{
    Console.WriteLine(item);
}
```

flcl42: (but for some magic reason, I guess, related to no default comparer for ForkId, it works for our case. But we should not rely on it)

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/IEthRpcModule.cs
```diff
@@ -289,7 +288,7 @@ ResultWrapper<TransactionForRpc> eth_getTransactionByBlockNumberAndIndex(
         [JsonRpcMethod(IsImplemented = true, Description = "Retrieves Account with code and no storageRoot via Address and Blocknumber", IsSharable = true)]
         ResultWrapper<AccountInfoForRpc?> eth_getAccountInfo([JsonRpcParameter(ExampleValue = "[\"0xaa00000000000000000000000000000000000000\", \"latest\"]")] Address accountAddress, BlockParameter? blockParameter = null);
 
-        [JsonRpcMethod(IsImplemented = true, Description = "Provides configuration data for the current and next fork", IsSharable = true)]
-        ResultWrapper<JsonNode> eth_config();
+        [JsonRpcMethod(IsImplemented = true, Description = "Provides configuration data for the current and next fork or the full fork schedule", IsSharable = true)]
+        ResultWrapper<JsonNode> eth_config([JsonRpcParameter(Description = "[Nethermind only] Additionally returns every known forks when true", ExampleValue = "[true]")] bool showAllForks = false);
```
Marchhill:
```suggestion
        ResultWrapper<JsonNode> eth_config([JsonRpcParameter(Description = "[Nethermind only] Additionally returns all scheduled forks when true", ExampleValue = "[true]")] bool showAllForks = false);
```

Marchhill (changes requested):
Looks good, I agree with @deffrian though seems like should be in debug namespace since it's not a standard feature

benaadams (dismissed):

stdevMac:
@flcl42 stale-PR check: this has been blocked ~7 months on the design objection about non-spec fields in eth_config (reviewers preferred a debug_ namespace). Still pursuing, or OK to close? If there's no reply in the next few days, we'll close this as part of the stale-PR cleanup — it can always be reopened or resubmitted.

flcl42:
keep it will work on it later

PR #9852, opened 2025-12-01
Don't cache current block cache

LukaszRozmej:
## Changes

- Don't cache current block cache
- This makes cache usable for block production where current block hash is yet unknown
- Guard if Hash is null for current block

#### Requires testing

- [x] Yes

- [x] Yes

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain/BlockhashCache.cs
+++ b/src/Nethermind/Nethermind.Blockchain/BlockhashCache.cs
@@ -142,7 +142,7 @@ public class BlockhashCache(IHeaderFinder headerFinder, ILogManager logManager) 
                             int length = FlatCacheLength(blockHeader);
                             hashes = new Hash256[length];
                             hashes[0] = blockHeader.ParentHash;
-                            Array.Copy(parentHashes, 0, hashes, 1, Math.Min(length - 1, MaxDepth - 1));
+                            Array.Copy(parentHashes, 0, hashes, 1, length - 1);
                             if (!emptyHash)
                             {
                                 _flatCache.Set(blockHeader.Hash, hashes);


flcl42 (approved):

src/Nethermind/Nethermind.Blockchain/BlockhashProvider.cs
```diff
@@ -28,24 +28,34 @@ public class BlockhashProvider(
...
-                ? hashes[depth]
-                : blockhashCache.GetHash(currentBlock, (int)depth)
-                  ?? throw new InvalidDataException("Hash cannot be found when executing BLOCKHASH operation");
+
+            return depth switch
+            {
+                <= 0 or > MaxDepth => ReturnOutOfBounds(currentBlock, number),
+                1 => currentBlock.ParentHash,
+                _ => hashes is not null
+                    ? hashes[depth - 1]
```
dipkakwani, line 49: Do we not require the out of bound check? `depth - 1 < hashes.length` ?

LukaszRozmej: IMO we shouldn't need it, because:
- we exit early on `number < 0`
- we exit early on depth `<= 0 or > MaxDepth`
- we exit early on depth 1
- So now we are sure to have depth 1..Math.Min(currentBlock.Number - 1, MaxDepth), now:
  - We store in flatcache only if we have full array
  - in prefetch we take from flat cache and slide the array by 1 (or expand it if we are < MaxDepth)
  - so all it all it should be implicitly guarded here that we are in bounds

There might be a flaw in my logic though, please point it out if there is.

dipkakwani: @LukaszRozmej Replying here for https://github.com/NethermindEth/nethermind/pull/9874#discussion_r2584984819
From the taiko logs we actually hit `IndexOutOfRangeException` due to the missing check here, but if hashes are always >= requested depth with the new fix, this won't be the case. But don't see any harm adding the check here.

dipkakwani: @LukaszRozmej I am still seeing the failure without the bound check with Taiko:
```
03 Dec 13:06:04 | JSON RPC request engine_getPayloadV2 
03 Dec 13:06:04 | GetPayloadV2 result: 1 (0x282894...37d573). 
03 Dec 13:06:04 | JSON RPC request engine_newPayloadV2 
03 Dec 13:06:04 | Received New Block:  1 (0x282894...37d573)      | limit    16,000,000 👆 | Hex: 0x4b00 
03 Dec 13:06:04 | Removing beacon pivot, previous pivot:  
03 Dec 13:06:04 | OnInvalidBlock: 0x282894580af8933e4c78a0674f64b20a2be4045c48583cee9db12e159437d573 0x587f846fc42691b04906c27f90cf7a76ba35c7eba0d6b388150af463037ec3a0 
03 Dec 13:06:04 | Block validator rejected the block 1 (0x282894580af8933e4c78a0674f64b20a2be4045c48583cee9db12e159437d573). 
03 Dec 13:06:04 | Rejected invalid block 1 (0x282894580af8933e4c78a0674f64b20a2be4045c48583cee9db12e159437d573),  ExtraData: K, reason: Anchor transaction must have valid selector 
03 Dec 13:06:04 | OnInvalidBlock: 0x282894580af8933e4c78a0674f64b20a2be4045c48583cee9db12e159437d573 0x587f846fc42691b04906c27f90cf7a76ba35c7eba0d6b388150af463037ec3a0 
```

LukaszRozmej: interesting it is on block 1 for what number?

dipkakwani: @LukaszRozmej Ah sorry my bad, I forgot to include my latest taiko changes which are still in PR for testing.

LukaszRozmej: So if there is an issue it should be unit-testable, but I don't see it

PR #9854, opened 2025-12-01
Perf/simple flat

asdacap:
- Requires:
  - #10076
  - #10145
  - #10080
- Not actually simple.
- Based on #9402 , brings compacted statediff, except the diff also include trie now.
- Works by having separate columns for flat and trie along with a trie node cache.
- Persistence is abstracted via `IPersistence` for very different db scheme.
- Slow trie commit time is mitigated via a complicated trie warmer.
- No snap sync in this PR. Import only via "--FlatDb.ImportFromPruningTrieState true".
  - Snap sync is in #10426   

## Note

- Some optimization was removed to make this PR clearer or because its does not improve the performance significantly. These are:
  - Warmup slot from prewarmer thread. This will added in a separate PR as it modifies the worldstate a bit.
  - HintSet. Very hard to notice if it help or not.
  - Custom bloom filter for address and slots. Although faster, its redundant with rocksdb bloom and only improve performance by a tiny amount compared to the complexity it brings.

## Design

global_state

- The global state orchestration lives in `FlatDbManager` 
  - At the back of it is `IPersistence` which represent the persistence layer.
    - It has a create write batch and reader which rely on rocksdb snapshot.
  - On top of it, there is two layer of `Snapshot`, backed by `SnapshotContent`.
    - Snapshot is just written accounts,slots, and trie nodes per range of block.
    - The first layer consists of exactly one block sized snapshot.
    - The second layer are compactions of multiple blocks worth of snaphots. This is called compacted snapshots and it default size is either 32 (16 in the diagram) blocks or 4 blocks. 32 block for full compaction and 4 block for "mid" compaction.
    - In front of it is a `TrieNodeCache` which is a special `TrieNode`... cache. It stores the node in a sharded hashtable array. This cache is addressed by both path and hash. Its put in front because its updated on the head instead of on persist. But its reorg safe as the trie node still need keccak to match.

readonly_snapshot_bundle

- The `ReadOnlySnapshotBundle` is a bundle of snapshot.
  - Its the reader used by `StateReader` and the various world state. Its read only, refcounted and shared.
  - The use of the compacted second layer allow the `ReadOnlySnapshotBundle` to have a small-ish number of layers to traverse which is about 10, instead of 128 on mainnet. You can make it lower at expense of about 50% of a full CPU core and some good amount of memory, but it does not look like it help much.

scope_with_bundle(1)

- A worldstate scope consist of:
  - SnapshotBundle which consists of:
    - TransientResource which are some pooled large variables.
    - SnapshotContent which are mutable that will be put into a Snapshot on commit.
    - ReadOnlySnapshotBundle which are collection of snapshot and persistence's reader.
  - TrieNodeWarmer which is a dedicated special warmer just for node.
  - Read goes through `ReadOnlySnapshotBundle`.
  - A special `HintGet` method is added to the world state scope interface which allow it to trigger the trie node warmer on what key it should warm up.
  - (Not in this PR) Additionally, a `WarmUpOutOfScope` is added to the existing scope provider interface which allow prewarmer to send key to trie node warmer from outside the scope (prewarmer kinda runs outside the main block processing).

## Persistence

- Technically you can put the persistence to rocksdb, LMDB or Paprika, but so far I cant find a faster configuration than just plain RocksDb.
- There are three "layout" right now.

### Flat

- Flat layout is the default consisting of a rocksdb database with 7 columns.
  - Metadata
  - Account
    - 14GB uncompressed.
    - Stores the account encoded with slim RLP encoding (same as snap sync). This causes the database to be very small so might as well disable compression.
    - Key is `<20-byte-hashed-address-prefix>` (first 20 bytes of the hashed address).
  - Storage
    - 62GB compressed.
    - Stores the slot value with leading zeros removed.
    - Key is `<4-byte-hashed-address-prefix><32-byte-hashed-slotindex><16-byte-hashed-address-suffix>` (52 bytes total). The address hash is split to help RocksDB's comparator skip bytes during comparison and help with index key shortening.
  - StateTopNodes
    - 648M compressed.
    - Stores state trie nodes where path nibble count is 5 or less.
    - Separate from StateNodes to reduce compaction and improve block cache hit rate.
    - Key is `<3-byte-path>` where the last 4 bits encode the path length.
  - StateNodes
    - 34GB compressed.
    - Stores state trie nodes where path nibble count is 6-15.
    - Key is `<8-byte-path>` where the last 4 bits encode the path length.
  - StorageNodes
    - 143GB compressed.
    - Stores storage trie nodes where path nibble count is 0-15.
    - Key is `<4-byte-hashed-address-prefix><8-byte-path><16-byte-hashed-address-suffix>` (28 bytes total).
  - FallbackNodes
    - Stores both state and storage trie nodes where path nibble count is 16+.
    - State nodes key: `<0x00><32-byte-path><1-byte-length>` (34 bytes total).
    - Storage nodes key: `<0x01><4-byte-address-prefix><32-byte-path><1-byte-length><16-byte-address-suffix>` (54 bytes total).
- It assumes a reasonably high memory of at least 32GB on mainnet.
  - Rocksdb partitioned index is disabled, meaning large in memory index.
  - A 1 GB block cache configurable via `--FlatDb.BlockCacheSizeBudget` is set to the account and storage db at a 30/70 split. 
  - For flat columns, `optimize_filters_for_hits` is set to false to keep last level bloom, which mean even more memory usage.
  - A dedicated 1GB block cache is given to the flat column. (might be removed).
  - Index + bloom take up around 2.3GB on mainnet. This is proportional to the database size.
  - Total database size is 260 GB. 

### PreimageFlat

- Is exactly like Flat, but account and storage keys use the raw address/slot instead of their hashes.
- It is the fastest configuration, but it cannot snap sync.

### FlatInTrie

- The purpose of FlatInTrie layout is to cater for lower spec machine or very large network by relying more on database compression and using a database configuration that consume less memory.
- To do that, the account and storage data is embedded within `StateNodes` and `StorageNodes` (reusing the same columns as trie nodes). The key encoding is the same.
- There are 5 column in total. 
  - Metadata
  - StateTopNodes
  - StateNodes
  - StorageNodes
  - FallbackNodes
- Rocksdb's partitioned index is turned on and `optimize_filters_for_hits` is turned on. This significantly reduce memory at the expense of latency.  
- The same 1GB block cache remains but put to the state nodes and storage nodes db. 
- This results in better compression and since partitioned index is enabled for this layout, lower memory usage.
- In consume around 16MB of fixed index size (partitioned index means most index is lazy loaded as part of block cache). `optimize_filters_for_hits` means last level SST does not have bloom which slows down null slot check.
- Freshly imported database size is 193GB. For reference the state db size at this point (its a backup a few months back) is 186GB.
- However, this comes at the cost of lower performance.

## Performance

Screenshot_2026-01-28_12-42-59

- Graph is Flat, FlatInTrie, HalfPath. Note: halfpath have a bug here, it does not prune reliably.
- Test is running with 48GB of RAM limit via systemd-run. 
- In general, its about 20% higher mgas/sec than halfpath.

## Types of changes

- [x] Optimization

#### Requires testing

- [x] Yes

- [x] Yes

### DIFF
--- a/src/Nethermind/Nethermind.State.Flat/Persistence/RefCountingPersistenceReader.cs
+++ b/src/Nethermind/Nethermind.State.Flat/Persistence/RefCountingPersistenceReader.cs
@@ -30,7 +30,7 @@ public class RefCountingPersistenceReader : RefCountingDisposable, IPersistence.
                 await Task.Delay(60_000);
                 if (Volatile.Read(ref _leases.Value) <= NoAccessors) return;
                 if (logger.IsWarn)
-                    logger.Warn($"Unexpected old snapshot created. Lease count {_leases.Value}");
+                    logger.Warn($"Unexpected old snapshot created. Lease count {_leases.Value}. State {CurrentState}");
             }
         });
     }


benaadams:
```
Unexpected old snapshot created. Lease count 7 at    at System.Environment.get_StackTrace()
   at Nethermind.State.Flat.Persistence.RefCountingPersistenceReader..ctor(IPersistenceReader innerReader, ILogger logger) in D:\GitHub\nethermind\src\Nethermind\Nethermind.State\Flat\Persistence\RefCountingPersistenceReader.cs:line 21
   at Nethermind.State.Flat.FlatDiffRepository.LeaseReader() in D:\GitHub\nethermind\src\Nethermind\Nethermind.State\Flat\FlatDiffRepository.cs:line 116
   at Nethermind.State.Flat.FlatDiffRepository.GatherCache(StateId baseBlock, SnapshotBundleUsage usage, Nullable`1 earliestExclusive) in D:\GitHub\nethermind\src\Nethermind\Nethermind.State\Flat\FlatDiffRepository.cs:line 412
   at Nethermind.State.WorldState.BeginScope(BlockHeader baseBlock)
   at Nethermind.Consensus.Processing.BranchProcessor.Process(BlockHeader baseBlock, IReadOnlyList`1 suggestedBlocks, ProcessingOptions options, IBlockTracer blockTracer, CancellationToken token) in D:\GitHub\nethermind\src\Nethermind\Nethermind.Consensus\Processing\BranchProcessor.cs:line 53
   at Nethermind.Consensus.Processing.BlockchainProcessor.ProcessBranch(ProcessingBranch& processingBranch, ProcessingOptions options, IBlockTracer tracer, CancellationToken token, String& error)
   at Nethermind.Consensus.Processing.BlockchainProcessor.Process(Block suggestedBlock, ProcessingOptions options, IBlockTracer tracer, CancellationToken token, String& error) in D:\GitHub\nethermind\src\Nethermind\Nethermind.Consensus\Processing\BlockchainProcessor.cs:line 462
   at Nethermind.Consensus.Processing.BlockchainProcessor.ProcessBlocks() in D:\GitHub\nethermind\src\Nethermind\Nethermind.Consensus\Processing\BlockchainProcessor.cs:line 353
   at Nethermind.Consensus.Processing.BlockchainProcessor.RunProcessingLoop() in D:\GitHub\nethermind\src\Nethermind\Nethermind.Consensus\Processing\BlockchainProcessor.cs:line 332
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.AsyncStateMachineBox`1.MoveNext()
   at System.Threading.Thread.StartCallback()
 ```
 
 Seems to continue fine; but not happy if restarted

**Comment by asdacap** [comment]
Yea, dont worry about that for now. That was for debugging before. The pruning is the trie cache pruning. It uses some shards of concurrent dictionary and clear them shard by shard until the total memory is under budget.

File: src/Nethermind/Nethermind.State.Flat/TransientResource.cs [hunk 1]
```diff
@@ -0,0 +1,91 @@
...
+            BloomFilter oldFilter = PrewarmedAddresses;
+            PrewarmedAddresses = null!;
+            try
+            {
+                oldFilter.Dispose();
+            }
+            finally
+            {
+                PrewarmedAddresses = new BloomFilter(newCapacity, bitsPerKey);
+            }
```
benaadams:
```suggestion
            // Create new filter before disposing old one to avoid null ref race condition
            BloomFilter newFilter = new BloomFilter(newCapacity, bitsPerKey);
            BloomFilter oldFilter = Interlocked.Exchange(ref PrewarmedAddresses, newFilter);
            oldFilter.Dispose();
```

Also would need `using System.Threading;` in header

src/Nethermind/Nethermind.State.Flat/SnapshotBundle.cs
```diff
@@ -0,0 +1,469 @@
...
+        _transientResource.Reset();
+
+        ExpandCurrentPooledContent();
+    }
+
+    private void GuardDispose() => ObjectDisposedException.ThrowIf(_isDisposed, this);
+
+    public void Dispose()
+    {
+        if (Interlocked.CompareExchange(ref _isDisposed, true, false)) return;
```
benaadams:
```suggestion
        if (Interlocked.Exchange(ref _isDisposed, true)) return;
```

src/Nethermind/Nethermind.State.Flat/PersistenceManager.cs
```diff
@@ -0,0 +1,321 @@
...
+
+            if (snapshot.From == currentPersistedState)
+            {
+                if (_logger.IsDebug) _logger.Debug($"Persisting compacted state {stateId}");
+
+                return snapshot;
+            }
+
+            snapshot.Dispose();
+        }
```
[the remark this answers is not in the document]
benaadams: Might need to add a `Do not use Linq when a simple loop would work` to the AGENTS file

src/Nethermind/Nethermind.Core.Test/Modules/PseudoNethermindModule.cs
```diff
@@ -73,4 +102,54 @@ protected override void Load(ContainerBuilder builder)
             }
         });
     }
+
+    /// <summary>
+    /// A LOT of test rely on the fact that trie store will assume state is available as long as the state root is
+    /// empty tree even if the blocknumber is not -1. This does not work with flat. We will ignore it for now.
+    /// </summary>
+    /// <param name="flatDbManager"></param>
+    private class FlatDbManagerTestCompat(IFlatDbManager flatDbManager) : IFlatDbManager
```
LukaszRozmej: minor: I would move this out of the Module class

src/Nethermind/Nethermind.Core/Utils/RefCountingDisposable.cs
```diff
@@ -0,0 +1,139 @@
...
+        if (TryAcquireLease() == false)
+        {
+            ThrowCouldNotAcquire();
+        }
+
+        [DoesNotReturn]
+        [StackTraceHidden]
+        static void ThrowCouldNotAcquire()
+        {
+            throw new Exception("The lease cannot be acquired");
```
LukaszRozmej: throw more meaningful exception type, maybe `InvalidOperationException` or custom one

src/Nethermind/Nethermind.Core/Utils/RefCountingDisposable.cs
```diff
@@ -0,0 +1,139 @@
...
+            if (prev == Disposing)
+            {
+                // Already disposed
+                return false;
+            }
+
+            // Try again with new starting value
+            current = prev;
+            // Add PAUSE instruction to reduce shared core contention
+            Thread.SpinWait(1);
```
LukaszRozmej: switch statement?

src/Nethermind/Nethermind.Core/Utils/RefCountingDisposable.cs
```diff
@@ -0,0 +1,139 @@
...
+        }
+    }
+
+    /// <summary>
+    /// Disposes it once, decreasing the lease count by 1.
+    /// </summary>
+    public void Dispose() => ReleaseLeaseOnce();
```
LukaszRozmej, line 74: A bit weird semantics.
Normally Dispose i dispose and we are done with the object.
Maybe Aquire should return something disposable that would decrease the count?

src/Nethermind/Nethermind.Core/Utils/RefCountingDisposable.cs
```diff
@@ -0,0 +1,139 @@
...
+        {
+            // set to disposed by this Release
+            CleanUp();
+        }
+
+        [DoesNotReturn]
+        [StackTraceHidden]
+        static void ThrowOverDisposed()
+        {
+            throw new Exception("The lease has already been disposed");
```
LukaszRozmej:
```suggestion
            throw new ObjectDisposedException("The lease has already been disposed");
```

src/Nethermind/Nethermind.State.Flat/Persistence/BloomFilter/BloomFilter.cs
```diff
@@ -0,0 +1,316 @@
...
+        }
+
+        _data = (byte*)NativeMemory.AlignedAlloc(_dataSize, alignment);
+        if (_data == null) throw new OutOfMemoryException();
+
+        // Hint the kernel to use huge pages BEFORE we touch the memory (Clear).
+        // This ensures that when Clear() triggers page faults, the kernel allocates 2MB physical pages immediately.
+        if (useHugePages)
+        {
+            Madvise(_data, _dataSize, MADV_HUGEPAGE);
```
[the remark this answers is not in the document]
asdacap: No

src/Nethermind/Nethermind.Core/Threading/ReadWriteLockBox.cs
```diff
@@ -0,0 +1,52 @@
...
+using System;
+using System.Threading;
+
+namespace Nethermind.Core.Threading;
+
+/// <summary>
+/// Rust style wrapper of locked item. Make it a bit easier to know which object this lock is protecting.
+/// </summary>
+/// <typeparam name="T"></typeparam>
+public class ReadWriteLockBox<T>
```
LukaszRozmej:
```suggestion
public struct ReadWriteLockBox<T>
```

src/Nethermind/Nethermind.Db.Rocks/HyperClockCacheWrapper.cs
```diff
@@ -18,7 +18,8 @@ public HyperClockCacheWrapper(ulong capacity = 32_000_000) : base(ownsHandle: tr
 
     protected override bool ReleaseHandle()
     {
-        RocksDbSharp.Native.Instance.rocksdb_cache_destroy(handle);
+        // Temporary disable to see if it fix crash
+        // RocksDbSharp.Native.Instance.rocksdb_cache_destroy(handle);
```
LukaszRozmej: Sounds bad

src/Nethermind/Nethermind.State.Flat/Persistence/BaseFlatPersistence.cs
```diff
@@ -0,0 +1,275 @@
...
+            return true;
+        }
+
+        public ValueHash256 CurrentKey => _currentKey;
+        public ReadOnlySpan<byte> CurrentValue => _currentValue;
+
+        public void Dispose() => _view.Dispose();
+    }
+
+    public struct StorageIterator : IPersistence.IFlatIterator
```
LukaszRozmej: Does it make any sense for those iterators to be structs? I think this is not utilized at the moment.

src/Nethermind/Nethermind.State.Flat/ScopeProvider/TrieWarmer.cs
```diff
@@ -0,0 +1,333 @@
...
+                    Thread t = new Thread(() =>
+                    {
+                        RunSecondaryWorker(_cancelTokenSource.Token);
+                    });
+                    t.Name = $"TrieWarmer-Secondary-{i}";
+                    t.Priority = ThreadPriority.Lowest;
+                    t.IsBackground = true;
+                    t.Start();
+                    tasks.Add(t);
+                }
```
LukaszRozmej: Do we actually want to spawn ne threads?

src/Nethermind/Nethermind.Init/Modules/FlatWorldStateModule.cs
```diff
@@ -0,0 +1,136 @@
...
+            .AddDatabase(DbNames.Preimage)
+
+            .AddSingleton<IPersistence, IFlatDbConfig, IProcessExitSource, ILogManager, IComponentContext>((flatDbConfig, exitSource, logManager, ctx) =>
+            {
+                IPersistence persistence = flatDbConfig.Layout switch
+                {
+                    FlatLayout.Flat => ctx.Resolve<RocksDbPersistence>(),
+                    FlatLayout.FlatInTrie => ctx.Resolve<FlatInTriePersistence>(),
+                    FlatLayout.PreimageFlat => ctx.Resolve<PreimageRocksdbPersistence>(),
+                    _ => throw new Exception($"Unsupported layout {flatDbConfig.Layout}")
```
LukaszRozmej: avoid base Exception

src/Nethermind/Nethermind.State.Flat/SlotValue.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Runtime.Intrinsics;
+using Nethermind.Core.Extensions;
+
+namespace Nethermind.State.Flat;
+
+/// <summary>
+/// Make storing slot value smaller than a byte[].
+/// </summary>
+[StructLayout(LayoutKind.Sequential, Pack = 32, Size = 32)]
+public struct SlotValue
```
flcl42: Might be inline array, not a big diff though

src/Nethermind/Nethermind.State.Flat/SnapshotCompactor.cs
```diff
@@ -0,0 +1,202 @@
...
+
+        Snapshot snapshot = _resourcePool.CreateSnapshot(from, to, usage);
+        ConcurrentDictionary<AddressAsKey, Account?> accounts = snapshot.Content.Accounts;
+        ConcurrentDictionary<(AddressAsKey, UInt256), SlotValue?> storages = snapshot.Content.Storages;
+        ConcurrentDictionary<AddressAsKey, bool> selfDestructedStorageAddresses = snapshot.Content.SelfDestructedStorageAddresses;
+        ConcurrentDictionary<(Hash256AsKey, TreePath), TrieNode> storageNodes = snapshot.Content.StorageNodes;
+        ConcurrentDictionary<TreePath, TrieNode> stateNodes = snapshot.Content.StateNodes;
+
+        HashSet<Address> addressToClear = new HashSet<Address>();
+        HashSet<Hash256AsKey> addressHashToClear = new HashSet<Hash256AsKey>();
```
LukaszRozmej: Looks like big opportunity for some pooling?
We can used PooledCollections which has pooled hash set.

src/Nethermind/Nethermind.State.Flat/SnapshotCompactor.cs
```diff
@@ -0,0 +1,202 @@
...
+                foreach (KeyValuePair<(Hash256AsKey, TreePath), TrieNode> kv in storageNodes)
+                {
+                    if (addressHashToClear.Contains(kv.Key.Item1))
+                    {
+                        storageNodes.Remove(kv.Key, out _);
+                    }
+                }
+            }
+
+            storages.AddOrUpdateRange(knownState.Storages);
```
flcl42: No way knownState.Storages can contain addressToClear?

src/Nethermind/Nethermind.State.Flat/SnapshotCompactor.cs
flcl42: (If you mean https://github.com/jtmueller/Collections.Pooled, we need to add it back)

src/Nethermind/Nethermind.State.Flat/SnapshotCompactor.cs
asdacap: It can? This written storages is after the clear.

src/Nethermind/Nethermind.Core/Utils/RefCountingDisposable.cs
asdacap: Maybe dont know. This is from paprika btw.

src/Nethermind/Nethermind.Db.Rocks/HyperClockCacheWrapper.cs
asdacap: It crash on exit.

src/Nethermind/Nethermind.State.Flat/Persistence/BaseFlatPersistence.cs
asdacap: Its used in verify trie.

src/Nethermind/Nethermind.State.Flat/ScopeProvider/TrieWarmer.cs
asdacap: Maybe. Seems to work.

src/Nethermind/Nethermind.Init/Modules/FlatWorldStateModule.cs
asdacap: @copilot do it.

src/Nethermind/Nethermind.State.Flat/SnapshotBundle.cs
```diff
@@ -0,0 +1,463 @@
...
+                    storageKeysToRemove.Add(kv.Key);
+                }
+            }
+
+            foreach ((Hash256AsKey, TreePath) key in storageKeysToRemove)
+            {
+                _changedStorageNodes.TryRemove(key, out _);
+            }
+
+            ArrayPoolListRef<(AddressAsKey, UInt256)> slotKeysToRemove = new(0);
```
flcl42: Needs using

src/Nethermind/Nethermind.State.Flat/FlatDbManager.cs
```diff
@@ -0,0 +1,410 @@
...
+    }
+
+    public bool HasStateForBlock(in StateId stateId)
+    {
+        if (_snapshotRepository.HasState(stateId)) return true;
+        if (_persistenceManager.GetCurrentPersistedStateId() == stateId) return true;
+        return false;
+    }
+
+    public async ValueTask DisposeAsync()
```
flcl42, line 391: Do we need here too:
ClearReadOnlyBundleCache();
_trieNodeCache.Clear();

src/Nethermind/Nethermind.State.Flat/SnapshotBundle.cs
LukaszRozmej: also not sure if good to start at 0

src/Nethermind/Nethermind.Db.Rocks/HyperClockCacheWrapper.cs
asdacap: Actually it still crash even with this comment

src/Nethermind/Nethermind.State.Flat/FlatDbManager.cs
asdacap: fixed

src/Nethermind/Nethermind.State.Flat/StateId.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Nethermind.State.Flat;
+
+public readonly record struct StateId(long BlockNumber, in ValueHash256 StateRoot) : IComparable<StateId>
+{
+    public StateId(BlockHeader? header) : this(header?.Number ?? -1, header?.StateRoot ?? Keccak.EmptyTreeHash)
+    {
+    }
+
+    public static StateId PreGenesis = new(-1, Keccak.EmptyTreeHash);
```
benaadams:
```suggestion
    public static readonly StateId PreGenesis = new(-1, Keccak.EmptyTreeHash);
```

src/Nethermind/Nethermind.State.Flat/PersistenceManager.cs
```diff
@@ -0,0 +1,318 @@
...
+
+            foreach (KeyValuePair<(AddressAsKey, UInt256), SlotValue?> kv in snapshot.Storages)
+            {
+                ((Address addr, UInt256 slot), SlotValue? value) = kv;
+
+                batch.SetStorage(addr, slot, value);
+            }
+
+            _trieNodesSortBuffer.Clear();
+            _trieNodesSortBuffer.AddRange(snapshot.StateNodeKeys.Select<TreePath, (Hash256AsKey, TreePath)>((path) => (new Hash256AsKey(Hash256.Zero), path)));
```
benaadams:
```suggestion
            foreach (TreePath path in snapshot.StateNodeKeys)
            {
                _trieNodesSortBuffer.Add((new Hash256AsKey(Hash256.Zero), path));
            }
```

src/Nethermind/Nethermind.State.Flat/Persistence/BaseFlatPersistence.cs
```diff
@@ -0,0 +1,257 @@
...
+            ReadOnlySpan<byte> storageKey = EncodeStorageKeyHashedWithShortPrefix(stackalloc byte[StorageKeyLength], address, slot);
+
+            Span<byte> buffer = stackalloc byte[40];
+            int resultSize = GetStorageBuffer(storageKey, buffer);
+            if (resultSize == 0) return false;
+
+            Span<byte> value = buffer[..resultSize];
+
+            // AI said: Use Unsafe to bypass the 'Slice' bounds check and property access
+            // This writes the variable-length DB value into the end of the 32-byte struct
```
benaadams:
```suggestion
            // Bypass bounds check on the slice - the length is already validated by the if guard above.
            // This writes the variable-length DB value into the end of the 32-byte struct.
```

PR #9855, opened 2025-12-01
Xdc : SignTx Creating and Handling

Demuirgos:
Fixes Closes Resolves #

## Changes

- _List the changes_

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Remarks

### DIFF
--- a/src/Nethermind/Nethermind.Xdc/XdcModule.cs
+++ b/src/Nethermind/Nethermind.Xdc/XdcModule.cs
@@ -21,6 +21,11 @@ using Nethermind.Init.Modules;
 using Nethermind.Specs.ChainSpecStyle;
 using Nethermind.Xdc.Contracts;
 using Nethermind.Xdc.Spec;
+using Nethermind.TxPool;
+using Nethermind.Logging;
+using Nethermind.Evm.TransactionProcessing;
+using Nethermind.Xdc.TxPool;
+using Nethermind.Api.Steps;
 
 namespace Nethermind.Xdc;
 


src/Nethermind/Nethermind.Core/Collections/ArrayPoolList.cs
ak88, line 1: revert

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -22,7 +22,7 @@
 
 namespace Nethermind.Evm.TransactionProcessing
 {
-    public sealed class TransactionProcessor(
+    public class TransactionProcessor(
```
LukaszRozmej: It is sealed for a reason, you want `TransactionProcessorBase`

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -963,6 +963,7 @@ private TransactionResult(ErrorType error = ErrorType.None, EvmExceptionType evm
             ErrorType.SenderNotSpecified => "sender not specified",
             ErrorType.TransactionSizeOverMaxInitCodeSize => "EIP-3860 - transaction size over max init code size",
             ErrorType.WrongTransactionNonce => "wrong transaction nonce",
+            ErrorType.ContainsBlacklistedAddress => "sender or receiver are blacklisted addresses",
```
LukaszRozmej: We don't want to have anything like that in main codebase.
Suggestion: Add `int->string` dictionary for custom error codes and use custom error.
Potentially convert `Error` to an `int`

src/Nethermind/Nethermind.Xdc/XdcBlockBuildingTransactionExecutor.cs
```diff
@@ -0,0 +1,110 @@
...
+using Nethermind.State.Proofs;
+using Nethermind.TxPool.Comparison;
+using System.Collections.Generic;
+using System.Threading;
+using Nethermind.TxPool;
+using static Nethermind.Consensus.Processing.BlockProcessor;
+
+namespace Nethermind.Xdc;
+
+internal class XdcBlockBuildingTransactionExecutor(
```
LukaszRozmej: Consider a decorator instead of inheritance

src/Nethermind/Nethermind.Xdc/XdcBlockValidationTransactionExecutor.cs
```diff
@@ -0,0 +1,52 @@
...
+using System;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Metrics = Nethermind.Evm.Metrics;
+
+namespace Nethermind.Xdc;
+
+internal class XdcBlockValidationTransactionExecutor(ITransactionProcessorAdapter txProcessorAdapter, IWorldState stateProvider, ISpecProvider specProvider)
```
LukaszRozmej: Consider a decorator instead of inheritance

src/Nethermind/Nethermind.Consensus/Producers/BlockToProduce.cs
```diff
@@ -9,6 +9,7 @@
 
 //TODO: Redo clique block producer
 [assembly: InternalsVisibleTo("Nethermind.Consensus.Clique")]
+[assembly: InternalsVisibleTo("Nethermind.Xdc")]
```
LukaszRozmej: Why needed?

flcl42 (approved):

src/Nethermind/Nethermind.Xdc.Test/SpecialTransactionsTests.cs
```diff
@@ -0,0 +1,470 @@
...
+            spec.IsEip1559Enabled = false;
+        });
+
+        blockChain.StartHotStuffModule();
+
+        XdcBlockHeader? head = blockChain.BlockTree.Head!.Header as XdcBlockHeader;
+        do
+        {
+            await blockChain.TriggerAndSimulateBlockProposalAndVoting();
+            await Task.Delay(blockChain.SpecProvider.GetXdcSpec(head!).MinePeriod.Seconds()); // to avoid tight loop
```
ak88: TriggerAndSimulateBlockProposalAndVoting() will do all the waiting and timeout for you. I dont see why you need to wait extra here.

src/Nethermind/Nethermind.Xdc.Test/SpecialTransactionsTests.cs
```diff
@@ -0,0 +1,470 @@
...
+            head = (XdcBlockHeader)blockChain.BlockTree.Head!.Header;
+        }
+        while (!IsTimeForOnchainSignature(blockChain.SpecProvider.GetXdcSpec(head), head.Number - 1));
+
+
+
+        Assert.That(blockChain.BlockTree.Head.Number, Is.EqualTo(mergeSignBlockRange + 1));
+
+        await Task.Delay(((XdcReleaseSpec)blockChain.SpecProvider.GetFinalSpec()).MinePeriod.Seconds()); // to avoid tight loop
```
ak88: Again why do you wait here?

src/Nethermind/Nethermind.Xdc/SignTransactionManager.cs
```diff
@@ -0,0 +1,201 @@
...
+using System.Linq;
+using System.Security.Cryptography;
+using System.Text;
+using System.Threading.Tasks;
+using static Microsoft.FSharp.Core.ByRefKinds;
+using static System.Net.Mime.MediaTypeNames;
+
+namespace Nethermind.Xdc;
+
+internal class SignTransactionManager(IDb stateDb, ISigner signer, ITxPool txPool) : ISignTransactionManager
```
ak88: Change the name from stateDb. It doesnt have anything to do with state db.

src/Nethermind/Nethermind.Xdc/SignTransactionManager.cs
```diff
@@ -0,0 +1,201 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Security.Cryptography;
+using System.Text;
+using System.Threading.Tasks;
+using static Microsoft.FSharp.Core.ByRefKinds;
+using static System.Net.Mime.MediaTypeNames;
```
ak88: remove

src/Nethermind/Nethermind.Xdc/SignTransactionManager.cs
```diff
@@ -0,0 +1,201 @@
...
+        UInt256 nonce = txPool.GetLatestPendingNonce(signer.Address);
+        Transaction transaction = CreateTxSign((UInt256)header.Number, header.Hash, nonce, spec.BlockSignersAddress, signer.Address);
+
+        await signer.Sign(transaction);
+
+        // add local somehow to tx pool
+        bool added = txPool.SubmitTx(transaction, TxHandlingOptions.PersistentBroadcast);
+        if (!added)
+        {
+            throw new Exception("Failed to add signed transaction to the pool.");
```
ak88: Never throw just Exception()

src/Nethermind/Nethermind.Xdc/SignTransactionManager.cs
```diff
@@ -0,0 +1,201 @@
...
+        }
+
+        long blockNumber = header.Number;
+        long checkNumber = blockNumber % spec.EpochLength;
+
+        PrivateKeyGenerator privateKeyGenerator = new PrivateKeyGenerator();
```
ak88: no need for this to be allocated on every call.

src/Nethermind/Nethermind.Xdc/SignTransactionManager.cs
```diff
@@ -0,0 +1,201 @@
...
+        PrivateKey randomPrivate = privateKeyGenerator.Generate();
+
+        byte[] randomKey = Encoding.UTF8.GetBytes("randomizeKey");
```
ak88: make it a static readonly

src/Nethermind/Nethermind.Xdc/SignTransactionManager.cs
```diff
@@ -0,0 +1,201 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using static Microsoft.FSharp.Core.ByRefKinds;
+using static System.Net.Mime.MediaTypeNames;
+
+namespace Nethermind.Xdc;
+
+internal class SignTransactionManager(IDb stateDb, ISigner signer, ITxPool txPool) : ISignTransactionManager
+{
+    public async Task CreateTransactionSign(XdcBlockHeader header, IXdcReleaseSpec spec)
```
ak88: Better name would be something like SubmitTransactionSign

src/Nethermind/Nethermind.Xdc/SignTransactionManager.cs
```diff
@@ -0,0 +1,201 @@
...
+        PrivateKeyGenerator privateKeyGenerator = new PrivateKeyGenerator();
+        PrivateKey randomPrivate = privateKeyGenerator.Generate();
+
+        byte[] randomKey = Encoding.UTF8.GetBytes("randomizeKey");
+
+        var exists = stateDb.KeyExists(randomKey);
+
+        if (exists)
+        {
+            if (checkNumber > 0 && spec.EpochBlockOpening <= checkNumber && spec.EpochBlockRandomize >= checkNumber)
```
ak88: move this check out above the KeyExists check and make it return on false to reduce nesting, and then remove the duplicate check below

src/Nethermind/Nethermind.Xdc/SignTransactionManager.cs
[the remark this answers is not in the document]
Demuirgos: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/NethermindEth/nethermind/pull/9855#discussion_r2657284703)

src/Nethermind/Nethermind.Xdc/XdcTransactionExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    public static bool IsSpecialTransaction(this Transaction currentTx, IXdcReleaseSpec spec)
+        => currentTx.To is not null && ((currentTx.To == spec.BlockSignerContract) || (currentTx.To == spec.RandomizeSMCBinary));
+    public static bool RequiresSpecialHandling(this Transaction currentTx, IXdcReleaseSpec spec)
+            => IsSignTransaction(currentTx, spec)
+            || IsLendingTransaction(currentTx, spec)
+            || IsTradingTransaction(currentTx, spec)
+            || IsLendingFinalizedTradeTransaction(currentTx, spec)
+            || IsTradingStateTransaction(currentTx, spec);
+    public static bool IsSignTransaction(this Transaction currentTx, IXdcReleaseSpec spec) => currentTx.To is not null && currentTx.To == spec.BlockSignerContract;
```
cicr99, line 22: Is this enough to consider it a sign transaction? There is also `getSigners` function in the same contract. In the rewardCalculator for example there is a method for this that also checks the data length is correct and the selector is the one corresponding to `sign` function

src/Nethermind/Nethermind.Xdc/XdcTransactionProcessor.cs
```diff
@@ -0,0 +1,199 @@
...
+                    return XdcTransactionResult.NonceTooHigh;
+                }
+                else if (nonce > tx.Nonce)
+                {
+                    return XdcTransactionResult.NonceTooLow;
+                }
+
+                WorldState.IncrementNonce(tx.SenderAddress);
+
+                return TransactionResult.Ok;
```
cicr99: you could remove this line if you wanted, the statement below is the same and would get executed

src/Nethermind/Nethermind.Xdc/XdcTransactionProcessor.cs
```diff
@@ -0,0 +1,199 @@
...
+                WorldState.RecalculateStateRoot();
+                stateRoot = WorldState.StateRoot;
+            }
+
+            var log = new LogEntry(tx.To, [], []);
+            tracer.MarkAsSuccess(tx.To, 0, [], [log], stateRoot);
+        }
+
+        return TransactionResult.Ok;
+    }
```
cicr99: Is this the same as ProcessEmptyTransaction?

src/Nethermind/Nethermind.Xdc/Spec/XdcReleaseSpec.cs
```diff
@@ -32,9 +32,21 @@ public class XdcReleaseSpec : ReleaseSpec, IXdcReleaseSpec
...
+    public long MergeSignRange { get; set; }
+    public long TIP2019Block { get; set; }
+    public Address[] BlackListedAddresses { get; set; }
+    public long EpochBlockSecret { get; set; }
+    public Address BlockSignerContract { get; set; }
+    public Address RandomizeSMCBinary { get; set; }
+    public Address XDCXLendingFinalizedTradeAddressBinary { get; set; }
+    public Address XDCXLendingAddressBinary { get; set; }
+    public Address XDCXAddressBinary { get; set; }
+    public Address TradingStateAddressBinary { get; set; }
```
cicr99: For these values to be set in the release spec:
- they need to be added in the configuration file `xdc.json` in the engine parameters section
- they have to be assigned through `CreateReleaseSpec` method in `XdcChainSpecBasedSpecProvider`

src/Nethermind/Nethermind.Xdc/Spec/XdcReleaseSpec.cs
```diff
@@ -32,9 +32,21 @@ public class XdcReleaseSpec : ReleaseSpec, IXdcReleaseSpec
     public List<V2ConfigParams> V2Configs { get; set; } = new List<V2ConfigParams>();
 
     public Address[] GenesisMasterNodes { get; set; }
+    public long BlackListHFNumber { get; set; }
+    public long EpochBlockOpening { get; set; }
+    public long EpochBlockRandomize { get; set; }
+    public long MergeSignRange { get; set; }
```
cicr99, line 35: There was a variable for this in the reward calculator, but since it is used by other components, it make sense that it goes here. It should be changed in reward calculator to use this one from the spec and remove the variable. The changes should be quite small, since it is only used in GetSigningTxCount method, which already receives the spec as an argument; and probably the tests should also be updated so the release spec returns this value (I see it is already added in WrapReleaseSpec, so only the unit test actually needs updating then)

src/Nethermind/Nethermind.Xdc/SignTransactionManager.cs
```diff
@@ -0,0 +1,194 @@
...
+internal class SignTransactionManager(IDb stateDb, ISigner signer, ITxPool txPool) : ISignTransactionManager
+{
+    public async Task CreateTransactionSign(XdcBlockHeader header, IXdcReleaseSpec spec)
+    {
+        UInt256 nonce = txPool.GetLatestPendingNonce(signer.Address);
+        Transaction transaction = CreateTxSign((UInt256)header.Number, header.Hash, nonce, spec.BlockSignerContract, signer.Address);
+
+        await signer.Sign(transaction);
+
+        // add local somehow to tx pool
```
cicr99: This comment seems outdated / misleading. Is there something I'm missing here?

src/Nethermind/Nethermind.Xdc/SignTransactionManager.cs
```diff
@@ -0,0 +1,194 @@
...
+        if (!added)
+        {
+            throw new InvalidOperationException("Failed to add signed transaction to the pool.");
+        }
+
+        long blockNumber = header.Number;
+        long checkNumber = blockNumber % spec.EpochLength;
+
+        PrivateKeyGenerator privateKeyGenerator = new PrivateKeyGenerator();
+        PrivateKey randomPrivate = privateKeyGenerator.Generate();
```
cicr99: This doesn't seem to be used

src/Nethermind/Nethermind.Xdc/SignTransactionManager.cs
```diff
@@ -0,0 +1,194 @@
...
+
+        if (exists)
+        {
+            if (checkNumber > 0 && spec.EpochBlockOpening <= checkNumber && spec.EpochBlockRandomize >= checkNumber)
+            {
+                var randomizeKeyValue = stateDb.Get(randomKey);
+                Transaction tx = CreateTxOpeningRandomize(nonce + 1, spec.RandomizeSMCBinary, randomizeKeyValue, signer.Address);
+                await signer.Sign(tx);
+
+                // add local somehow to tx pool
```
cicr99: same comment as above :)

src/Nethermind/Nethermind.Xdc/SignTransactionManager.cs
```diff
@@ -0,0 +1,194 @@
...
+            }
+        }
+        else
+        {
+            var randomizeKeyValue = RandStringByte(32);
+            if (checkNumber > 0 && spec.EpochBlockSecret <= checkNumber && spec.EpochBlockOpening > checkNumber)
+            {
+                Transaction tx = BuildTxSecretRandomize(nonce + 1, spec.RandomizeSMCBinary, (ulong)spec.EpochLength, randomizeKeyValue, signer.Address);
+                await signer.Sign(tx);
+                // add local somehow to tx pool
```
cicr99: same comment as above :)

src/Nethermind/Nethermind.Xdc.Test/Helpers/XdcTestBlockchain.cs
```diff
@@ -333,6 +352,25 @@ public Block Build()
...
+            var dummyCodeHashcode = Keccak.Compute(dummyCode);
+
+            state.InsertCode(finalSpec.BlockSignerContract, dummyCodeHashcode, dummyCode, genesisSpec!, true);
+            state.InsertCode(finalSpec.RandomizeSMCBinary, dummyCodeHashcode, dummyCode, genesisSpec!, true);
+
+
+            foreach (var nodeAddress in finalSpec.GenesisMasterNodes)
+            {
+                state.CreateAccount(nodeAddress, testConfiguration.AccountInitialValue);
+            }
```
cicr99: Wasn't this done already a few lines above?

src/Nethermind/Nethermind.Xdc/XdcTransactionExtensions.cs
cicr99: Already discussed this, in xdc repo, they don't appear to check anything else either

src/Nethermind/Nethermind.Xdc/XdcHotStuff.cs
```diff
@@ -241,12 +245,18 @@ internal async Task RunRoundChecks(CancellationToken ct)
             {
                 _highestSelfMinedRound = currentRound;
                 Task blockBuilder = BuildAndProposeBlock(roundParent, currentRound, spec, ct);
+
+                if ((roundParent.Number % spec.MergeSignRange == 0) || roundParent.Number < spec.TIP2019Block)
+                {
+                    await _signTransactionManager.CreateTransactionSign(roundParent, spec);
+                }
```
ak88: You need a check on if the node is a masternode - if not then dont do the tx sign.
And this has nothing to do with if the node is currently proposing or not.

ak88:
I dont think that randomize contract is used at all in V2 and so should be removed.

src/Nethermind/Nethermind.Xdc/XdcHotStuff.cs
Demuirgos: from what i understood SignTxs are submitted by the node that proposes the block, and they are submitted to the local txPool kinda doesnt make sense that every masternode submits a local tx right?

src/Nethermind/Nethermind.Xdc/XdcHotStuff.cs
```diff
@@ -241,12 +245,20 @@ internal async Task RunRoundChecks(CancellationToken ct)
             {
                 _highestSelfMinedRound = currentRound;
                 Task blockBuilder = BuildAndProposeBlock(roundParent, currentRound, spec, ct);
+
+            }
+
+            if (Array.Exists(epochInfo.Masternodes, (masternodeAddress) => _signer.Address == masternodeAddress)
+                && ((roundParent.Number % spec.MergeSignRange == 0) || roundParent.Number < spec.TIP2019Block))
```
cicr99: You should be able to remove the last part in the or statement: `roundParent.Number < spec.TIP2019Block`. From what I see in their code this variable is set to `tip2019Block: big.NewInt(1)` in mainnet, and 0 or 1 in the other cases, so this should always be false in our case, right?

src/Nethermind/Nethermind.Xdc/SignTransactionManager.cs
```diff
@@ -0,0 +1,60 @@
...
+    {
+        UInt256 nonce = txPool.GetLatestPendingNonce(signer.Address);
+        Transaction transaction = CreateTxSign((UInt256)header.Number, header.Hash, nonce, spec.BlockSignerContract, signer.Address);
+
+        await signer.Sign(transaction);
+
+        bool added = txPool.SubmitTx(transaction, TxHandlingOptions.PersistentBroadcast);
+        if (!added)
+        {
+            throw new SignTransactionPoolException("Failed to add signed transaction to the pool.");
```
ak88: I dont think this should throw. Just write an error to log.

src/Nethermind/Nethermind.Xdc/XdcTransactionProcessor.cs
```diff
@@ -0,0 +1,169 @@
...
+        ILogManager? logManager) : TransactionProcessorBase<EthereumGasPolicy>(blobBaseFeeCalculator, specProvider, worldState, virtualMachine, codeInfoRepository, logManager)
+{
+
+    protected override TransactionResult ValidateSender(Transaction tx, BlockHeader header, IReleaseSpec spec, ITxTracer tracer, ExecutionOptions opts)
+    {
+        var xdcSpec = spec as XdcReleaseSpec;
+        Address target = tx.To;
+        Address sender = tx.SenderAddress;
+
+        if (xdcSpec.BlackListHFNumber <= header.Number)
```
ak88:
```suggestion
        if (xdcSpec.BlackListHFNumber >= header.Number)
```

src/Nethermind/Nethermind.Xdc/XdcTransactionProcessor.cs
```diff
@@ -0,0 +1,169 @@
...
+                return TransactionResult.MalformedTransaction;
+            }
+
+            return TransactionResult.Ok;
+        }
+
+        return base.ValidateStatic(tx, header, spec, opts, intrinsicGas);
+    }
+
+    private bool IsBlackListed(IXdcReleaseSpec spec, Address sender)
```
ak88, line 68: check using a hashset instead

src/Nethermind/Nethermind.Xdc/XdcTransactionProcessor.cs
```diff
@@ -0,0 +1,169 @@
...
+            {
+                return TransactionResult.MalformedTransaction;
+            }
+
+            UInt256 blkNumber = new UInt256(tx.Data.Span.Slice(4, 32), true);
+            if (blkNumber >= header.Number || blkNumber <= (header.Number - (xdcSpec.EpochLength * 2)))
+            {
+                // Invalid block number in special transaction data
+                return TransactionResult.MalformedTransaction;
+            }
```
ak88: You are mixing up block production and block processing. These checks belong in block production, or better  as part of a TxPool filter.

Demuirgos: so when we get a block that includes these transactions we wont validate their form?

src/Nethermind/Nethermind.Xdc/Spec/XdcChainSpecEngineParameters.cs
```diff
@@ -37,6 +43,13 @@ public List<V2ConfigParams> V2Configs
         }
     }
 
+    public long TIP2019Block { get; internal set; }
+    public long MergeSignRange { get; internal set; }
+    public Address[] BlackListedAddresses { get; internal set; }
+    public long BlackListHFNumber { get; internal set; }
+    public long TipXDCX { get; internal set; }
+    public long TIPXDCXMinerDisable { get; internal set; }
+
```
ak88: these cannot have internal setters

src/Nethermind/Nethermind.Xdc.Test/SpecialTransactionsTests.cs
```diff
@@ -732,4 +728,185 @@ public async Task Valid_SpecialTx_NotSign_Call_EmptyTx_Handler()
...
+        receipts.Any(r => r.Recipient == spec.BlockSignerContract).Should().BeTrue();
+    }
+
+    [Test]
+    public async Task RandomizeTx_IncrementNonce_And_Is_Treated_As_Free()
+    {
+        var blockChain = await XdcTestBlockchain.Create(5, false);
+        blockChain.ChangeReleaseSpec((spec) =>
+        {
+            spec.IsEip1559Enabled = false;
```
ak88: Why not test with both condition? We have to support 1559.

src/Nethermind/Nethermind.Xdc/XdcTransactionProcessor.cs
```diff
@@ -0,0 +1,197 @@
...
+
+namespace Nethermind.Xdc;
+
+internal class XdcTransactionProcessor(
+        ITransactionProcessor.IBlobBaseFeeCalculator blobBaseFeeCalculator,
+        ISpecProvider? specProvider,
+        IWorldState? worldState,
+        IVirtualMachine? virtualMachine,
+        ICodeInfoRepository? codeInfoRepository,
+        ILogManager? logManager) : TransactionProcessorBase<EthereumGasPolicy>(blobBaseFeeCalculator, specProvider, worldState, virtualMachine, codeInfoRepository, logManager)
```
cicr99, line 26: Tiny nit: We could use `EthereumTransactionProcessorBase` instead of `TransactionProcessorBase<EthereumGasPolicy>`, though it doesn't change behavior

PR #9225, opened 2025-09-01
Update tests to fix missing test jsons

flcl42:
With the current commit of src/tests submodule some json are missing for some reason. Let's update the tests and see if they are failing also

## Changes

- Update src/tests

#### What types of changes does your code introduce?

- [ ] Bugfix (a non-breaking change that fixes an issue)
- [ ] New feature (a non-breaking change that adds functionality)
- [ ] Breaking change (a change that causes existing functionality not to work as expected)
- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

#### If yes, did you write tests?

- [ ] Yes
- [x] No

PR #9865, opened 2025-12-02
fix: only touch coinbase after successful transaction in state tests

bshastry:
When a transaction fails validation (e.g., insufficient balance), the state test runner should not modify state at all. Previously, the coinbase account was created in InitializeTestState before transaction execution, causing state root divergence from geth when transactions failed validation.

This fix moves coinbase creation to after successful transaction execution, matching geth's behavior:
- Only touch coinbase when txResult == Ok
- When validation fails, state remains unchanged from pre-state

This was a consensus bug in the test framework (not production code) introduced in PR #9225.

# fix: only touch coinbase after successful transaction in state tests

## Changes

- Move coinbase account creation inside the successful transaction block in `GeneralTestBase.cs`
- Remove premature coinbase creation from `InitializeTestState` method
- Add comment explaining why coinbase touch must happen after successful tx execution
- Ensure state root matches geth reference implementation for tests with invalid transactions

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

Manually verified with EEST-generated state test `CALL_Bounds2a[fork_Cancun-state_test--g0]` that previously produced divergent state roots:

- **Before fix**: Nethermind state root `0x823cc91f768e8afb9e1d4cfcdd1bf85e35781fbbb2122020783586da1f42ca30`
- **After fix**: Nethermind state root `0xd919d764bbc4f34ed6b03dbe84b9ef094a75146c08390a209b349c4c9156f629` (matches geth)

The test involves a transaction that fails validation due to insufficient sender balance (sender has 0 wei, needs 1,500,001 wei for gas * price + value). Previously, Nethermind was creating the coinbase account before transaction execution, causing the state root to differ from geth which only touches coinbase after successful execution.

**Note**: This bug only affects the test framework (`Ethereum.Test.Base`), not production code. The production `TransactionProcessor` correctly only touches coinbase in `PayFees` after `BuyGas` validation passes.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks

This bug was introduced in commit `f48e3d8e6b4` (PR #9225) which moved coinbase creation before transaction execution. The original code had a comment from @winsvega about adding a 0-wei reward for retesteth compatibility, but this should only happen after successful transaction execution, not when the transaction fails validation.

The same bug was independently discovered in Besu's state test runner (fixed in parallel). Both bugs are isolated to test frameworks only - production consensus code is not affected.

### DIFF
--- a/src/Nethermind/Ethereum.Test.Base/GeneralTestBase.cs
+++ b/src/Nethermind/Ethereum.Test.Base/GeneralTestBase.cs
@@ -180,6 +180,7 @@ namespace Ethereum.Test.Base
                 // For legacy tests with failed tx, we need to recalculate root since coinbase was created
                 if (test.IsLegacy)
                 {
+                    stateProvider.CommitTree(0);
                     stateProvider.RecalculateStateRoot();
                 }
                 else


src/Nethermind/Ethereum.Test.Base/GeneralTestBase.cs
```diff
@@ -154,6 +154,15 @@ protected EthereumTestResult RunTest(GeneralStateTest test, ITxTracer txTracer)
             {
                 stateProvider.Commit(specProvider.GetSpec((ForkActivation)1));
                 stateProvider.CommitTree(1);
+
+                // '@winsvega added a 0-wei reward to the miner, so we had to add that into the state test execution phase. He needed it for retesteth.'
+                // This must only happen after successful transaction execution, not when tx fails validation.
+                if (!stateProvider.AccountExists(test.CurrentCoinbase))
```
deffrian: We have stateProveider.CreatIfNotExists

src/Nethermind/Ethereum.Test.Base/GeneralTestBase.cs
```diff
@@ -154,6 +154,15 @@ protected EthereumTestResult RunTest(GeneralStateTest test, ITxTracer txTracer)
...
+                {
+                    stateProvider.CreateAccount(test.CurrentCoinbase, 0);
+                }
```
LukaszRozmej:
```suggestion
                stateProvider.CreateAccountIfNotExists(test.CurrentCoinbase, UInt256.Zero);
```

src/Nethermind/Ethereum.Test.Base/GeneralTestBase.cs
```diff
@@ -154,11 +164,28 @@ protected EthereumTestResult RunTest(GeneralStateTest test, ITxTracer txTracer)
...
+                stateProvider.Commit(specProvider.GetSpec((ForkActivation)1));
                 stateProvider.RecalculateStateRoot();
             }
             else
             {
-                stateProvider.Reset();
+                // For legacy tests with failed tx, we need to recalculate root since coinbase was created
+                if (test.IsLegacy)
+                {
+                    stateProvider.RecalculateStateRoot();
```
flcl42, line 184:
```suggestion
                    stateProvider.CommitTree(0);
                    stateProvider.RecalculateStateRoot();
```

PR #9867, opened 2025-12-02
refactor(blockchain): remove unused LinearExtrapolation overload

Snezhkko:
Remove the private 4-parameter constructor (long, DateTime, long, DateTime) from ChainSizes.LinearExtrapolation in 
Nethermind.Blockchain/KnownChainSizes.cs. The overload is unreferenced across the codebase and the class is private, so it cannot be used externally. Only the 3-parameter overload is used by CreateChainSizeInfo(). This reduces maintenance overhead and avoids confusion.


LukaszRozmej:
leave it for now

PR #9868, opened 2025-12-02
fix: include IsAuthenticated and modules in JsonRpcUrl.GetHashCode

sashass1315:
JsonRpcUrl.Equals compares IsAuthenticated and EnabledModules (case-insensitive set), but GetHashCode previously ignored both by casting EnabledModules to IStructuralEquatable (which HashSet<string> doesn’t implement) and not including IsAuthenticated. This could break hash-based collections when equal instances produce different hashes. The fix adds IsAuthenticated to the hash and computes a stable, case-/order-insensitive hash for EnabledModules to preserve the equals/hash code contract.

### DIFF
--- a/src/Nethermind/Nethermind.JsonRpc/JsonRpcUrl.cs
+++ b/src/Nethermind/Nethermind.JsonRpc/JsonRpcUrl.cs
@@ -14,18 +14,17 @@ namespace Nethermind.JsonRpc
     public class JsonRpcUrl : IEquatable<JsonRpcUrl>, ICloneable
     {
         private int? _hashCode;
-        private string _scheme = null!;
-        private string _host = null!;
-        private int _port;
-        private RpcEndpoint _rpcEndpoint;
-        private IReadOnlySet<string> _enabledModules = null!;
         public JsonRpcUrl(string scheme, string host, int port, RpcEndpoint rpcEndpoint, bool isAuthenticated, string[] enabledModules, long? maxRequestBodySize = null)
         {
+            string[] sortedModules = enabledModules
+                .OrderBy(x => x, StringComparer.OrdinalIgnoreCase)
+                .ToArray();
+
             Scheme = scheme;
             Host = host;
             Port = port;
             RpcEndpoint = rpcEndpoint;
-            EnabledModules = new HashSet<string>(enabledModules, StringComparer.OrdinalIgnoreCase);
+            EnabledModules = new HashSet<string>(sortedModules, StringComparer.OrdinalIgnoreCase);
             IsAuthenticated = isAuthenticated;
             MaxRequestBodySize = maxRequestBodySize;
         }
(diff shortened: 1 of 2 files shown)


benaadams (commented):
Make it quite expensive? Should probably cache it to a nullable `int?`

(Initialize if null)

src/Nethermind/Nethermind.JsonRpc/JsonRpcUrl.cs
```diff
@@ -113,10 +118,21 @@ public override bool Equals(object other)
...
+            foreach (string m in EnabledModules.OrderBy(x => x, StringComparer.OrdinalIgnoreCase))
+            {
+                modulesHash = HashCode.Combine(modulesHash, StringComparer.OrdinalIgnoreCase.GetHashCode(m));
+            }
 
-        public override int GetHashCode() => HashCode.Combine(Scheme, Host, Port, RpcEndpoint, EnabledModules as IStructuralEquatable);
+            int computed = HashCode.Combine(Scheme, Host, Port, RpcEndpoint, IsAuthenticated, modulesHash);
+            _hashCode = computed;
+            return computed;
+        }
```
LukaszRozmej:
```suggestion
            return _hashCode = HashCode.Combine(Scheme, Host, Port, RpcEndpoint, IsAuthenticated, modulesHash);
        }
```

LukaszRozmej (approved):

src/Nethermind/Nethermind.JsonRpc/JsonRpcUrl.cs
sashass1315: > 

This didn't compile

src/Nethermind/Nethermind.JsonRpc/JsonRpcUrl.cs
LukaszRozmej: We could sort `enabledModules` before adding them to the `HashSet` -> thus having order guaranteed (`HashSet` keeps the order).

Alternatively we could use SortedSet

src/Nethermind/Nethermind.JsonRpc/JsonRpcUrl.cs
```diff
@@ -78,11 +84,11 @@ public static JsonRpcUrl Parse(string packedUrlValue)
...
-        public string Scheme { get; set; }
-        public string Host { get; set; }
-        public int Port { get; set; }
-        public RpcEndpoint RpcEndpoint { get; set; }
-        public IReadOnlySet<string> EnabledModules { get; set; }
+        public string Scheme { get => _scheme; set { _scheme = value; _hashCode = null; } }
+        public string Host { get => _host; set { _host = value; _hashCode = null; } }
+        public int Port { get => _port; set { _port = value; _hashCode = null; } }
+        public RpcEndpoint RpcEndpoint { get => _rpcEndpoint; set { _rpcEndpoint = value; _hashCode = null; } }
+        public IReadOnlySet<string> EnabledModules { get => _enabledModules; set { _enabledModules = new HashSet<string>(value, StringComparer.OrdinalIgnoreCase); _hashCode = null; } }
```
LukaszRozmej: We probably can avoid having those set'ers. And make the whole class readonly?
This needs just a bit of refactoring.

LukaszRozmej (changes requested):
I thought about it a bit and:
-

stdevMac:
@LukaszRozmej ping on the March question ('is this still valid or can we close it?') — the author did respond to review back then, so it just needs your call: rebase-and-merge or close? If there's no reply in the next few days, we'll close this as part of the stale-PR cleanup — it can always be reopened or resubmitted.

PR #9870, opened 2025-12-03
fix: remove redundant BindingFlags.GetProperty from GetProperties

radik878:
Type.GetProperties(BindingFlags) ignores BindingFlags.GetProperty, which is only meaningful for InvokeMember. Dropped it to avoid confusion and reflect the intended filter set.


LukaszRozmej:
doesn't really change anything

PR #9872, opened 2025-12-03
Improve SendBlobs tool

flcl42:
## Changes

- Reduce code repetition
- Add seed option to get the same blobs on every run if needed
- Make Osaka fork default fork, which affects blob network wrapper format

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/tools/SendBlobs/BlobSender.cs
+++ b/tools/SendBlobs/BlobSender.cs
@@ -83,8 +83,6 @@ internal class BlobSender
 
         int signerIndex = -1;
 
-        ulong excessBlobs = (ulong)blobTxCounts.Sum(bc => bc.blobCount) / 2;
-
         foreach ((int txCount, int blobCount, string @break) txs in blobTxCounts)
         {
             int txCount = txs.txCount;
@@ -143,6 +141,7 @@ internal class BlobSender
                     return;
                 }
 
+                ulong excessBlobs = (ulong)blobCount / 2;
                 (UInt256 maxGasPrice, UInt256 maxPriorityFeePerGas, UInt256 maxFeePerBlobGas) = await GetGasPrices(null, maxPriorityFeeGasArgs, maxFeePerBlobGasArgs, blockResult, excessBlobs, spec);
 
                 maxPriorityFeePerGas *= feeMultiplier;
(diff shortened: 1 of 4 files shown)


tools/SendBlobs/RpcHelper.cs
```diff
@@ -0,0 +1,52 @@
...
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using Nethermind.Core;
+using Nethermind.Core.Extensions;
+using Nethermind.Int256;
+using Nethermind.JsonRpc.Client;
+
+namespace SendBlobs;
+
+internal static class RpcHelper
```
alexb5dh: Won't these look better as extension methods, given that it's for this project only?

tools/SendBlobs/RpcHelper.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        string? nonceString = await rpcClient.Post<string>("eth_getTransactionCount", address, "latest");
+        return nonceString is null ? null : Bytes.FromHexString(nonceString).AsSpan().ReadEthUInt64();
+    }
+
+    public static async Task<UInt256> GetGasPriceAsync(IJsonRpcClient rpcClient)
+    {
+        ArgumentNullException.ThrowIfNull(rpcClient);
+
+        string gasPriceRes = await rpcClient.Post<string>("eth_gasPrice") ?? "0x1";
```
alexb5dh: Falling back to defaut value here and below seems a bit inconsistent with `GetTransactionCountAsync`, which returns `null` instead (and causes caller to throw an exception).

tools/SendBlobs/RpcHelper.cs
flcl42: Good idea!

tools/SendBlobs/RpcHelper.cs
flcl42: Let's stop when server does not respond or responds with unexpected null

PR #9881, opened 2025-12-04
XDC Reward handler

cicr99:


src/Nethermind/Nethermind.Xdc.Test/Helpers/XdcTestBlockchain.cs
```diff
@@ -335,6 +355,11 @@ public Block Build()
         }
     }
 
+    private class NullRewardCalculator : IRewardCalculator
```
ak88:
```suggestion
    private class ZeroRewardCalculator : IRewardCalculator
```

src/Nethermind/Nethermind.Xdc/XdcRewardCalculator.cs
```diff
@@ -0,0 +1,243 @@
...
+        {
+            if (block is null)
+                throw new ArgumentNullException(nameof(block));
+            if (block.Header is not XdcBlockHeader xdcHeader)
+                throw new InvalidOperationException("Only supports XDC headers");
+
+            var rewards = new List<BlockReward>();
+            var number = xdcHeader.Number;
+            IXdcReleaseSpec spec = specProvider.GetXdcSpec(xdcHeader, xdcHeader.ExtraConsensusData.BlockRound);
+            if (number == spec.SwitchBlock + 1) return rewards.ToArray();
```
ak88: Return Array.Empty<BlockReward>().

src/Nethermind/Nethermind.Xdc/XdcRewardCalculator.cs
```diff
@@ -0,0 +1,243 @@
...
+
+        private (Dictionary<Address, long> Signers, long Count) GetSigningTxCount(long number, XdcBlockHeader header, IXdcReleaseSpec spec)
+        {
+            long signEpochCount = 1, rewardEpochCount = 2, epochCount = 0, endBlockNumber = 0, startBlockNumber = 0, signingCount = 0;
+            var signers = new Dictionary<Address, long>();
+            var blockNumberToHash = new Dictionary<long, Hash256>();
+            var hashToSigningAddress = new Dictionary<Hash256, HashSet<Address>>();
+            var masternodes = new HashSet<Address>();
+
+            if (number == 0) return (signers, signingCount);
```
ak88: move to top

src/Nethermind/Nethermind.Xdc/XdcRewardCalculator.cs
```diff
@@ -0,0 +1,243 @@
...
+        {
+            Transaction[] signingTxs = txs.Where(t => IsSigningTransaction(t, spec)).ToArray();
+            _signingTxsCache.Set(hash, signingTxs);
+            return signingTxs;
+        }
+
+        // Signing transaction ABI (Solidity):
+        // function sign(uint256 _blockNumber, bytes32 _blockHash)
+        // Calldata = 4-byte selector + 32-byte big-endian uint + 32-byte bytes32 = 68 bytes total.
+        private bool IsSigningTransaction(Transaction tx, IXdcReleaseSpec spec)
```
ak88, line 163: Would imagine this would be used other places as well. Just to keep in mind @Demuirgos .

src/Nethermind/Nethermind.Xdc/XdcRewardCalculator.cs
```diff
@@ -0,0 +1,243 @@
...
+        {
+            Address owner = masternodeVotingContract.GetCandidateOwner(header, masternodeAddress);
+
+            // 90% of the reward goes to the masternode
+            UInt256 masterReward = reward * 90 / 100;
+
+            // 10% of the reward goes to the foundation wallet
+            UInt256 foundationReward = reward / 10;
+
+            return (new BlockReward(owner, masterReward), foundationReward);
```
ak88, line 244: we need to make sure the rounding acts 100% the same.
Does XDC use any floating point numbers in their calculation?

src/Nethermind/Nethermind.Xdc/XdcHotStuff.cs
ak88, line 1: This is not correct. Block rewards are done in BlockProcessor through the interface. 
There should be no changes in this file.

ak88 (changes requested):
As is the rewards are not actually applied. You need to inject the RewardCalculator in XdcModule 

Think we also could use some unittests with odd reward amounts, that validates against rewards pre calculated using XDC  repo.

src/Nethermind/Nethermind.Xdc/XdcRewardCalculator.cs
cicr99: No, they are using Big.Int for all the calculation. Added a unit test that checks this is being done the same way as them

src/Nethermind/Nethermind.Xdc/XdcHotStuff.cs
cicr99: Removed this logic from XdcHotStuff and added the check for the block being an epoch switch at the beginning of XdcRewardCalculator

src/Nethermind/Nethermind.Xdc.Test/ModuleTests/RewardTests.cs
```diff
@@ -0,0 +1,317 @@
...
+using Nethermind.Xdc.Types;
+using NSubstitute;
+using NUnit.Framework;
+
+namespace Nethermind.Xdc.Test.ModuleTests;
+
+public class RewardTests
+{
+    [Test]
+    public async Task TestHookRewardV2()
```
ak88, line 30: Maybe add a link to the test in XDC repo so there is an explanation for the naming

ak88 (approved):

PR #9886, opened 2025-12-05
fix(evm): use IsTracingOpLevelStorage for op-level storage events

ANtutov:
Switch TLOAD, TSTORE, SSTORE (un/metered) and SLOAD to guard Set/LoadOperation* calls with IsTracingOpLevelStorage rather than IsTracingStorage.
This matches the ITxTracer contract and wrapper tracers, ensuring op-level traces are emitted when only op-level storage tracing is enabled.


ANtutov:
> @ANtutov please fix

@LukaszRozmej Hi, my bad sorry, I see this problem was solved in #9846 if I'm not mistaken, but with different solution:


> @ANtutov please fix

image
I guess this problem was fixed with different solution (on screenshot)
so I think I should close PR

PR #9889, opened 2025-12-05
Fix hive getLogs

deffrian:
Fixes Closes Resolves #

```
eth_getLogs/filter-error-blockHash-and-range (nethermind)
eth_getLogs/filter-error-future-block-range (nethermind)

```
## Changes

- Refactor eth_getLogs a bit
- Add checks

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_


src/Nethermind/Nethermind.Facade/Filters/FilterStore.cs
```diff
@@ -202,48 +202,40 @@ private static TopicExpression GetTopicExpression(FilterTopic? filterTopic)
             return AnyTopic.Instance;
         }
 
-        private static AddressFilter GetAddress(object? address) =>
-            address switch
+        private static AddressFilter GetAddress(AddressAsKey[]? address)
```
alexb5dh: Parameter should probably be named `addresses`, just to avoid confusion.

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/Filter.cs
```diff
@@ -31,44 +32,65 @@ public void ReadJson(JsonElement filter, JsonSerializerOptions options)
...
+                    throw new ArgumentException("either (fromBlock and toBlock) or blockHash have to be specified");
+                }
+
+                FromBlock = new(new Hash256(blockHashElement.ToString()));
+                ToBlock = FromBlock;
             }
             else
             {
-                FromBlock = ToBlock = BlockParameterConverter.GetBlockParameter(blockHash);
+                FromBlock = hasFromBlock ? new BlockParameter(LongConverter.FromString(fromBlockElement.ToString())) : BlockParameter.Earliest;
```
alexb5dh: `BlockParameterConverter` also handled all non-integer values like `earliest`, `latest`, etc. I don't see this functionality in `LongConverter`, are these values still handled correctly?

In case if not - would be good to add some unit tests on these values specifically after a fix.

src/Nethermind/Nethermind.Facade/Filters/AddressFilter.cs
```diff
@@ -19,9 +19,16 @@ public AddressFilter(Address address)
             Address = address;
         }
 
-        public AddressFilter(HashSet<AddressAsKey> addresses)
+        public AddressFilter(AddressAsKey[] addresses)
```
alexb5dh: What was the reason behind replacing `HashSet` with an array?

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/Filter.cs
```diff
@@ -31,44 +32,65 @@ public void ReadJson(JsonElement filter, JsonSerializerOptions options)
...
-            if (blockHash is null)
+            if (hasBlockHash)
             {
-                filter.TryGetProperty("fromBlock"u8, out JsonElement fromBlockElement);
-                FromBlock = BlockParameterConverter.GetBlockParameter(fromBlockElement.ToString());
-                filter.TryGetProperty("toBlock"u8, out JsonElement toBlockElement);
-                ToBlock = BlockParameterConverter.GetBlockParameter(toBlockElement.ToString());
+                if (hasFromBlock || hasToBlock)
+                {
+                    throw new ArgumentException("either (fromBlock and toBlock) or blockHash have to be specified");
```
alexb5dh: Honestly, geth message looks better to me, but it may be subjective:
```
cannot specify both BlockHash and FromBlock/ToBlock
```

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/Filter.cs
alexb5dh: I also see that earliest/latest params were removed/replaced in many tests, was this intentional?

deffrian: Yes, geth supports only hex numbers or null(for earliest latest)

src/Nethermind/Nethermind.Facade/Filters/AddressFilter.cs
deffrian: I think AddressFilter should decide how addresses are stored inside

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/Filter.cs
deffrian: My bad, they actually support it. Got confused

src/Nethermind/Nethermind.Facade/Filters/AddressFilter.cs
alexb5dh: How about accepting `IEnumerable` instead? I see in some places we are calling `ToArray` to immediately follow with `ToHashSet` after.

src/Nethermind/Nethermind.Facade/Filters/AddressFilter.cs
```diff
@@ -9,39 +9,27 @@ namespace Nethermind.Blockchain.Filters
...
-            }
-
-            return Address is null || Address == address;
-        }
+        public bool Accepts(Address address) => Addresses.Count == 0 || Addresses.Contains(address);
 
         public bool Accepts(ref AddressStructRef address)
         {
-            if (Addresses?.Count > 0)
+            if (Addresses.Count > 0)
```
Marchhill, line 32: why not use `Addresses.Contains`?

src/Nethermind/Nethermind.Facade/Filters/AddressFilter.cs
```diff
@@ -51,55 +39,39 @@ public bool Accepts(ref AddressStructRef address)
...
 
-            return Address is null || Address == address;
+            return true;
         }
 
         public bool Matches(Bloom bloom)
         {
-            if (Addresses is not null)
+            bool result = true;
+            var indexes = AddressesBloomExtracts;
```
Marchhill: unneeded variable? Also use type instead of `var`

src/Nethermind/Nethermind.Facade/Filters/FilterStore.cs
```diff
@@ -202,48 +202,40 @@ private static TopicExpression GetTopicExpression(FilterTopic? filterTopic)
             return AnyTopic.Instance;
         }
 
-        private static AddressFilter GetAddress(object? address) =>
-            address switch
+        private static AddressFilter GetAddress(AddressAsKey[]? addresses)
+        {
```
Marchhill: could be one-liner

src/Nethermind/Nethermind.Facade/Filters/FilterStore.cs
```diff
@@ -202,48 +202,40 @@ private static TopicExpression GetTopicExpression(FilterTopic? filterTopic)
...
+                    Topic = topics[0]
+                };
             }
-
-            return obj is not IEnumerable<string> topics
-                ? null
-                : new FilterTopic
+            else
+            {
+                return new FilterTopic()
```
Marchhill, line 220: could `FilterTopic` also be refactored to just have `Topics`?

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
```diff
@@ -325,10 +325,10 @@ public bool TryGetLogs(int filterId, out IEnumerable<FilterLog> filterLogs, Canc
             return filter is not null;
         }
 
-        public int NewFilter(BlockParameter? fromBlock, BlockParameter? toBlock,
-            object? address = null, IEnumerable<object>? topics = null)
+        public int NewFilter(BlockParameter fromBlock, BlockParameter toBlock,
```
Marchhill, line 328: why not nullable here but nullable in interface?

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/Filter.cs
```diff
@@ -31,44 +32,69 @@ public void ReadJson(JsonElement filter, JsonSerializerOptions options)
...
+    private static AddressAsKey[]? GetAddress(JsonElement token, JsonSerializerOptions options)
+    {
+        switch (token.ValueKind)
+        {
+            case JsonValueKind.Undefined or JsonValueKind.Null:
+                return null;
+            case JsonValueKind.String:
+                return [new AddressAsKey(new Address(token.ToString()))];
+            case JsonValueKind.Array:
+                var enumerator = token.EnumerateArray();
```
Marchhill, line 84: use explicit type

src/Nethermind/Nethermind.Facade/Filters/AddressFilter.cs
deffrian: If it's empty we pass all

src/Nethermind/Nethermind.Facade/Filters/FilterStore.cs
deffrian: I'm not sure why we have this complicated machinery with SpecificTopic and OrExpression. I just don't want to touch it

src/Nethermind/Nethermind.Facade/Filters/AddressFilter.cs
```diff
@@ -9,39 +9,27 @@ namespace Nethermind.Blockchain.Filters
...
+        public AddressFilter(Address address) : this([address])
         {
-            Address = address;
         }
 
-        public AddressFilter(HashSet<AddressAsKey> addresses)
+        public AddressFilter(IEnumerable<AddressAsKey> addresses)
         {
-            Addresses = addresses;
+            Addresses = addresses.ToHashSet();
```
LukaszRozmej, line 22: Unnecessary allocation

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/Filter.cs
```diff
@@ -31,44 +32,69 @@ public void ReadJson(JsonElement filter, JsonSerializerOptions options)
...
+                List<AddressAsKey> result = new();
+                while (enumerator.MoveNext())
+                {
+                    result.Add(new(new Address(enumerator.Current.ToString())));
+                }
+
+                return result.ToArray();
```
LukaszRozmej, line 91: So you:
1. Create a list
2. Convert it to Array
3. Later Convert it to HashSet

Couldn't we just do HashSet here and use `token.GetArrayLength()` to hint it the capacity?

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/Filter.cs
```diff
@@ -77,16 +103,27 @@ public void ReadJson(JsonElement filter, JsonSerializerOptions options)
...
+                    break;
+                case JsonValueKind.Array:
+                    JsonElement.ArrayEnumerator enumerator = token.EnumerateArray();
+                    List<Hash256> result = new();
+                    while (enumerator.MoveNext())
+                    {
+                        result.Add(new(enumerator.Current.ToString()));
+                    }
+
+                    yield return result.ToArray();
```
LukaszRozmej, line 122: Same here

PR #9896, opened 2025-12-07
Remove dead concurrency toggle in BLS MSM precompiles

0xcharry:
remove DisableConcurrency flag from BlsConst and drop sequential MSM branches that were unreachable, keep the single parallel decoding path in G1MSMPrecompile and G2MSMPrecompile, eliminating #pragma suppressions, behavior stays identical while the code becomes simpler and warning-free


LukaszRozmej:
It's there for a reason

PR #9902, opened 2025-12-08
fix: correct slice index in Address.TryParseVariableLength

reject-i:
Used value.Length instead of span.Length when slicing overflow addresses with "0x" prefix. This caused 2 extra hex chars to be removed.


src/Nethermind/Nethermind.Core.Test/AddressTests.cs
```diff
@@ -183,6 +183,7 @@ public bool Is_PointEvaluationPrecompile_properly_activated(IReleaseSpec spec) =
     [TestCase(Address.SystemUserHex, false)]
     [TestCase("2" + Address.SystemUserHex, false)]
     [TestCase("2" + Address.SystemUserHex, true)]
+    [TestCase("0x00" + "fffffffffffffffffffffffffffffffffffffffe", true)]
```
LukaszRozmej, line 187:
```suggestion
    [TestCase("0x00" + Address.SystemUserHex, true)]
```

LukaszRozmej (approved):

src/Nethermind/Nethermind.Core.Test/AddressTests.cs
```diff
@@ -183,6 +183,7 @@ public bool Is_PointEvaluationPrecompile_properly_activated(IReleaseSpec spec) =
     [TestCase(Address.SystemUserHex, false)]
     [TestCase("2" + Address.SystemUserHex, false)]
     [TestCase("2" + Address.SystemUserHex, true)]
+    [TestCase("0x00" + Address.SystemUserHex, true)]
```
LukaszRozmej, line 186:
```suggestion
    [TestCase("0x00" + Address.SystemUserHex, true)]
    [TestCase("0x00" + "fffffffffffffffffffffffffffffffffffffffe", true)]
```

PR #9904, opened 2025-12-09
fix: add SyncServer.HintBlock call to eth/69 BlockRangeUpdate handler

letmehateu:
Fixes Closes Resolves #

## Changes

- **Eth69ProtocolHandler.cs**: Added `SyncServer.HintBlock()` call in `Handle(BlockRangeUpdateMessage)` to notify the sync server about the latest block in the announced range
- **SyncServerTests.cs**: 
  - Added test `Eth69_does_not_request_peer_refresh_on_known_block_range_update()` to verify known blocks don't trigger peer refresh
  - Added test `Eth69_requests_peer_refresh_on_unknown_block_range_update()` to verify unknown blocks trigger peer refresh
  - Removed TODO comment requesting these tests

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks

### DIFF
--- a/src/Nethermind/Nethermind.Synchronization.Test/SyncServerTests.cs
+++ b/src/Nethermind/Nethermind.Synchronization.Test/SyncServerTests.cs
@@ -40,7 +40,6 @@ using BlockTree = Nethermind.Blockchain.BlockTree;
 
 namespace Nethermind.Synchronization.Test;
 
-// TODO: add tests for new block notification for eth/69
 [Parallelizable(ParallelScope.All)]
 public class SyncServerTests
 {


src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/V69/Eth69ProtocolHandler.cs
```diff
@@ -136,6 +136,7 @@ private void Handle(BlockRangeUpdateMessage blockRangeUpdate)
         _remoteHeadBlockHash = blockRangeUpdate.LatestBlockHash;
         HeadNumber = blockRangeUpdate.LatestBlock;
         HeadHash = blockRangeUpdate.LatestBlockHash;
+        SyncServer.HintBlock(blockRangeUpdate.LatestBlockHash, blockRangeUpdate.LatestBlock, this);
```
LukaszRozmej, line 139: Good question is it actually needed. We manually update `HeadNumber` and `HeadHash` here. In post-merge networks we will also have `CanGossipBlocks` set to `false`. The only thing the sync server does here is `RefreshTotalDifficulty`.

So the only thing where this would make a change is non-merge network that still relies on block gossip, but then we would already know that from block gossiping, making this call redundant.

LukaszRozmej (commented):
Probably not needed (check comment) but will wait for @alexb5dh, @marcindsobczak and others to share their opinions.

src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/V69/Eth69ProtocolHandler.cs
alexb5dh: Also don't think this will change anything.

src/Nethermind/Nethermind.Synchronization.Test/SyncServerTests.cs
```diff
@@ -75,6 +74,24 @@ public void Requests_peer_refresh_on_unknown_hints()
...
+        ctx.PeerPool.DidNotReceiveWithAnyArgs().RefreshTotalDifficulty(null!, null!);
+    }
+
+    [Test]
+    public void Eth69_requests_peer_refresh_on_unknown_block_range_update()
+    {
+        Context ctx = new();
+        ctx.BlockTree.IsKnownBlock(100, TestItem.KeccakB).ReturnsForAnyArgs(false);
+        ctx.SyncServer.HintBlock(TestItem.KeccakB, 100, ctx.NodeWhoSentTheBlock);
+        ctx.PeerPool.Received().ReceivedWithAnyArgs();
```
alexb5dh, line 92: Looks like checked method is missing here.

PR #9908, opened 2025-12-09
Optimize ColumnDb MultiGet allocations

0xxFloki:
replace the LINQ Select(...).ToArray() in ColumnDb’s indexer with a direct Array.Fill, eliminate the per-call delegate and enumerable allocations while keeping the returned data unchanged

### DIFF
--- a/src/Nethermind/Nethermind.Db.Rocks/ColumnDb.cs
+++ b/src/Nethermind/Nethermind.Db.Rocks/ColumnDb.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using Nethermind.Core;
 using Nethermind.Core.Crypto;
 using RocksDbSharp;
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Db.Rocks/ColumnDb.cs
```diff
@@ -62,8 +62,16 @@ public void Merge(ReadOnlySpan<byte> key, ReadOnlySpan<byte> value, WriteFlags w
...
-    public KeyValuePair<byte[], byte[]?>[] this[byte[][] keys] =>
-        _rocksDb.MultiGet(keys, keys.Select(k => _columnFamily).ToArray());
+    public KeyValuePair<byte[], byte[]?>[] this[byte[][] keys]
+    {
+        get
+        {
+            ColumnFamilyHandle[] columnFamilies = new ColumnFamilyHandle[keys.Length];
+            Array.Fill(columnFamilies, _columnFamily);
+
+            return _rocksDb.MultiGet(keys, columnFamilies);
```
LukaszRozmej, line 70: To be honest, proper optimization would be to somehow avoid that allocation at all

0xxFloki: now everything should be perfect, thanks for pointing that out

0xxFloki: i think all good))

PR #9910, opened 2025-12-10
Disable publishing to Downloads

stdevMac:
Remove from release the upload to the downloads page


kamilchodola:
More like temporary removal due to provider issue and whole page is down now with not easy way to quickly restore it so to avoid failed steps we will temporarily remove this option as GH releases are just as fine.

PR #9911, opened 2025-12-10
fix: unify ExpectedErrors type to string[] in test classes

avorylli:
Changed ExpectedErrors property type from List<string> to string[] in CheckHealthTest class to match CheckHealthPostMergeTest. 

This ensures consistency across test data classes and aligns with the actual return type from CheckHealthResult.Errors (IEnumerable<string>).


LukaszRozmej:
meaningless change

PR #9916, opened 2025-12-10
Move ExecutionEnvironment to Rentable sealed class

LukaszRozmej:
## Changes

- Move ExecutionEnvironment to Rentable sealed class
- Avoid `System.Buffer.BulkMoveWithWriteBarrier(ref Byte, ref Byte, UIntPtr)` on World trace, which took about 20% of the time!

image

- [x] Optimization

#### Requires testing

- [x] Yes

- [x] No

#### Notes on testing

Existing tests + running nodes should suffice.

## Remarks

This needs complex performance evaluation on different nodes and benchmarks.

### DIFF
--- a/src/Nethermind/Nethermind.Evm/EvmState.cs
+++ b/src/Nethermind/Nethermind.Evm/EvmState.cs
@@ -100,10 +100,6 @@ public sealed class EvmState : IDisposable // TODO: rename to CallState
     private StackAccessTracker _accessTracker;
     private Snapshot _snapshot;
 
-#if DEBUG
-    private StackTrace? _creationStackTrace;
-#endif
-
     /// <summary>
     /// Rent a top level <see cref="EvmState"/>.
     /// </summary>
@@ -267,6 +263,9 @@ public sealed class EvmState : IDisposable // TODO: rename to CallState
     }
 
 #if DEBUG
+
+    private StackTrace? _creationStackTrace;
+
     ~EvmState()
     {
         if (!_isDisposed)
--- a/src/Nethermind/Nethermind.Evm/ExecutionEnvironment.cs
+++ b/src/Nethermind/Nethermind.Evm/ExecutionEnvironment.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Concurrent;
+using System.Diagnostics;
 using Nethermind.Core;
 using Nethermind.Evm.CodeAnalysis;
 using Nethermind.Int256;
@@ -48,7 +49,7 @@ namespace Nethermind.Evm
 
         /// <summary>
         /// Value information passed (it is different from transfer value in DELEGATECALL.
-        /// DELEGATECALL behaves like a library call and it uses the value information from the caller even
+        /// DELEGATECALL behaves like a library call, and it uses the value information from the caller even
         /// as no transfer happens.
         /// </summary>
         public ref readonly UInt256 Value => ref _value;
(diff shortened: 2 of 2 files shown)


src/Nethermind/Nethermind.Evm/EvmState.cs
```diff
@@ -255,7 +255,11 @@ public void Dispose()
         _memory.Dispose();
         _memory = default;
         _accessTracker = default;
-        _env = default;
+        if (_env is not null)
+        {
+            _env.Return();
+        }
```
alexb5dh:
```suggestion
        _env?.Return();
```

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -683,56 +684,56 @@ private int ExecuteEvmCall<TTracingInst>(
 
             ExecutionType executionType = tx.IsContractCreation ? ((spec.IsEofEnabled && tx.IsEofContractCreation) ? ExecutionType.TXCREATE : ExecutionType.CREATE) : ExecutionType.TRANSACTION;
 
-            using (EvmState state = EvmState.RentTopLevel(gasAvailable, executionType, in env, in accessedItems, in snapshot))
+            using (EvmState state = EvmState.RentTopLevel(gasAvailable, executionType, env, in accessedItems, in snapshot))
```
alexb5dh, line 686: It may be worth adding `env.Return()` if `EvmState` was not created (as of now only its `Dispose` call is doing `Return` here).

Probably always under `FailContractCreate` label, and maybe under some condition under `Complete` label. Or make `Rent` method idempotent and call it in any case.

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -683,56 +684,56 @@ private int ExecuteEvmCall<TTracingInst>(
...
                 substate = VirtualMachine.ExecuteTransaction<TTracingInst>(state, WorldState, tracer);
                 Metrics.IncrementOpCodes(VirtualMachine.OpCodeCount);
                 gasAvailable = state.GasAvailable;
-            }
 
-            if (tracer.IsTracingAccess)
-            {
-                tracer.ReportAccess(accessedItems.AccessedAddresses, accessedItems.AccessedStorageCells);
-            }
+                if (tracer.IsTracingAccess)
```
alexb5dh, line 692: Was indentation wrong before?

LukaszRozmej: It went under using in line 687

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
LukaszRozmej: solved in different way on TopLevel frames

src/Nethermind/Nethermind.Evm/ExecutionEnvironment.cs
```diff
@@ -2,60 +2,103 @@
...
+        {
+            CodeInfo = null!;
+            ExecutingAccount = null!;
+            Caller = null!;
+            CodeSource = null;
+            CallDepth = 0;
+            _transferValue = default;
+            _value = default;
+            InputData = default;
+            _pool.Enqueue(this);
```
alexb5dh: How about adding `isDisposed` flag (or maybe just checking for `CodeInfo == null`) to make `Dispose` safe to call multiple times? Will make class simpler to use.

benaadams (approved):

PR #9930, opened 2025-12-13
perf: reuse ancestor headers in UnclesValidator

yyhrnk:
UnclesValidator used to walk the parent chain twice per uncle: once inside IsKin via FindParentHeader and once more when checking if an uncle was already included by an ancestor via repeated FindBlock calls. This change caches a bounded list of ancestor headers for the block once, reuses them for the kinship check, resolves ancestor blocks only when needed to inspect their uncles for duplicates, and removes the stale performance Todo attribute. The original validation semantics are preserved and existing UnclesValidatorTests continue to pass.


stdevMac:
@LukaszRozmej @benaadams is this still valid or we can close it?

LukaszRozmej:
@claude review

LukaszRozmej:
@claude re-review

LukaszRozmej (commented):
Two concerns about the new approach:

### 1. Eager ancestor walk regresses the 0-uncle case

The original code walks ancestors only *inside* the `for (int i = 0; i < uncles.Length; i++)` loop, so when `uncles.Length == 0` the method returns `true` without touching the block tree. The new code does 6 `FindParentHeader` calls + a heap allocation **unconditionally**, before checking whether there's anything to validate.

`MergeUnclesValidator` short-circuits post-merge mainnet, but `UnclesValidator` is still hit during pre-merge sync / archive replay of PoW history and on PoW-only chains — where the overwhelming majority of blocks also have zero uncles. We'd be adding a fixed cost per block to a path that previously paid nothing.

Suggest hoisting the walk behind `if (uncles.Length == 0) return true;`, or lazy-initializing the ancestors array on first iteration of the uncles loop.

### 2. Use a stack-allocated buffer instead of `new BlockHeader[6]`

The `BlockHeader[maxAncestorsDepth]` allocation can be avoided. Options:

- `ArrayPoolListRef<BlockHeader>` (with `using var` for disposal), or
- A C# inline-array struct, e.g.:
  ```csharp
  [InlineArray(6)] private struct AncestorsBuffer { private BlockHeader? _e0; }
  ...
  AncestorsBuffer buffer = default;
  Span<BlockHeader?> ancestors = buffer;
  ```

The allocation is small but happens per block on a validation hot path, so it's worth removing — especially once point (1) is fixed and we only allocate when there are uncles to check.

src/Nethermind/Nethermind.Consensus/Validators/UnclesValidator.cs
```diff
@@ -2,22 +2,32 @@
...
 using Nethermind.Logging;
 
 namespace Nethermind.Consensus.Validators
 {
-    [Todo(Improve.Performance, "We execute the search up the tree twice - once for IsKin and once for HasAlreadyBeenIncluded")]
     public class UnclesValidator(IBlockTree? blockTree, IHeaderValidator? headerValidator, ILogManager? logManager) : IUnclesValidator
     {
+        // Yellow Paper §11.1: an uncle must share an ancestor with the block within 6 generations.
+        private const int RelationshipLevel = 6;
+        // Yellow Paper §11.1: an uncle is invalid if it was already included by one of the 5 direct ancestors.
```
LukaszRozmej:
```suggestion
        private const int RelationshipLevel = 6;
```

LukaszRozmej (approved):

flcl42 (approved):

PR #9931, opened 2025-12-14
Fix hanging scenario if persisted node not able to be pruned

asdacap:
- Fix blocked `BeginScope` during pruning persisted nodes. 

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- Make processing range 4500000 to 5000000 faster.

### DIFF
--- a/src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
+++ b/src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
@@ -607,7 +607,7 @@ public sealed class TrieStore : ITrieStore, IPruningTrieStore
                         PrunePersistedNodes();
                     }
 
-                    if (_lastPrunedShardIdx - startingShard >= _shardedDirtyNodeCount && _pruningStrategy.ShouldPrunePersistedNode(CaptureCurrentState()))
+                    if (!IsInCommitBufferMode && _lastPrunedShardIdx - startingShard >= _shardedDirtyNodeCount && _pruningStrategy.ShouldPrunePersistedNode(CaptureCurrentState()))
                     {
                         // A persisted nodes that was recommitted and is still within pruning boundary cannot be pruned.
                         // This should be rare but can happen, notably in mainnet block 4500000 around there, But this


src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -588,17 +588,37 @@ internal void SyncPruneCheck()
...
-                while (i < maxTry && _pruningStrategy.ShouldPrunePersistedNode(CaptureCurrentState()))
+                try
                 {
-                    PrunePersistedNodes();
-                    i++;
+                    // When `_pruningLock` is held, the begin commit will check for _toBePersistedBlockNumber in order
+                    // to decide which block to be used as the boundary for the commit buffer. This number was re-set
+                    // to -1 in `PersistAndPruneDirtyCache`. So we need to re-set it here, otherwrise `BeginScope` will hang
+                    // until the prune persisted node loop is completed.
+                    _toBePersistedBlockNumber = LastPersistedBlockNumber;
```
LukaszRozmej, line 597: Should this be done only if `PersistAndPruneDirtyCache` actually was invoked in the above `if`?

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -588,17 +588,37 @@ internal void SyncPruneCheck()
...
+
+                    // `PrunePersistedNodes` only work on part of the partition at any one time. With commit buffer,
+                    // it is possible that the commit buffer once flushed will immediately trigger another prune, which
+                    // mean `PrunePersistedNodes` was not able to re-trigger multiple time, which make the persisted node
+                    // cache even bigger which causes longer prune which causes bigger commit buffer, etc.
+                    // So we loop it here until `ShouldPrunePersistedNode` return false.
+                    int startingShard = _lastPrunedShardIdx;
+                    while (_lastPrunedShardIdx - startingShard < _shardedDirtyNodeCount && _pruningStrategy.ShouldPrunePersistedNode(CaptureCurrentState()))
```
LukaszRozmej, line 605: minor: `TrieStoreState` could be converted to `struct` and passed via `in`

LukaszRozmej (approved):

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
asdacap: Still needed as _pruningLock is held.

PR #9935, opened 2025-12-14
Faster protocol parsing, add Ethrex as node/brand type

benaadams:
## Changes

- Fast protocol parsing without materialising `string` or `.SequenceEquals` (via utf8 as number representation)
- Add Ethrex as client (plus a number of peer types to reduce Unknowns)

Aside probably better if `Protocol` was an enum, but is a pretty major refactor

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Network.Stats/Model/Node.cs
+++ b/src/Nethermind/Nethermind.Network.Stats/Model/Node.cs
@@ -2,7 +2,10 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 
 using System;
+using System.Linq;
 using System.Net;
+using System.Text.RegularExpressions;
+using FastEnumUtility;
 using Nethermind.Config;
 using Nethermind.Core.Crypto;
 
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Network.Contract/P2P/ProtocolParser.cs
```diff
@@ -0,0 +1,84 @@
...
+                if (key2 == AA)
+                {
+                    protocol = Protocol.AA;
+                    return true;
+                }
+                break;
+        }
+        return false;
+    }
+}
```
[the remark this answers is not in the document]
benaadams: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/NethermindEth/nethermind/pull/9935#discussion_r2617394633)

src/Nethermind/Nethermind.Network.Contract/P2P/ProtocolParser.cs
```diff
@@ -0,0 +1,84 @@
...
+                ushort key2 = (ushort)(protocolSpan[0] | (protocolSpan[1] << 8));
+                if (key2 == AA)
+                {
+                    protocol = Protocol.AA;
+                    return true;
+                }
+                break;
+        }
+        return false;
+    }
```
LukaszRozmej, line 83: why not start the comparision by switch on the first letter and then do the rest?

src/Nethermind/Nethermind.Network.Stats/Model/Node.cs
```diff
@@ -227,6 +227,10 @@ public static NodeClientType RecognizeClientType(string clientId)
             {
                 return NodeClientType.EthereumJS;
             }
+            else if (clientId.Contains(nameof(NodeClientType.Ethrex), StringComparison.OrdinalIgnoreCase))
```
LukaszRozmej: Do switch on the first letter to reduce the number of comparisions? (Maybe Regex would be better?)

src/Nethermind/Nethermind.Network.Contract/P2P/ProtocolParser.cs
benaadams: Don't know if we can use the first letter without checking size; at which point might as well use size as first comparison

LukaszRozmej: It is just a bit less readable, but fine.

src/Nethermind/Nethermind.Network.Stats/Model/Node.cs
LukaszRozmej: I do think this one can be improved with single compiled `Regex` pass?

LukaszRozmej: @copilot please generate optimized regex for the `RecognizeClientType` method.

benaadams: Is an ignore case contains (rather than starts with)

src/Nethermind/Nethermind.Network.Stats/Model/Node.cs
```diff
@@ -193,56 +196,66 @@ public bool Equals(Node other)
...
+                            nameof(NodeClientType.Besu),
+                            nameof(NodeClientType.Erigon),
+                            nameof(NodeClientType.Nimbus),
+                            nameof(NodeClientType.Ethrex),
+                            nameof(NodeClientType.EthereumJS),
+                            nameof(NodeClientType.OpenEthereum),
+                            nameof(NodeClientType.Parity),
+                        })
+                        .OrderByDescending(name => name.Length))),
+            RegexOptions.Compiled | RegexOptions.IgnoreCase);
```
LukaszRozmej, line 236: @copilot Should `CultureInvariant` or `NonBacktracking` be good here in any way?

PR #9938, opened 2025-12-15
Drop eth66 and eth67 support, add eth69 as a default

benaadams:
## Changes

- Drop eth66 and eth67 as default supported protocol (Geth dropped last year); all mainnet clients support eth69

#### Requires testing

- [x] Yes - updated tests


src/Nethermind/Nethermind.Network/ProtocolsManager.cs
```diff
@@ -41,9 +41,8 @@ public class ProtocolsManager : IProtocolsManager
     {
         public static readonly IEnumerable<Capability> DefaultCapabilities = new Capability[]
         {
-            new(Protocol.Eth, 66),
-            new(Protocol.Eth, 67),
             new(Protocol.Eth, 68),
+            new(Protocol.Eth, 69),
             new(Protocol.NodeData, 1)
```
flcl42, line 45: @marcindsobczak do we still need NodeData?

src/Nethermind/Nethermind.Network.Test/P2P/P2PProtocolInfoProviderTests.cs
```diff
@@ -14,7 +14,7 @@ public class P2PProtocolInfoProviderTests
         public void DefaultCapabilitiesToString_ReturnExpectedResult()
         {
             string result = P2PProtocolInfoProvider.DefaultCapabilitiesToString();
-            Assert.That(result, Is.EqualTo("eth/68,eth/67,eth/66,nodedata/1"));
+            Assert.That(result, Is.EqualTo("eth/69,eth/68,nodedata/1"));
```
flcl42: @ak88 do you need 66, 67 for XDC?

src/Nethermind/Nethermind.Network/ProtocolsManager.cs
marcindsobczak: We have snap serving now, so we can remove all NodeData code at all. It was workaround to sync nodes in nethermind-only networks

marcindsobczak (approved):

src/Nethermind/Nethermind.Network.Test/P2P/P2PProtocolInfoProviderTests.cs
ak88: This are default capabilities which we are overwriting anyway.

src/Nethermind/Nethermind.Network/ProtocolsManager.cs
ak88: > We have snap serving now, so we can remove all NodeData code at all. It was workaround to sync nodes in nethermind-only networks

We need Node data for state sync on XDC since there is no snap support.

flcl42: Node data stays if you do not activate eth/67, so I guess in particular NodeData protocol(which duplicates pre-eth/67 functionality) is not required.

LukaszRozmej: I would keep node data, we have some mechanism that can fallback to it like on healing

MarekM25 (approved):

flcl42:
So you can confirm it works for all the networks right?

PR #9939, opened 2025-12-15
Adjust tx flood control to keep Besu as peer more

benaadams:
## Changes

- After looking to why I never have Besu peers; turns out they send too many txs with old nonces; which could just be them lagging/sending tx while verifying blocks (which are then included) etc. e.g. @LukaszRozmej has Besu peers but his machine is slow
- Also don't count rejected txs as a disconnect reason when syncing

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/V62/TxFloodController.cs
+++ b/src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/V62/TxFloodController.cs
@@ -38,7 +38,8 @@ internal class TxFloodController(Eth62ProtocolHandler protocolHandler, ITimestam
             }
 
             if (accepted is
-                { Id:
+                {
+                    Id:
                     TxResultCode.Invalid or
                     TxResultCode.GasLimitExceeded or
                     TxResultCode.MaxTxSizeExceeded or
--- a/src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/V62/TxFloodController.cs
+++ b/src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/V62/TxFloodController.cs
@@ -1,72 +1,92 @@
 // SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
 // SPDX-License-Identifier: LGPL-3.0-only
 
+using System;
 using Nethermind.Core;
 using Nethermind.Logging;
 using Nethermind.Stats.Model;
 using Nethermind.TxPool;
-using System;
 
-namespace Nethermind.Network.P2P.Subprotocols.Eth.V62
+namespace Nethermind.Network.P2P.Subprotocols.Eth.V62;
+
+internal class TxFloodController(Eth62ProtocolHandler protocolHandler, ITimestamper timestamper, ILogger logger)
 {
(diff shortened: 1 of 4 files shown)


src/Nethermind/Nethermind.TxPool/AcceptTxResult.cs
```diff
@@ -3,132 +3,155 @@
...
+    SenderIsContract,
+    NonceTooFarInFuture,
+    PendingTxsOfConflictingType,
+    NotSupportedTxType,
+    MaxTxSizeExceeded,
+    NotCurrentNonceForDelegation,
+    DelegatorHasPendingTx,
+    PermissionDenied = 100,
+    Syncing = 503
+}
```
LukaszRozmej, line 31: So I think this needs to be extendable by plugins. Which is still is just pointing out why we used int

src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/V62/TxFloodController.cs
```diff
@@ -1,72 +1,93 @@
...
+    private long _notAcceptedSinceLastCheck;
+    private readonly Random _random = new();
 
-        public void Report(AcceptTxResult accepted)
-        {
-            TryReset();
+    internal bool IsDowngraded { get; private set; }
 
-            if (!accepted)
+    public void Report(in AcceptTxResult accepted)
```
LukaszRozmej, line 28: Can we rename? `accepted` suggest it was accepted, while we are explicitly checking for rejections

```suggestion
    public void Report(in AcceptTxResult acceptedResult)
```

src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/V62/TxFloodController.cs
```diff
@@ -1,72 +1,93 @@
...
+                    TxResultCode.GasLimitExceeded or
+                    TxResultCode.MaxTxSizeExceeded or
+                    TxResultCode.Int256Overflow or
+                    TxResultCode.SenderIsContract or
+                    TxResultCode.FailedToResolveSender or
+                    TxResultCode.NotSupportedTxType
+                })
+            {
+                if (logger.IsDebug) logger.Debug($"Disconnecting {_protocolHandler} due to invalid tx received");
+                _protocolHandler.Disconnect(DisconnectReason.InvalidTx, accepted.Id.ToString());
```
LukaszRozmej, line 53: Just use the message!
No alloc + plugin extensibility

```suggestion
                _protocolHandler.Disconnect(DisconnectReason.InvalidTx, accepted.Message);
```

LukaszRozmej (approved):

flcl42:
Can't be/shouldn't be fixed on besu side?

benaadams:
> Can't be/shouldn't be fixed on besu side?

Is a gap where if your node processes block; all the txs for that block will be invalid but node that hasn't validated block will see those txs as still valid and send them, so need some allowance

This is also why I think a faster machine (e.g. mine) currently will drop the peers but a slower machine will keep them (e.g. Lukasz's)

PR #9946, opened 2025-12-15
Gnosis Balancer permanent changes

Marchhill:
## Changes

- Gnosis balancer fork
- Override code based on timestamp AuRa

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus.AuRa/Config/AuRaChainSpecEngineParameters.cs
+++ b/src/Nethermind/Nethermind.Consensus.AuRa/Config/AuRaChainSpecEngineParameters.cs
@@ -57,6 +57,20 @@ public class AuRaChainSpecEngineParameters : IChainSpecEngineParameters
     public IDictionary<long, IDictionary<Address, byte[]>> RewriteBytecode { get; set; } = new Dictionary<long, IDictionary<Address, byte[]>>();
     public IDictionary<ulong, IDictionary<Address, byte[]>> RewriteBytecodeTimestamp { get; set; } = new Dictionary<ulong, IDictionary<Address, byte[]>>();
 
+    public IEnumerable<(ulong, Address, byte[])> RewriteBytecodeTimestampParsed
+    {
+        get
+        {
+            foreach (KeyValuePair<ulong, IDictionary<Address, byte[]>> timestampOverrides in RewriteBytecodeTimestamp)
+            {
+                foreach (KeyValuePair<Address, byte[]> addressOverride in timestampOverrides.Value)
+                {
+                    yield return (timestampOverrides.Key, addressOverride.Key, addressOverride.Value);
+                }
+            }
+        }
+    }
+
     public Address WithdrawalContractAddress { get; set; }
 
     private AuRaParameters.Validator? _validators;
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Specs/GnosisForks/19_BalancerGnosis.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Nethermind.Specs.GnosisForks;
+
+public class BalancerGnosis : PragueGnosis
+{
+    private static IReleaseSpec _instance;
+
+    private BalancerGnosis() : base()
+    {
+        Name = "Balancer";
```
LukaszRozmej: Shouldn't this initiate the `rewriteBytecodeTimestamp` somehow?

src/Nethermind/Nethermind.Consensus.AuRa/ContractRewriter.cs
```diff
@@ -8,26 +8,45 @@
...
         bool result = false;
-        if (_contractOverrides.TryGetValue(blockNumber, out IDictionary<Address, byte[]> overrides))
+        foreach (KeyValuePair<ulong, IDictionary<Address, byte[]>> overrides in _contractOverridesTimestamp)
         {
-            foreach (KeyValuePair<Address, byte[]> contractOverride in overrides)
+            if (timestamp >= overrides.Key && parentTimestamp < overrides.Key)
             {
-                stateProvider.InsertCode(contractOverride.Key, contractOverride.Value, spec);
-                result = true;
+                result |= InsertOverwriteCode(overrides.Value, stateProvider, spec);
```
LukaszRozmej: I kind of dislike this. I think this would work better as a list/array than dictionary.
And if we would expect bigger lists we could do binary search. For now we can leave it as iteration, but list would be better.

LukaszRozmej (commented):
I know it was done this way, but could we change it to list for master?

src/Nethermind/Nethermind.Specs/GnosisForks/19_BalancerGnosis.cs
Marchhill: I don't think so, we're not making any changes to release spec. The code overwrite is handled in `AuRaBlockProccessor`

src/Nethermind/Nethermind.Consensus.AuRa/ContractRewriter.cs
Marchhill: I think dictionary is cleaner, you don't have to duplicate timestamp if you're doing multiple code overwrites at the same time (like we did for the shadowfork). also hopefully we won't have to do many code overwrites in future

src/Nethermind/Nethermind.Specs/GnosisForks/19_BalancerGnosis.cs
LukaszRozmej: ok how `GnosisSpecProvider` knows about `RewriteBytecodeTimestamp`? (when not parsing chainspec)

src/Nethermind/Nethermind.Consensus.AuRa/ContractRewriter.cs
LukaszRozmej: You could have dictionary in spec but convert it to array internally

src/Nethermind/Nethermind.Specs/GnosisForks/19_BalancerGnosis.cs
Marchhill: could probably get rid of it and just use Prague actually

Marchhill: it doesn't change the release spec so it doesn't need to

PR #9949, opened 2025-12-16
feat: Support Witness Generation

hudem1:
Add support for `debug_executionWitness` RPC endpoint.

We'll need that for Arbitrum project as well to be a validator.

## Changes

- _List the changes_

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Remarks

### DIFF
--- a/src/Nethermind/Nethermind.Evm/CacheCodeInfoRepository.cs
+++ b/src/Nethermind/Nethermind.Evm/CacheCodeInfoRepository.cs
@@ -112,7 +112,7 @@ public class CacheCodeInfoRepository : ICodeInfoRepository
 
         internal void Clear()
         {
-            foreach (ClockCache<ValueHash256, CodeInfo> cache in _codeCache._caches)
+            foreach (ClockCache<ValueHash256, CodeInfo> cache in _caches)
             {
                 cache.Clear();
             }
--- a/src/Nethermind/Nethermind.Evm.Benchmark/EvmOpcodesBenchmark.cs
+++ b/src/Nethermind/Nethermind.Evm.Benchmark/EvmOpcodesBenchmark.cs
@@ -313,7 +313,7 @@ public unsafe class EvmOpcodesBenchmark
     public void CleanupStack()
     {
         _stateProvider.Reset(resetBlockChanges: true);
-        CodeInfoRepository.Clear();
+        CacheCodeInfoRepository.Clear();
     }
 
     [IterationSetup]


src/Nethermind/Nethermind.Init/Modules/WorldStateModule.cs
```diff
@@ -73,6 +73,8 @@ dbFactory is not MemDbFactory
 
             // Most config actually done in factory. We just call `Build` and then get back components from its output.
             .AddSingleton<MainPruningTrieStoreFactory>() // This part is done separately so that triestore can be obtained in test.
+            .AddSingleton<IReadOnlyTrieStore>(ctx =>
+                ctx.Resolve<MainPruningTrieStoreFactory>().PruningTrieStore.AsReadOnly())
```
hudem1, line 77: Not sure it's okay to expose that even in readonly.
I need the trie store dependency (in readonly) to create my trie node capturing trie store wrapper in `WitnessGeneratingBlockProcessingEnvFactory`.

A possible alternative i think would be to get the worldStateManager through the DI (the manager contains a reference to the trie store) and add a function to that manager similar to `CreateOverridableWorldScope` to create a world scope using the witness capturing trie store, which I could then call in `WitnessGeneratingBlockProcessingEnvFactory.CreateScope()`

src/Nethermind/Nethermind.Consensus/Stateless/NoCacheCodeInfoRepository.cs
```diff
@@ -0,0 +1,116 @@
...
+using Nethermind.Evm.EvmObjectFormat;
+using Nethermind.Evm.State;
+
+namespace Nethermind.Consensus.Stateless;
+
+/// <summary>
+/// Same implementation as CodeInfoRepository but without cache so that it always goes through
+/// the world state, which captures touched bytecodes.
+/// </summary>
+public sealed class NoCacheCodeInfoRepository : ICodeInfoRepository
```
hudem1: This class is used for capturing bytecodes. Without it, I have to use the regular `CodeInfoRepository`, which has a static `private static readonly CodeLruCache _codeCache = new();`. Therefore, even if i spawn a new environment for witness recording request, the static cache persists and it already has some bytecodes in its cache, which don't get recorded by the `WitnessGeneratingWorldState` as not reached.

I initially implemented a wrapper code info repository wrapping/decorating the regular `CodeInfoRepository` and whenever it calls `GetCachedCodeInfo`, it also records that bytecode. But then, sometimes I believe some bytecodes could be gotten directly using `WorldState.GetCode()` without going through the code info repository. So, i'd need to have bytecode recording logic in 2 places, which I thought was not very clean. So, I opted for that no cache code info rep.

But a disadvantage is that using that no cache code info rep, if we need to update the regular CodeInfoRepository, this no cache class should also be updated...

src/Nethermind/Nethermind.Consensus/Stateless/WitnessGeneratingWorldState.cs
```diff
@@ -0,0 +1,297 @@
...
+
+    public (byte[][] Codes, byte[][] Keys) GetWitness()
+    {
+        // Keys should be ordered like: <address1><address2><slot1-address2><slot2-address2><address3><slot1-address3>
+        List<byte[]> keys = _storageSlots
+            .SelectMany(kvp =>
+                new[] { kvp.Key.Bytes }
+                .Concat(kvp.Value.Select(slot => slot.ToBigEndian()))
+            )
+            .ToList();
```
hudem1: Keys seem to be ordered that way, see [geth PR comment](https://github.com/ethereum/go-ethereum/pull/32216#issuecomment-3209287671).

Additionally, you'll see in my PR `keys` are not used anywhere.
It seems they are not really useful for now.
Geth doesn't have the `keys` recording implemented yet.
And Reth doesn't seem to be using them either, even if they record them.

hudem1: I cannot find any standard / source of truth for the `debug_executionWitness` explaining exactly the keys etc. Isn't there something like that for debug api ?

src/Nethermind/Nethermind.Consensus/Stateless/WitnessCapturingTrieStore.cs
```diff
@@ -0,0 +1,69 @@
...
+    public byte[]? LoadRlp(Hash256? address, in TreePath path, Hash256 hash, ReadFlags flags = ReadFlags.None)
+    {
+        byte[]? rlp = TryLoadRlp(address, in path, hash, flags);
+        if (rlp is null) throw new MissingTrieNodeException("Missing RLP node", address, path, hash);
+        return rlp;
+    }
+
+    public byte[]? TryLoadRlp(Hash256? address, in TreePath path, Hash256 hash, ReadFlags flags = ReadFlags.None)
+    {
+        byte[]? code = _nodeStorage.Get(address, in path, hash, flags) ?? baseStore.TryLoadRlp(address, in path, hash, flags);
```
damian-orzechowski: Just a minor thing - name `code` might be confusing here

src/Nethermind/Nethermind.Consensus/Stateless/WitnessGeneratingBlockProcessingEnvFactory.cs
```diff
@@ -0,0 +1,66 @@
...
+        WitnessCapturingTrieStore trieStore = new(readOnlyDbProvider.StateDb, readOnlyTrieStore);
+        IStateReader stateReader = new StateReader(trieStore, readOnlyDbProvider.CodeDb, logManager);
+        IWorldState worldState = new WorldState(trieStore, readOnlyDbProvider.CodeDb, logManager, null, true);
+
+        ILifetimeScope envLifetimeScope = rootLifetimeScope.BeginLifetimeScope((builder) => builder
+            .AddScoped<IStateReader>(stateReader)
+            .AddScoped<IWorldState>(worldState)
+            .AddScoped<WitnessCapturingTrieStore>(trieStore)
+            .AddScoped<IWitnessGeneratingBlockProcessingEnv>(builder =>
+                new WitnessGeneratingBlockProcessingEnv(
```
damian-orzechowski, line 53: Wouldn't this work without explicit `resolve` calls?

src/Nethermind/Nethermind.Init/Modules/WorldStateModule.cs
damian-orzechowski: I'm not sure this is necessary. Even when having access to `PruningTrieStore` or `MainPruningTrieStoreFactory` you can call `AsReadOnly` in your init code.

src/Nethermind/Nethermind.Consensus/Stateless/WitnessCapturingTrieStore.cs
```diff
@@ -0,0 +1,69 @@
+// SPDX-FileCopyrightText: 2024 Demerzel Solutions Limited
```
damian-orzechowski: Nit pick - probably copied SPDX

src/Nethermind/Nethermind.Consensus/Stateless/WitnessGeneratingBlockFinder.cs
```diff
@@ -0,0 +1,124 @@
...
+using System.Threading.Tasks;
+using Nethermind.Blockchain;
+using Nethermind.Blockchain.Find;
+using Nethermind.Core;
+using Nethermind.Core.Crypto;
+using Nethermind.Serialization.Rlp;
+
+namespace Nethermind.Consensus.Stateless;
+
+public class WitnessGeneratingBlockFinder(IBlockFinder inner, IBlockhashCache blockhashCache) : IBlockFinder, IBlockhashCache
```
damian-orzechowski: Is this only to support `BLOCKHASH` opcode?

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
```diff
@@ -33,13 +33,15 @@
...
 using Nethermind.State.OverridableEnv;
+using Nethermind.Consensus.Stateless;
 
 namespace Nethermind.Facade
 {
     [Todo(Improve.Refactor, "I want to remove BlockchainBridge, split it into something with logging, state and tx processing. Then we can start using independent modules.")]
     public class BlockchainBridge(
         IOverridableEnv<BlockchainBridge.BlockProcessingComponents> processingEnv,
         Lazy<ISimulateReadOnlyBlocksProcessingEnv> lazySimulateProcessingEnv,
+        IWitnessGeneratingBlockProcessingEnvFactory witnessGeneratingBlockProcessingEnvFactory,
```
damian-orzechowski: Should it be lazy as the `simulate` env?

src/Nethermind/Nethermind.Consensus/Stateless/WitnessGeneratingBlockProcessingEnvFactory.cs
hudem1: I tried and we definitely need at least the `.AddScoped(IWorldState)` when resolving `builder.Resolve<IRewardCalculator>()`, which uses `.AddScoped<IRewardCalculator, IRewardCalculatorSource, ITransactionProcessor>((rewardSource, txP) => rewardSource.Get(txP))` in `BlockProcessingModule`.

src/Nethermind/Nethermind.Consensus/Stateless/WitnessGeneratingBlockFinder.cs
hudem1: yes from what I've read and understand, we can fetch headers only through BLOCKHASH opcode.
And we always have to include the parent one as well

src/Nethermind/Nethermind.Init/Modules/WorldStateModule.cs
hudem1: I tried initially but `MainPruningTrieStoreFactory` from Nethermind.Init is not accessible from `Nethermind.Consensus`. And it seems we don't have access to the PruningTrieStore outside of `MainPruningTrieStoreFactory` or `WorldStateManager`

src/Nethermind/Nethermind.Consensus/Stateless/WitnessGeneratingBlockFinder.cs
hudem1: I realized I could simplify the logic, pushed new logic for capturing headers

src/Nethermind/Nethermind.Consensus/Stateless/WitnessCollector.cs
```diff
@@ -0,0 +1,41 @@
...
+
+            (byte[][] codes, byte[][] keys) = worldState.GetWitness();
+
+            return new Witness()
+            {
+                Headers = headerFinder.GetWitnessHeaders(parentHeader.Hash),
+                Codes = codes,
+                State = trieStore.TouchedNodesRlp,
+                Keys = keys
+            };
```
smartprogrammer93: seems like the witness here is missing something? can you really prove the prestate in the MPT using the state root of the parent with only this info? i think we are missing the proof here.

src/Nethermind/Nethermind.Consensus/Stateless/NoCacheCodeInfoRepository.cs
AnkushinDaniil: What if wrap `NoCacheCodeInfoRepository` instead?

src/Nethermind/Nethermind.JsonRpc/Modules/DebugModule/IDebugRpcModule.cs
```diff
@@ -123,4 +124,7 @@ ResultWrapper<IReadOnlyList<SimulateBlockResult<GethLikeTxTrace>>> debug_simulat
 
     [JsonRpcMethod(Description = "Executes a list of bundles of transactions without creating transactions on the blockchain and returns their traces", IsImplemented = true, IsSharable = false)]
     ResultWrapper<IEnumerable<IEnumerable<GethLikeTxTrace>>> debug_traceCallMany(TransactionBundle[] bundles, BlockParameter? blockParameter = null, GethTraceOptions? options = null);
+
+    [JsonRpcMethod(Description = "Reprocess the existing block whose parameters are specified by blockParameter and return the generated execution witness")]
```
LukaszRozmej: @rubo ?

src/Nethermind/Nethermind.JsonRpc.Test/Modules/DebugRpcModuleTests.ExecutionWitness.cs
```diff
@@ -0,0 +1,126 @@
...
+            .WithGasLimit(200_000)
+            .SignedAndResolved(TestItem.PrivateKeyA)
+            .TestObject;
+        await blockchain.AddBlock(callTx);
+
+        Block? block = blockchain.BlockTree.FindBlock(blockNumber, BlockTreeLookupOptions.RequireCanonical);
+        block.Should().NotBeNull();
+        block!.Hash.Should().NotBeNull();
+
+        Hash256 parentHash = block.Header.ParentHash!;
```
LukaszRozmej: Can you break it down a bit to helper methods?

src/Nethermind/Nethermind.JsonRpc.Test/Modules/DebugRpcModuleTests.ExecutionWitness.cs
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        Witness witness = response.Should().BeOfType<JsonRpcSuccessResponse>()
+            .Which.Result.Should().BeOfType<Witness>()
+            .Subject;
+
+        witness.Headers.Should().NotBeNull();
+        witness.Headers.Length.Should().BeGreaterThan(0);
+        witness.State.Should().NotBeNull();
+        witness.State.Length.Should().BeGreaterThan(0);
```
LukaszRozmej:
```suggestion
        witness.Headers.Should().NotBeEmpty();
        witness.State.Should().NotBeEmpty();
```

src/Nethermind/Nethermind.JsonRpc.Test/Modules/DebugRpcModuleTests.ExecutionWitness.cs
```diff
@@ -0,0 +1,126 @@
...
+            blockchain.SpecProvider,
+            Always.Valid,
+            blockchain.LogManager);
+
+        using var scope = statelessEnv.WorldState.BeginScope(parent!);
+        (Block processed, _) = statelessEnv.BlockProcessor.ProcessOne(
+            block,
+            ProcessingOptions.ReadOnlyChain,
+            NullBlockTracer.Instance,
+            blockchain.SpecProvider.GetSpec(block.Header));
```
LukaszRozmej: Move to reprocess with witness helper method?

src/Nethermind/Nethermind.Consensus/Stateless/NoCacheCodeInfoRepository.cs
LukaszRozmej: if possible `CodeInfoRepository` (could be renamed to `CacheCodeInfoRepository`) should wrap `NoCacheCodeInfoRepository` (could be renamed to `CodeInfoRepository`).

src/Nethermind/Nethermind.Consensus/Stateless/NoCacheCodeInfoRepository.cs
```diff
@@ -0,0 +1,116 @@
...
+using Nethermind.Evm.CodeAnalysis;
+using Nethermind.Evm.EvmObjectFormat;
+using Nethermind.Evm.State;
+
+namespace Nethermind.Consensus.Stateless;
+
+/// <summary>
+/// Same implementation as CodeInfoRepository but without cache so that it always goes through
+/// the world state, which captures touched bytecodes.
+/// </summary>
```
LukaszRozmej: Those are generally `<remarks> ` not `<summary>`, you are using the docs a bit wrong in multiple files.

src/Nethermind/Nethermind.Consensus/Stateless/WitnessCapturingTrieStore.cs
```diff
@@ -0,0 +1,69 @@
...
+    public void Dispose()
+    {
+        baseStore.Dispose();
+    }
+
+    public TrieNode FindCachedOrUnknown(Hash256? address, in TreePath path, Hash256 hash)
+    {
+        TrieNode node = baseStore.FindCachedOrUnknown(address, in path, hash);
+        if (node.NodeType != NodeType.Unknown)
+            _rlpCollector.TryAdd(node.Keccak, node.FullRlp.Span.ToArray());
```
LukaszRozmej: If it is already an array you save on allocation:

```suggestion
            _rlpCollector.TryAdd(node.Keccak, node.FullRlp.ToArray()!);
```

src/Nethermind/Nethermind.Consensus/Stateless/WitnessGeneratingHeaderFinder.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            _lowestRequestedHeader = blockNumber;
+        }
+        return header;
+    }
+
+    public byte[][] GetWitnessHeaders(Hash256 parentHash)
+    {
+        HeaderDecoder decoder = new();
+        List<byte[]> headers = new();
```
LukaszRozmej: You can use `ArrayPoolListRef` here to avoid some allocations.

src/Nethermind/Nethermind.Consensus/Stateless/WitnessGeneratingWorldState.cs
```diff
@@ -0,0 +1,297 @@
...
+
+        HashSet<byte[]> codes = new(_bytecodes.Values, Bytes.EqualityComparer);
+
+        return (codes.ToArray(), keys.ToArray());
```
LukaszRozmej: Lot of allocations, not all necessary

src/Nethermind/Nethermind.Init/Modules/WorldStateModule.cs
LukaszRozmej: @asdacap maybe you have better idea?

src/Nethermind/Nethermind.Consensus/Stateless/WitnessCollector.cs
hudem1: The State fields contains all the trie nodes accessed during execution as well during state root recomputation (might access new nodes not accessed during execution).

So, State fields also contain intermediate nodes on the path from the root to the leaves, making up for the proof for each leaf. That's why, stateless re-execution still works even though it traverses the trie and accesses intermediate nodes on the path from the root to the leaves

src/Nethermind/Nethermind.Init/Modules/WorldStateModule.cs
asdacap: No easy way. Perhaps extending IWorldStateScopeProvider with trie nodes read/writes.

asdacap: So, technically, you can rewrote the whole scope provider as long as you have trie nodes/rlp read access. So adding that to the IWorldStateScope then writing a custom.... whole thing... would allow you to not use IReadOnlyTrieStore.

hudem1: Okay sounds a bit overkill ? let's do that as a refactor if we think it's really worth it then because we need this PR for arbitrum for now if that's okay 🙏

src/Nethermind/Nethermind.JsonRpc/Modules/DebugModule/IDebugRpcModule.cs
rubo:
```suggestion
    [JsonRpcMethod(Description = "Reprocesses the existing block with the parameters specified and returns the generated execution witness.")]
```

src/Nethermind/Nethermind.Consensus/Stateless/NoCacheCodeInfoRepository.cs
hudem1: I realized there was an other file also named the same way `CachedCodeInfoRepository`.
It seems to be caching precompile results (with a block as its scope), whereas my `CacheCodeInfoRepository` (coming from the pre-existing `CodeInfoRepository`) caches bytecode objects (with process-wide scope). But the naming are very similar, should I rename the existing `CachedCodeInfoRepository` to `PrecompileCachedCodeInfoRepository` (something related to precompiles) ?

src/Nethermind/Nethermind.Consensus/Stateless/WitnessGeneratingHeaderFinder.cs
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    public byte[][] GetWitnessHeaders(Hash256 parentHash)
+    {
+        HeaderDecoder decoder = new();
+        using ArrayPoolListRef<byte[]> headers = new();
+
+        Hash256 currentHash = parentHash;
+        BlockHeader childHeader = inner.Get(currentHash) ?? throw new ArgumentException($"Parent {currentHash} is not found");
+        headers.Add(decoder.Encode(childHeader).Bytes);
```
LukaszRozmej: allocates

src/Nethermind/Nethermind.Consensus/Stateless/WitnessGeneratingHeaderFinder.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        if (_lowestRequestedHeader is not null)
+        {
+            for (long i = childHeader.Number - 1; i >= _lowestRequestedHeader.Value; i--)
+            {
+                currentHash = childHeader.ParentHash;
+                childHeader = inner.Get(currentHash, i) ?? throw new ArgumentException($"Unable to get requested header at hash {currentHash} and number {i} during witness generation");
+                headers.Add(decoder.Encode(childHeader).Bytes);
```
LukaszRozmej: allocates

src/Nethermind/Nethermind.Consensus/Stateless/WitnessGeneratingHeaderFinder.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            _lowestRequestedHeader = blockNumber;
+        }
+        return header;
+    }
+
+    public byte[][] GetWitnessHeaders(Hash256 parentHash)
+    {
+        HeaderDecoder decoder = new();
+        using ArrayPoolListRef<byte[]> headers = new();
```
LukaszRozmej: better to use some starting capacity

src/Nethermind/Nethermind.Consensus/Stateless/WitnessGeneratingWorldState.cs
```diff
@@ -0,0 +1,334 @@
...
+                keys[i++] = slot.ToBigEndian();
+        }
+
+        return new Witness()
+        {
+            Codes = _bytecodes.Values.ToArray(),
+            State = stateNodes.ToArray(),
+            Keys = keys,
+            Headers = headerFinder.GetWitnessHeaders(parentHeader.Hash)
+        };
```
LukaszRozmej: Could we use ArrayPoolList's more to avoid so many allocations?

src/Nethermind/Nethermind.Evm/CacheCodeInfoRepository.cs
```diff
@@ -0,0 +1,128 @@
...
+using Nethermind.Core.Caching;
+using Nethermind.Core.Crypto;
+using Nethermind.Core.Specs;
+using Nethermind.Evm.CodeAnalysis;
+using Nethermind.Evm.EvmObjectFormat;
+using Nethermind.Evm.State;
+
+namespace Nethermind.Evm;
+
+public class CacheCodeInfoRepository(IWorldState worldState, IPrecompileProvider precompileProvider) : CodeInfoRepository(worldState, precompileProvider)
```
LukaszRozmej: Consider composition rather than inheritance?

hudem1: Done

src/Nethermind/Nethermind.Consensus/Stateless/WitnessGeneratingWorldState.cs
hudem1: Done

src/Nethermind/Nethermind.Consensus/Stateless/WitnessGeneratingWorldState.cs
LukaszRozmej: If we are doing pooling this could be `PooledSet` from https://github.com/jtmueller/Collections.Pooled

src/Nethermind/Nethermind.Consensus/Stateless/Witness.cs
```diff
@@ -13,12 +13,20 @@
...
+public class Witness : IDisposable
 {
-    public byte[][] Codes;
-    public byte[][] State;
-    public byte[][] Keys;
-    public byte[][] Headers;
+    public required IOwnedReadOnlyList<byte[]> Codes { get; init; }
+    public required IOwnedReadOnlyList<byte[]> State { get; init; }
+    public required IOwnedReadOnlyList<byte[]> Keys { get; init; }
+    public required IOwnedReadOnlyList<byte[]> Headers { get; init; }
```
LukaszRozmej, line 19: In theory some of the inner arrays also could be pooled, but that's a harder task

src/Nethermind/Nethermind.Consensus/Stateless/WitnessGeneratingWorldState.cs
hudem1: Done

src/Nethermind/Nethermind.Consensus/Stateless/Witness.cs
hudem1: I'd be happy to try and do that but could it be as a refactor that i'd do next week ? Because i think we are expected to merge the witness generation related PRs this week for Arbitrum and i still have a few things to implement there

src/Nethermind/Nethermind.Consensus/Stateless/WitnessGeneratingHeaderFinder.cs
```diff
@@ -0,0 +1,45 @@
+using System;
```
rubo, line 4: Missing file header:

```suggestion
// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
// SPDX-License-Identifier: LGPL-3.0-only

using System;
```

tools/StatelessExecution/SetupCli.cs
rubo, line 1: This is gone, so when merging conflicts, just take the master version, i.e. delete.

PR #9952, opened 2025-12-16
Optimise MemDb for ZK

LukaszRozmej:
## Changes

- Adds `ZK` compiler flag
- Makes MemDb dependent on `Dictionary` rather than `ConcurrentDictionary` for ZK
- Removes some old garbage

- [x] Refactoring

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Db/MemDb.cs
+++ b/src/Nethermind/Nethermind.Db/MemDb.cs
@@ -18,8 +18,14 @@ namespace Nethermind.Db
         public long ReadsCount { get; private set; }
         public long WritesCount { get; private set; }
 
+#if ZK
+        private readonly Dictionary<byte[], byte[]?> _db;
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Db/MemDb.cs
```diff
@@ -80,58 +90,39 @@ public virtual byte[]? this[ReadOnlySpan<byte> key]
...
+#endif
         }
 
-        public bool KeyExists(ReadOnlySpan<byte> key)
-        {
-            return _spanDb.ContainsKey(key);
-        }
-
-        public IDb Innermost => this;
+        public bool KeyExists(ReadOnlySpan<byte> key) => _spanDb.ContainsKey(key);
```
maximmenshikov, line 100: is it a newly enforced style? because normally it is better not to introduce style changes unless the line is changed

maximmenshikov (approved):

PR #9953, opened 2025-12-17
fix: Correct DistributeFunds method spelling

zeroprooff:
Corrected method name from DitributeFunds to DistributeFunds in FundsDistributor and updated the call in SetupCli.


LukaszRozmej:
not meaningful

PR #9956, opened 2025-12-17
Add background scheduler stats

LukaszRozmej:
## Changes

- Adds per-request-type queue-depth stats to the background scheduler, reported in the queue-full drop warning and in the block-processing debug log.
- `IBackgroundTaskScheduler.TryScheduleTask` replaces the free-text `source` argument with a `TReq : IBackgroundTaskRequest<TReq>` constraint, so the reported name is derived from the request type instead of being passed at every call site.

#### Requires testing

- [x] Yes

- [x] Yes

- [x] Yes

`IBackgroundTaskScheduler.TryScheduleTask` changed shape: it gained a `where TReq : notnull, IBackgroundTaskRequest<TReq>` constraint and dropped the optional `string? source` argument. Out-of-tree plugins that call or implement this interface need updating — a request type now declares its own reporting bucket:

```csharp
private readonly record struct MyRequest(...) : IBackgroundTaskRequest<MyRequest>;
```

`TaskId` has a default implementation, so a plain request type needs no body. A wrapper request should override it with `BackgroundTaskTypeId<TWrapped>.Id` so stats report the wrapped type rather than the wrapper's generic name.


src/Nethermind/Nethermind.Network/P2P/Utils/BackgroundTaskSchedulerWrapper.cs
```diff
@@ -21,30 +21,48 @@ namespace Nethermind.Network.P2P.Utils;
...
+    }
+
+    private readonly record struct RequestValueTaskWithResponse<TReq, TRes>(in TReq Request, Func<TReq, CancellationToken, ValueTask<TRes>> FulfillFunc) : IToString
+    {
+        public override string ToString() => typeof(TReq).FullName!;
+    }
+
+    private readonly record struct RequestValueTask<TReq>(in TReq Request, Func<TReq, CancellationToken, ValueTask> BackgroundTask)
+    {
+        public override string ToString() => Request is IToString ? Request.ToString()! : typeof(TReq).FullName!;
```
benaadams: Is very long key (includes Namespace)

src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
```diff
@@ -41,6 +41,7 @@ public class BackgroundTaskScheduler : IBackgroundTaskScheduler, IAsyncDisposabl
     private readonly IChainHeadInfoProvider _headInfo;
     private readonly int _capacity;
     private long _queueCount;
+    private readonly ConcurrentDictionary<string, int> _stats = new();
```
benaadams: Not keen on the string as key

src/Nethermind/Nethermind.Network/P2P/Utils/BackgroundTaskSchedulerWrapper.cs
LukaszRozmej: yes but there are eth/66 stuff so it is kind of by design?

src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
LukaszRozmej: any other ideas with low implementation cost and upkeep?
Is Type better?

src/Nethermind/Nethermind.Network/P2P/Utils/BackgroundTaskSchedulerWrapper.cs
flcl42:
```
public interface INamedBackgroundTask<TTask> where TTask : notnull, INamedBackgroundTask<TTask>
{
    static abstract BackgroundTaskKind Kind { get; }
}
public enum BackgroundTaskKind
{
    GetAccountRangeMessage,
    GetBlockBodiesMessage,
    GetBlockHeadersMessage,
    GetByteCodesMessage,
    GetNodeDataMessage,
    GetPooledTransactionsMessage,
    GetReceiptsMessage,
    GetStorageRangeMessage,
    GetTrieNodesMessage,
    HistoryPruner,
    TransactionsMessage,
}
```
as an idea

src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
LukaszRozmej: changed to int

flcl42 (approved):
IBackgroundTaskRequest/BackgroundTaskTypeId potentially could have more generic names to be reused

AnkushinDaniil (approved):

src/Nethermind/Nethermind.Consensus/Scheduler/IBackgroundTaskRequest.cs
```diff
@@ -0,0 +1,74 @@
+// SPDX-FileCopyrightText: 2026 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System;
+using System.Threading;
+
+namespace Nethermind.Consensus.Scheduler;
+
+/// <summary>
+/// Marker for request types scheduled via <see cref="IBackgroundTaskScheduler"/>.
```
Marchhill, line 10: long comments

src/Nethermind/Nethermind.Consensus/Scheduler/IBackgroundTaskRequest.cs
```diff
@@ -0,0 +1,74 @@
...
+}
+
+/// <summary>
+/// Maps task ids back to <see cref="Type"/> for diagnostics (drop-warning rendering, stats dumps).
+/// Ids are produced by <see cref="Interlocked.Increment(ref int)"/>, so they are dense from 0 upward.
+/// </summary>
+public static class BackgroundTaskTypeRegistry
+{
+    private static int s_next = -1;
+    private static Type?[] s_types = new Type?[16];
```
Marchhill, line 38: why s_?

Marchhill (approved):

src/Nethermind/Nethermind.Network/P2P/Utils/BackgroundTaskSchedulerWrapper.cs
LukaszRozmej: Fixed. Keys are now the *wrapped* message type, resolved via `Type.Name`, so no namespace and no generic arity: `GetBlockBodiesMessage`, `GetPooledTransactionsMessage`, `TransactionsRequest`, ... — the same names the removed `source` argument produced.

This was actually broken before this push: the wrapper structs registered `typeof(SyncServeTaskRequest<TReq, TRes>)`, and `Type.Name` of a closed generic is ``SyncServeTaskRequest`2`` for *every* instantiation. So the stats dump read `(SyncServeTaskRequest`2: 5), (SyncServeTaskRequest`2: 3), ...` — several distinct buckets under one indistinguishable name. Now each wrapper does `TaskId => BackgroundTaskTypeId<TReq>.Id`, which also merges the sync-serve and background-task wrappers over the same message into one bucket.

@flcl42 re the explicit `BackgroundTaskKind` enum — I went with an auto-assigned id instead: it needs no manual upkeep when a task type is added (including from a plugin, which an enum in Nethermind.Consensus can't cover), and it yields exactly the names in your list. The cost is one `Type` stored per id in a fixed-size table, read only when a diagnostic string is rendered.

src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
LukaszRozmej: Yes, `int` now — and the storage behind it changed too. `ConcurrentDictionary.AddOrUpdate` takes a bucket `Monitor` lock on the update path, and this was calling it on *both* the enqueue and the dequeue path, i.e. a new lock on the P2P hot path.

Since the ids are dense from 0, the counters are now a flat `int[64]` indexed by task id, updated with a single `Interlocked.Add`. Micro-benchmark, 4M increment/decrement pairs over 8 keys:

```
threads     ConcurrentDictionary         int[]     speedup
1                      109.2 ms       7.7 ms       14.3x
2                       87.0 ms      21.4 ms        4.1x
4                       67.1 ms      55.3 ms        1.2x
8                      133.3 ms      57.7 ms        2.3x
```

Never slower, ~27ns -> ~2ns uncontended. Against the no-stats baseline the remaining cost is ~2ns per enqueue on a path that already does an interlocked increment plus a prioritized-channel write.

Also fixed while in here: the increment now happens *before* the channel write (a consumer could previously decrement ahead of it and undercount), and is undone if the write fails.

src/Nethermind/Nethermind.Consensus/Scheduler/IBackgroundTaskRequest.cs
LukaszRozmej: Trimmed — the two `<remarks>` blocks are down to the non-obvious parts (why a wrapper should point at the type it wraps, and what happens past the id cap).

src/Nethermind/Nethermind.Consensus/Scheduler/IBackgroundTaskRequest.cs
LukaszRozmej: Gone — the growable table those statics backed is replaced by a fixed-size one, so there is no lock and no mutable static array left. That also removed a race: `Register` incremented the id, grew the array, then wrote, so a concurrent grow could `Array.Copy` before another thread's write landed and silently drop a type (rendered as `unknown`).

LukaszRozmej:
@claude review

src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
```diff
@@ -238,16 +246,46 @@ public bool TryScheduleTask<TReq>(TReq request, Func<TReq, CancellationToken, Ta
...
+            int count = Volatile.Read(ref _stats[id]);
+            if (count > 0)
+            {
+                if (builder.Length > 0) builder.Append(", ");
+                builder.Append('(').Append(BackgroundTaskTypeRegistry.GetName(id)).Append(": ").Append(count).Append(')');
+            }
+        }
+
+        return builder.ToString();
+    }
```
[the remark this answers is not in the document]
LukaszRozmej: Confirmed and fixed in f4dba4b. `V62.Messages.GetBlockHeadersMessage` and `V66.Messages.GetBlockHeadersMessage` are indeed distinct types with the same `Type.Name`, and both reach `HandleInBackground`.

I didn't merge them in `FormatStats` though — merging makes the output unambiguous but throws away the eth/62-vs-eth/66 distinction, which is exactly what you'd want to see when one of them saturates the queue. Instead `GetName` qualifies a name only when another *registered* type shares it:

```csharp
public static string? GetName(int id)
{
    Type? type = (uint)id < MaxTaskTypes ? Volatile.Read(ref Types[id]) : null;
    if (type is null) return null;

    for (int other = 0; other < MaxTaskTypes; other++)
    {
        Type? candidate = Volatile.Read(ref Types[other]);
        if (candidate is not null && candidate != type && candidate.Name == type.Name)
        {
            return type.FullName;
        }
    }

    return type.Name;
}
```

Common names stay short (which is what benaadams asked for above), colliding ones get the namespace. O(64) per render on a path that runs at most once per 10 s. `GetStats` then needs no merge and its keys are unique. Regression test in `BackgroundTaskTypeRegistryTests`.

src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/V72/SparseBlobPoolPeerRegistry.cs
```diff
@@ -2626,7 +2624,11 @@ private enum InFlightReservationResult
 
     private readonly record struct TrackedStateKey(ValueHash256 Hash, long Revision);
 
-    private readonly record struct ScheduledRegistryRequest(SparseBlobPoolPeerRegistry Registry);
+    private readonly record struct ScheduledRegistryRequest(SparseBlobPoolPeerRegistry Registry)
```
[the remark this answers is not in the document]
LukaszRozmej: Good catch, fixed in f4dba4b — split into `CustodyUpdateRequest` and `MaintenanceSweepRequest`. This was a real regression against the two `source:` strings, and the `TaskId` boilerplate is gone now, so it costs two lines:

```csharp
private readonly record struct CustodyUpdateRequest(SparseBlobPoolPeerRegistry Registry)
    : IBackgroundTaskRequest<CustodyUpdateRequest>;

private readonly record struct MaintenanceSweepRequest(SparseBlobPoolPeerRegistry Registry)
    : IBackgroundTaskRequest<MaintenanceSweepRequest>;
```

src/Nethermind/Nethermind.Consensus/Scheduler/IBackgroundTaskRequest.cs
```diff
@@ -0,0 +1,60 @@
...
+    internal static int Register(Type type)
+    {
+        int id = Interlocked.Increment(ref _lastId);
+        if ((uint)id < MaxTaskTypes)
+        {
+            Types[id] = type;
+        }
+
+        return id;
+    }
```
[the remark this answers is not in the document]
LukaszRozmej: Both taken in f4dba4b.

Overflow: added `Debug.Assert(id < MaxTaskTypes, ...)` in `Register`. A runtime warn isn't available there (static registry, no logger), but the assert fires for whoever adds the 65th type in a debug build, which is the person who needs to see it.

Publication: the slot is now written with `Volatile.Write` and `GetName` reads it with `Volatile.Read`, returning `null` for an unclaimed id. I also dropped `Count` — deriving the bound from `_lastId` was the part that let a caller walk to an id whose slot hadn't been published yet. `GetStats` now scans the whole table and skips unpublished slots, so a reader sees a type or nothing.

src/Nethermind/Nethermind.Consensus/Scheduler/IBackgroundTaskRequest.cs
```diff
@@ -0,0 +1,60 @@
...
+/// </summary>
+/// <remarks>
+/// Implement as <c>public static int TaskId =&gt; BackgroundTaskTypeId&lt;TNamed&gt;.Id;</c>. A wrapper request
+/// should point at the type it wraps, so that stats read as the underlying message name instead of the
+/// wrapper's generic type name, which is identical for every instantiation.
+/// </remarks>
+public interface IBackgroundTaskRequest<T> where T : IBackgroundTaskRequest<T>
+{
+    static abstract int TaskId { get; }
+}
```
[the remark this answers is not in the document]
LukaszRozmej: Agreed, done in f4dba4b. `static virtual int TaskId => BackgroundTaskTypeId<T>.Id;` is the default now, and `HistoryPrunerRequest`, `DetectRequest`, `BenchmarkRequest`, `TestRequest` plus the two new `SparseBlobPoolPeerRegistry` requests are each a single line with no body. The wrapper structs in `BackgroundTaskSchedulerWrapper` still override with `BackgroundTaskTypeId<TReq>.Id`, and the `<remarks>` now covers just that case — plus the converse, that one request type used for two unrelated workloads should be split (which is exactly what the blob-registry comment above turned out to be).

src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
```diff
@@ -37,6 +38,7 @@ public class BackgroundTaskScheduler : IBackgroundTaskScheduler, IAsyncDisposabl
     private readonly Lock _blockProcessingLock = new();
     private int _activeBlockProcessingBranches;
     private long _queueCount;
+    private readonly int[] _stats = new int[BackgroundTaskTypeRegistry.MaxTaskTypes];
```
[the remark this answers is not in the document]
LukaszRozmej: Measured it rather than guessing, and I'm leaving the layout packed.

The missing term in the earlier table is that **every** enqueue already does `Interlocked.Increment(ref _queueCount)` — a single counter shared by all task types — so one contended line is paid whether or not stats exist. Adding that baseline and a 64-byte-strided variant (4M ops, each thread driving a *different* task type, i.e. the worst case for false sharing):

```
threads    queueCount only    + packed stats   + strided stats
1                   7.6 ms           24.6 ms           24.3 ms
2                  26.1 ms           31.8 ms           28.0 ms
4                  27.0 ms          122.7 ms          103.0 ms
8                  41.8 ms           96.1 ms          128.5 ms
```

Striding helps at 4 threads (~16%) and hurts at 8, i.e. it's inside the noise of this rig rather than a consistent win. Given that, 4 KiB per scheduler plus `id * Stride` arithmetic in three places isn't worth it. Worth revisiting if a real profile ever shows the stats line hot — the numbers above are a tight loop with zero work between enqueues, whereas the real path spends microseconds in the prioritized-channel write and `Activity` rent between touches.

src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
```diff
@@ -272,7 +310,24 @@ public async ValueTask DisposeAsync()
...
+        Dictionary<string, int> stats = new(count);
+        for (int id = 0; id < count; id++)
+        {
+            string name = BackgroundTaskTypeRegistry.GetName(id);
+            // Distinct types can share a simple name, so buckets are merged rather than overwritten
+            stats[name] = (stats.TryGetValue(name, out int existing) ? existing : 0) + Volatile.Read(ref _stats[id]);
+        }
+
+        return stats;
+    }
```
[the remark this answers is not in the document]
LukaszRozmej: Fair. `GetStats()` is now `internal` (`Nethermind.Consensus.Test` already has `InternalsVisibleTo`), so it adds no public surface, and I took your suggestion for the consumer:

```csharp
if (_logger.IsDebug) _logger.Debug($"Block processing starting, background queue depth: {depth} [{FormatStats()}]");
```

It uses `FormatStats()` rather than `GetStats()` so the Debug path allocates one string instead of a dictionary, and reads identically to the drop warning.

LukaszRozmej:
@claude re-review

src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
```diff
@@ -96,7 +98,7 @@ private void BranchProcessorOnBranchesProcessing(object? sender, BlocksProcessin
             if (_activeBlockProcessingBranches++ == 0)
             {
                 long depth = Volatile.Read(ref _queueCount);
-                if (_logger.IsDebug) _logger.Debug($"Block processing starting, background queue depth: {depth}");
+                if (_logger.IsDebug) _logger.Debug($"Block processing starting, background queue depth: {depth} [{FormatStats()}]");
```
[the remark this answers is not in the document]
LukaszRozmej: Fixed in a705181, and the important half of this isn't the microseconds — it's that the string was being built *before* `_blockProcessorCancellationTokenSource.Cancel()`, which is the thing that actually pauses background work for block processing. Delaying that under the lock every worker takes is the wrong order regardless of cost.

So both parts are addressed:

1. The lock now covers only the state change; the depth is captured inside and the Debug line is emitted after it:

```csharp
depth = Volatile.Read(ref _queueCount);
_blockProcessingDoneSignal = new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
_blockProcessorCancellationTokenSource.Cancel();
}

if (_logger.IsDebug) _logger.Debug($"Block processing starting, background queue depth: {depth} [{FormatStats()}]");
```

2. Took the memoization rather than leaving the quadratic factor: names are resolved once, when the type registers, so `GetName` is `Volatile.Read(ref Names[id])` and the ambiguity scan runs at most 64 times over the process lifetime instead of on every render. Registration is now serialised too, which as a bonus makes the collision fixup deterministic when two colliding types register concurrently — before, both could scan and miss each other and keep their short names.

src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
```diff
@@ -272,7 +310,24 @@ public async ValueTask DisposeAsync()
...
+    /// Snapshot of currently queued task counts, keyed by the request type's reported name.
+    /// </summary>
+    internal IReadOnlyDictionary<string, int> GetStats()
+    {
+        Dictionary<string, int> stats = [];
+        for (int id = 0; id < _stats.Length; id++)
+        {
+            if (BackgroundTaskTypeRegistry.GetName(id) is string name)
+            {
+                stats[name] = Volatile.Read(ref _stats[id]);
```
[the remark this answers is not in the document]
LukaszRozmej: Taken in a705181 — it's one line on a path that runs at most once per 10 s, and I'd rather the method be total than rely on `GetName` never colliding.

Agreed on the residual case being cross-assembly `FullName` equality; the ambiguity check escalates `Name` -> `FullName` and stops there. I left `FormatStats` per-id: in that case it would print the same label twice, which is visible and harmless, whereas silently dropping a bucket from `GetStats` was not. Escalating a second tier (`AssemblyQualifiedName`, or the id) to make names provably unique felt like more machinery than a case that needs a plugin to duplicate an in-tree namespace *and* type name warrants.

src/Nethermind/Nethermind.Consensus/Scheduler/IBackgroundTaskRequest.cs
```diff
@@ -0,0 +1,84 @@
...
+    private static int _lastId = -1;
+
+    internal static int Register(Type type)
+    {
+        int id = Interlocked.Increment(ref _lastId);
+        Debug.Assert(id < MaxTaskTypes, $"More than {MaxTaskTypes} background task types; raise {nameof(MaxTaskTypes)} or the extra ones go unreported.");
+
+        if ((uint)id < MaxTaskTypes)
+        {
+            // Published so a concurrent GetName sees either the type or nothing, never a torn slot
```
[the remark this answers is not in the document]
LukaszRozmej: You're right, and per AGENTS.md that's worth fixing rather than leaving. Reference-typed element writes are atomic on every runtime we target, so "torn slot" describes something that cannot happen; the guarantee is ordering against `Interlocked.Increment(ref _lastId)` having already published the id.

Reworded in a705181 — the comment now sits on the name table, since names are resolved at registration:

```csharp
Types[id] = type;
// Ordered after the type's name is resolved, and after the writes above, so a concurrent
// GetName sees either a fully resolved name or nothing
Volatile.Write(ref Names[id], name);
```

LukaszRozmej:
@claude re-review

src/Nethermind/Nethermind.Consensus/Scheduler/IBackgroundTaskRequest.cs
```diff
@@ -0,0 +1,87 @@
...
+            for (int other = 0; other < MaxTaskTypes; other++)
+            {
+                // Simple names are not unique — eth/62 and eth/66 both declare a GetBlockHeadersMessage —
+                // so a collision qualifies both this type and the one it collides with
+                if (Types[other] is Type candidate && candidate.Name == name)
+                {
+                    name = type.FullName!;
+                    Volatile.Write(ref Names[other], candidate.FullName);
+                }
+            }
```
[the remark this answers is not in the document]
LukaszRozmej: Right on both counts — the loop was already find-first, because reassigning `name` inside it made every later `candidate.Name == name` compare fail, and nothing in the code said so. Fixed in 11a2ba2: compare against `type.Name` (which doesn't move) and `break`, with the invariant written down.

```csharp
// Simple names are not unique — eth/62 and eth/66 both declare a GetBlockHeadersMessage —
// so a collision qualifies both this type and the one it collides with. One match is
// enough: any further type of that name was already qualified when it registered.
if (Types[other] is Type candidate && candidate.Name == type.Name)
{
    name = type.FullName!;
    Volatile.Write(ref Names[other], candidate.FullName);
    break;
}
```

src/Nethermind/Nethermind.Consensus.Test/Scheduler/BackgroundTaskTypeRegistryTests.cs
```diff
@@ -0,0 +1,39 @@
...
+        using (Assert.EnterMultipleScope())
+        {
+            Assert.That(BackgroundTaskTypeRegistry.GetName(first), Is.EqualTo(typeof(CollidingRequest).FullName));
+            Assert.That(BackgroundTaskTypeRegistry.GetName(second), Is.EqualTo(typeof(Other.CollidingRequest).FullName));
+        }
+    }
+
+    [Test]
+    public void Unclaimed_id_has_no_name() =>
+        Assert.That(BackgroundTaskTypeRegistry.GetName(BackgroundTaskTypeRegistry.MaxTaskTypes), Is.Null);
```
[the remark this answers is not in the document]
LukaszRozmej: Fixed in 11a2ba2 — `[Values(MaxTaskTypes - 1, MaxTaskTypes)]` so one case takes the in-range unclaimed slot and the other the range guard.

Also took the related note from your summary: the `?? "unknown"` in `FormatStats` was unreachable, since `AddToStats` refuses ids past the cap and so a non-zero count implies a registered name. It now pattern-matches the name the same way `GetStats` does, which drops the dead fallback without losing the null check. The fallback on the drop-warn path stays — that one is reachable.

LukaszRozmej:
@claude re-review

src/Nethermind/Nethermind.Network/P2P/Utils/BackgroundTaskSchedulerWrapper.cs
```diff
@@ -80,12 +80,12 @@ internal bool TryScheduleSyncServe<TReq, TRes>(TReq request, Func<TReq, Cancella
         return true;
     }
 
-    internal bool TryScheduleBackgroundTask<TReq>(TReq request, Func<TReq, CancellationToken, ValueTask> fulfillFunc, string? source = null) =>
-        TryScheduleBackgroundTask(new BackgroundTaskRequest<TReq>(handler, request, fulfillFunc), request, source);
+    internal bool TryScheduleBackgroundTask<TReq>(TReq request, Func<TReq, CancellationToken, ValueTask> fulfillFunc) =>
```
dipkakwani, line 83: `IBackgroundTaskRequest<TReq>` constraint is possibly not reaching the payload on the network paths. `BackgroundTaskSchedulerWrapper.TryScheduleBackgroundTask<TReq>` and the two `TryScheduleSyncServe` overloads put no constraint on `TReq`.
 
The constraint is satisfied by the wrapper struct instead. But the reported name comes from `TReq`. So the name is only right because each wrapper hand-writes `TaskId => BackgroundTaskTypeId<TReq>.Id`. If that is omitted, we get the wrapper's own name, which is the same for every TReq.

src/Nethermind/Nethermind.Network.Test/P2P/Subprotocols/Eth/V72/Eth72ProtocolHandlerTests.cs
```diff
@@ -5342,25 +5342,28 @@ private void ParkEmptyCellsResponse(Hash256 hash, BlobCellMask cellMask)
...
         public bool TryScheduleTask<TReq>(
             TReq request,
             Func<TReq, CancellationToken, Task> fulfillFunc,
-            TimeSpan? timeout = null,
-            string? source = null)
-            => source != rejectedSource
-            && RunImmediatelyScheduler.Instance.TryScheduleTask(request, fulfillFunc, timeout, source);
+            TimeSpan? timeout = null)
+            where TReq : notnull, IBackgroundTaskRequest<TReq>
+            => BackgroundTaskTypeRegistry.GetName(TReq.TaskId) != rejectedSource
```
dipkakwani: The reject now depends on a registry name that can change. Two types sharing the simple name `ClaimedCellsResponse` makes it `FullName`. An id past `MaxTaskTypes` makes it null. Then nothing matches and SourceRejectingBackgroundTaskScheduler stops rejecting.

src/Nethermind/Nethermind.Network/P2P/Utils/BackgroundTaskSchedulerWrapper.cs
LukaszRozmej: You're right, and it's the sharp edge of making `TaskId` defaulted: for a wrapper the default is actively wrong, and the compiler can't tell. I don't think the constraint can express it — the wrapper is what's scheduled, `TReq` is just one of its type arguments, so any constraint on `TReq` would still leave the *choice* of which type to report to the wrapper's override.

What I could do is stop it being "only right because someone hand-wrote it", so d96fb73 pins it with a test over all four scheduling paths (both `TryScheduleSyncServe` overloads, the `ISyncServeRequestHandler` one, and `TryScheduleBackgroundTask`):

```csharp
[Test]
public void Scheduling_reports_the_wrapped_message_type([Values] SchedulingPath path)
{
    ...
    Assert.That(scheduler.ReportedName, Is.EqualTo(nameof(TestRequestMessage)));
}
```

I checked it actually bites rather than just passing — deleting the override on `SyncServeTaskRequest` fails it with:

```
Expected: "TestRequestMessage"
But was:  "SyncServeTaskRequest`2"
```

which is exactly the collapse you describe. Happy to go further and restructure so the wrapper can't report itself at all, but that means `TryScheduleTask` taking the name type as a separate parameter, and I'd rather not widen that interface again in this PR unless you think it's worth it.

src/Nethermind/Nethermind.Network.Test/P2P/Subprotocols/Eth/V72/Eth72ProtocolHandlerTests.cs
LukaszRozmej: Agreed — matching on the reported name made the test depend on registry state that this PR itself can change, and the failure mode is the bad one: the task gets scheduled inline, the throttle path under test never runs, and the assertions are all `DidNotReceive` so it could have passed while testing nothing.

Fixed in d96fb73 on both counts. It now matches the wrapped request's own `Type.Name`, which no amount of collision qualification or id-cap overflow can move:

```csharp
foreach (Type wrapped in typeof(TReq).GenericTypeArguments)
{
    if (wrapped.Name == rejectedRequest)
    {
        Rejected++;
        return false;
    }
}

return RunImmediatelyScheduler.Instance.TryScheduleTask(request, fulfillFunc, timeout);
```

and the test asserts the rejection happened, so if the match ever stops working it fails loudly instead of silently:

```csharp
Assert.That(rejectingScheduler.Rejected, Is.GreaterThan(0), "the cells response must actually have been rejected");
```

dipkakwani (approved):

PR #9970, opened 2025-12-18
Fix: Correct memory access validation in TrySaveByte for MSTORE8 operation

0xlupin:
Fixes memory access validation in `EvmPooledMemory.TrySaveByte` to check for 1-byte access instead of 32-byte word access.

### DIFF
--- a/src/Nethermind/Nethermind.Evm/EvmPooledMemory.cs
+++ b/src/Nethermind/Nethermind.Evm/EvmPooledMemory.cs
@@ -45,7 +45,7 @@ public struct EvmPooledMemory : IEvmMemory
 
     public bool TrySaveByte(in UInt256 location, byte value)
     {
-        CheckMemoryAccessViolation(in location, 1, out _, out bool isViolation);
+        CheckMemoryAccessViolation(in location, 1, out ulong newLength, out bool isViolation);
         if (isViolation) return false;
 
         UpdateSize(newLength);


src/Nethermind/Nethermind.Evm/EvmPooledMemory.cs
```diff
@@ -45,10 +45,11 @@ ref Unsafe.Add(ref MemoryMarshal.GetArrayDataReference(_memory), offset),
 
     public bool TrySaveByte(in UInt256 location, byte value)
     {
-        CheckMemoryAccessViolation(in location, WordSize, out _, out bool outOfGas);
+        // MSTORE8 only touches a single byte; validate against the actual byte length.
```
LukaszRozmej:
```suggestion
```

src/Nethermind/Nethermind.Evm/EvmPooledMemory.cs
```diff
@@ -45,10 +45,10 @@ ref Unsafe.Add(ref MemoryMarshal.GetArrayDataReference(_memory), offset),
 
     public bool TrySaveByte(in UInt256 location, byte value)
     {
-        CheckMemoryAccessViolation(in location, WordSize, out _, out bool isViolation);
+        CheckMemoryAccessViolation(in location, 1, out _, out bool isViolation);
```
LukaszRozmej:
```suggestion
        CheckMemoryAccessViolation(in location, 1, out ulong newLength, out bool isViolation);
```

Demuirgos (approved):

PR #9971, opened 2025-12-18
Remove rpc from hive worflow

deffrian:
rpc == rpc-compat

### DIFF
--- a/.github/workflows/hive-tests.yml
+++ b/.github/workflows/hive-tests.yml
@@ -10,13 +10,12 @@ on:
         description: Test suite
         type: choice
         options:
-          - ''
+          - 'all'
           - devp2p
           - ethereum/consensus
           - ethereum/eest
           - ethereum/engine
           - ethereum/graphql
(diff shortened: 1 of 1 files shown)


.github/workflows/hive-tests.yml
```diff
@@ -10,13 +10,12 @@ on:
...
         type: choice
         options:
-          - ''
+          - 'all'
           - devp2p
           - ethereum/consensus
           - ethereum/eest
           - ethereum/engine
           - ethereum/graphql
-          - ethereum/rpc
```
rubo, line 19: If they are the same, maybe removing `rpc-compat` instead, as simple `rpc` reads better? Or is there a reason?

deffrian: it's called rpc-compat here. So probably better leave rpc-compat for consistency 
https://hive.ethpandaops.io/#/group/generic

rubo (approved):

PR #9972, opened 2025-12-18
Fixes #9577: resolve race condition in StateSyncFeedTests.Big_test

### LINKED ISSUE #9577: Big_test is failing with missing trie nodes
To reproduce the issue, you can increase the repeat count [here](https://github.com/NethermindEth/nethermind/blob/bugfix/trie-test-logging/src/Nethermind/Nethermind.Synchronization.Test/FastSync/StateSyncFeedTests.cs#L41)

It’s worth improving the logging to capture the failing case more accurately.

Please investigate whether this is caused by a test setup issue or a problem in the code itself. In either case, it should be fixed.

If we confirm that the issue lies in the code, we could consider running this test multiple times in the CI pipelines to ensure stability.

igladun:
Fixes #9577

## Changes

- Remove static `BlockTree` field in `StateSyncFeedTestsBase` that was shared across parallel test fixtures
- Add `TestChainLength` constant (100) to replace the shared state dependency
- Add per-test `_blockTree` field and `SetBlockTree()` method to `SyncPeerMock`
- Update `GetHeadBlockHeader()` to use the per-test block tree instance

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

#### Notes on testing

Verified by running `Big_test` 100 times with 4 parallel workers:

| Execution | Workers | Iterations | Passed | Failed |
|-----------|---------|------------|--------|--------|
| Before fix | 4 | 100 | ~81 | ~19 (19% failure rate) |
| After fix | 4 | 100 | 100 | 0 |

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks

The root cause was `SyncPeerMock.GetHeadBlockHeader()` referencing a static `BlockTree` field shared across all parallel test fixtures. When tests ran concurrently and modified their block trees, they interfered with each other through this shared state.

To reproduce the issue and the fix run
`docker build -f Dockerfile.big-test --build-arg REPEAT_COUNT=100 -t big-test-verify . && docker run --rm --cpus=4 --memory=3g big-test-verify`

> FROM mcr.microsoft.com/dotnet/nightly/sdk:10.0-preview
> 
> ARG REPEAT_COUNT=100
> ARG TEST_WORKERS=4
> 
> ENV DOTNET_CLI_TELEMETRY_OPTOUT=1
> ENV DOTNET_NOLOGO=1
> ENV TEST_WORKERS=${TEST_WORKERS}
> 
> WORKDIR /nethermind
> 
> COPY src/Nethermind src/Nethermind
> COPY Directory.*.props .
> COPY global.json .
> COPY nuget.config .
> 
> RUN echo '{"sdk":{"rollForward":"latestMajor","allowPrerelease":true}}' > global.json && \
>     sed -i '/<RuntimeIdentifiers>/d' src/Nethermind/Nethermind.Runner/Nethermind.Runner.csproj && \
>     sed -i 's/<TreatWarningsAsErrors>true<\/TreatWarningsAsErrors>/<TreatWarningsAsErrors>false<\/TreatWarningsAsErrors>/g' Directory.Build.props && \
>     sed -i "s/private const int TestRepeatCount = [0-9]*;/private const int TestRepeatCount = ${REPEAT_COUNT};/" \
>         src/Nethermind/Nethermind.Synchronization.Test/FastSync/StateSyncFeedTests.cs
> 
> WORKDIR /nethermind/src/Nethermind
> 
> RUN dotnet restore Nethermind.Synchronization.Test/Nethermind.Synchronization.Test.csproj && \
>     dotnet build Nethermind.Synchronization.Test/Nethermind.Synchronization.Test.csproj -c Release --no-restore
> 
> CMD dotnet test Nethermind.Synchronization.Test/Nethermind.Synchronization.Test.csproj \
>     -c Release --no-build --filter "FullyQualifiedName~Big_test" \
>     --logger "console;verbosity=normal" -- NUnit.NumberOfTestWorkers=${TEST_WORKERS}

### DIFF
--- a/src/Nethermind/Nethermind.Synchronization.Test/FastSync/StateSyncFeedTestsBase.cs
+++ b/src/Nethermind/Nethermind.Synchronization.Test/FastSync/StateSyncFeedTestsBase.cs
@@ -124,9 +124,9 @@ public abstract class StateSyncFeedTestsBase(int defaultPeerCount = 1, int defau
             .AddSingleton<INodeStorage>(dbContext.LocalNodeStorage)
 
             // Use factory function to make it lazy in case test need to replace IBlockTree
-            // Use unique ID per test to avoid cache collisions during parallel execution
+            // Cache key includes type name so different inherited test classes don't share the same blocktree
             .AddSingleton<IBlockTree>((ctx) => CachedBlockTreeBuilder.BuildCached(
-                $"{GetType().Name}{dbContext.RemoteStateTree.RootHash}{TestChainLength}{Guid.NewGuid()}",
+                $"{GetType().Name}{dbContext.RemoteStateTree.RootHash}{TestChainLength}",
                 () => Build.A.BlockTree().WithStateRoot(dbContext.RemoteStateTree.RootHash).OfChainLength(TestChainLength)))
 
             .Add<SafeContext>();


src/Nethermind/Nethermind.Synchronization.Test/FastSync/StateSyncFeedTestsBase.cs
```diff
@@ -122,8 +125,8 @@ protected ContainerBuilder BuildTestContainerBuilder(DbContext dbContext, int sy
 
             // Use factory function to make it lazy in case test need to replace IBlockTree
             .AddSingleton<IBlockTree>((ctx) => CachedBlockTreeBuilder.BuildCached(
-                $"{nameof(StateSyncFeedTestsBase)}{dbContext.RemoteStateTree.RootHash}{BlockTree.BestSuggestedHeader!.Number}",
-                () => Build.A.BlockTree().WithStateRoot(dbContext.RemoteStateTree.RootHash).OfChainLength((int)BlockTree.BestSuggestedHeader!.Number)))
+                $"{nameof(StateSyncFeedTestsBase)}{dbContext.RemoteStateTree.RootHash}{TestChainLength}",
```
flcl42: Maybe we can also use real type name: GetType().Name instead of nameof(StateSyncFeedTestsBase). So different inherited test sets will not share the same blocktree.
cc @asdacap

igladun: @flcl42 ,  applied your suggestion and additionally added `Guid.NewGuid()` for extra safety

flcl42: Guid.NewGuid() seems to make cache items never be reused, does not it?

igladun: @flcl42 , and if we use unique ID per test fixture instance ? Will it be ok?

flcl42: Yes

src/Nethermind/Nethermind.Synchronization.Test/FastSync/StateSyncFeedTestsBase.cs
```diff
@@ -121,9 +124,10 @@ protected ContainerBuilder BuildTestContainerBuilder(DbContext dbContext, int sy
             .AddSingleton<INodeStorage>(dbContext.LocalNodeStorage)
 
             // Use factory function to make it lazy in case test need to replace IBlockTree
+            // Use unique ID per test to avoid cache collisions during parallel execution
             .AddSingleton<IBlockTree>((ctx) => CachedBlockTreeBuilder.BuildCached(
-                $"{nameof(StateSyncFeedTestsBase)}{dbContext.RemoteStateTree.RootHash}{BlockTree.BestSuggestedHeader!.Number}",
-                () => Build.A.BlockTree().WithStateRoot(dbContext.RemoteStateTree.RootHash).OfChainLength((int)BlockTree.BestSuggestedHeader!.Number)))
+                $"{GetType().Name}{dbContext.RemoteStateTree.RootHash}{TestChainLength}{Guid.NewGuid()}",
+                () => Build.A.BlockTree().WithStateRoot(dbContext.RemoteStateTree.RootHash).OfChainLength(TestChainLength)))
```
flcl42, line 131:
```suggestion
                $"{GetType().Name)}{dbContext.RemoteStateTree.RootHash}{BlockTree.BestSuggestedHeader!.Number}",
                () => Build.A.BlockTree().WithStateRoot(dbContext.RemoteStateTree.RootHash).OfChainLength((int)BlockTree.BestSuggestedHeader!.Number)))
```

Something like this, it still caches based on the chain length, but considers difference in types

igladun: hi @flcl42 ,
I couldn't apply it exactly as written because the static BlockTree field was removed as part of this fix. It was the root cause of the race condition (shared state across parallel test fixtures)

I've also removed the per instance _testInstanceId Guid as you suggested, so caching now works properly across test instances of the same type

Is it looks good now for you?

flcl42 (approved):

PR #9976, opened 2025-12-18
refactor: Remove redundant try-catch in SszProperty.StaticLength

Aleksandr1732:
Removes an empty try-catch block in `SszProperty.StaticLength` property that was catching and immediately re-throwing exceptions without any additional handling.

### DIFF
--- a/src/Nethermind/Nethermind.Serialization.SszGenerator/SszProperty.cs
+++ b/src/Nethermind/Nethermind.Serialization.SszGenerator/SszProperty.cs
@@ -81,6 +81,7 @@ class SszProperty
             {
                 return 4;
             }
+
             return Kind switch
             {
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Serialization.SszGenerator/SszProperty.cs
```diff
@@ -81,16 +81,12 @@ public int StaticLength
             {
                 return 4;
             }
-            try
+            return Kind switch
```
flcl42, line 85: Need an empty line before return

LukaszRozmej, line 85:
```suggestion

            return Kind switch
```

PR #9982, opened 2025-12-19
 refactor(trie): remove redundant condition check in IsValidWithOneNodeLess

0xlupin:
Eliminates duplicate logic in `TrieNode.IsValidWithOneNodeLess` property getter.

### DIFF
--- a/src/Nethermind/Nethermind.Trie/TrieNode.cs
+++ b/src/Nethermind/Nethermind.Trie/TrieNode.cs
@@ -230,7 +230,12 @@ namespace Nethermind.Trie
                 int nonEmptyNodes = 0;
                 for (int i = 0; i < BranchesCount; i++)
                 {
-                    if (!IsChildNull(i) && ++nonEmptyNodes > 2)
+                    if (!IsChildNull(i))
+                    {
+                        nonEmptyNodes++;
+                    }
+
+                    if (nonEmptyNodes > 2)
                     {
                         return true;
                     }


src/Nethermind/Nethermind.Trie/TrieNode.cs
```diff
@@ -230,18 +230,13 @@ public bool IsValidWithOneNodeLess
                 int nonEmptyNodes = 0;
                 for (int i = 0; i < BranchesCount; i++)
                 {
-                    if (!IsChildNull(i))
-                    {
-                        nonEmptyNodes++;
-                    }
-
-                    if (nonEmptyNodes > 2)
+                   if (!IsChildNull(i) && ++nonEmptyNodes > 2)
```
LukaszRozmej:
```suggestion
                    if (!IsChildNull(i) && ++nonEmptyNodes > 2)
```

src/Nethermind/Nethermind.Trie/TrieNode.cs
```diff
@@ -230,18 +230,13 @@ public bool IsValidWithOneNodeLess
                 int nonEmptyNodes = 0;
                 for (int i = 0; i < BranchesCount; i++)
                 {
-                    if (!IsChildNull(i))
-                    {
-                        nonEmptyNodes++;
-                    }
-
-                    if (nonEmptyNodes > 2)
+                    if (!IsChildNull(i) && ++nonEmptyNodes > 2)
```
benaadams: Don't really like Pre/Post Increment as part of a conditional as is cognitive load
```suggestion
                    if (!IsChildNull(i))
                    {
                        nonEmptyNodes++;
                    }

                    if (nonEmptyNodes > 2)
```

PR #9984, opened 2025-12-20
Fix/json converters exception type

Galoretka:
## Changes

- Replace InvalidOperationException with JsonException in BooleanConverter and BigIntegerConverter when the JSON token type is invalid. JsonException is the expected exception type for malformed JSON payloads and allows the serializer to enrich errors with JSON path information. This change aligns these converters with the rest of Nethermind.Serialization.Json and provides consistent error reporting for deserialization failures.


src/Nethermind/Nethermind.Serialization.Json/BigIntegerConverter.cs
```diff
@@ -17,7 +17,7 @@ public override BigInteger Read(ref Utf8JsonReader reader, Type typeToConvert, J
             {
                 JsonTokenType.Number => new BigInteger(reader.GetInt64()),
                 JsonTokenType.String => BigInteger.Parse(reader.GetString()),
-                _ => throw new InvalidOperationException()
+                _ => throw new JsonException()
```
LukaszRozmej:
```suggestion
                _ => throw new JsonException($"Cannot convert {reader.TokenType} to {nameof(BigInteger)}")
```

src/Nethermind/Nethermind.Serialization.Json/BooleanConverter.cs
```diff
@@ -44,7 +44,7 @@ public override bool Read(
                 }
             }
 
-            throw new InvalidOperationException();
+            throw new JsonException();
```
LukaszRozmej:
```suggestion
            throw new JsonException($"Cannot convert {reader.TokenType} to {nameof(bool)}");
```

Galoretka: > 

Applied

PR #9985, opened 2025-12-20
test: replace Task.Delay with After() in ContractDataStoreTests

forkfury:
### Fix
Replaced fragile Task.Delay() calls in ContractDataStoreTests with NUnit's After() assertion modifier to reliably wait for async operations. This removes race conditions and makes tests deterministic.
### Changes
- Removed all await Task.Delay(...) calls from 5 test methods
- Converted async Task test methods to synchronous void methods
- Added Assert.That with After(1000, 100) for consistent timeout-based waiting
- Removed unused using System.Threading.Tasks; import

Affected tests:
- returns_data_from_receipts_on_non_consecutive_with_not_incremental_changes
- returns_data_from_receipts_on_consecutive_with_not_incremental_changes
- returns_data_from_receipts_on_consecutive_with_incremental_changes
- returns_unmodified_data_from_empty_receipts_on_consecutive_with_incremental_changes
- returns_data_from_receipts_on_consecutive_with_incremental_changes_with_identity

### DIFF
--- a/src/Nethermind/Nethermind.AuRa.Test/Contract/ContractDataStoreWithLocalDataTests.cs
+++ b/src/Nethermind/Nethermind.AuRa.Test/Contract/ContractDataStoreWithLocalDataTests.cs
@@ -86,7 +86,7 @@ namespace Nethermind.AuRa.Test.Contract
 
             Assert.That(
                 () => testCase.ContractDataStore.GetItemsFromContractAtBlock(secondBlock.Header),
-                Is.EquivalentTo(expected.Cast<object>()).After(1000, 100)
+                Is.EquivalentTo(expected.Cast<object>()).After(200, 20)
             );
 
             localDataSource.Data.Returns(new[] { TestItem.AddressC, TestItem.AddressD });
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.AuRa.Test/Contract/ContractDataStoreTests.cs
```diff
@@ -109,9 +108,10 @@ public async Task returns_data_from_receipts_on_non_consecutive_with_not_increme
         testCase.ContractDataStore.GetItemsFromContractAtBlock(blockHeader);
         testCase.BlockTree.NewHeadBlock += Raise.EventWith(new BlockEventArgs(secondBlock));
 
-        await Task.Delay(10); // delay for refresh from contract as its async
-
-        testCase.ContractDataStore.GetItemsFromContractAtBlock(secondBlock.Header).Should().BeEquivalentTo(expected.Cast<object>());
+        Assert.That(
+            () => testCase.ContractDataStore.GetItemsFromContractAtBlock(secondBlock.Header),
+            Is.EquivalentTo(expected.Cast<object>()).After(1000, 100)
```
LukaszRozmej: looking at the delay, maybe `1000, 100` is an overkill and should be more like `200, 20`? This applies to other tests too.

PR #9987, opened 2025-12-20
refactor: remove unnecessary Task.Run from SetupKeyStore

Snezhkko:
SetupKeyStore.Execute was wrapping all initialization work in Task.Run even though steps already run on background threads via EthereumStepsManager. This extra Task.Run brought no benefit, but added overhead and made the control flow harder to follow. The method now runs synchronously on the existing step thread, explicitly respects a pre-cancelled CancellationToken at the start, and returns Task.CompletedTask after finishing the setup. This keeps the effective behaviour the same while aligning the step with other init steps and simplifying error handling.


Snezhkko:
> Tests fail

fixed

src/Nethermind/Nethermind.Runner.Test/EthereumRunnerTests.cs
```diff
@@ -176,6 +176,9 @@ public async Task Smoke_cancel((string file, ConfigProvider configProvider) test
             return;
         }
 
+        if (testCase.file.Contains("none.json")) Assert.Ignore("engine port missing");
+        if (testCase.file.Contains("radius_testnet-sepolia.json")) Assert.Ignore("sequencer url not specified");
+
```
LukaszRozmej: This shouldn't be needed here
```suggestion
```

flcl42: Adding just `if (testCase.file.Contains("none.json")) Assert.Ignore("engine port missing");` helps indeed, we have similar check above

PR #9991, opened 2025-12-21
refactor(tests): Remove unused variables in Enode exception tests

marukai67:
Simplified exception assertion tests in PeerManagerTests by eliminating redundant local variables.

### DIFF
--- a/src/Nethermind/Nethermind.Network.Test/PeerManagerTests.cs
+++ b/src/Nethermind/Nethermind.Network.Test/PeerManagerTests.cs
@@ -121,17 +121,14 @@ namespace Nethermind.Network.Test
         [Test]
         public void Will_return_exception_in_port() =>
             Assert.Throws<ArgumentException>(static () => new Enode(enode3String));
-        
 
         [Test]
         public void Will_return_exception_in_dns() =>
             Assert.Throws<ArgumentException>(static () => new Enode(enode4String));
-        
 
         [Test]
         public void Will_return_exception_when_there_is_no_port() =>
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Network.Test/PeerManagerTests.cs
```diff
@@ -121,28 +121,19 @@ public async Task Will_discard_a_duplicate_incoming_session()
         [Test]
         public void Will_return_exception_in_port()
         {
-            Assert.Throws<ArgumentException>(static delegate
-            {
-                Enode unused = new(enode3String);
-            });
+            Assert.Throws<ArgumentException>(static () => new Enode(enode3String));
```
LukaszRozmej: Methods are one-liners. Use `=>` pattern rather than `{ }`

marukai67: done :)

src/Nethermind/Nethermind.Network.Test/PeerManagerTests.cs
```diff
@@ -162,31 +150,19 @@ public void Will_parse_port_correctly_when_there_is_only_one()
...
 
         [Test]
-        public void Will_return_exception_on_wrong_ports_part()
-        {
-            Assert.Throws<ArgumentException>(static delegate
-            {
-                Enode unused = new(enode8String);
-            });
-        }
+        ublic void Will_return_exception_on_wrong_ports_part() =>
```
LukaszRozmej:
```suggestion
        public void Will_return_exception_on_wrong_ports_part() =>
```

PR #9994, opened 2025-12-21
perf: reduce allocations in TypeExtensions

0xxFloki:
Refactored `TypeExtensions` to eliminate redundant allocations and improve code clarity while maintaining identical functionality.

### DIFF
--- a/src/Nethermind/Nethermind.Core/Extensions/TypeExtensions.cs
+++ b/src/Nethermind/Nethermind.Core/Extensions/TypeExtensions.cs
@@ -24,10 +24,16 @@ public static class TypeExtensions
         foreach (Type implementation in implementations)
         {
             Type[] allInterfaces = implementation.GetInterfaces();
-            using ArrayPoolListRef<Type> interfaces = allInterfaces
-                .Where(iface => !baseInterfaces.Contains(iface))
-                .ToPooledListRef(allInterfaces.Length);
-            
+            using ArrayPoolListRef<Type> interfaces = new(allInterfaces.Length);
+
+            foreach (Type iface in allInterfaces)
+            {
+                if (!baseInterfaces.Contains(iface))
+                {
+                    interfaces.Add(iface);
+                }
+            }
+
             if (interfaces.Contains(interfaceType))
             {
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Core/Extensions/TypeExtensions.cs
```diff
@@ -22,9 +22,7 @@ public static Type GetDirectInterfaceImplementation(this Type interfaceType)
 
         foreach (Type implementation in implementations)
         {
-            List<Type> interfaces = implementation.GetInterfaces().ToList();
-
-            interfaces.RemoveAll(i => baseInterfaces.Contains(i));
+            IEnumerable<Type> interfaces = implementation.GetInterfaces().Except(baseInterfaces);
```
LukaszRozmej: Except allocates `HashSet`, so you changed one allocation for other.
You did manage to reduce O(N^2) checks to O(N).

What can be done better:
- Use `ToPooledListRef()` extension method to avoid allocation (remember about disposing it).
- We could also use `PooledSet` to reduce the comparisions, but this is probably not worth it as we only expect a few interface implementations per item.

0xxFloki: now everything should be great))

src/Nethermind/Nethermind.Core/Extensions/TypeExtensions.cs
```diff
@@ -22,10 +23,11 @@ public static Type GetDirectInterfaceImplementation(this Type interfaceType)
 
         foreach (Type implementation in implementations)
         {
-            List<Type> interfaces = implementation.GetInterfaces().ToList();
-
-            interfaces.RemoveAll(i => baseInterfaces.Contains(i));
-
+            Type[] allInterfaces = implementation.GetInterfaces();
+            using ArrayPoolListRef<Type> interfaces = allInterfaces
+                .Where(iface => !baseInterfaces.Contains(iface))
```
LukaszRozmej: still allocates an iterators, maybe we can avoid that?

PR #9996, opened 2025-12-22
fix(evm): align TrySaveByte bounds with memory cost

MozirDmitriy:
## Changes

- Previously `EvmPooledMemory.TrySaveByte` used WordSize (32) as the length when calling `CheckMemoryAccessViolation`, while `UpdateSize` and the actual write only extended memory by a single byte. This created an inconsistent boundary behavior near the upper address range where `CalculateMemoryCost(location, length = 1)` would allow an access, but `TrySaveByte` could still reject it as out-of-gas. The method now checks bounds with a length of 1, making the access semantics consistent with the gas/memory cost calculation.

- [x] Refactoring

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

A regression test `SaveByte_AtMaxAllowedLocation_ShouldSucceedAfterCostCheck` was added to verify that a one-byte write at the maximum allowed location succeeds after a successful cost calculation.

### DIFF
--- a/src/Nethermind/Nethermind.Evm.Test/EvmPooledMemoryTests.cs
+++ b/src/Nethermind/Nethermind.Evm.Test/EvmPooledMemoryTests.cs
@@ -187,6 +187,19 @@ public class EvmPooledMemoryTests : EvmMemoryTestsBase
     }
 
     [Test]
+    public void SaveByte_AtMaxAllowedLocation_ShouldSucceedAfterCostCheck()
+    {
+        EvmPooledMemory memory = new();
+        // Use a large but realistic offset to avoid huge allocations while
+        // still validating that a successful cost calculation allows a byte write.
+        UInt256 location = (UInt256)(1024 * 1024); // 1 MiB
+        long cost = memory.CalculateMemoryCost(in location, UInt256.One, out bool outOfGas);
+        Assert.That(outOfGas, Is.EqualTo(false));
+        bool result = memory.TrySaveByte(in location, 0x42);
+        Assert.That(result, Is.EqualTo(true));
+    }
+
+    [Test]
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Evm.Test/EvmPooledMemoryTests.cs
```diff
@@ -186,6 +186,19 @@ public void SaveByte_LocationExceedsULong_ShouldReturnOutOfGas()
...
     }
 
+    [Test]
+    public void SaveByte_AtMaxAllowedLocation_ShouldSucceedAfterCostCheck()
+    {
+        EvmPooledMemory memory = new();
+        // Use a large but realistic offset to avoid huge allocations while
+        // still validating that a successful cost calculation allows a byte write.
+        UInt256 location = (UInt256)(1024 * 1024); // 1 MiB
+        long cost = memory.CalculateMemoryCost(in location, UInt256.One, out bool outOfGas);
```
LukaszRozmej, line 196:
```suggestion
        memory.CalculateMemoryCost(in location, UInt256.One, out bool outOfGas);
```

src/Nethermind/Nethermind.Evm.Test/EvmPooledMemoryTests.cs
```diff
@@ -186,6 +186,19 @@ public void SaveByte_LocationExceedsULong_ShouldReturnOutOfGas()
         Assert.That(outOfGas, Is.EqualTo(true));
     }
 
+    [Test]
+    public void SaveByte_AtMaxAllowedLocation_ShouldSucceedAfterCostCheck()
```
LukaszRozmej, line 190: This test passes without the changes in the PR

PR #9998, opened 2025-12-22
Taiko Shasta Changes - II

dipkakwani:
Closes https://github.com/NethermindEth/Surge/issues/216 
More Taiko Shasta fork changes based on the latest changes done in taiko-geth. Also fixes a few existing differences in taiko-geth and NMC implementation.

## Changes
- Update anchorv4 selector
- Fix validation to check selector as per the current fork
- Validate shasta extra data length
- Filter out all transactions from the golden touch account
- Validate `unclehash` & `withdrawalroot` (not shasta related, existing checks in taiko-geth)
- Validate timestamp in header
- Update L1Origin/HeadL1Origin
- **Fix NOTE:** Add fix for error while resetting the chain to the genesis block using `forkChoiceUpdated` (required for integration testing). Error: `Block's 0 (0xa85bf5...5a45de) main chain predecessor cannot be found and it will not be set as head`. This occurs since `TryGetBranch` tries to find the parent block during `forkChoiceUpdated`.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Taiko/TaikoHeaderValidator.cs
+++ b/src/Nethermind/Nethermind.Taiko/TaikoHeaderValidator.cs
@@ -19,12 +19,12 @@ public class TaikoHeaderValidator(
     IBlockTree? blockTree,
     ISealValidator? sealValidator,
     ISpecProvider? specProvider,
-    IL1OriginStore? l1OriginStore,
+    IL1OriginStore l1OriginStore,
     ITimestamper? timestamper,
     ILogManager? logManager)
     : HeaderValidator(blockTree, sealValidator, specProvider, logManager)
 {
-    private readonly IL1OriginStore? _l1OriginStore = l1OriginStore;
+    private readonly IL1OriginStore _l1OriginStore = l1OriginStore ?? throw new ArgumentNullException(nameof(l1OriginStore));
     private readonly ITimestamper _timestamper = timestamper ?? Timestamper.Default;
 
     // EIP-4396 calculation parameters.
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Taiko/TaikoHeaderValidator.cs
```diff
@@ -140,12 +179,50 @@ private static UInt256 ClampEip4396BaseFeeShasta(UInt256 baseFee)
...
+            {
+                error = BlockErrorMessages.InvalidTimestamp;
+                if (_logger.IsWarn) _logger.Warn($"Invalid block header ({header.Hash}) - timestamp before parent");
+                return false;
+            }
+        }
+
+        // If not a preconfirmation block, check that timestamp is not in the future
+        if (_l1OriginStore is not null)
+        {
```
smartprogrammer93: the store should never be null. so do this check in constructor. make the store not nullable

src/Nethermind/Nethermind.Taiko/TaikoHeaderValidator.cs
```diff
@@ -140,12 +179,50 @@ private static UInt256 ClampEip4396BaseFeeShasta(UInt256 baseFee)
...
+            L1Origin? l1Origin = _l1OriginStore.ReadL1Origin((UInt256)header.Number);
+            if (l1Origin is not null && !l1Origin.IsPreconfBlock)
```
smartprogrammer93: invert if to reduce nesting

flcl42 (approved):

PR #10002, opened 2025-12-22
Skip SST file size checks when MaxOpenFiles is specified

wurdum:
## Changes

It takes 10-15 mins to load databases (~400 gb) on Mac OS, where `MaxOpenFiles` is specified. Adding the config option fixes the issue by reducing open time to seconds.

Feature docs https://github.com/facebook/rocksdb/wiki/Speed-Up-DB-Open/cd1d15e5c4d6b37c604de257f40f9be21cb327e5#opening-db-with-optionsparanoid_checks--true--optionsskip_checking_sst_file_sizes_on_db_open--false

- [x] Other: config

#### Requires testing

- [x] Yes

- [x] No


src/Nethermind/Nethermind.Db.Rocks/Config/RocksDbConfigFactory.cs
```diff
@@ -35,6 +35,9 @@ public IRocksDbConfig GetForDatabase(string databaseName, string? columnName)
                 }
 
                 dbConfig.MaxOpenFiles = perDbLimit;
+
+                // Drastically speeds up DB open time on systems with limited file descriptors
+                dbConfig.RocksDbOptions += "skip_checking_sst_file_sizes_on_db_open=true;";
```
LukaszRozmej: Should we make this OS dependant?

wurdum: Yeah, makes sense. I'd also introduced a config parameter `IDbConfig.SkipCheckingSstFileSizesOnDbOpen` set to `true` for Mac OS, and `false` for others.

@LukaszRozmej @asdacap, what do you think?

wurdum: Updated to have a config parameter and be enabled by default on MacOS.

asdacap (approved):

PR #10006, opened 2025-12-22
fix: delegate raw converters read to base converters

ANtutov:
Removed duplicated parsing logic from LongRawJsonConverter and ULongRawJsonConverter by delegating their Read methods to LongConverter and ULongConverter. It keeps the existing Write behavior (always writing numeric JSON values) while ensuring that any future adjustments to number parsing in the base converters are automatically reflected in the raw variants.

### DIFF
--- a/src/Nethermind/Nethermind.Serialization.Json/LongRawJsonConverter.cs
+++ b/src/Nethermind/Nethermind.Serialization.Json/LongRawJsonConverter.cs
@@ -10,14 +10,12 @@ namespace Nethermind.Serialization.Json;
 
 public class LongRawJsonConverter : JsonConverter<long>
 {
-    private readonly LongConverter _converter = new();
-
     public override long Read(
         ref Utf8JsonReader reader,
         Type typeToConvert,
         JsonSerializerOptions options)
     {
-        return _converter.Read(ref reader, typeToConvert, options);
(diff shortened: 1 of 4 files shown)


src/Nethermind/Nethermind.Serialization.Json/LongRawJsonConverter.cs
```diff
@@ -10,28 +10,14 @@ namespace Nethermind.Serialization.Json;
...
-                return LongConverter.FromString(reader.ValueSpan);
-            }
-            else
-            {
-                return LongConverter.FromString(reader.ValueSequence.ToArray());
-            }
-        }
-
-        throw new JsonException();
+        return _converter.Read(ref reader, typeToConvert, options);
```
LukaszRozmej: Maybe better to make them both reference a static method in `LongConverter` instead?

ANtutov: > Maybe better to make them both reference a static method in `LongConverter` instead?

Ok, i will make changes till tomorrow

ANtutov: > Maybe better to make them both reference a static method in `LongConverter` instead?

corrected this

PR #10012, opened 2025-12-23
fix: use requested block number in SimulateBlockhashProvider

Fibonacci747:
## Changes

- SimulateBlockhashProvider was passing BestKnownNumber as the requested block number when delegating to the underlying IBlockhashProvider for BestSuggestedHeader. This broke the IBlockhashProvider contract and could return an incorrect, but non-null, block hash instead of treating the requested block as out of range or unknown. The implementation now forwards the original requested block number while still using BestSuggestedHeader as the current block context, preserving correct BLOCKHASH semantics in simulate environments.

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No


deffrian (changes requested):
We don't need `SimulateBlockhashProvider.cs` at all.
`bestKnown` is always bigger then `number`

PR #10013, opened 2025-12-23
Remove redundant topic filtering allocations in FilterStore

splinter012:
simplify GetTopicsFilter to build topic expressions in a single pass, drop temporary FilterTopic structs and extra ToArray allocations, preserve existing behavior for null/empty/single/multiple topic groups

### DIFF
--- a/src/Nethermind/Nethermind.Facade/Filters/FilterStore.cs
+++ b/src/Nethermind/Nethermind.Facade/Filters/FilterStore.cs
@@ -172,25 +172,15 @@ namespace Nethermind.Blockchain.Filters
                 return SequenceTopicsFilter.AnyTopic;
             }
 
-            List<TopicExpression> expressions = new();
+            FilterTopic?[] filterTopics = GetFilterTopics(topics) ?? Array.Empty<FilterTopic?>();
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Facade/Filters/FilterStore.cs
```diff
@@ -172,64 +172,29 @@ private static TopicsFilter GetTopicsFilter(IEnumerable<Hash256[]?>? topics = nu
...
+                else
                 {
-                    return new OrExpression(filterTopic.Topics.Select(
-                        static t => new SpecificTopic(t)).ToArray<TopicExpression>());
+                    expressions.Add(new OrExpression(group.Select(static t => new SpecificTopic(t)).ToArray<TopicExpression>()));
                 }
             }
 
-            return AnyTopic.Instance;
+            return new SequenceTopicsFilter(expressions.ToArray());
```
LukaszRozmej: We could avoid allocating `expressions` as a `List` and then doing `ToArray` as we can just allocate correct array size to begin with

splinter012: ready

PR #10015, opened 2025-12-23
fix: correct BlobTxMissingBlobs error message

eeemmmmmm:
BlobTxMissingBlobs had the same message as BlobTxMissingBlobVersionedHashes (copy-paste error).
The constant is used when txBlobCount < MinBlobsPerTransaction, so the message should be about missing blobs, not missing blob hashes.

### DIFF
--- a/src/Nethermind/Nethermind.Core/Messages/TxErrorMessages.cs
+++ b/src/Nethermind/Nethermind.Core/Messages/TxErrorMessages.cs
@@ -44,8 +44,8 @@ public static class TxErrorMessages
     public static string BlobTxGasLimitExceeded(ulong totalDataGas, ulong maxBlobGas) =>
         $"BlobTxGasLimitExceeded: Transaction's totalDataGas={totalDataGas} exceeded MaxBlobGas per transaction={maxBlobGas}.";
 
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Core/Messages/TxErrorMessages.cs
```diff
@@ -45,7 +45,7 @@ public static string BlobTxGasLimitExceeded(ulong totalDataGas, ulong maxBlobGas
         $"BlobTxGasLimitExceeded: Transaction's totalDataGas={totalDataGas} exceeded MaxBlobGas per transaction={maxBlobGas}.";
 
     public const string BlobTxMissingBlobs =
-        "blob transaction missing blob hashes";
+        "blob transaction must have at least one blob";
```
LukaszRozmej: Maybe better do it by param:

```
    public static readonly string BlobTxMissingBlobs = 
        $"blob transaction must have at least {Eip4844Constants.MinBlobsPerTransaction} blob";
```

eeemmmmmm: Sure, thanks for the suggestion!

PR #10020, opened 2025-12-24
Surge: Fix Profitabilty Check Bypass

dipkakwani:
Fixes https://github.com/NethermindEth/Surge/issues/180

In Surge, the profitability check can be bypassed by increasing the transaction's gas limit while doing a transaction (keeping the default gas price in wallet). Since the proposer does not know the exact gas used by each transaction, it uses the gas provided by the user for profitability check.

## Changes
- Filter out transactions exceeding maximum threshold gas limit in RPC `taikoAuth_txPoolContent`, compared against the actual gas spent.
- New config `Surge.MaxGasLimitRatio`, disabled by default (0 value).

## Types of changes

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


src/Nethermind/Nethermind.Taiko/Rpc/TaikoEngineRpcModule.cs
```diff
@@ -54,7 +55,8 @@ public class TaikoEngineRpcModule(IAsyncHandler<byte[], ExecutionPayload?> getPa
         IBlockFinder blockFinder,
         IShareableTxProcessorSource txProcessorSource,
         IRlpStreamDecoder<Transaction> txDecoder,
-        IL1OriginStore l1OriginStore) :
+        IL1OriginStore l1OriginStore,
+        ISurgeConfig surgeConfig) :
```
flcl42, line 59: Is it always available, even if taiko only runs?

src/Nethermind/Nethermind.Taiko/Rpc/TaikoEngineRpcModule.cs
```diff
@@ -208,6 +210,14 @@ void CommitAndDisposeBatch(Batch batch)
                         while (i < txSource.Length && txSource[i].SenderAddress == tx.SenderAddress) i++;
                         continue;
                     }
+
+                    // For Surge, filter out any transaction with very high gas limit
+                    if (surgeConfig.MaxGasLimitRatio > 0 && tx.GasLimit > tx.SpentGas * surgeConfig.MaxGasLimitRatio)
```
flcl42, line 215: Shouldn't be opposite like tx.SpentGas > tx.GasLimit * surgeConfig.MaxGasLimitRatio

smartprogrammer93: no it shouldnt. this is the correct logic actually. here we are trying to prevent someone from setting a very very high gas limit compared to the actual gas spent. ratio here is an int, so it is either 1, 2, 3 etc. if ratio was a fraction then the formula would be different. but for now, this is correct

src/Nethermind/Nethermind.Taiko/Rpc/TaikoEngineRpcModule.cs
smartprogrammer93: I believe so, @dipkakwani  to confirm

dipkakwani: @flcl42 (sorry for late reply) Yes, it is always available, with taiko it should be using all the default values. 
We also use it for `SurgeGasPriceOracle` initialization here https://github.com/NethermindEth/nethermind/blob/14d4a39103f3e4abfd2c570c3e2aefa2f8deb5af/src/Nethermind/Nethermind.Taiko/TaikoPlugin.cs#L168

I think we resolve all the configs here, so this should work fine
 https://github.com/NethermindEth/nethermind/blob/14d4a39103f3e4abfd2c570c3e2aefa2f8deb5af/src/Nethermind/Nethermind.Config/ConfigRegistrationSource.cs#L34

flcl42 (approved):

PR #10021, opened 2025-12-24
perf: avoid rlp encode-decode in simulate receipt recovery

Bashmunta:
## Changes

When a block is served from the in-memory simulate dictionary we already have the full Block with all transactions available. In this case we can construct ReceiptRecoveryBlock directly from the Block instead of re-encoding it to RLP and decoding it back again. This removes unnecessary CPU work and allocations while keeping the same behavior as other parts of the codebase that use ReceiptRecoveryBlock(Block) for in-memory blocks.

- [x] Optimization

### DIFF
--- a/src/Nethermind/Nethermind.Facade/Simulate/SimulateDictionaryBlockStore.cs
+++ b/src/Nethermind/Nethermind.Facade/Simulate/SimulateDictionaryBlockStore.cs
@@ -53,14 +53,10 @@ public class SimulateDictionaryBlockStore(IBlockStore readonlyBaseBlockStore) : 
         return readonlyBaseBlockStore.GetRlp(blockNumber, blockHash);
     }
 
-    public ReceiptRecoveryBlock? GetReceiptRecoveryBlock(long blockNumber, Hash256 blockHash)
-    {
-        if (_blockNumDict.TryGetValue(blockNumber, out Block block))
-        {
-            return new ReceiptRecoveryBlock(block);
-        }
-        return readonlyBaseBlockStore.GetReceiptRecoveryBlock(blockNumber, blockHash);
-    }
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Facade/Simulate/SimulateDictionaryBlockStore.cs
```diff
@@ -57,9 +57,7 @@ public void Delete(long blockNumber, Hash256 blockHash)
     {
         if (_blockNumDict.TryGetValue(blockNumber, out Block block))
         {
-            using NettyRlpStream newRlp = _blockDecoder.EncodeToNewNettyStream(block);
-            using var memoryManager = new CappedArrayMemoryManager(newRlp.Data);
-            return _blockDecoder.DecodeToReceiptRecoveryBlock(memoryManager, memoryManager.Memory, RlpBehaviors.None);
+            return new ReceiptRecoveryBlock(block);
```
LukaszRozmej:
```
    public ReceiptRecoveryBlock? GetReceiptRecoveryBlock(long blockNumber, Hash256 blockHash) => 
        _blockNumDict.TryGetValue(blockNumber, out Block block) 
            ? new ReceiptRecoveryBlock(block) 
            : readonlyBaseBlockStore.GetReceiptRecoveryBlock(blockNumber, blockHash);
```

Bashmunta: > ```
>     public ReceiptRecoveryBlock? GetReceiptRecoveryBlock(long blockNumber, Hash256 blockHash) => 
>         _blockNumDict.TryGetValue(blockNumber, out Block block) 
>             ? new ReceiptRecoveryBlock(block) 
>             : readonlyBaseBlockStore.GetReceiptRecoveryBlock(blockNumber, blockHash);
> ```

fixed

PR #10030, opened 2025-12-25
perf: avoid StackTrace when creating loggers from file path

mattvaux:
Use CallerFilePath parameter to extract logger name instead of expensive StackTrace operations. The filePath is already available via [CallerFilePath] attribute but was previously ignored, causing unnecessary performance overhead during logger initialization.

This change extracts the namespace and class name from the file path structure, maintaining the same logger naming convention while eliminating the StackTrace cost.


LukaszRozmej:
I don't think it is a good idea as there is no guarantee of anything here.
Way better refactor is to remove unnamed loggers. Leave only generic (prefered) and explicitly named ones.
Feel free to open PR for that.

PR #10033, opened 2025-12-25
fix: handle full UInt256 range in HexConvert.ToUInt256

dizer-ti:
ToUInt256 was converting via ulong, limiting values to 2^64.
Now correctly parses hex strings to full UInt256 range using Bytes.FromHexString.


LukaszRozmej (changes requested):
UInt256 has a Parse Method.

To be honest I think we could just drop the `HexConverter` class here and use UInt256 methods.

LukaszRozmej:
Feel free to open another PR without HexConverter it seems redundant

PR #10036, opened 2025-12-25
Kurahin

kurahin:
In TxTracer the IsTracing flag is computed once in the base constructor from the tracing feature flags. ParityLikeTxTracer and SimulateTxTracer were enabling their tracing flags (IsTracingReceipt, IsTracingActions, IsTracingInstructions, etc.) only in their own constructors after the base constructor had already run, so the IsTracing field stayed false.

VirtualMachine uses ITxTracer.IsTracing (which is backed by TxTracer.IsTracing) to decide whether a tracer is “connected” when building TransactionSubstate. With IsTracing stuck at false, top-level failures and reverts executed under ParityLikeTxTracer or SimulateTxTracer were treated as if no tracer was connected, so TransactionSubstate.Error stayed as a generic "error"/"revert" and revert data was not decoded.

This change explicitly recomputes IsTracing at the end of the ParityLikeTxTracer and SimulateTxTracer constructors, using the same boolean combination of tracing flags as ITxTracer’s default implementation. This makes IsTracing accurately reflect the actual tracing configuration, ensures TransactionSubstate sees the tracer as connected, and allows revert/error messages to be properly decoded when these tracers are used.

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain/Tracing/ParityStyle/ParityLikeTxTracer.cs
+++ b/src/Nethermind/Nethermind.Blockchain/Tracing/ParityStyle/ParityLikeTxTracer.cs
@@ -66,6 +66,19 @@ public class ParityLikeTxTracer : TxTracer
             IsTracingCode = true;
             IsTracingReceipt = true;
         }
+
+        IsTracing = IsTracingReceipt
+                    || IsTracingActions
+                    || IsTracingOpLevelStorage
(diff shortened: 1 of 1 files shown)


LukaszRozmej (changes requested):
Potentially good catch, but we need an universal mechanism for any class derived from `TxTracer`

LukaszRozmej:
Close in favor of #10053

PR #10037, opened 2025-12-25
fix: resource cleanup in BeaconHeadersSyncFeed

ANtutov:
BeaconHeadersSyncFeed.PostFinishCleanUp was only clearing the _pending and _sent collections without disposing the underlying HeadersSyncBatch instances. This meant that any owned response lists kept their buffers until the feed itself was disposed, which is inconsistent with the base HeadersSyncFeed behavior. The cleanup now matches the fast headers implementation by disposing pending and sent batches and resetting the headers progress queue, preventing unnecessary resource retention across beacon sync cycles.


LukaszRozmej (changes requested):
So basically we just shouldn't override `PostFinishCleanUp`? Maybe we should skip overriding something else too?

asdacap (approved):

benaadams (approved):

PR #10045, opened 2025-12-26
refactor: remove redundant HexConvert class

dizer-ti:
Follow-up to https://github.com/NethermindEth/nethermind/pull/10033#issuecomment-3692666959

Remove HexConvert class and use standard methods instead:
- UInt256.Parse() for hex to UInt256 conversion
- Convert.ToUInt64() for hex to ulong conversion


tools/SendBlobs/HexConvert.cs
```diff
@@ -1,18 +0,0 @@
...
-internal static class HexConvert
-{
-    public static UInt256 ToUInt256(string s)
-    {
-        return (UInt256)ToUInt64(s);
-    }
-
-    public static ulong ToUInt64(string s)
-    {
-        return Convert.ToUInt64(s, s.StartsWith("0x") ? 16 : 10);
```
alexb5dh, line 16: I'm ok with replacing `ToUInt256` with `UInt256.Parse`, however `ToUInt64` is simpler to use than `Convert.ToUInt64(s, 16)` and handles both 10 and 16 bases.

LukaszRozmej:
> @LukaszRozmej @alexb5dh could you please align on whether to keep ToUInt64 or remove it entirely?

Yeah agree with @alexb5dh that suggestion was suboptimal and previous was better on `ToUInt64`. Rest is good.

Issue #9913, opened 2025-12-10
Failed to initialize validators list on block 43572715 (0xa47f8734f602fa1370fda16917dac92514d5e01b6333ac0030283a1610691146)


fengmao:
**Description**
when i try to run nethermind geth client with snapshot from https://publicnode.com/snapshots#gnosis, it works fine first, when reach block 43572715, it throws error and node got shutdown, the environment is k8s statefulset

**Steps to Reproduce**
In order to replicate the behavior, please provide a detailed list of steps:
1、download snapshot from publicnodes
2、untar snapshot files
3、start k8s statefulset node

**Actual behavior**
```
Step StepWrapper              failed after 4,599ms Autofac.Core.DependencyResolutionException: An exception was thrown while activating λ:Nethermind.Init.Modules.MainProcessingContext -> Nethermind.Init.Modules.Ma │
│  ---> Autofac.Core.DependencyResolutionException: An exception was thrown while invoking the constructor 'Void .ctor(Autofac.ILifetimeScope, Nethermind.Blockchain.Receipts.IReceiptConfig, Nethermind.Config.IBlocksConfig, Nethermind │
│  ---> Autofac.Core.DependencyResolutionException: An exception was thrown while activating Nethermind.Init.Modules.MainProcessingContext+Components -> Nethermind.Consensus.Processing.BranchProcessor -> Nethermind.Consensus.Processi │
│  ---> Nethermind.Consensus.AuRa.AuRaException: Failed to initialize validators list on block 43572715 (0xa47f8734f602fa1370fda16917dac92514d5e01b6333ac0030283a1610691146)                                                              │
│    at Nethermind.Consensus.AuRa.Validators.ContractBasedValidator.LoadValidatorsFromContract(BlockHeader parentHeader)                                                                                                                  │
│    at Nethermind.Consensus.AuRa.Validators.ContractBasedValidator.SetFinalizationManager(IAuRaBlockFinalizationManager finalizationManager, BlockHeader parentHeader)                                                                   │
│    at Nethermind.Consensus.AuRa.Validators.ContractBasedValidator..ctor(IValidatorContract validatorContract, IBlockTree blockTree, IReceiptFinder receiptFinder, IValidatorStore validatorStore, IValidSealerStrategy validSealerStrat │
[...]

PR #10048, opened 2025-12-26
Implement zkVM compatibility

rubo:
> [!NOTE]  
> These changes are preliminary. The idea is to start a discussion on how to apply/re-implement the required changes correctly.

## Changes

- [x] Brought the work done on the stateless execution from the [mmenshikov/feature/stateless_execv2](https://github.com/NethermindEth/nethermind/tree/mmenshikov/feature/stateless_execv2) branch
- [x] Added another version of Nethermind.Numerics.Int256 package for zkVM without randomization source like in [Hash256](https://github.com/NethermindEth/nethermind/blob/feature/zkvm/src/Nethermind/Nethermind.Core/Crypto/Hash256.cs#L21-L26). It'd be good to find a better solution.
- [ ] Introduced conditional compilation to activate the abovementioned changes when needed (partially done)
- [ ] Added workflows to test the build and execution of zkVM-specific code

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [x] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [ ] No


src/Nethermind/Nethermind.Core/Crypto/Sha2.cs
rubo, line 1: @deffrian If this is not something ZKVM-specific and is just a regular SHA-256 implementation, I think we can replace this.

src/Nethermind/Nethermind.Core/Extensions/SpanExtensions.cs
rubo, line 1: @tanishqjasoria Could you please explain why we need these changes. Your initial commit says "temp change".

src/Nethermind/Nethermind.Core/Crypto/Sha2.cs
deffrian: Yes, it's regular sha. I think `System.Security.Cryptography.SHA256` is not working on risc. But maybe i'm wrong

src/Nethermind/Nethermind.Consensus/Stateless/WitnessCollector.cs
```diff
@@ -0,0 +1,31 @@
...
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using Nethermind.Blockchain.Tracing;
+using Nethermind.Consensus.Processing;
+using Nethermind.Core;
+using Nethermind.Core.Specs;
+
+namespace Nethermind.Consensus.Stateless;
+
+public interface IWitnessCollector
```
LukaszRozmej: Can be removed in favor of #9949

src/Nethermind/Nethermind.Consensus/Stateless/WitnessGeneratingBlockHashProvider.cs
```diff
@@ -0,0 +1,49 @@
...
+using Nethermind.Blockchain;
+using Nethermind.Core;
+using Nethermind.Core.Crypto;
+using Nethermind.Core.Specs;
+using Nethermind.Evm;
+using Nethermind.Serialization.Rlp;
+
+namespace Nethermind.Consensus.Stateless;
+
+public class WitnessGeneratingBlockHashProvider(IBlockhashProvider blockHashProvider, IReadOnlyBlockTree blockTree) : IBlockhashProvider
```
LukaszRozmej: Can be removed in favor of #9949

src/Nethermind/Nethermind.Consensus/Stateless/WitnessGeneratingBlockProcessingEnv.cs
```diff
@@ -0,0 +1,70 @@
...
+using Nethermind.Evm;
+using Nethermind.Evm.GasPolicy;
+using Nethermind.Evm.State;
+using Nethermind.Evm.TransactionProcessing;
+using Nethermind.Logging;
+using Nethermind.State;
+
+namespace Nethermind.Consensus.Stateless;
+
+public interface IWitnessGeneratingBlockProcessingEnv
```
LukaszRozmej: Can be removed in favor of #9949

src/Nethermind/Nethermind.Consensus/Stateless/WitnessGeneratingBlockProcessingEnvFactory.cs
```diff
@@ -0,0 +1,46 @@
...
+using Nethermind.Core.Specs;
+using Nethermind.Evm;
+using Nethermind.Evm.State;
+using Nethermind.Logging;
+using Nethermind.State;
+using Nethermind.State.OverridableEnv;
+
+namespace Nethermind.Consensus.Stateless;
+
+public interface IWitnessGeneratingBlockProcessingEnvFactory
```
LukaszRozmej: Can be removed in favor of #9913

src/Nethermind/Nethermind.Consensus/Stateless/WitnessGeneratingWorldState.cs
```diff
@@ -0,0 +1,286 @@
...
+using Nethermind.Core.Specs;
+using Nethermind.Evm.State;
+using Nethermind.Evm.Tracing.State;
+using Nethermind.Int256;
+using Nethermind.State;
+using Nethermind.State.Proofs;
+
+namespace Nethermind.Consensus.Stateless;
+
+public class WitnessGeneratingWorldState(WorldState inner) : IWorldState
```
LukaszRozmej: Can be removed in favor of #9949

src/Nethermind/Nethermind.Core/Extensions/SpanExtensions.cs
```diff
@@ -204,76 +204,28 @@ public static int FastHash(this Span<byte> input)
         [SkipLocalsInit]
         public static int FastHash(this ReadOnlySpan<byte> input)
         {
-            // Very fast hardware accelerated non-cryptographic hash function
-            var length = input.Length;
-            if (length == 0) return 0;
+            //TODO: just to remove ssl dependency - maybe wrong
```
LukaszRozmej: We need `#if ZKVM` here

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
```diff
@@ -398,6 +400,18 @@ public IEnumerable<FilterLog> FindLogs(LogFilter filter, CancellationToken cance
             return logFinder.FindLogs(filter, cancellationToken);
         }
 
+        public bool ExecuteWitness(Block blockParameter, Witness witness)
```
LukaszRozmej: Can be removed in favor of #9949

src/Nethermind/Nethermind.Facade/IBlockchainBridge.cs
```diff
@@ -50,5 +51,6 @@ public interface IBlockchainBridge : ILogFinder
         bool TryGetLogs(int filterId, out IEnumerable<FilterLog> filterLogs, CancellationToken cancellationToken = default);
         void RunTreeVisitor<TCtx>(ITreeVisitor<TCtx> treeVisitor, Hash256 stateRoot) where TCtx : struct, INodeContext<TCtx>;
         bool HasStateForBlock(BlockHeader? baseBlock);
+        Witness GenerateExecutionWitness(BlockHeader parent, Block block);
```
LukaszRozmej: Can be removed in favor of #9949

src/Nethermind/Nethermind.Init/Modules/RpcModules.cs
```diff
@@ -84,6 +85,7 @@ protected override void Load(ContainerBuilder builder)
                     .AddSingleton<FilterStore, ITimerFactory, IJsonRpcConfig>((timerFactory, rpcConfig) => new FilterStore(timerFactory, rpcConfig.FiltersTimeout))
                     .AddSingleton<FilterManager>()
                     .AddSingleton<ISimulateReadOnlyBlocksProcessingEnvFactory, SimulateReadOnlyBlocksProcessingEnvFactory>()
+                    .AddSingleton<IWitnessGeneratingBlockProcessingEnvFactory, WitnessGeneratingBlockProcessingEnvFactory>()
```
LukaszRozmej: Can be removed in favor of #9949

src/Nethermind/Nethermind.JsonRpc/Modules/DebugModule/DebugRpcModule.cs
```diff
@@ -640,4 +641,21 @@ private static ResultWrapper<TResult> GetRlpDecodingFailureResult<TResult>(Rlp b
         error = default!;
         return block;
     }
+
+    public ResultWrapper<Witness> debug_executionWitness(BlockParameter blockParameter)
```
LukaszRozmej: Can be removed in favor of #9949

src/Nethermind/Nethermind.JsonRpc/Modules/DebugModule/IDebugRpcModule.cs
```diff
@@ -123,4 +124,7 @@ ResultWrapper<IReadOnlyList<SimulateBlockResult<GethLikeTxTrace>>> debug_simulat
 
     [JsonRpcMethod(Description = "Executes a list of bundles of transactions without creating transactions on the blockchain and returns their traces", IsImplemented = true, IsSharable = false)]
     ResultWrapper<IEnumerable<IEnumerable<GethLikeTxTrace>>> debug_traceCallMany(TransactionBundle[] bundles, BlockParameter? blockParameter = null, GethTraceOptions? options = null);
+
+    [JsonRpcMethod(Description = "TODO")]
+    ResultWrapper<Witness> debug_executionWitness(BlockParameter blockParameter);
```
LukaszRozmej: Can be removed in favor of #9949

src/Nethermind/Nethermind.Logging/PathUtils.cs
```diff
@@ -16,10 +16,17 @@ public static class PathUtils
     static PathUtils()
     {
         Process process = Process.GetCurrentProcess();
+
         if (process.ProcessName.StartsWith("dotnet", StringComparison.OrdinalIgnoreCase)
             || process.ProcessName.Equals("ReSharperTestRunner", StringComparison.OrdinalIgnoreCase))
         {
-            ExecutingDirectory = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+            ExecutingDirectory =
+#if ZKVM
```
LukaszRozmej: Why this even runs in ZKVM context?

src/Nethermind/Nethermind.Serialization.Rlp/TxDecoder.cs
```diff
@@ -36,15 +36,24 @@ public sealed class GeneratedTxDecoder : TxDecoder<GeneratedTransaction>;
     protected TxDecoder(Func<T>? transactionFactory = null)
     {
         Func<T> factory = transactionFactory ?? (static () => new T());
+        RegisterDecoder(TxType.Legacy, new LegacyTxDecoder<T>(factory));
+        RegisterDecoder(TxType.AccessList, new AccessListTxDecoder<T>(factory));
+        RegisterDecoder(TxType.EIP1559, new EIP1559TxDecoder<T>(factory));
+        RegisterDecoder(TxType.Blob, new BlobTxDecoder<T>(factory));
+        RegisterDecoder(TxType.SetCode, new SetCodeTxDecoder<T>(factory));
```
LukaszRozmej, line 43: Why this makes a difference?

src/Nethermind/Nethermind.Serialization.Rlp/TxDecoder.cs
```diff
@@ -36,15 +36,24 @@ public sealed class GeneratedTxDecoder : TxDecoder<GeneratedTransaction>;
...
+        /*
```
LukaszRozmej: Remove if not needed

src/Nethermind/Nethermind.Serialization.Rlp/TxDecoder.cs
```diff
@@ -36,15 +36,24 @@ public sealed class GeneratedTxDecoder : TxDecoder<GeneratedTransaction>;
...
         RegisterDecoder(new AccessListTxDecoder<T>(factory));
         RegisterDecoder(new EIP1559TxDecoder<T>(factory));
         RegisterDecoder(new BlobTxDecoder<T>(factory));
         RegisterDecoder(new SetCodeTxDecoder<T>(factory));
+        */
     }
 
     public void RegisterDecoder(ITxDecoder decoder) => _decoders[(int)decoder.Type] = decoder;
 
+    public void RegisterDecoder(TxType type, ITxDecoder decoder) => _decoders[(int)type] = decoder;
```
LukaszRozmej, line 48: Above one should call this one

Directory.Packages.props
```diff
@@ -59,7 +61,7 @@
     <PackageVersion Include="Nethermind.Libp2p" Version="1.0.0-preview.45" />
     <PackageVersion Include="Nethermind.Libp2p.Protocols.PubsubPeerDiscovery" Version="1.0.0-preview.45" />
     <PackageVersion Include="Nethermind.MclBindings" Version="1.0.4" />
-    <PackageVersion Include="Nethermind.Numerics.Int256" Version="1.3.6" />
+    <PackageVersion Include="Nethermind.Numerics.Int256" Version="$(Int256Version)" />
```
LukaszRozmej: Do we have special version of Int256 package? Maybe it should also be reviewed and showcased?

tools/StatelessExecution/Program.cs
```diff
@@ -1,11 +1,426 @@
...
+                baseBlock = header;
+            }
+        }
+
+        if (baseBlock is null)
+        {
+            // Invalid witness headers
+            return 4;
+        }
+
```
LukaszRozmej, line 397: Is this a hardcoded block? Looks weird.
Can we make it proper?

Directory.Packages.props
rubo: [Here you go](https://github.com/NethermindEth/int256/commit/1334daf9ffb5dbe381273e1d9d54288298192346)

tools/StatelessExecution/Program.cs
rubo: cc @tanishqjasoria @deffrian

src/Nethermind/Nethermind.Serialization.Rlp/TxDecoder.cs
rubo: cc @maximmenshikov

src/Nethermind/Nethermind.Logging/PathUtils.cs
rubo: cc @tanishqjasoria

Directory.Packages.props
maximmenshikov: At some point we started hacking minipal to generate random numbers differently:
https://github.com/NethermindEth/bflat-riscv64/blob/riscv64_zk/src/bflat/modules/rng_stupid/module.c
Probably this int256 hack is no longer needed, but we need to check which function is eventually called in NativeAOT.

tools/StatelessExecution/Program.cs
maximmenshikov: @LukaszRozmej I believe for ease of testing it was hardcoded. We can eventually make blocks passed through zk input arguments.
(NOT through args - current args will break NativeAOT actually).

src/Nethermind/Nethermind.Core/Crypto/Hash256.cs
```diff
@@ -21,7 +18,12 @@ namespace Nethermind.Core.Crypto
         // Ensure that hashes are different for every run of the node and every node, so if are any hash collisions on
         // one node they will not be the same on another node or across a restart so hash collision cannot be used to degrade
         // the performance of the network as a whole.
-        private static readonly uint s_instanceRandom = (uint)System.Security.Cryptography.RandomNumberGenerator.GetInt32(int.MinValue, int.MaxValue);
+        private static readonly uint s_instanceRandom =
+#if ZKVM
+        98316501; // TODO: remove ssl
+#else
+        (uint)System.Security.Cryptography.RandomNumberGenerator.GetInt32(int.MinValue, int.MaxValue);
```
maximmenshikov, line 25: could be that it is solved by minipal fix, but need to double check

src/Nethermind/Nethermind.Serialization.Rlp/TxDecoder.cs
maximmenshikov: I believe there was a problem accessing decoder's type so early

tools/StatelessExecution/Program.cs
LukaszRozmej: Yes it should be passed through arguments.
What is the difference - arguments vs args that you mean?

Directory.Packages.props
LukaszRozmej: Makes sense, can we do something like:
- Make `s_instanceRandom` settable in UInt256 package and set it to some predefined value.
- In main repo set it to random value `#If not ZKVM`

So the code can be unified?

tools/StatelessExecution/Program.cs
maximmenshikov: There must be zkVM-specific mechanism to pass inputs/outputs. args[] can't be used because there is no actualy storage for it anywhere. Having args[] in the code will break the NativeAOT bootstrap in runtime.

LukaszRozmej:
pushed a commit, things to still be done:
- Change Int256 package and unify it
- Check SHA2 and potentially remove it
- Proper I/O read and write - this needs to be ZISK only
- Why we need to change PathUtils?
- Pipeline to build and verify
- Optional: Randomness, but we can do without it

tools/StatelessExecution/Program.cs
tanishqjasoria: yes, we had the version that took in arguments for witness and block through arguments, but we need something different for zkVM (as @maximmenshikov pointed out)

src/Nethermind/Nethermind.Core/Extensions/SpanExtensions.cs
tanishqjasoria: these were again due to our zkVm fix not supporting SSL library (thus not supporting`BitOperations.Crc32C`).
so i used the logic we were using before switching to Crc32
cc @maximmenshikov

src/Nethermind/Nethermind.Logging/PathUtils.cs
tanishqjasoria: it should not, but dotnet still compiles this as we complie the entire Nethermind project.
so we can ommit this change, as this is mostly throws a warning and it wont be used when executing in ZKVM context.

tools/StatelessExecution/Program.cs
tanishqjasoria: i think what we can do here is seperate this part in two tools?

1. StatelessExecution
 - this takes input through CLI and decodes it
 - then call the stateless execution module with these input

2. zkVM execution
-  the input is hardcodesd( can be changed when we have proper logic to provide input to zkVm) 
- then call the same stateless module as before

src/Nethermind/Nethermind.Core/Extensions/SpanExtensions.cs
rubo: > thus not supporting`BitOperations.Crc32C`

But we still use in `BitOperations.Crc32C` in [UInt256](https://github.com/NethermindEth/int256/blob/main/src/Nethermind.Int256/UInt256.cs#L3307-L3315) with no problem. Do I miss something?

tanishqjasoria: ohh interesting, i thought the local UInt256 packagae also included the fix where we removed CRC32.
@deffrian didn't we remove the use of CRC32 in a lot of places, including the UInt256 package?

tools/StatelessExecution/Program.cs
LukaszRozmej: Makes sense, the 2. Will probably be different per zkVM, we already have a hook how to do it for ZISK

src/Nethermind/Nethermind.Logging/PathUtils.cs
rubo: Sounds like reverting is better :)

rubo: Checked. No issues at build time, but the original code doesn't return a path, while the modified version does. If it's not gonna be used on zkVM, we can revert the change.

src/Nethermind/Nethermind.Core/Extensions/SpanExtensions.cs
rubo: All changes are reverted but randomizer

PR #10051, opened 2025-12-27
fix: track deleted accounts across transactions for zombie recreation

AnkushinDaniil:
Fixes Closes Resolves #

 ## Changes

  - Add `_deletedThisBlock` HashSet to track accounts deleted during block processing
  - Modify `CreateEmptyAccountIfDeletedOrNew` to check block-level deletions for cross-transaction zombie account recreation
  - Track deletions in commit logic: empty account cleanup, EIP-158 implicit deletion, and explicit delete operations
  - Clear deletion tracking in `ResetBlockChanges`

  ## Types of changes

  - [ ] Optimization
  - [ ] Refactoring
  - [ ] Documentation update
  - [ ] Build-related changes
  - [ ] Other: _Description_

  ## Testing

  #### Requires testing

  - [x] Yes
  - [ ] No

  - [ ] Yes
  - [x] No

  #### Notes on testing

  Tested via Arbitrum plugin tests which compare state roots against Nitro reference implementation

  ## Documentation

  - [ ] Yes
  - [x] No

  - [ ] Yes
  - [x] No

  ## Remarks

  Required for Arbitrum's pre-FixZombieAccounts behavior where zero-value transfers must recreate deleted accounts as empty "zombie" accounts. After the WorldState refactor, `CreateEmptyAccountIfDeletedOrNew` could only detect deletions within the same transaction's cache. This change enables detection of accounts deleted in earlier transactions within the same block.

### DIFF
--- a/src/Nethermind/Nethermind.State/StateProvider.cs
+++ b/src/Nethermind/Nethermind.State/StateProvider.cs
@@ -34,6 +34,9 @@ namespace Nethermind.State
         private readonly Dictionary<AddressAsKey, StackList<int>> _intraTxCache = new();
         private readonly HashSet<AddressAsKey> _committedThisRound = new();
         private readonly HashSet<AddressAsKey> _nullAccountReads = new();
+        // Tracks accounts deleted during this block
+        // Used by CreateEmptyAccountIfDeletedOrNew to detect cross-transaction deletions
+        private readonly HashSet<AddressAsKey> _deletedThisBlock = new();
         // Only guarding against hot duplicates so filter doesn't need to be too big
         // Note:
         // False negatives are fine as they will just result in a overwrite set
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.State/StateProvider.cs
```diff
@@ -448,12 +451,30 @@ public void CreateEmptyAccountIfDeletedOrNew(Address address)
...
+            if (!_deletedThisBlock.Contains(address)) return;
+            {
+                // Account was deleted earlier in this block
+                _needsStateRootUpdate = true;
+                if (_logger.IsTrace) TraceBlockLevel(address);
+
+                Account account = Account.TotallyEmpty;
+                StackList<int> stack = SetupCache(address);
+                PushRecreateEmpty(address, account, stack);
+            }
```
LukaszRozmej, line 468: This looks weird - like it belongs to the `if` but it does in fact is in else

LukaszRozmej (approved):

PR #10058, opened 2025-12-28
fix: remove dead code in BlockchainTestBase static constructor

PivasDesant:
Remove useless null check for _logManager. The field is already initialized as readonly, so the ??= operator never runs. Looks like leftover code.


flcl42:
Thanks! It's hard to count such pr as a contribution significant enough to be merged though

PR #10063, opened 2025-12-29
fix: normalize discovery bootnodes merge from chainspec

MozirDmitriy:
DiscoveryConfig.Bootnodes could be null and was concatenated using '+=' with chainspec bootnodes, producing a leading comma in the final enode list. At the same time the code tried to guard against a null ChainSpec.Bootnodes array using 'is not null', even though the array is always initialized by the chainspec loader.

Normalized discovery bootnodes handling by treating null and empty as the same state using string.IsNullOrEmpty and by relying on ChainSpec.Bootnodes.Length != 0 instead of a misleading null check. As a result, the merged bootnodes string is well-formed and better matches the actual contracts of both configs.

### DIFF
--- a/src/Nethermind/Nethermind.Init/Modules/DiscoveryModule.cs
+++ b/src/Nethermind/Nethermind.Init/Modules/DiscoveryModule.cs
@@ -81,7 +81,7 @@ public class DiscoveryModule(IInitConfig initConfig, INetworkConfig networkConfi
                     discoveryConfig.Bootnodes = string.Join(",", chainSpec.Bootnodes.Select(static bn => bn.ToString()));
                 }
 
-                if (networkConfig.DiscoveryDns == null)
+                if (networkConfig.DiscoveryDns is null)
                 {
                     string chainName = BlockchainIds.GetBlockchainName(chainSpec!.NetworkId).ToLowerInvariant();
                     networkConfig.DiscoveryDns = $"all.{chainName}.ethdisco.net";


src/Nethermind/Nethermind.Init/Modules/DiscoveryModule.cs
```diff
@@ -69,14 +69,14 @@ protected override void Load(ContainerBuilder builder)
                 IDiscoveryConfig discoveryConfig = ctx.Resolve<IDiscoveryConfig>();
 
                 // Was in `UpdateDiscoveryConfig` step.
-                if (discoveryConfig.Bootnodes != string.Empty)
+                if (!string.IsNullOrEmpty(discoveryConfig.Bootnodes))
```
LukaszRozmej, line 72: How this could be `null` as it is `string.Empty` by default?

src/Nethermind/Nethermind.Init/Modules/DiscoveryModule.cs
```diff
@@ -69,14 +69,14 @@ protected override void Load(ContainerBuilder builder)
...
                 {
                     if (chainSpec.Bootnodes.Length != 0)
                     {
                         discoveryConfig.Bootnodes += "," + string.Join(",", chainSpec.Bootnodes.Select(static bn => bn.ToString()));
                     }
                 }
-                else if (chainSpec.Bootnodes is not null)
+                else if (chainSpec.Bootnodes.Length != 0)
```
LukaszRozmej: Maybe better to guard for null and emptiness here.

LukaszRozmej:
```suggestion
                else if (chainSpec.Bootnodes?.Length > 0)
```

src/Nethermind/Nethermind.Init/Modules/DiscoveryModule.cs
LukaszRozmej:
```suggestion
                if (networkConfig.DiscoveryDns is null)
```

PR #10073, opened 2025-12-30
Should update child if parent has empty keccak as well

damian-orzechowski:
Fixes Closes Resolves
https://github.com/NethermindEth/nethermind-arbitrum/issues/509

Using `BulkSet` can omit changes in specific scenarios and result in incorrect trie.

## Changes

- `BulkSet` on `PatrciaTrie` can fail to update node's value when creating nested `MakeFakeBranch`. In such case, original value is retained and may not be updated when collapsed further on in `MaybeCombineNode`. This change allows to update child node  also when parent hash is empty (case above), 

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

Since the code uses `index` of `StorageCell` it's very difficult to construct an arbitrary trie in unit tests to test such specific cases.
Test case to consider:
  - 2 changes under path: `0x8c`
    - change `0x8cea` to a value
    - change `0x8cee` to a `[]`

Causes 2 calls to `MakeFakeBranch` and `ShouldUpdateChild` returns false for node under `0x8cea` (parent `Keccak` is `null`). As 2nd change is `[]` branches collapse retaining old value of leaf node under `0x8c`

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain.Test/FullPruning/CopyTreeVisitorTests.cs
+++ b/src/Nethermind/Nethermind.Blockchain.Test/FullPruning/CopyTreeVisitorTests.cs
@@ -93,18 +93,19 @@ public class CopyTreeVisitorTests
         // Use TestWorldStateFactory.CreateForTest() with the custom DbProvider
         (IWorldState worldState, IStateReader stateReader) = TestWorldStateFactory.CreateForTestWithStateReader(dbProvider, logManager);
 
+        BlockHeader? baseBlock = Build.A.BlockHeader.WithStateRoot(trie.RootHash).TestObject;
         if (_keyScheme == INodeStorage.KeyScheme.Hash)
         {
             NodeStorage nodeStorage = new NodeStorage(pruningContext, _keyScheme);
             using CopyTreeVisitor<NoopTreePathContextWithStorage> copyTreeVisitor = new(nodeStorage, writeFlags, logManager, cancellationToken);
-            stateReader.RunTreeVisitor(copyTreeVisitor, trie.RootHash, visitingOptions);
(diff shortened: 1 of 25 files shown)


src/Nethermind/Nethermind.Trie/PatriciaTree.cs
```diff
@@ -719,6 +719,7 @@ internal bool ShouldUpdateChild(TrieNode? parent, TrieNode? oldChild, TrieNode?
             if (oldChild is null && newChild is null) return false;
             if (!ReferenceEquals(oldChild, newChild)) return true;
             if (newChild.Keccak is null && parent.Keccak is not null) return true; // So that recalculate root knows to recalculate the parent root.
+            if (newChild.Keccak is null && parent.Keccak is null) return true; // parent's hash can also be null depending on nesting level - still need to update child, otherwise combine will remain original value
```
LukaszRozmej: Combine this line and the one before to just 
```suggestion
            if (newChild.Keccak is null) return true; // parent's hash can also be null depending on nesting level - still need to update child, otherwise combine will remain original value
```

asdacap: If only bulkset is failing, can you make another `ShouldUpdateChild` but just for bulk? This method is used by setting one by one too.

damian-orzechowski: Yes, so the case is gradually traversing down until one by one, so I think it makes sense for it to remain global

Issue #10074, opened 2025-12-31
`taiko_lastBlockIDByBatchID` returns hex string instead of decimal number, causing Go client unmarshaling error


jmadibekov:
## Problem

The `taiko_lastBlockIDByBatchID` RPC method returns the block ID as a hex string (e.g., `"0x2"`), but the taiko-client expects a raw decimal number (e.g., `2`).

This causes the following error in integration tests when running against Nethermind:

```bash
failed to get last block ID by batch ID 1: math/big: cannot unmarshal "\"0x2\"" into a *big.Int
```

## Call Flow

The test `TestOnBatchProposed` triggers this error through the following call chain:

### 1. Test calls GetProposalByIDShasta

```go
// prover/prover_test.go:186
payload, eventLog, err := s.RPCClient.GetProposalByIDShasta(context.Background(), common.Big1)
```

### 2. GetProposalByIDShasta calls LastBlockIDByBatchID

```go
// pkg/rpc/methods.go:1524
blockID, err := c.L2.LastBlockIDByBatchID(ctxWithTimeout, proposalID)
if err != nil {
    return nil, nil, fmt.Errorf("failed to get last block ID by batch ID %d: %w", proposalID, err)
}
```

### 3. taiko-geth client makes the RPC call

```go
// taiko-geth/ethclient/taiko_api.go:44-53
func (ec *Client) LastBlockIDByBatchID(ctx context.Context, batchID *big.Int) (*big.Int, error) {
    var res *big.Int

    if err := ec.c.CallContext(ctx, &res, "taiko_lastBlockIDByBatchID", hexutil.EncodeBig(batchID)); err != nil {
        return nil, err
    }

    return res, nil
}
```

The result is unmarshaled into `*big.Int`, which expects a raw JSON number like `2`.

### 4. taiko-geth server returns *big.Int

```go
// taiko-geth/eth/taiko_api_backend.go:84-95
func (s *TaikoAPIBackend) LastBlockIDByBatchID(batchID *math.HexOrDecimal256) (*big.Int, error) {
    blockID, err := rawdb.ReadBatchToLastBlockID(s.eth.ChainDb(), (*big.Int)(batchID))
    if err != nil {
        return nil, err
    }
    if blockID != nil {
        return blockID, nil
    }

    return s.getLastBlockByBatchId((*big.Int)(batchID))
}
```

taiko-geth returns `*big.Int` which Go marshals as a raw number (e.g., `2`).

## Root Cause

[...]

PR #10075, opened 2025-12-31
Fix `taiko_lastBlockIDByBatchID` to return raw decimal instead of hex string

jmadibekov:
Fixes https://github.com/NethermindEth/nethermind/issues/10074

## Changes

- Add `BlockId` record struct with custom JSON converter that serializes as raw decimal number
- Update `taiko_lastBlockIDByBatchID` return type from `UInt256` to `BlockId?`
- The new converter uses `WriteNumberValue()` to output `2` instead of `"0x2"`

## Types of changes

#### Requires testing

- [x] Yes

- [x] Yes

#### Notes on testing

Verified by running taiko-client integration tests against Nethermind:

```bash
cd ~/code/surge-taiko-mono/packages/taiko-client
L2_NODE=l2_nmc PACKAGE=prover make test
```

- Before fix: `TestProverTestSuite/TestOnBatchProposed` fails with unmarshaling error
- After fix: All prover tests pass

- [ ] No

- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Taiko/Rpc/TaikoExtendedEthModule.cs
+++ b/src/Nethermind/Nethermind.Taiko/Rpc/TaikoExtendedEthModule.cs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 
 using System;
+using System.Text.Json.Serialization;
 using System.Threading.Tasks;
 using Nethermind.Blockchain.Find;
 using Nethermind.Blockchain.Synchronization;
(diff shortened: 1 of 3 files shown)


dipkakwani:
We have seen such issues in the past, and they were fixed from the Taiko's end, since RPCs do not match the standard eth API Geth's interface. Taiko uses `*big.Int` but it should use `hexutil.Big` as return type, but looks like they missed fixing `taiko_lastBlockIDByBatchID`

Examples: https://github.com/taikoxyz/taiko-mono/issues/20799
https://github.com/taikoxyz/taiko-mono/issues/20233

src/Nethermind/Nethermind.Taiko/Rpc/TaikoExtendedEthModule.cs
```diff
@@ -12,13 +13,27 @@
...
+[JsonConverter(typeof(BlockIdConverter))]
+public readonly record struct BlockId(long Value);
+
+public class BlockIdConverter : JsonConverter<BlockId>
+{
+    public override BlockId Read(ref System.Text.Json.Utf8JsonReader reader, Type t, System.Text.Json.JsonSerializerOptions o)
+        => new(reader.TokenType == System.Text.Json.JsonTokenType.Number ? reader.GetInt64() : long.Parse(reader.GetString()!));
+    public override void Write(System.Text.Json.Utf8JsonWriter writer, BlockId value, System.Text.Json.JsonSerializerOptions o)
+        => writer.WriteNumberValue(value.Value);
+}
```
LukaszRozmej, line 28: Please put it in its own file

PR #10108, opened 2026-01-05
Update GrpcClient.cs

forkfury:
## Fix

Fixes thread safety issue with `_connected` flag and prevents infinite loop in connection logic when gRPC channel fails to establish connection.

## Changes

- Made `_connected` field `volatile` to ensure proper visibility across threads
- Added timeout protection (30 seconds) in `TryStartAsync` to prevent infinite waiting
- Added check for `ChannelState.Shutdown` to detect terminal channel state early
- Added `System.Diagnostics` import for `Stopwatch` usage

The connection loop now throws `TimeoutException` or `InvalidOperationException` when appropriate, preventing the client from hanging indefinitely on connection failures.

### DIFF
--- a/src/Nethermind/Nethermind.Grpc/Clients/GrpcClient.cs
+++ b/src/Nethermind/Nethermind.Grpc/Clients/GrpcClient.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


forkfury:
> We are currently not using GRPc, maybe we should delete it...?

@LukaszRozmej what exactly do you wanna delete, the module? should I make a pr with it?

PR #10112, opened 2026-01-06
Don't queue prunes

benaadams:
## Changes

- Is both a race condition where multiple Prunes may be kicked off and also where regular pruning may be kicked off when another type of pruning is happening.
- The current implementation ensures correctness; as they will all wait on the lock; however we don't want to build up a queue of Pruning (behind the lock); so instead add an early exit if the lock is held

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
+++ b/src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
@@ -574,12 +574,17 @@ public sealed class TrieStore : ITrieStore, IPruningTrieStore
         TrieStoreState state = CaptureCurrentState();
         if ((_pruningStrategy.ShouldPruneDirtyNode(state) || _pruningStrategy.ShouldPrunePersistedNode(state)) && _pruningTask.IsCompleted)
         {
-            _pruningTask = Task.Run(TrySyncPrune);
+            _pruningTask = TrySyncPrune();
         }
     }
 
     private async Task TrySyncPrune()
     {
+        // Delay for 3 things:
+        // 1. Move to background thread
+        // 2. Allow kick off (block processing) some time to finish before starting to prune (prune in block gap)
+        // 3. If we are n+1 Task passing the .IsCompleted check but not going to be first to pass lock,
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -571,60 +571,90 @@ private TrieStoreState CaptureCurrentState()
 
     public void Prune()
     {
-        var state = CaptureCurrentState();
+        TrieStoreState state = CaptureCurrentState();
         if ((_pruningStrategy.ShouldPruneDirtyNode(state) || _pruningStrategy.ShouldPrunePersistedNode(state)) && _pruningTask.IsCompleted)
         {
-            _pruningTask = Task.Run(SyncPruneCheck);
+            _pruningTask = Task.Run(TrySyncPrune);
```
LukaszRozmej: `Task.Run` is redundant here?
```suggestion
            _pruningTask = TrySyncPrune;
```

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -571,60 +571,90 @@ private TrieStoreState CaptureCurrentState()
...
         }
     }
 
-    internal void SyncPruneCheck()
+    private async Task TrySyncPrune()
+    {
+        await Task.Delay(10);
```
LukaszRozmej, line 588: why `Delay`? maybe `Yield`?

LukaszRozmej (approved):

asdacap (approved):

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
benaadams: Done

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
benaadams: Added reasoning

PR #10113, opened 2026-01-06
Save space locally on builds

LukaszRozmej:
## Changes

- Add Directory.Build.targets (also to Solution Items)
- Skip copying dependencies and native runtimes for library projects (this already saves 50%! 30GB->15GB)
- Share native runtimes across projects during local builds with a shared location and junctions/symlinks
- All those changes depend on `SaveDiskSpace` variable set to `true`, so it is opt-in (you can set environment variable)

This could be adopted for CI if needed

- [x] Build-related changes

#### Requires testing

- [x] Yes

- [x] No

#### Notes on testing

Current github actions need to be checked/tested.

## Remarks

So when I build in Debug/Release I get about ~30GB in Artifacts. If you count Debug/Release it is 60GB. If you count that I have 3 copies of the repo due to working on different stuff semi-simultaneously, it is... 180GB!

This PR brings it down to 4GB (8x), so even x6 it is ~24GB.

### DIFF
--- a/src/Nethermind/Directory.Build.targets
+++ b/src/Nethermind/Directory.Build.targets
@@ -1,35 +1,28 @@
 <Project>
 
-  <!-- Skip copying dependencies and native runtimes for library projects. -->
-  <PropertyGroup Condition="'$(OutputType)' != 'Exe' and '$(OutputType)' != 'WinExe'">
-    <CopyLocalLockFileAssemblies>false</CopyLocalLockFileAssemblies>
-    <CopyLocalRuntimeTargetAssets>false</CopyLocalRuntimeTargetAssets>
+  <!-- Share native runtimes dir across projects during with a shared location and junctions/symlinks.
+    Enable with `-p:SaveDiskSpace=true` or an environment variable. -->
+  <PropertyGroup Condition="'$(SaveDiskSpace)' == 'true'">
+    <_SharedRuntimesDir>$(ArtifactsPath)\runtimes\$(Configuration)</_SharedRuntimesDir>
   </PropertyGroup>
 
-  <!-- Share native runtimes across projects during local builds with a shared location and junctions/symlinks.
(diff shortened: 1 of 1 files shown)


Directory.Build.targets
```diff
@@ -0,0 +1,56 @@
...
+      <RuntimeTargetsCopyLocalItems Remove="@(RuntimeTargetsCopyLocalItems)"
+          Condition="($([System.String]::Copy('%(Identity)').Contains('\runtimes\')) or $([System.String]::Copy('%(Identity)').Contains('/runtimes/'))) and
+                     !$([System.String]::Copy('%(Identity)').Contains('\runtimes\$(_NethermindRuntimePrefix)')) and
+                     !$([System.String]::Copy('%(Identity)').Contains('/runtimes/$(_NethermindRuntimePrefix)'))" />
+      <NativeCopyLocalItems Remove="@(NativeCopyLocalItems)"
+          Condition="($([System.String]::Copy('%(Identity)').Contains('\runtimes\')) or $([System.String]::Copy('%(Identity)').Contains('/runtimes/'))) and
+                     !$([System.String]::Copy('%(Identity)').Contains('\runtimes\$(_NethermindRuntimePrefix)')) and
+                     !$([System.String]::Copy('%(Identity)').Contains('/runtimes/$(_NethermindRuntimePrefix)'))" />
+    </ItemGroup>
+  </Target>
```
LukaszRozmej: This could potentially be dropped, it only saves about 200MB as below de-duplicates native libraries

LukaszRozmej: Ok I removed that, doesn't really make sense to have this complexity.

LukaszRozmej:
> > Would you be happy if I make it opt-in rather than opt-out? Like only do it if one specifies a variable.
> 
> That seems like a fair middle ground.

done

flcl42 (approved):

rubo:
> done

I'll check shortly

rubo:
https://github.com/NethermindEth/nethermind/pull/10122

PR #10117, opened 2026-01-06
Allow claude code to do stuff without asking all the time

LukaszRozmej:
## Changes

- Allow claude code to do stuff without asking all the time

- [x] Other: AI instrumentation for Claude

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/.claude/settings.json
+++ b/src/Nethermind/.claude/settings.json
@@ -1,0 +1,16 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(dotnet build:*)",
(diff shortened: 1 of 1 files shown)


emlautarom1:
I remember discussing some time ago that we generally avoid having editor specific configurations in this repo. It's fine by me, but I would expect to eventually have VSCode/Rider/Visual Studio specific and other AI tools specific-settings if we go this route.

rubo (commented):
I see no reason for having this in the repo. If it works fine for someone locally, great. Why push to upstream?
We can add this to `.gitignore`, so that it doesn't mess with other stuff.

PR #10120, opened 2026-01-06
Optimize EVM

benaadams:
## Changes

Major EVM performance optimization focusing on hot path execution, stack operations, memory handling, and transaction processing. This commit refactors core EVM components  to use SIMD intrinsics, ref-based stack access, and reduces allocations in critical paths. Also introduces a new JitAsm tooling for JIT disassembly analysis and instruction benchmarks.

  **EVM Stack (EvmStack.cs)**

  - Converted to ref partial struct with direct ref-based stack access (ref byte _stack) instead of Span<byte>
  - Added SIMD-optimized PushBytes using AVX2/AVX512 with vectorized byte packing
  - Specialized push methods (Push10Bytes, Push11Bytes, etc.) for common operand sizes
  - Added ReadUInt256FromSlot/WriteUInt256ToSlot with AVX2 shuffle and Avx512Vbmi.VL.PermuteVar32x8 support
  - Optimized partial byte packing with PackHiU64/PackLoU64 for non-aligned pushes
  - Return EvmExceptionType directly instead of throwing for overflow checks

  **Virtual Machine (VirtualMachine.cs)**

  - Split ProcessCallFrames into exception-handling wrapper and hot loop (ProcessCallFramesCore)
  - Used [SkipLocalsInit] throughout to reduce stack zeroing overhead
  - Separated precompile, top-level, and exception paths into [NoInlining] methods to reduce hot loop stack frame
  - Converted exception type from enum to inline bit checks (e.g., (executionType & ExecutionType.IsCreate) != 0)
  - Changed _previousCallResult from ReadOnlyMemory<byte>? to non-nullable ReadOnlyMemory<byte>
  - Added OpcodeResult return type for opcode dispatch

  **Transaction Processor (TransactionProcessor.cs)**

  - Added fast path for simple ETH transfers (non-contract, no auth list, no precompile)
  - Extracted EVM execution path to ExecuteEvm<TLogTracing> with [NoInlining]
  - Generic TLogTracing parameter eliminates runtime trace-level branching
  - Deferred StackAccessTracker creation until needed (slow path only)
  - Optimized UpdateMetrics with reciprocal multiplication and incremental averaging

  **Memory & Pooling (EvmPooledMemory.cs)**

  - Extracted Rent() to [NoInlining] method to reduce hot path overhead
  - Added CalculateMemoryCost(ulong) overload avoiding UInt256 boxing
  - Direct u0 field access instead of implicit cast for location/length
  - Extracted ComputeMemoryExpansionCost to separate method

  **Execution Type (ExecutionType.cs)**

  - Changed from plain enum to [Flags] enum with embedded type bits (IsCreate, IsCall, IsEof)
  - Enables single-instruction type checks via bitmask instead of multi-case switch

  **Instruction Handlers**

  - Refactored bitwise, call, control flow, math, and stack instructions to return OpcodeResult
  - Added inline gas policy via ref TGasPolicy parameter
  - Reduced redundant stack bounds checks by combining pop/push in single operations

  **New: PreWarmCache (PreWarmCache.cs)**

  - Lock-free set-associative cache with epoch-based O(1) Clear()
  - Uses parallel arrays for hash/key/value to avoid boxing
  - Spinlock-per-slot with Interlocked.CompareExchange

  **New: JitAsm Tool (tools/JitAsm/)**

  - CLI tool for JIT disassembly of specific methods
  - Supports method resolution via Type.Method or Assembly!Type.Method syntax
  - Includes tests for disassembly parsing, method resolution, and static ctor detection

  **Benchmarks**

  - Added Nethermind.Evm.InstructionsBenchmark project with benchmarks for arithmetic, bitwise, comparison, shifts, stack, and memory operations

  **Other**

  - Address.Memory property for direct ReadOnlyMemory<byte> access
  - CallResult refactored to use ReadOnlyMemory<byte> instead of Output wrapper
  - TransactionSubstate simplified output handling with OutputBytes property
  - Removed unused IHashSetEnumerableCollection and IToArrayCollection interfaces

- [x] Optimization

#### Requires testing

- [x] No

- [x] No

- [x] No


LukaszRozmej:
Any benchmarks?

src/Nethermind/Nethermind.Evm.SourceGenerators/StackPushBytesGenerator.cs
```diff
@@ -0,0 +1,718 @@
...
+                if (src is null)
+                    continue;
+
+                var hintName = $"{GetSafeHintName(type)}.StackPushBytes.g.cs";
+                spc.AddSource(hintName, src);
+            }
+        });
+    }
+
+    private static string? EmitForType(SourceProductionContext spc, INamedTypeSymbol? type, Candidate[] methods)
```
Demuirgos: isnt there a way to have a template? it would provide a clearer idea on the structure of the emitted type

benaadams:
@copilot summarise the changes for PR body text

src/Nethermind/Nethermind.Evm.SourceGenerators/StackPushBytesGenerator.cs
benaadams: Changed to use more interpolated strings; though you can view the output as it is generated for the compilation 

So 10 bytes

image

And 30 bytes

image

src/Nethermind/Nethermind.Evm.SourceGenerators/Nethermind.Evm.SourceGenerators.csproj
```diff
@@ -0,0 +1,13 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
rubo: AI?

benaadams: Needs to be `netstandard2.0` for roslyn plugin (source compilation) 😭

LukaszRozmej (commented):
The source generation seems overcomplicated to me. Can't we just do it by hand?

Looking at generated methods they all have similar pattern.

Top part is differs by one param - the count
```
        if (TTracingInst.IsActive)
        {
            _tracer.TraceBytes(in value, 16); // <- count goes here
        }

        uint headOffset = (uint)Head;
        uint newOffset = headOffset + 1;
        ref Vector256<byte> head = ref Unsafe.As<byte, Vector256<byte>>(ref Unsafe.Add(ref MemoryMarshal.GetReference(_bytes), (nint)(headOffset * WordSize)));
        if (newOffset >= MaxStackSize)
        {
            return EvmExceptionType.StackOverflow;
        }

        Head = (int)newOffset;
```

while bottom part is more complicated and has more variations, example:

```
        if (Vector256.IsHardwareAccelerated)
        {
            head = Vector256.Create(
                0UL,
                (ulong)Unsafe.ReadUnaligned<ushort>(ref value) << 48,
                Unsafe.ReadUnaligned<ulong>(ref Unsafe.Add(ref value, 2)),
                Unsafe.ReadUnaligned<ulong>(ref Unsafe.Add(ref value, 10))
            ).AsByte();
        }
        else
        {
            ref Vector128<ulong> head128 = ref Unsafe.As<Vector256<byte>, Vector128<ulong>>(ref head);

            head128 = Vector128.Create(
                0UL,
                (ulong)Unsafe.ReadUnaligned<ushort>(ref value) << 48
            );

            Unsafe.Add(ref head128, 1) = Vector128.Create(
                Unsafe.ReadUnaligned<ulong>(ref Unsafe.Add(ref value, 2)),
                Unsafe.ReadUnaligned<ulong>(ref Unsafe.Add(ref value, 10))
            );
        }
```

but can be broken down to simple stuff:
- for V256 - we create one vector that's it
- for V128 we basically interpret something as 128 then create 2 other 128's.

This code could be made generic statics with those and we could extract and inline all of this Something like:

```
    public partial EvmExceptionType PushBytes<TOp, TOpTTracingInst>(ref byte value)
        where TTracingInst : struct, global::Nethermind.Core.IFlag
        where TOp : IOpCount // or something else
    {
        if (TTracingInst.IsActive)
        {
            _tracer.TraceBytes(in value, TOp.Count);
        }

        uint headOffset = (uint)Head;
        uint newOffset = headOffset + 1;
        ref Vector256<byte> head = ref Unsafe.As<byte, Vector256<byte>>(ref Unsafe.Add(ref MemoryMarshal.GetReference(_bytes), (nint)(headOffset * WordSize)));
        if (newOffset >= MaxStackSize)
        {
            return EvmExceptionType.StackOverflow;
        }

        Head = (int)newOffset;

        if (Vector256.IsHardwareAccelerated)
        {
            head = TOp.Create256Vector();
       }
        else
        {
            Unsafe.Add(ref head128, 1) = TOp.Create128Vector()
        }

        return EvmExceptionType.None;
    }
```

pass the correct params, aggressive inline those and you are done without all the obfuscation of code generation

tools/JitAsm/Program.cs
```diff
@@ -0,0 +1,274 @@
...
+        if (type is not null) return type;
+
+        // Try searching all types
+        foreach (var t in assembly.GetTypes())
+        {
+            if (t.FullName == typeName || t.Name == typeName)
+            {
+                return t;
+            }
+        }
```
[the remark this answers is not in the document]
benaadams: No Linq

tools/JitAsm/MethodResolver.cs
```diff
@@ -0,0 +1,366 @@
...
+        // Try nested types
+        foreach (var t in types)
+        {
+            if (t.FullName is not null && typeName.StartsWith(t.FullName + "+"))
+            {
+                var nestedName = typeName[(t.FullName.Length + 1)..];
+                var nested = t.GetNestedType(nestedName, BindingFlags.Public | BindingFlags.NonPublic);
+                if (nested is not null) return nested;
+            }
+        }
```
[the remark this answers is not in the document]
benaadams: No Linq

tools/JitAsm/Program.cs
```diff
@@ -0,0 +1,274 @@
...
+
+            // Initialize static constructors if requested
+            foreach (var cctorTypeName in cctorsToInit)
+            {
+                var cctorType = ResolveType(assembly, cctorTypeName);
+                if (cctorType is not null)
+                {
+                    RuntimeHelpers.RunClassConstructor(cctorType.TypeHandle);
+                }
+            }
```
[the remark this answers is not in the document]
benaadams: No Linq

tools/JitAsm/StaticCtorDetector.cs
```diff
@@ -0,0 +1,130 @@
...
+
+        // Detect direct .cctor calls
+        foreach (Match match in CctorCallPattern().Matches(disassemblyOutput))
+        {
+            var typeName = NormalizeTypeName(match.Groups["type"].Value);
+            if (!string.IsNullOrEmpty(typeName))
+            {
+                detectedTypes.Add(typeName);
+            }
+        }
```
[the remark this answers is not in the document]
benaadams: No Linq

src/Nethermind/Nethermind.State.Test/PreWarmCacheTests.cs
```diff
@@ -0,0 +1,579 @@
...
+    }
+
+    [Test]
+    public void Concurrent_clears_during_operations_are_safe()
+    {
+        var cache = new PreWarmCache<StorageCell, byte[]>();
+        const int threadCount = 4;
+        const int operationsPerThread = 5_000;
+        var exceptions = new ConcurrentBag<Exception>();
+        var cts = new CancellationTokenSource();
```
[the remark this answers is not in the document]
benaadams: Meh

MarekM25 (commented):
Let's test it with some EVM fuzzers before merging

src/Nethermind/Nethermind.Core/Address.cs
```diff
@@ -20,23 +21,27 @@ namespace Nethermind.Core
     [JsonConverter(typeof(AddressConverter))]
     [TypeConverter(typeof(AddressTypeConverter))]
     [DebuggerDisplay("{ToString()}")]
-    public class Address : IEquatable<Address>, IComparable<Address>
+    public sealed class Address : MemoryManager<byte>, IEquatable<Address>, IComparable<Address>
```
LukaszRozmej: does it make sense to implement `MemoryManager`, I don't see any custom code for it?

src/Nethermind/Nethermind.Blockchain/CachedCodeInfoRepository.cs
```diff
@@ -21,19 +22,26 @@ public class CachedCodeInfoRepository(
...
+    private FrozenDictionary<AddressAsKey, PrecompileInfo> CachedPrecompiles { get; } = precompileCache is null
         ? precompileProvider.GetPrecompiles()
         : precompileProvider.GetPrecompiles().ToFrozenDictionary(kvp => kvp.Key, kvp => CreateCachedPrecompile(kvp, precompileCache));
 
     public ICodeInfo GetCachedCodeInfo(Address codeSource, bool followDelegation, IReleaseSpec vmSpec,
         out Address? delegationAddress)
     {
-        if (vmSpec.IsPrecompile(codeSource) && _cachedPrecompile.TryGetValue(codeSource, out var cachedCodeInfo))
+        PrecompileInfo cachedCodeInfo;
+        if (vmSpec.IsPrecompile(codeSource) && TryGetCachedPrecompile(codeSource, out cachedCodeInfo))
```
LukaszRozmej:
```suggestion
        if (vmSpec.IsPrecompile(codeSource) && TryGetCachedPrecompile(codeSource, out PrecompileInfo cachedCodeInfo))
```

src/Nethermind/Nethermind.Blockchain/CachedCodeInfoRepository.cs
```diff
@@ -21,19 +22,26 @@ public class CachedCodeInfoRepository(
...
             delegationAddress = null;
             return cachedCodeInfo;
         }
         return baseCodeInfoRepository.GetCachedCodeInfo(codeSource, followDelegation, vmSpec, out delegationAddress);
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        bool TryGetCachedPrecompile(Address codeSource, out PrecompileInfo cachedCodeInfo)
+        {
+            return CachedPrecompiles.TryGetValue(codeSource, out cachedCodeInfo);
+        }
```
LukaszRozmej: Where possible simplify?

```suggestion
        bool TryGetCachedPrecompile(Address codeSource, out PrecompileInfo cachedCodeInfo) =>
            CachedPrecompiles.TryGetValue(codeSource, out cachedCodeInfo);
```

src/Nethermind/Nethermind.Core/Extensions/SpanExtensions.cs
```diff
@@ -197,6 +200,42 @@ public static ArrayPoolListRef<T> ToPooledListRef<T>(this in ReadOnlySpan<T> spa
...
+            }
+            else // Arm.Aes.IsSupported
+            {
+                Vector128<byte> key = Unsafe.As<byte, Vector128<byte>>(ref start);
+                Vector128<byte> data = Unsafe.As<byte, Vector128<byte>>(ref Unsafe.Add(ref start, 16));
+                key ^= Vector128.CreateScalar(seed).AsByte();
+                Vector128<byte> mixed = Arm.Aes.MixColumns(Arm.Aes.Encrypt(data, key));
+                ulong compressed = mixed.AsUInt64().GetElement(0) ^ mixed.AsUInt64().GetElement(1);
+                return (int)(uint)(compressed ^ (compressed >> 32));
+            }
```
LukaszRozmej:
```suggestion
            Vector128<byte> key = Unsafe.As<byte, Vector128<byte>>(ref start);
            Vector128<byte> data = Unsafe.As<byte, Vector128<byte>>(ref Unsafe.Add(ref start, 16));
            key ^= Vector128.CreateScalar(seed).AsByte();
            Vector128<byte> mixed = x64.Aes.IsSupported ? x64.Aes.Encrypt(data, key) : Arm.Aes.MixColumns(Arm.Aes.Encrypt(data, key));            
            ulong compressed = mixed.AsUInt64().GetElement(0) ^ mixed.AsUInt64().GetElement(1);
            return (int)(uint)(compressed ^ (compressed >> 32));
```

src/Nethermind/Nethermind.Core/Extensions/SpanExtensions.cs
```diff
@@ -234,15 +274,39 @@ public static int FastHash(this ReadOnlySpan<byte> input)
...
+            {
+                Vector128<byte> key = Unsafe.As<byte, Vector128<byte>>(ref start);
+                Vector128<byte> data = Unsafe.As<byte, Vector128<byte>>(ref Unsafe.Add(ref start, 16));
+                // Mix in the instance-random seed
+                key ^= Vector128.CreateScalar(seed).AsByte();
+                // ARM needs explicit MixColumns for equivalent diffusion
+                Vector128<byte> mixed = Arm.Aes.MixColumns(Arm.Aes.Encrypt(data, key));
+                ulong compressed = mixed.AsUInt64().GetElement(0) ^ mixed.AsUInt64().GetElement(1);
+                return (int)(uint)(compressed ^ (compressed >> 32));
+            }
```
LukaszRozmej: same here!

src/Nethermind/Nethermind.Evm.InstructionsBenchmark/Arithmetic/AddSubMulBenchmark.cs
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    /// <summary>
+    /// Benchmark of just the pop operation for 2 UInt256 values.
+    /// </summary>
+    [Benchmark]
+    public (UInt256, UInt256) PopTwo()
+    {
+        SetupStack(in LargeA, in LargeB);
+        var stack = new EvmStack(2, ref _stackBuffer[0], _codeBuffer);
```
LukaszRozmej, line 129: Those 2 lines should probably be in some benchmark setup and not in benchmark itself

src/Nethermind/Nethermind.Evm.InstructionsBenchmark/Arithmetic/DivModBenchmark.cs
```diff
@@ -0,0 +1,130 @@
...
+    {
+        BenchmarkHelpers.WriteStackSlot(_stackBuffer, 0, in b);
+        BenchmarkHelpers.WriteStackSlot(_stackBuffer, 1, in a);
+    }
+
+    [Benchmark(Baseline = true)]
+    public UInt256 Div_SmallDivisor()
+    {
+        SetupStack(in Dividend, in SmallDivisor);
+        var stack = new EvmStack(2, ref _stackBuffer[0], _codeBuffer);
```
LukaszRozmej, line 45: move to benchmark setup

src/Nethermind/Nethermind.Evm.InstructionsBenchmark/Bitwise/AndOrXorNotBenchmark.cs
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    /// <summary>
+    /// Bitwise AND using vector operations (current implementation).
+    /// </summary>
+    [Benchmark(Baseline = true)]
+    public Word And_Vector()
+    {
+        SetupStack(in Pattern1, in Pattern2);
+        var stack = new EvmStack(2, ref _stackBuffer[0], _codeBuffer);
```
LukaszRozmej, line 58: same

src/Nethermind/Nethermind.Evm.InstructionsBenchmark/Bitwise/ByteBenchmark.cs
```diff
@@ -0,0 +1,148 @@
...
+        BenchmarkHelpers.WriteStackSlot(_stackBuffer, 0, value);
+        BenchmarkHelpers.ClearStackSlot(_stackBuffer, 1);
+        BenchmarkHelpers.WriteStackSlotU64(_stackBuffer, 1, position);
+    }
+
+    [Benchmark(Baseline = true)]
+    public byte Byte_Position0()
+    {
+        SetupByteStack(0, in TestValue);
+        var stack = new EvmStack(2, ref _stackBuffer[0], _codeBuffer);
```
LukaszRozmej, line 44: same

src/Nethermind/Nethermind.Evm.Test/VmStateTests.cs
```diff
@@ -217,7 +217,7 @@ public void Can_dispose_without_init()
         public void Can_dispose_after_init()
         {
             VmState<EthereumGasPolicy> vmState = CreateEvmState();
-            vmState.InitializeStacks();
+            vmState.InitializeStacks(null, default, out _);
```
LukaszRozmej: Should those be some kind of default values? Should we have `InitializeStacks` without them? (At least as extension method in tests?)

src/Nethermind/Nethermind.Evm/EvmObjectFormat/EofCodeValidator.cs
```diff
@@ -42,18 +42,17 @@ public static bool IsEof(ReadOnlyMemory<byte> container, [NotNullWhen(true)] out
     /// </summary>
     /// <param name="container">Machine code to be checked</param>
     /// <returns></returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
     public static bool IsEof(ReadOnlySpan<byte> container, [NotNullWhen(true)] out byte version)
     {
-        if (container.Length > MAGIC.Length)
+        // Direct byte comparison avoids StartsWith overhead and static cctor check for MAGIC property
+        if (container.Length > VERSION_OFFSET && container[0] == 0xEF && container[1] == 0x00)
```
LukaszRozmej: Maybe named const?
```suggestion
        if (container.Length > VERSION_OFFSET && container[0] == MAGIC_0 && container[1] == MAGIC_1)
```

src/Nethermind/Nethermind.Evm/GasPolicy/EthereumGasPolicy.cs
```diff
@@ -125,6 +124,17 @@ public static bool UpdateMemoryCost(ref EthereumGasPolicy gas,
...
 
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static bool UpdateMemoryCost(ref EthereumGasPolicy gas,
+        in UInt256 position,
+        ulong length, VmState<EthereumGasPolicy> vmState)
+    {
+        long memoryCost = vmState.Memory.CalculateMemoryCost(in position, length, out bool outOfGas);
+        if (memoryCost == 0L)
+            return !outOfGas;
+        return UpdateGas(ref gas, memoryCost);
```
LukaszRozmej, line 133:
```suggestion
        return  memoryCost == 0L ? !outOfGas : UpdateGas(ref gas, memoryCost);
```

src/Nethermind/Nethermind.Evm/GasPolicy/EthereumGasPolicy.cs
```diff
@@ -179,10 +189,10 @@ public static void OnAfterInstructionTrace(in EthereumGasPolicy gas) { }
     public static EthereumGasPolicy Max(in EthereumGasPolicy a, in EthereumGasPolicy b) =>
         a.Value >= b.Value ? a : b;
 
-    public static EthereumGasPolicy CalculateIntrinsicGas(Transaction tx, IReleaseSpec spec)
+    public static EthereumGasPolicy CalculateIntrinsicGas(Transaction tx, IReleaseSpec spec, long tokensInCallData)
```
LukaszRozmej, line 198: `tokensInCallData` not used?

src/Nethermind/Nethermind.Evm/VirtualMachine.Warmup.cs
```diff
@@ -71,39 +71,52 @@ public static void WarmUpEvmInstructions(IWorldState state, ICodeInfoRepository
         RunTransactions(processor, state, spec);
     }
 
+    private static readonly Address recipient1 = new("0x0000000000000000000000000000000010000000");
+    private static readonly Address recipient2 = new("0x0000000000000000000000000000000000000100");
```
LukaszRozmej, line 72: we don't need those to live forever, make them local variables

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -148,225 +149,352 @@ public unsafe partial class VirtualMachine<TGasPolicy>(
...
+            if (inCancelable)
+            {
+                ProcessCallFrames<TTracingInst, OffFlag, OnFlag>(spec, out substate);
+            }
+            else
             {
-                ReturnDataBuffer = Array.Empty<byte>();
+                ProcessCallFrames<TTracingInst, OffFlag, OffFlag>(spec, out substate);
             }
+        }
```
LukaszRozmej, line 216: switch?

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -148,225 +149,352 @@ public unsafe partial class VirtualMachine<TGasPolicy>(
...
+
+            if (callResult.IsException)
+            {
+                HandleException<TTracingActions>(currentState, in callResult);
+                goto ClearOutput;
+            }
+
+            VmState<TGasPolicy> previousState = currentState;
+            // Mark for disposal - outer method will dispose if exception occurs
+            _stateToDispose = previousState;
```
LukaszRozmej: Maybe just `return` it and remove the field altogether?

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -148,225 +149,352 @@ public unsafe partial class VirtualMachine<TGasPolicy>(
...
+                return;
+            }
+            catch (EvmException) { /* Handled after finally */ }
+            catch (OverflowException) { /* Handled after finally */ }
+            finally
+            {
+                // Dispose any state that was pending when exception occurred
+                _stateToDispose?.Dispose();
+                _stateToDispose = null;
+            }
```
LukaszRozmej, line 254:
```suggestion
        {
            {
                // Inner loop runs without exception handling - JIT can optimize aggressively
                using _ = ProcessCallFramesCore<TTracingInst, TTracingActions, TCancellable>(spec, out substate)
            }
```

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -148,225 +149,352 @@ public unsafe partial class VirtualMachine<TGasPolicy>(
...
+
+            // Recovery path - only reached after catching an exception
+            VmState<TGasPolicy> current = _currentState;
```
LukaszRozmej: no dispose here?

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -148,225 +149,352 @@ public unsafe partial class VirtualMachine<TGasPolicy>(
...
+        Unsafe.SkipInit(out substate);
+        callResult = ExecutePrecompile<TTracingActions>(currentState, out Exception? precompileFailure, out string? substateError);
+        if (precompileFailure is not null)
+        {
+            if (currentState.IsTopLevel)
             {
-                _currentState = null;
-                return failSubstate;
+                substate = HandleTopLevelPrecompileFailure<TTracingInst, TTracingActions>(currentState, precompileFailure, substateError);
+                return 1; // Return with substate
```
LukaszRozmej, line 416: make it an enum for clarity?

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -148,225 +149,352 @@ public unsafe partial class VirtualMachine<TGasPolicy>(
...
+        where TTracingInst : struct, IFlag
+        where TTracingActions : struct, IFlag
     {
-        _previousCallResult = previousState.Env.ExecutingAccount.Bytes;
+        if (callResult.IsException)
+        {
+            substate = HandleTopLevelException<TTracingActions>(currentState, in callResult);
+            return;
+        }
+        substate = PrepareTopLevelSubstate<TTracingActions>(currentState, in callResult);
```
LukaszRozmej:
```suggestion
        substate =callResult.IsException 
             ? HandleTopLevelException<TTracingActions>(currentState, in callResult) 
             : PrepareTopLevelSubstate<TTracingActions>(currentState, in callResult);
```

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -148,225 +149,352 @@ public unsafe partial class VirtualMachine<TGasPolicy>(
...
-        _previousCallResult = previousState.ExecutionType.IsAnyCallEof() ? EofStatusCode.SuccessBytes :
-            callResult.PrecompileSuccess.HasValue
-            ? (callResult.PrecompileSuccess.Value ? StatusCode.SuccessBytes : StatusCode.FailureBytes)
-            : StatusCode.SuccessBytes;
-        previousCallOutput = callResult.Output.Bytes.Span.SliceWithZeroPadding(0, Math.Min(callResult.Output.Bytes.Length, (int)previousState.OutputLength));
+        // ReturnDataBuffer is already set by DispatchExecution/RunPrecompile
+        ReadOnlySpan<byte> returnSpan = ReturnDataBuffer.Span;
+
+        // Determine status code - EOF uses different codes than legacy
+        bool isEof = previousState.ExecutionType.IsAnyCallEof();
```
LukaszRozmej: should we make eof unreachable for now - to be JITed out?

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -148,225 +149,352 @@ public unsafe partial class VirtualMachine<TGasPolicy>(
...
+    {
+        // Capture code to deploy before PrepareCreateData clears ReturnDataBuffer
+        ReadOnlyMemory<byte> codeToDeployMemory = ReturnDataBuffer;
+        PrepareCreateData(previousState);
+        return executionType.IsAnyCreateLegacy() ?
+            HandleLegacyCreate<TTracingActions>(
+                in callResult,
+                previousState,
+                gasAvailableForCodeDeposit,
+                codeToDeployMemory) :
```
LukaszRozmej: maybe already pass Span's here?

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -610,17 +759,20 @@ protected void HandleRevert(VmState<TGasPolicy> previousState, in CallResult cal
...
+    protected TransactionSubstate HandleTopLevelPrecompileFailure<TTracingInst, TTracingActions>(VmState<TGasPolicy> currentState, Exception failure, string? substateError)
         where TTracingInst : struct, IFlag
+        where TTracingActions : struct, IFlag
     {
+        _currentState = null;
         // Log the exception if trace logging is enabled.
         if (_logger.IsTrace)
         {
-            _logger.Trace($"exception ({failure.GetType().Name}) in {_currentState.ExecutionType} at depth {_currentState.Env.CallDepth} - restoring snapshot");
+            _logger.Trace($"exception ({failure.GetType().Name}) in {currentState.ExecutionType} at depth {currentState.Env.CallDepth} - restoring snapshot");
```
LukaszRozmej: Should those go to separate methods on hotpath? Like we have ThrowXXX to have LogXXX? So we can decrease the bytecode size and don't do inlining?

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -857,17 +1107,34 @@ protected void TraceTransactionActionStart(VmState<TGasPolicy> currentState)
...
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
     private void PrepareOpcodes<TTracingInst>(IReleaseSpec spec)
         where TTracingInst : struct, IFlag
     {
-        // Check if tracing instructions are inactive.
+        if (!TTracingInst.IsActive)
+        {
+            object? instructions = spec.EvmInstructionsNoTrace;
+            // Fast path: cache populated and past the PGO refresh window
+            if (instructions is not null && _txCount >= 500_000)
```
LukaszRozmej: make 500k a named const

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -990,70 +1257,85 @@ private CallResult RunPrecompile(VmState<TGasPolicy> state)
...
-                shouldRevert: !success,
-                exceptionType: !success ? EvmExceptionType.PrecompileFailure : EvmExceptionType.None
-            )
+            (ResultType result, byte[]? data, string? error) = precompile.Run(callData, spec);
+            ReturnDataBuffer = data ?? Array.Empty<byte>();
+            if (result == ResultType.Success)
+            {
+                return new CallResult(precompileSuccess: true, shouldRevert: false, EvmExceptionType.None);
+            }
+            return new CallResult(precompileSuccess: false, shouldRevert: true, EvmExceptionType.PrecompileFailure)
```
LukaszRozmej, line 1142:
```suggestion
            return result == ResultType.Success            
                ? new CallResult(precompileSuccess: true, shouldRevert: false, EvmExceptionType.None)
                : new CallResult(precompileSuccess: false, shouldRevert: true, EvmExceptionType.PrecompileFailure)
```

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -990,70 +1257,85 @@ private CallResult RunPrecompile(VmState<TGasPolicy> state)
...
+    {
+        // https://github.com/ethereum/EIPs/blob/master/EIPS/eip-161.md
+        // An additional issue was found in Parity,
+        // where the Parity client incorrectly failed
+        // to revert empty account deletions in a more limited set of contexts
+        // involving out-of-gas calls to precompiled contracts;
+        // the new Geth behavior matches Parity's,
+        // and empty accounts will cease to be a source of concern in general
+        // in about one week once the state clearing process finishes.
+        if (spec.ClearEmptyAccountWhenTouched &&
```
LukaszRozmej, line 1191:
```suggestion
        if (!_parityTouchBugAccount.ShouldDelete && 
            spec.ClearEmptyAccountWhenTouched &&
```

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -1204,173 +1500,243 @@ protected CallResult ExecuteCall<TTracingInst>(
     /// which minimizes overhead and allows aggressive inlining and compile-time optimizations.
     /// </remarks>
     [SkipLocalsInit]
-    protected virtual unsafe CallResult RunByteCode<TTracingInst, TCancelable>(
+    protected virtual CallResult RunByteCode<TTracingInst, TCancellable>(
```
LukaszRozmej: Is there a point for this method? (virtuallity?)

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -1204,173 +1500,243 @@ protected CallResult ExecuteCall<TTracingInst>(
...
+            opCodeCount++;
+
+            // Dispatch via function pointer (calli). The handler returns the next PC and exception classification.
+            result = opcodeMethods[(nuint)(byte)instruction](this, ref stack, ref gas, result.ProgramCounter + 1);
+
+            // Call gas policy hook after instruction execution.
+            TGasPolicy.OnAfterInstructionTrace(in gas);
+
+            // If gas is exhausted, jump to the out-of-gas handler.
+            if (TGasPolicy.GetRemainingGas(in gas) < 0)
```
LukaszRozmej, line 1542: Should `OnAfterInstructionTrace` return remaining gas?

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -1204,173 +1500,243 @@ protected CallResult ExecuteCall<TTracingInst>(
...
+                // Ideally all the opcodes would return OutOfGas when they exhaust gas,
+                // so we could remove this check. Is a safety net.
+                goto OutOfGas;
+            }
+
+            // If an exception occurred, the loop will naturally exit due to the packed Value field.
+
+            // Trace completion is only meaningful for the non-throwing/non-terminating path.
+            if (TTracingInst.IsActive && result.Exception == EvmExceptionType.None)
+                EndInstructionTrace(TGasPolicy.GetRemainingGas(in gas));
```
LukaszRozmej, line 1553: we already called GetRemainingGas above?

src/Nethermind/Nethermind.State/TrieStoreScopeProvider.cs
```diff
@@ -115,18 +115,19 @@ public IWorldStateScopeProvider.IWorldStateWriteBatch StartWriteBatch(int estima
...
         {
             using var blockCommitter = _scopeProvider._trieStore.BeginBlockCommit(blockNumber);
-
-            // Note: These all runs in about 0.4ms. So the little overhead like attempting to sort the tasks
-            // may make it worst. Always check on mainnet.
             using ArrayPoolList<Task> commitTask = new ArrayPoolList<Task>(_storages.Count);
-            foreach (KeyValuePair<AddressAsKey, StorageTree> storage in _storages)
+            // Commit storages with most writes first to minimise waiting
+            foreach (KeyValuePair<AddressAsKey, StorageTree> storage in _storages
+                .Where(kv => kv.Value.OutstandingWritesEstimate > 0)
```
LukaszRozmej, line 121: Is the `OutstandingWritesEstimate` reliable? If it is just an estimate can we filter out something crucial here?

@asdacap ?

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Call.cs
```diff
@@ -128,102 +132,140 @@ public static EvmExceptionType InstructionCall<TGasPolicy, TOpCall, TTracingInst
...
 
-        // If gasLimit exceeds the host's representable range, treat as out-of-gas.
-        if (gasLimit >= long.MaxValue) goto OutOfGas;
-
-        long gasLimitUl = (long)gasLimit;
         if (!TGasPolicy.UpdateGas(ref gas, gasLimitUl)) goto OutOfGas;
 
         // Add call stipend if value is being transferred.
-        if (!transferValue.IsZero)
+        bool tracingRefunds = vm.TxTracer.IsTracingRefunds;
```
LukaszRozmej: not  `TTracingRefunds`?

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Environment.cs
```diff
@@ -833,17 +817,22 @@ public static EvmExceptionType InstructionBlockHash<TGasPolicy, TTracingInst>(Vi
...
 
         // Push the block hash bytes if available; otherwise, push a 32-byte zero value.
-        stack.PushBytes<TTracingInst>(blockHash is not null ? blockHash.Bytes : BytesZero32);
-
-        // If block hash tracing is enabled and a valid block hash was obtained, report it.
-        if (vm.TxTracer.IsTracingBlockHash && blockHash is not null)
+        if (blockHash is not null)
+        {
+            // If block hash tracing is enabled and a valid block hash was obtained, report it.
+            if (vm.TxTracer.IsTracingBlockHash)
```
LukaszRozmej, line 766: `TTracingBlockHash`?

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Shifts.cs
```diff
@@ -50,37 +49,36 @@ public interface IOpShift
...
         }
         else
         {
             // Otherwise, pop the value to be shifted.
             if (!stack.PopUInt256(out UInt256 b)) goto StackUnderflow;
-            // Perform the shift operation using the specific implementation.
-            TOpShift.Operation(in a, in b, out UInt256 result);
-            stack.PushUInt256<TTracingInst>(in result);
+            // Perform the shift operation directly with int shift amount (no UInt256 construction).
+            TOpShift.Operation((int)a, in b, out UInt256 result);
```
LukaszRozmej, line 76: cast?

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Stack.cs
```diff
@@ -88,46 +93,46 @@ public struct Op1 : IOpCount
...
-    public struct Op2 : IOpCount { public static int Count => 2; }
+    public struct Op2 : IOpCount
+    {
+        public static int Count => 2;
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static EvmExceptionType Push<TTracingInst>(int length, ref EvmStack stack, int programCounter)
+            where TTracingInst : struct, IFlag
+        {
+            throw new NotSupportedException($"Use the {nameof(InstructionPush2)} opcode instead");
```
LukaszRozmej, line 118: why this way?

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Storage.cs
```diff
@@ -109,21 +111,22 @@ public static EvmExceptionType InstructionTStore<TGasPolicy>(VirtualMachine<TGas
         vm.WorldState.SetTransientState(in storageCell, !bytes.IsZero() ? bytes.ToArray() : BytesZero32);
 
         // If storage tracing is enabled, retrieve the current stored value and log the operation.
-        if (vm.TxTracer.IsTracingStorage)
+        ITxTracer txTracer = vm.TxTracer;
+        if (txTracer.IsTracingStorage)
```
LukaszRozmej, line 110: why not `TTracingStorage`?

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Storage.cs
```diff
@@ -366,7 +378,7 @@ internal static EvmExceptionType InstructionSStoreUnmetered<TGasPolicy, TTracing
         StorageCell storageCell = new(vmState.Env.ExecutingAccount, in result);
 
         // Charge gas based on whether this is a cold or warm storage access.
-        if (!TGasPolicy.ConsumeStorageAccessGas(ref gas, in vmState.AccessTracker, vm.TxTracer.IsTracingAccess, in storageCell, StorageAccessType.SSTORE, spec))
+        if (!TGasPolicy.ConsumeStorageAccessGas(ref gas, in vmState.AccessTracker, txTracer.IsTracingAccess, in storageCell, StorageAccessType.SSTORE, spec))
```
LukaszRozmej: TTracingAccess?

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Storage.cs
```diff
@@ -452,12 +461,15 @@ internal static EvmExceptionType InstructionSStoreMetered<TGasPolicy, TTracingIn
         if (vmState.IsStatic) goto StaticCallViolation;
 
         IReleaseSpec spec = vm.Spec;
+        ITxTracer txTracer = vm.TxTracer;
+        bool isTracingRefunds = txTracer.IsTracingRefunds;
+        bool isTracingAccess = txTracer.IsTracingAccess;
```
LukaszRozmej, line 459: TTracingRefunds?
TTracingAccess?

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Storage.cs
```diff
@@ -595,6 +592,25 @@ private static void TraceSstore<TGasPolicy>(VirtualMachine<TGasPolicy> vm, bool
...
+    /// <summary>
+    /// Updates the refund counter and optionally reports the refund change to the tracer.
+    /// Extracted to reduce register pressure in the main SSTORE hot path.
+    /// </summary>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    private static void UpdateRefund<TGasPolicy>(VmState<TGasPolicy> vmState, ITxTracer txTracer, long delta, bool isTracingRefunds)
+        where TGasPolicy : struct, IGasPolicy<TGasPolicy>
+    {
+        vmState.Refund += delta;
+        if (isTracingRefunds)
```
LukaszRozmej, line 598: TTracingRefunds?

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Storage.cs
```diff
@@ -655,21 +671,30 @@ internal static EvmExceptionType InstructionSLoad<TGasPolicy, TTracingInst>(Virt
...
+        if (!result.IsUint64 || result.u0 >= (uint)span.Length)
+        {
+            return new(programCounter, stack.PushZero<TTracingInst>());
+        }
+        else
+        {
+            uint offset = (uint)result.u0;
+            uint available = (uint)span.Length - offset;
+            uint copiedLength = available >= 32 ? 32 : available;
+            return new(programCounter, stack.PushRightPaddedBytes<TTracingInst>(ref Unsafe.Add(ref MemoryMarshal.GetReference(span), offset), copiedLength));
```
LukaszRozmej:
```suggestion
        ReadOnlySpan<byte> span = vm.VmState.Env.InputData.Span;
        uint spanLength = (uint)span.Length;
        if (!result.IsUint64 || result.u0 >= spanLength)
        {
            return new(programCounter, stack.PushZero<TTracingInst>());
        }
        else
        {
            uint offset = (uint)result.u0;
            uint copiedLength = spanLength - offset;
            if (copiedLength > 32) copiedLength = 32;
            return new(programCounter, stack.PushRightPaddedBytes<TTracingInst>(ref Unsafe.Add(ref MemoryMarshal.GetReference(span), offset), copiedLength));
```

src/Nethermind/Nethermind.State/OverridableEnv/OverridableCodeInfoRepository.cs
```diff
@@ -55,13 +55,33 @@ public void MovePrecompile(IReleaseSpec vmSpec, Address precompileAddr, Address
...
         codeInfoRepository.SetDelegation(codeSource, authority, spec);
 
-    public bool TryGetDelegation(Address address, IReleaseSpec vmSpec,
-        [NotNullWhen(true)] out Address? delegatedAddress)
+    public bool TryGetDelegation(Address address, IReleaseSpec vmSpec, out ICodeInfo codeInfo, [NotNullWhen(true)] out Address? delegatedAddress)
     {
         delegatedAddress = null;
+
+        // Check precompile overrides first (same order as GetCachedCodeInfo)
+        if (_precompileOverrides.TryGetValue(address, out var precompile))
```
LukaszRozmej, line 63: no var pelase

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -164,19 +164,35 @@ public virtual TransactionResult Warmup(Transaction transaction, ITxTracer txTra
...
+
+                TransactionResult result = Execute<OnFlag>(tx, tracer, opts);
+                if (Logger.IsTrace) Logger.Trace($"Tx {tx.Hash} was executed, {result}");
+                return result;
             }
+            else
+            {
+                if (tx.IsSystem() || opts == ExecutionOptions.SkipValidation)
+                {
+                    _systemTransactionProcessor ??= new SystemTransactionProcessor<TGasPolicy>(_blobBaseFeeCalculator, SpecProvider, WorldState, VirtualMachine, _codeInfoRepository, _logManager);
```
LukaszRozmej: hide behind `SystemTransactionProcessor` property to lower the bytecode lenght?

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -260,25 +280,11 @@ protected virtual TransactionResult Execute(Transaction tx, ITxTracer tracer, Ex
...
-                }
-                else
-                {
-                    LogEntry[] logs = substate.Logs.Count != 0 ? substate.Logs.ToArray() : [];
-                    tracer.MarkAsSuccess(env.ExecutingAccount, spentGas, substate.Output.Bytes.ToArray(), logs, stateRoot);
-                }
+                TraceReceipt(tracer, spec, executingAccount, spentGas, statusCode, in substate);
             }
 
+            if (disposeAccessTracker) accessTracker.Dispose();
```
LukaszRozmej: why needed?

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -290,6 +296,110 @@ protected virtual TransactionResult CalculateAvailableGas(Transaction tx, in Int
...
+            in IntrinsicGas<TGasPolicy> intrinsicGas,
+            in UInt256 premiumPerGas,
+            in UInt256 blobBaseFee,
+            TGasPolicy gasAvailable,
+            ref StackAccessTracker accessTracker,
+            ref bool disposeAccessTracker,
+            out int statusCode,
+            out TransactionSubstate substate,
+            out GasConsumed spentGas,
+            out Address? executingAccount)
```
LukaszRozmej, line 357: not to many reference parameters?

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -290,6 +296,110 @@ protected virtual TransactionResult CalculateAvailableGas(Transaction tx, in Int
...
+            Unsafe.SkipInit(out spentGas);
+            executingAccount = tx.To;
+
+            int delegationRefunds = (!spec.IsEip7702Enabled || !tx.HasAuthorizationList) ? 0 : ProcessDelegations(tx, spec, accessTracker);
+
+            TransactionResult result;
+            if (!(result = BuildExecutionEnvironment(tx, spec, _codeInfoRepository, accessTracker, out ExecutionEnvironment e)))
+            {
+                accessTracker.Dispose();
+                disposeAccessTracker = false;
```
LukaszRozmej: Can we ignore it and always dispose in parent scope?

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -371,18 +481,27 @@ bool IsValidForExecution(
...
         {
             if (opts is ExecutionOptions.Commit or ExecutionOptions.None && (effectiveGasPrice[2] | effectiveGasPrice[3]) == 0)
             {
-                float gasPrice = (float)((double)effectiveGasPrice / 1_000_000_000.0);
+                // Fast path: most gas prices fit in 64 bits (< ~18 ETH/gas)
+                // Use multiplication by reciprocal instead of division
+                const float ToGwei = 1e-9f;
+                float gasPrice = effectiveGasPrice[1] == 0
+                    ? effectiveGasPrice[0] * ToGwei
+                    : (float)((double)effectiveGasPrice * 1e-9);
```
LukaszRozmej, line 524: Could we have a `TMetrics` switch? Could potentially help in zk-context.
@maximmenshikov

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -659,19 +785,98 @@ private TransactionResult BuildExecutionEnvironment(
...
+        {
+            // Transfer: subtract from sender, add to recipient (skip if self-transfer)
+            if (tx.SenderAddress != recipient)
+            {
+                PayValue(tx, spec, opts);
+                WorldState.AddToBalanceAndCreateIfNotExists(recipient, tx.Value, spec);
+            }
+
+            // Maintain action tracing if enabled
+            if (tracer.IsTracingActions)
```
LukaszRozmej, line 859: TTracingActions?

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -659,19 +785,98 @@ private TransactionResult BuildExecutionEnvironment(
...
+            {
+                TraceAction(tx, recipient, tracer, gasAvailable);
+            }
+
+            if (tracer.IsTracingAccess)
```
LukaszRozmej, line 864: TTracingAccess?

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -743,35 +947,47 @@ private int ExecuteEvmCall<TTracingInst>(
...
+                long destroyRefund = isTracingRefunds ? RefundOf.Destroy(spec.IsEip3529Enabled) : 0;
+
+                foreach (Address toBeDestroyed in destroyList)
+                {
+                    if (TLogTracing.IsActive) Logger.Trace($"Destroying account {toBeDestroyed}");
+
+                    WorldState.ClearStorage(toBeDestroyed);
+                    WorldState.DeleteAccount(toBeDestroyed);
+
+                    if (isTracingRefunds)
```
LukaszRozmej, line 1016: TTraceRefunds?

src/Nethermind/Nethermind.Evm/EvmStack.cs
```diff
@@ -61,25 +77,102 @@ private ref Word PushedHead()
...
+        nuint q = length >> 3;
+        nuint r = length & 7;
+
+        ulong partial = r == 0 ? 0UL : PackHiU64(ref src, r);
+
+        ref byte p = ref Unsafe.Add(ref src, (int)r);
+
+        Vector128<ulong> lo, hi;
+
+        if (q == 0)
```
LukaszRozmej, line 120: switch?

src/Nethermind/Nethermind.Evm/EvmStack.cs
```diff
@@ -100,175 +193,1306 @@ public void PushBytes<TTracingInst>(scoped in ZeroPaddedSpan value)
...
+        nuint q = length >> 3; // full 8-byte chunks: 0..3
+        nuint r = length & 7;  // remainder: 0..7
+
+        // The partial bytes (if any) live at src + 8*q
+        ref byte tail = ref Unsafe.Add(ref src, (int)(q << 3));
+        ulong partial = r == 0 ? 0UL : PackLoU64(ref tail, r);
+
+        Vector128<ulong> lo, hi;
+
+        if (q == 0)
```
LukaszRozmej, line 243: switch?

src/Nethermind/Nethermind.Evm/EvmStack.cs
```diff
@@ -100,175 +193,1306 @@ public void PushBytes<TTracingInst>(scoped in ZeroPaddedSpan value)
...
+
+    [SkipLocalsInit]
+    private static EvmExceptionType PushBytesPartialZeroPadded(ref byte dst, ref byte src, nuint length)
+    {
+        nuint q = length >> 3; // full 8-byte chunks: 0..3
+        nuint r = length & 7;  // remainder: 0..7
+
+        // The partial bytes (if any) live at src + 8*q
+        ref byte tail = ref Unsafe.Add(ref src, (int)(q << 3));
+        ulong partial = r == 0 ? 0UL : PackLoU64(ref tail, r);
```
LukaszRozmej, line 239: handle underflow in `PackLoU64`
```suggestion
        ulong partial = PackLoU64(ref tail, r);
```

src/Nethermind/Nethermind.Evm/EvmStack.cs
```diff
@@ -61,25 +77,102 @@ private ref Word PushedHead()
...
+        }
+    }
+
+    [SkipLocalsInit]
+    private static void PushBytesPartial(ref byte dst, ref byte src, nuint length)
+    {
+        nuint q = length >> 3;
+        nuint r = length & 7;
+
+        ulong partial = r == 0 ? 0UL : PackHiU64(ref src, r);
```
LukaszRozmej, line 114: Handle underflow in PackHiU64
```suggestion
        ulong partial = PackHiU64(ref src, r);
```

src/Nethermind/Nethermind.Evm/EvmStack.cs
```diff
@@ -61,25 +77,102 @@ private ref Word PushedHead()
...
+            0 => (ulong)src << 56,
+            1 => (ulong)Unsafe.ReadUnaligned<ushort>(ref src) << 48,
+            2 => ((ulong)Unsafe.ReadUnaligned<ushort>(ref src) << 40) |
+               ((ulong)Unsafe.Add(ref src, 2) << 56),
+            3 => (ulong)Unsafe.ReadUnaligned<uint>(ref src) << 32,
+            4 => ((ulong)Unsafe.ReadUnaligned<uint>(ref src) << 24) |
+               ((ulong)Unsafe.Add(ref src, 4) << 56),
+            5 => ((ulong)Unsafe.ReadUnaligned<uint>(ref src) << 16) |
+               ((ulong)Unsafe.ReadUnaligned<ushort>(ref Unsafe.Add(ref src, 4)) << 48),
+            _ => ((ulong)Unsafe.ReadUnaligned<uint>(ref src) << 8) |
```
LukaszRozmej, line 169: nuint.MaxValue underflow handling here

src/Nethermind/Nethermind.Evm/EvmStack.cs
```diff
@@ -100,175 +193,1306 @@ public void PushBytes<TTracingInst>(scoped in ZeroPaddedSpan value)
...
+            0 => (ulong)src,
+            1 => Unsafe.ReadUnaligned<ushort>(ref src),
+            2 => (ulong)Unsafe.ReadUnaligned<ushort>(ref src) |
+               ((ulong)Unsafe.Add(ref src, 2) << 16),
+            3 => Unsafe.ReadUnaligned<uint>(ref src),
+            4 => (ulong)Unsafe.ReadUnaligned<uint>(ref src) |
+               ((ulong)Unsafe.Add(ref src, 4) << 32),
+            5 => (ulong)Unsafe.ReadUnaligned<uint>(ref src) |
+               ((ulong)Unsafe.ReadUnaligned<ushort>(ref Unsafe.Add(ref src, 4)) << 32),
+            _ => (ulong)Unsafe.ReadUnaligned<uint>(ref src) |
```
LukaszRozmej, line 299: nuint.MaxValue underflow handling here

src/Nethermind/Nethermind.Evm/EvmStack.cs
```diff
@@ -100,175 +193,1306 @@ public void PushBytes<TTracingInst>(scoped in ZeroPaddedSpan value)
...
+        // Zero entire word with single vector store, then fill non-zero lanes with scalar stores.
+        // This avoids expensive vpinsrq + vinserti128 dependency chain.
+        head = default;
+        ref ulong headU64 = ref Unsafe.As<Word, ulong>(ref head);
+        Unsafe.Add(ref headU64, 2) = (ulong)Unsafe.ReadUnaligned<ushort>(ref value) << 48;
+        Unsafe.Add(ref headU64, 3) = Unsafe.ReadUnaligned<ulong>(ref Unsafe.Add(ref value, 2));
 
-        // Single 32-byte store
-        head = CreateWordFromUInt64(lane3);
+        return EvmExceptionType.None;
```
LukaszRozmej, line 442: All those methods are very similar with basically difference in only 2 last `Unsafe` calls. Can we make them reusable somehow? Maybe with `TPushX` structs?

src/Nethermind/Nethermind.Evm/EvmStack.cs
```diff
@@ -428,27 +1699,386 @@ public bool PopUInt256(out UInt256 result)
...
+                ulong b0 = Unsafe.ReadUnaligned<ulong>(ref Unsafe.Add(ref bytes, 24));
+                ulong a0 = Unsafe.ReadUnaligned<ulong>(ref Unsafe.Add(ref bytes, 56));
+
+                b = new UInt256(b0, b1, b2, b3);
+                a = new UInt256(a0, a1, a2, a3);
+            }
+        }
+
+        return true;
+    }
```
LukaszRozmej, line 1798: Can we do here just 2x PopUInt256Unchecked?

src/Nethermind/Nethermind.Evm/EvmStack.cs
```diff
@@ -428,27 +1699,386 @@ public bool PopUInt256(out UInt256 result)
...
+        if (newHead < 0)
+        {
+            return false;
+        }
+        Head = newHead;
+
+        ref byte bytes = ref Unsafe.Add(ref _stack, (nint)((uint)newHead * WordSize));
+        // Memory layout: [c @ +0] [b @ +32] [a @ +64]
+
+        if (Avx2.IsSupported)
```
LukaszRozmej, line 1825: Here do 3x PopUInt256Unchecked?

src/Nethermind/Nethermind.Evm/EvmStack.cs
```diff
@@ -428,27 +1699,386 @@ public bool PopUInt256(out UInt256 result)
...
+        if (newHead < 0)
+        {
+            return false;
+        }
+        Head = newHead;
+
+        ref byte bytes = ref Unsafe.Add(ref _stack, (nint)((uint)newHead * WordSize));
+        // Memory layout: [d @ +0] [c @ +32] [b @ +64] [a @ +96]
+
+        if (Avx2.IsSupported)
```
LukaszRozmej, line 1949: 4x PopUInt256Unchecked?

src/Nethermind/Nethermind.Evm/IntrinsicGasCalculator.cs
```diff
@@ -41,8 +41,10 @@ public static class IntrinsicGasCalculator
     public static IntrinsicGas<TGasPolicy> Calculate<TGasPolicy>(Transaction transaction, IReleaseSpec releaseSpec)
         where TGasPolicy : struct, IGasPolicy<TGasPolicy>
     {
-        TGasPolicy standard = TGasPolicy.CalculateIntrinsicGas(transaction, releaseSpec);
-        long floorCost = CalculateFloorCost(transaction, releaseSpec);
+        // Compute tokensInCallData once - CountZeros uses SIMD
+        long tokensInCallData = CalculateTokensInCallData(transaction, releaseSpec);
```
LukaszRozmej: should it be ulong?

src/Nethermind/Nethermind.State/PreWarmCache.cs
```diff
@@ -0,0 +1,187 @@
...
+    where TValue : class?
+{
+    /// <summary>
+    /// Number of cache entries. Must be a power of 2 for mask operations.
+    /// </summary>
+    private const int Count = 1 << 14; // 16384
+
+    private const int BucketMask = Count - 1;
+
+    // Bit layout: [Lock:1][Epoch:12][Hash:18][Occupied:1]
```
LukaszRozmej: wouldn't it be safer to use 64bit to avoid hash collisions better?

src/Nethermind/Nethermind.State/PreWarmCache.cs
```diff
@@ -0,0 +1,187 @@
...
+namespace Nethermind.State;
+
+/// <summary>
+/// A high-performance one-way set-associative cache similar to <see cref="Nethermind.Core.Crypto.KeccakCache"/>.
+/// Uses parallel arrays for hash codes, keys, and values to avoid boxing.
+/// Supports reference type values and O(1) Clear() via epoch-based invalidation.
+/// </summary>
+/// <typeparam name="TKey">The key type (must be a struct implementing IEquatable)</typeparam>
+/// <typeparam name="TValue">The value type (must be a reference type)</typeparam>
+public sealed class PreWarmCache<TKey, TValue>
```
LukaszRozmej, line 19: How much difference does it make compared to ConcurrentDictionary?
What are the potential drawbacks?

tools/JitAsm/MethodResolver.cs
```diff
@@ -0,0 +1,366 @@
...
+
+using System.Reflection;
+
+namespace JitAsm;
+
+internal sealed class MethodResolver(Assembly assembly)
+{
+    private static readonly Dictionary<string, Type> TypeAliases = new(StringComparer.OrdinalIgnoreCase)
+    {
+        ["bool"] = typeof(bool),
```
LukaszRozmej, line 12: use nameof?

LukaszRozmej (commented):
Benchmark setups should be fixed.
One liner methods to use `=>` - would save a lot of lines.
The excess use of pattern `where T : struct, IFlag` will cause issues in AOT, can you advise @maximmenshikov how to handle it in AOT so it generates multiple version of this - basically removing the generic?
Ask EF testing team to do fuzzing tests!

LukaszRozmej (commented):
Benchmark setups should be fixed.
One liner methods to use `=>` - would save a lot of lines.
The excess use of pattern `where T : struct, IFlag` will cause issues in AOT, can you advise @maximmenshikov how to handle it in AOT so it generates multiple version of this - basically removing the generic?
Ask EF testing team to do fuzzing tests!

src/Nethermind/Nethermind.Runner/runtimeconfig.template.json
```diff
@@ -2,6 +2,7 @@
   "configProperties": {
     "System.Net.Http.EnableActivityPropagation": false,
     "System.Diagnostics.Tracing.EventSource.IsSupported": false,
-    "System.Diagnostics.Metrics.Meter.IsSupported": false
+    "System.Diagnostics.Metrics.Meter.IsSupported": false,
+    "System.Runtime.TieredCompilation.CallCountingDelayMs": 50
```
LukaszRozmej, line 7: context?

src/Nethermind/Ethereum.Test.Base/BlockchainTestBase.cs
```diff
@@ -223,6 +223,7 @@ protected async Task<EthereumTestResult> RunTest(BlockchainTest test, Stopwatch?
 
             // Caches are cleared async, which is a problem as read for the MainWorldState with prewarmer is not correct if its not cleared.
             preWarmer?.ClearCaches();
+            preWarmer?.ClearStateCaches();
```
LukaszRozmej: Should it be part of `ClearCaches`?

LukaszRozmej:
Cross-block cache is definitely risky. Would be better as next-PR.
I would prefer we started rigorous testing process and proceed to merging the PR rather than bloat it more.

stdevMac:
@benaadams stale-PR check: this mega-PR is being re-landed piecemeal (#11804 and friends), and the scope/regression feedback was never resolved here. OK to close in favor of the extractions? If there's no reply in the next few days, we'll close this as part of the stale-PR cleanup — it can always be reopened or resubmitted.

benaadams:
Superceded by https://github.com/NethermindEth/nethermind/pull/13128

PR #10123, opened 2026-01-06
fix: block history pruning getting stuck

### LINKED ISSUE #10119: Node is trying to remove one more block (at least in logs)
It seems that Nethermind node with auto history pruning it keeps trying to remove one block over and over again (at least in logs) and can't finish automatic pruning. Reported by Osin from Obol

```
Pruning historical blocks up to timestamp 1736164775 (#21565278). Estimated 1 blocks will be deleted. 
06 Jan 11:59:49 | Historical block pruning in progress... Deleted 0 blocks, with 1 remaining. 
06 Jan 11:59:49 | Completed pruning operation up to timestamp 1736164775. Deleted 1 blocks up to #21565278.
```

hirako2000:
Fixes #10119

## Changes

Adds block to history pruner to handle the edge case where some block was already pruned, to move the delete pointer.

- [x] Bugfix
- [ ] New feature
- [ ] Breaking change
- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

All existing unit tests for HistoryPruner still pass.
Added a test to reproduce the issue.

Running the full tests suites locally yields non deterministic errors, but all appear unrelated to this changeset.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks

I suspect far many more than 1 block were skipping pruning if my understanding of the bug is right, it may be that the loop returning the blocks to prune may have stopped at the first already pruned block. Thus potentially causing no more pruning at any subsequent run: stuck on infinite attempt to prune a ghost block. _If so Release notes may require an explanation_.

### DIFF
--- a/src/Nethermind/Nethermind.History/HistoryPruner.cs
+++ b/src/Nethermind/Nethermind.History/HistoryPruner.cs
@@ -148,7 +148,7 @@ public class HistoryPruner : IHistoryPruner
                     {
                         int attempts = 0;
                         long from = _cutoffPointer.Value;
-                        using ArrayPoolListRef<(long Number, Hash256 Hash, Block? Block)> x = GetBlocksByNumber(from, to, b =>
+                        using ArrayPoolListRef<Block> x = GetBlocksByNumber(from, to, b =>
                         {
                             if (attempts >= MaxOptimisticSearchAttempts)
                             {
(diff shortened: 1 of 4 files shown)


src/Nethermind/Nethermind.History/HistoryPruner.cs
```diff
@@ -458,6 +458,18 @@ private void PruneBlocksAndReceipts(ulong? cutoffTimestamp, CancellationToken ca
                 deletedBlocks++;
                 Metrics.BlocksPruned++;
             }
+
+            if (CutoffBlockNumber is not null)
```
Marchhill, line 462: no longer relevant

Marchhill:
obsolete after refactoring

src/Nethermind/Nethermind.History/HistoryPruner.cs
hirako2000: Thanks for the comment, I closed this PR due to lack of further assistance to get it merged.

I believe the bug is well present, affecting a minority of deployment that may restart the nodes and have a somewhat out of sync tracking of what got purged at previous runs.

PR #10134, opened 2026-01-07
Add some guards to RetryCache

flcl42:
## Changes

- Don't use `PooledSet`s
- Clean up on dispose

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No


src/Nethermind/Nethermind.TxPool/RetryCache.cs
```diff
@@ -12,68 +12,101 @@
...
+                            finally
+                            {
+                                _handlerBagsPool.Return(requests);
+                            }
                         }
                     }
                 }
+                catch (Exception ex)
+                {
+                    if (_logger.IsError) _logger.Error($"Unexpected error in {nameof(TResourceId)} retry cache loop", ex);
```
LukaszRozmej, line 96: clear everything

src/Nethermind/Nethermind.TxPool/RetryCache.cs
```diff
@@ -12,68 +12,101 @@
...
     where TResourceId : struct, IEquatable<TResourceId>
 {
     private readonly int _timeoutMs;
+    private readonly CancellationToken _token;
     private readonly int _checkMs;
+    private readonly int _expiringQueueLimit;
 
-    private readonly ConcurrentDictionary<TResourceId, PooledSet<IMessageHandler<TMessage>>> _retryRequests = new();
+    private static readonly ObjectPool<ConcurrentDictionary<IMessageHandler<TMessage>, bool>> _handlerBagsPool = new DefaultObjectPool<ConcurrentDictionary<IMessageHandler<TMessage>, bool>>(new ConcurrentBagPolicy<IMessageHandler<TMessage>, bool>());
+    private readonly ConcurrentDictionary<TResourceId, ConcurrentDictionary<IMessageHandler<TMessage>, bool>> _retryRequests = new();
```
LukaszRozmej: ConcurrentHashSet or List with lock()

src/Nethermind/Nethermind.TxPool/RetryCache.cs
```diff
@@ -12,68 +12,101 @@
...
                                 {
-                                    retryHandler.HandleMessage(TMessage.New(item.ResourceId));
+                                    _requestingResources.Set(item.ResourceId);
                                 }
-                                catch (Exception ex)
+
+                                if (_logger.IsTrace) _logger.Trace($"Sending retry requests for {item.ResourceId} after timeout");
+
+
+                                foreach (IMessageHandler<TMessage> retryHandler in requests.Keys)
```
LukaszRozmej: Can't this still have race condition with other access? It would be fine but we are pooling the collection, so not really? We can probably still use collection that was returned to pool?

src/Nethermind/Nethermind.TxPool/RetryCache.cs
```diff
@@ -85,13 +118,13 @@ public AnnounceResult Announced(TResourceId resourceId, IMessageHandler<TMessage
...
-                return [];
-            }, (resourceId, dict) =>
+                return _handlerBagsPool.Get();
+            }, (resourceId, requests) =>
             {
                 if (_logger.IsTrace) _logger.Trace($"Announced {resourceId} by {retryHandler}: UPDATE");
 
-                dict.Add(retryHandler);
-                return dict;
+                requests[retryHandler] = true;
```
LukaszRozmej: IMO we should limit the number of retryHandlers per resource to like max 8

src/Nethermind/Nethermind.TxPool/RetryCache.cs
```diff
@@ -12,68 +12,101 @@
...
+                {
+                    if (_logger.IsError) _logger.Error($"Unexpected error in {nameof(TResourceId)} retry cache loop", ex);
+                }
             }
+
+            if (_logger.IsDebug) _logger.Debug($"{nameof(TResourceId)} retry cache stopped");
         }, token);
     }
 
     public AnnounceResult Announced(TResourceId resourceId, IMessageHandler<TMessage> retryHandler)
```
LukaszRozmej:
```suggestion
    public AnnounceResult Announced(in TResourceId resourceId, IMessageHandler<TMessage> retryHandler)
```

src/Nethermind/Nethermind.TxPool/RetryCache.cs
LukaszRozmej:
```suggestion
    public void Received(in TResourceId resourceId)
```

LukaszRozmej (changes requested):

src/Nethermind/Nethermind.TxPool/RetryCache.cs
flcl42: `ConcurrentHashSet` looks great

src/Nethermind/Nethermind.TxPool/RetryCache.cs
flcl42: Replaced with other collections, but Keys are a safe snapshot

src/Nethermind/Nethermind.TxPool/RetryCache.cs
flcl42: Done

src/Nethermind/Nethermind.Runner/packages.lock.json
```diff
@@ -1157,6 +1157,7 @@
         "type": "Project",
         "dependencies": {
           "Collections.Pooled": "[1.0.82, )",
```
LukaszRozmej: Can we remove `Collections.Pooled`?

src/Nethermind/Nethermind.TxPool/Nethermind.TxPool.csproj
```diff
@@ -7,6 +7,7 @@
   <ItemGroup>
     <PackageReference Include="NonBlocking" />
     <PackageReference Include="Collections.Pooled" />
```
LukaszRozmej: Remove?

src/Nethermind/Nethermind.TxPool/RetryCache.cs
```diff
@@ -12,68 +13,104 @@
...
+                        _expiringQueue.TryDequeue(out item);
 
+                        if (_retryRequests.TryRemove(item.ResourceId, out ConcurrentHashSet<IMessageHandler<TMessage>>? requests))
+                        {
 
-                            foreach (IMessageHandler<TMessage> retryHandler in requests)
+                            try
                             {
-                                try
+                                if (requests.Count > 0)
```
LukaszRozmej: Try avoiding calling `Count` on `ConcurrentHashSet` as it is very slow - it has to take all locks on the collection.

Maybe better to do below:

src/Nethermind/Nethermind.TxPool/RetryCache.cs
```diff
@@ -12,68 +13,104 @@
...
                                 }
-                                catch (Exception ex)
+
+                                if (_logger.IsTrace) _logger.Trace($"Sending retry requests for {item.ResourceId} after timeout");
+
+
+                                foreach (IMessageHandler<TMessage> retryHandler in requests)
                                 {
-                                    if (_logger.IsTrace) _logger.Error($"Failed to send retry request to {retryHandler} for {item.ResourceId}", ex);
+                                    try
```
LukaszRozmej:
```suggestion
                                    set |= _requestingResources.Set(item.ResourceId)
                                    try
```

src/Nethermind/Nethermind.TxPool/RetryCache.cs
```diff
@@ -85,36 +122,67 @@ public AnnounceResult Announced(TResourceId resourceId, IMessageHandler<TMessage
...
+                if (requests.Count < _maxRetryRequests)
+                {
+                    requests.Add(retryHandler);
+                }
+
+                return requests;
             });
 
-            return added ? AnnounceResult.New : AnnounceResult.Enqueued;
+            return added ? AnnounceResult.RequestRequired : AnnounceResult.Delayed;
```
LukaszRozmej: instead of `added` use `result` of type `AnnounceResult`

src/Nethermind/Nethermind.TxPool/RetryCache.cs
```diff
@@ -85,36 +122,67 @@ public AnnounceResult Announced(TResourceId resourceId, IMessageHandler<TMessage
...
+        {
+            _handlerBagsPool.Return(requests);
+        }
+
+        _retryRequests.Clear();
+        _expiringQueue.Clear();
+        _requestingResources.Clear();
+    }
+
+    public void Dispose() => Clear();
```
LukaszRozmej: IMO we should await first the Task spawned in constructor, we can make it AsyncDIsposable

src/Nethermind/Nethermind.TxPool/RetryCache.cs
```diff
@@ -85,36 +122,67 @@ public AnnounceResult Announced(TResourceId resourceId, IMessageHandler<TMessage
...
+    private void Clear()
+    {
+        foreach (ConcurrentHashSet<IMessageHandler<TMessage>> requests in _retryRequests.Values)
+        {
+            _handlerBagsPool.Return(requests);
+        }
+
+        _retryRequests.Clear();
+        _expiringQueue.Clear();
+        _requestingResources.Clear();
```
LukaszRozmej: This order is IMO slightly more safe - clear the queue first.
```suggestion
        _expiringQueue.Clear();
        _requestingResources.Clear();
        
        foreach (ConcurrentHashSet<IMessageHandler<TMessage>> requests in _retryRequests.Values)
        {
            _handlerBagsPool.Return(requests);
        }
        
        _retryRequests.Clear();
```

src/Nethermind/Nethermind.TxPool/RetryCache.cs
```diff
@@ -12,68 +13,104 @@
...
 
-    public AnnounceResult Announced(TResourceId resourceId, IMessageHandler<TMessage> retryHandler)
+    public AnnounceResult Announced(in TResourceId resourceId, IMessageHandler<TMessage> retryHandler)
     {
+        if (_token.IsCancellationRequested)
+        {
+            return AnnounceResult.RequestRequired;
+        }
+
+        if (_expiringQueue.Count > _expiringQueueLimit)
```
LukaszRozmej: We could do:

```
if (_retryRequests.TryGet(out set)) { set.Add; return Delayed; }
else return AnnounceResult.RequestRequired;
```

src/Nethermind/Nethermind.TxPool/TxPool.cs
LukaszRozmej: This is probably not needed now as won't pass to the `RetryCache` in the first place.
```suggestion
```

LukaszRozmej, line 538:
```suggestion
```

LukaszRozmej, line 498: Won't it be best to do it ASAP here?
```suggestion
_retryCache.Received(tx.Hash!);
```

LukaszRozmej: If not ASAP above, why do we have `if (accepted != AcceptTxResult.Invalid)` here?

src/Nethermind/Nethermind.TxPool/RetryCache.cs
flcl42: Used IsEmpty

src/Nethermind/Nethermind.TxPool/RetryCache.cs
flcl42: Added atomic counter

src/Nethermind/Nethermind.TxPool/TxPool.cs
flcl42: Transaction may appear without blobs, so we need to hope another peer will return it with blobs

src/Nethermind/Nethermind.Runner/packages.lock.json
```diff
@@ -1156,7 +1156,8 @@
       "nethermind.txpool": {
         "type": "Project",
         "dependencies": {
-          "Collections.Pooled": "[1.0.82, )",
+          "Collections.Pooled": "(, )",
```
LukaszRozmej: Do we still need it?

src/Nethermind/Nethermind.TxPool/RetryCache.cs
```diff
@@ -12,109 +13,193 @@
...
+                                    {
+                                        try
+                                        {
+                                            retryHandler.HandleMessage(TMessage.New(item.ResourceId));
+                                        }
+                                        catch (Exception ex)
+                                        {
+                                            if (_logger.IsTrace) _logger.Error($"Failed to send retry request to {retryHandler} for {item.ResourceId}", ex);
+                                        }
+                                    }
```
LukaszRozmej: It still requires locks:

```
  public bool IsEmpty
  {
    get
    {
      if (!this.AreAllBucketsEmpty())
        return false;
      int locksAcquired = 0;
      try
      {
        this.AcquireAllLocks(ref locksAcquired);
        return this.AreAllBucketsEmpty();
      }
      finally
      {
        this.ReleaseLocks(0, locksAcquired);
      }
    }
  }
```

Alternative without it:
```
                                    bool set = false;
                                    foreach (IMessageHandler<TMessage> retryHandler in requests)
                                    {
                                        if (!set)
                                        {
                                            _requestingResources.Set(item.ResourceId);
                                            set = true;
                                            if (_logger.IsTrace) _logger.Trace($"Sending retry requests for {item.ResourceId} after timeout");
                                        }

                                        try
                                        {
                                            retryHandler.HandleMessage(TMessage.New(item.ResourceId));
                                        }
                                        catch (Exception ex)
                                        {
                                            if (_logger.IsTrace) _logger.Error($"Failed to send retry request to {retryHandler} for {item.ResourceId}", ex);
                                        }
                                    }
```

flcl42: Omg, fixed

src/Nethermind/Nethermind.Runner/packages.lock.json
flcl42: Nop

src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -577,7 +576,10 @@ private void TryConvertProofVersion(Transaction tx)
             }
         }
 
-        public AnnounceResult AnnounceTx(ValueHash256 txhash, IMessageHandler<PooledTransactionRequestMessage> retryHandler) => _retryCache.Announced(txhash, retryHandler);
+        public AnnounceResult NoitifyAboutTx(Hash256 hash, IMessageHandler<PooledTransactionRequestMessage> retryHandler) =>
```
marcindsobczak, line 579: I'm not a big fan of this renaming, but if you decide to keep it, please fix typo.

flcl42: Yeah, I failed to push the fix before merge, will be in another pr

PR #10139, opened 2026-01-07
fix: dispose JsonDocument for batch JSON-RPC requests

Snezhkko:
Previously, successful batch JSON-RPC requests never disposed the JsonDocument instance created during parsing. Single requests and error/limit paths correctly attached JsonDocument.Dispose() via JsonRpcResponse.AddDisposable, but the normal batch path only registered disposal for the ArrayPoolList of requests. As a result, the pooled JsonDocument buffers were effectively leaked, increasing GC and memory pressure under sustained batch load. This change attaches JsonDocument.Dispose() to the JsonRpcBatchResult lifetime so that the document is released when the batch result is disposed

### DIFF
--- a/src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
+++ b/src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
@@ -184,6 +184,8 @@ public class JsonRpcProcessor : IJsonRpcProcessor
                 }
                 catch (Exception ex)
                 {
+                    jsonDocument?.Dispose();
+                    collection?.Dispose();
                     deserializationFailureResult = GetParsingError(startTime, in buffer, "Error during parsing/validation.", ex);
                 }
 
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
LukaszRozmej, line 177: What about here?

LukaszRozmej, line 183: What about here?

LukaszRozmej, line 195: What about here?

LukaszRozmej, line 222: What about `collection` here and in previous comments?

Snezhkko: > What about `collection` here and in previous comments?

Added disposal in the generic `catch` so we don’t leak `JsonDocument` / `collection` when DeserializeObjectOrArray throws, and removed the extra AddDisposable(() => collection.Dispose()) from the batch path — the `collection` is already disposed in IterateRequest’s finally, now both it and `JsonDocument` are disposed exactly once on every path.

PR #10148, opened 2026-01-08
Override default CLI option alias

### LINKED ISSUE #10147: More readable plugin command line options
We're facing an issue while developing [Grandine.NethermindPlugin]. Grandine CL has quite a lot of long arguments. For example:

```
--verify-configuration-file
--terminal-total-difficulty-override
--terminal-block-hash-override
--terminal-block-hash-activation-epoch-override
```

To provide these arguments via nethermind plugin, we define them in the interface:

```csharp
namespace Grandine.NethermindPlugin;

using Nethermind.Config;

public interface IGrandineConfig : IConfig
{
    /* ... */

    [ConfigItem(Description = "Verify that configuration matches YAML_FILE")]
    public string? VerifyConfigurationFile { get; set; }

    [ConfigItem(Description = "Override TERMINAL_TOTAL_DIFFICULTY")]
    public string? TerminalTotalDifficultyOverride { get; set; }

    [ConfigItem(Description = "Override TERMINAL_BLOCK_HASH")]
    public string? TerminalBlockHashOverride { get; set; }

    [ConfigItem(Description = "Override TERMINAL_BLOCK_HASH_ACTIVATION_EPOCH")]
    public string? TerminalBlockHashActivationEpochOverride { get; set; }

    /* ... */
}
```

But defining options in such way produces unreadable CLI argument variants. Take a look at `nethermind --help` output:

```
  --grandine-verifyconfigurationfile, --Grandine.VerifyConfigurationFile <value>                                                    Verify that configuration matches YAML_FILE
  --grandine-terminaltotaldifficultyoverride, --Grandine.TerminalTotalDifficultyOverride <value>                                    Override TERMINAL_TOTAL_DIFFICULTY
  --grandine-terminalblockhashoverride, --Grandine.TerminalBlockHashOverride <value>                                                Override TERMINAL_BLOCK_HASH
  --grandine-terminalblockhashactivationepochoverride, --Grandine.TerminalBlockHashActivationEpochOverride <value>                  Override TERMINAL_BLOCK_HASH_ACTIVATION_EPOCH
```

It would be very convenient if every CLI argument would match grandine's CLI argument, just with `--grandine` prefix:

```
--grandine-verify-configuration-file
--grandine-terminal-total-difficulty-override
--grandine-terminal-block-hash-override
--grandine-terminal-block-hash-activation-epoch-override
```

That lowercase argument variants are properly converted from `PascalCase` to `kebab-case` by default. However, that would be a breaking change for all plugins, so this probably not an option.

Here are all alternatives I can think of, listed from most to least preferable:

1. An attribute for whole configuration interface.
   It would be nice to have some kind of attribute that can be placed on top of configuration interface, just like [serde(rename_all)]:
   ```csharp
   [RenameAllCliOptions]
   public interface IGrandineConfig : IConfig 
   {
       /* All configuration options are defined the same as before */
   }
   ```
   Maybe not worth to introduce new attribute, but better to use existing [ConfigCategory]?
   ```csharp
   [ConfigCategory(RenameAllCliOptions = true)]
   public interface IGrandineConfig : IConfig 
   {
       /* All configuration options are defined the same as before */
   }
   ```
2. Option to override default CLI argument name in ConfigItem attribute.
   ```csharp
   [ConfigItem(CliOptionName = "--verify-configuration-file")]
   public string? VerifyConfigurationFile { get; set; }
   ```
3. Option to define additional CLI argument aliases in ConfigItem attribute.
   ```csharp
   [ConfigItem(CliAlias = "--verify-configuration-file")]
   public string? VerifyConfigurationFile { get; set; }
   ```
   This is the least preferable, as it will create third alias to the option, thus increasing amount of information, displayed in help message.

**Additional context**
We're generating Config implementation automatically, so difference between proposed 1 and 2 options are not a big deal for us.

[Grandine.NethermindPlugin]: https://github.com/grandinetech/grandine/tree/develop/bindings/csharp/Grandine.NethermindPlugin
[serde(rename_all)]: https://serde.rs/attr-rename.html
[ConfigCategory]: https://github.com/NethermindEth/nethermind/blob/f1fc64a22294bf0b523a07b6be84df27e5563b79/src/Nethermind/Nethermind.Config/ConfigCategoryAttribute.cs#L8

ArtiomTr:
Resolves #10147

## Changes

- Implemented additional property on ConfigItemAttribute, that allows to override default lowercase CLI option with provided one.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [ ] No

Not sure, probably not?

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

Probably would be good to mention this in plugin development docs, not sure.

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Config/ConfigItemAttribute.cs
+++ b/src/Nethermind/Nethermind.Config/ConfigItemAttribute.cs
@@ -3,25 +3,22 @@
 
 using System;
 
-namespace Nethermind.Config
+namespace Nethermind.Config;
+
+[AttributeUsage(AttributeTargets.Property)]
+public class ConfigItemAttribute : Attribute
 {
-    public class ConfigItemAttribute : Attribute
-    {
-        public string Description { get; set; }
+    public string Description { get; set; }
 
-        public string DefaultValue { get; set; }
+    public string DefaultValue { get; set; }
 
-        public bool HiddenFromDocs { get; set; }
+    public bool HiddenFromDocs { get; set; }
 
-        public bool DisabledForCli { get; set; }
+    public bool DisabledForCli { get; set; }
 
-        public string EnvironmentVariable { get; set; }
+    public string EnvironmentVariable { get; set; }
 
-        public bool IsPortOption { get; set; }
+    public bool IsPortOption { get; set; }
 
-        /// <summary>
-        /// Overrides the auto-generated kebab-case CLI option name.
-        /// </summary>
-        public string CliOptionName { get; set; }
-    }
+    public string CliOptionAlias { get; set; }
 }
--- a/src/Nethermind/Nethermind.Runner/Program.cs
+++ b/src/Nethermind/Nethermind.Runner/Program.cs
@@ -213,10 +213,13 @@ async Task<int> RunAsync(ParseResult parseResult, PluginLoader pluginLoader, Can
 
 void AddConfigurationOptions(Command command)
 {
-    static Option CreateOption<T>(string name, Type configType, string? cliOptionName) =>
-        new Option<T>(
-            $"--{ConfigExtensions.GetCategoryName(configType)}.{name}",
-            cliOptionName is not null ? cliOptionName : $"--{ConfigExtensions.GetCategoryName(configType)}-{name}".ToLowerInvariant());
+    static Option CreateOption<T>(Type configType, string name, string? alias)
+    {
+        var category = ConfigExtensions.GetCategoryName(configType);
+        alias = string.IsNullOrWhiteSpace(alias) ? name : alias;
+
+        return new Option<T>($"--{category}.{name}", $"--{category}-{alias}".ToLowerInvariant());
+    }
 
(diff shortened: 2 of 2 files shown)


rubo (commented):
I'd love we've gone this way from the very beginning, but it is what it is. Also, I'd like the following changes:

- When the `CliOptionName` (shorten to `CliOption`) is specified, it is the only name for the given configuration parameter. The less naming variants we have, the better. Currently we have two, as the lowercase variant has been added later, so we kept the older one not to break users' setups.
- No need to include `--` in the name, it must be added later automatically along with enforced lowercase.
- In the #10147 you want to a have a prefix but your changes don't have that. Forgot or changed your mind?

This said, if you don't mind, I can apply the above changes myself. Also, we need to update our doc generator to reflect this changes in the docs. Thanks.

rubo (approved):

flcl42 (approved):

PR #10154, opened 2026-01-08
perf: eliminate redundant allocations in SendData byte normalization

0xxFloki:
Optimizes byte data normalization in `BlobSender.SendData()` by replacing LINQ-based transformation with a pre-allocated buffer approach.

### DIFF
--- a/tools/SendBlobs/BlobSender.cs
+++ b/tools/SendBlobs/BlobSender.cs
@@ -204,9 +204,18 @@ internal class BlobSender
         bool waitForInclusion,
         IReleaseSpec spec)
     {
-        data = data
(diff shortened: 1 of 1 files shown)


tools/SendBlobs/BlobSender.cs
```diff
@@ -204,9 +204,18 @@ public async Task SendData(
...
+        for (int i = 0; i < data.Length; i++)
+        {
+            byte value = data[i];
+            if (i % 32 == 0 && value >= 0x73)
+            {
+                normalized.Add(32);
+            }
+            normalized.Add(value);
+        }
+        data = normalized.ToArray();
```
LukaszRozmej, line 218: In theory we could do 2 scans, one to calculate length and the other to do actual normalization. This way we could just allocate array and avoid allocating list.

But this is just BlobSender, so probably performance and GC pressure are not that crucial

PR #10156, opened 2026-01-08
fix: gRPC client reconnect to shutdown previous channel

radik878:
Ensure GrpcClient properly shuts down the existing gRPC channel before creating a new one during reconnect. This avoids leaking channels and aligns the client-side lifecycle with recommended gRPC resource management practices.

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain/Receipts/PersistentReceiptStorage.cs
+++ b/src/Nethermind/Nethermind.Blockchain/Receipts/PersistentReceiptStorage.cs
@@ -223,7 +223,12 @@ namespace Nethermind.Blockchain.Receipts
                 return true;
             }
 
-            var result = CanGetReceiptsByHash(blockNumber);
+            if (!CanGetReceiptsByHash(blockNumber))
+            {
+                iterator = new ReceiptsIterator();
+                return false;
+            }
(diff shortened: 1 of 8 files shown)


src/Nethermind/Nethermind.Grpc/Clients/GrpcClient.cs
```diff
@@ -136,6 +136,7 @@ await stream.ResponseStream.MoveNext(cancellationToken))
         private async Task TryReconnectAsync()
         {
             _connected = false;
+            await (_channel?.ShutdownAsync() ?? Task.CompletedTask);
```
LukaszRozmej:
```suggestion
            await Task StopAsync();
```

src/Nethermind/Nethermind.Grpc/Clients/GrpcClient.cs
```diff
@@ -135,7 +135,7 @@ await stream.ResponseStream.MoveNext(cancellationToken))
 
         private async Task TryReconnectAsync()
         {
-            _connected = false;
+            await Task StopAsync();
```
LukaszRozmej:
```suggestion
            await StopAsync();
```

PR #10158, opened 2026-01-09
fix(sync): use INetworkConfig constructor for SyncPeerPool DI registration

Savid:
## Changes

- Fix SyncPeerPool DI registration to use INetworkConfig constructor instead of constructor with default parameters
- SyncPeerPool.PeerMaxCount now correctly reads from Network.MaxActivePeers config instead of always defaulting to 100

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

Manually verified that SyncPeerPool.PeerMaxCount now matches Network.MaxActivePeers config value. Previously, the DI container was selecting the constructor with default parameter `peersMaxCount = 100`, causing the config to be ignored.

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

SyncPeerPool now correctly respects the configured `Network.MaxActivePeers` value. Previously, it always used a hardcoded default of 100 peers regardless of configuration due to incorrect DI constructor selection.

## Remarks

Root cause: SyncPeerPool has two constructors - one taking INetworkConfig and one with [default parameters (`peersMaxCount = 100`)](https://github.com/NethermindEth/nethermind/blob/master/src/Nethermind/Nethermind.Synchronization/Peers/SyncPeerPool.cs#L78). The `.AddSingleton<SyncPeerPool>()` registration allowed Autofac to pick a constructor automatically, and it selected the one with defaults, bypassing INetworkConfig entirely.

Fix: Changed to explicit factory registration that specifies INetworkConfig as a dependency.

### DIFF
--- a/src/Nethermind/Nethermind.Synchronization.Test/SynchronizerModuleTests.cs
+++ b/src/Nethermind/Nethermind.Synchronization.Test/SynchronizerModuleTests.cs
@@ -8,9 +8,11 @@ using Nethermind.Config;
 using Nethermind.Core;
 using Nethermind.Core.Test.Builders;
 using Nethermind.Core.Test.Modules;
+using Nethermind.Network.Config;
 using Nethermind.State;
 using Nethermind.Synchronization.FastSync;
 using Nethermind.Synchronization.ParallelSync;
+using Nethermind.Synchronization.Peers;
 using NSubstitute;
 using NUnit.Framework;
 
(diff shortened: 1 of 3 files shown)


MarekM25 (commented):
Is it possible to write the test?

asdacap (approved):

PR #10159, opened 2026-01-09
Optimize success response of engine_getBlobsV2 (and slow down non-success)

marcindsobczak:
## Changes

- remove the logic for checking number of available blobs and quick failing if not all blobs available
It will slow our response down a lot if some blob is missing (and we return empty), but success responses should be slightly faster

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No


benaadams:
Superseded by #10453

PR #9950, opened 2025-12-16
Comments on Xdc README

wgr523:
## Changes

`src/Nethermind/Nethermind.Xdc/README.md`

#### What types of changes does your code introduce?

- [ ] Bugfix (a non-breaking change that fixes an issue)
- [ ] New feature (a non-breaking change that adds functionality)
- [ ] Breaking change (a change that causes existing functionality not to work as expected)
- [ ] Optimization
- [ ] Refactoring
- [x] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

#### Requires documentation update

- [ ] Yes
- [x] No

PR #10160, opened 2026-01-09
Comments and changes from on XDC README 

ak88:
Resolves #9950

Some updates to our README from XDC team.

Replied to comments in #9950

### DIFF
--- a/src/Nethermind/Nethermind.Xdc/README.md
+++ b/src/Nethermind/Nethermind.Xdc/README.md
@@ -559,7 +559,7 @@ XdcBlockHeader {
     Hash256 ParentHash
     Address Beneficiary
     Hash256 StateRoot
-    long Number // in Golang, it's `*big.Int`. Is it okay to use `long` here?
+    long Number
     ulong Timestamp
     ...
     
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Xdc/README.md
```diff
@@ -568,7 +559,7 @@ XdcBlockHeader {
     Hash256 ParentHash
     Address Beneficiary
     Hash256 StateRoot
-    long Number
+    long Number // in Golang, it's `*big.Int`. Is it okay to use `long` here?
```
cicr99: comment is not removed here

src/Nethermind/Nethermind.Xdc/README.md
```diff
@@ -404,7 +395,7 @@ Phase 4: QC AGGREGATION
 Phase 5: QC COMMITMENT
 ───────────────────────
 ┌────────────────────────────────┐
-│ CommitCertificate()            │
+│ CommitQC()                     │
```
cicr99, line 398: I believe they didn't agree with the use of "commit" here and suggested "process" instead . Anyway, on the Happy Path chart below, it has been changed to Process QC, so I would say to decide on one and have it both be the same

PR #10161, opened 2026-01-09
Remove duplicate GetContentLength calls in receipt encoder

splinter012:
Compute GetContentLength once in ReceiptArrayStorageDecoder.Encode and reuse for both compact and legacy branches, keep behavior identical while avoiding redundant work and extra allocations, no other code paths touched; lints clean on modified file

### DIFF
--- a/src/Nethermind/Nethermind.Serialization.Rlp/ReceiptArrayStorageDecoder.cs
+++ b/src/Nethermind/Nethermind.Serialization.Rlp/ReceiptArrayStorageDecoder.cs
@@ -80,9 +80,11 @@ public sealed class ReceiptArrayStorageDecoder(bool compactEncoding = true) : Rl
             return;
         }
 
-        if (compactEncoding && (rlpBehaviors & RlpBehaviors.Storage) != 0)
+        bool useCompactEncoding = compactEncoding && (rlpBehaviors & RlpBehaviors.Storage) != 0;
+        int totalLength = GetContentLength(items, rlpBehaviors);
+
+        if (useCompactEncoding)
         {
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Serialization.Rlp/ReceiptArrayStorageDecoder.cs
LukaszRozmej, line 94: You could set `encoder` variable and unify this part after `else` too.

src/Nethermind/Nethermind.Serialization.Rlp/ReceiptArrayStorageDecoder.cs
```diff
@@ -80,9 +80,11 @@ public override void Encode(RlpStream stream, TxReceipt[] items, RlpBehaviors rl
...
 
-        if (compactEncoding && (rlpBehaviors & RlpBehaviors.Storage) != 0)
+        bool useCompactEncoding = compactEncoding && (rlpBehaviors & RlpBehaviors.Storage) != 0;
+        int totalLength = GetContentLength(items, rlpBehaviors);
+
+        if (useCompactEncoding)
         {
-            int totalLength = GetContentLength(items, rlpBehaviors);
             stream.WriteByte(CompactEncoding);
             stream.StartSequence(totalLength - 1);
```
LukaszRozmej, line 89: And do something like

```
var decoder = Decoder;
if (useCompactEncoding)
{
  stream.WriteByte(CompactEncoding);
  totalLength -= 1;
  decoder = CompactDecoder;
}
stream.StartSequence(totalLength);

PR #10166, opened 2026-01-09
fix: GetDefault returns value instead of tuple

endorina:
Fix GetDefault method to return the default value directly instead of wrapping it in a tuple.

The method was incorrectly returning (false, value) tuple, which caused the Value property to contain a tuple object instead of the actual default value when IsSet is false. This fix ensures GetDefault returns the default value as expected by all callers.


src/Nethermind/Nethermind.Config/ConfigSourceHelper.cs
```diff
@@ -101,7 +101,7 @@ public static object ParseValue(Type valueType, string valueString, string categ
         private static bool IsNullString(string valueString) =>
             valueString?.Equals("null", StringComparison.OrdinalIgnoreCase) ?? true;
 
-        public static object GetDefault(Type type) => type.IsValueType ? (false, Activator.CreateInstance(type)) : (false, null);
+        public static object GetDefault(Type type) => type.IsValueType && !type.IsValueTuple() ? Activator.CreateInstance(type) : null;
```
LukaszRozmej: Why `!type.IsValueTuple()`?

src/Nethermind/Nethermind.Config/ConfigSourceHelperTests.cs
```diff
@@ -0,0 +1,42 @@
...
+        Assert.That(ConfigSourceHelper.GetDefault(typeof(bool)), Is.EqualTo(false), "bool should return false");
+        Assert.That(ConfigSourceHelper.GetDefault(typeof(byte)), Is.EqualTo((byte)0), "byte should return 0");
+    }
+
+    [Test]
+    public void GetDefault_returns_null_for_nullable_value_types()
+    {
+        Assert.That(ConfigSourceHelper.GetDefault(typeof(int?)), Is.Null, "int? should return null");
+        Assert.That(ConfigSourceHelper.GetDefault(typeof(bool?)), Is.Null, "bool? should return null");
+    }
```
LukaszRozmej: This should be one test with test cases

src/Nethermind/Nethermind.Config/ConfigSourceHelper.cs
endorina: `ValueTuple` types are excxluded because they need specific values for each field. Using` Activator.CreateInstance` would create a default like `(0, null)` for `(int, string)`, which isn't useful in configuragtion. So `GetDefault` returns `null` for `ValueTuple` types, same as for reference types.

LukaszRozmej: But you cannot assign `null` to value tuple

src/Nethermind/Nethermind.Config/ConfigSourceHelperTests.cs
```diff
@@ -0,0 +1,40 @@
+// SPDX-FileCopyrightText: 2022 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System;
+using System.Collections.Generic;
+using NUnit.Framework;
```
LukaszRozmej, line 7:
```suggestion
using NUnit;
using NUnit.Framework;
```

src/Nethermind/Nethermind.Config/ConfigSourceHelperTests.cs
```diff
@@ -0,0 +1,41 @@
+// SPDX-FileCopyrightText: 2022 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System;
+using System.Collections.Generic;
+using NUnit;
```
LukaszRozmej, line 6: Please move to test project

src/Nethermind/Nethermind.Config/ConfigSourceHelperTests.cs
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    [TestCaseSource(nameof(GetDefaultTestCases))]
+    public void GetDefault_returns_expected_value(Type type, object? expected)
+    {
+        object? result = ConfigSourceHelper.GetDefault(type);
+        Assert.That(result, Is.EqualTo(expected));
+    }
+}
+
```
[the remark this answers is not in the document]
LukaszRozmej: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/NethermindEth/nethermind/pull/10166#discussion_r2690456599)

PR #10167, opened 2026-01-09
perf: Remove redundant ClearPreHash call in Transaction.PoolPolicy.Return

marukai67:
Removes redundant `ClearPreHash()` call in `Transaction.PoolPolicy.Return` method.


MarekM25 (changes requested):
Please elaborate why it is redundant

PR #10168, opened 2026-01-09
Fix disposing of a pooled array

flcl42:
## Changes

- Fix disposing of a pooled array
- Add a minimal delay for the case when there is one node found on local network and in tries to use it to discover other nodes instantly, which happens too often

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


src/Nethermind/Nethermind.Network.Discovery/Discv5/DiscoveryV5App.cs
```diff
@@ -208,64 +208,72 @@ public async IAsyncEnumerable<Node> DiscoverNodes([EnumeratorCancellation] Cance
     {
         Channel<Node> discoveredNodesChannel = Channel.CreateBounded<Node>(1);
 
-        async Task DiscoverAsync(IEnumerable<IEnr> startingNode, ArrayPoolSpan<byte> nodeId)
+        async Task DiscoverAsync(IEnumerable<IEnr> startingNode, ArrayPoolSpan<byte> nodeId, bool disposeNodeId = false)
```
LukaszRozmej: Make it `true` by default and pass `false` where you don't want to dispose - easier to check and reason about exceptions

```suggestion
        async Task DiscoverAsync(IEnumerable<IEnr> startingNode, ArrayPoolSpan<byte> nodeId, bool disposeNodeId = true)
```

src/Nethermind/Nethermind.Network.Discovery/Discv5/DiscoveryV5App.cs
```diff
@@ -294,6 +302,7 @@ static int[] GetDistances(byte[] srcNodeId, in ArrayPoolSpan<byte> destNodeId)
                     }
 
                     await Task.WhenAll(discoverTasks);
+                    await Task.Delay(TimeSpan.FromSeconds(2), token);
```
alexb5dh, line 305: Will throw `Task/OperationCancelledException`  in case of cancellation. Since any error is already handled below, may be worth catching it when token matches and ignoring - so we don't have error in logs if client is shutdown during this call.

flcl42: Good idea

PR #10169, opened 2026-01-09
fix: use Array.Empty instead of [] in AccessListForRpc.ToAccessList

forkfury:
## Fix

Replaces array literal `[]` with `Array.Empty<UInt256>()` in `AccessListForRpc.ToAccessList()` to avoid allocating a new empty array on each iteration when `StorageKeys` is null.

## Changes

- Changed null coalescing operator from `?? []` to `?? Array.Empty<UInt256>()` on line 58
- Matches existing project pattern used in `StorageCellIndexConverter` and other parts of the codebase
- Reduces unnecessary allocations in hot path when processing access lists with null storage keys

### DIFF
--- a/src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/AccessListForRpc.cs
+++ b/src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/AccessListForRpc.cs
@@ -46,7 +46,7 @@ public class AccessListForRpc
 
     public static AccessListForRpc FromAccessList(AccessList? accessList) =>
         accessList is null
-        ? new AccessListForRpc([])
+        ? new AccessListForRpc()
         : new AccessListForRpc(accessList.Select(static item => new Item(item.Address, [.. item.StorageKeys])));
 
     public AccessList ToAccessList()


MarekM25 (approved):

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/AccessListForRpc.cs
LukaszRozmej:
```suggestion
        ? new AccessListForRpc()
```

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/AccessListForRpc.cs
```diff
@@ -55,7 +55,7 @@ public AccessList ToAccessList()
         foreach (Item item in _items)
         {
             builder.AddAddress(item.Address);
-            foreach (UInt256 index in item.StorageKeys ?? [])
+            foreach (UInt256 index in item.StorageKeys ?? Array.Empty<UInt256>())
```
rubo, line 58: This brings no value but bloat

PR #10173, opened 2026-01-10
Optimize engine_getBlobs more

LukaszRozmej:
Based on #10159

## Changes

- Move under one lock
- Move db access out of lock
- Parallelize db access
- use structs instead of classes

- [x] Optimization

#### Requires testing

- [x] Yes

- [x] No


benaadams:
Have a JsonSerializer bypass for blobs in https://github.com/NethermindEth/nethermind/pull/10453 so they go directly to output stream

benaadams:
Superseded by #10453

PR #10177, opened 2026-01-11
Add AGENTS.md

rubo:
This PR aims to cover the 4 major AI agents by adding the standard [AGENTS.md](https://agents.md/), so everyone can use their favorite tool guided by the same rules.

## Changes

- Added AGENTS.md based on CLAUDE.md (respected by Codex and Copilot)
- Replaced CLAUDE.md with AGENTS.md import
- Added GEMINI.md and imported AGENTS.md
- Removed copilot-instructions.md

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [x] Other: AI rules

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -4,9 +4,9 @@ This guide helps to get started with the Nethermind Ethereum execution client re
 
 ## Repo structure
 
-- [src/Nethermind](./src/Nethermind/) directory contains the Nethermind codebase
-- [tools](./tools/) directory contains various servicing tools for testing, monitoring, etc.
-- [scripts](./scripts/) directory contains the build scripts and stuff used by GitHub Actions workflows
+- [src/Nethermind](./src/Nethermind/): The Nethermind codebase
+- [tools](./tools/): Various servicing tools for testing, monitoring, etc.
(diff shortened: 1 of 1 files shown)


benaadams (approved):

rubo:
> @benaadams Wondering whether we can share this with Copilot as well. Having a single source makes it easy to maintain. Otherwise, they quickly become outdated like the one of Copilot.

Turns out Copilot [respects](https://docs.github.com/en/enterprise-cloud@latest/copilot/tutorials/coding-agent/get-the-best-results#adding-custom-instructions-to-your-repository) AGENTS.md, so we can drop copilot-instructions.md bringing whatever we need to AGENTS.md.

.gitignore
```diff
@@ -435,9 +435,10 @@ FodyWeavers.xsd
 ## macOS
 .DS_Store
 
+# AI
+.claude/
+.gemini/
```
LukaszRozmej, line 440: In theory you can have those in subfolders too with additional config, this would prefer:


```suggestion
**/.claude/
**/.gemini/
```

rubo: It already covers all the cases. You don't need wildcards.

PR #10180, opened 2026-01-11
Remove redundant File.Delete after File.Move

eeemmmmmm:
File.Move already removes the source file, making the subsequent File.Delete call unnecessary.


eeemmmmmm:
@rubo please reopen the PR and I will add one more similar change (last)

I can’t add anything to this PR: https://github.com/NethermindEth/nethermind/pull/10128

PR #10184, opened 2026-01-11
fix(tools): add missing receiverOption to reclaim command

rnkrtt:
The `reclaim` command in SendBlobs was broken - `receiverOption` was declared with `Required = true` and used in the handler, but never added to the command. This caused a NullReferenceException when trying to create an Address from null. Other commands in the same file (SetupExecute, SetupSendFileCommand) correctly add this option.


LukaszRozmej:
duplicate of #10136

PR #10194, opened 2026-01-12
Some minor changes on xdc code

wgr523:
## Changes

- In `public bool VerifyCertificate` should check that round == 0 if it is switch block.
- In `XdcHotstuff.cs`, update highest voted round should be done after making sure to send the vote

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Xdc/QuorumCertificateManager.cs
+++ b/src/Nethermind/Nethermind.Xdc/QuorumCertificateManager.cs
@@ -201,7 +201,8 @@ internal class QuorumCertificateManager : IQuorumCertificateManager
         if (certificateTarget.Number == spec.SwitchBlock)
         {
             //Do not check round info on genesis block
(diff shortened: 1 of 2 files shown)


stdevMac:
@wgr523 stale-PR check: hunk 2 was obsoleted by the HotStuff rewrite (#11709/#11711), but the hunk-1 'Round != 0' check in QuorumCertificateManager.VerifyCertificate doesn't appear on master and may be a real fix. @batrr @cicr99 @ak88 — should that check be salvaged (as an issue or a small fresh PR) before we close this? If there's no reply in the next few days, we'll close this as part of the stale-PR cleanup — it can always be reopened or resubmitted.

ak88:
stale PR

PR #10200, opened 2026-01-12
chore: remove unused connectionRounds counter from PeerManager

Fibonacci747:
## Changes

- Remove the connectionRounds counter from PeerManager as it was never incremented or used outside of a debug log. Keeping it in the log always as 0 is misleading and does not provide any useful signal.

- [x] Refactoring


LukaszRozmej:
this won't even compile

Issue #10210, opened 2026-01-13
Fix txpool race condition during chain resets in integration tests


jmadibekov:
When running taiko-client integration tests with Nethermind (l2-nmc), several tests fail with txpool errors:

- `nonce too low`
- `AlreadyKnown`
- `ReplacementNotAllowed`

The same tests pass when running with taiko-geth (l2-geth).

## Root Cause

The issue is a combination of:

1. Different hash caching strategies between geth and Nethermind
2. Race condition between chain head changes and txpool processing

In the integration tests, chain resets happen via Engine API calls:

```
forkchoiceUpdatedV2 -> newPayloadV2 -> getPayloadV2 -> newPayloadV2 -> forkchoiceUpdatedV2
```

This sets the chain head back to an earlier block (e.g., from block 5 back to block 1).

### Why taiko-geth passes

In taiko-geth, the "AlreadyKnown" check is simple - it just looks at transactions currently in the pool:

```go
// core/txpool/legacypool/legacypool.go
func (pool *LegacyPool) add(tx *types.Transaction) (replaced bool, err error) {
    hash := tx.Hash()
    if pool.all.Get(hash) != nil {  // Only checks current pool contents
        return false, txpool.ErrAlreadyKnown
    }
    // ...
}
```

`pool.all` is a simple map of transactions currently in the pool. When a transaction is included in a block, it's removed from `pool.all`. There's no separate "seen hash" cache.

When the chain resets to an earlier block:

1. The old transactions were already processed and removed from `pool.all`
2. When resubmitting, `pool.all.Get(hash)` returns nil
3. Transaction is accepted

### Why Nethermind fails

Nethermind has an additional `HashCache` that persists transaction hashes even after they're processed:

```csharp
// Nethermind.TxPool/HashCache.cs
internal class HashCache
{
    private readonly ClockKeyCache<ValueHash256> _longTermCache;  // Persists across blocks
    private readonly ClockKeyCache<ValueHash256> _currentBlockCache;
    
    public bool Get(Hash256 hash)
    {
        return _currentBlockCache.Get(hash) || _longTermCache.Get(hash);
    }
}
```

The `AlreadyKnownTxFilter` uses this cache:

```csharp
// Nethermind.TxPool/Filters/AlreadyKnownTxFilter.cs
public AcceptTxResult Accept(...)
{
    if (_hashCache.Get(tx.Hash!))  // Checks persistent cache
    {
        return AcceptTxResult.AlreadyKnown;
[...]

PR #10211, opened 2026-01-13
 Add taiko debug RPC for integration test support

jmadibekov:
Fixes https://github.com/NethermindEth/nethermind/issues/10210

## Changes

- Add `taikoDebug_clearTxPoolForReorg` RPC method 

- [x] Other: Taiko specific change to make integration tests to pass

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

#### Notes on testing

Tested manually with taiko-client integration tests. All previously failing tests now pass:

- TestShastaProposalWithMultipleBlocks 
- TestShastaProposalWithOneBlobAndMultipleBlocks 
- TestShastaProposalWithTooMuchBlocks 
- TestShastaProposalsWithForcedInclusion 
- TestShastaProposalsWithInvalidForcedInclusion 

To be specific, with this change, now the whole test suite of integration tests pass in `surge-taiko-mono`; I ran via the following command:

```bash
cd ~/code/surge-taiko-mono/packages/taiko-client
PACAYA_FORK_TAIKO_MONO=~/code/pacaya-fork-taiko-mono \
SHASTA_FORK_TAIKO_MONO=~/code/shasta-fork-taiko-mono \
L2_NODE=l2_nmc \
make test
```

More details in this PR: https://github.com/NethermindEth/surge-taiko-mono/pull/245

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Taiko/Rpc/TaikoEngineRpcModule.cs
+++ b/src/Nethermind/Nethermind.Taiko/Rpc/TaikoEngineRpcModule.cs
@@ -378,13 +378,7 @@ public class TaikoEngineRpcModule(IAsyncHandler<byte[], ExecutionPayload?> getPa
         return ResultWrapper<L1Origin>.Success(l1Origin);
     }
 
-    /// <summary>
-    /// Clears txpool state (hash cache, account cache, pending transactions) after a chain reorg.
-    /// This is specifically designed for Taiko integration tests where the chain is reset to a base block.
-    /// After a reorg, stale txpool caches would reject transaction resubmissions with "already known" or "nonce too low".
-    /// Pending transactions must also be cleared because tests resubmit transactions with the same hash/nonce,
-    /// which would be rejected as "ReplacementNotAllowed" if they remain in the pool.
-    /// </summary>
+    /// <inheritdoc />
     public ResultWrapper<bool> taikoDebug_clearTxPoolForReorg()
     {
         txPool.ResetTxPoolState();
--- a/src/Nethermind/Nethermind.TxPool/ITxPool.cs
+++ b/src/Nethermind/Nethermind.TxPool/ITxPool.cs
@@ -64,12 +64,6 @@ namespace Nethermind.TxPool
         long PendingTransactionsAdded { get; }
 
         /// <summary>
-        /// The last block number that the txpool has processed.
-        /// This may be different from the blockchain head if txpool is still processing head changes.
-        /// </summary>
-        long LastProcessedBlockNumber { get; }
-
-        /// <summary>
         /// Resets txpool state by clearing all caches (hash cache, account cache) 
         /// and removing all pending transactions. Used for integration testing after chain reorgs.
         /// </summary>
--- a/src/Nethermind/Nethermind.TxPool/TxPool.cs
+++ b/src/Nethermind/Nethermind.TxPool/TxPool.cs
@@ -481,8 +481,9 @@ namespace Nethermind.TxPool
         public bool AcceptTxWhenNotSynced { get; set; }
         public bool SupportsBlobs { get; }
         public long PendingTransactionsAdded => Volatile.Read(ref _pendingTransactionsAdded);
-        public long LastProcessedBlockNumber => Volatile.Read(ref _lastBlockNumber);
 
+        /// This is a debug/testing method that clears the entire txpool state.
+        /// Currently only used in the Taiko integration tests after chain reorgs.
         public void ResetTxPoolState()
         {
             _newHeadLock.EnterWriteLock();
@@ -506,6 +507,10 @@ namespace Nethermind.TxPool
                 {
                     RemoveTransaction(tx.Hash);
                 }
+
+                // Update metrics after removal
+                Metrics.TransactionCount = _transactions.Count;
+                Metrics.BlobTransactionCount = _blobTransactions.Count;
 
                 // Reset snapshots
                 _transactionSnapshot = null;
--- a/src/Nethermind/Nethermind.TxPool/NullTxPool.cs
+++ b/src/Nethermind/Nethermind.TxPool/NullTxPool.cs
@@ -113,7 +113,6 @@ namespace Nethermind.TxPool
             remove { }
         }
         public bool AcceptTxWhenNotSynced { get; set; }
-        public long LastProcessedBlockNumber => 0;
         public void ResetTxPoolState() { }
     }
 }


src/Nethermind/Nethermind.TxPool/ITxPool.cs
```diff
@@ -62,5 +62,18 @@ bool TryGetBlobAndProofV1(byte[] blobVersionedHash,
...
+        /// <summary>
+        /// The last block number that the txpool has processed.
+        /// This may be different from the blockchain head if txpool is still processing head changes.
+        /// </summary>
+        long LastProcessedBlockNumber { get; }
+
+        /// <summary>
+        /// Clears all txpool caches (hash cache and account cache).
+        /// This should be called during chain resets/deep reorgs to ensure
+        /// transactions can be resubmitted without AlreadyKnown errors.
```
smartprogrammer93:
```suggestion
```

src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -481,6 +481,41 @@ public void RemovePeer(PublicKey nodeId)
...
+        public void ClearAllCaches()
+        {
+            _newHeadLock.EnterWriteLock();
+            try
+            {
+                // Clear hash cache and account cache
+                _hashCache.ClearAll();
+                _accountCache.Reset();
+
+                // Also clear all pending transactions - they are invalid after a chain reset
```
smartprogrammer93:
```suggestion
                // Also clear all pending transactions
```

src/Nethermind/Nethermind.TxPool/HashCache.cs
```diff
@@ -58,5 +58,16 @@ public void ClearCurrentBlockCache()
         {
             _currentBlockCache.Clear();
         }
+
+        /// <summary>
+        /// Clears both long-term and current block caches.
+        /// This should be called during deep reorgs when the chain goes backwards,
+        /// as transactions from the orphaned chain may now be valid again.
```
smartprogrammer93:
```suggestion
```

smartprogrammer93 (approved):

src/Nethermind/Nethermind.Taiko/Rpc/TaikoEngineRpcModule.cs
```diff
@@ -4,8 +4,10 @@
 using System;
 using System.Buffers;
 using System.Collections.Generic;
+using System.IO;
```
dipkakwani: Unused import here.

src/Nethermind/Nethermind.Taiko/Rpc/TaikoEngineRpcModule.cs
```diff
@@ -377,4 +379,61 @@ public ResultWrapper<L1Origin> taikoAuth_setL1OriginSignature(UInt256 blockId, i
 
         return ResultWrapper<L1Origin>.Success(l1Origin);
     }
+
+    public async Task<ResultWrapper<bool>> taikoAuth_waitForTxPoolSync(long expectedBlockNumber, int timeoutMs = 5000)
+    {
+        using CancellationTokenSource cts = new(TimeSpan.FromMilliseconds(timeoutMs));
```
dipkakwani: Should we add early exit at the start of the function, in case the condition is already met (before creating new tasks/event handler)?

src/Nethermind/Nethermind.Taiko/Rpc/TaikoEngineRpcModule.cs
```diff
@@ -377,4 +379,61 @@ public ResultWrapper<L1Origin> taikoAuth_setL1OriginSignature(UInt256 blockId, i
...
+                txPoolLastBlock = txPool.LastProcessedBlockNumber;
+
+                // Consider synced if txpool has processed the expected block
+                if (txPoolLastBlock >= expectedBlockNumber)
+                {
+                    return ResultWrapper<bool>.Success(true);
+                }
+
+                // Wait a bit and try again
+                await Task.Delay(pollIntervalMs, cts.Token);
```
dipkakwani: This can throw an exception if we reach `timeoutMs`, but that is not being caught in the current code. We need to add `catch` block to handle such scenario and then return false/not-success to the caller in that case.

src/Nethermind/Nethermind.Taiko/Rpc/TaikoEngineRpcModule.cs
```diff
@@ -377,4 +379,61 @@ public ResultWrapper<L1Origin> taikoAuth_setL1OriginSignature(UInt256 blockId, i
 
         return ResultWrapper<L1Origin>.Success(l1Origin);
     }
+
+    public async Task<ResultWrapper<bool>> taikoAuth_waitForTxPoolSync(long expectedBlockNumber, int timeoutMs = 5000)
+    {
+        using CancellationTokenSource cts = new(TimeSpan.FromMilliseconds(timeoutMs));
+        TaskCompletionSource<bool> tcs = new();
+
+        void OnTxPoolHeadChanged(object? sender, Block block)
```
dipkakwani: We are not awaiting on this anywhere? This is potentially dead code in that case, since we are returning irrespective of the result from here?

src/Nethermind/Nethermind.Taiko/Rpc/TaikoEngineRpcModule.cs
```diff
@@ -377,4 +379,61 @@ public ResultWrapper<L1Origin> taikoAuth_setL1OriginSignature(UInt256 blockId, i
...
+        txPool.TxPoolHeadChanged += OnTxPoolHeadChanged;
+        try
+        {
+            // Check if this is a reorg scenario where txpool is ahead of blockchain
+            // In this case, directly clear caches because the async ProcessNewHeadLoop 
+            // may not process in time
+            long txPoolLastBlock = txPool.LastProcessedBlockNumber;
+            long? currentBlockchainHead = blockFinder.Head?.Number;
+
+            if (currentBlockchainHead.HasValue && txPoolLastBlock > currentBlockchainHead.Value)
```
dipkakwani: A potential race condition could occur here? These two values are independently updated and by the time you do comparison, we could see different results?

dipkakwani: I am not too sure if there is a more reliable way to identify a reorg. But if there is none, we might need some locking here?

src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -481,6 +481,41 @@ public void RemovePeer(PublicKey nodeId)
         public bool AcceptTxWhenNotSynced { get; set; }
         public bool SupportsBlobs { get; }
         public long PendingTransactionsAdded => Volatile.Read(ref _pendingTransactionsAdded);
+        public long LastProcessedBlockNumber => Volatile.Read(ref _lastBlockNumber);
+
+        public void ClearAllCaches()
```
dipkakwani: This is not only clearing caches, right? May be helpful to rename or decouple the functions.

dipkakwani (changes requested):

src/Nethermind/Nethermind.Taiko/Rpc/TaikoEngineRpcModule.cs
jmadibekov: Thanks, removed it.

src/Nethermind/Nethermind.TxPool/TxPool.cs
jmadibekov: Good point, renamed it!

src/Nethermind/Nethermind.Taiko/Rpc/TaikoEngineRpcModule.cs
jmadibekov: I initially fixed it locally by using a lock to atomically read these values, but now I simplified my code a lot, this issue is not relevant anymore.

src/Nethermind/Nethermind.TxPool/ITxPool.cs
```diff
@@ -62,5 +62,17 @@ bool TryGetBlobAndProofV1(byte[] blobVersionedHash,
         public bool AcceptTxWhenNotSynced { get; set; }
         bool SupportsBlobs { get; }
         long PendingTransactionsAdded { get; }
+
+        /// <summary>
+        /// The last block number that the txpool has processed.
+        /// This may be different from the blockchain head if txpool is still processing head changes.
+        /// </summary>
+        long LastProcessedBlockNumber { get; }
```
dipkakwani: This is not required anymore, it can be removed.

src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -481,6 +481,41 @@ public void RemovePeer(PublicKey nodeId)
...
+
+                // Clear blob transactions too
+                Transaction[] pendingBlobTxs = _blobTransactions.GetSnapshot();
+                foreach (Transaction tx in pendingBlobTxs)
+                {
+                    RemoveTransaction(tx.Hash);
+                }
+
+                // Reset snapshots
+                _transactionSnapshot = null;
```
dipkakwani, line 516: Do we also need to update the metrics after removal? 
```
Metrics.TransactionCount = _transactions.Count;
Metrics.BlobTransactionCount = _blobTransactions.Count;
```

src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -481,6 +481,41 @@ public void RemovePeer(PublicKey nodeId)
         public bool AcceptTxWhenNotSynced { get; set; }
         public bool SupportsBlobs { get; }
         public long PendingTransactionsAdded => Volatile.Read(ref _pendingTransactionsAdded);
+        public long LastProcessedBlockNumber => Volatile.Read(ref _lastBlockNumber);
+
+        public void ResetTxPoolState()
```
dipkakwani, line 487: Maybe add a comment here to indicate this is for debug purposes; looks like a risky function to me. :sweat_smile: 
If we were using it only for tests, we could have made it internal, similar to some of the other functions in this class.

dipkakwani (approved):

src/Nethermind/Nethermind.TxPool/TxPool.cs
jmadibekov: Added a comment.

PR #10212, opened 2026-01-13
ReleaseSpec: get rid of Default Interface Members to speed up compilation

maximmenshikov:
If Default Interface Members are used for hot path classes, ILCompiler compilation is several orders of magnitude slower.

The slowdown is in dotnet MetadataVirtualMethodAlgorithm.cs - ResolveInterfaceMethodToVirtualMethodOnTypeRecursive.

The solution is to get rid of default interface members and move the functionality to classes derived from the interface.

## Changes

- IReleaseSpec default interface members are moved to classes derived from the interface.

- [x] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

Everything has to be retested.

- [ ] Yes
- [x] No

I think it doesn't.

- [ ] Yes
- [x] No

## Remarks

We need to fix more IReleaseSpec-derived classes probably.

### DIFF
--- a/src/Nethermind/Nethermind.Core/Specs/IReleaseSpec.cs
+++ b/src/Nethermind/Nethermind.Core/Specs/IReleaseSpec.cs
@@ -12,11 +12,13 @@ namespace Nethermind.Core.Specs
     /// </summary>
     public interface IReleaseSpec : IEip1559Spec, IReceiptSpec
     {
+        // Compiling of default interface methods/properties is very slow, therefore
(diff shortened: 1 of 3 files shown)


maximmenshikov:
@benaadams it is expected - it is fully Stateless Executor feature now. We need to expand it to other ReleaseSpec (out of my scope)

PR #10223, opened 2026-01-14
Reduce closure allocations in RetryCache

benaadams:
## Changes

- Don't allocate capture on Update path; and only capture `this` and `result` on Add path

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.TxPool/RetryCache.cs
+++ b/src/Nethermind/Nethermind.TxPool/RetryCache.cs
@@ -43,7 +43,7 @@ public sealed class RetryCache<TMessage, TResourceId> : IAsyncDisposable
         _requestingResources = new(requestingCacheSize);
         _expiringQueueLimit = expiringQueueLimit;
         _maxRetryRequests = maxRetryRequests;
-        // Clousre capture
+        // Closure capture
         _announceUpdate = AnnounceUpdate;
 
         _mainLoopTask = Task.Run(async () =>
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.TxPool/RetryCache.cs
```diff
@@ -119,35 +122,41 @@ public AnnounceResult Announced(in TResourceId resourceId, IMessageHandler<TMess
...
             AnnounceResult result = AnnounceResult.Delayed;
-
-            _retryRequests.AddOrUpdate(resourceId, (resourceId) =>
+            _retryRequests.AddOrUpdate(resourceId, (resourceId, retryHandler) =>
             {
-                if (_logger.IsTrace) _logger.Trace($"Announced {resourceId} by {retryHandler}: NEW");
-
-                _expiringQueue.Enqueue((resourceId, DateTimeOffset.UtcNow.AddMilliseconds(_timeoutMs)));
-                Interlocked.Increment(ref _expiringQueueCounter);
+                return AnnounceAdd(resourceId, retryHandler, out result);
```
flcl42, line 127: May be `ref`

src/Nethermind/Nethermind.TxPool/RetryCache.cs
```diff
@@ -119,35 +122,41 @@ public AnnounceResult Announced(in TResourceId resourceId, IMessageHandler<TMess
...
-                if (requests.Count < _maxRetryRequests)
-                {
-                    requests.Add(retryHandler);
-                }
+        return AnnounceResult.RequestRequired;
+    }
 
-                return requests;
-            });
+    private ConcurrentHashSet<IMessageHandler<TMessage>> AnnounceUpdate(TResourceId resourceId, ConcurrentHashSet<IMessageHandler<TMessage>> requests, IMessageHandler<TMessage> retryHandler)
```
flcl42: Could you swap order, so add method goes first

flcl42 (approved):

src/Nethermind/Nethermind.TxPool/RetryCache.cs
benaadams: Don't think changes anything; really need ConcurrentDict to return value as to whether it added or updated

Will raise api request in runtime

flcl42: Initial value is not default so ref is more in place, just semantics ofc, no byte code diff

PR #10224, opened 2026-01-14
Validate state test tx as part of a block

flcl42:
## Changes

- Validate state test tx as part of a block

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain/Tracing/GethStyle/Custom/Native/GethLikeNativeTracerFactory.cs
+++ b/src/Nethermind/Nethermind.Blockchain/Tracing/GethStyle/Custom/Native/GethLikeNativeTracerFactory.cs
@@ -31,7 +31,7 @@ public static class GethLikeNativeTracerFactory
         RegisterTracer(NativeCallTracer.CallTracer, static (options, _, transaction, _) => new NativeCallTracer(transaction, options));
     }
 
-    private static void RegisterTracer(string tracerName, GethLikeNativeTracerFactoryDelegate tracerDelegate)
+    public static void RegisterTracer(string tracerName, GethLikeNativeTracerFactoryDelegate tracerDelegate)
     {
         _tracers.Add(tracerName, tracerDelegate);
     }
--- a/src/Nethermind/Nethermind.Evm/GasPolicy/EthereumGasPolicy.cs
+++ b/src/Nethermind/Nethermind.Evm/GasPolicy/EthereumGasPolicy.cs
@@ -170,6 +170,12 @@ public struct EthereumGasPolicy : IGasPolicy<EthereumGasPolicy>
         => Consume(ref gas, baseCost + dataCost);
 
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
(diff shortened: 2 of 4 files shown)


src/Nethermind/Ethereum.Test.Base/GeneralTestBase.cs
```diff
@@ -121,41 +126,34 @@ protected EthereumTestResult RunTest(GeneralStateTest test, ITxTracer txTracer)
...
-            ValidationResult txIsValid = new TxValidator(test.ChainId).IsWellFormed(test.Transaction, spec);
+            Transaction[] transactions = [test.Transaction];
+            Withdrawal[]? withdrawals = spec.WithdrawalsEnabled ? [] : null;
+            Block block = new(header, new BlockBody(transactions, [], withdrawals));
+
+            header.TxRoot = TxTrie.CalculateRoot(block.Transactions);
+            header.ReceiptsRoot ??= test.PostReceiptsRoot;
+            header.WithdrawalsRoot = test.CurrentWithdrawalsRoot ?? (spec.WithdrawalsEnabled ? PatriciaTree.EmptyTreeHash : null);
+            header.RequestsHash = test.RequestsHash ?? (spec.RequestsEnabled ? ExecutionRequestExtensions.EmptyRequestsHash : null);
+            header.Hash = header.CalculateHash();
```
LukaszRozmej: This is a bit confusing as the code to create the header is here: https://github.com/NethermindEth/nethermind/pull/10224/changes#diff-4e5040b59ac024fe8d63943a1bab2ac9e73576492300875a681eb3e2fb822e64R109-R128

Should this code be unified?

src/Nethermind/Ethereum.Test.Base/GeneralStateTestEnvJson.cs
```diff
@@ -20,7 +20,5 @@ public class GeneralStateTestEnvJson
         public Hash256? CurrentBeaconRoot { get; set; }
         public Hash256? CurrentWithdrawalsRoot { get; set; }
         public ulong? CurrentExcessBlobGas { get; set; }
-        public UInt256? ParentBlobGasUsed { get; set; }
-        public UInt256? ParentExcessBlobGas { get; set; }
```
LukaszRozmej, line 24: Will we now just ignore those test params?

src/Nethermind/Ethereum.Test.Base/GeneralTestBase.cs
flcl42: Good idea!

src/Nethermind/Ethereum.Test.Base/GeneralStateTestEnvJson.cs
flcl42: I found no such fields in any test, there are also not really used in our code. the only usage is replaced by currentExcessBlobGas which is really in the test jsons

PR #10227, opened 2026-01-14
Allow some delays to be configurable

benaadams:
## Changes

- Allow pruning delay and gc delay to be configurable

- [x] Other: config

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Db/IPruningConfig.cs
+++ b/src/Nethermind/Nethermind.Db/IPruningConfig.cs
@@ -104,6 +104,6 @@ public interface IPruningConfig : IConfig
     [ConfigItem(Description = "[TECHNICAL] Simulate long finalization by not moving finalized block pointer until after this depth.", DefaultValue = "0", HiddenFromDocs = true)]
     int SimulateLongFinalizationDepth { get; set; }
 
-    [ConfigItem(Description = "If in memory pruning is scheduled how long after newPayload before triggering. Very short will mean it will clash with forkchoice, too long and it will overlap with GC.", DefaultValue = "75", HiddenFromDocs = true)]
+    [ConfigItem(Description = "If in-memory pruning is scheduled, the duration between `newPayload` and the GC trigger. If too short, it may clash with fork choice; if too long, it may overlap with GC.", DefaultValue = "75", HiddenFromDocs = true)]
     int PruneDelayMilliseconds { get; set; }
 }


src/Nethermind/Nethermind.Db/IPruningConfig.cs
```diff
@@ -103,4 +103,7 @@ The max number of parallel tasks that can be used by full pruning.
 
     [ConfigItem(Description = "[TECHNICAL] Simulate long finalization by not moving finalized block pointer until after this depth.", DefaultValue = "0", HiddenFromDocs = true)]
     int SimulateLongFinalizationDepth { get; set; }
+
+    [ConfigItem(Description = "If in memory pruning is scheduled how long after newPayload before triggering. Very short will mean it will clash with forkchoice, too long and it will overlap with GC.", DefaultValue = "75", HiddenFromDocs = true)]
```
LukaszRozmej: Description should be different aand alligned with others.

src/Nethermind/Nethermind.Merge.Plugin/IMergeConfig.cs
```diff
@@ -72,4 +72,7 @@ The number of requests to the garbage collector (GC) to release the process memo
 
     [ConfigItem(Description = "[TECHNICAL] Simulate block production for every possible slot. Just for stress-testing purposes.", DefaultValue = "false", HiddenFromDocs = true)]
     bool SimulateBlockProduction { get; set; }
+
+    [ConfigItem(Description = "Delay in milliseconds after newPayload before triggering GC, if not set defaults to 1/8th of SecondsPerSlot", DefaultValue = null, HiddenFromDocs = true)]
+    int? PostBlockGcDelayMs { get; set; }
```
LukaszRozmej: This one is lot better

src/Nethermind/Nethermind.Db/IPruningConfig.cs
rubo: If I got it right, then maybe this is better:
```suggestion
    [ConfigItem(Description = "If in-memory pruning is scheduled, the duration between `newPayload` and the GC trigger. If too short, it may clash with fork choice; if too long, it may overlap with GC.", DefaultValue = "75", HiddenFromDocs = true)]
```

src/Nethermind/Nethermind.Merge.Plugin/IMergeConfig.cs
```diff
@@ -72,4 +72,7 @@ The number of requests to the garbage collector (GC) to release the process memo
 
     [ConfigItem(Description = "[TECHNICAL] Simulate block production for every possible slot. Just for stress-testing purposes.", DefaultValue = "false", HiddenFromDocs = true)]
     bool SimulateBlockProduction { get; set; }
+
+    [ConfigItem(Description = "Delay in milliseconds after newPayload before triggering GC, if not set defaults to 1/8th of SecondsPerSlot", DefaultValue = null, HiddenFromDocs = true)]
```
rubo:
```suggestion
    [ConfigItem(Description = "Delay, in milliseconds, between `newPayload` and GC trigger. If not set, defaults to 1/8th of `Blocks.SecondsPerSlot`.", DefaultValue = null, HiddenFromDocs = true)]
```

rubo (approved):

PR #10241, opened 2026-01-14
Optimize storage key handling

benaadams:
## Changes

- Returns early in `ComputeKeyWithLookup` when key is found (currently it recomputes the key anyway) and uses faster copy than `.CopyTo`
- Replaces byte[] with ValueHash256 for storage key lookup and computation in StorageTree for improved type safety and performance.
- Updates related method signatures and usages to use ValueHash256.
- Adds [SkipLocalsInit] attribute to several performance-critical methods in trie and pruning code.

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.State/StorageTree.cs
+++ b/src/Nethermind/Nethermind.State/StorageTree.cs
@@ -2,12 +2,8 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 
 using System;
-using System.Collections.Frozen;
-using System.Collections.Generic;
 using System.Runtime.CompilerServices;
-using System.Runtime.InteropServices;
 using Nethermind.Core;
-using Nethermind.Core.Collections;
 using Nethermind.Core.Crypto;
 using Nethermind.Core.Extensions;
 using Nethermind.Evm.State;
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.State/StorageTree.cs
```diff
@@ -52,27 +52,25 @@ public StorageTree(IScopedTrieStore? trieStore, Hash256 rootHash, ILogManager? l
...
+        private static void ComputeKey(in UInt256 index, out ValueHash256 key)
         {
-            index.ToBigEndian(key);
-
-            // We can't direct ComputeTo the key as its also the input, so need a separate variable
-            KeccakCache.ComputeTo(key, out ValueHash256 keyHash);
-            // Which we can then directly assign to fast update the key
-            Unsafe.As<byte, ValueHash256>(ref MemoryMarshal.GetReference(key)) = keyHash;
+            Span<byte> buffer = stackalloc byte[32];
+            index.ToBigEndian(buffer);
```
LukaszRozmej: Do we need the buffer? Can't we reuse key memory?

```suggestion
            Span<byte> buffer = key.BytesAsSpan;
            index.ToBigEndian(buffer);
```

src/Nethermind/Nethermind.State/StorageTree.cs
```diff
@@ -158,7 +155,7 @@ public void Set(in UInt256 index, byte[] value)
         {
             if (index < LookupSize)
             {
-                SetInternal(Lookup[index], value);
+                SetInternal(Lookup[index].Bytes, value);
```
LukaszRozmej: Probably slightly better if `SetInternal` takes `in ValueHash256` and gets bytes inside if needed

src/Nethermind/Nethermind.State/StorageTree.cs
```diff
@@ -92,25 +90,24 @@ public static BulkSetEntry CreateBulkSetEntry(ValueHash256 key, byte[]? value)
...
             return GetWithKeyGenerate(in index, storageRoot);
 
             [SkipLocalsInit]
             byte[] GetWithKeyGenerate(in UInt256 index, Hash256 storageRoot)
             {
-                Span<byte> key = stackalloc byte[32];
-                ComputeKey(index, key);
-                return GetArray(key, storageRoot);
+                ComputeKey(index, out ValueHash256 key);
+                return GetArray(key.Bytes, storageRoot);
```
LukaszRozmej: Not sure if the same is for `GetArray`

LukaszRozmej (approved):

src/Nethermind/Nethermind.State/StorageTree.cs
benaadams: Using buffer as reference for both in (via Span) and out would cause issues in .ComputeTo

Will check if it can easily be made safe

src/Nethermind/Nethermind.State/StorageTree.cs
benaadams: Changed

src/Nethermind/Nethermind.State/StorageTree.cs
benaadams: Changed

src/Nethermind/Nethermind.State/StorageTree.cs
benaadams: Isn't easy to change but using a second ValueHash256 is better than a stackalloc; so changed to that

Also using array rather than FrozenDict as can just lookup by index and get the ref

PR #10248, opened 2026-01-15
P2P: Deduplicate background sync sender logic

viktorking7:
- Deduplicated the “await fulfill → send response” path in `BackgroundTaskSchedulerWrapper` by centralizing it in a single helper method.


LukaszRozmej:
pointless

PR #10263, opened 2026-01-19
Fix/shared blockcache not used

asdacap:
- Backporting some fix from #9854 
- Fix shared blockcache not used due to wrong ordering of setting.
- Change shared blockcache to hyperclockcache which is the recommended default starting from rocksdb 8.5... or something.
- Make metric for block cache easier by explicitly specifying the shared blok cache as a separate pool.
- Since we've been running with the default blockcache size (32 mb when not specified), reduced memory hint to reflect.
- Add a normalizer so that `optimize_filters_for_hits` can be overridden which is important for flat.

- [X] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Db.Rocks/HyperClockCacheWrapper.cs
+++ b/src/Nethermind/Nethermind.Db.Rocks/HyperClockCacheWrapper.cs
@@ -2,31 +2,29 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 
 using System;
-using System.Threading;
-using Nethermind.Core.Extensions;
+using Microsoft.Win32.SafeHandles;
 using RocksDbSharp;
 
 namespace Nethermind.Db.Rocks;
 
-public class HyperClockCacheWrapper : IDisposable
+public class HyperClockCacheWrapper : SafeHandleZeroOrMinusOneIsInvalid
 {
-    private readonly IntPtr _cacheHandle;
-
-    public HyperClockCacheWrapper(ulong capacity = 32_000_000)
+    public HyperClockCacheWrapper(ulong capacity = 32_000_000) : base(ownsHandle: true)
     {
-        _cacheHandle = RocksDbSharp.Native.Instance.rocksdb_cache_create_hyper_clock(new UIntPtr(capacity), 0);
+        SetHandle(RocksDbSharp.Native.Instance.rocksdb_cache_create_hyper_clock(new UIntPtr(capacity), 0));
     }
 
-    private bool _isDisposed;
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Db.Rocks/HyperClockCacheWrapper.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        _cacheHandle = RocksDbSharp.Native.Instance.rocksdb_cache_create_hyper_clock(new UIntPtr(capacity), 0);
+    }
+
+    private bool _isDisposed;
+    public IntPtr? Handle => _cacheHandle;
+    public void Dispose()
+    {
+        if (Interlocked.CompareExchange(ref _isDisposed, true, false)) return;
+        RocksDbSharp.Native.Instance.rocksdb_cache_destroy(_cacheHandle);
```
LukaszRozmej:
```suggestion
        RocksDbSharp.Native.Instance.rocksdb_cache_destroy(_cacheHandle);
        _cacheHandle = IntPtr.Zero;
```

src/Nethermind/Nethermind.Db.Rocks/HyperClockCacheWrapper.cs
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    public long GetUsage()
+    {
+        return (long)Native.Instance.rocksdb_cache_get_usage(_cacheHandle);
```
LukaszRozmej: Add disposed check?

src/Nethermind/Nethermind.Db.Rocks/HyperClockCacheWrapper.cs
```diff
@@ -0,0 +1,32 @@
...
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System;
+using System.Threading;
+using Nethermind.Core.Extensions;
+using RocksDbSharp;
+
+namespace Nethermind.Db.Rocks;
+
+public class HyperClockCacheWrapper : IDisposable
```
LukaszRozmej: You probably should just derive from `SafeHandleZeroOrMinusOneIsInvalid` or `SafeHandle`: https://blog.benoitblanchon.fr/safehandle/

asdacap:
@copilot make use of MSFT's money by doing it.

PR #10267, opened 2026-01-19
Warmup threads should not update tx.SpentGas

damian-orzechowski:
## Changes

State prewarmer creates separate processing environments for multithreaded transaction processing (pooled `TransactionProcessor` instances with separate VMs and its state) ensuring separate execution. The only shared object seems to be `Block` itself and it's transactions, therefore whenever warmup threads make updates to transaction object, these may be reflected in main processing thread.
Issue spotted for this particular line on Arbitrum mainnet:
https://github.com/NethermindEth/nethermind/blob/df26ddd16f15affc3a6e919ef1356bda74699316/src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs#L225

Log excerpt:
Thread 44 - main processing thread, thread 11 - one of warmup threads overwriting gas spent to incorrect value of `291151`
```
2026-01-18 21:20:54.5613|DEBUG|lambda_method305|44|ExecuteTransaction multi gas {"unknown":0,"computation":1678842,"historyGrowth":5120,"storageAccess":84200,"storageGrowth":0,"l1Calldata":139957,"l2Calldata":80,"wasmComputation":0,"refund":0,"total":1908199} 
2026-01-18 21:20:54.5613|DEBUG|lambda_method305|11|ExecuteTransaction multi gas {"unknown":0,"computation":1679242,"historyGrowth":5120,"storageAccess":72600,"storageGrowth":0,"l1Calldata":139957,"l2Calldata":80,"wasmComputation":0,"refund":0,"total":1896999} 
2026-01-18 21:20:54.5613|DEBUG|lambda_method302|44|ExecuteEvmCall 0xafb6ead8172ddeba04eeb67b20d25ce9bda67e95b1a77b61d0125b2410bb9420 -> 285015 - 0 - 302351 
2026-01-18 21:20:54.5613|DEBUG|lambda_method302|11|ExecuteEvmCall 0xafb6ead8172ddeba04eeb67b20d25ce9bda67e95b1a77b61d0125b2410bb9420 -> 296215 - 0 - 291151 
2026-01-18 21:20:54.5613|DEBUG|lambda_method302|44|Tx 0xafb6ead8172ddeba04eeb67b20d25ce9bda67e95b1a77b61d0125b2410bb9420 - spent gas: 302351 - tx gas 587366 
2026-01-18 21:20:54.5613|DEBUG|lambda_method302|11|Tx 0xafb6ead8172ddeba04eeb67b20d25ce9bda67e95b1a77b61d0125b2410bb9420 - spent gas: 291151 - tx gas 587366 
```

The assumption here is that warmup threads execution can have different result (and gas consumption) as they are executed on different state (out of order) then main thread does. However, as this has not been spotted earlier, might be wrong.

Proposed change
- Add explicit `ExecutionOptions` and only update `transaction.SpentGas` from non warmup threads

Alternative options:
- Add check similar to `PrewarmerTxAdapter` -> `if (worldState.ScopeProvider is IPreBlockCaches { IsWarmWorldState: true })`, but this requires adding dependent library
- Copy transaction objects in `BlockCachePreWarmer` - general separation, but at a cost of data copy

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/.github/workflows/ci-taiko.yml
+++ b/.github/workflows/ci-taiko.yml
@@ -1,0 +1,173 @@
+name: "Nethermind/Ethereum Taiko Client CI Tests"
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+    paths:
+      - "src/Nethermind/Nethermind.Taiko/**"
+      - ".github/workflows/ci-taiko.yml"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  integration_tests:
+    if: >-
+      ${{ github.event.pull_request.draft == false
+      && !startsWith(github.head_ref, 'release-please') }}
+    name: Integration tests
+    runs-on: [ubuntu-latest]
+    timeout-minutes: 30
+    env:
+      OLD_FORK_TAIKO_MONO_DIR: old-fork-taiko-mono
+      TAIKO_MONO_MAIN_DIR: taiko-mono-main
+      PACAYA_FORK_TAIKO_MONO_DIR: pacaya-fork-taiko-mono
+      SHASTA_FORK_TAIKO_MONO_DIR: shasta-fork-taiko-mono
+
+    strategy:
+      matrix:
+        execution_node: [l2_nmc]
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: actions/checkout@v6
+        with:
+          repository: NethermindEth/surge-taiko-mono
+          path: ${{ env.TAIKO_MONO_MAIN_DIR }}
+          ref: surge-shasta
+
+      - name: Install Foundry
+        uses: foundry-rs/foundry-toolchain@v1
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: ${{ env.TAIKO_MONO_MAIN_DIR }}/go.mod
+          cache: true
+
+      - name: Set up Git to use HTTPS
+        shell: bash
+        run: |
+          git config --global url."https://github.com/".insteadOf "git@github.com:"
+
+      - name: Install pnpm
+        uses: pnpm/action-setup@v4
+        with:
+          version: 9
+          run_install: false
+
+      - name: Install Node.js
+        uses: actions/setup-node@v5
+        with:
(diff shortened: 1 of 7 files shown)


src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -222,7 +227,10 @@ protected virtual TransactionResult Execute(Transaction tx, ITxTracer tracer, Ex
                 ExecuteEvmCall<OnFlag>(tx, header, spec, tracer, opts, delegationRefunds, intrinsicGas, accessTracker, gasAvailable, env, out substate, out spentGas);
 
             PayFees(tx, header, spec, tracer, in substate, spentGas.SpentGas, premiumPerGas, blobBaseFee, statusCode);
-            tx.SpentGas = spentGas.SpentGas;
+
+            //only main thread updates transaction
+            if (!opts.HasFlag(ExecutionOptions.Warmup))
```
LukaszRozmej, line 232: maybe just reuse SkipValidation would be enough?

damian-orzechowski: not sure, might more difficult and affect `Trace` and `CallAndRestore`

asdacap:
Another idea: Clone the Tx in prewarmer.

Demuirgos (approved):

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
damian-orzechowski: checked and 2 places stand out:
- `SimulateTransactionProcessorAdapter` uses `SpentGas` after execution, so can get different results
- Taiko / Surge transaction processing could be affected: `TaikoEngineRpcModule`

I think adding a new flag is a safer option

LukaszRozmej (approved):

PR #10274, opened 2026-01-20
Xdc: Penalty Handler

Demuirgos:
Fixes Closes Resolves #

## Changes

- _List the changes_

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Remarks

### DIFF
--- a/src/Nethermind/Nethermind.Xdc.Test/ModuleTests/PenaltyTests.cs
+++ b/src/Nethermind/Nethermind.Xdc.Test/ModuleTests/PenaltyTests.cs
@@ -200,6 +200,7 @@ internal class PenaltyTests
 
         var blockHeaders = new XdcBlockHeader[chainSize];
         var hashToHeader = new Dictionary<Hash256, XdcBlockHeader>();
+        var hashToBlock = new Dictionary<Hash256, Block>();
         var blocks = new Block[chainSize];
 
         for (int i = 0; i < chainSize; i++)
@@ -218,10 +219,13 @@ internal class PenaltyTests
             Hash256 hash = blockHeaders[i].Hash ?? blockHeaders[i].CalculateHash().ToHash256();
             hashToHeader[hash] = blockHeaders[i];
             blocks[i] = new Block(blockHeaders[i]);
+            hashToBlock[hash] = blocks[i];
         }
 
         blockTree.FindHeader(Arg.Any<Hash256>(), Arg.Any<long>())
             .Returns(ci => hashToHeader[ci.ArgAt<Hash256>(0)]);
+        blockTree.FindBlock(Arg.Any<Hash256>(), Arg.Any<long>())
+            .Returns(ci => hashToBlock.TryGetValue(ci.ArgAt<Hash256>(0), out Block? block) ? block : null);
         blockTree.Head.Returns(blocks.Last());
 
         IXdcReleaseSpec xdcSpec = Substitute.For<IXdcReleaseSpec>();
--- a/src/Nethermind/Nethermind.Xdc/SigningTxCache.cs
+++ b/src/Nethermind/Nethermind.Xdc/SigningTxCache.cs
@@ -34,9 +34,7 @@ public class SigningTxCache : ISigningTxCache
 
         Block? block = _blockTree.FindBlock(blockHash, blockNumber);
         if (block is null)
-        {
-            return [];
-        }
+            throw new InvalidOperationException($"Expected block {blockHash} at number {blockNumber} to exist in block tree.");
 
         return CacheSigningTransactions(blockHash, block, spec);
     }


src/Nethermind/Nethermind.Xdc/IMasternodesCalculator.cs
ak88, line 1: Why do we need a new class for this?

src/Nethermind/Nethermind.Xdc/SigningTxCache.cs
ak88, line 1: Needs to be aligned with what we already have

src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
```diff
@@ -84,37 +83,6 @@ public void StoreSnapshot(Snapshot snapshot)
         _snapshotCache.Set(snapshot.HeaderHash, snapshot);
     }
 
-    public (Address[] Masternodes, Address[] PenalizedNodes) CalculateNextEpochMasternodes(long blockNumber, Hash256 parentHash, IXdcReleaseSpec spec)
```
ak88, line 87: I dont get why this is moved

src/Nethermind/Nethermind.Xdc/SigningTxCache.cs
```diff
@@ -0,0 +1,36 @@
...
+    {
+        if (_signingTxsCache.TryGet(blockHash, out Transaction[] signingTxs))
+        {
+            return signingTxs;
+        }
+
+        Block? block = blockTree.FindBlock(blockHash, blockNumber);
+        if (block is null)
+        {
+            return [];
```
ak88: in which scenario would we expect the requested block to not exists?

src/Nethermind/Nethermind.Xdc/SigningTxCache.cs
cicr99: Initially we had discussed putting the cache inside `XdcConsensusContext` and passing the context to both `PenaltyHandler` and `XdcRewardCalculator`.

At the end, this approach felt cleaner. Both classes had identical logic around block access and tx filtering when interacting with the cache, so centralizing it removed duplication. It also keeps responsibilities separate, since the context currently only holds consensus state.

Did you have something different in mind?

src/Nethermind/Nethermind.Xdc/SigningTxCache.cs
cicr99: I don't expect it to return `null` tbh, but thought to leave it as a guard. It also mirrors the original code in xdc side, they do `if block != nil` and only populate cache in that branch; otherwise `signingTxs` stays empty (check [here](https://github.com/XinFinOrg/XDPoSChain/blob/398d9b693d15610a765f3e13c7ff912c84eb533a/eth/hooks/engine_v2_hooks.go#L233))

src/Nethermind/Nethermind.Xdc/SnapshotManager.cs
Demuirgos: Circular dependency because of PenaltyHandler

cicr99: the cycle goes:
`EpochSwitchManager` 
-> `SnapshotManager` 
-> `PenaltyHandler` (because of this method) 
-> `EpochSwitchManager`

batrr (approved):

src/Nethermind/Nethermind.Xdc/PenaltyHandler.cs
```diff
@@ -1,16 +1,190 @@
...
+        if (limit == 0) return currentEpochSwitchInfo.Penalties;
+
+        var epochNumber = (ulong)spec.SwitchEpoch + currentEpochSwitchInfo.EpochSwitchBlockInfo.Round / (ulong)spec.EpochLength;
+        if (epochNumber < limit) return [];
+
+        BlockRoundInfo results = epochSwitchManager.GetBlockByEpochNumber(epochNumber - limit);
+        if (results is null) return [];
+
+        var header = (XdcBlockHeader)tree.FindHeader(results.Hash, results.BlockNumber);
+        if (header is null || header.PenaltiesAddress is null) return [];
```
alexb5dh:
```suggestion
        if (header?.PenaltiesAddress is null) return [];
```

src/Nethermind/Nethermind.Xdc/PenaltyHandler.cs
```diff
@@ -1,16 +1,190 @@
...
+        var header = tree.FindHeader(parentHash, parentNumber) as XdcBlockHeader;
+        IXdcReleaseSpec currentSpec = specProvider.GetXdcSpec(header!);
+        Address[] preMasternodes = epochSwitchManager.GetEpochSwitchInfo(parentHash)!.Masternodes;
+        var penalties = new HashSet<Address>();
+
+        int minMinerBlockPerEpoch = XdcConstants.MinimumMinerBlockPerEpoch;
+        if (currentSpec.TipUpgradePenalty <= number)
+        {
+            minMinerBlockPerEpoch = currentSpec.MinimumMinerBlockPerEpoch;
+        }
```
alexb5dh:
```suggestion
        int minMinerBlockPerEpoch = currentSpec.TipUpgradePenalty <= number
            ? currentSpec.MinimumMinerBlockPerEpoch
            : XdcConstants.MinimumMinerBlockPerEpoch;
```

src/Nethermind/Nethermind.Xdc/PenaltyHandler.cs
```diff
@@ -1,16 +1,190 @@
...
+
+            Address miner = parentHeader.Beneficiary ?? _ethereumEcdsa.RecoverAddress(new Signature(parentHeader.Validator.AsSpan(0, 64), parentHeader.Validator[64]), Keccak.Compute(_xdcHeaderDecoder.Encode(parentHeader, RlpBehaviors.ForSealing).Bytes));
+            minerStatistics[miner!] = minerStatistics.TryGetValue(miner, out int count) ? count + 1 : 1;
+
+            parentNumber--;
+            currentHash = parentHeader.ParentHash;
+            listBlockHash.Add(currentHash);
+        }
+
+        var header = tree.FindHeader(parentHash, parentNumber) as XdcBlockHeader;
```
alexb5dh: Direct cast is used in other places. Looks safer, given that it's always assumed as not-null.

src/Nethermind/Nethermind.Xdc/PenaltyHandler.cs
```diff
@@ -1,16 +1,190 @@
...
+            var isEpochSwitch = epochSwitchManager.IsEpochSwitchAtBlock(parentHeader);
+            if (isEpochSwitch)
+                break;
+
+            Address miner = parentHeader.Beneficiary ?? _ethereumEcdsa.RecoverAddress(new Signature(parentHeader.Validator.AsSpan(0, 64), parentHeader.Validator[64]), Keccak.Compute(_xdcHeaderDecoder.Encode(parentHeader, RlpBehaviors.ForSealing).Bytes));
+            minerStatistics[miner!] = minerStatistics.TryGetValue(miner, out int count) ? count + 1 : 1;
+
+            parentNumber--;
+            currentHash = parentHeader.ParentHash;
+            listBlockHash.Add(currentHash);
```
alexb5dh, line 66: Can `parentHeader.ParentHash` be null if `parentNumber == -1`?
Also, `-1` will probably cause an exception below.

src/Nethermind/Nethermind.Xdc/PenaltyHandler.cs
```diff
@@ -1,16 +1,190 @@
...
+        }
+
+        foreach (var (miner, total) in minerStatistics)
+        {
+            if (total < minMinerBlockPerEpoch)
+                penalties.Add(miner);
+        }
+        penalties.UnionWith(
+            preMasternodes.Where(address => !minerStatistics.ContainsKey(address))
+        );
```
alexb5dh, line 85:
```suggestion
        penalties.UnionWith(
            preMasternodes.Except(minerStatistics.Keys)
        );
```

src/Nethermind/Nethermind.Xdc/PenaltyHandler.cs
```diff
@@ -1,16 +1,190 @@
...
+
+        bool isTipUpgradePenalty = currentSpec.TipUpgradePenalty <= number;
+        if (!isTipUpgradePenalty)
+        {
+            long comebackHeight = (currentSpec.LimitPenaltyEpochV2 + 1) * currentSpec.EpochLength + currentSpec.SwitchBlock;
+            var penComebacks = new HashSet<Address>();
```
alexb5dh: Definition and assignment can be moved below, just redundant allocation ATM.

src/Nethermind/Nethermind.Xdc/PenaltyHandler.cs
alexb5dh: `parentNumber` is decremented in a loop above, while `parentHash` stays the same, so they will likely correspond to different blocks here.

From debugging `TestHookPenaltyV2Comeback` it looks like actual header is always fetched from the cache and is one with `hash == parentHash`, but not with `number == parentNumber` (may actually be a bug in `BlockTree`/`HeaderStore`). Some other tests just mock the header fetching. In general, it will probably just return `null` in case of cache miss.

src/Nethermind/Nethermind.Xdc/XdcTransactionExtensions.cs
```diff
@@ -25,6 +25,17 @@ public static bool RequiresSpecialHandling(this Transaction currentTx, IXdcRelea
...
 
+    public static bool IsSigningTransaction(this Transaction currentTx, IXdcReleaseSpec spec)
+    {
+        var targetIsSignContract = currentTx.To is not null && (currentTx.To == spec.BlockSignerContract);
+        if (!targetIsSignContract) return false;
+
+        if (currentTx.Data.Length != 68) return false;
+
+
+        return currentTx.Data.Span.Slice(0, 4).SequenceEqual(XdcConstants.SignMethod);
```
alexb5dh:
```suggestion
        if (currentTx.Data.Length != 68) return false;

        return currentTx.Data.Span.Slice(0, 4).SequenceEqual(XdcConstants.SignMethod);
```

src/Nethermind/Nethermind.Xdc/XdcTransactionExtensions.cs
```diff
@@ -25,6 +25,17 @@ public static bool RequiresSpecialHandling(this Transaction currentTx, IXdcRelea
     public static bool IsLendingFinalizedTradeTransaction(this Transaction currentTx, IXdcReleaseSpec spec) => currentTx.To is not null && currentTx.To == spec.XDCXLendingFinalizedTradeAddressBinary && spec.IsTIPXDCXMiner;
     public static bool IsTradingStateTransaction(this Transaction currentTx, IXdcReleaseSpec spec) => currentTx.To is not null && currentTx.To == spec.TradingStateAddressBinary && spec.IsTIPXDCXMiner;
 
+    public static bool IsSigningTransaction(this Transaction currentTx, IXdcReleaseSpec spec)
+    {
+        var targetIsSignContract = currentTx.To is not null && (currentTx.To == spec.BlockSignerContract);
```
alexb5dh, line 30: Can `BlockSignerContract` be `null`?

src/Nethermind/Nethermind.Xdc/XdcTransactionExtensions.cs
```diff
@@ -25,6 +25,17 @@ public static bool RequiresSpecialHandling(this Transaction currentTx, IXdcRelea
...
+        if (!targetIsSignContract) return false;
+
+        if (currentTx.Data.Length != 68) return false;
```
alexb5dh: Maybe move `68` to `XdcConstants` with some descriptive naming, same as `SignMethod`? Also used in `XdcRewardCalculator.ExtractBlockHashFromSigningTxData`.

src/Nethermind/Nethermind.Xdc.Test/ModuleTests/PenaltyTests.cs
```diff
@@ -0,0 +1,279 @@
...
+
+            blockHeaders[i].Beneficiary = masternodes[i % (masternodes.Length - 1)].Address;
+            if (!shouldPenalizeAtSwitch(i)) blockHeaders[i].Penalties = [];
+            Hash256 hash = blockHeaders[i].Hash ?? blockHeaders[i].CalculateHash().ToHash256();
+            hashToHeader[hash] = blockHeaders[i];
+            blocks[i] = new Block(blockHeaders[i]);
+        }
+
+        blockTree.FindHeader(Arg.Any<Hash256>(), Arg.Any<long>())
+            .Returns(ci => blockHeaders[(int)(long)ci.Args()[1]]);
```
alexb5dh:
```suggestion
            .Returns(ci => blockHeaders[ci.ArgAt<long>(1)]);
```

alexb5dh: Also, worth adding hash match verification to cover mentioned bug. Or fetch by hash instead of by number.

alexb5dh (commented):
- Potential bug in `PenaltyHandler` - the only reason for not approving.
- Minor code suggestions, may be subjective.
- Would be nice to have some PR description :wink:

src/Nethermind/Nethermind.Xdc/Spec/XdcChainSpecBasedSpecProvider.cs
```diff
@@ -48,6 +48,10 @@ protected override ReleaseSpec CreateReleaseSpec(ChainSpec chainSpec, long relea
         releaseSpec.XDCXAddressBinary = chainSpecEngineParameters.XDCXAddressBinary;
         releaseSpec.TradingStateAddressBinary = chainSpecEngineParameters.TradingStateAddressBinary;
 
+        releaseSpec.LimitPenaltyEpoch = chainSpecEngineParameters.LimitPenaltyEpoch;
+        releaseSpec.LimitPenaltyEpochV2 = chainSpecEngineParameters.LimitPenaltyEpochV2;
+        releaseSpec.RangeReturnSigner = chainSpecEngineParameters.RangeReturnSigner;
+
```
batrr, line 55: Where is `TipUpgradePenalty` set?

src/Nethermind/Nethermind.Xdc/PenaltyHandler.cs
cicr99: That path should be unreachable in normal flow. `HandlePenalties` is called at an epoch switch, and the backward scan should stop at the previous epoch switch (at latest the switchBlock). In the tests, where `SwitchBlock = 0`, it returns/break before getting to  `parentNumber--;`, so `parentNumber` should never become -1

src/Nethermind/Nethermind.Xdc/PenaltyHandler.cs
cicr99: Yes, good catch! Just need to fix the number passed there.

On the fetched header: I don't think it's a BlockTree bug. `FindHeader(hash, number)` uses number as an optional lookup hint, looking at the code, it first tries `(number, hash)`, and on miss falls back to hash-only lookup. That fallback is likely what we were hitting here.

src/Nethermind/Nethermind.Xdc/XdcTransactionExtensions.cs
cicr99: Well, I guess it would only happen if it is not set in the configuration file, which shouldn't be the case

alexb5dh: So can just do?
```suggestion
        var targetIsSignContract = currentTx.To == spec.BlockSignerContract
```

src/Nethermind/Nethermind.Xdc/Spec/XdcChainSpecBasedSpecProvider.cs
cicr99: It's not active on xdc side, so I actually hadn't pay attention to this. It's wired now 🙂

src/Nethermind/Nethermind.Xdc/XdcTransactionExtensions.cs
cicr99: In theory yes, but I’d prefer to keep the current form. While BlockSignerContract is expected to be set in config, we don’t currently enforce that with a guard. Keeping the check explicit seems safer

PR #10277, opened 2026-01-20
XDC eip 1559

ak88:
introduces base fee calculation, which is currently just hardcoded value 
https://vscode.dev/github/XinFinOrg/XDPoSChain/blob/dev-upgrade/common/gas.go#L7


src/Nethermind/Nethermind.Xdc/XdcBaseFeeCalculator.cs
```diff
@@ -0,0 +1,21 @@
...
+using Nethermind.Int256;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Nethermind.Xdc;
+
+internal class XdcBaseFeeCalculator : IBaseFeeCalculator
+{
+    public const long BaseFee = 12500000000;
```
alexb5dh, line 17: Is there any spec avaialble as a source for this value? If yes, would be helpful to reference it in a comment.

PR #10280, opened 2026-01-20
fix(optimism): validate deposit log index

Galoretka:
SourceHash was computed with (ulong)(log.LogIndex ?? 0), which allowed null or negative LogIndex values to silently map to an incorrect index and produce invalid source hashes.
Added explicit validation for LogIndex and reject null/negative values before computing SourceHash for deposit transactions.

### DIFF
--- a/.github/workflows/ci-taiko.yml
+++ b/.github/workflows/ci-taiko.yml
@@ -1,0 +1,173 @@
+name: "Nethermind/Ethereum Taiko Client CI Tests"
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+    paths:
+      - "src/Nethermind/Nethermind.Taiko/**"
+      - ".github/workflows/ci-taiko.yml"
+
+concurrency:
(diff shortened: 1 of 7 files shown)


src/Nethermind/Nethermind.Optimism/CL/Derivation/DepositTransactionBuilder.cs
```diff
@@ -141,7 +141,12 @@ static Hash256 ComputeSourceHash(Hash256 l1BlockHash, ulong logIndex)
         if (version == DepositEvent.Version0)
         {
             var depositLogEventV0 = DepositLogEventV0.FromBytes(log.Data);
-            var sourceHash = ComputeSourceHash(log.BlockHash, (ulong)(log.LogIndex ?? 0)); // TODO: Unsafe cast with possible null;
+            if (log.LogIndex is null || log.LogIndex < 0)
```
alexb5dh, line 144: Don't think negative check here adds anything, we don't have it many other places. Just coalesce with `ArgumentNullException`.

PR #10288, opened 2026-01-21
consensus: standardize slow block JSON output for cross-client metrics

CPerezz:
## Summary

Implement standardized JSON format for slow block logging to enable cross-client performance analysis and protocol research.

This change is part of the **Cross-Client Execution Metrics** initiative proposed by Gary Rong and CPerezz.

## Motivation

Standardized execution metrics are critical for:
- Cross-client performance comparison
- Network health monitoring
- Data-driven protocol research

**Real-world example:** The [EIP-7907 analysis](https://ethresear.ch/t/data-driven-analysis-on-eip-7907/23850) used execution metrics to measure code read latency, per-call overhead scaling, and block execution breakdown. Without standardized metrics across clients, such analysis cannot be validated cross-client.

## References

- Original proposal: https://hackmd.io/dg7rizTyTXuCf2LSa2LsyQ (Gary Rong)
- EIP-7907 analysis: https://ethresear.ch/t/data-driven-analysis-on-eip-7907/23850

## Configuration

Two new flags in `Blocks` config section:

| Flag | Default | Description |
|------|---------|-------------|
| `Blocks.SlowBlockThresholdMs` | `-1` (disabled) | Threshold in ms for logging slow blocks. Set to `1000` to log blocks slower than 1s. Set to `0` to log all blocks. |
| `Blocks.SlowBlockPerTxThresholdMs` | `-1` (disabled) | Per-transaction threshold in ms. When enabled, individual transactions exceeding this threshold are included in the slow block JSON. Set to `0` to log all transactions. |

Both flags default to `-1` (disabled). When disabled, there is **zero overhead** on block processing — no metric captures, no delta computation, no JSON serialization.

When enabled, the overhead is ~31μs per block (~0.08% on a 40ms block) from reading thread-local counters. Per-tx timing adds ~6μs for a 150-tx block when enabled.

### Example usage

```
# Log blocks slower than 1s with all transactions
--Blocks.SlowBlockThresholdMs=1000 --Blocks.SlowBlockPerTxThresholdMs=0

# Log blocks slower than 500ms, only transactions slower than 50ms
--Blocks.SlowBlockThresholdMs=500 --Blocks.SlowBlockPerTxThresholdMs=50

# Log every block, no per-tx breakdown
--Blocks.SlowBlockThresholdMs=0
```

## JSON Output Format

```json
{
  "level": "warn",
  "msg": "Slow block",
  "block": {
    "number": 22075123,
    "hash": "0xabc...",
    "gas_used": 29850000,
    "gas_limit": 30000000,
    "tx_count": 142,
    "blob_count": 6
  },
  "timing": {
    "execution_ms": 52.1,
    "evm_ms": 45.2,
    "blooms_ms": 1.3,
    "receipts_root_ms": 3.8,
    "commit_ms": 2.1,
    "storage_merkle_ms": 18.4,
    "state_root_ms": 5.2,
    "state_hash_ms": 23.6,
    "total_ms": 82.0
  },
  "throughput": {
    "mgas_per_sec": 364.1
  },
  "state_reads": {
    "accounts": 1842,
    "storage_slots": 3201,
    "code": 215,
    "code_bytes": 482310
  },
  "state_writes": {
    "accounts": 312,
    "accounts_deleted": 0,
    "storage_slots": 1805,
    "storage_slots_deleted": 12,
    "code": 3,
    "code_bytes": 8420,
    "eip7702_delegations_set": 0,
    "eip7702_delegations_cleared": 0
  },
  "cache": {
    "account": { "hits": 1520, "misses": 322, "hit_rate": 82.52 },
    "storage": { "hits": 2800, "misses": 401, "hit_rate": 87.48 },
    "code": { "hits": 180, "misses": 35, "hit_rate": 83.72 }
  },
  "evm": {
    "opcodes": 1542000,
    "sload": 3201,
    "sstore": 1805,
    "calls": 2412,
    "empty_calls": 48,
    "creates": 3,
    "self_destructs": 0,
    "contracts_analyzed": 118,
    "cached_contracts_used": 97
  },
  "transactions": [
    {
      "index": 42,
      "hash": "0xdef...",
      "gas_used": 8200000,
      "execution_ms": 12.3,
      "type": "Legacy",
      "to": "0x7a25..."
    },
    {
      "index": 107,
      "hash": "0x123...",
      "gas_used": 5100000,
      "execution_ms": 8.7,
      "type": "EIP1559",
      "to": "0xdead..."
    }
  ]
}
```

### Timing field definitions

| Field | Definition |
|-------|-----------|
| `total_ms` | Wall-clock block processing time |
| `execution_ms` | `total - state_hash - commit` (backwards-compatible) |
| `evm_ms` | `total - state_hash - commit - blooms - receipts_root` (pure EVM) |
| `blooms_ms` | Bloom filter calculation (parallel) |
| `receipts_root_ms` | Receipts Patricia trie root computation |
| `commit_ms` | State commit without root recalculation |
| `storage_merkle_ms` | Storage trie merkleization (`Commit(commitRoots: true)`) |
| `state_root_ms` | State root recalculation + commit tree |
| `state_hash_ms` | `storage_merkle + state_root` (total merkleization) |

The `transactions` array is only present when `SlowBlockPerTxThresholdMs >= 0`.

## Implementation

- All metrics use `ZeroContentionCounter` (lock-free thread-local counters) — no locks, no contention
- Timing uses `Stopwatch.GetTimestamp()` only at coarse block-phase boundaries (not per state read/write)
- JSON is serialized via `Utf8JsonWriter` + `ArrayBufferWriter<byte>` on the ThreadPool, not the block-processing thread
- Per-tx timing uses a thread-static array with one `Stopwatch.GetTimestamp()` per tx (~20ns)
- Delta computation and JSON output are queued to ThreadPool via `UnsafeQueueUserWorkItem`


src/Nethermind/Nethermind.Evm/Metrics.cs
```diff
@@ -100,6 +101,94 @@ public class Metrics
...
+    internal static long ThreadLocalCodeWrites => _codeWrites.ThreadLocalValue;
+    internal static void IncrementCodeWrites() => _codeWrites.Increment();
+
+    [CounterMetric]
+    [Description("Total bytes of code written during execution.")]
+    public static long CodeBytesWritten => _codeBytesWritten.GetTotalValue();
+    private static readonly ZeroContentionCounter _codeBytesWritten = new();
+    [Description("Total bytes of code written on thread.")]
+    internal static long ThreadLocalCodeBytesWritten => _codeBytesWritten.ThreadLocalValue;
+    internal static void IncrementCodeBytesWritten(int bytes) => _codeBytesWritten.Increment(bytes);
```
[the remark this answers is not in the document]
CPerezz: Good catch. Addressed

src/Nethermind/Nethermind.Consensus/Processing/ProcessingStats.cs
```diff
@@ -40,8 +42,15 @@ internal class ProcessingStats
...
         private readonly IStateReader _stateReader;
         private readonly ILogger _logger;
+        private readonly ILogger _slowBlockLogger;
         private readonly Stopwatch _runStopwatch = new();
 
+        /// <summary>
+        /// Threshold in milliseconds for slow block logging (default: 1000ms).
+        /// Set to 0 to log all blocks.
+        /// </summary>
+        private readonly long _slowBlockThresholdMs;
```
LukaszRozmej, line 53: This could be configurable

src/Nethermind/Nethermind.Core.Test/MetricsSnapshot.cs
```diff
@@ -0,0 +1,177 @@
...
+using DbMetrics = Nethermind.Db.Metrics;
+using EvmMetrics = Nethermind.Evm.Metrics;
+
+namespace Nethermind.Core.Test;
+
+/// <summary>
+/// Captures a snapshot of thread-local metrics for delta calculation during tests.
+/// Used to verify that specific operations increment the expected metrics.
+/// </summary>
+public class MetricsSnapshot
```
LukaszRozmej: Is this used?

LukaszRozmej (commented):
Looks good, added some questions.

Food for thought @benaadams 
With so many metrics, maybe it would be useful to have them continuous in memory, so we could make copies by copying memory region and potentially delta's using SIMD?

src/Nethermind/Nethermind.Consensus/Processing/ProcessingStats.cs
CPerezz: This was already addressed in b9f5d1a.

src/Nethermind/Nethermind.Core.Test/MetricsSnapshot.cs
CPerezz: Nope! That's leftovers! Thanks!

src/Nethermind/Nethermind.Consensus/Processing/ProcessingStats.cs
```diff
@@ -413,6 +526,139 @@ UInt256 CalculateBalanceChange(BlockHeader? startBlock, BlockHeader endBlock, Ad
...
+                long storageCacheMisses = data.CurrentStorageCacheMisses - data.StartStorageCacheMisses;
+                long codeCacheHits = data.CurrentCodeCacheHits - data.StartCodeCacheHits;
+                long codeCacheMisses = data.CurrentCodeCacheMisses - data.StartCodeCacheMisses;
+
+                // Calculate hit rates
+                double accountHitRate = CalculateHitRate(accountCacheHits, accountCacheMisses);
+                double storageHitRate = CalculateHitRate(storageCacheHits, storageCacheMisses);
+                double codeHitRate = CalculateHitRate(codeCacheHits, codeCacheMisses);
+
+                var slowBlockLog = new
```
flcl42: Can be specific type instead of anon object

src/Nethermind/Nethermind.Evm.Test/MetricsIntegrationTests.cs
```diff
@@ -0,0 +1,532 @@
...
+    private void ResetMetrics()
+    {
+        // Thread-local metrics are reset by capturing the current values
+        // For testing, we'll capture start values and compute deltas
+    }
+
+    private void DeployCode(Address address, byte[] code)
+    {
+        _stateProvider.CreateAccountIfNotExists(address, 0);
+        _stateProvider.InsertCode(address, code, _specProvider.GetSpec(ForkActivation.TimestampOnly(MainnetSpecProvider.PragueBlockTimestamp)));
```
flcl42: Can user just `Prague.Instance` or so instead of getting it from the spec provider

src/Nethermind/Nethermind.State/PartialStorageProviderBase.cs
```diff
@@ -49,6 +50,8 @@ public ReadOnlySpan<byte> Get(in StorageCell storageCell)
         /// <param name="newValue">Value to store</param>
         public void Set(in StorageCell storageCell, byte[] newValue)
         {
+            // Track storage write for execution metrics
+            EvmMetrics.IncrementStorageWrites();
```
flcl42, line 53: @benaadams @LukaszRozmej maybe we can make some generic argument trick to avoid calling metric if don't need to(if MetricsOff is passed metrics are excluded due to unconditional if)?

src/Nethermind/Nethermind.Evm.Test/MetricsIntegrationTests.cs
CPerezz: Yes but this will need to be maintained always. In fact, we would have to use Osaka here to be sure we are fwd. compatible no?

I updated anyways, but I'm not fully convinced..

flcl42: It's just so using mainnet timestamps for single spec provider does not look perfect to me. Btw you can init a _releaseSpec some where with value like Prague.Instance or Osaka.Instance and reuse it.

If you mean that you would better use just the top mainnet spec, we might try to introduce Mainnet.Instance that points to the most recent fork and we change it when time comes. Can't recall of something existing like this

Marchhill:
update PR here: https://github.com/NethermindEth/nethermind/pull/11400

PR #10289, opened 2026-01-21
feat: add prewarmer get operation timing metrics

asdacap:
## Summary
- Add an optional timing measurement in the prewarmer scope.
- Uses prewarmer scope to be able to differentiate between warmed up or not warmed up reads, which allow measuring the avg time of cache miss.
- Also measure commit and write batch latency.

## Preview

Screenshot_2026-01-26_11-21-32

## Testing

- Tested manually.

### DIFF
--- a/src/Nethermind/Nethermind.Monitoring.Test/MetricsTests.cs
+++ b/src/Nethermind/Nethermind.Monitoring.Test/MetricsTests.cs
@@ -50,6 +50,9 @@ public class MetricsTests
         [System.ComponentModel.Description("A test description")]
         [DetailedMetric]
         public static long DetailedMetric { get; set; }
+
+        [DetailedMetricOnFlag]
+        public static bool DetailedMetricsEnabled { get; set; }
     }
 
     public enum SomeEnum
@@ -131,6 +134,7 @@ public class MetricsTests
         Dictionary<string, MetricsController.IMetricUpdater> updater = metricsController._individualUpdater;
         string metricName = "TestMetrics.DetailedMetric";
         Assert.That(updater.ContainsKey(metricName), Is.EqualTo(enableDetailedMetric));
+        Assert.That(TestMetrics.DetailedMetricsEnabled, Is.EqualTo(enableDetailedMetric));
     }
 
     [Test]
--- a/src/Nethermind/Nethermind.Monitoring/Metrics/MetricsController.cs
+++ b/src/Nethermind/Nethermind.Monitoring/Metrics/MetricsController.cs
@@ -29,7 +29,6 @@ namespace Nethermind.Monitoring.Metrics
 {
     public partial class MetricsController : IMetricsController
     {
-        public const string DetailedMetricFlagName = "DetailedMetricsEnabled";
         private readonly int _intervalMilliseconds;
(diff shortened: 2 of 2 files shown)


src/Nethermind/Nethermind.Db/Metrics.cs
```diff
@@ -125,5 +128,15 @@ public static class Metrics
...
+
+        [DetailedMetric]
+        [Description("Prewarmer get operation times")]
+        [ExponentialPowerHistogramMetric(Start = 10, Factor = 1.5, Count = 30, LabelNames = ["part", "is_prewarmer"])]
+        public static IMetricObserver PrewarmerGetTime { get; set; } = NoopMetricObserver.Instance;
+    }
+
+    public readonly struct PrewarmerGetTimeLabel(string part, bool isPrewarmer) : IMetricLabels
+    {
+        public string[] Labels => [part, isPrewarmer ? "true" : "false"];
```
LukaszRozmej: Save it in a field:

```suggestion
        public string[] Labels { get; } = [part, isPrewarmer ? "true" : "false"];
```

src/Nethermind/Nethermind.Monitoring.Test/MetricsTests.cs
LukaszRozmej: Btw this should be some more defined Exception type

src/Nethermind/Nethermind.Monitoring.Test/MetricsTests.cs
```diff
@@ -202,6 +202,7 @@ private static void ForEachProperty(Action<PropertyInfo> verifier)
                 PropertyInfo[] properties = metricsType.GetProperties(BindingFlags.Static | BindingFlags.Public);
                 foreach (PropertyInfo property in properties)
                 {
+                    if (property.Name == MetricsController.DetailedMetricFlagName) continue;
```
LukaszRozmej: Maybe better to develop some general mechanism like attribute for skipping or something?

src/Nethermind/Nethermind.State/PrewarmerScopeProvider.cs
```diff
@@ -47,42 +51,79 @@ public IWorldStateScopeProvider.IStorageTree CreateStorageTree(Address address)
...
+        {
+            if (!_measureMetric)
+            {
+                baseScope.Commit(blockNumber);
+                return;
+            }
+
+            long sw = Stopwatch.GetTimestamp();
+            baseScope.Commit(blockNumber);
+            _metricObserver.Observe(Stopwatch.GetTimestamp() - sw, new PrewarmerGetTimeLabel("commit", populatePreBlockCache));
```
LukaszRozmej: Can you create this `PrewarmerGetTimeLabel` only once and save it in some static variables? Maybe have static Dictionary or array that has them mapped?

src/Nethermind/Nethermind.Monitoring.Test/MetricsTests.cs
asdacap: Yea, maybe a "DetailedMetricOnFlag"

PR #10290, opened 2026-01-21
Update SurgeGasPriceOracle for Surge Shasta

dipkakwani:
Closes https://github.com/NethermindEth/Surge/issues/253

## Changes

- Update gas pricing formula for Surge Shasta

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [x] Other: Update gas price oracle for the new fork

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Taiko/Rpc/SurgeGasPriceOracle.cs
+++ b/src/Nethermind/Nethermind.Taiko/Rpc/SurgeGasPriceOracle.cs
@@ -22,7 +22,7 @@ namespace Nethermind.Taiko.Rpc;
 public class SurgeGasPriceOracle : GasPriceOracle
 {
     private const string ClassName = nameof(SurgeGasPriceOracle);
-    private const int BlobGasPerBlob = 131072;
+    private const ulong BlobGasPerBlob = Eip4844Constants.GasPerBlob;
 
     // ABI signatures and encoded function selectors for TaikoInbox contract.
     private static readonly AbiSignature GetCoreStateSignature = new("getCoreState");
@@ -97,7 +97,6 @@ public class SurgeGasPriceOracle : GasPriceOracle
                 await UpdateInboxRingBufferFullAsync();
             }
         }
-
 
         // Get the fee history from the L1 client with RPC
         L1FeeHistoryResults? feeHistory = await GetL1FeeHistory();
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Taiko/Rpc/SurgeGasPriceOracle.cs
```diff
@@ -74,15 +84,21 @@ public override async ValueTask<UInt256> GetGasPriceEstimate()
...
-            averageGasUsage = await GetAverageGasUsageAcrossBatches();
+            // Since the head block has changed, we need to re-compute the average gas usage and
+            // update the inbox ring buffer full status
+            _cachedAverageGasUsage = GetAverageGasUsagePerBlock();
+            averageGasUsage = _cachedAverageGasUsage;
+
+            if (!_isInboxRingBufferFull)
+            {
+                await UpdateInboxRingBufferFullAsync();
+            }
```
jmadibekov, line 98: Do we ever check and update this `_isInboxRingBufferFull` once this value becomes true? I think even after the inbox ring buffer becomes full, you should keep checking whether it becomes available and update it back to false at some point.

src/Nethermind/Nethermind.Taiko.Test/SurgeGasPriceOracleTests.cs
jmadibekov, line 1: You might as well add a test on a scenario that I outlined in my other comment. That is, ring buffer becoming full -> not full to test out that boolean flag is properly updated and is on track.

src/Nethermind/Nethermind.Taiko/Rpc/SurgeGasPriceOracle.cs
dipkakwani: I don't think this can go back from true -> false, once it becomes full it is assumed to be always full.
@AnshuJalan can you confirm this?

AnshuJalan: Correct. Once full, it's always full i.e that slots in the buffer are non-zero and we simply keep overwriting them.
It's not "full" in the sense that you can't put more items in it.

jmadibekov: Ah, so we're using this boolean flag purely for the gas calculation and when it's not full gas price is different because we're writing on a zero slot. Makes sense now, thanks for clarifying!

jmadibekov (approved):

PR #10305, opened 2026-01-23
XDC Subnet block header

batrr:
## Changes

- Added XdcSubnetBlockHeader (includes NextValidators)
- Added XdcSubnetHeaderDecoder for RLP encoding/decoding
- Added BaseXdcHeaderDecoder to share common logic

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Xdc/XdcSubnetBlockHeader.cs
+++ b/src/Nethermind/Nethermind.Xdc/XdcSubnetBlockHeader.cs
@@ -41,7 +41,7 @@ public class XdcSubnetBlockHeader : XdcBlockHeader
         set { _nextValidatorsAddress = value; }
     }
 
-    public new ValueHash256 CalculateHash()
+    public override ValueHash256 CalculateHash()
     {
         KeccakRlpStream rlpStream = new KeccakRlpStream();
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Xdc/XdcSubnetBlockHeader.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            if (_nextValidatorsAddress is not null)
+                return _nextValidatorsAddress;
+            _nextValidatorsAddress = XdcExtensions.ExtractAddresses(NextValidators);
+            return _nextValidatorsAddress;
+        }
+        set { _nextValidatorsAddress = value; }
+    }
+
+    public new ValueHash256 CalculateHash()
```
ak88: This will unfortunately fail because `new` doesnt override, but just hides the function. So when other classes that doesnt see it actual type, it will call `CalculateHash` on the base class instead.

ak88 (approved):

cicr99 (approved):

PR #10306, opened 2026-01-23
fix: correct off-by-one in ArrayPoolListCore.RemoveAt

radik878:
## Changes

RemoveAt used an incorrect copy length (count - index instead of count - index - 1) when shifting elements left. This could construct a Memory slice that extends past the logical end of the array when capacity equals count, leading to an ArgumentOutOfRangeException and doing extra work beyond the valid range.
Adjusted the copy length in ArrayPoolListCore.RemoveAt to use count - index - 1 so only the valid tail segment [index + 1 .. count - 1] is shifted. This keeps copying within the logical range, fixes the potential out-of-range slice when capacity == count, and avoids unnecessary copying.

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Synchronization/SnapSync/SnapProviderHelper.cs
+++ b/src/Nethermind/Nethermind.Synchronization/SnapSync/SnapProviderHelper.cs
@@ -78,13 +78,10 @@ namespace Nethermind.Synchronization.SnapSync
                     codeHashes.Add(account.Account.CodeHash);
                 }
 
-                var account_ = account.Account;
-                Rlp rlp = account_ is null ? null : account_.IsTotallyEmpty ? StateTree.EmptyAccountRlp : Rlp.Encode(account_);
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Core/Collections/ArrayListCore.cs
```diff
@@ -192,7 +192,7 @@ public static bool RemoveAt(T[] array, ref int count, int index, bool shouldThro
             int start = index + 1;
             if (start < count)
             {
-                array.AsMemory(start, count - index).CopyTo(array.AsMemory(index));
+                array.AsMemory(start, count - index - 1).CopyTo(array.AsMemory(index));
```
LukaszRozmej: I think this was intended and easier to read:

```suggestion
                array.AsMemory(start, count - start).CopyTo(array.AsMemory(index));
```

PR #10321, opened 2026-01-26
fix(xdc-tests): Add proper disposal and prevent parallel execution

asdacap:
## Summary
- Add `using` declarations to all `XdcTestBlockchain` instances for proper resource cleanup
- Add `[NonParallelizable]` attribute to test classes that use `XdcTestBlockchain` to prevent race conditions
- Add `[TearDown]` method to `HeaderVerificationTests` to dispose the blockchain field
- Remove `Setup()` helper in `MineModuleTests`, inline with `using` in each test

## Test plan
- [x] Verified 33 tests pass: XdcTestBlockchainTests, TimeoutTests, VoteTests, HeaderVerificationTests, BlockInfoTests, RewardTests
- [ ] Note: Some tests in MineModuleTests, ProposedBlockTests, and XdcReorgModuleTests have pre-existing hang issues related to `useHotStuffModule: true` that require separate investigation

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/Nethermind/Nethermind.Xdc.Test/ModuleTests/HeaderVerificationTests.cs
+++ b/src/Nethermind/Nethermind.Xdc.Test/ModuleTests/HeaderVerificationTests.cs
@@ -22,6 +22,7 @@ using ISigner = Nethermind.Consensus.ISigner;
 
 namespace Nethermind.Xdc.Test.ModuleTests;
 
+[NonParallelizable]
 internal class HeaderVerificationTests
 {
     private XdcTestBlockchain xdcTestBlockchain;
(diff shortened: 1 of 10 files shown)


src/Nethermind/Nethermind.Xdc.Test/ModuleTests/HeaderVerificationTests.cs
```diff
@@ -22,6 +22,7 @@
 
 namespace Nethermind.Xdc.Test.ModuleTests;
 
+[NonParallelizable]
```
LukaszRozmej, line 25: It can probably be parallel with:

`[FixtureLifeCycle(LifeCycle.InstancePerTestCase)]`

LukaszRozmej (commented):
Parallelize with `[FixtureLifeCycle(LifeCycle.InstancePerTestCase)]`

PR #10323, opened 2026-01-26
fix: apply VerifyChecksum config to column families and snapshots

MozirDmitriy:
`VerifyChecksum` config option was only applied to the main `DbOnTheRocks` reader. Column families, snapshots, and `GetViewBetween` operations ignored this setting and always used RocksDB defaults. If user set `VerifyChecksum = false` for performance, it only worked for the main DB, not for column families — breaking expected behavior.

Changes:

- Added internal VerifyChecksum property to DbOnTheRocks to expose config value
- Applied SetVerifyChecksums() to all ReadOptions in:
  - ColumnDb constructor, FirstKey, LastKey, CreateSnapshot()
  - DbOnTheRocks.GetViewBetween(), CreateSnapshot()
  - ColumnsDb.ColumnDbSnapshot

### DIFF
--- a/src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
+++ b/src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
@@ -117,12 +117,7 @@ public partial class DbOnTheRocks : IDb, ITunableDb, IReadOnlyNativeKeyValueStor
         _db = Init(basePath, dbSettings.DbPath, dbConfig, logManager, columnFamilies, dbSettings.DeleteOnStart, sharedCache);
         _iteratorManager = new IteratorManager(_db, null, _readAheadReadOptions);
 
-        _reader = new RocksDbReader(this, () =>
-        {
-            ReadOptions readOptions = new ReadOptions();
-            readOptions.SetVerifyChecksums(_perTableDbConfig.VerifyChecksum ?? true);
-            return readOptions;
-        }, _iteratorManager, null);
+        _reader = new RocksDbReader(this, CreateReadOptions, _iteratorManager, null);
     }
 
     protected virtual RocksDb DoOpen(string path, (DbOptions Options, ColumnFamilies? Families) db)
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Db.Rocks/ColumnDb.cs
```diff
@@ -30,8 +30,9 @@ public ColumnDb(RocksDb rocksDb, DbOnTheRocks mainDb, string name)
         _iteratorManager = new DbOnTheRocks.IteratorManager(_rocksDb, _columnFamily, _mainDb._readAheadReadOptions);
         _reader = new RocksDbReader(mainDb, () =>
         {
-            // TODO: Verify checksum not set here.
-            return new ReadOptions();
+            ReadOptions readOptions = new ReadOptions();
+            readOptions.SetVerifyChecksums(mainDb.VerifyChecksum);
```
LukaszRozmej: Can we extract it to some `CreateReadOptions` method - so next time we won't miss it?

src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
```diff
@@ -1971,6 +1972,7 @@ public ISortedView GetViewBetween(ReadOnlySpan<byte> firstKey, ReadOnlySpan<byte
     internal ISortedView GetViewBetween(ReadOnlySpan<byte> firstKey, ReadOnlySpan<byte> lastKey, ColumnFamilyHandle? cf)
     {
         ReadOptions readOptions = new ReadOptions();
+        readOptions.SetVerifyChecksums(VerifyChecksum);
```
LukaszRozmej: Same here...

MozirDmitriy: > Same here...

Did it!

LukaszRozmej (approved):

PR #10329, opened 2026-01-26
Db options for XDC removed

ak88:
Utilizes the no validation option on PerTableDbConfig so XDC options can be removed.


src/Nethermind/Nethermind.Db.Rocks/Config/DbConfig.cs
```diff
@@ -271,4 +271,7 @@ public class DbConfig : IDbConfig
     public string L1OriginDbRocksDbOptions { get; set; } = "";
 
     public string? L1OriginDbAdditionalRocksDbOptions { get; set; }
+
+    public string SnapshotsDbRocksDbOptions { get; set; } = "";
```
alexb5dh: Is this common practice to add options used only in a single/few chain(s) directly to `DbConfig`? Any alternatives?
Maybe `IRocksDbConfigFactory` suggestion above will work?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Xdc.Test/ModuleTests/XdcRocksDbConfigFactoryTests.cs
```diff
@@ -0,0 +1,35 @@
...
+        public void CustomFactory_DoesNotThrow_And_Returns_ConfigForSnapshot()
+        {
+            IDbConfig dbConfig = new DbConfig();
+            var pruning = Substitute.For<IPruningConfig>();
+            var hw = Substitute.For<IHardwareInfo>();
+            var logManager = Substitute.For<ILogManager>();
+            var baseFactory = new RocksDbConfigFactory(dbConfig, pruning, hw, logManager, validateConfig: true);
+            var custom = new XdcRocksDbConfigFactory(baseFactory, dbConfig);
+
+            IRocksDbConfig config = custom.GetForDatabase("XdcSnapshots", null);
```
alexb5dh: Would suggest to make `XdcModule.SnapshotDbName` public and utiilize it here. Or vice versa. I suspect it will also be needed in other places later.

src/Nethermind/Nethermind.Xdc/XdcRocksDbConfigFactory.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Nethermind.Xdc;
+
+/// <summary>
+/// A custom RocksDb config factory for XDC that bypasses validation for the XdcSnapshots database.
+/// This is necessary because XdcSnapshots doesn't have dedicated RocksDb options in the config.
+/// </summary>
+public class XdcRocksDbConfigFactory(IRocksDbConfigFactory baseFactory, IDbConfig dbConfig) : IRocksDbConfigFactory
+{
+    private const string SnapshotDbName = "XdcSnapshots";
```
alexb5dh: Same comment regarding using `XdcModule.SnapshotDbName`. Seems better than having identical string literals accross codebase.

src/Nethermind/Nethermind.Xdc/XdcRocksDbConfigFactory.cs
```diff
@@ -0,0 +1,28 @@
...
+
+    public IRocksDbConfig GetForDatabase(string databaseName, string? columnName)
+    {
+        // For XdcSnapshots, create PerTableDbConfig with validate=false since it doesn't have
+        // XdcSnapshotsDbRocksDbOptions configured in IDbConfig
+        if (databaseName == SnapshotDbName)
+        {
+            return new PerTableDbConfig(dbConfig, databaseName, columnName, validate: false);
```
asdacap, line 22: I'd make a NoopTableDbConfig instead of using the PerTableDbConfig.

PR #10338, opened 2026-01-27
Special required XDC masternode sort

ak88:
This is a special XDC masternode sort which is ported from here:
https://github.com/XinFinOrg/XDPoSChain/blob/ab393862426405b672999eeaa7f418e1d6eeb8e6/common/sort/slice.go#L29

We have to sort in the exact same manner or we can get consensus issue. Their implementation can swap elements despite being equal so the default sort does not work.

### DIFF
--- a/src/Nethermind/Nethermind.Xdc.Test/XdcSortTests.cs
+++ b/src/Nethermind/Nethermind.Xdc.Test/XdcSortTests.cs
@@ -62,12 +62,12 @@ internal class XdcSortTests
                 Stake = 10_000_000
             })
             .ToArray();
-        
+
         // Sort is deterministic but not stable: equal elements are reordered from original positions
         expectedOrder = new[] { 5, 4, 3, 2, 1, 12, 11, 19, 17, 15, 13, 6, 14, 7, 16, 8, 18, 9, 0, 10 }
             .Select(i => new Address($"0x{i:D40}"))
             .ToArray();
-        
+
         yield return new TestCaseData(candidatesAndStake, expectedOrder);
     }
 


src/Nethermind/Nethermind.Xdc.Test/XdcSortTests.cs
```diff
@@ -0,0 +1,66 @@
...
+using Nethermind.Xdc.Contracts;
+using NUnit.Framework;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Nethermind.Xdc.Test;
+
+internal class XdcSortTests
+{
+    public static IEnumerable<TestCaseData> CandidatesWithStake()
```
flcl42, line 16: So looks like it sorts by -stake and then by -index. Is it so risky just to use standard sorting?

src/Nethermind/Nethermind.Xdc.Test/XdcSortTests.cs
```diff
@@ -0,0 +1,66 @@
...
+        candidatesAndStake =
+            [
+            new CandidateStake() { Address = TestItem.AddressB, Stake = 10_000_000},
+            new CandidateStake() { Address = TestItem.AddressC, Stake = 10_000_000},
+            new CandidateStake() { Address = TestItem.AddressA, Stake = 10_000_000}
+            ];
+        expectedOrder = [TestItem.AddressA, TestItem.AddressC, TestItem.AddressB];
+
+        yield return new TestCaseData(candidatesAndStake, expectedOrder);
+    }
```
batrr, line 72: Go’s sort is unstable (equal elements may be reordered), so let’s add a test with a more complex pattern to lock this behavior:

```suggestion
        // Test with 20 items with same stake to verify unstable sort behavior
        candidatesAndStake = Enumerable.Range(0, 20)
            .Select(i => new CandidateStake()
            {
                Address = new Address($"0x{i:D40}"),
                Stake = 10_000_000
            })
            .ToArray();
        
        // Sort is deterministic but not stable: equal elements are reordered from original positions
        expectedOrder = new[] { 5, 4, 3, 2, 1, 12, 11, 19, 17, 15, 13, 6, 14, 7, 16, 8, 18, 9, 0, 10 }
            .Select(i => new Address($"0x{i:D40}"))
            .ToArray();
        
        yield return new TestCaseData(candidatesAndStake, expectedOrder);
    }
```

src/Nethermind/Nethermind.Xdc.Test/XdcSortTests.cs
ak88: Yes their version will swap equal elements, so we cannot use default sorting algos

src/Nethermind/Nethermind.Xdc.Test/Contracts/MasternodeVotingContractTests.cs
```diff
@@ -27,6 +27,7 @@
 using NUnit.Framework;
 using System;
 using System.Collections.Generic;
+using System.Linq;
```
alexb5dh, line 30: This is probably not needed?

src/Nethermind/Nethermind.Xdc/XdcSort.cs
```diff
@@ -0,0 +1,250 @@
...
+            throw new ArgumentNullException(nameof(x));
+        if (less == null)
+            throw new ArgumentNullException(nameof(less));
+
+        int length = x.Count;
+        var data = new LessSwap<T>(x, less);
+        QuickSort_func(data, 0, length, MaxDepth(length));
+    }
+
+    private static int MaxDepth(int n)
```
alexb5dh, line 53: Can be replaced with `BitOperations`.

For example, for non-negative `n`:
```csharp
(32 - BitOperations.LeadingZeroCount((uint)n)) * 2
```
May need small tweaks if negative values are allowed.

src/Nethermind/Nethermind.Xdc.Test/XdcSortTests.cs
alexb5dh: Just a note: personally would prefer to do it as C# `OrderBy` followed by a single pass swapping range of equal values to an expected order - just to minimize amount of custom code - but is probably harder to implement and maintain.

alexb5dh (approved):

src/Nethermind/Nethermind.Xdc.Test/XdcSortTests.cs
flcl42: @ak88 they do no seem to be fully equal if we consider addition order

PR #10343, opened 2026-01-28
simple flat refactor

LukaszRozmej:


src/Nethermind/Nethermind.State.Flat/Persistence/BasePersistence.cs
```diff
@@ -77,70 +90,70 @@ public interface ITrieReader
...
         {
             using NettyRlpStream stream = _accountDecoder.EncodeToNewNettyStream(account);
-
-            flatWriteBatch.SetAccount(addrHash, stream.AsSpan());
+            _flatWriteBatch.SetAccount(addrHash, stream.AsSpan());
         }
     }
 
-    public readonly struct ToHashedFlatReader<TFlatReader>(
+    public struct ToHashedFlatReader<TFlatReader>(
```
LukaszRozmej, line 137: either `flatReader` needs to be readonly (readonly on `IFlatReader` methods), or this shouldn't be readonly, same below. Otherwise each internal access makes a copy!

PR #10353, opened 2026-01-29
fix: correct Bytes.BytesComparer length comparison ordering

asdacap:
## Summary

- Fixed `Bytes.BytesComparer.Compare` which had inverted length comparison logic
- Previously: `null > non-null`, `[prefix] > [prefix, more]` (incorrect)
- Now: `null < non-null`, `[prefix] < [prefix, more]` (correct lexicographic ordering)
- Simplified implementation using `SequenceCompareTo` for proper lexicographic comparison

## Test plan

- [x] Updated `BytesTests.Compares_bytes_properly` test cases with correct expected values
- [x] All existing tests pass (`dotnet test` on Core.Test, Db.Test, Trie.Test)
- [x] No unexpected behavior changes detected

### DIFF
--- a/src/Nethermind/Nethermind.Core/Extensions/Bytes.cs
+++ b/src/Nethermind/Nethermind.Core/Extensions/Bytes.cs
@@ -76,6 +76,8 @@ namespace Nethermind.Core.Extensions
                     return y is null ? 0 : -1;
                 }
 
+                if (y is null) return 1;
+
                 return x.SequenceCompareTo(y);
             }
 


src/Nethermind/Nethermind.Core/Extensions/Bytes.cs
flcl42, line 68: Probably `Comparer<byte[]?>`

src/Nethermind/Nethermind.Core/Extensions/Bytes.cs
```diff
@@ -100,7 +100,7 @@ public override int Compare(byte[]? x, byte[]? y)
                     }
                 }
 
-                return y.Length > x.Length ? 1 : 0;
+                return y.Length > x.Length ? -1 : 0;
             }
 
             public static int Compare(ReadOnlySpan<byte> x, ReadOnlySpan<byte> y)
```
flcl42, line 84: Wondering `x.SequenceCompareTo(y);` behaves the same?

asdacap: Yea, this is the fix to make it behave the same.

flcl42: Do can we replace `public static int Compare(ReadOnlySpan<byte> x, ReadOnlySpan<byte> y)` impl with just `SequenceCompareTo` call?

asdacap: This code is probably benchmarked before, so I would not do that.

LukaszRozmej: IMO we should as `SequenceCompareTo` probably uses SIMD

src/Nethermind/Nethermind.Core/Extensions/Bytes.cs
```diff
@@ -73,24 +73,24 @@ public override int Compare(byte[]? x, byte[]? y)
...
 
                 for (int i = 0; i < x.Length; i++)
                 {
                     if (y.Length <= i)
                     {
-                        return -1;
+                        return 1;
                     }
 
                     int result = x[i].CompareTo(y[i]);
```
LukaszRozmej: same here use `SequenceCompareTo`

src/Nethermind/Nethermind.Core/Extensions/Bytes.cs
LukaszRozmej: @asdacap ?

PR #10360, opened 2026-01-30
fix: avoid false overflow on modexp lengths

eeemmmmmm:
This change fixes a bad overflow check in the ModExp precompile. The old logic used a bitwise OR and could reject perfectly valid inputs as “too large.” Now we only flag real overflow sentinels, so valid modexp calls no longer fail incorrectly.


src/Nethermind/Nethermind.Evm.Precompiles/ModExpPrecompile.cs
```diff
@@ -121,7 +121,7 @@ private static bool ExceedsMaxInputSize(IReleaseSpec releaseSpec, uint baseLengt
     {
         return releaseSpec.IsEip7823Enabled
             ? (baseLength > ModExpMaxInputSizeEip7823 | expLength > ModExpMaxInputSizeEip7823 | modulusLength > ModExpMaxInputSizeEip7823)
```
LukaszRozmej: should this also be? I think it doesn't really matter?

```suggestion
            ? (baseLength > ModExpMaxInputSizeEip7823 || expLength > ModExpMaxInputSizeEip7823 || modulusLength > ModExpMaxInputSizeEip7823)
```

src/Nethermind/Nethermind.Evm.Precompiles/ModExpPrecompile.cs
```diff
@@ -121,7 +121,7 @@ private static bool ExceedsMaxInputSize(IReleaseSpec releaseSpec, uint baseLengt
...
-            : (baseLength | modulusLength) >= uint.MaxValue;
+            : baseLength == uint.MaxValue || modulusLength == uint.MaxValue;
```
benaadams: Should still be `>=` not `==`? What if the value is higher?

src/Nethermind/Nethermind.Evm.Precompiles/ModExpPrecompile.cs
eeemmmmmm: yeah, you're right it doesn't affect logic, but || is cleaner here. done

src/Nethermind/Nethermind.Evm.Precompiles/ModExpPrecompile.cs
eeemmmmmm: good point, changed to >= for consistency. though for uint it's technically the same since uint.MaxValue is the ceiling

PR #10366, opened 2026-02-01
Skip cache for identity precompile

benaadams:
## Changes

- Introduce a SupportsCaching flag on IPrecompile (default true) and mark IdentityPrecompile as non-caching. 
- Update CachedCodeInfoRepository to only wrap precompiles in a CachedPrecompile when SupportsCaching is true (preserving behavior when cache is null). 
- Add comprehensive tests: CachedCodeInfoRepositoryTests validate wrapping behavior, cache hits, multiple inputs, and interactions with Identity/SHA256 precompiles; SupportsCachingTests assert which precompiles enable caching by default.
- These changes prevent inefficient caching for Identity and ensure correct caching semantics.

- [x] Optimization

#### Requires testing

- [x] Yes

- [x] Yes

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain/CachedCodeInfoRepository.cs
+++ b/src/Nethermind/Nethermind.Blockchain/CachedCodeInfoRepository.cs
@@ -64,7 +64,7 @@ public class CachedCodeInfoRepository(
     {
         IPrecompile precompile = originalPrecompile.Value.Precompile!;
 
-        return !precompile.SupportsCaching 
+        return !precompile.SupportsCaching
             ? originalPrecompile.Value
             : new PrecompileInfo(new CachedPrecompile(originalPrecompile.Key.Value, precompile, cache));
     }


src/Nethermind/Nethermind.Blockchain/CachedCodeInfoRepository.cs
```diff
@@ -60,8 +60,17 @@ public bool TryGetDelegation(Address address, IReleaseSpec spec,
...
+        ConcurrentDictionary<PreBlockCaches.PrecompileCacheKey, Result<byte[]>> cache)
+    {
+        IPrecompile precompile = originalPrecompile.Value.Precompile!;
+
+        if (!precompile.SupportsCaching)
+        {
+            return originalPrecompile.Value;
+        }
+
+        return new PrecompileInfo(new CachedPrecompile(originalPrecompile.Key.Value, precompile, cache));
```
LukaszRozmej:
```suggestion
        return !precompile.SupportsCaching 
            ? originalPrecompile.Value
            : new PrecompileInfo(new CachedPrecompile(originalPrecompile.Key.Value, precompile, cache));
```

LukaszRozmej (approved):

PR #10377, opened 2026-02-02
Sync Taiko Geth Changes into NMC

dipkakwani:
Adds latest changes from taiko-geth into NMC.

## Changes
- TaikoHeaderValidator: Skip EIP-4396 verification when ancestor block is missing instead of returning validation error. Logs warning and continues. (taiko-geth commit 5d7599848)
- PreBuiltTxList: Change JSON property names from PascalCase to camelCase to match taiko-geth RPC response format. (taiko-geth commit f1856b667)

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [x] Other: Aligning NMC with the latest taiko-geth changes

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


src/Nethermind/Nethermind.Taiko/Rpc/PreBuiltTxList.cs
```diff
@@ -9,14 +9,14 @@ namespace Nethermind.Taiko.Rpc;
 
 public sealed class PreBuiltTxList(TransactionForRpc[] transactions, ulong estimatedGasUsed, ulong bytesLength)
 {
-    [JsonPropertyName("TxList")]
+    [JsonPropertyName("txList")]
```
dipkakwani: This is the default name formatting so can be removed. On the other hand, having it explicitly mentioned here doesn't look bad to me, can be helpful in some cases.

smartprogrammer93: Standard in nethermind is to not have it afaik. So lets stick with that

PR #10381, opened 2026-02-02
Add LINQ guideline to agents

benaadams:
## Changes

- Update AGENTS.md: Add a new guideline advising against recommending LINQ when a simple loop would suffice. Is annoying that copilot keeps suggesting it in code reviews

- [x] Other: Agent update

#### Requires testing

- [x] No

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,3 +1,0 @@
-# GitHub Copilot Instructions
-
-Please follow the guidelines in [AGENTS.md](../AGENTS.md) when working on this repository.


.github/copilot-instructions.md
rubo, line 1: Why is this?

benaadams: Not convinced copilot reviewer uses agents (vs asking it to make code); docs keep saying use .github/copilot-instructions.md so just hedging

https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions

rubo: It does. From your link above:

> You can create one or more AGENTS.md files, stored anywhere within the repository. When Copilot is working, the nearest AGENTS.md file in the directory tree will take precedence over other agent instructions files. For more information, see the [openai/agents.md repository](https://github.com/openai/agents.md).
Alternatively, you can use a single CLAUDE.md or GEMINI.md file stored in the root of the repository.

Also, see https://github.blog/changelog/2025-08-28-copilot-coding-agent-now-supports-agents-md-custom-instructions

rubo: Drop this file, and we're good to merge.

benaadams: Removed

PR #10386, opened 2026-02-02
XDC - Pay transaction fees to owner after TIPTRC21Fee activation

cicr99:
TIPTRC21Fee changes XDC fee distribution so that transaction fees are paid to the masternode owner instead of the coinbase (check [here](https://github.com/XinFinOrg/XDPoSChain/blob/76a53321df1da9747559cecd819b9a9529c2f0bb/core/state_transition.go#L461) for reference). This PR:
- Implements post-TIPTRC21Fee fee routing in the XdcTransactionProcessor
- Routes fees to the masternode owner when TIPTRC21Fee is enabled
- Preserves special transaction handling and fork-aware behavior via the XDC spec

The XDC codebase also uses TIPTRC21Fee in other contexts, which are intentionally out of scope for this PR:
- TRC21 token fee accounting (`GetGasFee`, token-based fee deduction in the state processor)
- Transaction pool gas price validation (`GetGasPrice`, `GetMinGasPrice`)
These paths are part of XDC’s `TRC21` token mechanics, which won't be implemented atm.


src/Nethermind/Nethermind.Xdc/XdcTransactionProcessor.cs
```diff
@@ -13,24 +12,66 @@
...
+        }
+
+        Address coinbase = header.GasBeneficiary!;
+        Address owner = _masternodeVotingContract.GetCandidateOwner(header, coinbase);
+
+        if (owner is null || owner == Address.Zero)
+            return;
+
+        UInt256 fee = premiumPerGas * (ulong)spentGas;
+        WorldState.AddToBalanceAndCreateIfNotExists(owner, fee, spec);
```
Demuirgos, line 71: does this actually use the TRC20 token or just plain base XDC token? i was expecting maybe a smart contract call or something similar

PR #10389, opened 2026-02-02
Optimize Json hex parsing

benaadams:
## Changes

  Optimize hex parsing and JSON byte array conversion with focus on eliminating GS
  cookies, reducing allocations, and improving SIMD throughput.

  ### HexConverter Improvements

  | Method | Optimization | Impact |
  |--------|--------------|--------|
  | `TryDecodeFromUtf8_Scalar` | 2x loop unrolling (4 hex chars → 2 bytes/iter) |  Better ILP, fewer loop iterations |
  | `TryDecodeFromUtf8_Vector128` | 2x unrolling (32 hex bytes/iter), hoisted shuffle vector | ~2x throughput |
  | `TryDecodeFromUtf8_Vector256` | 2x unrolling (64 hex bytes/iter), hoisted shuffle vector | ~2x throughput |
  | `TryDecodeFromUtf8_Vector512` | 2x unrolling (128 hex bytes/iter), hoisted shuffle vector | ~2x throughput |
  | All methods | Combined validity check with single OR | Fewer branches |
  | All methods | Added `[SkipLocalsInit]` | Reduced prolog zeroing |

  ### ByteArrayConverter Improvements

  | Change | Before | After |
  |--------|--------|-------|
  | Hex prefix constant | `static readonly ushort` (runtime init) | `const ushort`  (compile-time) |
  | Span slicing | `hex[2..]` slice operator | `MemoryMarshal.CreateReadOnlySpan` (no  bounds check) |
  | 2-byte buffer | `stackalloc byte[2]` | `ushort` + `MemoryMarshal.CreateSpan` (no  GS cookie) |
  | Output tracking | Span slicing per write | Index-based with `ref byte` (fewer  copies) |

  ### Bytes32Converter Improvements

  | Method | Before | After | Reduction |
  |--------|--------|-------|-----------|
  | `Read` | `stackalloc byte[64]` | `Vector512<byte>` | -11.5% code size |
  | `Write` | `stackalloc byte[32]` + early return | `Vector256<byte>` + single  Convert call | -27.4% code size |

  Key changes:
  - Vector types as fixed-size buffers eliminate GS cookie overhead
  - Single `ByteArrayConverter.Convert` call avoids JIT code duplication
  - Cold path `ThrowJsonException()` extracted with `[DoesNotReturn, StackTraceHidden]`
  - Fast hex prefix check using `Unsafe.As<byte, ushort>`

  ### MemoryByteConverter

  - Simplified `Read` method by reusing `ByteArrayConverter.Convert` instead of duplicating logic

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Serialization.Json/ByteArray32Converter.cs
+++ b/src/Nethermind/Nethermind.Serialization.Json/ByteArray32Converter.cs
@@ -18,7 +18,7 @@ namespace Nethermind.Serialization.Json
 
     public class Bytes32Converter : JsonConverter<byte[]>
     {
-        private const int HexPrefix = 0x7830;
+        private const ushort HexPrefix = 0x7830;
 
         [SkipLocalsInit]
         public override byte[] Read(


src/Nethermind/Nethermind.Serialization.Json/ByteArray32Converter.cs
```diff
@@ -2,55 +2,81 @@
...
+                ThrowJsonException();
             }
 
             if (hex.Length < 64)
             {
-                Span<byte> hex32 = stackalloc byte[64];
-                hex32.Fill((byte)'0');
-                hex.CopyTo(hex32[(64 - hex.Length)..]);
-                return Bytes.FromUtf8HexString(hex32);
+                // Use Vector512<byte> as 64-byte buffer instead of stackalloc (avoids GS cookie)
```
asdacap, line 42: What is a GS cookie?

benaadams: Security cookie that gets added to check if stack overwrite/return address hijacking and to failfast if it does 

https://learn.microsoft.com/en-us/cpp/build/reference/gs-buffer-security-check?view=msvc-170

PR #10394, opened 2026-02-03
(refactor) Make comparisons between transactions more readable

LukaszRozmej:
## Changes

- Make comparisons between transactions more readable for humans
- Add missing test

- [x] Refactoring

#### Requires testing

- [x] Yes

- [x] Yes

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Comparers/BlobTxPriorityComparer.cs
+++ b/src/Nethermind/Nethermind.Consensus/Comparers/BlobTxPriorityComparer.cs
@@ -9,8 +9,8 @@ namespace Nethermind.Consensus.Comparers;
 
 /// <summary>
 /// Tie-breaker comparer that prefers blob transactions over non-blob transactions.
-/// Returns <see cref="TxComparisonResult.FirstIsBetter"/> if only x supports blobs,
-/// <see cref="TxComparisonResult.SecondIsBetter"/> if only y supports blobs,
+/// Returns <see cref="TxComparisonResult.XFirst"/> if only x supports blobs,
+/// <see cref="TxComparisonResult.YFirst"/> if only y supports blobs,
 /// or <see cref="TxComparisonResult.Equal"/> if both or neither support blobs.
 /// </summary>
 public sealed class BlobTxPriorityComparer : IComparer<Transaction>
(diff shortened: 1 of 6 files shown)


src/Nethermind/Nethermind.Consensus/Producers/TxPoolTxSource.cs
```diff
@@ -138,7 +139,7 @@ private static IEnumerable<Transaction> PickBlobTxsBetterThanCurrentTx(ArrayPool
             while (selectedBlobTxs.Count > 0)
             {
                 Transaction blobTx = selectedBlobTxs[0];
-                if (comparer.Compare(blobTx, tx) > 0)
+                if (comparer.Compare(blobTx, tx) < Equal)
```
flcl42, line 142: Looks not obvious, can we compare using `==`?

src/Nethermind/Nethermind.Consensus/Comparers/BlobTxPriorityComparer.cs
```diff
@@ -0,0 +1,31 @@
...
+public sealed class BlobTxPriorityComparer : IComparer<Transaction>
+{
+    public static readonly BlobTxPriorityComparer Instance = new();
+
+    private BlobTxPriorityComparer() { }
+
+    public int Compare(Transaction? x, Transaction? y)
+    {
+        if (ReferenceEquals(x, y)) return TxComparisonResult.Equal;
+        if (x is null) return TxComparisonResult.FirstIsBetter;
```
flcl42: Why null is "better"?

src/Nethermind/Nethermind.Consensus/Producers/TxPoolTxSource.cs
LukaszRozmej: no we can't due to comparision logic on numerical types

src/Nethermind/Nethermind.Consensus/Comparers/BlobTxPriorityComparer.cs
LukaszRozmej: yeah it doesn't look good. `null` is first in ordering though, by default, thinking about it. Maybe you have some suggestions?

marcindsobczak (approved):

PR #10411, opened 2026-02-05
Add XdcBeaconSyncStrategy to provide target block height from sync config

cicr99:


src/Nethermind/Nethermind.Xdc/XdcBeaconSyncStrategy.cs
```diff
@@ -0,0 +1,36 @@
...
+public class XdcBeaconSyncStrategy : IBeaconSyncStrategy
+{
+    private readonly ISyncConfig _syncConfig;
+
+    public XdcBeaconSyncStrategy(ISyncConfig syncConfig)
+    {
+        _syncConfig = syncConfig;
+    }
+
+    public void AllowBeaconHeaderSync() { }
```
Demuirgos, line 20: Are these methods supposed to be empty?

batrr (approved):

PR #10415, opened 2026-02-05
Add SeqlockCache

benaadams:
## Changes

- Add `SeqlockCache<TKey, TValue>` — a high-performance 2-way skew-associative cache using seqlock pattern for lock-free reads
- Replace `ConcurrentDictionary` with `SeqlockCache` for `StorageCache` and `StateCache` in `PreBlockCaches`
- Add `IHash64bit<T>` interface for 64-bit hashing support
- Add `FastHash64For20Bytes` optimized hash function for Address (20-byte) data

### Performance

SeqlockCache provides 2.6–6.3x faster operations compared to ConcurrentDictionary for cache hit/set scenarios, and up to 46x faster for GetOrAdd misses due to elimination of lock contention overhead.

| Operation         | Speedup     |
|-------------------|-------------|
| TryGetValue (Hit) | 2.6x faster |
| Set (Existing)    | 6.3x faster |
| GetOrAdd (Hit)    | 3.2x faster |
| GetOrAdd (Miss)   | 46x faster  |

Zero allocations on hit paths. Code size 5–17x smaller than ConcurrentDictionary equivalents.

### Cache Design

- **2-way skew-associative**: each way uses independent hash bits for set indexing (bits 0-13 vs bits 42-55 of the 64-bit hash), breaking collision correlation between ways
- **Lock-free reads** via seqlock pattern — readers never take locks, zero stores on the read path
- **Best-effort writes** — writers skip on contention (suitable for cache semantics)
- **O(1) Clear()** via global epoch — no per-entry zeroing required
- **Software prefetch** — prefetches way 1's cache line while checking way 0, hiding L2/L3 latency in the split memory layout
- **Hash-bit alternating eviction** — when both ways are live, uses a hash bit to spread evictions across ways (zero extra memory)

### Hit Rate (32K entries, uniform random keys)

| Keys  | 1-Way Direct | 2-Way Skew |
|------:|:-------------|:-----------|
| 5K    | ~93%         | 99.7%      |
| 10K   | ~86%         | 98.2%      |
| 15K   | ~80%         | 95.1%      |
| 20K   | ~75%         | 91.0%      |

At 10K keys (typical workload): 1.8% miss rate vs 14.1% — **7.8x fewer evictions**.

### Header Layout (64-bit)

[Lock:1][Epoch:26][Hash:28][Seq:8][Occ:1]

Hash bit partitioning:
- Bits 0-13: way 0 set index
- Bits 14-41: hash signature (28-bit tag)
- Bits 42-55: way 1 set index (independent)

## Types of changes

- [x] Optimization

## Testing

- [x] Requires testing

If yes, did you write tests?

- [x] Tests written (19 tests covering correctness, concurrency, Clear, GetOrAdd)

### DIFF
--- a/src/Nethermind/Nethermind.Core/Collections/SeqlockCache.cs
+++ b/src/Nethermind/Nethermind.Core/Collections/SeqlockCache.cs
@@ -185,7 +185,18 @@ public sealed class SeqlockCache<TKey, TValue>
             return value;
         }
 
-        value = valueFactory(in key);
+        return GetOrAddMiss(in key, valueFactory, idx0, idx1, hashPart);
+    }
+
+    /// <summary>
+    /// Cold path for GetOrAdd: invokes factory and stores the result.
+    /// Kept out-of-line so the hot path (cache hit) compiles to a lean method body
+    /// with minimal register saves and stack frame.
+    /// </summary>
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    private TValue? GetOrAddMiss(in TKey key, ValueFactory valueFactory, int idx0, int idx1, long hashPart)
+    {
+        TValue? value = valueFactory(in key);
         SetCore(in key, value, idx0, idx1, hashPart);
         return value;
     }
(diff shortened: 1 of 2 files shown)


LukaszRozmej (approved):

src/Nethermind/Nethermind.Core/Collections/IHash64bit.cs
```diff
@@ -0,0 +1,22 @@
...
+/// The 64-bit hash should have good distribution across all bits.
+/// </remarks>
+public interface IHash64bit<TKey>
+{
+    /// <summary>
+    /// Returns a 64-bit hash code for the current instance.
+    /// </summary>
+    /// <returns>A 64-bit hash code with good distribution across all bits.</returns>
+    long GetHashCode64();
+    bool Equals(in TKey other);
```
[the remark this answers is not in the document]
LukaszRozmej: It is kind of right?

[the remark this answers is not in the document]
benaadams: I want the `in` param on Equals

src/Nethermind/Nethermind.State/PreBlockCaches.cs
```diff
@@ -19,23 +19,23 @@ public class PreBlockCaches
 
     private readonly Func<CacheType>[] _clearCaches;
 
-    private readonly ConcurrentDictionary<StorageCell, byte[]> _storageCache = new(LockPartitions, InitialCapacity);
-    private readonly ConcurrentDictionary<AddressAsKey, Account> _stateCache = new(LockPartitions, InitialCapacity);
+    private readonly SeqlockCache<StorageCell, byte[]> _storageCache = new();
+    private readonly SeqlockCache<AddressAsKey, Account> _stateCache = new();
     private readonly ConcurrentDictionary<NodeKey, byte[]?> _rlpCache = new(LockPartitions, InitialCapacity);
     private readonly ConcurrentDictionary<PrecompileCacheKey, Result<byte[]>> _precompileCache = new(LockPartitions, InitialCapacity);
```
LukaszRozmej: why not those?

benaadams: Precompile is of indeterminate sizes and the SeqLock doesn't actually clear the values so stuff could be hanging around for ages

_rlpCache is a bit weird now

benaadams:
> Does it allow hash collisions? Could you add tests for that including ones checking concurrency if so?

Overwrites if collides; are tests in SeqlockCacheTests.cs

src/Nethermind/Nethermind.State/PreBlockCaches.cs
benaadams: > why not those?

Added Rlp. Precompiles can be pretty big hough (input and output)

benaadams:
> Does it give any meaningful MGas measured improvement?

Mild improvement
image

PR #10416, opened 2026-02-05
fix(xdc): prevent InvalidOperationException in XdcPool.EndRound

andrewshab3:
## Changes

The original code iterated over `_items.Keys` while calling `_items.Remove()` inside the loop.This causes `InvalidOperationException: Collection was modified; enumeration operation may not execute` at runtime when any item matches the removal condition.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_


andrewshab3:
@LukaszRozmej @ak88 sorry, can you check please?

PR #10423, opened 2026-02-06
Use CodeInfo type instead of ICodeInfo

benaadams:
## Changes

- Devirtualize CodeInfo access
- Replace the ICodeInfo abstraction with a concrete CodeInfo type and adapt related APIs and implementations. CodeInfo was extended to carry precompile info, provide Code/CodeSpan, Version, IsPrecompile/IsEmpty semantics and background analysis. EofCodeInfo now derives from CodeInfo and provides EOF-specific data and versioning. PrecompileInfo was removed and precompiles are represented as CodeInfo (wrapping CachedPrecompile when needed).
- API changes: ICodeInfoRepository, IOverridableCodeInfoRepository and IPrecompileProvider signatures and caches now use CodeInfo; CodeInfoFactory and CodeLruCache updated accordingly. Call sites across the VM, instruction implementations, tracers, transaction processing, repositories and tests were updated to use CodeInfo directly and to perform EOF checks using 'is EofCodeInfo' where appropriate. JumpDestinationAnalyzer.MachineCode was made accessible for CodeInfo.Code.
- Overall this unifies EOF and precompile handling under a single CodeInfo model and simplifies caching and execution logic.

- [x] Optimization

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Evm/CodeAnalysis/CodeInfo.cs
+++ b/src/Nethermind/Nethermind.Evm/CodeAnalysis/CodeInfo.cs
@@ -11,27 +11,35 @@ namespace Nethermind.Evm.CodeAnalysis;
 public class CodeInfo : IThreadPoolWorkItem, IEquatable<CodeInfo>
 {
     public static CodeInfo Empty { get; } = new();
-    private static readonly JumpDestinationAnalyzer _emptyAnalyzer = new(Empty, skipAnalysis: true);
+    // Empty code sentinel
+    private static readonly JumpDestinationAnalyzer? _emptyAnalyzer = new(Empty, skipAnalysis: true);
 
-    private CodeInfo() { }
+    // Empty
+    private CodeInfo()
+    {
+        _analyzer = null;
+    }
 
+    // Eof
     protected CodeInfo(int version, ReadOnlyMemory<byte> code)
     {
         Version = version;
         Code = code;
-        _analyzer = code.Length == 0 ? _emptyAnalyzer : new JumpDestinationAnalyzer(this, skipAnalysis: true);
+        _analyzer = null;
     }
 
+    // Regular contract
     public CodeInfo(ReadOnlyMemory<byte> code)
     {
         Code = code;
         _analyzer = code.Length == 0 ? _emptyAnalyzer : new JumpDestinationAnalyzer(this);
     }
 
+    // Precompile
     public CodeInfo(IPrecompile? precompile)
     {
-        _analyzer = _emptyAnalyzer;
         Precompile = precompile;
+        _analyzer = null;
     }
 
     public ReadOnlyMemory<byte> Code { get; }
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Evm/CodeAnalysis/CodeInfo.cs
```diff
@@ -3,35 +3,85 @@
...
 
-    private readonly JumpDestinationAnalyzer _analyzer = code.Length == 0 ? _emptyAnalyzer : new JumpDestinationAnalyzer(code);
+    private CodeInfo() { }
 
-    public bool IsEmpty => ReferenceEquals(_analyzer, _emptyAnalyzer);
+    protected CodeInfo(int version, ReadOnlyMemory<byte> code)
+    {
+        Version = version;
+        Code = code;
+        _analyzer = code.Length == 0 ? _emptyAnalyzer : new JumpDestinationAnalyzer(this, skipAnalysis: true);
```
LukaszRozmej: Does the `skipanalysis` have any merit here? Can't it just be:

```suggestion
        _analyzer = _emptyAnalyzer;
```

src/Nethermind/Nethermind.Evm/CodeAnalysis/CodeInfo.cs
```diff
@@ -3,35 +3,85 @@
...
+        _analyzer = _emptyAnalyzer;
+        Precompile = precompile;
     }
 
+    public ReadOnlyMemory<byte> Code { get; }
+    public ReadOnlySpan<byte> CodeSpan => Code.Span;
+
+    public IPrecompile? Precompile { get; }
+
+    private readonly JumpDestinationAnalyzer _analyzer;
```
LukaszRozmej, line 50: If previous is true maybe:

```suggestion
    private readonly JumpDestinationAnalyzer _analyzer = _emptyAnalyzer;
```

src/Nethermind/Nethermind.Evm/CodeAnalysis/CodeInfo.cs
```diff
@@ -3,35 +3,85 @@
...
+        Code = code;
+        _analyzer = code.Length == 0 ? _emptyAnalyzer : new JumpDestinationAnalyzer(this, skipAnalysis: true);
+    }
 
-    public bool ValidateJump(int destination)
+    public CodeInfo(ReadOnlyMemory<byte> code)
     {
-        return _analyzer.ValidateJump(destination);
+        Code = code;
+        _analyzer = code.Length == 0 ? _emptyAnalyzer : new JumpDestinationAnalyzer(this);
```
LukaszRozmej, line 35: And have only here:

```suggestion
        if (code.Length != 0) _analyzer = new JumpDestinationAnalyzer(this);
```

src/Nethermind/Nethermind.Evm/CodeAnalysis/CodeInfo.cs
```diff
@@ -3,35 +3,85 @@
 
 using System;
 using System.Threading;
+using Nethermind.Core.Extensions;
+using Nethermind.Evm.Precompiles;
 
 namespace Nethermind.Evm.CodeAnalysis;
 
-public sealed class CodeInfo(ReadOnlyMemory<byte> code) : ICodeInfo, IThreadPoolWorkItem
+public class CodeInfo : IThreadPoolWorkItem, IEquatable<CodeInfo>
```
flcl42, line 11: Can be sealed?

Demuirgos (approved):

src/Nethermind/Nethermind.Evm/CodeAnalysis/CodeInfo.cs
LukaszRozmej: nope -> EofCodeInfo inherits

src/Nethermind/Nethermind.Evm/CodeAnalysis/CodeInfo.cs
benaadams: Ok, used null as non analyse state as that still works with the IsEmpty test

PR #10431, opened 2026-02-06
Parallelizable does not work on all XDC tests

ak88:
Parallelizable will actually make the tests run slower and break some tests, so i removed some of the added tags.


src/Nethermind/Nethermind.Xdc.Test/ModuleTests/ProposedBlockTests.cs
```diff
@@ -13,7 +13,6 @@
 
 namespace Nethermind.Xdc.Test;
 
-[Parallelizable(ParallelScope.All)]
```
alexb5dh, line 16: Even `ParallelScope.Self` won't work?

PR #10446, opened 2026-02-07
Optimize receipt canonicality with OnUpdateMainChain batching

benaadams:
## Changes

  - Use `OnUpdateMainChain` event for batched tx-index writes instead of per-block `EnsureCanonical` calls, reducing write batch overhead during  multi-block chain updates
  - Fire `OnUpdateMainChain` before the per-block `MoveToMain` loop so tx indices are committed before `NewCanonicalReceipts` notifies JSON-RPC  subscribers
  - Extract `ShouldIndexBlock` and `WriteCanonicalTxIndex` helpers to share logic between single-block and batch paths
  - Add `IDisposable` to `PersistentReceiptStorage` to unsubscribe both event handlers
  - Use `PutSpan` instead of `BytesToArray()` for non-compact tx index writes to avoid allocation
  - Add early return in `RemoveBlockTx` when block has no transactions
  - Simplify `BlockTree.UpdateMainChain` ascending-order check and DEBUG validation loop

- [x] Optimization
- [x] Refactoring

#### Requires testing

- [x] Yes

- [x] Yes


src/Nethermind/Nethermind.Core/BlockReplacementEventArgs.cs
```diff
@@ -3,8 +3,37 @@
...
-        public Block? PreviousBlock { get; } = previousBlock;
+        public BlockReplacementEventArgs(Block block)
+            : this(block, null)
+        {
+        }
+
+        public BlockReplacementEventArgs(Block block, Block? previousBlock)
+            : this(block, previousBlock, isPartOfMainChainUpdate: false, isLastInMainChainUpdate: false, mainChainUpdateId: 0)
+        {
+        }
```
LukaszRozmej:
```suggestion
        public BlockReplacementEventArgs(Block block, Block? previousBlock = null)
            : this(block, previousBlock, isPartOfMainChainUpdate: false, isLastInMainChainUpdate: false, mainChainUpdateId: 0)
        {
        }
```

src/Nethermind/Nethermind.Blockchain/Receipts/PersistentReceiptStorage.cs
```diff
@@ -360,16 +381,80 @@ private void RemoveBlockTx(Block block)
...
+                if (_pendingBatchId < 0)
+                {
+                    _pendingBatchId = batchId;
+                    _pendingBatchLastBlockNumber = _blockTree.FindBestSuggestedHeader()?.Number ?? 0;
+                    _pendingBatchBlockCount = 0;
+                }
+
+                if (ShouldIndexBlock(block, _pendingBatchLastBlockNumber, out Transaction[] transactions))
+                {
+                    _pendingBatch ??= _transactionDb.StartWriteBatch();
```
LukaszRozmej: Do we announce block before we do that? If yes I am a bit worried about subscriptions or json rpc. Maybe we should serve receipts from memory until we persist it? (Then persisting could be async?)

src/Nethermind/Nethermind.Core/BlockReplacementEventArgs.cs
benaadams: Done

src/Nethermind/Nethermind.Blockchain/Receipts/PersistentReceiptStorage.cs
benaadams: Added overlay and test

src/Nethermind/Nethermind.Blockchain/Receipts/PersistentReceiptStorage.cs
```diff
@@ -111,7 +113,13 @@ private void BlockTreeOnBlockAddedToMain(object? sender, BlockReplacementEventAr
 
         public Hash256 FindBlockHash(Hash256 txHash)
         {
-            var blockHashData = _transactionDb.Get(txHash);
+            // Check in-memory overlay first for entries pending batch commit.
+            if (_pendingTxIndex.IsEmpty
```
LukaszRozmej: takes all locks?

src/Nethermind/Nethermind.Blockchain/Receipts/PersistentReceiptStorage.cs
```diff
@@ -483,13 +494,13 @@ private void DiscardPendingBatch()
...
             try { _pendingBatch?.Clear(); _pendingBatch?.Dispose(); }
-            finally { _pendingBatch = null; _pendingBatchId = -1; _pendingBatchBlockCount = 0; }
+            finally { _pendingBatch = null; _pendingBatchId = -1; _pendingBatchBlockCount = 0; _pendingTxIndex.Clear(); }
         }
 
         private void FlushPendingBatchWrites()
         {
             try { _pendingBatch?.Dispose(); }
-            finally { _pendingBatch = null; _pendingBatchBlockCount = 0; }
+            finally { _pendingBatch = null; _pendingBatchBlockCount = 0; _pendingTxIndex.Clear(); }
```
LukaszRozmej: maybe do flushing in background thread?

src/Nethermind/Nethermind.Blockchain/Receipts/PersistentReceiptStorage.cs
benaadams: Good catch — `ConcurrentDictionary.IsEmpty` acquires all bucket locks for a consistent snapshot. Removed the `IsEmpty` short-circuit; `TryGetValue` is lock-free and handles the empty case equally fast (hashes, finds empty bucket, returns false).

src/Nethermind/Nethermind.Blockchain/Receipts/PersistentReceiptStorage.cs
benaadams: The overlay already ensures reads are served correctly during the batch window. Background flushing would require keeping the overlay alive until the async write completes (can't clear \ until RocksDB has the data), plus coordination with the next batch arriving. The added complexity doesn't seem worth it for a write that's already batched (single RocksDB WriteBatch instead of N individual writes). Could revisit if profiling shows the flush as a bottleneck.

src/Nethermind/Nethermind.Blockchain/BlockTree.cs
asdacap: Can just replace the event to use OnUpdateMainChain?

src/Nethermind/Nethermind.Blockchain/Receipts/PersistentReceiptStorage.cs
```diff
@@ -114,8 +114,8 @@ private void BlockTreeOnBlockAddedToMain(object? sender, BlockReplacementEventAr
         public Hash256 FindBlockHash(Hash256 txHash)
         {
             // Check in-memory overlay first for entries pending batch commit.
-            if (_pendingTxIndex.IsEmpty
-                || !_pendingTxIndex.TryGetValue(txHash, out byte[]? blockHashData))
+            // TryGetValue is lock-free; avoid IsEmpty which acquires all bucket locks.
```
LukaszRozmej: pointless comment :P
```suggestion
```

src/Nethermind/Nethermind.Blockchain/BlockTree.cs
benaadams: Done

src/Nethermind/Nethermind.Blockchain/Receipts/PersistentReceiptStorage.cs
benaadams: Changed notification order so is safe (as per @asdacap suggstion) and overlay not needed

src/Nethermind/Nethermind.Blockchain/Receipts/PersistentReceiptStorage.cs
```diff
@@ -69,33 +68,57 @@ public PersistentReceiptStorage(
...
+                return;
+            }
+
+            using IWriteBatch writeBatch = _transactionDb.StartWriteBatch();
+            foreach (Block block in blocks)
+            {
+                if (ShouldIndexBlock(block, lastBlockNumber, out Transaction[] transactions))
+                {
+                    WriteCanonicalTxIndex(block, transactions, writeBatch);
+                }
```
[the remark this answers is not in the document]
benaadams: OMG will you stop suggesting Linq

LukaszRozmej (approved):

stdevMac:
@benaadams stale-PR check: #11894 removed the OnUpdateMainChain API this hangs off, so it needs a ground-up rework. OK to close until/unless that happens? If there's no reply in the next few days, we'll close this as part of the stale-PR cleanup — it can always be reopened or resubmitted.

PR #10460, opened 2026-02-08
AGENTS shouldn't be repetitive when adding code

benaadams:
## Changes

- When fixing a bug, always add a regression test that fails without the fix and passes with it
- Keep changes minimal and focused: do not rename variables, reformat surrounding code, or refactor unrelated logic as part of a fix. Touch only what is necessary to solve the problem.
- Follow DRY: after making changes, review the result for duplicated logic. Extract repeated blocks (roughly 5+ lines) into shared methods, but do not over-extract trivial one-liners into their own methods.
- In generic types, move methods that do not depend on the type parameter to a non-generic base class or static helper to avoid redundant JIT instantiations per closed type.

- [x] Other: Agents

#### Requires testing

- [x] No

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -28,6 +28,10 @@ This guide helps to get started with the Nethermind Ethereum execution client re
 - Add tests to existing test files rather than creating new ones
 - Code comments must explain _why_, not _what_
 - **NEVER suggest using LINQ (`.Select()`, `.Where()`, `.Any()`, etc.) when a simple `foreach` or `for` loop would work.** LINQ has overhead and is less readable for simple iterations. Use LINQ only for complex queries where the declarative syntax significantly improves clarity.
+- When fixing a bug, always add a regression test that fails without the fix and passes with it
+- Keep changes minimal and focused: do not rename variables, reformat surrounding code, or refactor unrelated logic as part of a fix. Touch only what is necessary to solve the problem.
+- Follow DRY: after making changes, review the result for duplicated logic. Extract repeated blocks (roughly 5+ lines) into shared methods, but do not over-extract trivial one-liners into their own methods.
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -28,6 +28,10 @@ This guide helps to get started with the Nethermind Ethereum execution client re
 - Add tests to existing test files rather than creating new ones
 - Code comments must explain _why_, not _what_
 - **NEVER suggest using LINQ (`.Select()`, `.Where()`, `.Any()`, etc.) when a simple `foreach` or `for` loop would work.** LINQ has overhead and is less readable for simple iterations. Use LINQ only for complex queries where the declarative syntax significantly improves clarity.
+- When fixing a bug, always add a regression test that fails without the fix and passes with it
```
LukaszRozmej, line 31:
```suggestion
- When fixing a bug, always add a regression test that fails without the fix and passes with it
- When adding a test, check if previous tests can be reused with new test case
```

AGENTS.md
```diff
@@ -28,6 +28,10 @@ This guide helps to get started with the Nethermind Ethereum execution client re
...
+- Keep changes minimal and focused: do not rename variables, reformat surrounding code, or refactor unrelated logic as part of a fix. Touch only what is necessary to solve the problem.
+- Follow DRY: after making changes, review the result for duplicated logic. Extract repeated blocks (roughly 5+ lines) into shared methods, but do not over-extract trivial one-liners into their own methods.
+- In generic types, move methods that do not depend on the type parameter to a non-generic base class or static helper to avoid redundant JIT instantiations per closed type.
```
LukaszRozmej, line 34:
```suggestion
- In generic types, move methods that do not depend on the type parameter to a non-generic base class or static helper to avoid redundant JIT instantiations per closed type.
- Avoid `var` when declaring variables, the only acceptable exceptions are very long type names e.g. nested generic types
```

LukaszRozmej:
@copilot create new PR from my comments

PR #10462, opened 2026-02-08
fix: add missing yield break after shutdown response in `JsonRpcProcessor`

VolodymyrBg:
## Changes

- `ProcessAsync` https://github.com/NethermindEth/nethermind/blob/cae3a5445525910cde27e332bbd33618e2d0c68a/src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs#L129-L236 was missing a `yield break` after yielding the "Shutting down" error response. This caused the method to continue reading from PipeReader, parsing JSON, and dispatching RPC handlers even though the node is shutting down.

On WebSocket/IPC this resulted in the client receiving both the shutdown error AND the actual request response (double response). On HTTP it caused unnecessary work during shutdown (mitigated by the caller's `break`, but the processing still ran).

Added `yield break` so `ProcessAsync` stops immediately after the shutdown response.

#### Requires testing

- [x] Yes

- [x] Yes

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -23,12 +23,15 @@ This guide helps to get started with the Nethermind Ethereum execution client re
 - Use the `ArgumentNullException.ThrowIfNull` method for null checks and other similar methods
 - Use the `ObjectDisposedException.ThrowIf` method for disposal checks
 - Use documentation comments for all public APIs with proper structure
+- Avoid `var` when declaring variables, the only acceptable exceptions are very long type names e.g. nested generic types
 - Consider performance implications in high-throughput paths
 - Trust null annotations, do not add redundant null checks
+- When fixing a bug, always add a regression test that fails without the fix and passes with it
 - Add tests to existing test files rather than creating new ones
+- When adding multiple, similar tests write one test with test cases
+- When adding a test, check if previous tests can be reused with new test case
 - Code comments must explain _why_, not _what_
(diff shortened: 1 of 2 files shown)


LukaszRozmej (approved):

src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
```diff
@@ -132,6 +132,7 @@ public async IAsyncEnumerable<JsonRpcResult> ProcessAsync(PipeReader reader, Jso
         {
             JsonRpcErrorResponse response = _jsonRpcService.GetErrorResponse(ErrorCodes.ResourceUnavailable, "Shutting down");
             yield return JsonRpcResult.Single(RecordResponse(response, new RpcReport("Shutdown", 0, false)));
+            yield break;
         }
```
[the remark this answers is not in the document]
LukaszRozmej: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/NethermindEth/nethermind/pull/10462#discussion_r2780392032)

VolodymyrBg:
> need to handle finally correctly

So I handle this in this PR or Copilot?

src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
LukaszRozmej: Basically this try should span whole method

LukaszRozmej (commented):
> > need to handle finally correctly
> 
> So I handle this in this PR or Copilot?

I think copilot won't handle external PRs

src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs
VolodymyrBg: > Basically this try should span whole method

Done

PR #10472, opened 2026-02-09
fix: clear reference-type elements in Truncate to prevent pool memory leak

phrwlk:
## Changes

`ArrayPoolListCore<T>.Truncate` reduced count without clearing elements in the `[newLength..oldCount)` range. When the backing array was later returned to the pool via `Dispose`, `ClearToCount` only cleared up to the already-reduced count, leaving stale references in the pooled array. This prevented GC from collecting those objects until the array slots were overwritten.

Added `Array.Clear` for the truncated tail when T is or contains references, consistent with how `ReduceCount`, `RemoveAt`, and `Clear` already handle this in the same file. Fixes the leak for both `ArrayPoolList<T>` and `ArrayPoolListRef<T>`.

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -790,6 +790,7 @@
     "unreferred",
     "unrequested",
     "unresolve",
+    "unshifted",
     "unsub",
     "unsubscription",
     "unsynchronized",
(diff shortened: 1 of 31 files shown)


LukaszRozmej:
Same code is in `ReduceCount`, can you make it a reusable helper method?

LukaszRozmej (approved):

PR #10483, opened 2026-02-11
Remove fallback key lookups in NodeStorage

benaadams:
## Changes

- Remove cross-scheme fallback lookups in `NodeStorage.Get` and `NodeStorage.KeyExists`. Previously, if a lookup under the active scheme missed, a   second lookup was attempted using the other scheme (HalfPath → Hash or Hash → HalfPath). Now each method queries only the scheme-specific key,   eliminating redundant DB reads on every node access.

  ## Breaking Change

  **Nodes that were written under one key scheme and are being read under another will no longer be found.** Any database that contains trie nodes  stored under the old (Hash) scheme **must be fully migrated to HalfPath before upgrading**. Databases that have already completed a full pruning  cycle on a HalfPath build are unaffected.

  ### Who is affected

  - Operators running databases that were originally synced with the Hash key scheme and have **not** completed a full pruning pass on a HalfPath-enabled build.
  - Operators who downgrade from HalfPath back to Hash after writing new nodes.

  ### Release notes entry

  > `NodeStorage` no longer falls back to the alternate key scheme on read. Databases must be fully migrated to the configured key scheme (default: HalfPath) before upgrading. Run a full pruning cycle or resync on the previous release to complete migration.If you have changes key scheme and not resynced or fully pruned since.

  ## Types of changes

  - [x] Optimization

  ## Testing

  #### Requires testing

  - [x] Yes

  - [x] Yes

  #### Notes on testing

  - Updated `TreeStoreTests` and `CopyTreeVisitorTests` to pass the fixture's key scheme to standalone `NodeStorage` instances used for  verification.
  - Removed `When_EntryOfDifferentScheme_Should_StillBeAbleToRead` and simplified `When_KeyNotExist_Should_TryBothKeyType` since the cross-scheme  fallback no longer exists.
  - Fixed a pre-existing bug in `CopyTreeVisitorTests` where the Hash fixture's read-flags assertion expected `HintReadAhead` instead of  `HintCacheMiss`.
  - Added `scheme` parameter to `TestWorldStateFactory.CreateForTestWithStateReader`.

  ## Documentation

  - [x] Yes

  - [x] Yes

  > `NodeStorage` no longer falls back to the alternate key scheme on read. Databases must be fully migrated to the configured key scheme (default:   HalfPath) before upgrading. Run a full pruning cycle or resync on the previous release to complete migration.If you have changes key scheme and not resynced or fully pruned since.

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain.Test/FullPruning/CopyTreeVisitorTests.cs
+++ b/src/Nethermind/Nethermind.Blockchain.Test/FullPruning/CopyTreeVisitorTests.cs
@@ -87,7 +87,7 @@ public class CopyTreeVisitorTests(INodeStorage.KeyScheme scheme)
         dbProvider.CodeDb.Returns(new MemDb());
 
         // Use TestWorldStateFactory.CreateForTest() with the custom DbProvider
-        (IWorldState worldState, IStateReader stateReader) = TestWorldStateFactory.CreateForTestWithStateReader(dbProvider, logManager, scheme);
+        (_, IStateReader stateReader) = TestWorldStateFactory.CreateForTestWithStateReader(dbProvider, logManager, scheme);
 
         BlockHeader? baseBlock = Build.A.BlockHeader.WithStateRoot(trie.RootHash).TestObject;
         if (scheme == INodeStorage.KeyScheme.Hash)


asdacap:
Need to check if this break node during online conversion from Hash to Halfpath.

PR #10495, opened 2026-02-12
Replace PeerManager blocking waits with async equivalents

benaadams:
## Changes

- Replace `ManualResetEventSlim` with `SemaphoreSlim(0, 1)` for async-compatible signaling in the peer update loop
- Convert `EnsureAvailableActivePeerSlot` to async (`EnsureAvailableActivePeerSlotAsync`), replacing blocking
`Wait(100ms)` polling with `WaitAsync`
- Replace `Task.Factory.StartNew(..., LongRunning)` with `async Task` + `Task.Yield()` — `LongRunning` is meaningless for
async delegates
- `StopAsync` now awaits the loop task for clean shutdown instead of `await Task.CompletedTask`
- Route all signal sites through `SignalPeerUpdate()` helper guarded by lock + `CurrentCount` check
- Remove `#region` directives

- [x] Optimization
- [x] Refactoring

#### Requires testing

- [x] Yes

- [x] Yes

#### Notes on testing

Added 3 regression tests in `PeerManagerTests`:
- `Stop_does_not_hang` (parameterized) — stop completes within 5s while idle or with saturated slots
- `Disconnect_triggers_refill_without_blocking` — disconnect signal triggers new connections via async path
- `No_slot_available_before_deadline_does_not_deadlock_refill` — slot exhaustion deadline expiry doesn't prevent
subsequent refill

### DIFF
--- a/src/Nethermind/Nethermind.Network/PeerManager.cs
+++ b/src/Nethermind/Nethermind.Network/PeerManager.cs
@@ -59,6 +59,7 @@ namespace Nethermind.Network
         private int _newActiveNodes;
         private int _failedInitialConnect;
         private int _connectionRounds;
+        private bool _hasPeerSlot;
 
         private Timer? _peerUpdateTimer;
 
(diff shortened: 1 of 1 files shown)


benaadams:
Superceaded by https://github.com/NethermindEth/nethermind/pull/10753

PR #10500, opened 2026-02-12
Optimize Kestrel socket transport for lower-latency JSON-RPC dispatch

benaadams:
## Changes

- Configure `SocketTransportOptions` on the JSON-RPC Kestrel server to reduce thread hops and allocation overhead:
- `IOQueueCount = 0` — schedule I/O directly on the ThreadPool instead of dedicated queues
- `WaitForDataBeforeAllocatingBuffer = false` — pre-allocate read buffers for lower first-byte latency
- `UnsafePreferInlineScheduling = true` — run application continuations on the I/O thread
- Set `DOTNET_SYSTEM_NET_SOCKETS_INLINE_COMPLETIONS=1` at process startup so the runtime inlines socket completions on
the epoll/IOCP thread instead of queuing to the ThreadPool

- [x] Optimization

#### Requires testing

- [x] Yes

- [x] No

#### Notes on testing

Needs profiling under realistic JSON-RPC load (Engine API + public RPC) to measure latency/throughput improvement and
verify no regressions. `UnsafePreferInlineScheduling` can degrade performance if any handler does expensive synchronous
work — the Engine API fast lane and standard RPC paths are async I/O–bound so should benefit, but this should be
validated under load.

`DOTNET_SYSTEM_NET_SOCKETS_INLINE_COMPLETIONS` is process-wide and also affects p2p networking (DotNetty). Verify no
adverse effects on devp2p sync performance.

### DIFF
--- a/src/Nethermind/Nethermind.Runner/Program.cs
+++ b/src/Nethermind/Nethermind.Runner/Program.cs
@@ -42,6 +42,10 @@ using DotNettyLoggerFactory = DotNetty.Common.Internal.Logging.InternalLoggerFac
 using DotNettyLeakDetector = DotNetty.Common.ResourceLeakDetector;
 #endif
 
+// Inline socket I/O completions to reduce thread-pool hops on the networking hot path.
+// Must be set before any socket usage so SocketAsyncEngine picks it up during static init.
+Environment.SetEnvironmentVariable("DOTNET_SYSTEM_NET_SOCKETS_INLINE_COMPLETIONS", "1");
+
 DataFeed.StartTime = Environment.TickCount64;
 Console.Title = ProductInfo.Name;
 // Increase regex cache size as more added in log coloring matches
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Runner/Program.cs
```diff
@@ -42,6 +42,10 @@
 using DotNettyLeakDetector = DotNetty.Common.ResourceLeakDetector;
 #endif
 
+// Inline socket I/O completions to reduce thread-pool hops on the networking hot path.
+// Must be set before any socket usage so SocketAsyncEngine picks it up during static init.
+Environment.SetEnvironmentVariable("DOTNET_SYSTEM_NET_SOCKETS_INLINE_COMPLETIONS", "1");
```
LukaszRozmej, line 47: I don't think this will apply to current already started app

benaadams: Wasn't sure it would either; but with this you can follow the trace from socket data received event to runningEvm now (percentages also include P2P etc)
 
image

stdevMac:
@benaadams stale-PR check: marked DO NOT MERGE pending profiling that never happened, ~5 months silent. OK to close? If there's no reply in the next few days, we'll close this as part of the stale-PR cleanup — it can always be reopened or resubmitted.

PR #10503, opened 2026-02-12
RLP encode decode msg code

ak88:
RLPx message code are actually RLP encoded ints 
https://github.com/ethereum/devp2p/blob/master/rlpx.md#message-encoding
but we treat them as a single unencoded byte. 

This changes encoding and decoding of message codes to act according to spec.

### DIFF
--- a/src/Nethermind/Nethermind.Network/Rlpx/ZeroFrameMerger.cs
+++ b/src/Nethermind/Nethermind.Network/Rlpx/ZeroFrameMerger.cs
@@ -75,8 +75,6 @@ namespace Nethermind.Network.Rlpx
         {
             Rlp.ValueDecoderContext rlpContext = new(input.AsSpan());
             ulong rlpPacketType = rlpContext.DecodeULong();
-            if (rlpPacketType > byte.MaxValue)
-                throw new NotSupportedException($"Received unsupported packet code={rlpPacketType}. We currently only support values 0-255.");
             int read = rlpContext.Position;
             input.SkipBytes(read);
             IByteBuffer content;


src/Nethermind/Nethermind.Network/MessageSerializationService.cs
```diff
@@ -68,7 +68,10 @@ void WriteAdaptivePacketType(in IByteBuffer buffer)
             {
                 if (message is P2PMessage p2PMessage)
                 {
-                    buffer.WriteByte(p2PMessage.AdaptivePacketType);
+                    if (p2PMessage.AdaptivePacketType <= Rlp.EmptyArrayByte)
+                        buffer.WriteByte(p2PMessage.AdaptivePacketType);
+                    else
+                        buffer.WriteBytes(Rlp.Encode((long)p2PMessage.AdaptivePacketType).Bytes);
```
LukaszRozmej: this allocates a small array would be good to avoid.

src/Nethermind/Nethermind.Network/Rlpx/ZeroFrameMerger.cs
```diff
@@ -72,20 +73,23 @@ private void ReadChunk(IByteBuffer input, in FrameHeaderReader.FrameInfo frame)
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private void ReadFirstChunk(IChannelHandlerContext context, IByteBuffer input, in FrameHeaderReader.FrameInfo frame)
         {
-            byte packetTypeRlp = input.ReadByte();
+            Rlp.ValueDecoderContext rlpContext = new(input.AsSpan());
+            byte rlpPacketType = (byte)rlpContext.DecodeULong();
```
flcl42: It hides any packet with value > 255. Better to inform we can't handle it

flcl42 (commented):
Imho it's a hot path, and it's real to avoid ValueDecoderContext/Rlp usage.

src/Nethermind/Nethermind.Network/MessageSerializationService.cs
ak88: Its not really possible without refactoring the Rlp class, and it only applies for protocols who uses message codes above 128, which XDC is the only one so far.

src/Nethermind/Nethermind.Network/Rlpx/ZeroFrameMerger.cs
ak88: good point

src/Nethermind/Nethermind.Network/MessageSerializationService.cs
flcl42: You can add similar check in the other case, so regular path will be to just read byte, otherwise - use ValueDecoderContext

src/Nethermind/Nethermind.Network/MessageSerializationService.cs
```diff
@@ -68,7 +68,7 @@ void WriteAdaptivePacketType(in IByteBuffer buffer)
             {
                 if (message is P2PMessage p2PMessage)
                 {
-                    buffer.WriteByte(p2PMessage.AdaptivePacketType);
+                    buffer.WriteBytes(Rlp.Encode((long)p2PMessage.AdaptivePacketType).Bytes);
```
LukaszRozmej: Try avoiding allocation, maybe let's add a version where you can encode to a Span or directly to buffer? Or at least pool the array? 

Or a ValueDecoderContext with stackalloc buffer?

asdacap (approved):

src/Nethermind/Nethermind.Network/MessageSerializationService.cs
alexb5dh: https://github.com/NethermindEth/nethermind/pull/10610

PR #10505, opened 2026-02-12
fix: warn and retry when external IP resolves to 0.0.0.0

### LINKED ISSUE #10281: Output warning when external IP cannot be resolved
**Description**
When external IP cannot be resolved it should output a warning

**Actual behavior**
No warning and uses `0.0.0.0` which means no peers will accept connection

smartprogrammer93:
Fixes #10281

## Changes

- Added warning log when external IP cannot be resolved (resolves to 0.0.0.0 or None)
- Implemented non-blocking background retry mechanism (up to 5 attempts, every 60 seconds)
- Log error after all retries fail to inform users that node won't be reachable by peers
- Advise user to restart the node once the ip is resolved for it to take effect.

Wondering if we should restart the node once the externalIp is resolved or not

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [ ] No

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

Improved external IP resolution with automatic retries and better error messaging when the node's external IP cannot be determined.

### DIFF
--- a/src/Nethermind/Nethermind.Network/IPResolver.cs
+++ b/src/Nethermind/Nethermind.Network/IPResolver.cs
@@ -36,47 +36,33 @@ public class IPResolver : IIPResolver
             LocalIp = IPAddress.Loopback;
         }
 
-        try
-        {
-            ExternalIp = await InitializeExternalIp();
-        }
-        catch (Exception)
-        {
-            ExternalIp = IPAddress.None;
-        }
+        const int maxAttempts = 5;
+        const int delaySeconds = 2;
 
-        if (Equals(ExternalIp, IPAddress.Any) || Equals(ExternalIp, IPAddress.None))
+        for (int i = 0; i < maxAttempts; i++)
         {
-            if (_logger.IsWarn) _logger.Warn("External IP could not be resolved. Peers will not be able to connect.");
-            _ = RetryExternalIpAsync();
-        }
-    }
+            if (i > 0)
+            {
+                if (_logger.IsWarn) _logger.Warn($"External IP resolution failed (attempt {i}/{maxAttempts}). Retrying in {delaySeconds}s...");
+                await Task.Delay(TimeSpan.FromSeconds(delaySeconds));
+            }
 
-    private async Task RetryExternalIpAsync()
-    {
-        const int maxRetries = 5;
-        for (int i = 0; i < maxRetries; i++)
-        {
-            await Task.Delay(TimeSpan.FromSeconds(60));
             try
             {
-                IPAddress ip = await InitializeExternalIp();
-                if (!Equals(ip, IPAddress.Any) && !Equals(ip, IPAddress.None))
+                ExternalIp = await InitializeExternalIp();
+                if (!Equals(ExternalIp, IPAddress.Any) && !Equals(ExternalIp, IPAddress.None))
                 {
-                    ExternalIp = ip;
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Network/IPResolver.cs
```diff
@@ -44,6 +44,39 @@ public async Task Initialize()
...
+        for (int i = 0; i < maxRetries; i++)
+        {
+            await Task.Delay(TimeSpan.FromSeconds(60));
+            try
+            {
+                IPAddress ip = await InitializeExternalIp();
+                if (!Equals(ip, IPAddress.Any) && !Equals(ip, IPAddress.None))
+                {
+                    ExternalIp = ip;
+                    if (_logger.IsWarn) _logger.Warn($"External IP resolved: {ip}. Restart the node for peers to connect using the new IP.");
```
benaadams: I think the other person commenting on issue wants this part to fix without restart (rather than say warn restart)

smartprogrammer93: Agreed, but it is actually quite complicated to resolve without restart. If you think its worth the effort, ill do it

src/Nethermind/Nethermind.Network/IPResolver.cs
```diff
@@ -44,6 +44,39 @@ public async Task Initialize()
         {
             ExternalIp = IPAddress.None;
         }
+
+        if (Equals(ExternalIp, IPAddress.Any) || Equals(ExternalIp, IPAddress.None))
+        {
+            if (_logger.IsWarn) _logger.Warn("External IP could not be resolved. Peers will not be able to connect.");
+            _ = RetryExternalIpAsync();
```
asdacap: Might wanna block until 5 reply or something. Other component still use the ExternalIp and will not get updated if this failed.

LukaszRozmej: Agree better to block a bit longer

smartprogrammer93: Done — `Initialize()` now blocks with 5 attempts and 2s delays (~8s max extra wait).

flcl42:
ip resolution is a big problem - some code may use outdated values. I'd try to discover it in an endless cycle and allow to subscribe to update, so enrs and stuff can notify about update

src/Nethermind/Nethermind.Network/IPResolver.cs
```diff
@@ -36,14 +36,33 @@ public async Task Initialize()
...
+        const int maxAttempts = 5;
+        const int delaySeconds = 2;
+
+        for (int i = 0; i < maxAttempts; i++)
         {
-            ExternalIp = IPAddress.None;
+            if (i > 0)
+            {
+                if (_logger.IsWarn) _logger.Warn($"External IP resolution failed (attempt {i}/{maxAttempts}). Retrying in {delaySeconds}s...");
+                await Task.Delay(TimeSpan.FromSeconds(delaySeconds));
```
LukaszRozmej, line 47: Should we wait 2s in-between?

smartprogrammer93: Yes, assuming router is booting up and connecting

PR #10507, opened 2026-02-12
perf: inline identity precompile execution in CALL instruction

kamilchodola:
## Changes

- Add a precompile fast-path in `CALL` handling for the optimized execution path used by this branch (identity-focused, with guarded fallback to the existing path).
- Keep tracer/debug compatibility by using the existing execution path when tracing is active.
- Preserve EVM semantics around gas checks and state/snapshot handling (including value-transfer behavior).
- Introduce and wire `SupportsFastPath` on precompiles so fast-path eligibility is explicit.
- Ensure cached precompile wrappers preserve fast-path capability metadata.
- Align return-data copy behavior and RIPEMD handling with parity expectations.

- [x] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

- Added coverage in:
  - `Nethermind.Evm.Test/SupportsCachingTests.cs`
  - `Nethermind.Blockchain.Test/CachedCodeInfoRepositoryTests.cs`
  - `Nethermind.Core.Test/RipemdTests.cs`
- Ran branch benchmarks for relevant precompile scenarios (identity/ripemd) against `kch/gas-benchmarks-bdn` baseline with fixed BenchmarkDotNet settings (`--inprocess --warmupCount 10 --iterationCount 10 --launchCount 1`).
- CI workflow run for this branch was executed and validated during iteration.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Call.cs
```diff
@@ -302,6 +318,189 @@ static EvmExceptionType FastCall(VirtualMachine<TGasPolicy> vm, IReleaseSpec spe
             return EvmExceptionType.None;
         }
 
+        // Precompile inline execution: avoids frame allocation and state snapshot.
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static EvmExceptionType FastPrecompileCall(
```
LukaszRozmej, line 336: Can this go through standard call path with `<TFast> where TFast : IFlag` and based on this flag we can skip parts of that code?

This would be A LOT more maintainable.

LukaszRozmej (commented):
This looks unmaintaibable - maybe we could go with my comment?

src/Nethermind/Nethermind.Evm/Instructions/EvmInstructions.Call.cs
kamilchodola: please check now

benaadams:
Went through this carefully - gas accounting, state rollback, 63/64 rule, stipend handling all check out. A few things need fixing though:

---

**CallTests.cs - parity tests don't exercise the fast path**

`FastPathTracerWithOutput` only overrides `IsTracingActions => false` but inherits `IsTracingInstructions => true` from `TestAllTracerWithOutput`. The transaction processor picks `TTracingInst` based on `IsTracingInstructions`, so when that's `true` we get `TTracingInst = OnFlag`, which makes `!TTracingInst.IsActive` evaluate to `false` at the fast-path check. Both runs take the slow path -- the tests pass because they're comparing the same code path against itself.

Needs:
```csharp
internal class FastPathTracerWithOutput : TestAllTracerWithOutput
{
    public override bool IsTracingActions => false;
    public override bool IsTracingInstructions => false;
}
```

Would also be good to add an assertion or counter confirming the fast path was actually taken.

Also missing:
- No parity test with non-zero value transfer (that's the snapshot/restore branch)
- No CALLCODE parity test
- G1MSM, G2MSM, MapFpToG1, MapFp2ToG2, PairingCheck, PointEvaluation, ModExpPreEip2565 all have `SupportsFastPath => true` but no parity coverage

---

**Ripemd160Precompile.cs - needs a comment on why SupportsFastPath is false**

The fast path doesn't implement the EIP-161 Parity touch bug workaround from `RunPrecompile` in `VirtualMachine.cs`. If someone flips this to `true` without knowing that, it breaks consensus for historical sync. Needs a comment explaining this.

---

**EcRecoverPrecompile.cs - what's the exclusion reason?**

The fast path's `TryRunPrecompile` handles `DllNotFoundException` the same way (calls `Environment.Exit`). If it's just that the 3000-gas base cost makes frame overhead negligible, worth saying so in a comment.

---

**EvmInstructions.Call.cs line 409-414 - touch ordering comment**

The zero-value touch happens *after* success in the fast path, while the slow path touches *before* running the precompile. This works because RIPEMD is excluded, but the comment should mention *why* -- the Parity touch bug -- not just that we're mirroring the non-fast path.

---

**SupportsCachingTests.cs - Secp256r1Precompile missing**

`Secp256r1Precompile` isn't in `PrecompilesWithFastPathEnabled` and doesn't have an explicit `SupportsFastPath_ReturnsFalse` test. Should add one to document the exclusion is intentional.

PR #10513, opened 2026-02-13
fix(test): fix race condition in BlockchainProcessorTests (flaky test)

benaadams:
## Changes

- Wait for the block to be enqueued in the processor (not just added to the block tree) before returning from `Suggested()` in `BlockchainProcessorTests`
- Prevents a race where two concurrent `Enqueue` calls both see `_queueCount > 1` and enter the recovery queue in non-deterministic order, causing the processor to batch blocks together and deadlock the test

#### Requires testing

- [x] No

#### Notes on testing

The fix itself is in test infrastructure. All 18 `BlockchainProcessorTests` pass, including the previously flaky `Can_change_branch_on_invalid_block_when_invalid_branch_is_in_the_queue_and_recovery_queue_max_has_been_reached`.

- [x] No

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain.Test/BlockchainProcessorTests.cs
+++ b/src/Nethermind/Nethermind.Blockchain.Test/BlockchainProcessorTests.cs
@@ -288,17 +288,22 @@ public class BlockchainProcessorTests
             {
                 // Use Task.Run to avoid blocking when AllowSynchronousContinuations
                 // causes inline processing on the calling thread.
-                // Wait for BlockAdded to ensure the block's Enqueue (_queueCount increment)
-                // completes before the next Suggested call. This prevents a race where two
-                // concurrent Enqueue calls both see _queueCount > 1 and go to the recovery
-                // queue in non-deterministic order, causing the processor to batch blocks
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Blockchain.Test/BlockchainProcessorTests.cs
```diff
@@ -287,18 +287,62 @@ public ProcessingTestContext Suggested(Block block, BlockTreeSuggestOptions opti
...
+                TaskCompletionSource suggestCompleted = new(TaskCreationOptions.RunContinuationsAsynchronously);
+                void OnBlockAdded(object? sender, BlockEventArgs args)
                 {
-                    AddBlockResult result = _blockTree.SuggestBlock(block, options);
-                    if (result != AddBlockResult.Added)
+                    if (args.Block.Hash == block.Hash)
+                        suggestCompleted.TrySetResult();
+                }
+
+                ((IBlockProcessingQueue)_processor).BlockAdded += OnBlockAdded;
```
LukaszRozmej, line 309: Can you just cast it to a field in the constructor?

LukaszRozmej (approved):

PR #10514, opened 2026-02-13
perf: Cache intrinsic gas calculation to eliminate double computation

### LINKED ISSUE #9260: Refactor double calculation of intrinsic gas
Avoid calculating intrinsic gas twice - at validation and at execution:
https://github.com/NethermindEth/nethermind/blob/master/src/Nethermind/Nethermind.Consensus/Validators/TxValidator.cs#L121

smartprogrammer93:
Fixes #9260

## Changes

- Cache intrinsic gas (standard + floor) on `Transaction` using an embedded `IntrinsicGasCache` sealed class — published atomically via `Volatile.Write` / `Volatile.Read` (publication idiom)
- Cache invalidated automatically when a different spec is used (`ReferenceEquals` check)
- Replace `UInt256` with `long` arithmetic for init code ceiling division — avoids heap allocation
- `IntrinsicGas<T>` changed from `record struct` to plain `readonly struct` — avoids generated `Equals`/`GetHashCode`/`ToString` per generic instantiation

### Why a class + Volatile instead of a struct?

A `readonly struct` with multiple fields cannot be written atomically — concurrent readers can observe a torn state where `Spec` still points to the previous spec while the gas values have already been updated (or vice versa). A `lock` is correct but adds ~12–13 ns of uncontended monitor overhead on every call, erasing the cache benefit entirely.

The chosen approach uses an **immutable sealed class** with the [publication idiom](https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom):
- `Volatile.Write` on publish: release barrier ensures all constructor stores complete before the reference is visible to readers
- `Volatile.Read` on the hot path: acquire barrier ensures readers see the fully-constructed object
- On x86/x64 (TSO memory model): both are **compiler fences only** — zero additional CPU instructions
- On ARM: lightweight `dmb` barriers, negligible overhead
- Readers always see either `null` (cold) or a **complete, consistent snapshot** — no torn state possible under any concurrency

## Benchmarks

IntrinsicGasBenchmark (per-call cost of IntrinsicGasCalculator.Calculate)

  ┌──────────────┬───────────────────┬─────────────────────┬──────────────────────┬───────────────┐
  │  Benchmark   │ Master (no cache) │ Seqlock (cache hit) │ Seqlock (cache miss) │ Speedup (hit) │
  ├──────────────┼───────────────────┼─────────────────────┼──────────────────────┼───────────────┤
  │ SimpleTx     │ 2.679 ns          │ 0.756 ns            │ 5.508 ns             │ 3.5x          │
  ├──────────────┼───────────────────┼─────────────────────┼──────────────────────┼───────────────┤
  │ DataTx       │ 12.499 ns         │ 0.995 ns            │ 14.778 ns            │ 12.6x         │
  ├──────────────┼───────────────────┼─────────────────────┼──────────────────────┼───────────────┤
  │ AccessListTx │ 4.198 ns          │ 0.727 ns            │ 7.094 ns             │ 5.8x          │
  ├──────────────┼───────────────────┼─────────────────────┼──────────────────────┼───────────────┤
  │ SpecChange   │ 4.201 ns          │ —                   │ 7.280 ns             │ (always miss) │
  └──────────────┴───────────────────┴─────────────────────┴──────────────────────┴───────────────┘

  All variants: 0 B allocated.

  TxProcessingBenchmark (full transaction processing end-to-end)

  ┌──────────────┬──────────┬──────────────────┬────────────────────┐
  │  Benchmark   │  Master  │ Seqlock (cached) │ Seqlock (no cache) │
  ├──────────────┼──────────┼──────────────────┼────────────────────┤
  │ SimpleTx     │ 2.619 us │ 2.623 us         │ 2.525 us           │
  ├──────────────┼──────────┼──────────────────┼────────────────────┤
  │ DataTx       │ 2.534 us │ 2.483 us         │ 2.517 us           │
  ├──────────────┼──────────┼──────────────────┼────────────────────┤
  │ ContractCall │ 2.620 us │ 2.607 us         │ 2.610 us           │
  └──────────────┴──────────┴──────────────────┴────────────────────┘

  Full tx processing times are equivalent — intrinsic gas calculation (~5–13 ns) is <

## Types of changes

- [x] Optimization

#### Requires testing

- [x] Yes

#### If you wrote tests?

- [x] Yes
- `IntrinsicGasCalculatorTests.CacheBehaviorTests` (6 tests):
  - `Cache_InitiallyNullBeforeFirstCalculation`
  - `Cache_HitReturnsCachedValue`
  - `Cache_InvalidatedWhenSpecChanges`
  - `Cache_PragueFloorCosts`
  - `Cache_SpecTransition` (Homestead → Berlin → Prague)
  - `Cache_AccessListTransaction`

- [x] No

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Core/Transaction.cs
+++ b/src/Nethermind/Nethermind.Core/Transaction.cs
@@ -208,18 +208,18 @@ namespace Nethermind.Core
 
         protected int? _size = null;
 
-        // Intrinsic gas cache: immutable class published via Volatile.Write so that readers on all
-        // architectures always see either null or a fully-constructed snapshot — no torn reads.
-        // Volatile.Write/Read cost zero extra CPU instructions on x86/x64 (TSO); on ARM they emit
-        // lightweight barriers. The only overhead vs a struct is one pointer dereference on the hot path.
-        internal sealed class IntrinsicGasCache(long standard, long floor, IReleaseSpec? spec)
-        {
-            internal readonly long Standard = standard;
-            internal readonly long Floor = floor;
-            internal readonly IReleaseSpec? Spec = spec;
-        }
-
-        internal IntrinsicGasCache? _cachedIntrinsicGas;
+        // Seqlock-based intrinsic gas cache: three inline fields avoid heap allocation and pointer
+        // indirection while remaining fully thread-safe.
+        //
+        // Writers acquire the seqlock via CAS (even→odd), write data, then release by advancing to
+        // the next even number via Volatile.Write.  Readers validate the sequence number before and
+        // after reading — if it changed or is odd they fall back to recalculation.
+        //
+        // On x86/x64 (TSO) Volatile.Read/Write are just compiler fences — zero extra CPU cost.
+        // On ARM they emit lightweight acquire/release barriers.
+        internal int _gasCacheSeq;
+        internal long _cachedIntrinsicGasPacked;       // high 32: standard, low 32: floor
+        internal IReleaseSpec? _cachedIntrinsicGasSpec;
 
         /// <summary>
         /// Encoded transaction length
@@ -327,7 +327,9 @@ namespace Nethermind.Core
                 obj.IsServiceTransaction = default;
                 obj.PoolIndex = default;
                 obj._size = default;
-                obj._cachedIntrinsicGas = null;
+                obj._gasCacheSeq = 0;
+                obj._cachedIntrinsicGasPacked = 0;
+                obj._cachedIntrinsicGasSpec = null;
                 obj.AuthorizationList = default;
 
                 return true;
--- a/src/Nethermind/Nethermind.Evm/GasPolicy/EthereumGasPolicy.cs
+++ b/src/Nethermind/Nethermind.Evm/GasPolicy/EthereumGasPolicy.cs
@@ -179,14 +179,21 @@ public struct EthereumGasPolicy : IGasPolicy<EthereumGasPolicy>
 
     public static IntrinsicGas<EthereumGasPolicy> CalculateIntrinsicGas(Transaction tx, IReleaseSpec spec)
     {
-        // Volatile.Read: acquire barrier — ensures we see a fully-constructed object
-        // on all architectures (free on x64 TSO; lightweight barrier on ARM).
-        Transaction.IntrinsicGasCache? cached = Volatile.Read(ref tx._cachedIntrinsicGas);
-        if (cached is not null && ReferenceEquals(cached.Spec, spec))
+        // ── Seqlock read: check version, read data, re-check version ──────────
+        // If the version is odd a writer is active; if it changed between our two
+        // reads a write completed in between — both cases fall through to recalculation.
+        // On x86/x64 (TSO) Volatile.Read is just a compiler fence — zero CPU cost.
+        int s1 = Volatile.Read(ref tx._gasCacheSeq);
+        if ((s1 & 1) == 0)
         {
-            return new IntrinsicGas<EthereumGasPolicy>(
-                FromLong(cached.Standard),
-                FromLong(cached.Floor));
+            long packed = tx._cachedIntrinsicGasPacked;
+            IReleaseSpec? cachedSpec = tx._cachedIntrinsicGasSpec;
+            if (Volatile.Read(ref tx._gasCacheSeq) == s1 && ReferenceEquals(cachedSpec, spec))
+            {
+                return new IntrinsicGas<EthereumGasPolicy>(
+                    FromLong((uint)(packed >> 32)),
+                    FromLong((uint)packed));
+            }
         }
 
         long tokensInCallData = IGasPolicy<EthereumGasPolicy>.CalculateTokensInCallData(tx, spec);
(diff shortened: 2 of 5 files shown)


MarekM25:
@smartprogrammer93 is value worth complexity here? can you measure it. I have seen the AI bot has already generating such PR

LukaszRozmej:
I would close both this and the issue, not worth it.

src/Nethermind/Nethermind.Core/Transaction.cs
```diff
@@ -204,6 +207,13 @@ private void ClearPreHashInternal()
 
         protected int? _size = null;
 
+        // Intrinsic gas cache: plain long avoids Nullable<long> non-atomic writes on
+        // concurrent warmup paths. _cachedIntrinsicGasSpec is written last and acts as
+        // the sole validity indicator (non-null + ReferenceEquals ⇒ standard/floor valid).
+        internal long _cachedIntrinsicGasStandard;
+        internal long _cachedIntrinsicGasFloor;
+        internal IReleaseSpec? _cachedIntrinsicGasSpec;
```
LukaszRozmej: I think we should collapse those 3 fields into one struct (or class) so it will be better encapsulated

smartprogrammer93:
@LukaszRozmej @MarekM25 benchmarks in the pr description. Imo, worth it.

src/Nethermind/Nethermind.Benchmark.Runner/Program.cs
```diff
@@ -38,7 +38,7 @@ public DashboardConfig(params Job[] jobs)
 
     public class PrecompileBenchmarkConfig : DashboardConfig
     {
-        public PrecompileBenchmarkConfig() : base(Job.MediumRun.WithRuntime(CoreRuntime.Core90))
+        public PrecompileBenchmarkConfig() : base(Job.MediumRun.WithRuntime(CoreRuntime.CreateForNewVersion("net10.0", ".NET 10.0")))
```
LukaszRozmej: no enum flag? Maybe we need to update the package?

src/Nethermind/Nethermind.Evm/GasPolicy/EthereumGasPolicy.cs
```diff
@@ -178,13 +178,27 @@ public static EthereumGasPolicy Max(in EthereumGasPolicy a, in EthereumGasPolicy
...
                         + DataCost(tx, spec, tokensInCallData)
                         + CreateCost(tx, spec)
                         + IGasPolicy<EthereumGasPolicy>.AccessListCost(tx, spec)
                         + AuthorizationListCost(tx, spec);
         long floorCost = IGasPolicy<EthereumGasPolicy>.CalculateFloorCost(tokensInCallData, spec);
+
+        // Write Spec LAST so concurrent readers see it only after Standard/Floor are set
+        tx._cachedIntrinsicGas.Standard = standard;
+        tx._cachedIntrinsicGas.Floor = floorCost;
+        tx._cachedIntrinsicGas.Spec = spec;
```
LukaszRozmej: this should be a single write...
Come to thing of it maybe better to make it a class - so we construct new instance and just swap it?

src/Nethermind/Nethermind.Evm/GasPolicy/IGasPolicy.cs
```diff
@@ -273,10 +273,21 @@ protected static long CalculateFloorCost(long tokensInCallData, IReleaseSpec spe
...
+
+    public IntrinsicGas(TGasPolicy standard, TGasPolicy floorGas)
+    {
+        Standard = standard;
+        FloorGas = floorGas;
+        MinimalGas = TGasPolicy.Max(standard, floorGas);
+    }
+
     public static explicit operator TGasPolicy(IntrinsicGas<TGasPolicy> gas) => gas.MinimalGas;
 }
```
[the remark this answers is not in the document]
LukaszRozmej: what is the point of this change? Can we reverse it please?

src/Nethermind/Nethermind.Evm/GasPolicy/EthereumGasPolicy.cs
```diff
@@ -178,13 +178,27 @@ public static EthereumGasPolicy Max(in EthereumGasPolicy a, in EthereumGasPolicy
 
     public static IntrinsicGas<EthereumGasPolicy> CalculateIntrinsicGas(Transaction tx, IReleaseSpec spec)
     {
+        // Cache check: Spec written last acts as sole validity indicator
+        if (ReferenceEquals(tx._cachedIntrinsicGas.Spec, spec))
+        {
+            return new IntrinsicGas<EthereumGasPolicy>(
+                FromLong(tx._cachedIntrinsicGas.Standard),
+                FromLong(tx._cachedIntrinsicGas.Floor));
+        }
```
LukaszRozmej: Also here first take cached value to local variable!

src/Nethermind/Nethermind.Evm/GasPolicy/EthereumGasPolicy.cs
smartprogrammer93: Class was the initial implementation to unify them, but was too slow

LukaszRozmej: weird, why it would be slow? Can you double-check?

src/Nethermind/Nethermind.Evm/GasPolicy/EthereumGasPolicy.cs
```diff
@@ -194,10 +195,8 @@ public static IntrinsicGas<EthereumGasPolicy> CalculateIntrinsicGas(Transaction
                         + AuthorizationListCost(tx, spec);
         long floorCost = IGasPolicy<EthereumGasPolicy>.CalculateFloorCost(tokensInCallData, spec);
 
-        // Write Spec LAST so concurrent readers see it only after Standard/Floor are set
-        tx._cachedIntrinsicGas.Standard = standard;
-        tx._cachedIntrinsicGas.Floor = floorCost;
-        tx._cachedIntrinsicGas.Spec = spec;
+        // Spec field is last in struct layout; single assignment preserves write ordering on common architectures
+        tx._cachedIntrinsicGas = new Transaction.IntrinsicGasCache(standard, floorCost, spec);
```
LukaszRozmej: This could still give incorrect results on multi-threaded access, as the spec still can point to previous spec, while the cost can already be updated.

PR #10523, opened 2026-02-14
fix: fix incorrect error messages in EraImporter

andrewshab3:
## Changes

- Fix typo "then" → "than" in error message for upper bound check
- Fix wrong word "lower" → "higher" in error message for lower bound check — the condition `from < firstBlockInStore` means the store's lowest block is *higher* than requested, not lower

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_


smartprogrammer93:
Not meaningful pr, closing

PR #10529, opened 2026-02-14
fix(jsonrpc): include block identifier in pruned history error message

### LINKED ISSUE #9539: Log block request block number on history pruned error
```
22 Oct 09:56:32 | Error response handling JsonRpc Id:369 Method:eth_getBlockByHash | Code: 4444 Message: Pruned history unavailable
```

- The above shows the error message when block is not found. 
- Unfortunately it does not show which block it is trying to request.
- Change the error message so that the block request is known.

smartprogrammer93:
Closes #9539

## Changes

- Include the requested block number, hash, or type in the "pruned history unavailable" error message

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks

Note: Geth returns a generic `"pruned history unavailable"` without the block identifier. This change deviates from Geth's behavior by including the requested block number/hash/type in the error message for better debuggability.

### DIFF
--- a/src/Nethermind/Nethermind.JsonRpc/Modules/BlockFinderExtensions.cs
+++ b/src/Nethermind/Nethermind.JsonRpc/Modules/BlockFinderExtensions.cs
@@ -80,7 +80,7 @@ namespace Nethermind.JsonRpc.Modules
                 if (!allowNulls)
                 {
                     return new SearchResult<Block>(
-                        $"block {blockParameter} could not be found",
+                        $"Block {blockParameter} could not be found",
                         ErrorCodes.ResourceNotFound);
                 }
             }


src/Nethermind/Nethermind.JsonRpc/Modules/BlockFinderExtensions.cs
```diff
@@ -72,7 +72,9 @@ public static SearchResult<Block> SearchForBlock(this IBlockFinder blockFinder,
 
                 if (blockFinder.IsBlockPruned(blockParameter))
                 {
-                    return new SearchResult<Block>("Pruned history unavailable", ErrorCodes.PrunedHistoryUnavailable);
+                    return new SearchResult<Block>(
+                        $"pruned history unavailable for block {blockParameter.BlockHash?.ToString() ?? blockParameter.BlockNumber?.ToString() ?? blockParameter.Type.ToString()}",
```
asdacap: Is the capital P important?

asdacap (approved):

src/Nethermind/Nethermind.JsonRpc/Modules/BlockFinderExtensions.cs
smartprogrammer93: Yes, to match geth returned string, it needs to be a small 'p'

LukaszRozmej: Maybe just change `BlockParameter.ToString()`, so we can simplify to:
```suggestion
                        $"pruned history unavailable for block {blockParameter}",
```

LukaszRozmej (approved):

src/Nethermind/Nethermind.JsonRpc/Modules/BlockFinderExtensions.cs
smartprogrammer93: Done! Overrode `BlockParameter.ToString()` to return a clean human-readable string (lowercase type names for `latest`/`earliest`/etc., and the raw number or hash for block number/hash types). All three verbose nullable-chain patterns in `BlockFinderExtensions` are now simplified to just `{`blockParameter`}`.

PR #10531, opened 2026-02-14
Fix ConfigSourceHelper.GetDefault to return actual default value instead of tuple

prestoalvarez:
- Fixed `ConfigSourceHelper.GetDefault()` to return the actual default value instead of wrapping it in a tuple
- Added regression test to verify unset config values return correct defaults


LukaszRozmej:
fails

PR #10534, opened 2026-02-14
fix(rpc): prevent TrieException by checking BestPersistedState in HasStateForBlock

### LINKED ISSUE #9408: eth_getBalance responds with TrieException

smartprogrammer93:
Closes #9408

## Changes

- Add `BestPersistedState` check in `BlockchainBridge.HasStateForBlock` to prevent accessing partially-pruned trie state
- When a block is older than `BestPersistedState - 64` (reorg safety margin), return `false` before trie traversal
- Prevents `MissingTrieNodeException` from propagating as internal error on pruned nodes

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

4 unit tests added in `BlockchainBridgeTests.cs`:
- `HasStateForBlock_returns_false_when_block_older_than_BestPersistedState` — verifies blocks older than persisted boundary are rejected
- `HasStateForBlock_returns_true_when_block_within_BestPersistedState_range` — verifies recent blocks pass through
- `HasStateForBlock_returns_true_when_BestPersistedState_is_null` — verifies archive/unset mode falls back to HasRoot
- `HasStateForBlock_returns_false_when_root_does_not_exist` — verifies HasRoot still checked first

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Trie.Test/Pruning/TreeStoreTests.cs
+++ b/src/Nethermind/Nethermind.Trie.Test/Pruning/TreeStoreTests.cs
@@ -942,12 +942,6 @@ namespace Nethermind.Trie.Test.Pruning
                 trieStore.HasRoot(rootHashes[oldBlock], oldBlock).Should().BeFalse();
             }
 
-            // HasRoot without block number still returns true (root node exists)
-            if (oldBlock >= 0)
-            {
-                trieStore.HasRoot(rootHashes[oldBlock]).Should().BeTrue();
-            }
-
             trieStore.Dispose();
         }
 


src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
```diff
@@ -405,9 +407,23 @@ public void RunTreeVisitor<TCtx>(ITreeVisitor<TCtx> treeVisitor, BlockHeader? ba
...
-            return stateReader.HasStateForBlock(baseBlock);
+            if (!stateReader.HasStateForBlock(baseBlock))
+                return false;
+
+            // Reject blocks whose state may have been partially pruned (root exists but child nodes don't)
+            if (baseBlock is not null && pruningConfig.Mode != PruningMode.None)
+            {
+                long? bestPersistedState = blockTree.BestPersistedState;
+                if (bestPersistedState.HasValue &&
+                    baseBlock.Number < bestPersistedState.Value - Reorganization.MaxDepth)
```
LukaszRozmej: use `PruningConfig.PruningBoundary` instead `Reorganization.MaxDepth`
this should probably be `head - PruningConfig.PruningBoundary` though?

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
```diff
@@ -52,7 +53,8 @@ public class BlockchainBridge(
         ILogFinder logFinder,
         ISpecProvider specProvider,
         IBlocksConfig blocksConfig,
-        IMiningConfig miningConfig)
+        IMiningConfig miningConfig,
+        IPruningConfig pruningConfig)
```
asdacap: Please dont bind IPruningConfig with BlockchainBridge. This will break flat as flat does not use this. Instead add the logic within TrieStore.

smartprogrammer93: You're right, thanks for the catch. Refactored:  is now removed from  entirely. The pruning boundary check has been moved into  — a new overload on  with a default passthrough implementation (so flat/archive nodes are unaffected).  now passes the block number through to , so the check happens at the trie layer where it belongs.

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
smartprogrammer93: Fixed — switched to  (which maps to ) in the  implementation of . The check is now .

PR #10537, opened 2026-02-15
Fix XDC gas limit calculation 

ak88:
XDC devnet uses 1559 and also setting gas limit dynamically

### DIFF
--- a/src/Nethermind/Nethermind.Xdc/XdcBlockProducer.cs
+++ b/src/Nethermind/Nethermind.Xdc/XdcBlockProducer.cs
@@ -58,14 +58,9 @@ internal class XdcBlockProducer : BlockProducerBase
             blockAuthor,
             UInt256.Zero,
             parent.Number + 1,
-            //This should probably use TargetAdjustedGasLimitCalculator
             gasLimit,
             0,
-            extra)
-        {
-            //This will BestSuggestedBody in BlockTree, which may not be needed
-            //IsPostMerge = true
-        };
+            extra);
 
         IXdcReleaseSpec spec = specProvider.GetXdcSpec(xdcBlockHeader, currentRound);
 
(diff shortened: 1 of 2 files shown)


Demuirgos (approved):

src/Nethermind/Nethermind.Xdc/XdcBlockProducer.cs
```diff
@@ -51,19 +51,20 @@ protected override BlockHeader PrepareBlockHeader(BlockHeader parent, PayloadAtt
...
             parent.Number + 1,
             //This should probably use TargetAdjustedGasLimitCalculator
-            XdcConstants.TargetGasLimit,
+            gasLimit,
             0,
             extra)
         {
             //This will BestSuggestedBody in BlockTree, which may not be needed
-            IsPostMerge = true
+            //IsPostMerge = true
```
cicr99: Can this be removed then?

src/Nethermind/Nethermind.Xdc/Spec/XdcChainSpecBasedSpecProvider.cs
```diff
@@ -32,6 +32,7 @@ protected override ReleaseSpec CreateReleaseSpec(ChainSpec chainSpec, long relea
         releaseSpec.IsBlackListingEnabled = chainSpecEngineParameters.BlackListHFNumber <= releaseStartBlock;
         releaseSpec.IsTIP2019 = chainSpecEngineParameters.TIP2019Block <= releaseStartBlock;
         releaseSpec.IsTIPXDCXMiner = chainSpecEngineParameters.TipXDCX <= releaseStartBlock && releaseStartBlock < chainSpecEngineParameters.TIPXDCXMinerDisable;
+        releaseSpec.IsDynamicGasLimitBlock = chainSpecEngineParameters.DynamicGasLimitBlock <= releaseStartBlock;
```
cicr99, line 35: If `DynamicGasLimitBlock` is not set in `xdc.json`, this flag should always be true. Is that the expected behavior here?

ak88: true, i fixed it, but we will need to adjust all the block activations.

src/Nethermind/Nethermind.Xdc/XdcBlockProducer.cs
ak88: removed

PR #10538, opened 2026-02-15
Remove unused write batch allocation in TrieStore.PersistOnShutdown

marukai67:
Removes redundant `IWriteBatch` allocation in `PersistOnShutdown()` that was created but never used for any write operations.


src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -1285,14 +1285,12 @@ private void PersistOnShutdown()
...
 
-        INodeStorage.IWriteBatch writeBatch = _nodeStorage.StartWriteBatch();
         for (int index = 0; index < candidateSets.Count; index++)
         {
             BlockCommitSet blockCommitSet = candidateSets[index];
             if (_logger.IsDebug) _logger.Debug($"Persisting on disposal {blockCommitSet} (cache memory at {MemoryUsedByDirtyCache})");
             ParallelPersistBlockCommitSet(blockCommitSet, _persistedNodeRecorderNoop);
         }
-        writeBatch.Dispose();
         _nodeStorage.Flush(onlyWal: false);
```
LukaszRozmej, line 1280: `ParallelPersistBlockCommitSet` already calls `_nodeStorage.Flush` but with `onlyWal: true`. This should be optional parameter in `ParallelPersistBlockCommitSet` and could be removed from here.

asdacap: This should be removed completely, as this slowdown shutdown and if its in something like docker, the docker might resort to kill if it take too much time for shutdown.

asdacap (approved):

PR #10549, opened 2026-02-17
XDC FastSync 

batrr:
In XDC, header verification requires snapshots from previous blocks; however, these are not loaded during fast sync because previous headers are not processed normally. As a result, after fast sync completes, we need some snapshots from the gap blocks before.

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Xdc/XdcStateSyncPivot.cs
+++ b/src/Nethermind/Nethermind.Xdc/XdcStateSyncPivot.cs
@@ -21,7 +21,6 @@ public class XdcStateSyncPivot : IStateSyncPivot
     private readonly IBlockTree _blockTree;
     private readonly ISyncConfig _syncConfig;
     private readonly IStateReader _stateReader;
-    private readonly ILogger _logger;
     private readonly Queue<XdcBlockHeader> _targets = new();
     private XdcBlockHeader? _pivotHeader;
     private bool _initialized;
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Xdc/XdcStateSyncSnapshotManager.cs
```diff
@@ -0,0 +1,76 @@
...
+        IXdcReleaseSpec spec = _specProvider.GetXdcSpec(pivotHeader);
+
+        XdcBlockHeader epochSwitchHeader = pivotHeader;
+
+        while (!_epochSwitchManager.IsEpochSwitchAtBlock(epochSwitchHeader))
+        {
+            epochSwitchHeader = (XdcBlockHeader)_blockTree.FindHeader(epochSwitchHeader.ParentHash);
+        }
+
+        long gapBlockNum = epochSwitchHeader.Number - epochSwitchHeader.Number % spec.EpochLength - spec.Gap;
```
cicr99: This could return a negative number if `epochSwitchHeader.Number = 0` (or smaller than epochLength in general). Is that scenario possible here?

ak88: yes better do a Math.Min in this case

src/Nethermind/Nethermind.Xdc/XdcStateSyncPivot.cs
```diff
@@ -0,0 +1,85 @@
...
+        _blockTree = blockTree;
+        _syncConfig = syncConfig;
+        _stateReader = stateReader;
+        _syncSnapshotManager = syncSnapshotManager;
+        _logger = logManager?.GetClassLogger() ?? throw new ArgumentNullException(nameof(logManager));
+    }
+
+    public BlockHeader? GetPivotHeader()
+    {
+        EnsureInitialized();
```
asdacap, line 43: Probably rename to "RefreshTargets" since its called multiple time.

PR #10551, opened 2026-02-17
Rename RLP's [] and null

flcl42:
Should be:
Empty byte array = empty BYTE array = null = 0x80
Empty list = [] = 0xc0
Names picked from the paper (Appendix B)
image

Potentially affects:
- Access list encoder
- Receipts storage
- L1Origin encoding in OP
- XDC, where these constants are used
- More decoders where `IsNextItemNull` is used

## Changes

- Use proper RLP encodings for [] and null

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No


src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
```diff
@@ -35,15 +35,15 @@ public class Rlp
...
-        public const byte NullObjectByte = 192; // use bytes to avoid stack overflow
+        public static readonly Rlp OfNullOrZero = new(NullOrZeroByte);
 
-        internal const int LengthOfNull = 1;
+        public const byte EmptyArrayByte = 0xc0; // use bytes to avoid stack overflow
 
         public static readonly Rlp OfEmptyByteArray = new(EmptyArrayByte);
 
-        public static readonly Rlp OfEmptySequence = new(NullObjectByte);
+        internal const int LengthOfNull = 1;
```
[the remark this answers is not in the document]
flcl42: Renamed

benaadams (approved):
Everything was backwards?

src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
ak88: Isnt it weird we allow a list to be null here, since lists cannot be null in RLP?

src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
```diff
@@ -347,7 +347,7 @@ public static Rlp Encode<T>(T[]? items, RlpBehaviors behaviors = RlpBehaviors.No
         {
             if (items is null)
             {
-                return OfEmptySequence;
```
ak88, line 261: I think treating null arrays as an empty list is the more correct approach, since a null would end up being invalid RLP when you are expecting list.

src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
```diff
@@ -1559,6 +1564,9 @@ public byte[][] DecodeByteArrays()
                 return result;
             }
 
+            public byte[][] DecodeEnsureByteArrays()
```
ak88: Do we have scenarios where we want to decode RLP even if its invalid RLP?

src/Nethermind/Nethermind.Xdc/RLP/QuorumCertificateDecoder.cs
```diff
@@ -46,13 +50,17 @@ protected override QuorumCertificate DecodeInternal(ref Rlp.ValueDecoderContext
 
     protected override QuorumCertificate DecodeInternal(RlpStream rlpStream, RlpBehaviors rlpBehaviors = RlpBehaviors.None)
     {
-        int sequenceLength = rlpStream.ReadSequenceLength();
-        if (sequenceLength == 0)
+        if (rlpStream.IsNextItemNull())
+        {
+            rlpStream.ReadByte();
```
ak88: This would actually be invalid RLP. I think the old sequence check is correct.

src/Nethermind/Nethermind.Xdc/RLP/BaseSnapshotDecoder.cs
```diff
@@ -14,7 +14,10 @@ internal abstract class BaseSnapshotDecoder<T> : RlpValueDecoder<T> where T : Sn
     protected TResult DecodeBase<TResult>(ref Rlp.ValueDecoderContext decoderContext, Func<long, Hash256, Address[], TResult> createSnapshot, RlpBehaviors rlpBehaviors = RlpBehaviors.None) where TResult : Snapshot
     {
         if (decoderContext.IsNextItemNull())
```
ak88: I think these null checks in almost all the XDC decoders are wrong and should actually be sequence checks like in QuorumCertificateDecoder 
@cicr99

src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
flcl42: We actually can try to get rid of DecodeByteArrays that may return null
Do we expect signatureBytes for sure here: https://github.com/NethermindEth/nethermind/blob/e6ba00b6c0cc2c665889445489e9a4808528c52b/src/Nethermind/Nethermind.Xdc/RLP/QuorumCertificateDecoder.cs?plain=1#L29C1-L29C70 and here https://github.com/NethermindEth/nethermind/blob/e6ba00b6c0cc2c665889445489e9a4808528c52b/src/Nethermind/Nethermind.Xdc/RLP/TimeoutCertificateDecoder.cs?plain=1#L26C1-L26C70 ?

src/Nethermind/Nethermind.Xdc/RLP/QuorumCertificateDecoder.cs
flcl42: Yeah

src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
flcl42: Yeah, let's expect empty array at least

flcl42:
@ak88 @benaadams  I remade it so it's just renaming + some to be sure byte reads + small cleanup here and there. I think we could even raise an exception in XDC if it's not expected to get null, wdyt @ak88?

PR #10558, opened 2026-02-17
fix: fix inverted condition in SimpleFilePublicKeyDb silently dropping value updates

andrewshab3:
## Changes

Fix inverted equality check in `SimpleFilePublicKeyDb.Set` that caused updates to existing keys to be silently dropped

The condition `!Bytes.AreEqual(existingValue, value)` set `setValue = false` when the new value was *different* from the existing one — the exact opposite of the intended optimization. As a result, value updates were never persisted while identical values triggered redundant writes.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_


src/Nethermind/Nethermind.Db/SimpleFilePublicKeyDb.cs
LukaszRozmej: this can be simplified to:

```
bool setValue = !_cacheSpan.TryGetValue(key, out var existingValue) || !Bytes.AreEqual(existingValue, value);
```

LukaszRozmej (approved):

asdacap (approved):

PR #10559, opened 2026-02-17
Estimate Gas fix

### LINKED ISSUE #10552: `eth_estimateGas` intermittently returns incorrect results / reverts on v1.36.0 (regression from v1.35.x)
Description

eth_estimateGas intermittently fails or returns incorrect results on Nethermind v1.36.0. The same calls succeed consistently on v1.35.7/v1.35.8. Multiple independent node operators have confirmed this regression.

Steps to Reproduce

1. Run Nethermind v1.36.0
2. Send eth_estimateGas RPC calls under normal load
3. Observe intermittent failures — the same call may succeed on retry
4. Optionally: run v1.35.7/v1.35.8 side-by-side and send identical calls to both to confirm discrepancy

Actual behavior

eth_estimateGas intermittently returns a gas estimate of 0 or fails with one of:

Code: -32000 Message: execution reverted

Code: -32000 Message: Transaction execution fails

Failures are non-deterministic — retrying the same call moments later often succeeds. A side-by-side test of v1.35.7 and v1.36.0 against the same calls at the same time confirmed v1.36.0 sometimes returns 0 where v1.35.7 returns the correct estimate.

Expected behavior

eth_estimateGas should return consistent, correct gas estimates matching v1.35.x behavior.

Screenshots

Image

Desktop (please complete the following information):

• Operating System: Linux
• Version: v1.36.0 (regression not present in v1.35.7 / v1.35.8)
• Installation Method: Unknown
• Consensus Client: Unknown
• Network: Gnosis (at minimum, may affect others)

Node Config:
```
--config=gnosis
--datadir=/data
--log=INFO
--logger-config=/etc/nethermind/NLog.config
--Sync.SnapSync=true
--Sync.NetworkingEnabled=true
--JsonRpc.Enabled=true
--JsonRpc.Host=0.0.0.0
--JsonRpc.Port=8545
--JsonRpc.JwtSecretFile=/data/jwt.hex
--JsonRpc.EngineHost=0.0.0.0
--JsonRpc.EnginePort=8551
--Network.DiscoveryPort=30303
--Network.MaxActivePeers=20
--HealthChecks.Enabled=true
--Metrics.Enabled=true
--Metrics.ExposeHost=0.0.0.0
--Metrics.ExposePort=6060
```

Additional context

• Potentially related to PR #9822 (Allow Inner Reverts for eth_gasEstimate), which modified how contract reverts are handled during gas estimation. The original fix for #9064 may incorrectly treat inner reverts as top-level failures. Ben Adams identified this as a likely cause in Discord.
• Multiple independent node operators have downgraded to v1.35.8 as a workaround.

Script to reproduce:
```
import secrets
import sys
from datetime import datetime

from eth_account import Account
from web3 import Web3

rpc_node = "https://gnosis-node.localhost/"
SAFE_SINGLETON_ADDRESS = "0xd9Db270c1B5E3Bd161E8c8503c55cEABeE709552"
PROXY_FACTORY_ADDRESS = "0xa6B71E26C5e0845f74c812102Ca7114b6a896AB2"
FALLBACK_HANDLER_ADDRESS = "0xf48f2B2d2a534e402487b3ee7C18c33Aec0Fe5e4"
ZERO_ADDRESS = "0x0000000000000000000000000000000000000000"
PRIVATE_KEY = open("pk.txt").read().strip()
ACCOUNT = Account.from_key(PRIVATE_KEY)

# GnosisSafe setup function ABI
SAFE_SETUP_ABI = {
    "inputs": [
        {"internalType": "address[]", "name": "_owners", "type": "address[]"},
        {"internalType": "uint256", "name": "_threshold", "type": "uint256"},
        {"internalType": "address", "name": "to", "type": "address"},
        {"internalType": "bytes", "name": "data", "type": "bytes"},
        {"internalType": "address", "name": "fallbackHandler", "type": "address"},
        {"internalType": "address", "name": "paymentToken", "type": "address"},
        {"internalType": "uint256", "name": "payment", "type": "uint256"},
        {
            "internalType": "address payable",
            "name": "paymentReceiver",
            "type": "address",
        },
    ],
    "name": "setup",
    "outputs": [],
    "stateMutability": "nonpayable",
    "type": "function",
}

# GnosisSafeProxyFactory createProxyWithNonce function ABI
FACTORY_CREATE_PROXY_ABI = {
    "inputs": [
        {"internalType": "address", "name": "_singleton", "type": "address"},
        {"internalType": "bytes", "name": "initializer", "type": "bytes"},
        {"internalType": "uint256", "name": "saltNonce", "type": "uint256"},
    ],
    "name": "createProxyWithNonce",
    "outputs": [
        {"internalType": "contract GnosisSafeProxy", "name": "proxy", "type": "address"}
    ],
    "stateMutability": "nonpayable",
    "type": "function",
}


def get_estimation(rpc_url: str) -> int:
    web3 = Web3(Web3.HTTPProvider(rpc_url, request_kwargs={"timeout": 30}))
    safe_contract = web3.eth.contract(
        address=web3.to_checksum_address(SAFE_SINGLETON_ADDRESS),
        abi=[SAFE_SETUP_ABI],
    )
    factory_contract = web3.eth.contract(
        address=web3.to_checksum_address(PROXY_FACTORY_ADDRESS),
        abi=[FACTORY_CREATE_PROXY_ABI],
    )
    owner_address = web3.to_checksum_address(ACCOUNT.address)
    setup_call = safe_contract.functions.setup(
        [owner_address],
        1,
        web3.to_checksum_address(ZERO_ADDRESS),
        b"",
        web3.to_checksum_address(FALLBACK_HANDLER_ADDRESS),
        web3.to_checksum_address(ZERO_ADDRESS),
        0,
        web3.to_checksum_address(ZERO_ADDRESS),
    )
    initializer = setup_call._encode_transaction_data()
    salt_nonce = secrets.SystemRandom().randint(0, 2**256 - 1)
    return factory_contract.functions.createProxyWithNonce(
        web3.to_checksum_address(SAFE_SINGLETON_ADDRESS),
        initializer,
        salt_nonce,
    ).estimate_gas({"from": owner_address})


if __name__ == "__main__":
    for attempt in range(250):
        try:
            gas = get_estimation(rpc_node)
            prefix = f"[{datetime.now()}] Gnosis"
            print(f"{prefix} Gas Estimation (Attempt {attempt}): {gas}")
        except Exception as e:
            prefix = f"[{datetime.now()}] Gnosis"
            print(f"{prefix} Failed to estimate gas on attempt {attempt}: {e}")
            sys.exit(1)
```

svlachakis:
Fixes Closes Resolves #10552 

## Changes

PR #9822 fixed TopLevelRevert to be nesting-aware - it only triggers when the top-level call (nesting level 0) reverts. Inner reverts are correctly ignored. But OutOfGas was left as a global flag - it gets set when ANY call at ANY depth runs out of gas.

The gas estimation binary search checks both flags:

```
return result.TransactionExecuted && gasTracer.StatusCode == StatusCode.Success &&
       !gasTracer.OutOfGas && !gasTracer.TopLevelRevert;
```

So if a contract makes a gas-limited inner call that OOGs (even if the caller catches the failure and continues successfully), the binary search considers that gas level a failure.

---

### Why This Differs from Geth

In Geth, result.Failed() only checks the top-level execution outcome. The EVM internally handles inner call failures - the parent CALL gets 0 on the stack and continues. Inner OOG never propagates to the estimation logic.

In Nethermind (before this fix), inner OOG propagated through the tracer, causing the binary search to reject gas levels where the transaction actually succeeds.

---

### How This Could Cause the GnosisSafe Issue

GnosisSafe's createProxyWithNonce does CREATE2 → setup call → inner calls. If any subcall in that chain runs out of gas (gas-limited probe calls, limited-gas forwarding), Nethermind's estimation would fail at that binary search midpoint while Geth's would not.

The intermittent nature could fit because different gas levels in the binary search would trigger different inner call behaviors.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain.Test/TransactionProcessorTests.cs
+++ b/src/Nethermind/Nethermind.Blockchain.Test/TransactionProcessorTests.cs
@@ -474,32 +474,19 @@ public class TransactionProcessorTests
 
     private void ConfirmEnoughEstimate(Transaction tx, Block block, long estimate)
     {
+        var blkCtx = new BlockExecutionContext(block.Header, _specProvider.GetSpec(block.Header));
+
         CallOutputTracer outputTracer = new();
         tx.GasLimit = estimate;
-        TestContext.Out.WriteLine(tx.GasLimit);
-
-        GethLikeTxMemoryTracer gethTracer = new(tx, GethTraceOptions.Default);
-        var blkCtx = new BlockExecutionContext(block.Header, _specProvider.GetSpec(block.Header));
-        _transactionProcessor.CallAndRestore(tx, blkCtx, gethTracer);
-        string traceEnoughGas = new EthereumJsonSerializer().Serialize(gethTracer.BuildResult(), true);
-
         _transactionProcessor.CallAndRestore(tx, blkCtx, outputTracer);
-        traceEnoughGas.Should().NotContain("OutOfGas");
+        outputTracer.StatusCode.Should().Be(StatusCode.Success,
+            $"transaction should succeed at the estimate ({estimate})");
 
         outputTracer = new CallOutputTracer();
         tx.GasLimit = Math.Min(estimate - 1, estimate * 63 / 64);
-        TestContext.Out.WriteLine(tx.GasLimit);
-
-        gethTracer = new GethLikeTxMemoryTracer(tx, GethTraceOptions.Default);
-        _transactionProcessor.CallAndRestore(tx, blkCtx, gethTracer);
-
-        string traceOutOfGas = new EthereumJsonSerializer().Serialize(gethTracer.BuildResult(), true);
-        TestContext.Out.WriteLine(traceOutOfGas);
-
         _transactionProcessor.CallAndRestore(tx, blkCtx, outputTracer);
-
-        bool failed = traceEnoughGas.Contains("failed") || traceEnoughGas.Contains("OutOfGas");
-        failed.Should().BeTrue();
+        outputTracer.StatusCode.Should().Be(StatusCode.Failure,
+            $"transaction should fail below the estimate ({tx.GasLimit})");
     }
 
     [TestCase]


src/Nethermind/Nethermind.Blockchain.Test/TransactionProcessorTests.cs
```diff
@@ -466,10 +466,19 @@ public void Can_estimate_with_destroy_refund_and_below_intrinsic_pre_berlin()
         tracer.CalculateAdditionalGasRequired(tx, releaseSpec).Should().Be(24080);
         tracer.GasSpent.Should().Be(35228L);
         long estimate = estimator.Estimate(tx, block.Header, tracer, out string? err, 0);
-        estimate.Should().Be(59307);
+        estimate.Should().Be(54225);
```
svlachakis, line 469: The `Can_estimate_with_destroy_refund_and_below_intrinsic_pre_berlin` test previously expected an estimate of **59307**. With the nesting-aware OutOfGas fix, it now returns **54225** — a lower, more accurate estimate.

At this gas level, the inner `CALL` to the `SELFDESTRUCT` contract runs out of gas, but the top-level `CREATE` transaction still succeeds. The old estimate was inflated because the binary search rejected `54225` due to the inner OOG, even though the transaction completes successfully.

This matches Geth's behavior. Geth’s binary search implementation:
https://github.com/ethereum/go-ethereum/blob/master/eth/gasestimator/gasestimator.go

uses `result.Failed()` to decide whether to raise or lower the gas bound. `Failed()` is defined in: https://github.com/ethereum/go-ethereum/blob/master/core/state_transition.go

as:

```
result.Err != nil
```

`Err` reflects only the top-level EVM outcome. An inner call Out-Of-Gas is handled internally by the `CALL` opcode (which pushes `0` onto the stack) and does not propagate to `result.Err`. Therefore, Geth would accept `54225` as a valid estimate.

The test validation was changed from `ConfirmEnoughEstimate` (which uses `GethLikeTxTracer` and asserts no OOG at any call depth) to inline `CallOutputTracer.StatusCode` checks that verify the top-level transaction succeeds at the estimate and fails below it — matching what Geth actually validates.

From Geth’s `gasestimator.go`:

> "It is a bit pointless to return a perfect estimation, as changing network conditions require the caller to bump it up anyway. Since wallets tend to use 20-25% bump, allowing a small approximation error is fine (as long as it's upwards)."

src/Nethermind/Nethermind.Blockchain.Test/TransactionProcessorTests.cs
```diff
@@ -466,10 +466,19 @@ public void Can_estimate_with_destroy_refund_and_below_intrinsic_pre_berlin()
...
+        CallOutputTracer outputTracer = new();
+        tx.GasLimit = estimate;
+        var blkCtx2 = new BlockExecutionContext(block.Header, _specProvider.GetSpec(block.Header));
+        _transactionProcessor.CallAndRestore(tx, blkCtx2, outputTracer);
+        outputTracer.StatusCode.Should().Be(StatusCode.Success);
+
+        outputTracer = new CallOutputTracer();
+        tx.GasLimit = Math.Min(estimate - 1, estimate * 63 / 64);
+        _transactionProcessor.CallAndRestore(tx, blkCtx2, outputTracer);
+        outputTracer.StatusCode.Should().Be(StatusCode.Failure);
```
smartprogrammer93: whats thee point from all these additions, unnecessary imo

src/Nethermind/Nethermind.Blockchain.Test/TransactionProcessorTests.cs
smartprogrammer93: best to confirm this against geth

smartprogrammer93: I can confirm, geth returns the same exact value as the new implementation we have here.

src/Nethermind/Nethermind.Blockchain.Test/TransactionProcessorTests.cs
smartprogrammer93:
```suggestion
        ConfirmEnoughEstimate(tx, block, estimate);
```

svlachakis: It's the equilevant of that confirm estimate but without geth tracer because it fails

svlachakis: I refactored the shared `ConfirmEnoughEstimate` - all should pass now

PR #10565, opened 2026-02-18
Flat/further tuning

asdacap:
### 1. Read-only trie node separation
Trie nodes that are only read (not modified) are no longer written into the changed set. A separate read cache (`ReadStateNodes`/`ReadStorageNodes` in `TransientResource`) handles lookups, while `_changedStateNodes`/`_changedStorageNodes` are only populated on actual writes. A `_trieChanged` flag gates checking the changed dictionaries, avoiding unnecessary lookups when no writes have occurred. This reduces memory churn in the compaction pipeline.

### 2. Hierarchical compact pattern
Instead of two levels of compaction (mid/full), the compact size is now determined per block using `blockNumber & -blockNumber` (largest power-of-2 divisor), clamped to `compactSize`. This replaces the old `MidCompactSize` (renamed to `MinCompactSize`, default changed from 4 to 2) with a hierarchical power-of-2 scheme. `ResourcePool.Usage` now has `Compact2` through `Compact2048` (11 levels) with independent pools. Both `CompactSize` and `MinCompactSize` are validated to be powers of 2. This reduces the number of snapshots in a snapshot bundle slightly, but also allows increasing the compact size much higher for use in short slot networks.

### 3. Parallel compaction
`CompactSnapshotBundle` is now parallelized into 4 tasks: accounts, slots+selfdestruct, state tries, and storage tries.

### 4. Persistence on every block
`RunCompactJob` no longer gates persistence on `blockNumber % compactSize == 0`. Every block now triggers persistence, which is needed for the hierarchical compaction scheme.

### 5. Auto-adjust write buffer size
Measure the write buffer size used during persistence and adjust the write buffer size to make it optimal on all networks. New `WriteBufferAdjuster` class wraps column batches with `CountingWriteBatch` to track bytes written, then adjusts RocksDB write buffer via `SetWriteBuffer` (16MB–256MB, 1.5x multiplier, 20% hysteresis). Integrated into `RocksDbPersistence`, `PreimageRocksdbPersistence`, and `FlatInTriePersistence`. `IDb.SetWriteBuffer` added as a default interface method; `ColumnDb` implements it via `rocksdb_set_options_cf`.

- [X] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

#### Notes on testing

- [X] Mainnet starts normally.

### DIFF
--- a/src/Nethermind/Nethermind.State.Flat/SnapshotCompactor.cs
+++ b/src/Nethermind/Nethermind.State.Flat/SnapshotCompactor.cs
@@ -3,6 +3,7 @@
 
 using System.Collections.Concurrent;
 using System.Diagnostics;
+using Collections.Pooled;
 using Nethermind.Core;
 using Nethermind.Core.Collections;
 using Nethermind.Core.Crypto;
@@ -157,7 +158,7 @@ public class SnapshotCompactor : ISnapshotCompactor
         // Slots and Selfdestruct
         compactTask.Add(Task.Run(() =>
         {
-            HashSet<Address> addressToClear = new();
+            using PooledSet<Address> addressToClear = new();
 
             for (int i = 0; i < snapshots.Count; i++)
             {
@@ -207,7 +208,7 @@ public class SnapshotCompactor : ISnapshotCompactor
         // Storage tries
         compactTask.Add(Task.Run(() =>
         {
-            HashSet<Hash256AsKey> addressHashToClear = new();
+            using PooledSet<Hash256AsKey> addressHashToClear = new();
 
             for (int i = 0; i < snapshots.Count; i++)
             {
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -12,6 +12,7 @@
     <PackageVersion Include="BenchmarkDotNet.Diagnostics.Windows" Version="0.15.8" />
     <PackageVersion Include="BouncyCastle.Cryptography" Version="2.6.2" />
     <PackageVersion Include="Ckzg.Bindings" Version="2.1.5.1544" />
+    <PackageVersion Include="Collections.Pooled" Version="1.0.82" />
     <PackageVersion Include="CommunityToolkit.HighPerformance" Version="8.4.0" />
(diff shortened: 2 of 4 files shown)


src/Nethermind/Nethermind.State.Flat/SnapshotCompactor.cs
```diff
@@ -142,56 +142,101 @@ public Snapshot CompactSnapshotBundle(SnapshotPooledList snapshots)
...
-                    addressToClear.Add(address);
-                    addressHashToClear.Add(address.Value.ToAccountPath.ToCommitment());
+                    // Clear
+                    foreach (((AddressAsKey Address, UInt256) key, SlotValue? _) in storages)
+                    {
+                        if (addressToClear.Contains(key.Address))
+                        {
+                            storages.Remove(key, out _);
+                        }
+                    }
```
LukaszRozmej, line 191: Can you utilize here `AlternativeLookup` on `storages` to iterate over `addressToClear` and not `storages`?

src/Nethermind/Nethermind.State.Flat/SnapshotCompactor.cs
```diff
@@ -142,56 +142,101 @@ public Snapshot CompactSnapshotBundle(SnapshotPooledList snapshots)
...
+                accounts.AddOrUpdateRange(knownState.Accounts);
+            }
+        }));
 
-            addressToClear.Clear();
-            addressHashToClear.Clear();
+        // Slots and Selfdestruct
+        compactTask.Add(Task.Run(() =>
+        {
+            HashSet<Address> addressToClear = new();
```
LukaszRozmej: PooledSet?

https://github.com/jtmueller/Collections.Pooled

src/Nethermind/Nethermind.State.Flat/SnapshotCompactor.cs
asdacap: No, it need to group by address without the slot.

PR #10568, opened 2026-02-18
Fix memory leak in `TriggerWithCondition` by implementing IDisposable

sashass1315:
`TriggerWithCondition` was subscribing to the inner trigger's `TriggerBlockProduction` event but never unsubscribing or disposing the inner trigger, causing a memory leak. Added `IDisposable` implementation to properly unsubscribe from the event and dispose the inner trigger when `TriggerWithCondition` is disposed.


LukaszRozmej:
Not a problem, no leak in practice

PR #10578, opened 2026-02-19
witness generation optimizations and refactorings

LukaszRozmej:
This targets #9949

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Stateless/Witness.cs
+++ b/src/Nethermind/Nethermind.Consensus/Stateless/Witness.cs
@@ -23,6 +23,8 @@ public class Witness
 
 public static class WitnessExtensions
 {
+    private static readonly HeaderDecoder _decoder = new();
+
     extension(Witness witness)
     {
         public INodeStorage CreateNodeStorage()
@@ -53,11 +55,10 @@ public static class WitnessExtensions
         {
             byte[][] witnessHeaders = witness.Headers;
             ArrayPoolList<BlockHeader> headers = new(witnessHeaders.Length);
-            HeaderDecoder decoder = new();
             foreach (byte[] encodedHeader in witnessHeaders)
             {
                 Rlp.ValueDecoderContext stream = new(encodedHeader);
-                headers.Add(decoder.Decode(ref stream) ?? throw new ArgumentException());
+                headers.Add(_decoder.Decode(ref stream) ?? throw new ArgumentException());
             }
 
             return headers;


src/Nethermind/Nethermind.Consensus/Stateless/Witness.cs
```diff
@@ -5,74 +5,62 @@
...
-    [JsonIgnore]
-    public INodeStorage NodeStorage => _nodeStorage ??= CreateNodeStorage();
-
-    [JsonIgnore]
-    public IKeyValueStoreWithBatching CodeDb => _codeDb ??= CreateCodeDb();
-
-    private INodeStorage CreateNodeStorage()
+public static class WitnessExtensions
+{
+    extension(Witness witness)
```
hudem1, line 28: for understanding, why put these methods as extension and not directly inside the class ?

src/Nethermind/Nethermind.Consensus/Stateless/Witness.cs
```diff
@@ -5,74 +5,62 @@
...
+}
 
-    [JsonIgnore]
-    public IReadOnlyCollection<BlockHeader> DecodedHeaders => _decodedHeaders ??= DecodeHeaders();
-
-    [JsonIgnore]
-    public INodeStorage NodeStorage => _nodeStorage ??= CreateNodeStorage();
-
-    [JsonIgnore]
-    public IKeyValueStoreWithBatching CodeDb => _codeDb ??= CreateCodeDb();
```
hudem1, line 33: With your changes, those 3 methods DecodeHeaders(), CreateNodeStorage() and CreateCodeDb() are called every time stateless execution happens, whereas in that previous solution, they were called only once. I guess in practice, the witness won't even be used right after creation but rather sent as a response to the rpc endpoint, so, it doesn't matter i think. But by curiosity, why change that and call those 3 methods inside stateless execution ?

src/Nethermind/Nethermind.Consensus/Stateless/Witness.cs
```diff
@@ -5,74 +5,62 @@
...
-        List<BlockHeader> headers = new(Headers.Length);
-        HeaderDecoder decoder = new();
-        foreach (var encodedHeader in Headers)
+        public ArrayPoolList<BlockHeader> DecodeHeaders()
         {
-            Rlp.ValueDecoderContext stream = new(encodedHeader);
-            headers.Add(decoder.Decode(ref stream) ?? throw new ArgumentException());
+            byte[][] witnessHeaders = witness.Headers;
+            ArrayPoolList<BlockHeader> headers = new(witnessHeaders.Length);
+            HeaderDecoder decoder = new();
```
hudem1: maybe we could even have the decoder as a static readonly field ?

src/Nethermind/Nethermind.Consensus/Stateless/WitnessGeneratingWorldState.cs
```diff
@@ -78,50 +77,47 @@ public Witness GetWitness(BlockHeader parentHeader)
...
 
-    public Hash256 StateRoot { get => inner.StateRoot; }
+    public Hash256 StateRoot => inner.StateRoot;
 
     public bool IsInScope => inner.IsInScope;
 
     public IWorldStateScopeProvider ScopeProvider => inner.ScopeProvider;
 
-    public byte[] GetCode(Address address)
+    public byte[]? GetCode(Address address)
```
hudem1, line 112: why add the ?
the worldstate.getcode doesn't have this it seems and if does not exist, it'll just return empty [], no ?

src/Nethermind/Nethermind.Consensus/Stateless/Witness.cs
LukaszRozmej: Not really a strict reason, I think it looks better, making WItness more of a DTO.

src/Nethermind/Nethermind.Consensus/Stateless/Witness.cs
LukaszRozmej: I think it is called once per witness?

src/Nethermind/Nethermind.Consensus/Stateless/WitnessGeneratingWorldState.cs
LukaszRozmej: hm... it should have - make it consistent with interface

src/Nethermind/Nethermind.Consensus/Stateless/Witness.cs
hudem1: yes i guess when used it should be used once

PR #10584, opened 2026-02-19
Fix flaky test: Will_Trigger_ReorgBoundaryEvent_On_Prune

smartprogrammer93:
Closes #10583

## Changes

- Added a 50ms delay after  in the flaky test to ensure the  event handler completes before the assertion checks the value.

## Root Cause

The test was flaky due to a race condition:
1. When a block is committed,  event fires in  (line 490) with potentially stale 
2. Then pruning runs asynchronously, and  (line 753) fires the event AGAIN with the correct updated value after pruning updates 
3. Due to timing, sometimes the test assertion ran before the second event's value was available

The fix adds a small buffer delay after  returns to ensure the event handler has completed.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [x] Other: Test fix

#### Requires testing

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

#### Notes on testing

Test was run 10+ times locally without failure.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Trie.Test/Pruning/TreeStoreTests.cs
+++ b/src/Nethermind/Nethermind.Trie.Test/Pruning/TreeStoreTests.cs
@@ -937,11 +937,8 @@ namespace Nethermind.Trie.Test.Pruning
         }
 
         [Test]
-        [Retry(3)]
-        [NonParallelizable]
-        public async Task Will_Trigger_ReorgBoundaryEvent_On_Prune()
+        public void Will_Trigger_ReorgBoundaryEvent_On_Prune()
         {
-            // TODO: Check why slow
             MemDb memDb = new();
 
             using TrieStore fullTrieStore = CreateTrieStore(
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Trie.Test/Pruning/TreeStoreTests.cs
```diff
@@ -971,6 +971,14 @@ public async Task Will_Trigger_ReorgBoundaryEvent_On_Prune()
                 if (i > 4)
                 {
                     fullTrieStore.WaitForPruning();
+
+                    // The ReorgBoundaryReached event fires in two places:
+                    // 1. PushToMainCommitSetQueue (before pruning) - may have stale value
+                    // 2. SaveSnapshot (after pruning updates LastPersistedBlockNumber) - correct value
+                    // The polling assertion should eventually see the correct value from #2,
+                    // but we add a tiny delay to ensure the event handler has completed.
+                    await Task.Delay(50);
```
LukaszRozmej: why this 50ms would make a difference if the assert below has `.After(10000, 100))`? So already checks for 10s?

smartprogrammer93: lol, was some slop. fixed.

**Root cause:** `Prune()` in `FinishBlockCommit` is a no-op when `_pruningTask.IsCompleted == false` (i.e., a previous background pruning task is still running). In that case, `_pruningTask` still points to the old task, so `WaitForPruning()` only waits for *that* task — which may not have processed the current block's commit set at all. The event fires with a stale `LastPersistedBlockNumber` from a previous iteration, and no new pruning is ever triggered to produce the correct value. The `.After()` polling can't help because there's nothing scheduled to update `reorgBoundary`.

**Fix:** Replaced the entire `WaitForPruning()` + `Task.Delay(50)` + polling assertion pattern with `WaitForPruning()` + `SyncPruneQueue()`. `SyncPruneQueue()` (already used by other tests and marked "Testing purpose only") synchronously processes the commit set queue after any background work finishes. This makes the test fully deterministic — no timing dependencies, no polling, no `Retry(3)` needed.

PR #10587, opened 2026-02-20
fix: restore StrictHexFormat in BlockParameterConverterTests

newmanifold:
- `Cant_read_block_number_when_strict_hex_format_is_enabled` sets the static `EthereumJsonSerializer.StrictHexFormat = true` but never restored it, polluting other tests running afterward.
- Replaced with a try/finally block that saves and restores the original value. Otherwise tests fail in Linux ubuntu-24, Dotnet v10.0.103

### DIFF
--- a/src/Nethermind/Nethermind.JsonRpc.Test/Data/BlockParameterConverterTests.cs
+++ b/src/Nethermind/Nethermind.JsonRpc.Test/Data/BlockParameterConverterTests.cs
@@ -16,8 +16,6 @@ namespace Nethermind.JsonRpc.Test.Data
     [TestFixture]
     public class BlockParameterConverterTests : SerializationTestBase
     {
-        [TearDown]
-        public void TearDown() => EthereumJsonSerializer.StrictHexFormat = false;
         [TestCase("0", 0)]
         [TestCase("100", 100)]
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.JsonRpc.Test/Data/BlockParameterConverterTests.cs
```diff
@@ -15,6 +16,8 @@ namespace Nethermind.JsonRpc.Test.Data
     [TestFixture]
     public class BlockParameterConverterTests : SerializationTestBase
     {
+        [TearDown]
+        public void TearDown() => EthereumJsonSerializer.StrictHexFormat = false;
```
LukaszRozmej: should be done only in `Cant_read_block_number_when_strict_hex_format_is_enabled` and to previouys value only?

newmanifold: Done! 
Now that I think about it, with parallel test runs, is the above assignment to the static `EthereumJsonSerializer.StrictHexFormat` is there something that ensures ordering? 
I assume that a DI approach be better here, just like Hash256?

PR #10588, opened 2026-02-20
fix: fix thread safety of List<T> in FilterManager

letmehateu:
## Changes

- Add `lock` around all `List<T>` access points in `FilterManager` to prevent concurrent modification from TxPool event handlers and RPC polling threads
- Make `ToArray()` + `Clear()` atomic in `PollBlockHashes`, `PollLogs`, and `PollPendingTransactionHashes` to prevent data loss between the two calls

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

### DIFF
--- a/src/Nethermind/Nethermind.Facade/Filters/FilterManager.cs
+++ b/src/Nethermind/Nethermind.Facade/Filters/FilterManager.cs
@@ -131,7 +131,7 @@ namespace Nethermind.Blockchain.Filters
                 return [];
             }
 
-            using ArrayPoolListRef<Hash256> result = new(0);
+            using ArrayPoolListRef<Hash256> result = new(blockHashes.Count);
             while (blockHashes.TryDequeue(out Hash256? hash))
             {
                 result.Add(hash);
@@ -145,7 +145,7 @@ namespace Nethermind.Blockchain.Filters
             if (!_logs.TryGetValue(filterId, out ConcurrentQueue<FilterLog> logs))
                 return [];
 
-            using ArrayPoolListRef<FilterLog> result = new(0);
+            using ArrayPoolListRef<FilterLog> result = new(logs.Count);
             while (logs.TryDequeue(out FilterLog? log))
             {
                 result.Add(log);
@@ -159,7 +159,7 @@ namespace Nethermind.Blockchain.Filters
             if (!_pendingTransactions.TryGetValue(filterId, out ConcurrentQueue<Option<Hash256>>? pendingTransactions))
                 return [];
 
-            using ArrayPoolListRef<Hash256> result = new(0);
+            using ArrayPoolListRef<Hash256> result = new(pendingTransactions.Count);
             while (pendingTransactions.TryDequeue(out Option<Hash256>? option))
             {
                 if (!option.IsRemoved)


src/Nethermind/Nethermind.Facade/Filters/FilterManager.cs
```diff
@@ -204,22 +223,25 @@ private void StoreLogs(LogFilter filter, TxReceipt txReceipt, long logIndex, ulo
             }
 
             List<FilterLog> logs = _logs.GetOrAdd(filter.Id, static i => new List<FilterLog>());
-            for (int i = 0; i < txReceipt.Logs.Length; i++)
+            lock (logs)
```
LukaszRozmej: I think better way would be to create `ArrayPoolListRef`, create all the `FilterLog`s, then take the lock and do a quick `AddRange` with a span - minimizing the log time taken while not introducing any new allocations.

letmehateu: > I think better way would be to create `ArrayPoolListRef`, create all the `FilterLog`s, then take the lock and do a quick `AddRange` with a span - minimizing the log time taken while not introducing any new allocations.

Is better now?

src/Nethermind/Nethermind.Facade/Filters/FilterManager.cs
```diff
@@ -74,9 +75,11 @@ private void OnNewPendingTransaction(object sender, TxPool.TxEventArgs e)
...
                 int filterId = filter.Id;
                 List<Hash256> transactions = _pendingTransactions.GetOrAdd(filterId, static _ => new List<Hash256>());
-                transactions.Add(e.Transaction.Hash);
-                if (_logger.IsTrace) _logger.Trace($"Filter with id: {filterId} contains {transactions.Count} transactions.");
-
+                lock (transactions)
+                {
+                    transactions.Add(e.Transaction.Hash);
+                    if (_logger.IsTrace) _logger.Trace($"Filter with id: {filterId} contains {transactions.Count} transactions.");
+                }
```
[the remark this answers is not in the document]
LukaszRozmej: Those list are private it is fine.

src/Nethermind/Nethermind.Facade/Filters/FilterManager.cs
```diff
@@ -88,24 +91,32 @@ private void OnRemovedPendingTransaction(object sender, TxPool.TxEventArgs e)
...
         public FilterLog[] GetLogs(int filterId)
         {
             _filterStore.RefreshFilter(filterId);
-            _logs.TryGetValue(filterId, out List<FilterLog> logs);
-            return logs?.ToArray() ?? [];
+            if (_logs.TryGetValue(filterId, out List<FilterLog> logs))
+            {
+                lock (logs) return logs.ToArray();
+            }
+            return [];
```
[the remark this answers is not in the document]
LukaszRozmej: Yep this one makes sense, please fix

src/Nethermind/Nethermind.Facade/Filters/FilterManager.cs
```diff
@@ -203,23 +223,28 @@ private void StoreLogs(LogFilter filter, TxReceipt txReceipt, long logIndex, ulo
...
                 return;
             }
 
-            if (_logger.IsTrace) _logger.Trace($"Filter with id: {filter.Id} contains {logs.Count} logs.");
+            List<FilterLog> logs = _logs.GetOrAdd(filter.Id, static i => new List<FilterLog>());
+            lock (logs)
+            {
+                logs.AddRange(filteredLogs.AsSpan());
+                if (_logger.IsTrace) _logger.Trace($"Filter with id: {filter.Id} contains {logs.Count} logs.");
+            }
```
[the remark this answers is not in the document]
LukaszRozmej: kind of true too

src/Nethermind/Nethermind.Facade/Filters/FilterManager.cs
```diff
@@ -203,23 +240,30 @@ private void StoreLogs(LogFilter filter, TxReceipt txReceipt, long logIndex, ulo
                 return;
             }
 
+            // Fetch (or create) the list before processing logs so that a concurrent OnFilterRemoved
+            // cannot cause GetOrAdd to resurrect a removed entry in _logs after TryRemove has run.
             List<FilterLog> logs = _logs.GetOrAdd(filter.Id, static i => new List<FilterLog>());
```
LukaszRozmej: Move this lower, just before the lock

src/Nethermind/Nethermind.Facade/Filters/FilterManager.cs
```diff
@@ -17,13 +17,17 @@ namespace Nethermind.Blockchain.Filters
...
-        private readonly ConcurrentDictionary<int, List<Hash256>> _blockHashes =
+        private readonly ConcurrentDictionary<int, ConcurrentQueue<Hash256>> _blockHashes =
             new();
 
-        private readonly ConcurrentDictionary<int, List<Hash256>> _pendingTransactions =
+        private readonly ConcurrentDictionary<int, ConcurrentQueue<Hash256>> _pendingTransactions =
+            new();
+
+        // Tracks transaction hashes marked as removed before the next poll drains the queue.
+        private readonly ConcurrentDictionary<int, ConcurrentDictionary<Hash256, byte>> _removedTransactions =
```
LukaszRozmej: Question here, not sure:

Maybe better to make 
```suggestion
        private readonly ConcurrentDictionary<int, ConcurrentQueue<Option<Hash256>>> _pendingTransactions =
            new();                       
```

Where `Option<T>` is a class, but we can modify it to mark it removed, while not removing the item from the queue.

Then when removing items from the queue we can mark those items in a scan.
Then when going through the queue we can just filter-out removed items.

src/Nethermind/Nethermind.Facade/Filters/FilterManager.cs
```diff
@@ -124,39 +126,39 @@ public Hash256[] PollBlockHashes(int filterId)
...
             }
 
-            var existingBlockHashes = blockHashes.ToArray();
-            _blockHashes[filterId].Clear();
-
-            return existingBlockHashes;
+            List<Hash256> result = new();
+            while (blockHashes.TryDequeue(out Hash256? hash))
+                result.Add(hash);
+            return result.ToArray();
```
LukaszRozmej, line 139: Now this is a good question should we do snapshots? Does it make sense now?
We could optimize it in two ways:
1. Don't do snapshots at all, return the queue, drain the queue directly when serializing the response.
2. Do snapshots but use `ArrayPoolList` to avoid allocating the arrays.

WDYT?

flcl42:
> > Might be safer to use thread safe collections
> 
> Which one? We can make one with `List<T>` and `lock` for simplicity, but the `ConcurrentBag` sucks and I would stay away from it.

Concurrent queue may work indeed

src/Nethermind/Nethermind.Facade/Filters/FilterManager.cs
LukaszRozmej: I think this would be last thing I would kind of insist on, seems simpler

src/Nethermind/Nethermind.Facade/Filters/FilterManager.cs
```diff
@@ -124,39 +131,37 @@ public Hash256[] PollBlockHashes(int filterId)
                 return [];
             }
 
-            var existingBlockHashes = blockHashes.ToArray();
-            _blockHashes[filterId].Clear();
-
-            return existingBlockHashes;
+            List<Hash256> result = new();
+            while (blockHashes.TryDequeue(out Hash256? hash))
+                result.Add(hash);
```
flcl42, line 137: Could you add {} around the statement

src/Nethermind/Nethermind.Facade/Filters/FilterManager.cs
```diff
@@ -124,39 +131,37 @@ public Hash256[] PollBlockHashes(int filterId)
...
                 return [];
-            }
-
-            var existingLogs = logs.ToArray();
-            _logs[filterId].Clear();
 
-            return existingLogs;
+            List<FilterLog> result = new();
+            while (logs.TryDequeue(out FilterLog? log))
+                result.Add(log);
```
flcl42, line 151: And here

src/Nethermind/Nethermind.Facade/Filters/FilterManager.cs
```diff
@@ -124,39 +131,37 @@ public Hash256[] PollBlockHashes(int filterId)
...
+            if (!_pendingTransactions.TryGetValue(filterId, out ConcurrentQueue<Option<Hash256>>? pendingTransactions))
+                return [];
 
-            if (!_pendingTransactions.TryGetValue(filterId, out var pendingTransactions))
+            List<Hash256> result = new();
+            while (pendingTransactions.TryDequeue(out Option<Hash256>? option))
             {
-                return [];
+                if (!option.IsRemoved)
+                    result.Add(option.Value);
```
flcl42, line 167: And here

letmehateu: Done, thanks

src/Nethermind/Nethermind.Facade/Filters/FilterManager.cs
```diff
@@ -124,39 +131,43 @@ public Hash256[] PollBlockHashes(int filterId)
...
-            {
+            if (!_logs.TryGetValue(filterId, out ConcurrentQueue<FilterLog> logs))
                 return [];
-            }
-
-            var existingLogs = logs.ToArray();
-            _logs[filterId].Clear();
 
-            return existingLogs;
+            List<FilterLog> result = new();
```
LukaszRozmej: Should be `ArrayPoolListRef` instead of List

src/Nethermind/Nethermind.Facade/Filters/FilterManager.cs
```diff
@@ -124,39 +131,43 @@ public Hash256[] PollBlockHashes(int filterId)
                 return [];
             }
 
-            var existingBlockHashes = blockHashes.ToArray();
-            _blockHashes[filterId].Clear();
-
-            return existingBlockHashes;
+            List<Hash256> result = new();
```
LukaszRozmej: Should be `ArrayPoolListRef` instead of List

src/Nethermind/Nethermind.Facade/Filters/FilterManager.cs
```diff
@@ -124,39 +131,43 @@ public Hash256[] PollBlockHashes(int filterId)
...
         }
 
         public Hash256[] PollPendingTransactionHashes(int filterId)
         {
             _filterStore.RefreshFilter(filterId);
+            if (!_pendingTransactions.TryGetValue(filterId, out ConcurrentQueue<Option<Hash256>>? pendingTransactions))
+                return [];
 
-            if (!_pendingTransactions.TryGetValue(filterId, out var pendingTransactions))
+            List<Hash256> result = new();
```
LukaszRozmej: Should be `ArrayPoolListRef` instead of List

src/Nethermind/Nethermind.Facade/Filters/FilterManager.cs
```diff
@@ -87,32 +87,39 @@ private void OnRemovedPendingTransaction(object sender, TxPool.TxEventArgs e)
...
+                    continue;
 
+                // Scan the queue and mark the matching item as removed.
+                foreach (Option<Hash256> option in transactions)
+                {
+                    if (!option.IsRemoved && option.Value == e.Transaction.Hash)
+                    {
+                        option.MarkRemoved();
+                        if (_logger.IsTrace) _logger.Trace($"Filter with id: {filterId}: transaction {e.Transaction.Hash} marked as removed.");
+                        break;
```
LukaszRozmej: Can't we have duplicates in this queue?

src/Nethermind/Nethermind.Facade/Filters/FilterManager.cs
```diff
@@ -124,39 +131,43 @@ public Hash256[] PollBlockHashes(int filterId)
...
-            {
+            if (!_logs.TryGetValue(filterId, out ConcurrentQueue<FilterLog> logs))
                 return [];
-            }
-
-            var existingLogs = logs.ToArray();
-            _logs[filterId].Clear();
 
-            return existingLogs;
+            using ArrayPoolListRef<FilterLog> result = new(0);
```
LukaszRozmej: initialie them with queue count first

PR #10595, opened 2026-02-20
Resolve merge conflict with master

kamilchodola:
## Summary
- Merges latest `master` into `fix/large-rpc-buffers` (PR #9411) and resolves the conflict in `Startup.cs`
- Master refactored the inline request processing block into `ProcessJsonRpcRequestCoreAsync`. The `CountingStreamPipeWriter(ctx.Response.Body)` change from this PR is now applied inside that extracted method instead of in the old inline location.

## Test plan
- [ ] Verify JSON-RPC responses work correctly with the stream-based writer
- [ ] CI passes


MarekM25:
wtf 74k line PR?

PR #10599, opened 2026-02-21
refactor(trie): remove dead code in TrieNode.Decoder

aso20455:
Remove unreachable `|| data is null` in else branches where data was already checked for null.


LukaszRozmej:
_nullNode is not the same as null

PR #10591, opened 2026-02-20
Perf/single gas spec calculations

LukaszRozmej:
## Changes

- Introduce `SpecGasCosts` that stores calculations in `ReleaseSpec`

#### What types of changes does your code introduce?

- [x] Optimization

#### Requires testing

- [x] Yes

#### Notes on testing

Current tests should cover it

PR #10601, opened 2026-02-21
fix: lazy-initialize GasCosts in OverridableReleaseSpec and SpecProviderSubstitute

MarekM25:
## Changes

- `OverridableReleaseSpec.GasCosts`: was allocating `new SpecGasCosts(this)` on every property access. Added a backing field with `??=` lazy init, matching `ReleaseSpec` pattern.
- `SpecProviderSubstitute.GasCosts`: was using `Returns(_ => new SpecGasCosts(spec))` — the lambda caused NSubstitute to invoke the factory on every access. Changed to `Returns(new SpecGasCosts(spec))` to return a single cached instance, matching `ReleaseSpecSubstitute`.

Both changes follow the pattern established in `ReleaseSpec.cs`:
```csharp
private SpecGasCosts? _gasCosts;
SpecGasCosts IReleaseSpec.GasCosts => _gasCosts ??= new SpecGasCosts(this);
```

Depends on #10591. Can be retargeted to `master` after that merges.

## Type of change

## Testing

- [ ] Requires testing
- [ ] Requires no testing

## Documentation

- [ ] Requires documentation update
- [ ] Does not require documentation update

### DIFF
--- a/src/Nethermind/Nethermind.Core.Test/SpecProviderSubstitute.cs
+++ b/src/Nethermind/Nethermind.Core.Test/SpecProviderSubstitute.cs
@@ -14,7 +14,7 @@ public static class SpecProviderSubstitute
         IReleaseSpec genesis = ReleaseSpecSubstitute.Create();
         sub.GenesisSpec.Returns(genesis);
         IReleaseSpec spec = sub.GetSpec(Arg.Any<ForkActivation>());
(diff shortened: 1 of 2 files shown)


MarekM25:
@LukaszRozmej I'm experimenting, ignore for now

PR #10602, opened 2026-02-21
chore: add code review instructions for GitHub Copilot

smartprogrammer93:


LukaszRozmej:
Can we make something agent-agnostic or does it have to be tailored to copilot?

smartprogrammer93:
@LukaszRozmej it has to be tailored. Instructions here: https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions

PR #10604, opened 2026-02-21
fix(txpool): fix stale pending transaction snapshot after head update

bobtajson:
## Changes

- Mark `_transactionSnapshot` and `_blobTransactionSnapshot` fields as `volatile` so lock-free readers (`GetPendingTransactions`, `AddPeer`) always observe null-writes performed by other threads
 - Move snapshot clearing in `ProcessNewHeadLoop` from the end of the `try` block into the `finally` block, immediately before `ExitWriteLock()`

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_


src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -315,6 +311,12 @@ private async Task ProcessNewHeadLoop()
                     }
                     finally
                     {
+                        // Snapshot must be cleared inside the write lock so readers cannot
```
dipkakwani, line 314: Can you please add new regression test, since this issue was not caught by our existing tests? Maybe one simple test to check the contents of `GetPendingTransactions` after a head change and, if possible, another one for multiple concurrent `GetPendingTransactions` during head processing?

src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -315,6 +311,12 @@ private async Task ProcessNewHeadLoop()
...
+                        // regenerate it from a partially-updated _transactions collection.
+                        // Placed in finally to guarantee clearing even if an exception occurs
+                        // mid-update (otherwise readers could see a stale snapshot).
+                        _transactionSnapshot = null;
```
dipkakwani, line 318: We might also need to change the logging here to ensure it does not throw Null Reference Exception due to the new volatile type? https://github.com/NethermindEth/nethermind/blob/9a27f7adcd002e23940ef7da878e81968968ceee/src/Nethermind/Nethermind.TxPool/TxPool.cs#L465

dipkakwani (changes requested):

src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.cs
```diff
@@ -2467,6 +2467,101 @@ public async Task should_bring_back_reorganized_txs()
...
+                Volatile.Write(ref stopReading, true);
+                await Task.WhenAll(readers);
+
+                // After head processing completes, snapshot must be up-to-date
+                Transaction[] snapshot = _txPool.GetPendingTransactions();
+                snapshot.Length.Should().Be(1);
+                snapshot.Should().Contain(t => t.Hash == txB.Hash);
+                snapshot.Should().NotContain(t => t.Hash == txA.Hash);
+
+                // Re-create test state for the next attempt
```
dipkakwani, line 2461: Need to dispose the txpool before the next iteration:
```
await _txPool.DisposeAsync();
```

dipkakwani (approved):
LGTM now. Also verified that the newly added concurrent test fail with the current master branch and pass with the fix.

src/Nethermind/Nethermind.TxPool.Test/TxPoolTests.cs
bobtajson: > Need to dispose the txpool before the next iteration:
> 
> ```
> await _txPool.DisposeAsync();
> ```

Thank you, done

dipkakwani (approved):

newmanifold:
@claude review

LukaszRozmej (approved):

benaadams (approved):

LukaszRozmej:
please solve merge conflict

LukaszRozmej:
@claude review

LukaszRozmej:
@claude re-review

benaadams (approved):

PR #10610, opened 2026-02-22
#10503: no-alloc packet type writing

alexb5dh:
For #10503:
- Support for long/ulong RLP encoding to pre-allocated buffer
- More RLP long/ulong test cases
- Use no-alloc version for writing packet type during serialization

### DIFF
--- a/src/Nethermind/Nethermind.Core.Test/RlpTests.cs
+++ b/src/Nethermind/Nethermind.Core.Test/RlpTests.cs
@@ -62,7 +62,7 @@ namespace Nethermind.Core.Test
         }
 
         [Test]
-        public void Length_of_long_same_as_uint256([ValueSource(nameof(LongValues))] long value)
+        public void Length_of_ulong_same_as_uint256([ValueSource(nameof(ULongValues))] ulong value)
         {
             Assert.That(Rlp.LengthOf(value), Is.EqualTo(Rlp.LengthOf((UInt256)value)));
         }


src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
```diff
@@ -403,6 +403,48 @@ public static Rlp Encode(int value)
...
+                case < 0x1_0000_0000_0000:
+                    buffer[0] = 134; buffer[1] = (byte)(value >> 40); buffer[2] = (byte)(value >> 32); buffer[3] = (byte)(value >> 24); buffer[4] = (byte)(value >> 16); buffer[5] = (byte)(value >> 8); buffer[6] = (byte)value;
+                    return buffer[..7];
+                case < 0x100_0000_0000_0000:
+                    buffer[0] = 135; buffer[1] = (byte)(value >> 48); buffer[2] = (byte)(value >> 40); buffer[3] = (byte)(value >> 32); buffer[4] = (byte)(value >> 24); buffer[5] = (byte)(value >> 16); buffer[6] = (byte)(value >> 8); buffer[7] = (byte)value;
+                    return buffer[..8];
+                default:
+                    buffer[0] = 136; buffer[1] = (byte)(value >> 56); buffer[2] = (byte)(value >> 48); buffer[3] = (byte)(value >> 40); buffer[4] = (byte)(value >> 32); buffer[5] = (byte)(value >> 24); buffer[6] = (byte)(value >> 16); buffer[7] = (byte)(value >> 8); buffer[8] = (byte)value;
+                    return buffer[..9];
+            }
```
alexb5dh, line 448: Looks ugly, but should perform the same as non-buffered version (exluding allocation).

Alternative approach - less code, but may use a bit more CPU ops:
```csharp
public static Span<byte> Encode(ulong value, Span<byte> buffer)
{
    if (value == 0) { buffer[0] = 0x80; return buffer[..1]; }
    if (value < 0x80) { buffer[0] = (byte)value; return buffer[..1]; }

    int byteCount = 0;
    while (value > 0) { buffer[1 + byteCount++] = (byte)value; value >>= 8; }
    buffer[1..(1 + byteCount)].Reverse();
    buffer[0] = (byte)(0x80 + byteCount);
    return buffer[..(1 + byteCount)];
}
```
or even cleaner one, but always requires 9+ bytes array:
```csharp
// caller is responsible for allocating 9+ bytes array
public static int Encode(ulong value, Span<byte> buffer)
{
    int lz = BitOperations.LeadingZeroCount(value);
    int byteCount = sizeof(ulong) - lz / 8;

    if (byteCount == 0) { buffer[0] = 0x80; return 1; }
    if (value < 0x80)   { buffer[0] = (byte)value; return 1; }

    buffer[0] = (byte)(0x80 + byteCount);
    BinaryPrimitives.WriteUInt64BigEndian(buffer[1..], value << ((sizeof(ulong) - byteCount) * 8));
    return 1 + byteCount;
}
```

src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
```diff
@@ -403,6 +403,48 @@ public static Rlp Encode(int value)
...
+        }
+
+        // caller is responsible for allocating buffer large enough (max 9 bytes)
+        public static Span<byte> Encode(long value, Span<byte> buffer) => Encode((ulong)value, buffer);
```
alexb5dh: Added tests to confirm that simply casting `long` -> `ulong` will produce the same result as old version for negative numbers (which fallbacks to `BigInteger` in this case).

benaadams (approved):

src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
```diff
@@ -403,6 +403,48 @@ public static Rlp Encode(int value)
             _ => new(new byte[] { 136, (byte)(value >> 56), (byte)(value >> 48), (byte)(value >> 40), (byte)(value >> 32), (byte)(value >> 24), (byte)(value >> 16), (byte)(value >> 8), (byte)value }),
         };
 
+        // caller is responsible for allocating buffer large enough (max 9 bytes)
+        [SuppressMessage("ReSharper", "IntVariableOverflowInUncheckedContext")]
+        public static Span<byte> Encode(ulong value, Span<byte> buffer)
```
alexb5dh, line 408: Should I add `LengthOf` vs `buffer.Length` check? May reduce performance here - current implementation casts `ulong` to `UInt256`.

PR #10621, opened 2026-02-24
Fix XDC EIP-161

batrr:
## Problem

When transactions touch precompiles (e.g. `ecrecover` at 0x01), Nethermind persists empty precompile accounts instead of deleting them per EIP-161, causing state root mismatches with other clients.

## Root Cause

1. User tx touches ecrecover → change is batched.
2. `PayFees` calls `GetCandidateOwnerDuringProcessing(coinbase)` with `Address.SystemUser`.
3. SystemUser → `SystemTransactionProcessor` (shared `WorldState`).
4. `SystemTransactionProcessor` commits with `SystemTransactionReleaseSpec` (`IsEip158Enabled = false`).
5. Commit flushes all batched changes; empty precompile is persisted instead of deleted.

## Previous Temporary Solution

Fetch the owner **before** executing each transaction, when there are no changes from the current tx yet. 
The cached owner is used in `PayFees`, avoiding any system call during tx execution.
This is only a temporary solution that was enough to continue processing blocks on mainnet.

## New Solution

Instead of triggering a system transaction during processing, the owner is now retrieved **directly from the state**.

This removes the need to call `GetCandidateOwnerDuringProcessing` via `SystemTransactionProcessor`.


src/Nethermind/Nethermind.Xdc/XdcTransactionProcessor.cs
```diff
@@ -61,8 +61,7 @@ protected override void PayFees(
             return;
         }
 
-        Address coinbase = header.GasBeneficiary!;
-        Address owner = _masternodeVotingContract.GetCandidateOwner(this, header, coinbase);
+        Address owner = _masternodeVotingContract.GetCandidateOwner(WorldState, header.GasBeneficiary!);
 
         if (owner is null || owner == Address.Zero)
             return;
```
cicr99, line 67: There is this check in the function:
```
if (storageValue.Length != 21 || storageValue[0] != 0x01)
            throw new InvalidOperationException($"Unexpected storage value for validatorsState owner");
```
Is it possible for `isCandidate` to be true while the address is zero?
From the contract logic, I don't believe it is. If that's the case, perhaps this check after calling `GetCandidateOwner` is not actually needed, or we should throw here as well as it would be an inconsistency.

src/Nethermind/Nethermind.Xdc/Contracts/MasternodeVotingContract.cs
```diff
@@ -66,6 +66,23 @@ public Address GetCandidateOwner(ITransactionProcessor transactionProcessor, Blo
...
+        mappingKeyInput.Clear();
+        candidate.Bytes.CopyTo(mappingKeyInput.Slice(12, Address.Size));
+        mappingKeyInput[63] = ValidatorsStateSlot;
+        ValueHash256 slotHash = ValueKeccak.Compute(mappingKeyInput);
+        UInt256 slot = new(slotHash.Bytes, isBigEndian: true);
+
+        StorageCell cell = new(ContractAddress!, slot);
+        ReadOnlySpan<byte> storageValue = worldState.Get(cell);
+        if (storageValue.Length != 21 || storageValue[0] != 0x01)
+            throw new InvalidOperationException($"Unexpected storage value for validatorsState owner");
```
alexb5dh, line 82: Should we really throw here? Maybe return something like `Address.Zero` to be handled by `if` check in the caller? For example in case if candidate has resigned (`resign`) or was deleted in `voteInvalidKYC`, and maybe some other cases?

alexb5dh: Actually, I don't see XDC checking `isCandidate` at all?
https://github.com/XinFinOrg/XDPoSChain/blob/dev-upgrade/core/state/statedb.go#L1223

PR #10625, opened 2026-02-24
fix: update SetL1OriginSignature API to use byte[] and add validation

smartprogrammer93:
Closes https://github.com/taikoxyz/taiko-geth/pull/531

## Changes

- Change L1Origin.Signature from int[] to byte[]
- Add 65-byte validation in taikoAuth_setL1OriginSignature RPC method
- Update RLP encoder/decoder to handle byte[] type
- Update tests to use byte[] fixtures

## Types of changes

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: 

## Testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

## Documentation

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Taiko/L1Origin.cs
+++ b/src/Nethermind/Nethermind.Taiko/L1Origin.cs
@@ -12,7 +12,7 @@ public class L1Origin(UInt256 blockId,
     ValueHash256 l1BlockHash,
     int[]? buildPayloadArgsId,
     bool isForcedInclusion = false,
-    int[]? signature = null)
+    byte[]? signature = null)
 {
     public UInt256 BlockId { get; set; } = blockId;
     public ValueHash256? L2BlockHash { get; set; } = l2BlockHash;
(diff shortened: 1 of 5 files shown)


src/Nethermind/Nethermind.Taiko/L1Origin.cs
```diff
@@ -22,7 +22,7 @@ public class L1Origin(UInt256 blockId,
     // Taiko uses int-like serializer (Go's default encoding for byte arrays)
     public int[]? BuildPayloadArgsId { get; set; } = buildPayloadArgsId;
     public bool IsForcedInclusion { get; set; } = isForcedInclusion;
-    public int[]? Signature { get; set; } = signature;
+    public byte[]? Signature { get; set; } = signature;
```
dipkakwani, line 25: Taiko geth still uses fixed byte array for signature https://github.com/taikoxyz/taiko-geth/blob/3c2eee10ddf0749de9dee3636fb8f9fceca3ecfe/core/rawdb/taiko_l1_origin.go#L47
So modifying the type here could cause some serialization issue?
Taiko-geth's `[65]byte` is serialized to int-array (Go's default encoding, mentioned in the comment above), which is why we had it as int[] instead of byte[] here.

So after this change, NMC will return all signature in L1Origin as hex string, while taiko-geth will continue to return it as int array? Not sure why exactly the integration test pass, and if we now have some special handling for both types.

jmadibekov: @dipkakwani yes, taiko team did update the signature's API to hex in https://github.com/taikoxyz/taiko-geth/pull/531 but forgot to override this default encoding in the struct that is used by gencodec. 

Hence the integration tests are failing here. The fix needs to be on the taiko-geth, I'm working on it right now.

PR #10628, opened 2026-02-24
fix: warmup transactions bypass SystemTransactionProcessor due to Warmup flag

kamilchodola:
## Changes

Fixes a performance regression introduced by #10267 where the pre-warmer executes transactions through the **full heavy `Execute()` path** instead of the lightweight `SystemTransactionProcessor` path.

### Root Cause

PR #10267 added a `Warmup = 8` flag to `ExecutionOptions` and changed the warmup call from:
```
opts = SkipValidation           // value 4
```
to:
```
opts = Warmup | SkipValidation  // value 12
```

The routing logic in `ExecuteCore` used an **equality check**:
```csharp
if (tx.IsSystem() || opts == ExecutionOptions.SkipValidation)  // 12 != 4 → false
```

This broke the routing: warmup transactions no longer matched and fell through to the full `Execute()` path, which performs:
- **BuyGas** — reads/writes sender balance for gas reservation
- **IncrementNonce** — increments sender nonce in state
- **PayFees** — distributes gas fees to beneficiary
- **PayRefund** — refunds unused gas to sender
- **PayValue** — transfers transaction value

The `SystemTransactionProcessor` overrides all of these as **no-ops**, making it ~significantly~ cheaper for warmup/system transactions.

The same pattern existed in `SystemTransactionProcessor.Execute` where `opts != ExecutionOptions.SkipValidation` was used to decide whether to add the `OriginalValidate` flag. With the Warmup bit set, this comparison also broke, causing incorrect flag propagation.

### Fix

Mask out the `Warmup` flag before comparing `ExecutionOptions`:

**`TransactionProcessor.cs` (ExecuteCore):**
```csharp
// Before (broken):
if (tx.IsSystem() || opts == ExecutionOptions.SkipValidation)
// After (fixed):
if (tx.IsSystem() || (opts & ~ExecutionOptions.Warmup) == ExecutionOptions.SkipValidation)
```

**`SystemTransactionProcessor.cs` (Execute):**
```csharp
// Before (broken):
return base.Execute(tx, tracer, (opts != ExecutionOptions.SkipValidation && ...)
// After (fixed):
ExecutionOptions coreOpts = opts & ~ExecutionOptions.Warmup;
return base.Execute(tx, tracer, (coreOpts != ExecutionOptions.SkipValidation && ...)
```

### Impact

During block processing, the pre-warmer runs transactions ahead of the main thread to warm state caches. Without this fix, every warmup transaction performs unnecessary state mutations (gas accounting, nonce increment, fee distribution) that:
1. **Waste CPU** on operations that will be discarded
2. **Create unnecessary state writes** in the warmer thread
3. **Defeat the purpose** of using `SystemTransactionProcessor` for lightweight execution

On benchmarks running 10k+ sequential blocks, this compounds into a measurable performance regression.

## Type of change

## Testing

Added 3 regression tests to `TransactionProcessorTests`:
- `Warmup_does_not_update_SpentGas` — verifies warmup doesn't write to `tx.SpentGas`
- `Warmup_does_not_modify_sender_nonce` — verifies warmup doesn't increment sender nonce
- `Warmup_does_not_deduct_sender_balance` — verifies warmup doesn't deduct gas/value from sender

All 6 test runs pass (3 tests × 2 EIP-155 fixture variants).


benaadams (approved):

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -165,7 +165,7 @@ public virtual TransactionResult Warmup(Transaction transaction, ITxTracer txTra
         private TransactionResult ExecuteCore(Transaction tx, ITxTracer tracer, ExecutionOptions opts)
         {
             if (Logger.IsTrace) Logger.Trace($"Executing tx {tx.Hash}");
-            if (tx.IsSystem() || opts == ExecutionOptions.SkipValidation)
+            if (tx.IsSystem() || (opts & ~ExecutionOptions.Warmup) == ExecutionOptions.SkipValidation)
```
LukaszRozmej:
```suggestion
            if (tx.IsSystem() || (opts & ExecutionOptions.SkipValidation) == ExecutionOptions.SkipValidation)
```

LukaszRozmej (commented):
Shouldn't we check `ExecutionOptions.SkipValidation` as a flag? (Can use `HasFlag` or `&` -whatever)

src/Nethermind/Nethermind.Evm/TransactionProcessing/SystemTransactionProcessor.cs
```diff
@@ -42,7 +42,8 @@ protected override TransactionResult Execute(Transaction tx, ITxTracer tracer, E
             WorldState.CreateAccountIfNotExists(Address.SystemUser, UInt256.Zero, UInt256.Zero);
         }
 
-        return base.Execute(tx, tracer, (opts != ExecutionOptions.SkipValidation && !opts.HasFlag(ExecutionOptions.SkipValidationAndCommit))
+        ExecutionOptions coreOpts = opts & ~ExecutionOptions.Warmup;
+        return base.Execute(tx, tracer, (coreOpts != ExecutionOptions.SkipValidation && !coreOpts.HasFlag(ExecutionOptions.SkipValidationAndCommit))
```
LukaszRozmej:
```suggestion
        return base.Execute(tx, tracer, ((coreOpts & ExecutionOptions.SkipValidation) != ExecutionOptions.SkipValidation && !coreOpts.HasFlag(ExecutionOptions.SkipValidationAndCommit))
```

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
smartprogrammer93: according to claude:

> CommitAndRestore = Commit | Restore | SkipValidation = 7. With HasFlag(SkipValidation): (7 & 4) == 4 → TRUE → routes CallAndRestore to SystemTransactionProcessor.

which is not a desired outcome. same with 

> SkipValidationAndCommit = 5. (5 & 4) == 4 → TRUE → routes Trace() to SystemTransactionProcessor. Also a behavioral change.

smartprogrammer93:
## Benchmarks Against `master`

| Method                      | Master   | Fix      | Change        |
|----------------------------|----------|----------|--------------|
| Warmup_SimpleTx            | 3,239 ns | 1,009 ns | -69% faster  |
| Warmup_AccessListTx        | 3,435 ns | 1,154 ns | -66% faster  |
| Warmup_Eip1559Tx           | 3,128 ns | 1,043 ns | -67% faster  |
| Warmup_ContractCall        | 3,265 ns | 994 ns   | -70% faster  |
| CallAndRestore_SimpleTx    | 3,186 ns | 3,000 ns | -6% (noise)  |
| CallAndRestore_ContractCall| 3,390 ns | 3,166 ns | -7% (noise)  |

---

## Allocations

| Method               | Master  | Fix     | Change |
|----------------------|---------|---------|--------|
| Warmup_SimpleTx      | 4.95 KB | 1.73 KB | -65%   |
| Warmup_AccessListTx  | 5.04 KB | 1.81 KB | -64%   |
| Warmup_Eip1559Tx     | 4.95 KB | 1.74 KB | -65%   |
| Warmup_ContractCall  | 4.59 KB | 1.48 KB | -68%   |

---

## Detailed Fix Stats

| Method               | Mean     | Error   | StdDev   | Min     | Max     | Median  | P90      | P95      |
|----------------------|----------|---------|----------|---------|---------|---------|----------|----------|
| Warmup_SimpleTx      | 1,009 ns | 66.6 ns | 99.7 ns  | 892 ns  | 1,204 ns| 965 ns  | 1,152 ns | 1,185 ns |
| Warmup_AccessListTx  | 1,154 ns | 31.1 ns | 44.6 ns  | 1,070 ns| 1,251 ns| 1,158 ns| 1,208 ns | 1,226 ns |
| Warmup_Eip1559Tx     | 1,043 ns | 90.6 ns | 127.0 ns | 925 ns  | 1,296 ns| 972 ns  | 1,241 ns | 1,256 ns |
| Warmup_ContractCall  | 994 ns   | 73.3 ns | 105.2 ns | 884 ns  | 1,231 ns| 989 ns  | 1,132 ns | 1,196 ns |

asdacap (approved):

kamilchodola:
@claude This PR brings significant performance imporvement but it also increased the "Get" RocksDb method OwnTime per dottrace by 25k ms - making it not super optimal solution - any ideas how to make it even better?

PR #10629, opened 2026-02-24
feat(zkevm): Refactor stateless executor

### LINKED ISSUE #10618: Add BlockValidator to StatelessExecutor

### LINKED ISSUE #10630: Implement precompiles for Zisk
- [ ] `BLAKE2F`
- [ ] `BLS12_G1ADD`
- [ ] `BLS12_G1MSM`
- [ ] `BLS12_G2ADD`
- [ ] `BLS12_G2MSM`
- [ ] `BLS12_MAP_FP2_TO_G2`
- [ ] `BLS12_MAP_FP_TO_G1`
- [ ] `BLS12_PAIRING_CHECK`
- [ ] `BN254_ADD`
- [ ] `BN254_MUL`
- [x] `BN254_PAIRING` (unstable on our end)
- [ ] `ECREC`
- [ ] `ID`
- [ ] `KZG_POINT_EVALUATION` (seems like missing in Zisk)
- [ ] `MODEXP`
- [ ] `RIPEMD160`
- [ ] `SHA256`

### LINKED ISSUE #10632: Consider using native Keccak in Zisk
Currently, we use our own implementation of Keccak-256, while Zisk provides its own, likely a faster one.

### LINKED ISSUE #10633: Implement input serialization for Zisk
Implement input data serialization for Zisk guest that includes states, block, witness, config, etc.

rubo:
- Resolves #10633
- Resolves #10618
- Resolves #10630
- Resolves #10632

## Changes

- Introduced two projects: the `Nethermind.Stateless.ZiskGuest` as a guest for Zisk, and the `Nethermind.Stateless.Executor` with the core execution and serialization logic. The idea is to have a separate guest app for each supported zkVM that contains all host-specific functionality. The stateless executor itself is a regular library that can be compiled and run on any platform.
- Implemented input serialization.
- Added block validation.
- Refactored and optimized Zisk I/O API.
- Added Zisk cryptographic API bindings for precompiles, hashing, etc.
    - Precompiles are implemented as 3 partial classes to avoid runtime costs and are conditionally compiled. See below.
- Conditional compilation is implemented based on a file or directory name: `std` for standard implementations and `zkevm` for zkEVM implementations. The `.std.cs` files and `std/*.cs` files are excluded from build when `EnableZkEvm = true`, while the `.zkevm.cs` files and `zkevm/*.cs` files are excluded from build when `EnableZkEvm != true`. This vastly eliminates the need for `#if ZK_EVM` directives in the codebase.
- Added a tool to generate input files for Zisk.
- Added testing workflow with Mainnet blocks.
- Removed the old `StatelessExecution` project.

See the [README](https://github.com/NethermindEth/nethermind/blob/feature/stateless-zisk/src/Nethermind/Nethermind.Stateless.ZiskGuest/README.md)

- [x] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

Requires manual testing

### DIFF
--- a/src/Nethermind/Stateless.slnx
+++ b/src/Nethermind/Stateless.slnx
@@ -1,6 +1,7 @@
 <Solution>
   <Folder Name="/Nethermind/">
     <Project Path="Nethermind.Abi/Nethermind.Abi.csproj" />
+    <Project Path="Nethermind.Analyzers/Nethermind.Analyzers.csproj" />
     <Project Path="Nethermind.Blockchain/Nethermind.Blockchain.csproj" />
     <Project Path="Nethermind.Config/Nethermind.Config.csproj" />
     <Project Path="Nethermind.Consensus/Nethermind.Consensus.csproj" />


tools/Stateless/ZiskGuest/Zisk/IO.cs
```diff
@@ -0,0 +1,97 @@
+// SPDX-FileCopyrightText: 2026 Demerzel Solutions Limited
```
maximmenshikov: That is not the right place for it. libziskos is. Let's use the architecture that is already provided in bflat.

This is due to many reasons, most important one is that this file is 100% limited by Zisk version (making nmc inadvertently dependent on Zisk)

src/Nethermind/Nethermind.Stateless.ZiskGuest/Dockerfile.zisk
```diff
@@ -0,0 +1,20 @@
...
+
+RUN curl -L https://raw.githubusercontent.com/0xPolygonHermez/zisk/main/ziskup/ziskup | bash -s -- --nokey
+
+ENV PATH="/root/.zisk/bin:${PATH}"
+
+WORKDIR /n
+
+COPY bin/nethermind .
+
+CMD ["tail", "-f", "/dev/null"]
```
maximmenshikov: Huh?

tools/Stateless/ZiskGuest/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+FROM rust:latest
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+  libomp-dev \
+  libopenmpi-dev \
+  libsodium-dev && \
+  rm -rf /var/lib/apt/lists/*
+
+RUN curl -L https://raw.githubusercontent.com/0xPolygonHermez/zisk/main/ziskup/ziskup | bash -s -- --nokey
```
maximmenshikov: It is a bit fragile and insecure to run the script like this.

tools/Stateless/ZiskGuest/Makefile
```diff
@@ -0,0 +1,97 @@
...
+	dotnet build -c release -p:EnableZkvm=true
+
+build-zisk: build-docker-bflat
+	$(DOCKER_RUN) $(BFLAT_BUILD) --libc zisk $(BFLAT_REFS) $(BFLAT_SRC) \
+    --extlib https://github.com/NethermindEth/bflat-libziskos:pre-develop-0.16.0
+	mkdir -p bin; mv -f Program.patched bin/nethermind; rm -f Program Program.o
+
+build-zisk-sim: build-docker-bflat
+	$(DOCKER_RUN) $(BFLAT_BUILD) --libc zisk_sim $(BFLAT_REFS) $(BFLAT_SRC)
+	mkdir -p bin; mv -f Program bin/nethermind; rm -f Program.o
```
maximmenshikov: suggesting abs paths here and in other places

tools/Stateless/ZiskGuest/Zisk/IO.cs
rubo: The point of this particular project is to be dependent on a particular zkVM, Zisk in this case. It's way more convenient to have everything here than to have it scattered across different repos.

maximmenshikov: You can't have it working properly from here. Realistic precompiles need supplementary assembly because most syscalls can't be invoked using C#. That's what bflat handles through bflat external libraries.

Also, I don't see any point in remaking it again and again. We moved it from bflat to a separate repo to simplify management and because we discussed it.
Here, we move it to Nethermind without discussion.

rubo: > Here, we move it to Nethermind without discussion.

What are we doing now if not discussing?

> You can't have it working properly from here.

I don't see any reason why they can't work from here. But the real problem is that they don't work properly from anywhere at the moment, sadly.

maximmenshikov: Alright, discussing.

I said why. You can't implement system calls in C#.
It worked for Zisk because you have a complementary library that does it for you (and it brings you problems, in fact).

rubo: The guest app doesn't know whether the call is a syscall or whatever. It just calls a function in a linked library. A regular .NET p/invoke. What the native library does under the hood doesn't concern the guest. This said, we extracted Zisk functionality from the bflat-riscv64 repo to keep it generic and not be tied to a specific zkVM. The point of the bflat-libziskos repo is to provide a correctly built _native_ library for bflat to link. That's it. All the C# code is better off in a single place for quick changes, testing, and execution. Also, I believe it should be more convenient for you, too. You focus on native code/compiler stuff, while the C# stuff is left for others. This is my understanding of how things should be organized.

maximmenshikov: libziskos is a single point of responsibility for communication with ziskos and it is the most logical solution as native and managed parts are not splittable from each other.

libziskos builds ziskos itself, adds native layer if needed (and it _is_ needed for real precompiles, not ziskos library functions), it may technically provide an entrypoint for early initialization (likely we will need it), do wrappers, whatever.

Having it all in different modules would be a major source of errors.

maximmenshikov: libziskos is fully open for you to modify and it is not dependent on me, as I stressed multiple times.

rubo: I'm not convinced that it's the way to go, but I'd put this conversation on hold until we get working precompiles and see how we go from there. Thanks.

maximmenshikov: We have _working_ precompiles in both meanings - real "syscalls" and library calls. We have problems with some and it is the other problem, unrelated to the actual mechanism, and we will fix it.

Therefore, I don't see any benefit in making incomplete and unused implementation here in first place, please remove it.

tools/Stateless/ZiskGuest/Zisk/IO.cs
```diff
@@ -0,0 +1,97 @@
...
+    {
+        byte* sizePtr = (byte*)(INPUT_ADDR + 8);
+        ulong size = 0;
+
+        for (int i = 0; i < 8; i++)
+            size |= ((ulong)sizePtr[i]) << (i * 8);
+
+        if (size == 0) return [];
+
+        byte[] result = new byte[size];
```
LukaszRozmej: I know this might be kept in `bflat-libziskos`, but something that we can port there:
We probably don't need to allocate here, we can just use `ReadOnlySpan<byte>` with a pointer!

tools/Stateless/ZiskGuest/Zisk/IO.cs
```diff
@@ -0,0 +1,97 @@
...
+        *(baseAddr + 1 + id) = value;
+    }
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static void WriteChar(char c)
+    {
+        *(byte*)UART_ADDR = (byte)c;
+    }
+
+    public static void WriteString(string s)
```
LukaszRozmej:
```suggestion
    public static void WriteString(ReadOnlySpan<char> s)
```

tools/Stateless/ZiskGuest/Zisk/IO.cs
```diff
@@ -0,0 +1,97 @@
...
+    {
+        for (int i = 0; i < s.Length; i++)
+            WriteChar(s[i]);
```
LukaszRozmej: If we need to write a lot - (which we probably don't?) - this could be speedup:
1. By casting to 64bit items and write 64bit per instruction (with bounds checks) - 4x speedup from 16bits, 8x speedup if we would write in UTF-8 - btw good question what encoding the VM expects.
2. If Vectors are supported (probably not?) - we could go to even 128, 256 or 512 bits in one go.

Do we even need to write strings?

tools/Stateless/ZiskGuest/Zisk/IO.cs
```diff
@@ -0,0 +1,97 @@
...
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static uint ReadUInt32(byte[] data, int offset = 0)
+    {
+        if (data.Length < offset + 4) return 0;
+
+        uint result = 0;
+        for (int i = 0; i < 4; i++)
+            result |= ((uint)data[offset + i]) << (i * 8);
+        return result;
+    }
```
LukaszRozmej: Probably can be done better with `Unsafe`

tools/Stateless/ZiskGuest/Zisk/IO.cs
rubo: I just copied it from original repo. I plan to improve it but later. Things are fragile now, while this code works. Once we stabilize our stuff, we can experiment with better approaches here. Otherwise, we risk breaking there where it works.

tools/Stateless/ZiskGuest/Zisk/IO.cs
rubo: This is another advantage of having this code here for now. More people review, more changes/improvements suggested.

src/Nethermind/Nethermind.Stateless.Executor/InputSerializer.cs
```diff
@@ -0,0 +1,182 @@
...
+
+        var offset = 0;
+        var chainId = ReadUInt32(input, ref offset);
+        var blockLength = ReadInt32(input, ref offset);
+
+        IRlpValueDecoder<Block> blockDecoder = Rlp.GetValueDecoder<Block>()!; // cannot be null
+        Rlp.ValueDecoderContext blockContext = new(input.Slice(offset, blockLength));
+        Block block = blockDecoder.Decode(ref blockContext, RlpBehaviors.None);
+        blockContext.Check(blockLength);
+        offset += blockLength;
```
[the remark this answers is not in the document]
rubo: If the `blockLength` or any other value is wrong, deserialization throws in the end.

src/Nethermind/Nethermind.Stateless.Executor/InputSerializer.cs
```diff
@@ -0,0 +1,182 @@
...
+        ArrayPoolList<byte[]> output = new(count, count);
+
+        for (int i = 0; i < count; i++)
+        {
+            var len = ReadInt32(source, ref offset);
+
+            output[i] = source.Slice(offset, len).ToArray();
+            offset += len;
+        }
+
```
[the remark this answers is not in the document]
rubo: We need to keep it lightweight. So if something goes out of sync, it throws in the end, comparing offset and input length.

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -57,7 +57,7 @@ public bool TryCalculateBlobBaseFee(BlockHeader header, Transaction transaction,
     public abstract class TransactionProcessorBase<TGasPolicy> : ITransactionProcessor
         where TGasPolicy : struct, IGasPolicy<TGasPolicy>
     {
-        protected EthereumEcdsa Ecdsa { get; }
+        protected IEthereumEcdsa Ecdsa { get; }
```
rubo: @LukaszRozmej I'm not sure how much this may affect actual performance. Though we can have a specific implementation type under conditional compilation.

src/Nethermind/Nethermind.Core/ExecutionRequest/ExecutionRequestExtensions.cs
```diff
@@ -90,4 +87,18 @@ static byte[] FlatEncodeRequests(ExecutionRequest[] requests, int bufferSize, by
             return buffer.ToArray();
         }
     }
+
+#if ZK_EVM
```
LukaszRozmej, line 91: .zkevm file?

nuget.config
```diff
@@ -22,8 +22,8 @@
       <package pattern="Microsoft.Diagnostics.Runtime" />
       <package pattern="Microsoft.Diagnostics.Runtime.*" />
     </packageSource>
-    <!--<packageSource key="nugettest.org">
-      <package pattern="" />
-    </packageSource>-->
+    <packageSource key="nugettest.org">
+      <package pattern="Nethermind.ZiskBindings" />
+    </packageSource>
```
LukaszRozmej, line 27: Should we have stable package to merge to master?

tools/StatelessInputGen/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "profiles": {
```
LukaszRozmej, line 2: Not needed?

tools/StatelessInputGen/OwnedReadOnlyListConverter.cs
```diff
@@ -0,0 +1,94 @@
...
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System.Reflection;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Nethermind.Core.Collections;
+
+namespace Nethermind.StatelessInputGen;
+
+public class OwnedReadOnlyListConverter : JsonConverterFactory
```
LukaszRozmej, line 11: Can we potentially have a separate Witness classes to avoid pooled list in this project?
Or maybe we should just make `IOwnedReadOnlyList<>` equivalent to normal `List` under `ZK_EVM`?
This pooling isn't really netting us anything here right?

src/Nethermind/Nethermind.Stateless.Executor/InputSerializer.cs
```diff
@@ -0,0 +1,181 @@
...
+using Nethermind.Core;
+using Nethermind.Core.Collections;
+using Nethermind.Serialization.Rlp;
+
+namespace Nethermind.Stateless.Execution;
+
+/// <summary>
+/// Provides methods for zkVM input serialization. <see cref="Witness.Keys"/> are ignored.
+/// </summary>
+public static class InputSerializer
```
LukaszRozmej, line 16: Was there any point of using this over standarized `IRlpDecoder`s?

src/Nethermind/Nethermind.Stateless.ZiskGuest/Makefile
```diff
@@ -0,0 +1,105 @@
...
+		--mount type=bind,source="$(GUEST_DIR)",target=$(SRC_DIR) \
+		nethermindeth/bflat-riscv64:ce97bb96fb8a5c1b80d1108e5913208a856bad79 \
+		bflat build --arch riscv64 --os linux --stdlib dotnet --libc zisk \
+		--no-pie \
+		--no-pthread \
+		--no-stacktrace-data \
+		--no-globalization \
+		--nostdlibrefs \
+		$(BFLAT_REFS) \
+		-r $(BIN_DIR)/Nethermind.ZiskBindings.dll \
```
maximmenshikov: please remove this and next line as they are added by extlib

nuget.config
rubo: Not yet. Things are changing, and this package will likely be gone soon.

tools/StatelessInputGen/Properties/launchSettings.json
rubo: If you run locally, you're gonna need to add some params here like RPC endpoint. So added to make users life easier.

src/Nethermind/Nethermind.Stateless.Executor/InputSerializer.cs
rubo: I'm not sure what you mean. Please elaborate.

src/Nethermind/Nethermind.Stateless.ZiskGuest/Program.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Buffers.Binary;
+using Nethermind.Core;
+using Nethermind.Stateless.Execution;
+using Nethermind.ZiskBindings;
+
+namespace Nethermind.Stateless.ZiskGuest;
+
+class Program
+{
+    static void Main()
```
maximmenshikov: I propose to make it int and return 0. As far as I remember the NativeAOT required the real return value. It is quite likely that the return value brought back to NativeAOT might be garbage in that scenario.

benaadams (approved):
Presumably you could also swap out (most of) UInt256 by just changing the opcode registration?  🤔

tools/StatelessInputGen/OwnedReadOnlyListConverter.cs
rubo: Right, that was my initial approach because of the lack of a serializer. Then I decided to have the same class everywhere, but I don't have strong preferences here.

src/Nethermind/Nethermind.Stateless.Executor/InputSerializer.cs
LukaszRozmej: we have a standard how to design RLP serialization/deserialization and interfaces for it

rubo: Please point to it then

PR #10638, opened 2026-02-25
feat: reproducible benchmarks with retrospective and dual payload sets

kamilchodola:
## Summary
- Integrates retrospective multi-image benchmarking into the main workflow
- Adds `run_count` input for reproducibility testing with matrix-based parallel runs
- Runs both **superblocks** and **realblocks** on master push and PR label trigger
- Separates cache keys per payload_set so they don't overwrite each other
- Moves PR comment to a dedicated `report` job that aggregates both payload sets into two comparison tables
- Waits for parallel `publish-docker.yml` on push events instead of triggering a separate build

## Changes
- `.github/workflows/run-expb-reproducible-benchmarks.yml`: Major workflow restructuring

## Type of change
- [x] Non-breaking change (fix or feature that would not break existing functionality)

## Testing
- Workflow validated via YAML parser
- Cache key isolation ensures superblocks/realblocks data remain independent

### DIFF
--- a/.github/workflows/run-expb-reproducible-benchmarks.yml
+++ b/.github/workflows/run-expb-reproducible-benchmarks.yml
@@ -50,9 +50,9 @@ on:
         default: 1
       docker_images:
         description: >-
-          Force-test specific Docker image tags (comma-separated).
+          Force-test specific Docker images (comma-separated full image references including repo).
           Skips docker build; runs each image as a separate benchmark.
-          Example: master-21f1dca,master-9f78bef,flat
+          Example: nethermindeth/nethermind:master-21f1dca,nethermind/nethermind:latest
         required: false
         default: ""
       enable_retrospective:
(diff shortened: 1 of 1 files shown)


.github/workflows/run-expb-reproducible-benchmarks.yml
```diff
@@ -43,6 +43,33 @@ on:
...
+        required: false
+        type: number
+        default: 1
+      docker_images:
+        description: >-
+          Force-test specific Docker image tags (comma-separated).
+          Skips docker build; runs each image as a separate benchmark.
+          Example: master-21f1dca,master-9f78bef,flat
+        required: false
+        default: ""
```
stdevMac, line 57: Maybe worth adding a differentiator here for nethermind/nethermind and nethermindeth/nethermind in case we want to compare with official releases?

kamilchodola: @copilot can you fix this to ensure full docker image with repo needs to be applied?

PR #10643, opened 2026-02-25
Consider another override; scan more

flcl42:
## Changes

- Consider another override
- Scan more
- Account for Skipped txs
- Don't depend on broken json serializer

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_


src/Nethermind/Nethermind.State/ParallelWorldState.cs
```diff
@@ -324,6 +346,18 @@ public void ValidateBlockAccessList(ushort index, long gasRemaining)
...
+                {
+                    // skip generated with no changes (read-only entries from side-effects)
+                    generatedHead = generatedChanges.MoveNext() ? generatedChanges.Current : null;
+
+                    if (generatedHead is not null)
+                    {
+                        generatedReads += generatedHead.Value.Reads;
+                    }
+                    continue;
+                }
```
LukaszRozmej, line 320: duplicated code?

flcl42: Fixed

PR #10661, opened 2026-02-26
XDC Fix new block message

batrr:
When a **new block** is received over P2P (gossiped `NewBlockMessage`), the node can hang.

**Cause:** Block processing eventually calls `TrieStoreScopeProvider.Commit()`. It schedules tasks onto the same event loop thread and then blocks with `Task.WaitAll`, causing a self-deadlock.

Blocks with no transactions don’t trigger trie commits, so they process fine.
To reproduce on a private chain, create a transaction that modifies storage and process it from `NewBlockMessage` gossip.

Flow when a new block is received (stuck at Task.WaitAll):

```
ZeroNettyP2PHandler.ChannelRead0()
  → XdcProtocolHandler.HandleMessage()
  → Eth62ProtocolHandler.Handle()
  → SyncServer.AddNewBlock()
  → SyncServer.SyncBlock()
  → BlockTree.SuggestBlock()
  → … process block …
  → WorldState.CommitTree()
  → TrieStoreScopeProvider.Commit()
  → Task.WaitAll()   ← stuck here
```

### DIFF
--- a/src/Nethermind/Nethermind.Runner/configs/xdc-archive.json
+++ b/src/Nethermind/Nethermind.Runner/configs/xdc-archive.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://raw.githubusercontent.com/NethermindEth/core-scripts/refs/heads/main/schemas/config.json",
   "Init": {
-    "ChainSpecPath": "chainspec/xdc-test.json",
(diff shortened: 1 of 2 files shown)


batrr:
Just ran `joc-mainnet` - it hits the same deadlock.

PR #10665, opened 2026-02-27
Fix flaky ValidateSeal_SignatureIsInvalid_ReturnsFalse test

smartprogrammer93:
## Changes

- Create separate `XdcBlockHeader` instances for each test case in `InvalidSignatureCases()` to eliminate race condition from shared mutable state when test cases run in parallel

The test class is marked `[Parallelizable(ParallelScope.All)]`, so NUnit runs test cases concurrently. All 5 cases shared a single `XdcBlockHeader` instance, and each case mutates `header.Validator` before calling `ValidateSeal` (which also mutates `header.Author`). This caused intermittent `IndexOutOfRangeException` when one thread overwrote `Validator` with a different-length array while another thread was indexing into it.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [ ] No

#### Notes on testing

The fix itself is to the test infrastructure. Verified by running the affected test 5 times in a row with all 5 cases passing consistently.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Xdc.Test/XdcSealValidatorTests.cs
+++ b/src/Nethermind/Nethermind.Xdc.Test/XdcSealValidatorTests.cs
@@ -61,22 +61,27 @@ internal class XdcSealValidatorTests
 
     public static IEnumerable<TestCaseData> InvalidSignatureCases()
     {
-        XdcBlockHeader header = Build.A.XdcBlockHeader().WithBeneficiary(TestItem.AddressA).TestObject;
+        XdcBlockHeader header = Build.A.XdcBlockHeader().TestObject;
+        header.Beneficiary = TestItem.AddressA;
         yield return new TestCaseData(header, new byte[0]);
 
-        header = Build.A.XdcBlockHeader().WithBeneficiary(TestItem.AddressA).TestObject;
+        header = Build.A.XdcBlockHeader().TestObject;
+        header.Beneficiary = TestItem.AddressA;
         yield return new TestCaseData(header, new byte[65]);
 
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Xdc.Test/XdcSealValidatorTests.cs
```diff
@@ -61,17 +61,27 @@ public void ValidateSeal_ValidSeal_ReturnsTrue()
...
     {
-        XdcBlockHeader header =
-            Build.A.XdcBlockHeader()
-            .TestObject;
+        XdcBlockHeader header = Build.A.XdcBlockHeader().TestObject;
         header.Beneficiary = TestItem.AddressA;
         yield return new TestCaseData(header, new byte[0]);
+
+        header = Build.A.XdcBlockHeader().TestObject;
+        header.Beneficiary = TestItem.AddressA;
```
LukaszRozmej:
```suggestion
        header = Build.A.XdcBlockHeader().WithBeneficiary(TestItem.AddressA).TestObject;
```

smartprogrammer93: It is not like this to avoid having cast the blockHeader to xdcBlockHeader

LukaszRozmej: intent is clearer, separate build and use phase

batrr (approved):

PR #10669, opened 2026-02-27
Use 32-byte StorageValue struct instead of byte[] in storage trie

### LINKED ISSUE #8442: Use 32byte structs rather than byte[] in StorageTrie
## Is your feature request related to a problem? Please describe.

As well as allocating and having an indirection a `byte[]` 

Has 24 bytes of header

- Object header: 16 bytes total
   - 8 bytes for the sync block
   - 8 bytes for the method table pointer
- Array length field: 4 bytes (stored as an Int32)
- Padding: Up to 4 bytes added so that the total object size is a multiple of 8 bytes

Then 8 bytes for the pointer

So is already 32 bytes without any data; so might as well use a 32byte struct rather than a 64byte array (32bytes+32bytes); since all values are treated as unique and don't really share `byte[]` for leaf values

## Describe the solution you'd like

Use `using Word = Vector256<byte>` Words instead of `byte[]`; this should reduce GC objects considered and use less memory (since even 1 byte is 33 bytes)

## Describe alternatives you've considered

Leaving it unchanged

smartprogrammer93:
Closes #8442

Trie
image

Flat
image

Yellow is master, Grean is feature branch in both screenshots

## Changes

Replace heap-allocated `byte[]` with a fixed 32-byte `StorageValue` struct (`Vector256<byte>`-backed) across the full EVM → State → Trie storage path, eliminating per-slot allocations and `UInt256` conversions on every SLOAD/SSTORE/TLOAD/TSTORE.

### Core struct enhancements (`StorageValue`)
- Add `AggressiveInlining` on `AsSpan` / `AsReadOnlySpan` for zero-overhead access
- Add `FromBytes(byte[]?)` factory for converting legacy `byte[]` values
- Remove `Pack = 32` from `StructLayout` — fixes alignment propagation to `OriginalValue`, `StorageChangeTrace`, `Nullable<StorageValue>`
- Use `FastHash64For32Bytes` in `GetHashCode` for direct 32-byte hashing (avoids span allocation)
- Add `AggressiveInlining` to constructor
- Split `FromSpanWithoutLeadingZero` into inlined fast path (32-byte case) + `NoInlining` slow path (prevents `stackalloc` from blocking inlining)
- Add static `StorageValue.Equals(in, in)` and `NotEquals(in, in)` methods to avoid 64-byte copies on comparisons

### EVM hot-path optimizations
- Add `PushStorageValue` on `EvmStack` — zero-copy 32-byte push directly from `StorageValue`
- SLOAD/TLOAD: reinterpret stack slot as `StorageValue` via `Unsafe.As`, push without intermediate `byte[]`
- SSTORE/TSTORE: reinterpret stack slot as `StorageValue`, pass directly to state without `UInt256` roundtrip
- Unify `Set` / `SetTransientState` to a single `StorageValue` overload (remove `ReadOnlySpan<byte>` overloads)
- Use `StorageValue.Equals(in, in)` at all SSTORE comparison sites to avoid 64-byte value copies

### State layer changes
- Change `IWorldState.Get` return type from `ReadOnlySpan<byte>` to `StorageValue`
- Change `IWorldState.GetOriginal` return type from `byte[]` to `StorageValue`
- Pass `StorageValue` by `in` across the entire write chain (`IWorldState.Set`, `SetTransientState`, `WorldState`, `WorldStateMetricsDecorator`, `WitnessGeneratingWorldState`, `PartialStorageProviderBase`, `PersistentStorageProvider`)
- Update `PersistentStorageProvider`, `TransientStorageProvider`, `WorldState`, `WorldStateMetricsDecorator`
- Use zero-alloc `DecodeByteArraySpan()` in `StorageTree.GetStorageValue`

### Storage provider internals optimization
- **SoA change tracking**: Replace `List<Change>` with split `ChangeKey[]` + `StorageValue[]` arrays and `_changeCount` position counter — eliminates `List.Clear()` zeroing entirely on reset
- **Reference-free `InternalStorageKey`**: 56-byte value-type key inlining 20-byte address + `UInt256` index with `Pack=1`, avoiding `Address` class references in dictionary entries so `Dictionary.Clear()` skips entry zeroing. Uses two independent `FastHash` calls (index 32B + address 20B combined with multiply-XOR) for better instruction-level parallelism
- **Version-stamped `_originalValues`**: `_commitRound` counter replaces per-transaction `Clear()` (200 calls/block → 1), with reference-free `OriginalValue` struct entries; periodic trim when size exceeds 16K entries
- **Right-sized `_intraBlockCache` capacity (16)**: Dictionary cleared per-transaction (~200x/block), each transaction touches only 2-8 cells; capacity 16 reduces bucket zeroing from ~512 KB/block (at 8192) to ~1 KB/block
- **`ref readonly StorageValue`** in CommitCore loop: avoids 32-byte value copies on non-processed iterations (~75% of loop)
- **`CollectionsMarshal.AsSpan`** for `_keptValueCache` in Restore — avoids list indexer overhead
- **Single `HashSet.Add`** in CommitCore replaces `Contains` + `Add` double-lookup
- **Replace LINQ** in `UpdateRootHashesMultiThread` with explicit `foreach` + `Span.Sort`

### Prewarmer storage cache: eliminate byte[] round-trip
- **Remove `where TValue : class?`** constraint from `SeqlockCache` — enables storing value types directly. Guard `ReferenceEquals` fast-path with `RuntimeHelpers.IsReferenceOrContainsReferences<TValue>()` (JIT-eliminated for value types, preserves existing behavior for reference types)
- **Store `StorageValue` directly** in `SeqlockCache<StorageCell, StorageValue>` instead of `byte[]` — eliminates `ToArray()` allocation on every cache miss and `FromSpanWithoutLeadingZero` conversion on every cache hit
- **Delete `LoadFromTreeStorageBytes`** — factory delegate now points directly to `LoadFromTreeStorage` which already returns `StorageValue`
- Local profiling confirms: `StorageValue::FromSpanWithoutLeadingZero` gone from FlatState hotspots, `List::ToArray` dropped to 0.04% (3 samples out of 7,755 resolved)

### SlotValue consolidation
- Delete `Nethermind.State.Flat/SlotValue.cs` — replaced by `StorageValue` everywhere
- Update all 19 files in `Nethermind.State.Flat` and `Nethermind.State.Flat.Test`

### Caller updates (~65 files)
- Update all storage callers across Blockchain, EVM, State, Trie, JsonRpc, Optimism, Flashbots, Xdc, and test projects

### BlockProcessingBenchmark: FlatState backend variant
- Add `[Params(StateBackend.Trie, StateBackend.FlatState)]` so block processing performance can be compared between both state backends
- FlatState path uses a minimal `BenchmarkFlatDbManager` with snapshot retention and freeze-after-setup semantics
- Update CI comparison script to key benchmarks by `Method(Parameters)` so both backends appear as separate rows

- [x] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


src/Nethermind/Nethermind.Evm.Benchmark/BlockProcessingBenchmark.cs
```diff
@@ -196,7 +241,30 @@ public void GlobalSetup()
...
+            FlatDbConfig flatDbConfig = new() { TrieCacheMemoryBudget = 0 };
+            ResourcePool resourcePool = new(flatDbConfig);
+            TrieNodeCache trieNodeCache = new(flatDbConfig, LimboLogs.Instance);
+            BenchmarkFlatDbManager flatDbManager = new(resourcePool, trieNodeCache);
+            flatDbManagerRef = flatDbManager;
+            FlatScopeProvider flatScopeProvider = new(
+                dbProvider.CodeDb,
+                flatDbManager,
+                flatDbConfig,
+                new NoopTrieWarmer(),
```
asdacap, line 257: Not, the trie warmer can make a big difference.

src/Nethermind/Nethermind.Evm.Benchmark/ProfileRunner.cs
```diff
@@ -0,0 +1,344 @@
...
+using Nethermind.State.Flat.ScopeProvider;
+using Nethermind.Trie.Pruning;
+
+namespace Nethermind.Evm.Benchmark;
+
+/// <summary>
+/// Minimal runner that executes the storage-heavy benchmarks in a tight loop,
+/// suitable for profiling with dotnet-trace.
+/// </summary>
+public static class ProfileRunner
```
asdacap, line 46: What is this, why is it separate from the existing benchmark?

src/Nethermind/Nethermind.State.Flat/ReadOnlySnapshotBundle.cs
```diff
@@ -119,7 +119,7 @@ public int DetermineSelfDestructSnapshotIdx(Address address)
             }
         }
 
-        return value;
+        return (!found || outStorageValue.IsZero) ? null : outStorageValue;
```
asdacap, line 122: Null and zero is different here because of the uncertainty in the verkle/binarytree future eip.

src/Nethermind/Nethermind.State/PartialStorageProviderBase.cs
```diff
@@ -18,10 +22,15 @@ namespace Nethermind.State
...
     internal abstract class PartialStorageProviderBase
     {
-        protected readonly Dictionary<StorageCell, StackList<int>> _intraBlockCache = new();
+        private readonly Dictionary<InternalStorageKey, int> _intraBlockCache = new(16);
+        private StackList<int>?[] _stackPool = new StackList<int>[512];
+        private int _stackCount;
         protected readonly ILogger _logger;
-        protected readonly List<Change> _changes = new(Resettable.StartCapacity);
-        private readonly List<Change> _keptInCache = new();
+        protected ChangeKey[] _changeKeys = new ChangeKey[1_024];
```
asdacap, line 29: The change here is complicated and sensitive enough that it should be put in a separate PR.

src/Nethermind/Nethermind.Evm.Benchmark/BlockProcessingBenchmark.cs
asdacap: *Note

stdevMac:
@smartprogrammer93 stale-PR check: the EXPB run showed no win (P99 regressed), asdacap's review request went unanswered, and the flat slot-value work evolved independently (#11909). OK to close? If there's no reply in the next few days, we'll close this as part of the stale-PR cleanup — it can always be reopened or resubmitted.

smartprogrammer93:
Closing as stale and didnt see enough gain from it. maybe another jab at it from someone else. for me, this one is not worth continuing on.

PR #10670, opened 2026-02-28
fix(benchmark): fix inverted CI throughput bounds in GasColumnProvider

newmanifold:
BDN's confidence interval is measured in **nanoseconds** (time). Since throughput (`MGas/s = gas / nanoseconds`) is inversely proportional to time, the lower throughput bound corresponds to the *upper* time bound and vice versa.

The previous code passed `ci.Lower` (fastest time) to `CalculateMGasThroughput` for the `CI-Lower` column, producing a value *higher* than the mean — i.e. the bounds were swapped relative to their labels.

Example of benchmark run, Notice the Throughput CI-Lower > Throughput CI-Upper:
image

## Changes

- Fix the bound selection: `ci.Upper` for lower throughput, `ci.Lower` for upper throughput
- Extract `CalculateMGasThroughput` and `GetThroughputBound` as `internal static` methods on `GasColumnProvider` for testability
- Add `Nethermind.Precompiles.Benchmark.Test` project with regression test asserting `lower < upper` throughput bound
- Register test project in `Benchmarks.slnx`

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

- Ran `EcRecoverBenchmark` locally and confirmed `CI-Lower < Throughput mean < CI-Upper` after the fix
- Regression test in `Nethermind.Precompiles.Benchmark.Test` fails when the CI inversion is removed

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Precompiles.Benchmark/GasColumnProvider.cs
+++ b/src/Nethermind/Nethermind.Precompiles.Benchmark/GasColumnProvider.cs
@@ -30,11 +30,7 @@ public class GasColumnProvider : IColumnProvider
     /// <summary>
     /// Converts gas consumed and execution time into MGas/s throughput.
     /// </summary>
-    internal static double CalculateMGasThroughput(long gas, double nanoseconds)
-    {
-        double opThroughput = 1_000_000_000.0 / nanoseconds;
-        return gas * opThroughput / 1_000_000.0;
-    }
+    internal static double CalculateMGasThroughput(long gas, double nanoseconds) => gas * 1000.0 / nanoseconds;
 
     /// <summary>
     /// Computes the MGas/s throughput for a given Benchmark Statistics.


src/Nethermind/Nethermind.Precompiles.Benchmark/GasColumnProvider.cs
```diff
@@ -3,26 +3,51 @@
...
     public IEnumerable<IColumn> GetColumns(Summary summary) => Columns;
 
+    /// <summary>
+    /// Converts gas consumed and execution time into MGas/s throughput.
+    /// </summary>
+    internal static double CalculateMGasThroughput(long gas, double nanoseconds)
+    {
+        double opThroughput = 1_000_000_000.0 / nanoseconds;
+        return gas * opThroughput / 1_000_000.0;
+    }
```
LukaszRozmej:
```suggestion
    internal static double CalculateMGasThroughput(long gas, double nanoseconds) => gas * 1000.0 / nanoseconds;
```

PR #10671, opened 2026-02-28
fix: fix pending requests hanging forever on peer disconnect

bobtajson:
## Changes

- Fix `MessageQueue.CompleteAdding` to cancel all pending `TaskCompletionSource` instances when a peer disconnects, preventing sync dispatchers from hanging indefinitely
- Move `_isClosed` check inside the lock in `Send` to eliminate a race condition between `Send` and `CompleteAdding`

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

### DIFF
--- a/src/Nethermind/Nethermind.Network.Test/MessageQueueTests.cs
+++ b/src/Nethermind/Nethermind.Network.Test/MessageQueueTests.cs
@@ -161,6 +161,49 @@ public class MessageQueueTests
         _recordedSends.Count.Should().Be(0);
     }
 
+    [Test]
+    public void CompleteAdding_cancels_current_request()
+    {
+        Request<GetBlockHeadersMessage, IOwnedReadOnlyList<BlockHeader>> request = CreateRequest();
+
+        _queue.Send(request);
+
+        _queue.CompleteAdding();
+
+        request.CompletionSource.Task.IsCanceled.Should().BeTrue();
+    }
+
+    [Test]
+    public void CompleteAdding_cancels_queued_requests()
+    {
+        Request<GetBlockHeadersMessage, IOwnedReadOnlyList<BlockHeader>> request1 = CreateRequest();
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Network/P2P/MessageQueue.cs
```diff
@@ -20,14 +20,15 @@ public class MessageQueue<TMsg, TData>(Action<TMsg> send)
...
-            }
-
             lock (_lock)
             {
+                if (_isClosed)
+                {
+                    request.Message.TryDispose();
+                    request.CompletionSource.TrySetCanceled();
+                    return;
+                }
```
LukaszRozmej, line 30: why under lock? Ah probably so when it is closed we will dispose this message.

src/Nethermind/Nethermind.Network/P2P/MessageQueue.cs
```diff
@@ -66,7 +67,23 @@ public void Handle(TData data, long size)
...
+                    _currentRequest.CompletionSource.TrySetCanceled();
+                    _currentRequest = null;
+                }
+
+                while (_requestQueue.TryDequeue(out Request<TMsg, TData>? request))
+                {
+                    request.Message.TryDispose();
+                    request.CompletionSource.TrySetCanceled();
+                }
+            }
```
LukaszRozmej, line 86: Maybe move `Queue`->`ConcurrentQueue` and `_currentRequest` -> `Interlocked` and we don't need `_lock` at all?

LukaszRozmej: Not sure if it would be better?

bobtajson: I think it would be more code for the same result

PR #10673, opened 2026-02-28
chore: Replace `zkVM` with `zkEVM`

rubo:
## Changes

Following the EF parlance, replaced all `zkVM` with `zkEVM` to stay consistent with them. Also, the term itself is clearer and fit us better. 

- Renamed MSBuild properties
- Renamed compiler symbols
- Renamed files as `standard` and `zkEVM`
- Renamed GitHub labels

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No


LukaszRozmej (approved):
Kind of pointless and adds 2 more characters everywhere, now it will be longer to type :P
Potentially needs changes in #10120. Whatever in the end.

PR #10683, opened 2026-03-01
feat(eip-7981): increase access list cost to include token floor pricing

MarekM25:
## Changes

- Implements [EIP-7981](https://eips.ethereum.org/EIPS/eip-7981) — _Increase Access List Cost_
- Extends EIP-7623 floor pricing to also cover the data footprint of access lists, closing the loophole that allowed circumventing calldata floor pricing by combining large access lists with minimal calldata
- Adds `TOTAL_COST_FLOOR_PER_TOKEN * tokens_in_access_list` to both the standard access list cost and the EIP-7623 floor calculation
- Activated in the **Amsterdam** fork (`22_Amsterdam.cs`)

### Implementation details

- `IReleaseSpec` / `ReleaseSpec` / `ReleaseSpecDecorator` / `OverridableReleaseSpec` — new `IsEip7981Enabled` flag
- `IGasPolicy` — new `CalculateTokensInAccessList()` helper; `AccessListCost()` adds token floor cost when EIP-7981 enabled; `CalculateFloorCost()` includes access list tokens in total
- `EthereumGasPolicy.CalculateIntrinsicGas()` — computes access list tokens and passes them to floor cost calculation
- ChainSpec support (`ChainParameters`, `ChainSpecParamsJson`, `ChainSpecLoader`, `ChainSpecBasedSpecProvider`) — `Eip7981TransitionTimestamp`

## Type of change

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Testing

- [x] Added regression tests in `Eip7623Tests.cs`:
  - Access list with address — verifies token floor cost added to standard gas and floor gas (Amsterdam)
  - Access list with address + storage key — verifies combined token count (Amsterdam)
  - Access list in Prague — verifies no change before EIP-7981 activation

## Checklist

- [x] I have performed a self-review of my code
- [x] I have tested my code
- [ ] I have updated the documentation (if applicable)

### DIFF
--- a/src/Nethermind/Nethermind.Evm.Benchmark/AccessListTokensBenchmarks.cs
+++ b/src/Nethermind/Nethermind.Evm.Benchmark/AccessListTokensBenchmarks.cs
@@ -63,7 +63,7 @@ public class AccessListTokensBenchmarks
         if (accessList is null) return 0L;
 
         long tokens = 0;
-        long nonZeroMultiplier = _spec.GetTxDataNonZeroMultiplier();
+        long nonZeroMultiplier = _spec.GasCosts.TxDataNonZeroMultiplier;
         Span<byte> keyBytes = stackalloc byte[Nethermind.Core.Extensions.UInt256Extensions.ByteSize];
         foreach ((Address address, AccessList.StorageKeysEnumerable storageKeys) in accessList)
         {


src/Nethermind/Nethermind.Evm/GasPolicy/IGasPolicy.cs
```diff
@@ -257,7 +281,12 @@ public static long AccessListCost(Transaction transaction, IReleaseSpec spec)
...
 
             (int addressesCount, int storageKeysCount) = accessList.Count;
-            return addressesCount * GasCostOf.AccessAccountListEntry + storageKeysCount * GasCostOf.AccessStorageListEntry;
+            long cost = addressesCount * GasCostOf.AccessAccountListEntry + storageKeysCount * GasCostOf.AccessStorageListEntry;
+            if (spec.IsEip7981Enabled)
+            {
+                cost += GasCostOf.TotalCostFloorPerTokenEip7623 * CalculateTokensInAccessList(transaction, spec);
+            }
+            return cost;
         }
```
MarekM25, line 285: Double iteration of the access list: `CalculateTokensInAccessList` is called here inside `AccessListCost`, but `CalculateIntrinsicGas` already computed the same value on the line above its `AccessListCost` call.

Fix: accept the precomputed count as a parameter instead of recomputing it:
```csharp
public static long AccessListCost(Transaction transaction, IReleaseSpec spec, long tokensInAccessList = 0)
{
    // ...
    long cost = addressesCount * GasCostOf.AccessAccountListEntry + storageKeysCount * GasCostOf.AccessStorageListEntry;
    if (spec.IsEip7981Enabled)
        cost += GasCostOf.TotalCostFloorPerTokenEip7623 * tokensInAccessList;
    return cost;
}
```
Then in `CalculateIntrinsicGas` pass `tokensInAccessList` directly: `AccessListCost(tx, spec, tokensInAccessList)`. This removes the second traversal entirely and the standalone `CalculateTokensInAccessList` call in `CalculateIntrinsicGas` can be inlined into the `CalculateFloorCost` call (or kept, but only called once).

src/Nethermind/Nethermind.Benchmark.Runner/Program.cs
```diff
@@ -70,15 +65,17 @@ public static void Main(string[] args)
...
             else
             {
-                Assembly[] releaseAssemblies = additionalJobAssemblies
-                    .Union(simpleJobAssemblies)
-                    .Append(typeof(KeccakBenchmark).Assembly)
-                    .Distinct()
-                    .ToArray();
+                foreach (Assembly assembly in additionalJobAssemblies)
+                {
+                    BenchmarkRunner.Run(assembly, new DashboardConfig(Job.MediumRun), args);
```
MarekM25: Why did we change it?

src/Nethermind/Nethermind.Core/Eip2930/AccessList.cs
```diff
@@ -28,6 +28,7 @@ private AccessList(List<(Address address, int count)> addresses, List<UInt256> k
     public bool IsEmpty => _addresses.Count == 0;
     public (int AddressesCount, int StorageKeysCount) Count => (_addresses.Count, _keys.Count);
 
+
```
MarekM25: not needed change

src/Nethermind/Nethermind.Evm/GasPolicy/IGasPolicy.cs
MarekM25: Already fixed — `AccessListCost` now accepts `tokensInAccessList` as a required parameter (no default to avoid silent zeroing). `CalculateIntrinsicGas` computes it once and passes the precomputed value to both `AccessListCost` and `CalculateFloorCost`.

src/Nethermind/Nethermind.Benchmark.Runner/Program.cs
MarekM25: Unintended change — reverting.

src/Nethermind/Nethermind.Core/Eip2930/AccessList.cs
MarekM25: Reverting.

src/Nethermind/Nethermind.Evm.Benchmark/AccessListTokensBenchmarks.cs
```diff
@@ -0,0 +1,106 @@
...
+
+/// <summary>
+/// Compares two approaches for counting zero/non-zero bytes in access list storage keys:
+///   Span — serialize UInt256 via ToBigEndian into a stackalloc byte[32], then count zeros over the span
+///   Swar — borrow-safe SWAR directly on the four ulong limbs, no memory write (production path)
+/// </summary>
+[MemoryDiagnoser]
+public class AccessListTokensBenchmarks
+{
+    private static readonly IReleaseSpec _spec = Osaka.Instance;
```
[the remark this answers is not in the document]
MarekM25: Good catch — Osaka (fork 19) predates Amsterdam (fork 22), so \ is false and \ returns 0 immediately. Fixed to \.

src/Nethermind/Nethermind.Evm/GasPolicy/IGasPolicy.cs
```diff
@@ -246,30 +247,55 @@ protected static long CalculateTokensInCallData(Transaction transaction, IReleas
...
+        if (!spec.UseTxAccessLists)
+        {
+            ThrowInvalidDataException(spec);
         }
 
-        return 0;
+        (int addressesCount, int storageKeysCount) = accessList.Count;
+        return addressesCount * GasCostOf.AccessAccountListEntry
+            + storageKeysCount * GasCostOf.AccessStorageListEntry
+            + GasCostOf.TotalCostFloorPerTokenEip7623 * tokensInAccessList;
```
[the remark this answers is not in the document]
MarekM25: Applied — added the \ and early return so the token-floor addend is only applied when EIP-7981 is enabled.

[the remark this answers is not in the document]
MarekM25: Reverting — callers always pass the result of \ which already returns 0 when EIP-7981 is disabled, so the assert is unnecessary and adds noise.

src/Nethermind/Nethermind.Evm.Test/Eip7623Tests.cs
```diff
@@ -31,4 +34,176 @@ public void zero_data_transaction_floor_cost_should_be_10()
         cost.Should().Be(new EthereumIntrinsicGas(Standard: GasCostOf.Transaction + GasCostOf.TxDataZero,
             FloorGas: GasCostOf.Transaction + GasCostOf.TotalCostFloorPerTokenEip7623));
     }
+
+    [Test]
+    public void access_list_address_cost_includes_token_floor_with_eip7981()
```
newmanifold, line 39: These EIP-7981 specific tests should be in separate file, no?

src/Nethermind/Nethermind.Core/Extensions/UInt256Extensions.cs
```diff
@@ -66,6 +67,14 @@ public static ValueHash256 ToValueHash(this in UInt256 value)
         return Unsafe.As<Word, ValueHash256>(ref result);
     }
 
+    /// <summary>
+    /// Returns the number of zero-valued bytes in the 256-bit value.
+    /// Operates directly on the four <see cref="ulong"/> limbs — no serialization needed.
+    /// </summary>
+    public static int CountZeroBytes(this in UInt256 value) =>
+        value.u0.CountZeroBytes() + value.u1.CountZeroBytes()
```
newmanifold, line 75: I think we can possibly use SIMD here, similar to how it's used in Span<byte>.CountZeros, instead of performing four SWAR operations.I did some quick benchmarking on x86, and the SIMD version was about 3–4× faster ARM requires slightly different instructions (I am still testing that) to get similar bump but it was still faster than 4x Swar with Extract MSB approach.

This could be especially beneficial since this method will be used inside a loop in the access list.

PR #10685, opened 2026-03-01
perf: Refactor `WitnessExtensions.DecodeHeaders` method

rubo:
## Changes

- Set pre-sized `ArrayPoolList` to work around `GuardIndex`
- Used `Rlp.GetValueDecoder<BlockHeader>()` instead of creating a new `HeaderDecoder` instance
- Improved error handling (I'm not sure we should throw there, though)

- [x] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Stateless/Witness.cs
+++ b/src/Nethermind/Nethermind.Consensus/Stateless/Witness.cs
@@ -31,7 +31,8 @@ public class Witness : IDisposable
 
 public static class WitnessExtensions
 {
-    private static readonly IRlpValueDecoder<BlockHeader> _decoder = Rlp.GetValueDecoder<BlockHeader>();
+    private static readonly IRlpValueDecoder<BlockHeader> _decoder =
+        Rlp.GetValueDecoder<BlockHeader>() ?? new HeaderDecoder();
 
     extension(Witness witness)
     {


src/Nethermind/Nethermind.Consensus/Stateless/Witness.cs
rubo: @LukaszRozmej What about `Rlp.GetStreamDecoder<BlockHeader>()` instead?

LukaszRozmej: sure, no difference practically

rubo: Doesn't it return an already existing instance instead of creating a new one?

rubo:
> There is practically no difference, array pool list is pre-allocated

Well, setting an item with indexer `[]` fails, while `Add()` works, which an indication of resizing.

LukaszRozmej:
> > There is practically no difference, array pool list is pre-allocated
> 
> Well, setting an item with indexer `[]` fails, while `Add()` works, which an indication of resizing.

Can you elaborate? 
I think you are confusing internal count manipulation with resizing the internal buffer.

benaadams (approved):

PR #10688, opened 2026-03-02
fix: return InternalError instead of ParseError for unhandled exceptions in JSON-RPC

bobtajson:
## Changes

The catch-all branch in `ReturnErrorResponse` returned `-32700 Parse error` for unhandled exceptions during method execution. Per JSON-RPC spec, `-32700` means malformed JSON. Changed to `-32603 Internal error` to match the spec and the other branches in the same method.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

### DIFF
--- a/src/Nethermind/Nethermind.JsonRpc.Test/JsonRpcServiceTests.cs
+++ b/src/Nethermind/Nethermind.JsonRpc.Test/JsonRpcServiceTests.cs
@@ -256,7 +256,6 @@ public class JsonRpcServiceTests
 
             (false, Build.A.Block.WithWithdrawals(null).TestObject)
         };
-    
     [Test]
     public async Task Unhandled_exception_returns_InternalError()
     {


bobtajson:
@benaadams can you please retrigger CI?

src/Nethermind/Nethermind.JsonRpc.Test/JsonRpcServiceTests.cs
```diff
@@ -255,4 +256,18 @@ public void BlockForRpc_should_expose_withdrawals_if_any((bool Expected, Block B
 
             (false, Build.A.Block.WithWithdrawals(null).TestObject)
         };
+
```
LukaszRozmej:
```suggestion

```

PR #10689, opened 2026-03-02
fix: Use AND-NOT instead of XOR to clear flags in PeerInfo

bobtajson:
## Changes

Free and WakeUp used ^= (XOR) to clear allocation/sleeping flags. XOR toggles bits — if a bit is already clear, it gets set instead of staying clear.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -834,7 +834,6 @@
     "unrequested",
     "unresolve",
     "unshifted",
-    "unsleeping",
     "unsub",
     "unsubscription",
     "unsynchronized",


src/Nethermind/Nethermind.Synchronization.Test/PeerInfoTests.cs
```diff
@@ -139,5 +139,34 @@ public void Cannot_allocate_subcontext_of_sleeping()
...
+            peerInfo.TryAllocate(AllocationContexts.Headers);
+
+            peerInfo.Free(AllocationContexts.Bodies);
+
+            peerInfo.IsAllocated(AllocationContexts.Headers).Should().BeTrue();
+            peerInfo.IsAllocated(AllocationContexts.Bodies).Should().BeFalse();
+        }
+
+        [Test]
+        public void WakeUp_does_not_set_unsleeping_context()
```
LukaszRozmej: `unsleeping` is not a word, maybe you can name it better?

bobtajson: > `unsleeping` is not a word, maybe you can name it better?

renamed to WakeUp_does_not_put_awake_context_to_sleep
Is it better?

benaadams:
Thank you for your contribution

PR #10690, opened 2026-03-02
Fix Eth62ProtocolHandler to handle NewBlockMessage in background

batrr:
When a **new block** is received over P2P (gossiped `NewBlockMessage`), the node can hang.

**Cause:** Block processing eventually calls `TrieStoreScopeProvider.Commit()`. It schedules tasks onto the same event loop thread and then blocks with `Task.WaitAll`, causing a self-deadlock.

Blocks with no transactions don’t trigger trie commits, so they process fine.
To reproduce on a private chain, create a transaction that modifies storage and process it from `NewBlockMessage` gossip.
The hang was observed on XDC and JOC 

Flow when a new block is received (stuck at Task.WaitAll):

```
ZeroNettyP2PHandler.ChannelRead0()
  → XdcProtocolHandler.HandleMessage()
  → Eth62ProtocolHandler.Handle()
  → SyncServer.AddNewBlock()
  → SyncServer.SyncBlock()
  → BlockTree.SuggestBlock()
  → … process block …
  → WorldState.CommitTree()
  → TrieStoreScopeProvider.Commit()
  → Task.WaitAll()   ← stuck here
```

### DIFF
--- a/src/Nethermind/Nethermind.Network.Test/P2P/Subprotocols/Eth/V62/Eth62ProtocolHandlerTests.cs
+++ b/src/Nethermind/Nethermind.Network.Test/P2P/Subprotocols/Eth/V62/Eth62ProtocolHandlerTests.cs
@@ -306,7 +306,7 @@ namespace Nethermind.Network.Test.P2P.Subprotocols.Eth.V62
         }
 
         [Test]
-        public void Throws_if_adding_new_block_fails()
+        public void Disconnects_peer_if_adding_new_block_fails()
         {
             using NewBlockMessage newBlockMessage = new();
             newBlockMessage.Block = Build.A.Block.WithParent(_genesisBlock).TestObject;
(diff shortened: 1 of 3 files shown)


src/Nethermind/Nethermind.Network.Test/P2P/Subprotocols/Eth/V62/Eth62ProtocolHandlerTests.cs
```diff
@@ -317,10 +317,10 @@ public void Throws_if_adding_new_block_fails()
...
             getBlockHeadersPacket.ReadByte();
 
-            _syncManager.WhenForAnyArgs(w => w.AddNewBlock(null, _handler)).Do(ci => throw new Exception());
-            Assert.Throws<Exception>(
-                () => _handler.HandleMessage(
-                    new ZeroPacket(getBlockHeadersPacket) { PacketType = Eth62MessageCode.NewBlock }));
+            _syncManager.WhenForAnyArgs(w => w.AddNewBlock(null!, _handler)).Do(_ => throw new Exception());
+            _handler.HandleMessage(new ZeroPacket(getBlockHeadersPacket) { PacketType = Eth62MessageCode.NewBlock });
+
+            _session.Received().InitiateDisconnect(DisconnectReason.BackgroundTaskFailure, Arg.Any<string>());
```
ak88, line 323: Are we still sending correct DisconnectReason when back ground tasks fails?

batrr: Default:
https://github.com/NethermindEth/nethermind/blob/9b3d6c49038d58a35fee5c748910f9e29d0b7556/src/Nethermind/Nethermind.Network/P2P/ProtocolHandlers/ZeroNettyP2PHandler.cs#L118-L130

For background tasks:
https://github.com/NethermindEth/nethermind/blob/9b3d6c49038d58a35fee5c748910f9e29d0b7556/src/Nethermind/Nethermind.Network/P2P/Utils/BackgroundTaskSchedulerWrapper.cs#L55-L63

Now we use more specific disconnect reasons, but we also disconnect on internal exceptions.
But the only usage of `IInternalNethermindException` is `TrieException`.

ak88: @flcl42 what is your opinion here?

flcl42: Looks ok

PR #10692, opened 2026-03-02
feat(evm): add `ConsumeCodeDeposit` to `IGasPolicy`

AnkushinDaniil:
Add `ConsumeCodeDeposit` method to `IGasPolicy` interface for multi-dimensional gas tracking.
Currently `ConsumeCodeDeposit` accepts cost as a parameter that violates gas policy principle that cost must be calculated inside gas policy. This PR is a hot fix. Proper handling of this gas consumption case requires more changes and refactoring that will be addressed in the future massive gas policy refactoring

## Linked PR

https://github.com/NethermindEth/nethermind-arbitrum/pull/new/daniil/multigas-code-deposit


src/Nethermind/Nethermind.Evm/VirtualMachine.cs
LukaszRozmej, line 409: This is now confusing and not used

LukaszRozmej, line 410: this check doesn't make sense

PR #10695, opened 2026-03-02
fix: remove redundant SetNonce call in GeneralTestBase

reallesee:
CreateAccount already accepts nonce as a parameter, so calling SetNonce separately right after was unnecessary. This aligns GeneralTestBase with BlockchainTestBase which uses the single-call approach.

### DIFF
--- a/src/Nethermind/Ethereum.Test.Base/GeneralTestBase.cs
+++ b/src/Nethermind/Ethereum.Test.Base/GeneralTestBase.cs
@@ -210,8 +210,8 @@ namespace Ethereum.Test.Base
                         storageItem.Value.WithoutLeadingZeros().ToArray());
                 }
 
-            stateProvider.CreateAccount(accountState.Key, accountState.Value.Balance, accountState.Value.Nonce);
-            stateProvider.InsertCode(accountState.Key, accountState.Value.Code, specProvider.GenesisSpec);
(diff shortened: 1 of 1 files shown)


src/Nethermind/Ethereum.Test.Base/GeneralTestBase.cs
```diff
@@ -210,9 +210,8 @@ public static void InitializeTestState(Dictionary<Address, AccountState> preStat
                         storageItem.Value.WithoutLeadingZeros().ToArray());
                 }
 
-                stateProvider.CreateAccount(accountState.Key, accountState.Value.Balance);
-                stateProvider.InsertCode(accountState.Key, accountState.Value.Code, specProvider.GenesisSpec);
-                stateProvider.SetNonce(accountState.Key, accountState.Value.Nonce);
+            stateProvider.CreateAccount(accountState.Key, accountState.Value.Balance, accountState.Value.Nonce);
+            stateProvider.InsertCode(accountState.Key, accountState.Value.Code, specProvider.GenesisSpec);
```
LukaszRozmej:
```suggestion
                stateProvider.CreateAccount(accountState.Key, accountState.Value.Balance, accountState.Value.Nonce);
                stateProvider.InsertCode(accountState.Key, accountState.Value.Code, specProvider.GenesisSpec);
```

PR #10699, opened 2026-03-03
fix: Flat selfdestruct critical case

asdacap:
## Summary
- Fix selfdestruct edge case in flat world state scope.
- Repreducable on early mainnet blocks.
- Parameterize `StateProviderTests`, `StorageProviderTests`, and `ScopeProviderTests` to run against both trie and flat scope providers
- Add `CreateFlatScopeProvider` helper to `TestWorldStateFactory`

## Test plan
- [x] All parameterized tests pass for both `useFlat=true` and `useFlat=false` fixtures
- [x] Verify no regressions in existing state tests
- [x] Era mainnet replay

### DIFF
--- a/src/Nethermind/Nethermind.State.Flat/ScopeProvider/FlatWorldStateScope.cs
+++ b/src/Nethermind/Nethermind.State.Flat/ScopeProvider/FlatWorldStateScope.cs
@@ -286,10 +286,6 @@ public sealed class FlatWorldStateScope : IWorldStateScopeProvider.IScope, ITrie
             [MethodImpl(MethodImplOptions.NoInlining)]
             void Trace(Address address, Hash256 storageRoot, Account? account) =>
                 logger.Trace($"Update {address} S {account?.StorageRoot} -> {storageRoot}");
-
-            [DoesNotReturn, StackTraceHidden]
-            static Account ThrowNullAccount(Address address) =>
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.State.Flat/ScopeProvider/FlatWorldStateScope.cs
```diff
@@ -255,6 +255,12 @@ public void Dispose()
...
                     if (!_dirtyAccounts.TryGetValue(key, out Account? account)) account = scope.Get(key);
                     if (account == null && storageRoot == Keccak.EmptyTreeHash) continue;
+                    if (account == null)
+                    {
+                        using var wb = CreateStorageWriteBatch(entry.Item1, 0);
+                        wb.Clear();
+                        continue;
+                    }
                     account ??= ThrowNullAccount(key);
                     account = account!.WithChangedStorageRoot(storageRoot);
```
LukaszRozmej:
```suggestion
                    if (account is null)
                    {
                        if (storageRoot == Keccak.EmptyTreeHash) continue;
                        using var wb = CreateStorageWriteBatch(entry.Item1, 0);
                        wb.Clear();
                        continue;
                    }
                    account = account.WithChangedStorageRoot(storageRoot);
```

PR #10705, opened 2026-03-03
Use actual blob gas fee for any tx type

flcl42:
## Changes

- Use actual blob gas fee for any tx type

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Facade.Test/BlockchainBridgeTests.cs
+++ b/src/Nethermind/Nethermind.Facade.Test/BlockchainBridgeTests.cs
@@ -284,19 +284,15 @@ public class BlockchainBridgeTests
             Arg.Any<ITxTracer>());
     }
 
-    private static IEnumerable<TestCaseData> BlobBaseFeeTestCases()
-    {
-        string[] methods = [nameof(IBlockchainBridge.Call), nameof(IBlockchainBridge.EstimateGas), nameof(IBlockchainBridge.CreateAccessList)];
-        ulong[] excessBlobGasValues = [0, 100];
+    private static Action<IBlockchainBridge, BlockHeader, Transaction>[] BridgeCallSources() =>
+    [
+        (bridge, header, tx) => bridge.Call(header, tx),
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Facade.Test/BlockchainBridgeTests.cs
```diff
@@ -282,6 +284,51 @@ public void Call_sets_maxFeePerBlobGas()
...
+            case nameof(IBlockchainBridge.Call):
+                _blockchainBridge.Call(header, tx);
+                break;
+            case nameof(IBlockchainBridge.EstimateGas):
+                _blockchainBridge.EstimateGas(header, tx, 1);
+                break;
+            case nameof(IBlockchainBridge.CreateAccessList):
+                _blockchainBridge.CreateAccessList(header, tx, default, false);
+                break;
+        }
```
LukaszRozmej: maybe instead of this, just pass pass a delegate?

src/Nethermind/Nethermind.Facade.Test/BlockchainBridgeTests.cs
```diff
@@ -282,6 +284,51 @@ public void Call_sets_maxFeePerBlobGas()
             Arg.Any<ITxTracer>());
     }
 
+    private static IEnumerable<TestCaseData> BlobBaseFeeTestCases()
+    {
+        string[] methods = [nameof(IBlockchainBridge.Call), nameof(IBlockchainBridge.EstimateGas), nameof(IBlockchainBridge.CreateAccessList)];
+        ulong[] excessBlobGasValues = [0, 100];
```
LukaszRozmej: this could be done with [Values] instead of TestCaseSource - probably not with delegates though

marcindsobczak (approved):

PR #10716, opened 2026-03-04
Fix XDC Transaction Filter

batrr:
XDC Filter was rejecting valid signing txs from peers, which caused disconnects.

Reference:
https://github.com/XinFinOrg/XDPoSChain/blob/41fb331a2e7d93683d4c895964b05c7fc46f53bd/consensus/XDPoS/engines/engine_v2/engine.go#L559
https://github.com/XinFinOrg/XDPoSChain/blob/41fb331a2e7d93683d4c895964b05c7fc46f53bd/core/txpool/validation.go#L101

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Xdc/TxPool/SignTransactionFilter.cs
+++ b/src/Nethermind/Nethermind.Xdc/TxPool/SignTransactionFilter.cs
@@ -69,13 +69,7 @@ internal sealed class SignTransactionFilter(ISnapshotManager snapshotManager, IB
         return AcceptTxResult.Accepted;
     }
 
-    private static bool IsEpochCandidate(Snapshot? snapshot, Address? senderAddress)
-    {
-        if (snapshot is null || senderAddress is null)
-        {
-            return false;
-        }
-
-        return snapshot.NextEpochCandidates.Contains(senderAddress);
-    }
+    private static bool IsEpochCandidate(Snapshot? snapshot, Address? senderAddress) =>
+        snapshot is not null && senderAddress is not null &&
+        snapshot.NextEpochCandidates.Contains(senderAddress);
 }


src/Nethermind/Nethermind.Xdc/TxPool/SignTransactionFilter.cs
```diff
@@ -27,36 +27,62 @@ private AcceptTxResult ValidateSignTransaction(Transaction tx, long headerNumber
...
-            return AcceptTxResult.Invalid;
+            return AcceptTxResult.Invalid.WithMessage("Sign transaction data length is less than required length");
         }
 
         UInt256 blkNumber = new UInt256(tx.Data.Span.Slice(4, 32), true);
         if (blkNumber > headerNumber || blkNumber <= (headerNumber - (xdcSpec.EpochLength * 2)))
         {
             // Invalid block number in special transaction data
-            return AcceptTxResult.Invalid;
+            return AcceptTxResult.Invalid.WithMessage("Sign transaction block number is out of range");
```
ak88, line 38: Maybe AlreadyKnown known here so we dont disconnect. A Failure here is not really a violation.

src/Nethermind/Nethermind.Xdc/TxPool/SignTransactionFilter.cs
```diff
@@ -27,36 +27,62 @@ private AcceptTxResult ValidateSignTransaction(Transaction tx, long headerNumber
...
-            if (tx.IsSignTransaction(spec) && !ValidateSignTransaction(tx, headerNumber, spec))
+            AcceptTxResult result = ValidateSignTransaction(tx, headerNumber, spec);
+            if (result != AcceptTxResult.Accepted)
             {
-                return AcceptTxResult.Invalid;
+                return result;
             }
+        }
+
+        Snapshot? snapshot = snapshotManager.GetSnapshotByBlockNumber(headerNumber, spec);
```
ak88, line 62: Should probably use actual master node list from epoch switch. There can be many candidates.

src/Nethermind/Nethermind.Xdc/TxPool/SignTransactionFilter.cs
batrr: This could still cause a disconnect.
https://github.com/NethermindEth/nethermind/blob/0887540ec5bf69b1b8a006ae62a59257e904019f/src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/V62/TxFloodController.cs#L42-L48

src/Nethermind/Nethermind.Xdc/TxPool/SignTransactionFilter.cs
```diff
@@ -27,36 +27,62 @@ private AcceptTxResult ValidateSignTransaction(Transaction tx, long headerNumber
...
 
     public AcceptTxResult Accept(Transaction tx, ref TxFilteringState state, TxHandlingOptions txHandlingOptions)
     {
-        var (headerNumber, spec) = GetSpecAndHeader();
+        if (!tx.IsSpecialTransaction((IXdcReleaseSpec)specProvider.GetFinalSpec()))
+        {
+            return AcceptTxResult.Accepted;
+        }
+
+        (long headerNumber, IXdcReleaseSpec spec) = GetSpecAndHeader();
```
cicr99: Why get the spec twice? We should be able to use the same

src/Nethermind/Nethermind.Xdc/TxPool/SignTransactionFilter.cs
```diff
@@ -27,36 +27,62 @@ private AcceptTxResult ValidateSignTransaction(Transaction tx, long headerNumber
...
+        {
+            if (candidate == senderAddress)
             {
-                return AcceptTxResult.Invalid;
+                return true;
             }
         }
 
-        return AcceptTxResult.Accepted;
+        return false;
```
cicr99: This could be simplified using Any() instead of the manual loop
```suggestion
        return snapshot.NextEpochCandidates.Any(candidate => candidate == senderAddress);
```

src/Nethermind/Nethermind.Xdc/TxPool/SignTransactionFilter.cs
batrr: Let's keep it as is - same as in go implementation and will need full candidate list after `TIPUpgradeReward`  (protectors + observers also send sign txs)

src/Nethermind/Nethermind.Xdc/TxPool/SignTransactionFilter.cs
```diff
@@ -27,36 +28,54 @@ private AcceptTxResult ValidateSignTransaction(Transaction tx, long headerNumber
...
+
+    private static bool IsEpochCandidate(Snapshot? snapshot, Address? senderAddress)
+    {
+        if (snapshot is null || senderAddress is null)
+        {
+            return false;
+        }
+
+        return snapshot.NextEpochCandidates.Contains(senderAddress);
+    }
```
alexb5dh:
```suggestion
    private static bool IsEpochCandidate(Snapshot? snapshot, Address? senderAddress) =>
        snapshot is not null && senderAddress is not null &&
        snapshot.NextEpochCandidates.Contains(senderAddress);
```

PR #10717, opened 2026-03-04
fix: widen SeqlockCache seq counter to 16 bits to prevent ABA flake

benaadams:
## Changes

- Widened the seqlock sequence counter from 8 bits (128 distinct values) to 16 bits (32,768 distinct values) to eliminate an ABA race condition
- Shrunk the hash signature from 28 bits to 20 bits to make room (still gives 1-in-1M false positive rate)
- Updated the eviction tiebreaker bit from `1L << 9` to `1L << 17` to stay within the new HashMask range

The 8-bit Seq field could wrap around under heavy contention (4 writers on 2 slots with identical hashes), causing readers to accept torn key/value pairs when `h1 == h2` held despite interleaved writes. This manifested as a flaky `Concurrent_read_write_returns_correct_value_for_key` test failure on CI.

#### Requires testing

- [x] No

#### Notes on testing

Existing `Concurrent_read_write_returns_correct_value_for_key` test covers this scenario — it was the test that caught the ABA bug. All 20 SeqlockCache tests pass.

### DIFF
--- a/src/Nethermind/Nethermind.Core/Collections/SeqlockCache.cs
+++ b/src/Nethermind/Nethermind.Core/Collections/SeqlockCache.cs
@@ -22,15 +22,15 @@ namespace Nethermind.Core.Collections;
 ///
 /// Hash bit partitioning (64-bit hash):
 ///   Bits  0-13: way 0 set index (14 bits)
-///   Bits 14-41: hash signature stored in header (28 bits)
+///   Bits 22-41: hash signature stored in header (20 bits)
 ///   Bits 42-55: way 1 set index (14 bits, independent from way 0)
 ///
 /// Header layout (64-bit):
-/// [Lock:1][Epoch:26][Hash:28][Seq:8][Occ:1]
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Core/Collections/SeqlockCache.cs
```diff
@@ -49,16 +49,16 @@ public sealed class SeqlockCache<TKey, TValue>
     private const int SetMask = Sets - 1;
 
     // Header bit layout:
-    // [Lock:1][Epoch:26][Hash:28][Seq:8][Occ:1]
+    // [Lock:1][Epoch:26][Hash:20][Seq:16][Occ:1]
```
LukaszRozmej, line 52: Can we keep the hash bigger but make epoch smaller? The way we use it epoch is per block?

benaadams: hash is for performance (key is still compared) but epoch is for correctness on O(1) clear, so if it wraps that's a problem as old entries will be accepted as valid

PR #10720, opened 2026-03-04
perf: reduce GC pressure in flat/snap sync hot paths

smartprogrammer93:
## Changes

- Add `in ValueHash256` overloads to `SetAccountRaw`/`SetStorageRaw`/`GetAccountRaw`/`TryGetStorageRaw` across the full interface chain (`IPersistence`, `BasePersistence`, all implementors), eliminating 15+ `ToCommitment()` heap allocations per sync cycle
- Eliminate double RocksDB snapshot in `FlatTreeSyncStore.SaveNode` by sharing a single `IPersistenceReader` between the write batch and `ReadExistingNode`
- Batch `Interlocked.Add` for metrics in all 4 snap tree implementations (`FlatSnapStateTree`, `FlatSnapStorageTree`, `PatriciaSnapStateTree`, `PatriciaSnapStorageTree`), replacing per-entry memory barriers with a single post-loop atomic add
- Remove LINQ `Select` allocation in `FlatStateRootIndex.ResetAvailableStateRoots`, replaced with a `for` loop
- Replace `Dictionary<FlatDbColumns, T>` with array indexing in `WriteBufferAdjuster`, avoiding enum key boxing and hash table overhead
- Fix `ArrayPoolList` leak in `FlatSnapServer.GetTrieNodes` on early return (correctness bug)
- Eliminate `.ToArray()` allocation in `FlatSnapServer.GetStorageRanges` by accepting `ReadOnlySpan<byte>` in `GetAccountByPath`
- Use `AccountDecoder.Instance` singleton instead of allocating a new `AccountDecoder` instance in `FlatSnapServer`

- [x] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

Existing tests updated to match `in ValueHash256` interface signatures (NSubstitute `Arg.Any<ValueHash256>()` matchers). All 98 flat state tests and 85 sync tests pass.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.State.Flat/Persistence/WriteBufferAdjuster.cs
+++ b/src/Nethermind/Nethermind.State.Flat/Persistence/WriteBufferAdjuster.cs
@@ -1,6 +1,7 @@
 // SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
 // SPDX-License-Identifier: LGPL-3.0-only
 
+using System.Runtime.CompilerServices;
 using Nethermind.Core;
 using Nethermind.Db;
 
@@ -8,15 +9,27 @@ namespace Nethermind.State.Flat.Persistence;
 
 internal class WriteBufferAdjuster(IColumnsDb<FlatDbColumns> db)
 {
-    private const int ColumnCount = 7; // Number of FlatDbColumns values
+    private const int ColumnCount = (int)FlatDbColumns.FallbackNodes + 1;
     private const long MinWriteBufferSize = 16L * 1024 * 1024;   // 16 MB floor
     private const long MaxWriteBufferSize = 256L * 1024 * 1024;  // 256 MB cap
 
     private bool _syncBufferSet;
 
-    private readonly long[] _lastWriteBufferSize = new long[ColumnCount];
-    private readonly CountingWriteBatch?[] _activeCounters = new CountingWriteBatch?[ColumnCount];
+    private WriteBufferSizeBuffer _lastWriteBufferSize;
+    private ActiveCounterBuffer _activeCounters;
     private int _activeCounterCount;
+
+    [InlineArray(ColumnCount)]
+    private struct WriteBufferSizeBuffer
+    {
+        private long _element0;
+    }
+
+    [InlineArray(ColumnCount)]
+    private struct ActiveCounterBuffer
+    {
+        private CountingWriteBatch? _element0;
+    }
 
     public IWriteBatch Wrap(IColumnsWriteBatch<FlatDbColumns> batch, FlatDbColumns column, WriteFlags flags)
     {


src/Nethermind/Nethermind.State.Flat/Persistence/WriteBufferAdjuster.cs
```diff
@@ -8,13 +8,15 @@ namespace Nethermind.State.Flat.Persistence;
...
+    private const int ColumnCount = 7; // Number of FlatDbColumns values
     private const long MinWriteBufferSize = 16L * 1024 * 1024;   // 16 MB floor
     private const long MaxWriteBufferSize = 256L * 1024 * 1024;  // 256 MB cap
 
     private bool _syncBufferSet;
 
-    private readonly Dictionary<FlatDbColumns, long> _lastWriteBufferSize = new();
-    private readonly Dictionary<FlatDbColumns, CountingWriteBatch> _activeCounters = new();
+    private readonly long[] _lastWriteBufferSize = new long[ColumnCount];
+    private readonly CountingWriteBatch?[] _activeCounters = new CountingWriteBatch?[ColumnCount];
```
LukaszRozmej: Those arrays could be inline arrays on struct - so will be closer in memory https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-12.0/inline-arrays

src/Nethermind/Nethermind.State.Flat/Persistence/WriteBufferAdjuster.cs
```diff
@@ -8,13 +8,15 @@ namespace Nethermind.State.Flat.Persistence;
 
 internal class WriteBufferAdjuster(IColumnsDb<FlatDbColumns> db)
 {
+    private const int ColumnCount = 7; // Number of FlatDbColumns values
```
LukaszRozmej:
```suggestion
    private const int ColumnCount = (int)FlatDbColumns.MaxValue;
```

and add:

```
public enum FlatDbColumns
{
    FallbackNodes,
    MaxValue = FallbackNodes
}

```

smartprogrammer93: couldnt do it this way because rocksDb iterates over the enum keys, so it errors out.

LukaszRozmej: Add `FlatDbColumnsStats` with `MaxValue` const next to `FlatDbColumns` so you can't miss it.

LukaszRozmej: but current state is acceptable

LukaszRozmej: Maybe make this const public and add a test that it is correct?

LukaszRozmej: Or Debug.Assert somewhere in code?

PR #10721, opened 2026-03-04
Fix Eth62ProtocolHandler to handle NewBlockMessage in background #2

batrr:
The earlier fix used `HandleInBackground`, which runs on `BackgroundTaskScheduler`. That worked fine when there were no or little nested calls. But with enough state changes, `TrieStoreScopeProvider.Commit()` starts spawning tasks for each storage tree, and each `PatriciaTree.Commit()` can spawn more tasks and block with `Task.WaitAll`.

Since all of this runs on the same scheduler, we end up hitting a deadlock - the background handler only has `BackgroundTaskConcurrency = 2` threads, which isn’t enough when trie commits start creating nested tasks.

Using `Task.Run()` moves the work to the thread pool, which has enough threads to handle these nested tasks.

Refer: https://github.com/NethermindEth/nethermind/pull/10690


src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/V62/Eth62ProtocolHandler.cs
```diff
@@ -307,21 +309,14 @@ private void Handle(NewBlockHashesMessage newBlockHashes)
             }
         }
 
-        private ValueTask Handle(NewBlockMessage msg, CancellationToken cancellationToken)
-        {
-            try
+        // Task.Run avoids BackgroundTaskScheduler's 2-thread limit (default config) 
+        // and deadlock risk when AddNewBlock spawns nested tasks.
+        private Task Handle(NewBlockMessage msg) => Task.Run(() =>
```
LukaszRozmej: No need to handle exceptions?

batrr: The default handler already logs and disconnects, so catching here would just duplicate that.


https://github.com/NethermindEth/nethermind/blob/9b3d6c49038d58a35fee5c748910f9e29d0b7556/src/Nethermind/Nethermind.Network/P2P/ProtocolHandlers/ZeroNettyP2PHandler.cs#L106-L130

src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/V62/Eth62ProtocolHandler.cs
```diff
@@ -307,21 +309,14 @@ private void Handle(NewBlockHashesMessage newBlockHashes)
...
             }
-            catch (Exception e)
-            {
-                if (Logger.IsDebug) Logger.Debug($"Handling {msg} from {Node:c} failed: " + e.Message);
-                throw;
-            }
-
-            return ValueTask.CompletedTask;
-        }
+        );
```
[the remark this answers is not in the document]
LukaszRozmej: I agree those should be queued to avoid DOS

src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/V62/Eth62ProtocolHandler.cs
LukaszRozmej: you are blocking network thread then, not good either

src/Nethermind/Nethermind.State/TrieStoreScopeProvider.cs
```diff
@@ -127,15 +128,14 @@ public void Commit(long blockNumber)
                         StorageTree st = (StorageTree)ctx;
                         st.Commit();
                         blockCommitter.ReturnConcurrencyQuota();
-                    }, storage.Value));
+                    }, storage.Value, CancellationToken.None, TaskCreationOptions.None, TaskScheduler.Default));
                 }
                 else
                 {
                     storage.Value.Commit();
                 }
```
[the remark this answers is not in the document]
batrr: ohhh

PR #10722, opened 2026-03-04
fix: background task scheduler — remove re-queue loop, add cancellation bail-outs

kamilchodola:
## Changes

- **Remove re-queue/throttle loop**: The `goto Throttle` path re-queued unexpired tasks during block processing, causing `_queueCount` to grow monotonically toward the capacity cap and dropping tasks under sustained load. Replaced with a single-pass drain.
- **Hybrid cancelled-token handling during block processing**: Tasks run with a cancelled CancellationToken so handlers can bail out quickly. If queue depth exceeds 80% capacity, tasks are disposed without execution to prevent overflow.
- **Stale-token refresh**: Inner drain loop breaks after processing a cancelled task so the outer loop creates a fresh linked CTS — prevents stale cancellation after block processing ends.
- **Early cancellation bail-out in P2P handlers**: `SyncPeerProtocolHandlerBase` methods (`FindHeaders`, `GetBlockBodies`, `GetReceipts`) check cancellation at loop entry before DB lookups. `HandleSlow` disposes the transaction list when cancelled before processing (prevents infinite reschedule loop).
- **Rate-limited drop warnings**: Throttled to once per 10 seconds.
- **Source tagging**: `TryScheduleTask` accepts `source` parameter for drop log triage.
- **New metrics**: `TotalBackgroundTasksDropped`, `TotalBackgroundTasksExecuted`, `TotalBackgroundTasksCancelled`.

- [x] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

- 12 BackgroundTaskScheduler tests (8 unit + 4 parameterized burst) all pass.
- 50 Eth62ProtocolHandler tests pass with the HandleSlow and interface changes.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
+++ b/src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
@@ -154,11 +154,11 @@ public class BackgroundTaskScheduler : IBackgroundTaskScheduler, IAsyncDisposabl
 
         WaitForBlockProcessing:
             cts.Dispose();
-            // Async-wait for block processing to finish instead of polling/re-queuing in a loop
+            // Wait for block processing to finish, but wake up periodically to drain expired tasks
             TaskCompletionSource? signal = _blockProcessingDoneSignal;
-            if (signal is not null)
+            if (signal is not null && !signal.Task.IsCompleted)
             {
-                await signal.Task.WaitAsync(_mainCancellationTokenSource.Token);
+                await Task.WhenAny(signal.Task, Task.Delay(100, _mainCancellationTokenSource.Token));
             }
         }
     }
--- a/src/Nethermind/Nethermind.Consensus.Test/Scheduler/BackgroundTaskSchedulerTests.cs
+++ b/src/Nethermind/Nethermind.Consensus.Test/Scheduler/BackgroundTaskSchedulerTests.cs
@@ -112,10 +112,10 @@ public class BackgroundTaskSchedulerTests
                 if (token.IsCancellationRequested)
                     Interlocked.Increment(ref cancelledCount);
                 return Task.CompletedTask;
-            });
+            }, TimeSpan.FromMilliseconds(1));
         }
 
-        // Tasks during block processing run with a cancelled token
+        // Expired tasks during block processing run with a cancelled token
         Assert.That(() => Volatile.Read(ref cancelledCount), Is.EqualTo(5).After(2000, 10));
 
         // After block processing, new tasks execute with active token
@@ -136,7 +136,7 @@ public class BackgroundTaskSchedulerTests
     }
 
     [Test]
-    public async Task Test_task_during_block_processing_gets_cancelled_token_and_exits()
+    public async Task Test_expired_task_during_block_processing_gets_cancelled_token_and_exits()
     {
         await using BackgroundTaskScheduler scheduler = new(_branchProcessor, _chainHeadInfo, 2, 65536, LimboLogs.Instance);
         _branchProcessor.BlocksProcessing += Raise.EventWith(new BlocksProcessingEventArgs(null));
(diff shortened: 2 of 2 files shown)


src/Nethermind/Nethermind.Consensus.Test/Scheduler/BackgroundTaskSchedulerBurstTests.cs
```diff
@@ -0,0 +1,88 @@
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
```
stdevMac: Update year template

src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
```diff
@@ -189,12 +156,48 @@ public bool TryScheduleTask<TReq>(TReq request, Func<TReq, CancellationToken, Ta
...
+        DisposeDeep(request);
         return false;
     }
 
+    /// <summary>
+    /// Recursively disposes IDisposable members buried inside ValueTuples.
+    /// Requests are often wrapped as ((actualRequest, func), outerFunc) by BackgroundTaskSchedulerWrapper,
+    /// so a plain TryDispose() on the outer tuple is a no-op — the inner IOwnedReadOnlyList etc. would leak.
+    /// </summary>
+    private static void DisposeDeep<T>(T value)
```
asdacap: Remove, this is unreliable.

src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
```diff
@@ -119,26 +110,16 @@ private async Task StartChannel()
                 while (_taskQueue.Reader.TryRead(out IActivity activity))
                 {
                     Interlocked.Decrement(ref _queueCount);
+                    UpdateQueueCount();
+
+                    await activity.Do(token);
```
asdacap: No? Like the point is if there is a chance that the activity can run AFTER the block processing, (token.IsCancellationRequested is true on block processing), we want to queue it and run after block processing.

src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
kamilchodola: Removed in b22f496176. Reverted to `request.TryDispose()`.

src/Nethermind/Nethermind.Consensus/Scheduler/BackgroundTaskScheduler.cs
kamilchodola: You're right — fixed in b22f496176. Tasks with remaining deadline are now re-queued and the consumer async-waits on _blockProcessingDoneSignal (a TaskCompletionSource) until block processing finishes. Only expired tasks run with cancelled token for cleanup. No polling loop — single await instead of the old goto Throttle + Task.Delay(1) spin.

src/Nethermind/Nethermind.Consensus.Test/Scheduler/BackgroundTaskSchedulerBurstTests.cs
kamilchodola: Already updated to 2026 in 2bef304fc3.

src/Nethermind/Nethermind.Network/P2P/Utils/BackgroundTaskSchedulerWrapper.cs
```diff
@@ -20,14 +21,28 @@ namespace Nethermind.Network.P2P.Utils;
 /// <param name="backgroundTaskScheduler"></param>
 public class BackgroundTaskSchedulerWrapper(ProtocolHandlerBase handler, IBackgroundTaskScheduler backgroundTaskScheduler)
 {
-    internal bool TryScheduleSyncServe<TReq, TRes>(TReq request, Func<TReq, CancellationToken, Task<TRes>> fulfillFunc) where TRes : P2PMessage =>
-        TryScheduleBackgroundTask((request, fulfillFunc), BackgroundSyncSender);
+    internal bool TryScheduleSyncServe<TReq, TRes>(TReq request, Func<TReq, CancellationToken, Task<TRes>> fulfillFunc) where TRes : P2PMessage
+    {
+        if (!TryScheduleBackgroundTask((request, fulfillFunc), BackgroundSyncSender, typeof(TReq).Name))
+        {
+            request.TryDispose();
```
LukaszRozmej, line 28: Why should they be disposed here, rather in the caller? Just for convienience?

PR #10723, opened 2026-03-04
fix(wallet): remove static modifier from DevWallet key seed

zeroprooff:
`_keySeed` was static but mutated in constructor. Each new DevWallet instance continued from where previous one stopped, producing different accounts.

tests didn't catch this because they cache a single instance in OneTimeSetUp

### DIFF
--- a/src/Nethermind/Nethermind.Wallet/DevWallet.cs
+++ b/src/Nethermind/Nethermind.Wallet/DevWallet.cs
@@ -17,7 +17,6 @@ namespace Nethermind.Wallet
     public class DevWallet : IWallet
     {
         private const string AnyPassword = "#DEV_ACCOUNT_NETHERMIND_ANY_PASSWORD#";
-        private static readonly byte[] _keySeed = new byte[32];
         private readonly ILogger _logger;
         private readonly Dictionary<Address, bool> _isUnlocked = new Dictionary<Address, bool>();
         private readonly Dictionary<Address, PrivateKey> _keys = new Dictionary<Address, PrivateKey>();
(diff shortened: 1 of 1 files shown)


AnkushinDaniil (commented):
The branch is conflicting with master now - DevWallet.cs was modernized there, so it needs a rebase (and the green CI is from March).

src/Nethermind/Nethermind.Wallet/DevWallet.cs
```diff
@@ -29,14 +28,15 @@ public DevWallet(IWalletConfig walletConfig, ILogManager logManager)
         {
             _logger = logManager?.GetClassLogger() ?? throw new ArgumentNullException(nameof(logManager));
 
-            _keySeed[31] = 1;
+            byte[] keySeed = new byte[32];
```
AnkushinDaniil, line 31: I think the described bug doesn't exist - the old ctor resets _keySeed[31] = 1 on every call and PrivateKey copies the buffer, so sequential instances already got the same accounts. The change still makes sense as a fix for concurrent construction (two ctors racing on the shared buffer), so maybe update the description to say that?

src/Nethermind/Nethermind.Wallet/DevWallet.cs
```diff
@@ -29,14 +28,15 @@ public DevWallet(IWalletConfig walletConfig, ILogManager logManager)
...
+            keySeed[31] = 1;
             for (int i = 0; i < walletConfig?.DevAccounts; i++)
             {
-                PrivateKey key = new PrivateKey(_keySeed);
+                PrivateKey key = new PrivateKey(keySeed);
                 _keys.Add(key.Address, key);
                 _passwords.Add(key.Address, AnyPassword);
                 _isUnlocked.Add(key.Address, true);
-                _keySeed[31]++;
+                keySeed[31]++;
```
AnkushinDaniil, line 39: Could you add a regression test? It probably needs to construct several DevWallets in parallel and assert they return the same accounts - a sequential comparison passes on both old and new code.

PR #10725, opened 2026-03-05
Move eest downloads to stable location

benaadams:
## Changes

- Currently fixtures are wiped on each test rebuild and then need to be redownloaded. They are pretty big so cache in a stable `eest` location rather than the `bin` directory

- [x] Optimization
- [x] Build-related changes

#### Requires testing

- [x] No

### DIFF
--- a/src/Nethermind/Ethereum.Blockchain.Pyspec.Test/LoadPyspecTestsStrategy.cs
+++ b/src/Nethermind/Ethereum.Blockchain.Pyspec.Test/LoadPyspecTestsStrategy.cs
@@ -92,7 +92,8 @@ public class LoadPyspecTestsStrategy : ITestLoadStrategy
     private static void DownloadAndExtract(string archiveVersion, string archiveName, string testsDirectoryName)
     {
         // Clean up any partial extraction from a previous interrupted attempt.
-        Directory.Delete(testsDirectoryName, true);
+        if (Directory.Exists(testsDirectoryName))
+            Directory.Delete(testsDirectoryName, true);
 
         Directory.CreateDirectory(testsDirectoryName);
 


src/Nethermind/Ethereum.Blockchain.Pyspec.Test/LoadPyspecTestsStrategy.cs
```diff
@@ -40,7 +40,64 @@ public IEnumerable<EthereumTest> Load(string testsDir, string wildcard = null)
...
+    /// Falls back to <see cref="AppContext.BaseDirectory"/> if the expected layout is not found.
+    /// </summary>
+    private static string GetFixturesCacheDirectory()
+    {
+        // AppContext.BaseDirectory is typically: .../artifacts/bin/Ethereum.Blockchain.Pyspec.Test/release/
+        // We want:                              .../artifacts/eest/
+        DirectoryInfo dir = new(AppContext.BaseDirectory);
+        while (dir is not null)
+        {
+            if (dir.Parent != null && string.Equals(dir.Name, "bin", StringComparison.OrdinalIgnoreCase)
```
flcl42: Won't it always stop in artifacts/bin never reaching artifacts?

benaadams: Uses `dir.Parent.FullName`

LukaszRozmej:
```suggestion
            if (dir.Parent is not null && string.Equals(dir.Name, "bin", StringComparison.OrdinalIgnoreCase)
```

src/Nethermind/Ethereum.Blockchain.Pyspec.Test/LoadPyspecTestsStrategy.cs
```diff
@@ -40,18 +40,73 @@ public IEnumerable<EthereumTest> Load(string testsDir, string wildcard = null)
...
+        {
+            mutex.ReleaseMutex();
+        }
+    }
+
+    private static void DownloadAndExtract(string archiveVersion, string archiveName, string testsDirectoryName)
     {
+        // Clean up any partial extraction from a previous interrupted attempt.
+        if (Directory.Exists(testsDirectoryName))
+            Directory.Delete(testsDirectoryName, true);
```
LukaszRozmej:
```suggestion
        Directory.Delete(testsDirectoryName, true);
```

benaadams: Turns out this throws `DirectoryNotFoundException`

PR #10726, opened 2026-03-05
XDC Block proposal fixes

alexb5dh:
Fixes for block proposal on XDC

## Changes

- Made block improvement check to use round number instead of TD
- Fix invalid header being used for voting
- Use zero-timeout-certificate as highest by default
- Use correct round number for highest-commit-block
- Log reason for skipping vote
- Improve some nullability checks

#### Requires testing

- [x] Yes

- **In progress**


src/Nethermind/Nethermind.Xdc/XdcHotStuff.cs
```diff
@@ -250,10 +250,15 @@ internal async Task RunRoundChecks(CancellationToken ct)
...
+            if (isMyTurn)
             {
-                _highestSelfMinedRound = currentRound;
-                Task blockBuilder = BuildAndProposeBlock(roundParent, currentRound, spec, ct);
+                // miningParent is null when highestQC block is not in the local chain yet
+                if (FindHeaderToBuildOn(_xdcContext.HighestQC) is { } miningParent &&
+                    IsItTimeToPropose(miningParent, currentRound, spec))
+                {
+                    _highestSelfMinedRound = currentRound;
+                    _ = BuildAndProposeBlock(miningParent, currentRound, spec, ct); // TODO: wait for it to finish?
```
alexb5dh: Wasn't awaited before, should it be?

PR #10734, opened 2026-03-06
perf: eliminate CappedArraySource heap allocation in SpanSource

kamilchodola:
## Changes

- Replace the discriminated-union `object _obj` field in `SpanSource` with explicit `byte[]? _array + int _length` fields
- Remove the `CappedArraySource` sealed class that was heap-allocated on every `SafeRentBuffer` call
- Eliminate type-checking overhead (3 branches) on every `Span`/`Length`/`IsNull` access
- `SpanSource(CappedArray<byte>)` now extracts `UnderlyingArray` + `Length` directly — zero allocations

## Types of changes

- [x] Optimization

## Context

PR #8719 introduced `SpanSource` replacing `RlpFactory`/`CappedArray<byte>` in trie hot paths. The design used a discriminated union over `object _obj` (either `byte[]` or a `CappedArraySource` wrapper class). When `SafeRentBuffer` is called with a pool (the common path during trie node encode/decode), it called `new CappedArraySource(capped)` — a heap allocation on every call. This happens thousands of times per block (every trie node encode/decode), adding significant GC pressure.

Benchmark comparison between `79c0da0` (before #8719) and `eb4a682` (after) showed:
- **AVG block processing: 42.6ms → 51.2ms (+20%)**
- **P90 block processing: 102ms → 135ms (+32%)**

The P90 amplification is consistent with GC pauses from short-lived Gen0 `CappedArraySource` objects.

This fix replaces the single `object _obj` with `byte[]? _array + int _length` (struct grows 8→16 bytes), eliminating all heap allocations and type checking. The `TryGetCappedArray` method reconstructs `CappedArray<byte>` from the stored fields when needed.

#### Requires testing

- [x] Yes

- [x] No (existing tests cover this — all 3 SpanSource tests and 418 Trie tests pass)

#### Notes on testing

- SpanSource unit tests: 3/3 pass
- Trie tests: 418/418 pass (5 skipped)
- Full solution builds with 0 errors, 0 warnings
- Needs benchmark validation to confirm the regression is recovered

### DIFF
--- a/src/Nethermind/Nethermind.Core/Buffers/ICappedArrayPool.cs
+++ b/src/Nethermind/Nethermind.Core/Buffers/ICappedArrayPool.cs
@@ -12,15 +12,8 @@ public interface ICappedArrayPool
 
 public static class CappedArrayPoolExtensions
 {
-    public static CappedArray<byte> SafeRent(this ICappedArrayPool? pool, int size)
-    {
-        if (pool is null)
-        {
-            return new CappedArray<byte>(new byte[size]);
-        }
-
-        return pool.Rent(size);
-    }
+    public static CappedArray<byte> SafeRent(this ICappedArrayPool? pool, int size) =>
+        pool?.Rent(size) ?? new CappedArray<byte>(new byte[size]);
 
     public static void SafeReturn(this ICappedArrayPool? pool, in CappedArray<byte> buffer)
     {


src/Nethermind/Nethermind.Core/Buffers/ICappedArrayPool.cs
```diff
@@ -9,3 +9,24 @@ public interface ICappedArrayPool
...
+{
+    public static CappedArray<byte> SafeRent(this ICappedArrayPool? pool, int size)
+    {
+        if (pool is null)
+        {
+            return new CappedArray<byte>(new byte[size]);
+        }
+
+        return pool.Rent(size);
+    }
```
LukaszRozmej:
```suggestion
    public static CappedArray<byte> SafeRent(this ICappedArrayPool? pool, int size) =>
        pool?.Rent(size) ?? new CappedArray<byte>(new byte[size]);
```

PR #10735, opened 2026-03-06
Snapshot - Big Files & Improvements

### LINKED ISSUE #715: Sepolia Snapshot Support (Big Files)

svlachakis:
Fixes Closes Resolves https://github.com/NethermindEth/nethermind-arbitrum/issues/715

### Bug fixes

- **Resumable downloads** - `HttpClient` auto-redirect was silently stripping the `Range` header on every CDN redirect, causing full re-downloads from byte 0 on every retry. Fixed with a manual redirect loop that preserves the `Range` header.
- **HTTP 416 (range not satisfiable)** is now handled gracefully as *"already complete"* instead of crashing node startup.
- **HttpRequestException (TCP drops, DNS failures)** is now retried - previously only `IOException` was caught, so most real network failures were fatal.
- **Server returning 200 instead of 206 on resume** now skips already-downloaded bytes instead of overwriting the file.

### New functionality

- **`.tar.zst`, `.tar.gz`, `.tar`, `.zst`, `.gz` archive support** - previously only `.zip` was supported; all others threw errors.
- **`StripComponents` config option** — controls `tar --strip-components` to strip leading path components from the archive.

- [ ] Optimization
- [X] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

Tested manually on VM downloading (630GB) and extracting (850GB) `.zst` Arbitrum Sepolia snapshot.

Download time: 3hrs
Extraction time: 1.5hr


LukaszRozmej:
@claude review

src/Nethermind/Nethermind.Init.Snapshot/InitDatabaseSnapshot.cs
```diff
@@ -1,242 +1,182 @@
...
+/// </summary>
 public class InitDatabaseSnapshot : InitDatabase
 {
-    private const int BufferSize = 8192;
-
     private readonly INethermindApi _api;
     private readonly ILogger _logger;
 
-    public InitDatabaseSnapshot(INethermindApi api) : base()
+    public InitDatabaseSnapshot(INethermindApi api)
```
LukaszRozmej: minor -> default constructor

src/Nethermind/Nethermind.Init.Snapshot/InitDatabaseSnapshot.cs
```diff
@@ -1,242 +1,182 @@
...
-                if (_logger.IsInfo)
-                    _logger.Info("Snapshot checksum verified.");
-            }
-            else if (_logger.IsWarn)
-                _logger.Warn("Snapshot checksum is not configured");
-            SetCheckpoint(snapshotConfig, Stage.Verified);
+            if (_logger.IsWarn)
+                _logger.Warn($"Deleting invalid snapshot file '{snapshotPath}' and resetting checkpoint for re-download on next run.");
+            if (File.Exists(snapshotPath))
+                File.Delete(snapshotPath);
```
LukaszRozmej: redundant?

```suggestion
            File.Delete(snapshotPath);
```

src/Nethermind/Nethermind.Init.Snapshot/SnapshotCheckpoint.cs
```diff
@@ -0,0 +1,35 @@
...
+internal sealed class SnapshotCheckpoint(ISnapshotConfig config)
+{
+    private readonly string _path = Path.Combine(config.SnapshotDirectory, $"checkpoint_{config.SnapshotFileName}");
+
+    public SnapshotStage Read()
+    {
+        if (!File.Exists(_path))
+            return SnapshotStage.Started;
+
+        string content = File.ReadAllText(_path).Trim();
```
LukaszRozmej: Just for caution, I would read only first line.

src/Nethermind/Nethermind.Init.Snapshot/SnapshotCheckpoint.cs
```diff
@@ -0,0 +1,35 @@
+// SPDX-FileCopyrightText: 2023 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+namespace Nethermind.Init.Snapshot;
+
+/// <summary>
+/// Persists the current stage of a snapshot initialization to disk so that progress
+/// survives process restarts.
+/// </summary>
+internal sealed class SnapshotCheckpoint(ISnapshotConfig config)
```
LukaszRozmej: Storing enum values in multiple files seems suboptimal.
Can you explain the usage, have you considered other approaches?

src/Nethermind/Nethermind.Init.Snapshot/SnapshotExtractor.cs
```diff
@@ -0,0 +1,109 @@
...
+            ".zst" or ".zstd" => new DecompressionStream(fileStream),
+            ".gz" => new GZipStream(fileStream, CompressionMode.Decompress),
+            _ => fileStream
+        };
+
+    /// <summary>
+    /// Strips <paramref name="count"/> leading path components from <paramref name="entryName"/>.
+    /// Returns <c>null</c> if the entry should be skipped (not enough components).
+    /// </summary>
+    private static string? StripLeadingComponents(string entryName, int count)
```
LukaszRozmej, line 101: minor: Do we need to convert it back to string, can we return `ReadOnlySpan<char>`?

src/Nethermind/Nethermind.Init.Snapshot/SnapshotDownloader.cs
```diff
@@ -0,0 +1,170 @@
...
+            // Server returned the full file, but we already have part of it —
+            // skip the already-downloaded portion in the stream and append the rest.
+            HttpStatusCode.OK when existingSize > 0 =>
+                (FileMode.Append, existingSize, response.Content.Headers.ContentLength),
+
+            HttpStatusCode.OK =>
+                (FileMode.Create, 0L, response.Content.Headers.ContentLength),
+
+            _ => throw new IOException($"Unexpected HTTP status: {response.StatusCode}")
+        };
```
LukaszRozmej: Helper method?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Init.Snapshot/SnapshotCheckpoint.cs
```diff
@@ -0,0 +1,35 @@
...
+        if (!File.Exists(_path))
+            return SnapshotStage.Started;
+
+        string content = File.ReadAllText(_path).Trim();
+        return Enum.TryParse(content, out SnapshotStage stage)
+            ? stage
+            : SnapshotStage.Started;
+    }
+
+    public void Advance(SnapshotStage stage) => File.WriteAllText(_path, stage.ToString());
```
AnkushinDaniil: What happens if the process crashes right after `checkpoint.Advance(SnapshotStage.Extracted)` but before the OS flushes the write to disk? I think that `File.WriteAllText` doesn't guarantee durability, the OS may buffer it for seconds. I'm afraid that in that scenario, the next restart would see a stale checkpoint and could delete a valid database, triggering a full re-download.
Would it make sense to use an atomic write pattern here?
```suggestion
    public void Advance(SnapshotStage stage)
      {
          string tempPath = $"{_path}.tmp";
          using (FileStream fs = new(tempPath, FileMode.Create, FileAccess.Write,
              FileShare.None, 4096, FileOptions.WriteThrough))
          {
              byte[] data = Encoding.UTF8.GetBytes(stage.ToString());
              fs.Write(data);
              fs.Flush(flushToDisk: true);
          }
          File.Move(tempPath, _path, overwrite: true);
      }
```

src/Nethermind/Nethermind.Init.Snapshot/SnapshotDownloader.cs
```diff
@@ -0,0 +1,170 @@
...
+        {
+            // Server honored the Range request — append the remaining bytes.
+            HttpStatusCode.PartialContent =>
+                (FileMode.Append, 0L, existingSize + response.Content.Headers.ContentLength),
+
+            // Server returned the full file, but we already have part of it —
+            // skip the already-downloaded portion in the stream and append the rest.
+            HttpStatusCode.OK when existingSize > 0 =>
+                (FileMode.Append, existingSize, response.Content.Headers.ContentLength),
+
```
AnkushinDaniil: Wouldn't this produce a file containing `[old partial data][full response minus skipped prefix]`? If the partial download was interrupted mid-chunk, the boundaries wouldn't align and the file would be silently corrupt. Checksum would eventually catch it, but only after downloading the full data again.
Would it be safer to just truncate and restart in this case?
```suggestion
            HttpStatusCode.OK when existingSize > 0 =>
              {
                  if (_logger.IsWarn)
                      _logger.Warn("Server does not support range requests. Restarting download from the beginning.");
                  return (FileMode.Create, 0L, response.Content.Headers.ContentLength);
              }
```
The append-with-skip approach seems valid only if we verify via ETag/Last-Modified that the server content hasn't changed between the partial download and the resume, wdyt?

src/Nethermind/Nethermind.Init.Snapshot/InitDatabaseSnapshot.cs
```diff
@@ -1,242 +1,182 @@
...
-            string stringStage = File.ReadAllText(checkpointPath);
-            return Enum.Parse<Stage>(stringStage);
-        }
-        else
-        {
-            return Stage.Start;
-        }
+        byte[] expected = Bytes.FromHexString(expectedChecksum);
+        await using FileStream fileStream = File.OpenRead(filePath);
+        byte[] actual = await SHA256.HashDataAsync(fileStream, cancellationToken).ConfigureAwait(false);
```
AnkushinDaniil: `HA256.HashDataAsync` could take minutes depending on disk speed. During that time there's no log output. Have you thought about using `IncrementalHash` with periodic progress logging?

src/Nethermind/Nethermind.Init.Snapshot/InitDatabaseSnapshot.cs
```diff
@@ -1,242 +1,182 @@
...
+            {
+                if (_logger.IsError)
+                    _logger.Error($"Snapshot download failed with permanent HTTP error {(int?)e.StatusCode}. Aborting.");
+                throw;
+            }
+            catch (Exception e) when (e is IOException or HttpRequestException)
+            {
+                if (_logger.IsError)
+                    _logger.Error($"Snapshot download failed. Retrying in 5 seconds. Error: {e}");
+                await Task.Delay(TimeSpan.FromSeconds(5), cancellationToken).ConfigureAwait(false);
```
AnkushinDaniil: Would it make sense to use an exponential backoff instead of a fixed 5-second delay?

src/Nethermind/Nethermind.Init.Snapshot/InitDatabaseSnapshot.cs
```diff
@@ -1,242 +1,182 @@
...
+                _logger.Warn($"Deleting invalid snapshot file '{snapshotPath}' and resetting checkpoint for re-download on next run.");
+            if (File.Exists(snapshotPath))
+                File.Delete(snapshotPath);
+            checkpoint.Advance(SnapshotStage.Started);
+            return;
         }
 
-        await ExtractSnapshotTo(snapshotFileName, dbPath, cancellationToken);
-        SetCheckpoint(snapshotConfig, Stage.Extracted);
+        await ExtractAsync(snapshotPath, dbPath, snapshotConfig.StripComponents, checkpoint, cancellationToken).ConfigureAwait(false);
```
AnkushinDaniil, line 88: It seems that a quick free-space check before starting extraction could save a lot of pain

src/Nethermind/Nethermind.Init.Snapshot/InitDatabaseSnapshot.cs
```diff
@@ -1,242 +1,182 @@
...
-                _logger.Info($"Extracting snapshot to {dbPath}. Do not interrupt!");
+                _logger.Info($"Verifying snapshot checksum {config.Checksum}.");
 
-            ZipFile.ExtractToDirectory(snapshotPath, dbPath);
-        }, cancellationToken);
+            bool valid = await ComputeAndCompareChecksumAsync(snapshotPath, config.Checksum, cancellationToken).ConfigureAwait(false);
+            if (!valid)
+            {
+                if (_logger.IsError)
+                    _logger.Error("Snapshot checksum verification failed. Aborting snapshot initialization, but the node will continue running.");
```
AnkushinDaniil: Should expected and actual values be included into the log?

src/Nethermind/Nethermind.Init.Snapshot/ISnapshotConfig.cs
```diff
@@ -21,4 +21,7 @@ public interface ISnapshotConfig : IConfig
 
     [ConfigItem(Description = "The name of the snapshot file.", DefaultValue = "snapshot.zip")]
     public string SnapshotFileName { get; set; }
+
+    [ConfigItem(Description = "Number of leading path components to strip when extracting a tar archive (passed as --strip-components to tar). Set this to match the depth of the snapshot path embedded in the archive.", DefaultValue = "1")]
+    public int StripComponents { get; set; }
```
AnkushinDaniil: I'd add bounds check for it

src/Nethermind/Nethermind.Init.Snapshot/SnapshotExtractor.cs
```diff
@@ -0,0 +1,109 @@
...
+
+    private static void ExtractTar(string archivePath, string destinationPath, string extension, int stripComponents, CancellationToken cancellationToken)
+    {
+        Directory.CreateDirectory(destinationPath);
+
+        using FileStream fileStream = File.OpenRead(archivePath);
+        using Stream decompressedStream = OpenDecompressedStream(fileStream, extension);
+        using TarReader tarReader = new(decompressedStream);
+
+        string destinationRoot = destinationPath.TrimEnd(Path.DirectorySeparatorChar) + Path.DirectorySeparatorChar;
```
AnkushinDaniil, line 64: Does it correctly work on windows?

src/Nethermind/Nethermind.Init.Snapshot/SnapshotCheckpoint.cs
svlachakis: Fixed - now uses `File.ReadLines(_path).FirstOrDefault()?.Trim()` which reads only the first line without loading the whole file.

src/Nethermind/Nethermind.Init.Snapshot/SnapshotDownloader.cs
svlachakis: Fixed, extracted into a ResolveCopyStrategy private static method.

src/Nethermind/Nethermind.Init.Snapshot/SnapshotCheckpoint.cs
svlachakis: Fixed. Advance now writes to a `.tmp` file with `FileOptions.WriteThrough`, calls `Flush(flushToDisk: true)`, then `File.Move(overwrite: true)`. A crash mid-write will leave the old checkpoint intact.

src/Nethermind/Nethermind.Init.Snapshot/InitDatabaseSnapshot.cs
svlachakis: Fixed, the delay now starts at 5s and doubles on each failure, capped at 300s (5 minutes). The current delay is included in the log message.

src/Nethermind/Nethermind.Init.Snapshot/InitDatabaseSnapshot.cs
svlachakis: Fixed

src/Nethermind/Nethermind.Init.Snapshot/InitDatabaseSnapshot.cs
svlachakis: done

src/Nethermind/Nethermind.Init.Snapshot/ISnapshotConfig.cs
svlachakis: done

src/Nethermind/Nethermind.Init.Snapshot/SnapshotExtractor.cs
svlachakis: Yes, it should work correctly on Windows. `Path.GetFullPath` normalizes separators to the OS convention (\ on Windows), and `Path.DirectorySeparatorChar` produces the same character.

Both `destinationEntryPath` and `destinationRoot` are produced by `Path.GetFullPath`, so `StartsWith` compares two strings in the same normalized form regardless of OS.

src/Nethermind/Nethermind.Init.Snapshot/SnapshotCheckpoint.cs
svlachakis: There is only ever one checkpoint file per snapshot, named checkpoint_{snapshotFileName} (e.g.
checkpoint_snapshot.tar.gz), stored in the same snapshot directory as the download.

It contains a single line: the name of the last completed stage.

I did consider alternatives:

1. Marker files (e.g. snapshot.tar.gz.downloaded) - presence/absence as the signal. No parsing needed, but leaves multiple files to clean up and is harder to extend if stages change.
2. Inferring state from disk (file size vs expected, db directory exists, etc.) - no extra file at all, but fragile: we can't distinguish "downloaded" from "verified" without re-hashing on every restart.
3. Embedding stage in the snapshot filename - same problem as marker files.

I think a single plain-text file with an enum value seemed like the right balance: one extra file, zero schema, human-readable, easy to delete to force a full restart. 

Happy to change the approach if you have a preference.

src/Nethermind/Nethermind.Init.Snapshot/SnapshotExtractor.cs
svlachakis: The result is passed directly to `Path.Combine(destinationPath, strippedPath)`, which has no `ReadOnlySpan<char> `overload - it requires a string.

Returning a span would just push the `.ToString()` call to the call site with no allocation saved.

Additionally, `ReadOnlySpan<char>` can't be nullable, so we'd lose the clean null-means-skip return and have to use `.IsEmpty`, which collapses two distinct cases ("not enough components" vs "empty component after stripping") into one.

src/Nethermind/Nethermind.Init.Snapshot/SnapshotDownloader.cs
svlachakis: The boundary concern isn't quite right I think. The skip is byte-exact (existingSize bytes from the stream, matching the file size on disk), so chunk boundaries don't factor in.

The file would be correct regardless of where the previous download was interrupted.

The real risk is content change: if the server's file changed between the partial download and the resume, we'd splice old bytes with new bytes, producing a corrupt file.

The checksum catches this, but only after downloading the full response. An ETag/Last-Modified check could detect it upfront.

That said, we've gone with truncate-and-restart on 200 since it's simpler.

The trade-off is that for servers that never support range requests, resume is effectively disabled - each retry starts from zero.

I think that maybe we could add an ETag check in the future if that ever becomes an issue..

src/Nethermind/Nethermind.Init.Snapshot/InitDatabaseSnapshot.cs
svlachakis: Implemented. `ComputeChecksumAsync` now uses `IncrementalHash` fed in 64 KB `ArrayPool` chunks, with a progress log every 30 seconds (Snapshot checksum progress: X%).

The hash result is identical; only the visibility changes.

src/Nethermind/Nethermind.Init.Snapshot/InitDatabaseSnapshot.cs
```diff
@@ -1,242 +1,226 @@
...
-            await Task.Delay(TimeSpan.FromSeconds(5), cancellationToken);
+            try
+            {
+                await downloader.DownloadAsync(url, destinationPath, cancellationToken).ConfigureAwait(false);
+                break;
+            }
+            catch (HttpRequestException e) when (
+                e.StatusCode is >= HttpStatusCode.BadRequest and < HttpStatusCode.InternalServerError
+                    and not HttpStatusCode.TooManyRequests
+                    and not HttpStatusCode.RequestedRangeNotSatisfiable)
```
flcl42, line 118: Can we still try to download without passing range if range error happens?

flcl42 (approved):

src/Nethermind/Nethermind.Init.Snapshot/InitDatabaseSnapshot.cs
svlachakis: Yes

damian-orzechowski (approved):

PR #10736, opened 2026-03-06
fix: correct RocksDB MaxOpenFiles formula to divide by DB count

kamilchodola:
## Summary

- Fix RocksDB `MaxOpenFiles` auto-limit formula to properly divide by estimated database count (~15)
- The OS file descriptor limit is shared across all databases, but `MaxOpenFiles` is set per database

### Problem

PR #9504 introduced auto-detection of the system open file limit to prevent macOS crashes during sync. The formula `perDbLimit = 80% × systemLimit` was applied **per database**, but with ~15 databases the combined limit was `15 × 0.8 × limit = 12× the OS limit`. This:

1. Didn't actually prevent the macOS crash it was designed to fix
2. Caused a ~43% P90 block processing regression by setting an unnecessarily high per-DB limit that still forced RocksDB into LRU file handle caching mode

### Fix

Divide by estimated DB count: `perDbLimit = max(128, systemLimit × 0.8 / 15)`

| System limit | Before (per DB) | Before (combined) | After (per DB) | After (combined) |
|---|---|---|---|---|
| 256 (macOS) | 256 | 3840 ❌ | 128 | 1920 ✅ |
| 1024 (low Linux) | 819 | 12285 ❌ | 128 | 1920 ✅ |
| 65536 (server) | 52428 | 786420 ❌ | 3495 | 52425 ✅ |

## Changes

## Test plan

- Updated `RocksDbConfigFactoryTests` — all 11 tests pass
- Benchmark validation: run `realblocks` payload set to verify P90 recovery


asdacap:
You cant divide it by num of DB. Some db take up a lot of file, others dont. Thats why per db it is 80%.

LukaszRozmej (approved):

asdacap (approved):

benaadams (approved):

kamilchodola:
@claude can you do review of this PR once more to check if it can introduce any regresison on RocksDB Get method on Ubuntu 24?

PR #10739, opened 2026-03-06
Expand retry cache

flcl42:
## Changes

- Widen the cache for the case of just synced node getting lots of new tx hashes

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.TxPool/RetryCache.cs
+++ b/src/Nethermind/Nethermind.TxPool/RetryCache.cs
@@ -112,9 +112,9 @@ public sealed class RetryCache<TMessage, TResourceId> : IAsyncDisposable
             return AnnounceResult.RequestRequired;
         }
 
-        if (_expiringQueueCounter > _expiringQueueLimit)
+        if (_expiringQueueCounter >= _expiringQueueLimit)
         {
-            if (_logger.IsDebug) _logger.Warn($"{nameof(TResourceId)} retry queue is full");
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.TxPool/RetryCache.cs
```diff
@@ -112,9 +112,9 @@ public AnnounceResult Announced(in TResourceId resourceId, IMessageHandler<TMess
             return AnnounceResult.RequestRequired;
         }
 
-        if (_expiringQueueCounter > _expiringQueueLimit)
+        if (_expiringQueueCounter >= _expiringQueueLimit)
         {
-            if (_logger.IsDebug) _logger.Warn($"{nameof(TResourceId)} retry queue is full");
+            if (_logger.IsTrace) _logger.Trace($"{nameof(TResourceId)} retry queue is full");
```
benaadams: These should be `typeof` not `nameof`; did a fix in https://github.com/NethermindEth/nethermind/pull/10753
```suggestion
            if (_logger.IsTrace) _logger.Trace($"{typeof(TResourceId)} retry queue is full");
```

flcl42: +.Name

LukaszRozmej (approved):

benaadams (approved):

svlachakis (approved):

PR #10741, opened 2026-03-06
fix: Make method non-virtual for zkEVM

rubo:
## Changes

Made the `HeaderValidator.Validate` method non-virtual for zkEVM. This class is now used for block validation in stateless execution.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

Tested manually in Zisk

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Validators/HeaderValidator.cs
+++ b/src/Nethermind/Nethermind.Consensus/Validators/HeaderValidator.cs
@@ -74,11 +74,10 @@ namespace Nethermind.Consensus.Validators
         public bool Validate(BlockHeader header, BlockHeader parent, bool isUncle, out string? error) =>
             Validate<OffFlag>(header, parent, isUncle, out error);
 
-        protected
 #if !ZK_EVM
         virtual
 #endif
-        bool Validate<TOrphaned>(BlockHeader header, BlockHeader? parent, bool isUncle, out string? error) where TOrphaned : struct, IFlag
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Consensus/Validators/HeaderValidator.cs
```diff
@@ -74,7 +74,11 @@ public bool Validate(BlockHeader header, BlockHeader parent, bool isUncle = fals
         public bool Validate(BlockHeader header, BlockHeader parent, bool isUncle, out string? error) =>
             Validate<OffFlag>(header, parent, isUncle, out error);
 
-        protected virtual bool Validate<TOrphaned>(BlockHeader header, BlockHeader? parent, bool isUncle, out string? error) where TOrphaned : struct, IFlag
+        protected
+#if !ZK_EVM
+        virtual
+#endif
+        bool Validate<TOrphaned>(BlockHeader header, BlockHeader? parent, bool isUncle, out string? error) where TOrphaned : struct, IFlag
```
LukaszRozmej: a bit less ugly:

```suggestion       
#if !ZK_EVM
        virtual
#endif
        protected bool Validate<TOrphaned>(BlockHeader header, BlockHeader? parent, bool isUncle, out string? error) where TOrphaned : struct, IFlag
```

PR #10745, opened 2026-03-07
refactor: remove NethermindConstructorFinder, use explicit factory registrations

kamilchodola:
## Changes

Removes `FindConstructorsWith(NethermindConstructorFinder.Instance)` from `CommonNethermindConfig()`, which was applied to every Autofac registration globally. Types that relied on `[UseConstructorForDependencyInjection]` for constructor disambiguation now use explicit factory registrations instead.

### What changed
- **`ContainerBuilderExtensions.cs`**: Removed `FindConstructorsWith(NethermindConstructorFinder.Instance)` from `CommonNethermindConfig()`
- **`ChainHeadInfoProvider.cs`**: Removed `[UseConstructorForDependencyInjection]` attribute, added explicit factory in `NethermindModule.cs`
- **`TxPoolInfoProvider.cs`**: Removed attribute, explicit factory already existed in `RpcModules.cs`
- **`SyncPeerPool.cs`**: Removed attribute, added explicit factory in `Synchronizer.cs`
- **`TraceStorePruner.cs`**: Removed attribute (single constructor used by DI)
- **`NethermindConstructorFinderTests.cs`**: Deleted (tests the removed feature)

### Why

The `NethermindConstructorFinder` uses reflection + LINQ (`GetCustomAttribute<>()` + `.Where().ToArray()`) to find constructors marked with `[UseConstructorForDependencyInjection]`. It was registered globally via `CommonNethermindConfig()` on every DI registration, but only 4 types actually used the attribute.

Debug logging confirmed it fires 374 times (332 at startup + 42 during initial block processing via child DI scopes). Autofac caches results after first call per registration, and the prewarmer pools its DI scopes, so steady-state block processing sees zero calls.

Explicit factory registrations are simpler, more direct, and don't require a global reflection-based constructor finder applied to all 264+ registered types.

### Benchmark results (10x master vs 10x fix, rebased)

| Metric | Master (10 runs) | Fix (10 runs) |
|--------|-----------------|---------------|
| AVG | 39.25ms (σ=0.83) | 39.64ms (σ=0.44) |
| MED | 20.72ms | 20.86ms |

**Performance neutral** — within noise at 10-sample size. This is a code quality / simplification change, not a performance improvement.

### Types of changes

- [x] Refactoring

## Testing

- [x] Existing tests pass (`SyncPeerPool_should_use_INetworkConfig_MaxActivePeers` fixed via explicit factory)
- [x] Deleted `NethermindConstructorFinderTests` (tested the removed feature)
- [x] Verified via debug logging: 0 `FindConstructors` calls on fix branch vs 374 on master
- [x] 10x10 benchmark comparison confirms no regression


src/Nethermind/Nethermind.Core/Container/NethermindConstructorFinder.cs
```diff
@@ -21,6 +21,9 @@ private NethermindConstructorFinder()
 
     public ConstructorInfo[] FindConstructors(Type targetType)
     {
+        Console.Error.WriteLine($"[NethermindConstructorFinder] FindConstructors called for {targetType.Name}");
+        Console.Error.Flush();
```
benaadams: Shouldn't use `Console` probably shouldn't log here either?

kamilchodola: Correct — this is a temporary debug line added to verify when FindConstructors is called during runtime (startup vs block processing). Will be removed before merging. Was added specifically for the investigation — results in the PR comments above.

PR #10752, opened 2026-03-08
fix(dns): correct Base64 padding calculation in EnrRecordParser

zeroprooff:
Fixed incorrect padding logic in AddPadding method. The old code added `mod4` padding characters instead of `(4 - mod4) % 4`. This caused 3 padding chars to be added when length % 4 == 3, but Base64 only allows max 2. The variable was also misleadingly named `mod3` while computing % 4.


MarekM25 (changes requested):
No tests that prove the issue

PR #10759, opened 2026-03-09
Use generic decoding results for block and tx; remove IRlpObjectDecoder; single RlpDecoder base class and interface

flcl42:
## Changes

- Use generic decoding results for block and tx
- Remove IRlpObjectDecoder
- Use generic decoding
- Single RlpDecoder base class and interface
- Fix null-array RLP length/encoding consistency

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No


src/Nethermind/Nethermind.Core.Test/Encoding/CompactReceiptDecoderTests.cs
```diff
@@ -274,5 +274,19 @@ private void AssertStorageLegacyReceipt(TxReceipt txReceipt, TxReceipt deseriali
...
+            RlpStream stream = new(decoder.GetLength(txReceipt, behaviors));
+            decoder.Encode(stream, txReceipt, behaviors);
+            return new Rlp(stream.Data.ToArray()!);
+        }
+
+        private static Rlp EncodeReceipts(CompactReceiptStorageDecoder decoder, TxReceipt[] receipts, RlpBehaviors behaviors = RlpBehaviors.None)
+        {
+            RlpStream stream = new(decoder.GetLength(receipts, behaviors));
+            decoder.Encode(stream, receipts, behaviors);
+            return new Rlp(stream.Data.ToArray()!);
```
LukaszRozmej: Can we replace all those helper methods with one generic one on `Rlp`?

src/Nethermind/Nethermind.Core.Test/Builders/TestItem.cs
```diff
@@ -186,10 +186,17 @@ public static byte[] GenerateIndexedAccountRlp(int index, AccountDecoder? accoun
             accountDecoder ??= _accountDecoder;
 
             Account account = GenerateIndexedAccount(index);
-            byte[] value = accountDecoder.Encode(account).Bytes;
+            byte[] value = EncodeAccount(accountDecoder, account);
             return value;
         }
 
+        private static byte[] EncodeAccount(AccountDecoder accountDecoder, Account account)
```
LukaszRozmej: this one too

src/Nethermind/Nethermind.Core.Test/Encoding/LogEntryDecoderTests.cs
```diff
@@ -50,6 +50,13 @@ public void Can_do_roundtrip(bool valueDecode, bool useDecoderInstance)
         AssertLogEntriesEqual(logEntry, decoded);
     }
 
+    private static Rlp EncodeLogEntry(LogEntryDecoder decoder, LogEntry logEntry)
```
LukaszRozmej: this one too

src/Nethermind/Nethermind.Core.Test/Encoding/ReceiptDecoderTests.cs
```diff
@@ -299,5 +299,19 @@ private void AssertStorageReceipt(TxReceipt txReceipt, TxReceipt? deserialized)
...
+            RlpStream stream = new(decoder.GetLength(txReceipt, behaviors));
+            decoder.Encode(stream, txReceipt, behaviors);
+            return new Rlp(stream.Data.ToArray()!);
+        }
+
+        private static Rlp EncodeReceipts(ReceiptStorageDecoder decoder, TxReceipt[] receipts, RlpBehaviors behaviors = RlpBehaviors.None)
+        {
+            RlpStream stream = new(decoder.GetLength(receipts, behaviors));
+            decoder.Encode(stream, receipts, behaviors);
+            return new Rlp(stream.Data.ToArray()!);
```
LukaszRozmej: this one too

src/Nethermind/Nethermind.Db.Test/CompressingStoreTests.cs
```diff
@@ -56,7 +56,7 @@ public void EOA()
     {
         Context ctx = new();
 
-        Rlp encoded = new AccountDecoder().Encode(new(1));
+        Rlp encoded = Rlp.Encode((Account)new(1));
```
LukaszRozmej:
```suggestion
        Rlp encoded = Rlp.Encode(new Account(1));
```

src/Nethermind/Nethermind.Db.Test/CompressingStoreTests.cs
```diff
@@ -68,7 +68,7 @@ public void EOAWithSPan()
...
```
LukaszRozmej: same

src/Nethermind/Nethermind.Merge.Plugin.Test/SszRest/SszCodecTests.cs
```diff
@@ -162,7 +162,7 @@ private static IEnumerable<TestCaseData> NonEmptyEncodings()
         yield return new TestCaseData((Action<IBufferWriter<byte>>)(w =>
         {
             ExecutionPayloadV3 ep = SszTestData.MakeV3Payload();
-            Block block = ep.TryGetBlock().Block!;
+            Block block = ep.TryGetBlock().Data!;
```
LukaszRozmej: Explicit casts - better - exception will be thrown directly
```suggestion
            Block block = (Block)ep.TryGetBlock();
```

LukaszRozmej (commented):
If we are not checking if `Result` is successfull or not - do explicit casting

src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
```diff
@@ -281,15 +281,22 @@ public static Rlp Encode<T>(T item, RlpBehaviors behaviors = RlpBehaviors.None)
...
             if (rlpStreamEncoder is not null)
             {
-                int totalLength = rlpStreamEncoder.GetLength(item, behaviors);
-                RlpStream stream = new(totalLength);
-                rlpStreamEncoder.Encode(stream, item, behaviors);
-                return new Rlp(stream.Data.ToArray());
+                return Encode(rlpStreamEncoder, item, behaviors);
             }
 
             throw new RlpException($"{nameof(Rlp)} does not support encoding {typeof(T).Name}");
```
LukaszRozmej: this should be one expression

src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
```diff
@@ -281,15 +281,22 @@ public static Rlp Encode<T>(T item, RlpBehaviors behaviors = RlpBehaviors.None)
...
         }
 
+        public static Rlp Encode<T>(IRlpStreamEncoder<T> rlpStreamEncoder, T item, RlpBehaviors behaviors = RlpBehaviors.None)
+        {
+            ArgumentNullException.ThrowIfNull(rlpStreamEncoder);
+
+            int totalLength = rlpStreamEncoder.GetLength(item, behaviors);
+            RlpStream stream = new(totalLength);
+            rlpStreamEncoder.Encode(stream, item, behaviors);
+            return new Rlp(stream.Data.ToArray());
```
LukaszRozmej: Can we avoid allocating `Rlp` and array?

src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
```diff
@@ -300,15 +307,22 @@ public static Rlp Encode<T>(T[] items, RlpBehaviors behaviors = RlpBehaviors.Non
...
             if (rlpStreamEncoder is not null)
             {
-                int totalLength = rlpStreamEncoder.GetLength(items, behaviors);
-                RlpStream stream = new(totalLength);
-                rlpStreamEncoder.Encode(stream, items, behaviors);
-                return new Rlp(stream.Data.ToArray());
+                return Encode(rlpStreamEncoder, items, behaviors);
             }
 
             throw new RlpException($"{nameof(Rlp)} does not support encoding {typeof(T).Name}");
```
LukaszRozmej: one expression

src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
```diff
@@ -300,15 +307,22 @@ public static Rlp Encode<T>(T[] items, RlpBehaviors behaviors = RlpBehaviors.Non
...
         }
 
+        public static Rlp Encode<T>(IRlpStreamEncoder<T> rlpStreamEncoder, T[] items, RlpBehaviors behaviors = RlpBehaviors.None)
+        {
+            ArgumentNullException.ThrowIfNull(rlpStreamEncoder);
+
+            int totalLength = rlpStreamEncoder.GetLength(items, behaviors);
+            RlpStream stream = new(totalLength);
+            rlpStreamEncoder.Encode(stream, items, behaviors);
+            return new Rlp(stream.Data.ToArray());
```
LukaszRozmej: same here

LukaszRozmej (approved):

benaadams (approved):

svlachakis:
@claude please review the following minor points:

1. IRlpDecoder<T> is misnamed and violates ISP — Nethermind.Serialization.Rlp/IRlpDecoder.cs:14-22.

The interface called “Decoder” carries only GetLength(T, RlpBehaviors) — a pure encoder concern. IRlpStreamEncoder<T> (an encoder) inherits from it; IRlpValueDecoder<T> (the real decoder) also inherits from it. A pure decoder consumer is therefore forced to depend on GetLength.

This PR was the natural moment to rename it to IRlpSizable<T> (or fold GetLength into IRlpStreamEncoder<T> and make IRlpValueDecoder<T> standalone). It removes one bad abstraction (IRlpObjectDecoder) while leaving the deeper one in place.

2. Two public Encode<T> overloads were added for a single test caller — Rlp.cs:281,291.

Both are one-line pass-throughs to EncodeWithStreamEncoder. At the call site, Rlp.Encode(decoder, item) reads as “encode both decoder and item”, but actually means “use decoder to encode item”.

This is speculative generality combined with an ambiguous verb. Inline these into the private helper or rename them to EncodeWith / EncodeUsing.

3. RlpDecoderExtensions.GetContentLength silently fixes a behavior bug in a PR labeled “refactor / no testing” — RlpDecoderExtensions.cs:213-218.

The old code called decoder.GetLength(items[i]) for null items; the new code short-circuits to Rlp.OfEmptyList.Length. This is now internally consistent with the array Encode loop (which writes Rlp.OfEmptyList for nulls), so this is a real correctness fix for any decoder where GetLength(null) != 1.

The PR description says “no testing required”, but this change should be called out explicitly and covered with a regression test.

LukaszRozmej:
## Review: state at HEAD of `improve-decoding`

### What's already addressed since earlier reviews
- `PendingValidatorsDecoder.Encode(RlpStream, …)` null path now calls `EncodeNullObject()` — earlier NRE regression fixed.
- `LogEntryDecoder` legacy `Rlp Encode(LogEntry?)` removed; `Rlp.Encode(LogEntry)` static overload deleted.
- The speculative `Rlp.Encode<T>(IRlpStreamEncoder<T>, T …)` overloads from earlier commits did not land.
- `IRlpDecoderTests.Array_encoding_uses_empty_list_for_null_withdrawals` regression-tests the `GetContentLength` null-array consistency fix.

### Still worth raising

**🟡 Medium — `CompactLogEntryDecoder.Decode(…)!` hides a contract gap**

```csharp
LogEntry IRlpDecoder<LogEntry>.Decode(ref Rlp.ValueDecoderContext ctx, RlpBehaviors b)
    => Decode(ref ctx, b)!;
```

The underlying `Decode` returns `LogEntry?` and returns `null` for an empty-list item, but the explicit interface impl claims non-null `LogEntry` via `!`. A caller through `IRlpDecoder<LogEntry>` will silently get `null` despite the type. `LogEntryDecoder` has the same shape (`RlpDecoder<LogEntry>` with `DecodeInternal` returning `LogEntry?`). Prefer `IRlpDecoder<LogEntry?>` / `RlpDecoder<LogEntry?>` so nullability is honest, or fail-fast in the explicit impl.

**🟡 Medium — Behavior fix in a "no testing required" PR**

The `GetContentLength(T?[]?)` change to `items[i] is null ? Rlp.OfEmptyList.Length : decoder.GetLength(items[i])` is a real correctness fix for any decoder where `GetLength(null) ≠ 1`. A regression test is now in place, but the PR description still has *Requires testing → No*. Worth ticking *Bugfix* alongside *Refactoring* so this is discoverable in the changelog and bisect history.

**🟡 Medium — Hot-path encoding now dispatches through DIM**

`Rlp.Encode<T>(T)` routes through `GetDecoder<T>().Encode(item, behaviors)` where `Encode(T) → Rlp` is a default interface method on `IRlpDecoder<T>`. That's a permanent interface dispatch on encode paths that previously had direct sealed-class calls. Probably negligible with PGO, but it's worth a `Nethermind.Benchmarks.Rlp` run before merge. If it does move, the fix is cheap: surface the DIMs as instance methods on `RlpDecoder<T>` so concrete callers devirtualize.

**🔵 Low — `IRlpDecoder<T>` is no longer "just a decoder"**

The flattened interface now bundles ~20 members covering `GetLength`, `Encode(RlpStream)`, `Encode(T) → Rlp`, `Decode`, `DecodeArray`, `DecodeComplete`, `DecodeGuardNotNull`, four `EncodeToNewNettyStream` overloads, `EncodeToCappedArray`, plus array `Encode`/`GetContentLength`/`GetLength`. Half of those are encoder concerns; the name is now actively misleading. Rename out of scope here, but `IRlpCodec<T>` / `IRlpSerializer<T>` would describe the shape honestly — worth a follow-up.

**🔵 Low — `CompactLogEntryDecoder` skips `RlpDecoder<T>`**

It implements `IRlpDecoder<LogEntry>` directly with `static` Encode/GetLength and forwards via explicit interface impls. As a result it doesn't get the `IndexOutOfRangeException`/`ArgumentOutOfRangeException` → `RlpException` wrapping that `RlpDecoder<T>.Decode` provides. Likely intentional given the static call surface, but inconsistent with every other decoder.

**🔵 Low — SPDX years**

New files use `2026`. Existing files touched purely for the refactor that now read `2026` are slightly off-spec (SPDX year = first publication). Trivial.

### Verdict

Two Mediums worth resolving before merge: honest nullability on `LogEntry`/`CompactLogEntryDecoder` interface impls, and updating the PR description to acknowledge the array-null behavior fix. The DIM-dispatch concern can be a follow-up gated on benchmark numbers.

flcl42:
@claude review again

src/Nethermind/Nethermind.Blockchain/Blocks/BadBlockStore.cs
```diff
@@ -13,7 +13,7 @@ namespace Nethermind.Blockchain.Blocks;
 
 public class BadBlockStore(IDb blockDb, long maxSize) : IBadBlockStore
 {
-    private readonly BlockDecoder _blockDecoder = new();
+    private readonly IRlpDecoder<Block> _blockDecoder = new BlockDecoder();
```
benaadams: Why interface and not concrete type?

src/Nethermind/Nethermind.Blockchain/Blocks/BlockStore.cs
```diff
@@ -12,23 +12,33 @@
...
 
-public class BlockStore([KeyFilter(DbNames.Blocks)] IDb blockDb, IHeaderDecoder headerDecoder = null) : IBlockStore, IClearableCache
+public class BlockStore : IBlockStore, IClearableCache
 {
-    private readonly BlockDecoder _blockDecoder = new(headerDecoder ?? new HeaderDecoder());
     public const int CacheSize = 128 + 32;
 
+    private readonly IDb _blockDb;
+    private readonly BlockDecoder _blockDecoder;
+    private readonly IRlpDecoder<Block> _blockRlpDecoder;
```
benaadams: Why interface and not concrete type? Only one constructor and it creates concrete type

src/Nethermind/Nethermind.Blockchain/Receipts/PersistentReceiptStorage.cs
```diff
@@ -29,7 +29,8 @@ public class PersistentReceiptStorage : IReceiptStorage
         private readonly IDb _transactionDb;
         private static readonly Hash256 MigrationBlockNumberKey = Keccak.Compute(nameof(MigratedBlockNumber));
         private long _migratedBlockNumber;
-        private readonly ReceiptArrayStorageDecoder _storageDecoder = ReceiptArrayStorageDecoder.Instance;
+        private readonly ReceiptArrayStorageDecoder _storageDecoder;
+        private readonly IRlpDecoder<TxReceipt[]> _storageRlpDecoder;
```
benaadams: Why interface and not concrete type?

src/Nethermind/Nethermind.Blockchain/Receipts/ReceiptsRootCalculator.cs
```diff
@@ -12,8 +12,8 @@ namespace Nethermind.Blockchain.Receipts;
 public class ReceiptsRootCalculator : IReceiptsRootCalculator
 {
     public static readonly ReceiptsRootCalculator Instance = new();
-    private static readonly IRlpStreamEncoder<TxReceipt> _decoder = Rlp.GetStreamEncoder<TxReceipt>(RlpDecoderKey.Trie)!;
-    private static readonly IRlpStreamEncoder<TxReceipt> _skipStateDecoder = new ReceiptMessageDecoder(skipStateAndStatus: true);
+    private static readonly IRlpDecoder<TxReceipt> _decoder = Rlp.GetDecoder<TxReceipt>(RlpDecoderKey.Trie)!;
+    private static readonly IRlpDecoder<TxReceipt> _skipStateDecoder = new ReceiptMessageDecoder(skipStateAndStatus: true);
```
benaadams: Why interface and not concrete type?

src/Nethermind/Nethermind.Consensus.AuRa/Validators/ValidatorStore.cs
```diff
@@ -15,7 +15,7 @@ public class ValidatorStore : IValidatorStore
     {
         internal static readonly Hash256 LatestFinalizedValidatorsBlockNumberKey = Keccak.Compute("LatestFinalizedValidatorsBlockNumber");
         internal static readonly Hash256 PendingValidatorsKey = Keccak.Compute("PendingValidators");
-        private static readonly PendingValidatorsDecoder PendingValidatorsDecoder = new();
+        private static readonly IRlpDecoder<PendingValidators> PendingValidatorsDecoder = new PendingValidatorsDecoder();
```
benaadams: Why interface and not concrete type?

src/Nethermind/Nethermind.Consensus.Clique/SnapshotManager.cs
```diff
@@ -222,25 +222,17 @@ private static Hash256 GetSnapshotKey(Hash256 blockHash)
             return new Hash256(keyBytes);
         }
 
-        private readonly SnapshotDecoder _decoder = new();
+        private readonly IRlpDecoder<Snapshot> _decoder = new SnapshotDecoder();
```
benaadams: Why interface and not concrete type?

src/Nethermind/Nethermind.Era1/EraReader.cs
```diff
@@ -21,7 +21,7 @@ namespace Nethermind.Era1;
 /// </summary>
 public class EraReader(E2StoreReader e2) : IAsyncEnumerable<(Block, TxReceipt[])>, IDisposable
 {
-    private readonly ReceiptMessageDecoder _receiptDecoder = new();
+    private readonly IRlpDecoder<TxReceipt> _receiptDecoder = new ReceiptMessageDecoder();
```
benaadams: Why interface and not concrete type?

src/Nethermind/Nethermind.Era1/EraWriter.cs
```diff
@@ -19,9 +19,9 @@ public class EraWriter : IDisposable
     private long _totalWritten;
     private readonly ArrayPoolList<long> _entryIndexes;
 
-    private readonly HeaderDecoder _headerDecoder = new();
-    private readonly BlockBodyDecoder _blockBodyDecoder = new();
-    private readonly ReceiptMessageDecoder _receiptDecoder = new();
+    private readonly IRlpDecoder<BlockHeader> _headerDecoder = new HeaderDecoder();
+    private readonly IRlpDecoder<BlockBody> _blockBodyDecoder = BlockBodyDecoder.Instance;
+    private readonly IRlpDecoder<TxReceipt> _receiptDecoder = new ReceiptMessageDecoder();
```
benaadams: Why interfaces and not concrete type?

src/Nethermind/Nethermind.EraE/Archive/EraWriter.cs
```diff
@@ -36,8 +36,8 @@ public sealed class EraWriter : IDisposable
     private const int BeaconSlotSeconds = 12;
     private const int IndexFieldSize = 8; // sizeof(long) — each ComponentIndex field is a little-endian int64
 
-    private readonly HeaderDecoder _headerDecoder = new();
-    private readonly BlockBodyDecoder _blockBodyDecoder = new();
+    private readonly IRlpDecoder<BlockHeader> _headerDecoder = new HeaderDecoder();
+    private readonly IRlpDecoder<BlockBody> _blockBodyDecoder = BlockBodyDecoder.Instance;
```
benaadams: Why interface and not concrete types?

src/Nethermind/Nethermind.ExternalSigner.Plugin/ClefWallet.cs
```diff
@@ -17,7 +17,7 @@ namespace Nethermind.ExternalSigner.Plugin
 {
     public class ClefWallet(IJsonRpcClient rpcClient) : IWallet
     {
-        private readonly HeaderDecoder _headerDecoder = new();
+        private readonly IRlpDecoder<BlockHeader> _headerDecoder = new HeaderDecoder();
```
benaadams: Why interface and not concrete type?

src/Nethermind/Nethermind.Facade/Simulate/SimulateDictionaryBlockStore.cs
```diff
@@ -13,7 +13,7 @@ public class SimulateDictionaryBlockStore(IBlockStore readonlyBaseBlockStore) :
 {
     private readonly Dictionary<Hash256AsKey, Block> _blockDict = [];
     private readonly Dictionary<long, Block> _blockNumDict = [];
-    private readonly BlockDecoder _blockDecoder = new();
+    private readonly IRlpDecoder<Block> _blockDecoder = new BlockDecoder();
```
benaadams: Why interface and not concrete type?

benaadams (commented):
Not sure why using interfaces rather than concrete types when initialised to concrete type?

flcl42:
@benaadams we can't call default interface methods otherwise x9. Fixed in 2 places cos we don't rely on the interface there

benaadams:
> @benaadams we can't call default interface methods otherwise x9. Fixed in 2 places cos we don't rely on the interface there

Can be extension method instead?

benaadams:
> > @benaadams we can't call default interface methods otherwise x9. Fixed in 2 places cos we don't rely on the interface there
> 
> Can be extension method instead?

Or virtual method

flcl42:
@claude  review again

flcl42:
@benaadams @LukaszRozmej could you check it one more time? It's already enough imho for this pr

In the future I'd also:
- replace EncodeToNewNettyStream with encode to a disposable span like ArrayPoolSpan
- rethink DecodeComplete/DecodeCompleteNotNull/DecodeGuardNotNull(may become default behavior)
- rethink whether we should use nullable everywhere in decoder type parameters
- whether we can replace Rlp.GetDecoder with specificdecoder.Instance (considering late overrides in plugins it seems still possible)

LukaszRozmej (approved):
@alexb5dh please review

benaadams (approved):

src/Nethermind/Nethermind.Blockchain.Test/Receipts/PersistentReceiptStorageTests.cs
```diff
@@ -38,6 +38,7 @@ public class PersistentReceiptStorageTests(bool useCompactReceipts)
     private ReceiptConfig _receiptConfig = null!;
     private PersistentReceiptStorage _storage = null!;
     private ReceiptArrayStorageDecoder _decoder = null!;
+    private IRlpDecoder<TxReceipt[]> _rlpDecoder = null!;
```
alexb5dh: So have you agreed on whether direct type or interface should be used in cases like this? 😅

Seems like can be a direct one after switch to extension methods, but probably won't bring any benefits.

src/Nethermind/Nethermind.Serialization.Rlp/IRlpDecoder.cs
```diff
@@ -2,135 +2,64 @@
...
-            /// and verifies that the end of the stream has been reached.
-            /// </summary>
-            public T DecodeComplete(ReadOnlySpan<byte> bytes, RlpBehaviors rlpBehaviors = RlpBehaviors.None)
-            {
-                Rlp.ValueDecoderContext context = new(bytes);
-                return decoder.DecodeComplete(ref context, rlpBehaviors);
-            }
-        }
-
-        extension<T>(IRlpValueDecoder<T> decoder) where T : class
```
alexb5dh, line 66: Why make `DecodeGuard*`, `DecodeComplete*` interface methods now, instead of extensions as before?

src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
```diff
@@ -124,39 +126,19 @@ internal static byte[] ByteSpanToArray(ReadOnlySpan<byte> span)
...
-                byte[][] arrays = RlpStream.SingleByteArrays;
-                if ((uint)value < (uint)arrays.Length)
-                {
-                    return arrays[value].ToPooledList();
-                }
-            }
-
-            return span.ToPooledList();
-        }
+        internal static ArrayPoolList<byte> ByteSpanToArrayPool(ReadOnlySpan<byte> span) => span.Length == 0 ? ArrayPoolList<byte>.Empty() : span.ToPooledList();
```
alexb5dh: Was planning to ask about backwards compatibility, but looks like this method is not being used at all (now?).

src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
```diff
@@ -1879,6 +1844,23 @@ private static void ThrowUnexpectedCount(int count, int expected) =>
         [StackTraceHidden]
         private static void ThrowBufferTooSmall(Span<byte> buffer, int minLength) =>
             throw new ArgumentException($"Buffer is too small. Minimal length: {minLength}, actual length: {buffer.Length}");
+
+        private static CappedArray<byte>[] CreatePreEncodes()
+        {
+            const int MaxCache = 1024;
```
alexb5dh: Nitpicking: other local constants use camelCase

src/Nethermind/Nethermind.Core.Test/Encoding/AuthorizationTupleDecoderTests.cs
```diff
@@ -29,10 +28,18 @@ public void Encode_TupleHasValues_TupleCanBeDecodedToEquivalentTuple(Authorizati
...
-        sut.Decode(ref ctx).Should().BeEquivalentTo(item);
+        AuthorizationTuple decoded = sut.Decode(ref ctx);
+        Assert.Multiple(() =>
+        {
+            Assert.That(decoded.ChainId, Is.EqualTo(item.ChainId));
+            Assert.That(decoded.CodeAddress, Is.EqualTo(item.CodeAddress));
+            Assert.That(decoded.Nonce, Is.EqualTo(item.Nonce));
+            Assert.That(decoded.AuthoritySignature, Is.EqualTo(item.AuthoritySignature));
+            Assert.That(decoded.Authority, Is.EqualTo(item.Authority));
+        });
```
alexb5dh:
```suggestion
        Assert.That(decoded, Is.EqualTo(item).UsingPropertiesComparer());
```

src/Nethermind/Nethermind.JsonRpc.Test/Modules/Eth/EthRpcModuleTests.SignTransaction.cs
```diff
@@ -189,8 +189,9 @@ public void SignTransactionResult_Dispose_ReturnsRentedBufferToPool()
         JsonRpcSuccessResponse response = new() { Result = result };
         response.Dispose();
 
-        pool.Returned.Should().ContainSingle().Which.Should().BeSameAs(rented,
-            "the rented buffer must come back to the pool — otherwise pooling is strictly " +
+        Assert.That(pool.Returned, Has.Count.EqualTo(1));
+        Assert.That(pool.Returned[0], Is.SameAs(rented),
+            "the rented buffer must come back to the pool - otherwise pooling is strictly " +
```
alexb5dh:
```suggestion
        Assert.That(pool.Returned, Is.EqualTo([rented]),
            "the rented buffer must come back to the pool - otherwise pooling is strictly " +
```

src/Nethermind/Nethermind.Synchronization/Trie/SnapRangeRecovery.cs
```diff
@@ -122,7 +120,7 @@ public class SnapRangeRecovery(ISyncPeerPool peerPool, ILogManager logManager) :
             ValueHash256 slotPath = default;
             if (pathAndAccounts.Length > 0)
             {
-                accountRlp = _accountDecoder.Encode(pathAndAccounts[0].Account).Bytes;
+                accountRlp = Rlp.Encode(pathAndAccounts[0].Account).Bytes;
```
alexb5dh: Why most places now use private static decoder instance, but `Account` specifically always go through `Rlp`?

src/Nethermind/Nethermind.Blockchain.Test/Receipts/ReceiptsIteratorTests.cs
```diff
@@ -103,7 +104,7 @@ public void SmokeTest()
 
     private ReceiptsIterator CreateIterator(TxReceipt[] receipts, Block block)
     {
-        using NettyRlpStream stream = _decoder.EncodeToNewNettyStream(receipts, RlpBehaviors.Storage);
+        using NettyRlpStream stream = _rlpDecoder.EncodeToNewNettyStream(receipts, RlpBehaviors.Storage);
```
alexb5dh: Does using `_decoder` directly changes behavior?

src/Nethermind/Nethermind.Xdc.Test/ModuleTests/SyncInfoDecoderTests.cs
```diff
@@ -171,4 +169,31 @@ public void Decode_EmptyByteArray_ValueDecoderContext_ReturnsNull()
 
         Assert.That(decoded, Is.Null);
     }
+
+    private static void AssertSyncInfo(SyncInfo actual, SyncInfo expected)
```
alexb5dh: Probably all these helpers can be replaced with
```
Assert.That(decoded, Is.EqualTo(expected).UsingPropertiesComparer());
```

src/Nethermind/Nethermind.Xdc.Test/VoteDecoderTests.cs
```diff
@@ -200,4 +200,19 @@ public void Vote_PoolKey_ReturnsRoundAndHash()
         Assert.That(round, Is.EqualTo(1UL));
         Assert.That(hash, Is.Not.EqualTo(Hash256.Zero)); // Should be computed hash
     }
+
+    private static void AssertVote(Vote actual, Vote expected)
```
alexb5dh: `UsingPropertiesComparer`?

alexb5dh (approved):
Could you elaborate on many changes like this:
Image


and this:
Image

src/Nethermind/Nethermind.Serialization.Rlp/Rlp.cs
flcl42: Yeah let's remove it

flcl42:
@alexb5dh Snapshot decoding is ugly along with using of .Instance, improved both

src/Nethermind/Nethermind.Serialization.Rlp/IRlpDecoder.cs
flcl42: Because I want to make it a single encode/decode API with all this options considered (whether nulls are allowed and if the entire chunk of data is expected to be decoded with no redundant trailing bytes)

src/Nethermind/Nethermind.Xdc.Test/ModuleTests/SyncInfoDecoderTests.cs
```diff
@@ -112,10 +110,9 @@ public void Encode_UseBothRlpStreamAndValueDecoderContext_IsEquivalentAfterReenc
         Rlp.ValueDecoderContext decoderContext = new(stream.Data.AsSpan());
         SyncInfo decodedContext = decoder.Decode(ref decoderContext);
 
-        // Both should be equivalent to original
-        decodedStream.Should().BeEquivalentTo(syncInfo);
-        decodedContext.Should().BeEquivalentTo(syncInfo);
-        decodedStream.Should().BeEquivalentTo(decodedContext);
+        Assert.That(decodedStream, Is.EqualTo(syncInfo).UsingPropertiesComparer());
+        Assert.That(decodedContext, Is.EqualTo(syncInfo).UsingPropertiesComparer());
+        Assert.That(decodedStream, Is.EqualTo(decodedContext).UsingPropertiesComparer());
```
alexb5dh: 3rd comparison seems redundant?

alexb5dh (approved):

src/Nethermind/Nethermind.Blockchain.Test/Receipts/PersistentReceiptStorageTests.cs
flcl42: Direct is better for now

LukaszRozmej (approved):

benaadams (approved):

PR #10760, opened 2026-03-09
[perf] Optimize IsNonZeroAccount.

LukaszRozmej:
## Changes

- Optimize IsNonZeroAccount
- Deduplicate tests

- [x] Optimization


benaadams (approved):

src/Nethermind/Nethermind.State/WorldState.cs
```diff
@@ -308,17 +308,11 @@ public bool AccountExists(Address address)
...
-            accountExists = _stateProvider.AccountExists(address);
-            if (accountExists)
-            {
-                Account account = _stateProvider.GetAccount(address);
-                return account.IsContract || !account.Nonce.IsZero || !_persistentStorageProvider.IsStorageEmpty(address);
-            }
-
-            return false;
+            Account account = _stateProvider.GetAccount(address);
+            return account.IsContract || !account.Nonce.IsZero || !_persistentStorageProvider.IsStorageEmpty(address);
```
benaadams: Can this just be a property of account? (I think are many atm that might fit?)

LukaszRozmej: I was thinking about it but decided against it, the naming is a bit confusing (suggest potentially empty ballance), but I couldn't think of better name. This is the only place for this check so it is fine.

PR #10782, opened 2026-03-11
XDC - Trc21 State Reader

cicr99:
- Introduce `ITrc21StateReader` which:
    - Reads token fee capacities from issuer contract storage.
    - Implements per-root cache.
    - Exposes token tx validation helper.
- Add TRC21-related fields to XDC spec/release spec
- Implement unit tests

Links for context on xdc side:
- [trc21_reader](https://github.com/XinFinOrg/XDPoSChain/blob/dev-upgrade/core/state/trc21_reader.go)
- [trc21 contract](https://github.com/XinFinOrg/XDPoSChain/blob/dev-upgrade/contracts/trc21issuer/contract/TRC21.sol)
- [trc21Issuer contract](https://github.com/XinFinOrg/XDPoSChain/blob/dev-upgrade/contracts/trc21issuer/contract/TRC21Issuer.sol)

### DIFF
--- a/src/Nethermind/Nethermind.Xdc.Test/SpecialTransactionsTests.cs
+++ b/src/Nethermind/Nethermind.Xdc.Test/SpecialTransactionsTests.cs
@@ -233,7 +233,8 @@ internal class SpecialTransactionsTests
 
         var moqVm = new VirtualMachine(new BlockhashProvider(new BlockhashCache(blockChain.Container.Resolve<IHeaderFinder>(), NullLogManager.Instance), blockChain.MainWorldState, NullLogManager.Instance), blockChain.SpecProvider, NullLogManager.Instance);
 
-        var transactionProcessor = new XdcTransactionProcessor(BlobBaseFeeCalculator.Instance, blockChain.SpecProvider, blockChain.MainWorldState, moqVm, Substitute.For<ICodeInfoRepository>(), NullLogManager.Instance, Substitute.For<IMasternodeVotingContract>());
+        var transactionProcessor = new XdcTransactionProcessor(BlobBaseFeeCalculator.Instance, blockChain.SpecProvider, blockChain.MainWorldState, moqVm, Substitute.For<ICodeInfoRepository>(), NullLogManager.Instance, Substitute.For<IMasternodeVotingContract>(),
+            Substitute.For<ITrc21StateReader>());
 
 
         XdcBlockHeader head = (XdcBlockHeader)blockChain.BlockTree.Head!.Header!;
@@ -292,7 +293,8 @@ internal class SpecialTransactionsTests
             moqVm,
             Substitute.For<ICodeInfoRepository>(),
             NullLogManager.Instance,
-            Substitute.For<IMasternodeVotingContract>());
+            Substitute.For<IMasternodeVotingContract>(),
+            Substitute.For<ITrc21StateReader>());
 
 
         XdcBlockHeader head = (XdcBlockHeader)blockChain.BlockTree.Head!.Header!;
@@ -383,7 +385,8 @@ internal class SpecialTransactionsTests
             moqVm,
             Substitute.For<ICodeInfoRepository>(),
             NullLogManager.Instance,
-            Substitute.For<IMasternodeVotingContract>());
+            Substitute.For<IMasternodeVotingContract>(),
+            Substitute.For<ITrc21StateReader>());
 
 
         XdcBlockHeader head = (XdcBlockHeader)blockChain.BlockTree.Head!.Header!;
@@ -439,7 +442,8 @@ internal class SpecialTransactionsTests
             moqVm,
             Substitute.For<ICodeInfoRepository>(),
             NullLogManager.Instance,
-            Substitute.For<IMasternodeVotingContract>());
+            Substitute.For<IMasternodeVotingContract>(),
+            Substitute.For<ITrc21StateReader>());
 
 
         XdcBlockHeader head = (XdcBlockHeader)blockChain.BlockTree.Head!.Header!;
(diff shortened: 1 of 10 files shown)


src/Nethermind/Chains/xdc.json
```diff
@@ -133,6 +133,7 @@
 
         "masternodeVotingContract": "0x0000000000000000000000000000000000000088",
         "blockSignerContract": "0x0000000000000000000000000000000000000089",
+        "trc21IssuerContract": "0x8c0faeb5C6bEd2129b8674F262Fd45c4e9468bee",
```
alexb5dh, line 137: Is this something that should be present in XDC mainnet?

src/Nethermind/Nethermind.Xdc/Contracts/Trc21StateReader.cs
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+        UInt256 senderBalanceSlot = CalculateMappingSlot(from.ToHash(), TokenBalancesSlot);
+        UInt256 balance = ReadStorage(baseBlock, token, senderBalanceSlot);
+
+        if (!balance.IsZero)
+        {
+            UInt256 minFee = ReadStorage(baseBlock, token, TokenMinFeeSlot);
+            UInt256 transferValue = ExtractTransferValue(data);
+            return balance >= minFee + transferValue;
```
alexb5dh: Worth checking for overflow? I think it may be possible to build an attack utilizing this.

```suggestion
            return !UInt256.AddOverflow(minFee, transferValue, out UInt256 required) && balance >= required;
```

src/Nethermind/Nethermind.Xdc/Contracts/Trc21StateReader.cs
```diff
@@ -0,0 +1,144 @@
...
+            return LoadFeeCapacities(baseBlock);
+        }
+
+        if (!_capacityCache.TryGet(baseBlock.StateRoot, out Dictionary<Address, UInt256>? cached) || cached is null)
+        {
+            cached = LoadFeeCapacities(baseBlock);
+            _capacityCache.Set(baseBlock.StateRoot, cached);
+        }
+
+        return Clone(cached);
```
alexb5dh: Change return type to `IReadOnlyDictionary<Address, UInt256>` and just return `cached`? Or `cached.AsReadOnly()`, but this will allocate.

src/Nethermind/Nethermind.Xdc/XdcModule.cs
```diff
@@ -71,6 +72,7 @@ protected override void Load(ContainerBuilder builder)
                 IAbiEncoder,
                 ISpecProvider,
                 IReadOnlyTxProcessingEnvFactory>(CreateVotingContract)
+            .AddSingleton<ITrc21StateReader, IStateReader, ISpecProvider>((stateReader, specProvider) => new Trc21StateReader(stateReader, specProvider))
```
alexb5dh: Does it need a custom factory function?

src/Nethermind/Chains/xdc.json
cicr99: I believe so, yes

src/Nethermind/Nethermind.Xdc/Contracts/Trc21StateReader.cs
cicr99: very nice catch!

src/Nethermind/Chains/xdc.json
alexb5dh: But there's nothing at https://xdcscan.com/address/0x8c0faeb5C6bEd2129b8674F262Fd45c4e9468bee 🤔

batrr: It seems the contract was never deployed

src/Nethermind/Nethermind.Xdc/Contracts/Trc21StateReader.cs
cicr99: yes, it makes sense. I originally went with the clone as it was safest, but it's true is more expensive. Updated it

src/Nethermind/Chains/xdc.json
cicr99: yep, checked both in mainnet and testnet and the contract doesn't seem to have ever been deployed. We were under the impression after talking to them that this feature had been used in the past but not anymore, however this indicates, it might have never been released even.

src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -110,7 +111,8 @@ public TxPool(IEthereumEcdsa ecdsa,
             ITxGossipPolicy? transactionsGossipPolicy = null,
             IIncomingTxFilter? incomingTxFilter = null,
             [KeyFilter(ITxValidator.HeadTxValidatorKey)] ITxValidator? headTxValidator = null,
-            bool thereIsPriorityContract = false)
+            bool thereIsPriorityContract = false,
+            ITxPoolCostAndFundsProvider? costAndFundsProvider = null)
```
LukaszRozmej, line 115: I don't like it at all.

IMO better solution would be to give plugins a way to create/ammend those filter classes.
And then you could create a compleatly customised `BalanceZeroFilter` - compleatly different class, no code shared. Or derive from `BalanceZeroFilter` and override something with your checks.

src/Nethermind/Nethermind.TxPool/TxPool.cs
```diff
@@ -110,7 +111,8 @@ public TxPool(IEthereumEcdsa ecdsa,
             ITxGossipPolicy? transactionsGossipPolicy = null,
             IIncomingTxFilter? incomingTxFilter = null,
             [KeyFilter(ITxValidator.HeadTxValidatorKey)] ITxValidator? headTxValidator = null,
-            bool thereIsPriorityContract = false)
+            bool thereIsPriorityContract = false,
```
LukaszRozmej, line 114: If you do this, please move this to AuRa plugin too!

LukaszRozmej, line 114: This probably should be moved to AuRa plugin if that mechanism is available

src/Nethermind/Nethermind.TxPool/TxPool.cs
cicr99: Yes, I agree that would be definitely a better solution. The changes in this PR seemed like the minimal path to restore correct behavior for xdc filters without broad refactor risk, as reworking txpool filter construction would entail much larger modifications

PR #10785, opened 2026-03-11
Refactor use base class for hash comparison

ak88:
Just a base class for hash comparisons in XDC

### DIFF
--- a/src/Nethermind/Nethermind.Xdc/Types/RlpHashEqualityBase.cs
+++ b/src/Nethermind/Nethermind.Xdc/Types/RlpHashEqualityBase.cs
@@ -8,7 +8,7 @@ namespace Nethermind.Xdc.Types;
 
 public abstract class RlpHashEqualityBase
 {
-    private Hash256 _hash;
+    private Hash256? _hash;
 
     public Hash256 Hash => _hash ??= CalculateHash();
 


src/Nethermind/Nethermind.Xdc/Types/QuorumCertificate.cs
```diff
@@ -2,12 +2,18 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 
 using Nethermind.Core.Crypto;
+using Nethermind.Crypto;
+using Nethermind.Serialization.Rlp;
+using System.Text;
```
batrr, line 7:
```suggestion
```

src/Nethermind/Nethermind.Xdc/Types/RlpHashEqualityBase.cs
```diff
@@ -0,0 +1,26 @@
...
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using Nethermind.Core.Crypto;
+using Nethermind.Crypto;
+
+namespace Nethermind.Xdc.Types;
+
+public abstract class RlpHashEqualityBase
+{
+    private Hash256 _hash;
```
batrr:
```suggestion
    private Hash256? _hash;
```

src/Nethermind/Nethermind.Xdc/Types/RlpHashEqualityBase.cs
```diff
@@ -0,0 +1,26 @@
...
+    private Hash256 CalculateHash()
+    {
+        KeccakRlpStream stream = new();
+        Encode(stream);
+        return stream.GetHash();
+    }
+    protected abstract void Encode(KeccakRlpStream stream);
+
+    public override bool Equals(object? obj) =>
+                obj is RlpHashEqualityBase other && Hash == other.Hash;
```
batrr, line 24: `Vote` can be equal to `Timeout` here. Unlikely though.

ak88: Because of RLP sequence it would not be possible, but could theoretically happen with two object with the exact same RLP sequence

PR #10803, opened 2026-03-12
XDC Random Value

batrr:
XDC uses the block number as a random number, not the difficulty.

## Changes

  - Add `BlockExecutionContext.WithPrevRandao` static factory for custom RNG injection
  - Override `CreateBlockExecutionContext` to  pass xdc specific `PREVRANDAO`  [xdc](https://github.com/XinFinOrg/XDPoSChain/blob/0b257ecb015fc8bf54f250d38d61117f0a301b39/core/evm.go#L46)

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

### DIFF
--- a/src/Nethermind/Nethermind.Xdc/XdcBlockProcessor.cs
+++ b/src/Nethermind/Nethermind.Xdc/XdcBlockProcessor.cs
@@ -25,9 +25,10 @@ internal class XdcBlockProcessor : BlockProcessor
     {
     }
 
+    // Match Go's big.Int.Bytes() behavior: zero produces empty bytes, not [0x00].
     protected override BlockExecutionContext CreateBlockExecutionContext(BlockHeader header, IReleaseSpec spec) =>
         BlockExecutionContext.WithPrevRandao(header, spec,
-            ValueKeccak.Compute(header.Number.ToBigEndianSpanWithoutLeadingZeros(out _)));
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Xdc/XdcBlockProcessor.cs
```diff
@@ -22,6 +25,10 @@ public XdcBlockProcessor(ISpecProvider specProvider, IBlockValidator blockValida
     {
     }
 
+    protected override BlockExecutionContext CreateBlockExecutionContext(BlockHeader header, IReleaseSpec spec) =>
+        BlockExecutionContext.WithPrevRandao(header, spec,
+            ValueKeccak.Compute(header.Number.ToBigEndianSpanWithoutLeadingZeros(out _)));
```
alexb5dh: So for block value of `0` (which we probably won't hit?) I think it will be different from XDC version, as `big.Int.Bytes()` [returns](https://stackoverflow.com/questions/67908738/why-big-newint0-bytes-returns-instead-of-0-in-go) `[]` in Go, while our  version returns `[0]`.

But in general, +1 for a test comparing this function output against static pre-computed XDC values for a few inputs, as I'm not so confident our `ToBigEndianSpanWithoutLeadingZeros` always matches `big.Int.Bytes()`.

PR #10809, opened 2026-03-12
Fix/xdc block building

ak88:
Fixes for XDC block proposals

### DIFF
--- a/src/Nethermind/Nethermind.Xdc.Test/ModuleTests/TimeoutTests.cs
+++ b/src/Nethermind/Nethermind.Xdc.Test/ModuleTests/TimeoutTests.cs
@@ -66,7 +66,7 @@ public class TimeoutTests
 
         // Sanity check: round hasn’t advanced, HighestTC not set to this round yet
         Assert.That(ctx.CurrentRound, Is.EqualTo(round));
-        Assert.That(ctx.HighestTC.Round, Is.EqualTo(0));        
+        Assert.That(ctx.HighestTC.Round, Is.EqualTo(0));
 
         // Send timeout message with wrong gap so it doesn't reach threshold yet
         await blockchain.TimeoutCertificateManager.HandleTimeoutVote(XdcTestHelper.BuildSignedTimeout(signers.Last(), round, 1350));


src/Nethermind/Nethermind.Xdc/XdcBlockTree.cs
```diff
@@ -77,4 +77,34 @@ protected override AddBlockResult Suggest(Block? block, BlockHeader header, Bloc
...
     }
 
+    protected override bool HeadImprovementRequirementsSatisfied(BlockHeader header)
+    {
+        if (base.HeadImprovementRequirementsSatisfied(header))
+            return true;
+
+        return header is XdcBlockHeader newBlock && Head?.Header is XdcBlockHeader headBlock &&
+            IsSameTdButSelfMined(newBlock, headBlock);
+    }
```
batrr, line 87: What about using rounds here?

src/Nethermind/Nethermind.Xdc/XdcHotStuff.cs
```diff
@@ -334,40 +339,53 @@ private async Task CommitCertificateAndVote(XdcBlockHeader head, EpochSwitchInfo
...
+                if (_logger.IsDebug)
+                    _logger.Debug($"Round {votingRound}: Skipped voting (not in masternode set)");
                 return;
             }
 
-            // Check voting rule
-            bool canVote = _votesManager.VerifyVotingRules(head);
-            if (!canVote)
+            // Check voting rule            
+            if (!head.IsSelfMined && !_votesManager.VerifyVotingRules(head, out string error))
```
batrr, line 362: `!head.IsSelfMined` seems to be unnecessary,

src/Nethermind/Nethermind.Xdc/XdcHotStuff.cs
```diff
@@ -399,17 +417,26 @@ private bool IsItTimeToPropose(XdcBlockHeader parent, ulong round, IXdcReleaseSp
...
-            if ((long)parent.Timestamp + spec.MinePeriod > DateTimeOffset.UtcNow.ToUnixTimeSeconds())
+            var now = DateTimeOffset.UtcNow.ToUnixTimeSeconds();
+            if ((long)parent.Timestamp + spec.MinePeriod > now)
             {
                 //Not enough time has passed since last block
                 return false;
             }
 
+            var fallbackPeriod = spec.TimeoutPeriod / 2;
+            if ((long)parent.Timestamp + fallbackPeriod < now)
```
batrr, line 427: Nice. 
It would be good to have a better explanation. 
And it seems to be outside XDPoS 2.0 spec?

src/Nethermind/Nethermind.Xdc/QuorumCertificateManager.cs
```diff
@@ -86,7 +86,7 @@ public void CommitCertificate(QuorumCertificate qc)
         }
     }
 
-    private bool CommitBlock(IBlockTree chain, XdcBlockHeader proposedBlockHeader, ulong proposedRound, QuorumCertificate proposedQuorumCert, [NotNullWhen(false)] out string? error)
+    private bool CommitBlock(XdcBlockHeader proposedBlockHeader, ulong proposedRound, QuorumCertificate proposedQuorumCert, [NotNullWhen(false)] out string? error)
```
batrr, line 89: `proposedQuorumCert` is unused

ak88: fix

src/Nethermind/Nethermind.Xdc/XdcBlockTree.cs
ak88: I cant find any mention of using rounds as canonical chain in white papers, and there is nothing in code either.

src/Nethermind/Nethermind.Xdc/XdcHotStuff.cs
ak88: We skip the voting rules check if we mined the block itself, because we want to always vote on our own block.

src/Nethermind/Nethermind.Xdc/XdcHotStuff.cs
ak88: It is not part of consensus, but rather up to block producers. XDC will build a block with whatever it currently has and not wait for voting to finish.

ak88: By waiting some extra seconds we have a higher chance of not forking.

PR #10325, opened 2026-01-26
Bal devnet 2

Marchhill:
## Changes

- Devnet 2 Eips: https://notes.ethereum.org/@ethpandaops/bal-devnet-2
- [EIP-7708](https://eips.ethereum.org/EIPS/eip-7708) ETH transfers emit a log
- [EIP-7778](https://eips.ethereum.org/EIPS/eip-7778) Block Gas Accounting without Refunds
- [EIP-7843](https://eips.ethereum.org/EIPS/eip-7843) SLOTNUM opcode
- [EIP-8024](https://eips.ethereum.org/EIPS/eip-8024) Backward compatible SWAPN, DUPN, EXCHANGE

#### What types of changes does your code introduce?

- [x] New feature (a non-breaking change that adds functionality)

#### Requires testing

- [x] Yes
- [ ] No

#### If yes, did you write tests?

- [x] Yes
- [ ] No

PR #10810, opened 2026-03-12
Restore Taiko FCU timestamp validation allowing equal timestamps

jmadibekov:
## Changes

- Add `IsPayloadTimestampValid` virtual method to `ForkchoiceUpdatedHandler` (base: `>` for strict increase)
- Override in `TaikoForkchoiceUpdatedHandler` as a one-liner using `>=` (allowing equal timestamps)
- Reduce `timeout-minutes` from 45 to 20 in `ci-surge.yml` and `ci-taiko.yml`, add step-level 15m timeout on the test step so timeouts surface as failures instead of cancellations
- Add unit tests covering equal, greater, and lesser timestamp scenarios

PR #10325 renamed `IsPayloadAttributesTimestampValid` → `ArePayloadAttributesTimestampAndSlotNumberValid`, removed `virtual`, and deleted the Taiko override. In Taiko, multiple L2 blocks can be derived from a single L1 block sharing the same L1 anchor timestamp, so equal timestamps must be allowed.

#### Requires testing

- [x] Yes

- [x] Yes

#### Notes on testing

Added 3 test cases in `TaikoEngineApiTests` verifying equal timestamps are accepted, greater timestamps are accepted, and lesser timestamps are rejected.

Integration tests passed on this PR:
- Surge: https://github.com/NethermindEth/nethermind/actions/runs/23011831745
- Taiko: https://github.com/NethermindEth/nethermind/actions/runs/23011831680

- [ ] No

- [ ] No

### DIFF
--- a/.agents/rules/coding-style.md
+++ b/.agents/rules/coding-style.md
@@ -12,6 +12,7 @@
 - Use documentation comments with proper structure (`<summary>`, `<param>`, `<returns>`) for all public APIs
 - Avoid `var` — spell out types (exception: very long nested generic types)
 - Prefer low-allocation code patterns
+- Use `Array.Empty<T>()` or `[]` instead of `new T[0]` — avoids allocating a new empty array each time. In attribute arguments (e.g. `[Attr(new string[0])]`), `new T[0]` is acceptable because `Array.Empty<T>()` is not a compile-time constant.
 - Consider performance implications in high-throughput paths
 - **No LINQ** when a simple `for`/`foreach` works — use LINQ only for complex queries
 - Trust null annotations, don't add redundant null checks
--- a/.agents/rules/test-infrastructure.md
+++ b/.agents/rules/test-infrastructure.md
@@ -69,5 +69,5 @@ The rule: **if production modules already wire a component, use them — don't c
 ## Test guidelines
 
 - Add tests to existing test files rather than creating new ones
-- When adding similar tests, write one test with test cases (`[TestCase(...)]`)
-- Check if previous tests can be reused with a new test case
+- **Do not duplicate test methods that differ only in parameters** — use `[TestCase(...)]` or `[TestCaseSource(...)]` to parameterize a single method
+- Before writing a new test, check if an existing test can be extended with another `[TestCase]` or use `[TestCaseSource]`
--- a/.agents/skills/review/SKILL.md
+++ b/.agents/skills/review/SKILL.md
@@ -2,7 +2,15 @@
 name: review
 description: Deep code review for an Ethereum execution client. Checks consensus correctness, security, robustness, performance, DI patterns, breaking changes, and observability. Use when asked to "review", "check this PR", "look for bugs", "audit", or "review my changes".
 allowed-tools:
-  [Bash(git diff*), Bash(git merge-base*), Bash(git log*), Bash(git status*), Read, Grep, Glob]
+  [
+    Bash(git diff*),
+    Bash(git merge-base*),
+    Bash(git log*),
+    Bash(git status*),
+    Read,
+    Grep,
+    Glob,
+  ]
 ---
 
 # Code review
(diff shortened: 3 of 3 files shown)


src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
```diff
@@ -288,10 +288,12 @@ protected virtual bool IsOnMainChainBehindHead(Block newHeadBlock, ForkchoiceSta
...
 
+    protected virtual ulong GetMinRequiredTimestamp(Block newHeadBlock) => newHeadBlock.Timestamp + 1;
+
     protected bool ArePayloadAttributesTimestampAndSlotNumberValid(Block newHeadBlock, ForkchoiceStateV1 forkchoiceState, PayloadAttributes payloadAttributes,
         [NotNullWhen(false)] out ResultWrapper<ForkchoiceUpdatedV1Result>? errorResult)
     {
-        if (newHeadBlock.Timestamp >= payloadAttributes.Timestamp)
+        if (payloadAttributes.Timestamp < GetMinRequiredTimestamp(newHeadBlock))
         {
             string error = $"Payload timestamp {payloadAttributes.Timestamp} must be greater than block timestamp {newHeadBlock.Timestamp}.";
```
LukaszRozmej: This message should be adjusted to use `GetMinRequiredTimestamp` too - save it in variable.

LukaszRozmej: Also not sure which would be better here `<` or `<=` (which would require amending `GetMinRequiredTimestamp` accordingly), WDYT?

src/Nethermind/Nethermind.Taiko/Rpc/TaikoForkchoiceUpdatedHandler.cs
```diff
@@ -77,6 +77,10 @@ protected override bool IsOnMainChainBehindHead(Block newHeadBlock, ForkchoiceSt
         return blockHeader;
     }
 
+    // Taiko Pacaya allows equal timestamps because multiple L2 blocks can be derived
+    // from a single L1 block, all sharing the same L1 anchor timestamp.
+    protected override ulong GetMinRequiredTimestamp(Block newHeadBlock) => newHeadBlock.Timestamp;
```
dipkakwani: This change will allow equal timestamps for all the forks, not just Pacaya right? 
We need to have conditions to return based on the current fork, similar to taiko geth here: https://github.com/taikoxyz/taiko-geth/blob/1fda81a9372dd79056427487caef38cee2be75b5/consensus/taiko/consensus.go#L158-L168

src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
```diff
@@ -288,10 +288,12 @@ protected virtual bool IsOnMainChainBehindHead(Block newHeadBlock, ForkchoiceSta
         return null;
     }
 
+    protected virtual ulong GetMinRequiredTimestamp(Block newHeadBlock) => newHeadBlock.Timestamp + 1;
```
dipkakwani: This function looks a bit random, the intention of adding 1 to the timestamp is not clear. 
Do you think it wil be better to have the complete logic for checking timestamp as virtual, instead of only the minimum required timestamp? Something like `IsPayloadTimestampValid` in the `if` statement of line 296, and each module can override that function?

dipkakwani: Just noticed the linked PR, not sure why it was removed. The existing `IsPayloadTimestampValid` https://github.com/NethermindEth/nethermind/pull/10325/changes#diff-74bac6db7a134940f36bb275f9bd355e45f5428bf5f30cd4bd2b7fccd2b1b186

dipkakwani: Also noticed Lukaz comment, I think we can still make `IsPayloadTimestampValid` one-liner in the base class instead of using `GetMinRequiredTimestamp`.

src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
jmadibekov: Regarding the first comment, done! Regarding the second, I'd rather leave it as-is (that is `<`) because I find that more readable.

src/Nethermind/Nethermind.Taiko/Rpc/TaikoForkchoiceUpdatedHandler.cs
jmadibekov: Actually, we do already have similar per-fork rules during block validation: https://github.com/NethermindEth/nethermind/blob/master/src/Nethermind/Nethermind.Taiko/TaikoHeaderValidator.cs#L204-L226. 

And when it comes to fcu, nmc is actually correctly mirroring here taiko-geth because taiko-geth also allows equal timestamps for all forks unconditionally: https://github.com/taikoxyz/taiko-geth/blob/taiko/miner/worker.go#L174-L187. 

So I don't think there is a need for this.

src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
jmadibekov: I like that, done!

src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
jmadibekov: Actually, I went ahead with the approach @dipkakwani suggested below because I found that even more elegant :)

LukaszRozmej: But the error message for taiko is now wrong?

PR #10814, opened 2026-03-12
Simplify releasing & fix `IByteBuffer` leaks in tests

alexb5dh:
## Changes

- Adds `DisposableByteBuffer` decorator for `IByteBuffer`, implementing `IDisposable` and calling `Release` during `Dispose` - to be used in tests
- Simplifies existing `IByteBuffer` releases in tests
- Fixes `IByteBuffer` leaks in tests

- [x] Bugfix (tests only)


src/Nethermind/Nethermind.Serialization.Rlp/ByteBufferExtensions.cs
```diff
@@ -111,5 +111,7 @@ public static Memory<byte> AsMemory(this IByteBuffer buffer, int? startIndex = n
             return buffer.Array.AsMemory()
                 .Slice(buffer.ArrayOffset + startIdx, buffer.WriterIndex - startIdx);
         }
+
+        public static DisposableByteBuffer AsDisposable(this IByteBuffer buffer) => new(buffer);
```
LukaszRozmej: I would drop this method and instead add implicit cast from `IByteBuffer` to `DisposableByteBuffer`

alexb5dh: > User-defined conversions are not allowed to convert from or to interface-types.

😢

src/Nethermind/Nethermind.Network.Test/DisposableByteBuffer.cs
```diff
@@ -0,0 +1,224 @@
...
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using DotNetty.Buffers;
+using DotNetty.Common;
+using DotNetty.Common.Utilities;
+
+namespace Nethermind.Serialization.Rlp;
+
+public sealed class DisposableByteBuffer(IByteBuffer inner) : IByteBuffer, IDisposable
```
LukaszRozmej, line 15: Maybe it should live in tests?
For non-test potentially struct would be better if it is meant to be disposed in same method as created.
Maybe even ref struct.

alexb5dh: I did think about `struct`, but changed my mind after realising it will most likely get boxed anyway.
Will check via `ref struct` so.

alexb5dh: Adding it to non-test project was a mistake, will fix.

src/Nethermind/Nethermind.Network.Test/P2P/Subprotocols/Snap/Messages/GetTrieNodesMessageSerializerTests.cs
```diff
@@ -107,33 +107,26 @@ public void NullPathGroup()
...
-                buffer.Release();
-                buffer2.Release();
-                msg.Dispose();
-            }
+            serializer.Serialize(buffer2, deserialized);
+            byte[] secondBytes = new byte[buffer2.ReadableBytes];
+            buffer2.GetBytes(buffer2.ReaderIndex, secondBytes);
+
+            secondBytes.Should().BeEquivalentTo(firstBytes);
+            msg.Dispose();
```
newmanifold: This `msg.Dispose` should still be in try/finally, no?

alexb5dh: Fixed via another `using`

PR #10818, opened 2026-03-13
fix(simulate): correct LogIndex calculation

zeroprooff:
LogIndex used txIndex + i which breaks when tx has more than 1 log. Added cumulative counter to track log position in block.


LukaszRozmej (changes requested):
missing test showing the issue

PR #10821, opened 2026-03-13
XDC - TRC21 related logic in XdcTransactionProcessor

cicr99:
This PR adds TRC21-aware fee handling to XDC transaction execution, focused on `XdcTransactionProcessor`

**Changes:**
- Added TRC21-aware effective gas price logic:
    - Special XDC txs return 0 gas price.
    - TRC21 txs use XDC tiered gas price (before tip block / after tip block / 50x block).
    - Reference: [Gas price adjustment](https://github.com/XinFinOrg/XDPoSChain/blob/be36b32b3e4d999a5504e22c4d69d2c521860c63/core/state_transition.go#L174)
- Added TRC21-aware BuyGas path:
    - For TRC21 txs, checks sender native balance only for transfer value.
    - Validates token fee capacity against gasLimit * effectiveGasPrice.
    -  Reference: [buyGas function](https://github.com/XinFinOrg/XDPoSChain/blob/be36b32b3e4d999a5504e22c4d69d2c521860c63/core/state_transition.go#L274)
- Added TRC21-aware refund behavior in XdcTransactionProcessor:
    - TRC21 path skips native refund credit.
    -  Reference: [returnGas function](https://github.com/XinFinOrg/XDPoSChain/blob/be36b32b3e4d999a5504e22c4d69d2c521860c63/core/state_transition.go#L593)

### DIFF
--- a/src/Nethermind/Nethermind.Xdc/XdcTransactionProcessor.cs
+++ b/src/Nethermind/Nethermind.Xdc/XdcTransactionProcessor.cs
@@ -301,7 +301,7 @@ internal class XdcTransactionProcessor : EthereumTransactionProcessorBase
     {
         if (blockNumber >= spec.BlockNumberGas50x)
         {
-            return XdcConstants.GasPrice50x;
+            return XdcConstants.Trc21GasPrice50x;
         }
 
         if (blockNumber > spec.TipTrc21FeeBlock)
--- a/src/Nethermind/Nethermind.Xdc.Test/TxPool/XdcTxPoolTestHelper.cs
+++ b/src/Nethermind/Nethermind.Xdc.Test/TxPool/XdcTxPoolTestHelper.cs
@@ -77,7 +77,7 @@ internal static class XdcTxPoolTestHelper
         public bool IsValid { get; set; } = true;
         public int ValidateCalls { get; private set; }
 
-        public Dictionary<Address, UInt256> GetFeeCapacities(XdcBlockHeader? baseBlock) => new(FeeCapacities);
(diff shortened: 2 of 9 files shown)


src/Nethermind/Nethermind.Xdc/XdcTransactionProcessor.cs
```diff
@@ -79,19 +82,46 @@ protected override TransactionResult BuyGas(Transaction tx, IReleaseSpec spec, I
...
+        bool validate = ShouldValidateGas(tx, opts);
+        if (validate && !TryCalculatePremiumPerGas(tx, header.BaseFeePerGas, out premiumPerGas))
+        {
+            TraceLogInvalidTx(tx, "MINER_PREMIUM_IS_NEGATIVE");
+            return TransactionResult.MinerPremiumNegative;
         }
-        return base.BuyGas(tx, spec, tracer, opts, effectiveGasPrice, out premiumPerGas, out senderReservedGasPayment, out blobBaseFee);
+
+        UInt256 senderBalance = WorldState.GetBalance(tx.SenderAddress!);
+        if (UInt256.SubtractUnderflow(in senderBalance, in tx.ValueRef, out _))
```
batrr, line 105: Do we need this check here?

cicr99: Yes, I believe we have to keep it. On xdc side they still do this check via `CanTransfer` function in `TransitionDb`, not in `buyGas`. Nethermind’s flow does not have a check like this anywhere else before  `PayValue`, so without this check an insufficient-value tx can reach `SubtractFromBalance` and fail as a state exception instead of a normal tx validation result.

PR #10822, opened 2026-03-13
Make state available if not pruned

flcl42:
## Changes

- Make state available if not prunable

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Trie.Test/Pruning/TreeStoreTests.cs
+++ b/src/Nethermind/Nethermind.Trie.Test/Pruning/TreeStoreTests.cs
@@ -946,7 +946,7 @@ namespace Nethermind.Trie.Test.Pruning
         }
 
         [Test]
-        public void HasRoot_with_block_number_allows_old_blocks_in_archive_mode([Values(false, true)] bool trackPastKeys)
+        public void HasRoot_with_block_number_allows_old_blocks_in_archive_mode([Values] bool trackPastKeys)
         {
             int pruningBoundary = 4;
             // Archive mode: shouldPrune always true, deleteObsoleteKeys = false


src/Nethermind/Nethermind.Trie.Test/Pruning/TreeStoreTests.cs
```diff
@@ -945,6 +945,52 @@ public void HasRoot_with_block_number_rejects_pruned_state()
             trieStore.Dispose();
         }
 
+        [Test]
+        public void HasRoot_with_block_number_allows_old_blocks_in_archive_mode()
+        {
+            int pruningBoundary = 4;
+            // Archive mode: shouldPrune always true, deleteObsoleteKeys = false
+            TestPruningStrategy testPruningStrategy = new TestPruningStrategy(shouldPrune: true, deleteObsoleteKeys: false);
```
alexb5dh, line 955: Test seems to pass with any value of `deleteObsoleteKeys` when `TrackPastKeys = false` - because of `TrieStore._deleteOldNodes = _pruningStrategy.DeleteObsoleteKeys && _pastKeyTrackingEnabled`. As I understand this corresponds to real archive node setup, but maybe worth testing both values for `TrackPastKeys`?

flcl42: Made it check both paths

src/Nethermind/Nethermind.Trie.Test/Pruning/TreeStoreTests.cs
```diff
@@ -945,6 +945,54 @@ public void HasRoot_with_block_number_rejects_pruned_state()
             trieStore.Dispose();
         }
 
+        [Test]
+        public void HasRoot_with_block_number_allows_old_blocks_in_archive_mode([Values(false, true)] bool trackPastKeys)
```
alexb5dh:
```suggestion
        public void HasRoot_with_block_number_allows_old_blocks_in_archive_mode([Values] bool trackPastKeys)
```

PR #10831, opened 2026-03-14
devops: extract code-formatting steps into shared scripts

MarekM25:
## Changes

- Add `scripts/dotnet-format.sh` — wraps `dotnet format whitespace` for both `src/Nethermind/` and `tools/`, forwarding all arguments verbatim
- Add `scripts/cspell.sh` — wraps `npx cspell` using the repo's `cspell.json` config
- Update `.github/workflows/code-formatting.yml` to call `bash scripts/dotnet-format.sh --verify-no-changes` instead of inlining the commands

## Motivation

The CI workflow and local Claude Code hooks were duplicating the formatting logic. If flags or target directories change, they now need to be updated in one place (`scripts/dotnet-format.sh`) and both CI and local hooks pick it up automatically.

The `cspell-action` step is intentionally kept as a GitHub Action (it provides PR annotations and only checks changed files), but `scripts/cspell.sh` provides a local equivalent for hooks — both read the same `cspell.json`.

## Type of change

- [x] CI/CD/release pipeline

## Testing

- [ ] Workflow runs `bash scripts/dotnet-format.sh --verify-no-changes` and fails if formatting is off
- [ ] Local hook calls `scripts/dotnet-format.sh --include <file>` and `scripts/cspell.sh <file>` after each `.cs` edit


.claude/settings.json
```diff
@@ -0,0 +1,15 @@
+{
+  "hooks": {
+    "PostToolUse": [
+      {
+        "matcher": "Edit|Write",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "bash .claude/hooks/format-check.sh"
```
newmanifold, line 9: should cspell.sh check also be here? would be good if it can do the changes automatically rather than calling the skill

PR #10768, opened 2026-03-10
ci: Add Claude review

smartprogrammer93:

PR #10832, opened 2026-03-14
feat: Add Codex review workflow

rubo:
## Changes

This is a follow-up to #10768—adds a Codex review workflow. OpenAI uses a different approach, so the workflow is more complex but also offers greater flexibility and freedom. It's limited to owner, member, and collaborator roles. The reviews are submitted on behalf of the GitHub Actions bot, unlike those from Claude and Copilot.

- Added workflow
- Added review instructions

To trigger a review, just tag Codex like so: `@codex` or `@codex review`

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [x] Other: Codex review

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

Tested manually

### DIFF
--- a/.github/workflows/codex-review.yml
+++ b/.github/workflows/codex-review.yml
@@ -162,6 +162,7 @@ jobs:
           prompt-file: ${{ steps.prompt.outputs.prompt_file }}
 
   post-feedback:
+    name: Feedback
     runs-on: ubuntu-latest
     needs: codex
     if: needs.codex.outputs.final_message != ''


.github/workflows/codex-review.yml
```diff
@@ -0,0 +1,154 @@
...
+            core.setOutput('prompt_file', promptPath);
+
+      - name: Run Codex
+        id: run-codex
+        uses: openai/codex-action@v1
+        with:
+          openai-api-key: ${{ secrets.OPENAI_API_KEY }}
+          prompt-file: ${{ steps.prompt.outputs.prompt_file }}
+
+      - name: Report Codex feedback
```
newmanifold, line 170: https://github.com/openai/codex-action/blob/main/docs/security.md#recommendation-run-openaicodex-action-as-the-last-step-in-a-job

Not sure whether this recommendation is relevant there ?

rubo: Well, I was wondering why they have the reporting step as a separate job in their sample. That's the explanation, I guess. Will restore it to the original way then. Thanks.

PR #10833, opened 2026-03-14
fix: IByteBuffer leak in ZeroSnappyEncoder.Encode

newmanifold:
## Changes

- Replace `input.ReadBytes(packetTypeLen)` with `WriteBytes(input, packetTypeLen)` in `ZeroSnappyEncoder.Encode`
- `ReadBytes(n)` allocates a new pooled `IByteBuffer` from `input.Allocator` that is never released, leaking one buffer per outbound P2P message
- `WriteBytes(input, n)` copies bytes directly from source to destination without any intermediate allocation
- Add regression test using `PooledByteBufferAllocator` metrics to verify no active allocations grow after Encode

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Network.Test/Rlpx/SnappyTests.cs
+++ b/src/Nethermind/Nethermind.Network.Test/Rlpx/SnappyTests.cs
@@ -124,7 +124,7 @@ public class SnappyTests
         using DisposableByteBuffer output = allocator.Buffer().AsDisposable();
 
         input.WriteBytes(payload);
-        
+
         long activeBefore = allocator.Metric.HeapArenas().Sum(a => a.NumActiveAllocations);
 
         encoder.TestEncode(input, output);


src/Nethermind/Nethermind.Network.Test/Rlpx/SnappyTests.cs
```diff
@@ -98,4 +103,42 @@ public void Roundtrip_zero()
...
+    public void Encode_does_not_leak_pooled_buffers()
+    {
+        PooledByteBufferAllocator allocator = new();
+        ZeroSnappyEncoderForTest encoder = new();
+
+        // RLP-encoded packet type (0x01) followed by an RLP-encoded body
+        byte[] packetType = Rlp.Encode(1).Bytes;
+        byte[] body = Rlp.Encode(new byte[100]).Bytes;
+        byte[] payload = Bytes.Concat(packetType, body);
+        IByteBuffer input = allocator.Buffer();
```
LukaszRozmej: Could use pattern from https://github.com/NethermindEth/nethermind/pull/10814 (base this upon it)

src/Nethermind/Nethermind.Network.Test/Rlpx/SnappyTests.cs
```diff
@@ -98,4 +103,42 @@ public void Roundtrip_zero()
...
+        input.WriteBytes(payload);
+
+        IByteBuffer output = allocator.Buffer();
+        try
+        {
+            long activeBefore = allocator.Metric.HeapArenas().Sum(a => a.NumActiveAllocations);
+
+            encoder.TestEncode(input, output);
+
+            long activeAfter = allocator.Metric.HeapArenas().Sum(a => a.NumActiveAllocations);
```
LukaszRozmej: Should we have those checks for tests with any code touching ByteBuffers?

src/Nethermind/Nethermind.Network.Test/Rlpx/SnappyTests.cs
newmanifold: Will do it after it after https://github.com/NethermindEth/nethermind/pull/10814 is merged to master

src/Nethermind/Nethermind.Network.Test/Rlpx/SnappyTests.cs
newmanifold: Maybe, but i just realised that allocator stays till process's lifetime and it would require at least 4KiB at bare minimum. We can have a static allocator per test class and use that assuming that tests run sequentially.

PR #10839, opened 2026-03-15
fix: handle tuple disposal in TryDispose to prevent resource leaks

newmanifold:
## Changes

- Fix `TryDispose<T>` in `DisposableExtensions` to handle `ITuple` types by disposing each element individually
- Previously, `(item as IDisposable)?.Dispose()` silently no-oped on value tuples like `(OwnedBlockBodies, long)` and `(IOwnedReadOnlyList<TxReceipt[]>, long)`, leaking disposable components in `MessageDictionary.Handle` and `MessageQueue.Handle` error paths
- Add `ITuple` fallback that iterates tuple elements and disposes any that implement `IDisposable`
- Add `ValueTuple<T1, T2>` overload with `T1 : IDisposable` constraint to avoid boxing the tuple on the common path

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Core/Extensions/DisposableExtensions.cs
+++ b/src/Nethermind/Nethermind.Core/Extensions/DisposableExtensions.cs
@@ -14,8 +14,9 @@ public static class DisposableExtensions
     /// Attempts to dispose <paramref name="item"/> if it implements <see cref="IDisposable"/>.
     /// For <see cref="ITuple"/> values (e.g. value tuples) that don't implement IDisposable,
     /// each element is individually checked and disposed.
+    /// Accepts <c>object?</c> so the constrained overloads are preferred by the compiler.
     /// </summary>
-    public static void TryDispose<T>(this T item)
+    public static void TryDispose(this object? item)
     {
         if (item is IDisposable d)
         {
@@ -25,21 +26,28 @@ public static class DisposableExtensions
         {
             for (int i = 0; i < tuple.Length; i++)
             {
-                (tuple[i] as IDisposable)?.Dispose();
+                if (tuple[i] is IDisposable element)
+                    element.Dispose();
             }
         }
     }
 
     /// <summary>
+    /// Constrained overload for known-disposable types. Avoids boxing and the ITuple check.
+    /// </summary>
+    public static void TryDispose<T>(this T item) where T : IDisposable
+    {
+        item?.Dispose();
+    }
+
+    /// <summary>
     /// Overload for 2-element value tuples where the first element is <see cref="IDisposable"/>.
-    /// Avoids boxing the tuple into <see cref="ITuple"/>;
-    /// the compiler prefers this overload
-    /// for calls like <c>(OwnedResource, long).TryDispose()</c>.
+    /// Avoids boxing the tuple into <see cref="ITuple"/>.
     /// </summary>
     public static void TryDispose<T1, T2>(this in (T1, T2) item) where T1 : IDisposable
     {
-        item.Item1.Dispose();
-        (item.Item2 as IDisposable)?.Dispose();
+        item.Item1?.Dispose();
+        if (item.Item2 is IDisposable d2) d2.Dispose();
     }
 
     public static void DisposeItems<T>(this IEnumerable<T> items) where T : IDisposable


src/Nethermind/Nethermind.Core/Extensions/DisposableExtensions.cs
```diff
@@ -3,15 +3,31 @@
...
+        {
+            d.Dispose();
+        }
+        else if (item is ITuple tuple)
+        {
+            for (int i = 0; i < tuple.Length; i++)
+            {
+                (tuple[i] as IDisposable)?.Dispose();
+            }
+        }
```
LukaszRozmej, line 32: This will box. Which is fine, but maybe to avoid boxing add another overload. This one should be preferred in static resolution:

```
    public static void TryDispose<T>(this T tuple) where T : ITuple
    {
        for (int i = 0; i < tuple.Length; i++)
        {
            (tuple[i] as IDisposable)?.Dispose();
        }
    }
```

Could also add one for:

```
    public static void TryDispose<T>(this T item) where T : IDisposable
```

we could also add:

```
    public static void TryDispose<T1, T2>(this in ValueTuple<T1, T2> item) where T1 : IDisposable
```

and make sure when we use it first item is the disposable one.

LukaszRozmej:
> I don't fully like the fix, seems more like a hack. I'm thinking of introducing a `struct SizedResponse : IDisposable` with a Deconstruct method to replace the tuples, and constraining TData to `IDisposable` on MessageQueue and MessageDictionary. Looking at current usage, every `TData` type either implements `IDisposable` directly or is a tuple containing a disposable component so the constraint would fit.
> 
> Or if there is a better idiomatic way C# to handle this lmk

Not sure if worth it (but maybe yes, would have to see it). But even so It probably won't easily handle all the cases where we use ValueStructs in lists

src/Nethermind/Nethermind.Core/Extensions/DisposableExtensions.cs
newmanifold: Overload with 3 variant would work but for others 

`void TryDispose<T>(this T tuple) where T : ITuple`, 
`static void TryDispose<T>(this T item) where T : IDisposable`, compiler gives error.

We would need to remove current unconstrainted TryDispose

LukaszRozmej: the current one can just accept `object item` then

src/Nethermind/Nethermind.Core/Extensions/DisposableExtensions.cs
```diff
@@ -3,15 +3,43 @@
...
+    /// <summary>
+    /// Overload for 2-element value tuples where the first element is <see cref="IDisposable"/>.
+    /// Avoids boxing the tuple into <see cref="ITuple"/>;
+    /// the compiler prefers this overload
+    /// for calls like <c>(OwnedResource, long).TryDispose()</c>.
+    /// </summary>
+    public static void TryDispose<T1, T2>(this in (T1, T2) item) where T1 : IDisposable
+    {
+        item.Item1.Dispose();
+        (item.Item2 as IDisposable)?.Dispose();
```
LukaszRozmej: this will box to though :/

newmanifold: Yes, but if we remove this then T2 can still be disposable and we don't call dispose on it. Not sure whether thats correct.

What about these overload sets
- `public static void TryDispose(this object? item)` - old version
- `public static void TryDispose<T>(this T item) where T : IDisposable`
- `public static void TryDispose<T>(this in (T, long) item) where T : IDisposable` or `public static void TryDispose<T1, T2>(this in (T1, T2) item) where T1 : IDisposable  where T2: INumber<T2>` ?

But does compiler even calls other two overloads? i think compiler will always call `TryDispose(this object? item)` from inside MessageDictionary, as we doesn't constraint the type, so not sure whether they even help current usage

LukaszRozmej: Maybe we should mark all messages, thus `MessageBase` and `IEth66Message` as `IDisposable` and be done with it? Avoiding `TryDispose`?

newmanifold: We also need to mark TData as IDisposable to get rid of `TryDispose` in `MessageDictionary` and `MessageQueue`. We can do this by wrapping our response tuples in a disposable type.

For messages, `IEth66Message : IDisposable` would work. For MessageBase, there are classes that don't contain disposable/unmanaged resources those would need no-op Dispose methods. 

Instead, we can just add where `TMsg : MessageBase, IDisposable` on MessageQueue alongside `IEth66Message : IDisposable` with `SizedResponse` on `TData`.

Do we want this as part of this PR? or better do this as part of separate refactor?

LukaszRozmej: We can try in next PR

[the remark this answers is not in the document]
newmanifold: I think we have nullable type `analysis`, IDisposable? on T1 would fail to compile

PR #6722, opened 2024-02-14
Feature/snapsync server

asdacap:
- I'd like to merge snap sync server to master with its disabled so that I don't have to keep rebasing/merging.
- Think of it like a feature flag.
- Implementation is based on @tanishqjasoria's original implementation, but changed comparator to use TreePath and uses contextual visitor. It should work out of the box with any state db but might be slow.
- Here is how it perform. Metrics is nodes written measured from the syncing side. Only one peer and one thread. For reference, you'll need to sustain 1200k nodes per second to finish snap sync in 30 minute. Under most case, my SN850 will throttle it down to 900k nodes per second after about 5 minute.

| DB | Node per second |
| -- | -------|
| Geth | 600k-700k | 
| HalfPath | 170k-200k |
| Hash | 7k-8k |

- Hashdb work as expected as single thread IOPs is usually limited to 10k, compared to maximum nvme ssd throughput which is at least 500k. For the fun of it, when the serving node is using an optane drive, the node per second is 20k, which is disappointing. 
- It took about 27 hour to sync mainnet where the serving side uses optane drive. It'll probably take 2 to 3 days normally.
- This also means that turning it on by default is probably not a bad idea as it can't saturate any system resource except for background task queue.
- Note: Nethermind can't update snap sync pivot with only 1 node. Also, nethermind never use snap protocol in healing from another nethermind, so you'll need to modify something to test it.

## Changes

- Add snap server.
- Add tree path.

#### What types of changes does your code introduce?

- [X] New feature (a non-breaking change that adds functionality)

#### Requires testing

- [X] Yes
- [ ] No

#### If yes, did you write tests?

- [X] Yes
- [ ] No

#### Notes on testing

- Can confirm on halfpath can sync in about 3 hour.

PR #10842, opened 2026-03-16
fix(rlp): pass ValueDecoderContext by ref in DecodeStorageRootOnly

reject-i:
DecodeStorageRootOnly called DecodeStorageRoot without ref, so position changes were lost (ValueDecoderContext is a ref struct). Bug introduced in #6722 during RlpStream to ValueDecoderContext migration. Removed the duplicate non-ref DecodeStorageRoot overload and added a regression test.

### DIFF
--- a/.github/workflows/sync-supported-chains.yml
+++ b/.github/workflows/sync-supported-chains.yml
@@ -16,6 +16,10 @@ on:
         description: "Docker image to be used by action"
         default: ""
         required: false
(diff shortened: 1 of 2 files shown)


LukaszRozmej (approved):
fortunately this is one-off throw away ValueDecoderContext, so it doesn't really affect anything in current usages. But should be by `ref`

PR #10843, opened 2026-03-16
fix(db): return null for missing key in SimpleFilePublicKeyDb.Get

reject-i:
SimpleFilePublicKeyDb.Get used indexer directly which throws KeyNotFoundException for missing keys. Changed to TryGetValue to return null, matching IDb contract and behavior of other implementations like MemDb.

### DIFF
--- a/src/Nethermind/Nethermind.Db/SimpleFilePublicKeyDb.cs
+++ b/src/Nethermind/Nethermind.Db/SimpleFilePublicKeyDb.cs
@@ -60,7 +60,7 @@ namespace Nethermind.Db
             set => Set(key, value);
         }
 
-        public byte[]? Get(ReadOnlySpan<byte> key, ReadFlags flags = ReadFlags.None) 
+        public byte[]? Get(ReadOnlySpan<byte> key, ReadFlags flags = ReadFlags.None)
         {
             _cacheSpan.TryGetValue(key, out byte[]? value);
             return value;


LukaszRozmej (approved):

src/Nethermind/Nethermind.Db/SimpleFilePublicKeyDb.cs
```diff
@@ -60,7 +60,7 @@ public byte[]? this[ReadOnlySpan<byte> key]
             set => Set(key, value);
         }
 
-        public byte[]? Get(ReadOnlySpan<byte> key, ReadFlags flags = ReadFlags.None) => _cacheSpan[key];
+        public byte[]? Get(ReadOnlySpan<byte> key, ReadFlags flags = ReadFlags.None) => _cacheSpan.TryGetValue(key, out var value) ? value : null;
```
flcl42: Could you use specific type instead of var?

LukaszRozmej: you can avoid the branching too:

```
 _cacheSpan.TryGetValue(key, out byte[]? value)
return value;
```

LukaszRozmej:
```suggestion
        public byte[]? Get(ReadOnlySpan<byte> key, ReadFlags flags = ReadFlags.None) 
        {
            _cacheSpan.TryGetValue(key, out byte[]? value)
            return value;
        }
```

src/Nethermind/Nethermind.Db/SimpleFilePublicKeyDb.cs
```diff
@@ -60,7 +60,11 @@ public byte[]? this[ReadOnlySpan<byte> key]
             set => Set(key, value);
         }
 
-        public byte[]? Get(ReadOnlySpan<byte> key, ReadFlags flags = ReadFlags.None) => _cacheSpan[key];
+        public byte[]? Get(ReadOnlySpan<byte> key, ReadFlags flags = ReadFlags.None) 
+        {
+            _cacheSpan.TryGetValue(key, out byte[]? value)
```
LukaszRozmej:
```suggestion
            _cacheSpan.TryGetValue(key, out byte[]? value);
```

src/Nethermind/Nethermind.Db/SimpleFilePublicKeyDb.cs
```diff
@@ -60,7 +60,11 @@ public byte[]? this[ReadOnlySpan<byte> key]
             set => Set(key, value);
         }
 
-        public byte[]? Get(ReadOnlySpan<byte> key, ReadFlags flags = ReadFlags.None) => _cacheSpan[key];
+        public byte[]? Get(ReadOnlySpan<byte> key, ReadFlags flags = ReadFlags.None)
```
LukaszRozmej:
```suggestion
        public byte[]? Get(ReadOnlySpan<byte> key, ReadFlags flags = ReadFlags.None)
```

PR #10395, opened 2026-02-03
Harden rpc transaction validations

LukaszRozmej:
Replaces #10354

## Changes

- Harden RPC transaction validations

#### What types of changes does your code introduce?

- [x] Bugfix (a non-breaking change that fixes an issue)
- [x] Refactoring

#### Requires testing

- [x] Yes

#### If yes, did you write tests?

- [x] No

#### Notes on testing

Existing tests should do heavy lufting

PR #10846, opened 2026-03-16
fix: rpc-compat hive tests

deffrian:
Fixes Closes Resolves #

Fix prc-compat hive tests

## Changes

- Run rpc-compat tests on every pr
- Fix log index
- Remove zero fee error

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/.github/workflows/hive-tests.yml
+++ b/.github/workflows/hive-tests.yml
@@ -55,8 +55,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        test-suite: ${{ fromJson(
-            github.event_name == 'push' ||
+        test-suite: ${{ fromJson((
+            github.event_name == 'push' || 
             (github.event_name == 'workflow_dispatch' && github.event.inputs.test-suite == 'all')) &&
           '["ethereum/engine","ethereum/graphql","ethereum/rpc-compat","ethereum/sync"]' ||
           format('["{0}"]', github.event.inputs.test-suite)


.github/workflows/hive-tests.yml
```diff
@@ -1,6 +1,7 @@
 name: Hive tests
 
 on:
+  pull_request:
```
rubo: Is this a must? We already have a workflow carnival on every PR. I'd like some feedback on this.
@LukaszRozmej @benaadams

deffrian: Looks like everyone ignores master runs. Imo better to run rpc-compat tests on every pr, so we don't have these regressions again

LukaszRozmej: how long does it take?

src/Nethermind/Nethermind.Facade/Simulate/SimulateTxTracer.cs
```diff
@@ -42,17 +43,20 @@ public SimulateTxTracer(
         _currentBlockTimestamp = currentBlockTimestamp;
         _txIndex = txIndex;
         _logIndexStart = logIndexStart;
+        _isTracingTransfers = isTracingTransfers;
         IsTracingReceipt = true;
         IsTracingLogs = true;
-        IsTracingActions = isTracingTransfers;
+        IsTracingActions = true;
```
LukaszRozmej, line 49: Can you explain this change to me - I think it is functionally equivalent, but the previous setup was simpler. 
Basically if you still bail on actions with `if (!_isTracingTransfers) return;` then the only thing is that you call base. And base doesn't do anything.

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/EIP1559TransactionForRpc.cs
```diff
@@ -44,10 +44,6 @@ public override Result<Transaction> ToTransaction(bool validateUserInput = false
             if (GasPrice is not null && (MaxFeePerGas is not null || MaxPriorityFeePerGas is not null))
                 return RpcTransactionErrors.GasPriceInEip1559;
 
-            // Reject zero maxFeePerGas from user input
-            if (MaxFeePerGas?.IsZero == true)
-                return RpcTransactionErrors.ZeroMaxFeePerGas;
```
LukaszRozmej, line 49: Should we allow this from users? What does that mean for UX?

deffrian: Nothing really, the same as null. Hive tests expect us to accept zero, and I don't see a reason to not accept zero

src/Nethermind/Nethermind.Facade/Simulate/SimulateTxTracer.cs
deffrian: `public override void ReportActionError(EvmExceptionType evmExceptionType)`
We need to report evm exception regardless of `_isTracingTransfers`.

.github/workflows/hive-tests.yml
deffrian: 3 min

LukaszRozmej: should we run more test suites?

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/EIP1559TransactionForRpc.cs
LukaszRozmej: Zero MaxFeePerGas would mean it can't pay for basefee?
@flcl42 @marcindsobczak can you weigh in?

LukaszRozmej: Maybe `validateUserInput` should be `false` for hive?

LukaszRozmej: this was introduced in : https://github.com/NethermindEth/nethermind/pull/10395

flcl42: What is expected from the execution in case of MaxFeePerGas = 0?

deffrian: Zero base fee means we don't validate/charge gas:
```
protected static bool ShouldValidateGas(Transaction tx, ExecutionOptions opts)
            => !opts.HasFlag(ExecutionOptions.SkipValidation) || !tx.MaxFeePerGas.IsZero || !tx.MaxPriorityFeePerGas.IsZero;
```

.github/workflows/hive-tests.yml
deffrian: Others flake, so probably not a good idea to add them now

src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/EIP1559TransactionForRpc.cs
deffrian: I don't really see a reason to not allow zero. Maybe user want to disable gas validation explicitly

LukaszRozmej: What endpoint is it eth_call?

deffrian: and eth_simulate

LukaszRozmej: ok, should be fine

flcl42: If a proposer proposes a block with a tx with MaxFeePerGas = 0, will it fail when block will be received by a nethermind based validator?

PR #10848, opened 2026-03-17
Re-enable ReadyToRun AOT compilation

benaadams:
### ReadyToRun (R2R) AOT compilation
- Enable `PublishReadyToRun` and `PublishReadyToRunComposite` in `Nethermind.Runner.csproj` when a `RuntimeIdentifier` is specified, so crossgen2 AOT-compiles assemblies during publish
- Two-phase Docker restore: `dotnet restore --locked-mode` pins NuGet packages to the lock file, then `dotnet restore -r <rid> -p:PublishReadyToRun=true` fetches the crossgen2 runtime pack. The runtime pack version is deterministic — it's resolved from the SDK version, which is pinned by the Docker base image SHA digest. Per-RID lock files aren't feasible due to a [long-standing NuGet limitation](https://github.com/NuGet/Home/issues/8287) where `--locked-mode` rejects RID set changes.
- Add `publish-runner` CI job in `build-solutions.yml` to validate R2R publish on `linux-x64` and `linux-arm64`

### Profile-Guided Optimization (PGO)
- Add `Dockerfile.pgo` that builds Nethermind WITHOUT R2R and enables EventPipe to collect JIT profiling data during block processing
- Add `collect-pgo-profile.yml` workflow that:
  - Builds the PGO Docker image and runs it via EXPB against 5000 mainnet blocks
  - Converts the EventPipe `.nettrace` trace to `.mibc` using `dotnet-pgo create-mibc`
  - Extracts Nethermind assemblies from the Docker image and passes them as `--reference` for accurate generic method resolution
  - Runs weekly (Sunday 02:00 UTC) and creates a PR to update the checked-in `.mibc` profile
- Add `nethermind.mibc` (384KB) at `src/Nethermind/Nethermind.Runner/pgo/` — initial profile from 5000 mainnet blocks
- Add `Directory.Build.targets` that sets `PublishReadyToRunMibcPaths` when `PublishReadyToRun=true`, so crossgen2 uses the PGO profile for method layout, inlining, and hot/cold splitting — applies to all projects (Runner, BDN benchmarks)

### GC stability
- Set `DOTNET_GCRegionRange=0x800000000` (32 GiB cap) in production Dockerfiles to prevent `CLR_E_GC_OOM` (0x8013200E) crash loops when the runtime over-estimates available memory during GC initialization

### Build fixes
- Suppress `NU1900` (vulnerability metadata fetch failure) in `Directory.Build.props` so flaky NuGet feeds don't break builds
- Fix `Nethermind.KeyStore.csproj` overwriting inherited `$(NoWarn)` instead of appending
- Bump `actions/upload-artifact` to v7 (Node.js 24 support)

R2R was previously disabled in #7598 because the deep 26-level `ReleaseSpec` inheritance chain caused crossgen2 to take ~15 minutes. #10825 flattened that hierarchy, bringing AOT compilation down to ~10s, making it practical to re-enable.

## Types of changes

- [x] Optimization
- [x] Build-related changes

## Testing

- [x] CI validates R2R publish succeeds on linux-x64 and linux-arm64
- [x] PGO workflow produces valid `.mibc` artifact from mainnet blocks
- [x] Docker image builds successfully with R2R + PGO profile on both amd64 and arm64
- [x] Locally confirmed `PublishReadyToRunMibcPaths` is set and crossgen2 receives the profile

## Documentation

- [x] No documentation update required

## Remarks

- R2R is gated on `$(RuntimeIdentifier) != ''`, so regular `dotnet build` (no RID) is unaffected
- The `.mibc` profile is automatically updated weekly via the PGO workflow, which creates a PR with the new profile
- Docker image size increases slightly since R2R embeds precompiled native code alongside IL
- The `GCRegionRange` cap is virtual address space only — it doesn't limit actual heap usage, and is overridable via `-e DOTNET_GCRegionRange=...`

### DIFF
--- a/.github/workflows/run-expb-reproducible-benchmarks.yml
+++ b/.github/workflows/run-expb-reproducible-benchmarks.yml
@@ -633,7 +633,10 @@ jobs:
 
           trap on_terminate TERM INT
 
-          # Background: tail Nethermind container logs for visibility
+          # Save stdout fd before expb redirects it — container logs go here for live streaming
+          exec 3>&1
+
+          # Background: tail Nethermind container logs for live visibility
           (
             sleep 5
             for i in $(seq 1 60); do
@@ -642,8 +645,9 @@ jobs:
                 cid=$(docker ps --format '{{.ID}} {{.Image}}' 2>/dev/null | grep nethermind | head -1 | cut -d' ' -f1)
               fi
               if [[ -n "$cid" ]]; then
-                echo "=== Tailing Nethermind container ${cid} logs ==="
-                exec docker logs -f "$cid" 2>&1 | sed 's/^/[nethermind] /'
+                echo "=== Tailing Nethermind container ${cid} logs ===" >&3
+                docker logs -f "$cid" 2>&1 | sed 's/^/[nethermind] /' >&3
+                break
               fi
               sleep 2
             done
@@ -664,6 +668,7 @@ jobs:
           set -e
 
           kill -- -"${log_tail_pid}" 2>/dev/null || kill "${log_tail_pid}" 2>/dev/null || true
+          exec 3>&-
           cat "${RAW_RUN_LOG}"
           exit "${expb_exit_code}"
 
@@ -1384,7 +1389,10 @@ jobs:
 
           trap on_terminate TERM INT
 
-          # Background: tail Nethermind container logs for visibility
+          # Save stdout fd before expb redirects it — container logs go here for live streaming
+          exec 3>&1
+
+          # Background: tail Nethermind container logs for live visibility
           (
             sleep 5
             for i in $(seq 1 60); do
@@ -1393,8 +1401,9 @@ jobs:
                 cid=$(docker ps --format '{{.ID}} {{.Image}}' 2>/dev/null | grep nethermind | head -1 | cut -d' ' -f1)
               fi
               if [[ -n "$cid" ]]; then
-                echo "=== Tailing Nethermind container ${cid} logs ==="
-                exec docker logs -f "$cid" 2>&1 | sed 's/^/[nethermind] /'
+                echo "=== Tailing Nethermind container ${cid} logs ===" >&3
+                docker logs -f "$cid" 2>&1 | sed 's/^/[nethermind] /' >&3
+                break
               fi
               sleep 2
             done
@@ -1415,6 +1424,7 @@ jobs:
           set -e
 
           kill -- -"${log_tail_pid}" 2>/dev/null || kill "${log_tail_pid}" 2>/dev/null || true
+          exec 3>&-
           cat "${RAW_RUN_LOG}"
           exit "${expb_exit_code}"
 


Dockerfile
```diff
@@ -17,9 +17,10 @@ COPY global.json .
 COPY nuget.config .
 
 RUN arch=$([ "$TARGETARCH" = "amd64" ] && echo "x64" || echo "$TARGETARCH") && \
+  rid="linux-${arch}" && \
   cd src/Nethermind/Nethermind.Runner && \
   dotnet restore --locked-mode && \
-  dotnet publish -c $BUILD_CONFIG -a $arch -o /publish --no-restore --no-self-contained \
```
rubo, line 22: Why --no-restore gone? This could break reproducible builds

Dockerfile.chiseled
```diff
@@ -17,9 +17,10 @@ COPY global.json .
 COPY nuget.config .
 
 RUN arch=$([ "$TARGETARCH" = "amd64" ] && echo "x64" || echo "$TARGETARCH") && \
+  rid="linux-${arch}" && \
```
rubo: Can be made inline, no need for a variable

Dockerfile
benaadams: Fixed; though took a long time to get there 😅

Dockerfile.chiseled
benaadams: Fixed

src/Nethermind/Nethermind.Runner/Nethermind.Runner.csproj
```diff
@@ -16,6 +16,12 @@
     <UserSecretsId>03db39d0-4200-473e-9ff8-4a48d496381f</UserSecretsId>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(RuntimeIdentifier)' != ''">
+    <PublishReadyToRun>true</PublishReadyToRun>
+    <PublishReadyToRunComposite>true</PublishReadyToRunComposite>
+    <OptimizationPreference>Speed</OptimizationPreference>
+  </PropertyGroup>
```
LukaszRozmej, line 23: Should this be in files that apply to every project?

Directory.Packages.props
```diff
@@ -8,8 +8,8 @@
     <PackageVersion Include="AspNetCore.HealthChecks.UI.InMemory.Storage" Version="9.0.0" />
     <PackageVersion Include="Autofac" Version="9.1.0" />
     <PackageVersion Include="Autofac.Extensions.DependencyInjection" Version="10.0.0" />
-    <PackageVersion Include="BenchmarkDotNet" Version="0.15.8" />
-    <PackageVersion Include="BenchmarkDotNet.Diagnostics.Windows" Version="0.15.8" />
+    <PackageVersion Include="BenchmarkDotNet" Version="0.16.0-nightly.20260310.466" />
+    <PackageVersion Include="BenchmarkDotNet.Diagnostics.Windows" Version="0.16.0-nightly.20260310.466" />
```
LukaszRozmej, line 12: Should we use un-stable version?

benaadams: stable version doesn't have R2R profile; they've introduced it recently 😢

src/Nethermind/Nethermind.Runner/Nethermind.Runner.csproj
benaadams: Only kicks in during publish. Would be nice to do for nettest but I doubt people publish it to use it vs donet run

Dockerfile
```diff
@@ -12,14 +12,15 @@ ARG TARGETARCH
...
-COPY Directory.*.props .
+COPY Directory.*.props Directory.Build.targets ./
 COPY global.json .
 COPY nuget.config .
 
 RUN arch=$([ "$TARGETARCH" = "amd64" ] && echo "x64" || echo "$TARGETARCH") && \
   cd src/Nethermind/Nethermind.Runner && \
   dotnet restore --locked-mode && \
-  dotnet publish -c $BUILD_CONFIG -a $arch -o /publish --no-restore --no-self-contained \
+  dotnet restore -r "linux-${arch}" -p:PublishReadyToRun=true && \
```
rubo, line 22: This may break reproducibility

benaadams: Should have restored to lock; then restore platform to locked docker .NET runtime

benaadams: It won't restore the platform specific runtime stuff without not having the lock/second restore

> The offline --source approach doesn't work with PackageSourceMapping — NuGet requires sources to be named in the   config's mapping section. The crossgen2 runtime pack is already deterministic from the SDK version (pinned by Docker   image SHA digest), so the standard restore is reproducible.

rubo: Looks like an AI reply, so not that credible. This needs further testing.

benaadams: Most of the commits in the PR are trying to get it to compile with locked in any other way (unsuccessfully)

Can come up with a better way? Even tried platform specific lockfiles but couldn't get that to work either

rubo: I recall it was hard to keep it reproducible while not making disruptive and annoying changes for the team. Now it gets even harder. I'll try to take a look today.

benaadams: Suggestion is think can only create the right lock file if do it on the actual individual platforms (so have right sdk); but that might have just been AI blowing smoke

benaadams: i.e. since its local install that's doing the packages for the platform

PR #10854, opened 2026-03-18
feat: upgrade Claude CI/CD review workflow

newmanifold:
## Summary

Upgrades the Claude Code Action workflow to use dual-review.

### Changes
- **Dual-review strategy**: Dispatches a subagent to invoke the `review` skill while running an independent review in parallel, then aggregates and deduplicates findings
- **Opus model**: Switches from default Sonnet to Opus for higher review quality
- **Scoped permissions**: Adds `Agent` and `Skill` tools, denies `curl`/`wget`/`env`/`printenv` to prevent token exfiltration
- **Progress tracking**: Enables `track_progress` for live status updates on PRs
- **Full git history**: Adds `fetch-depth: 0` for `git merge-base`/`git show` support

### Security
- Exfiltration vectors (`curl`, `wget`, `env`, `printenv`) explicitly denied

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Enhancement to an existing feature
- [ ] Optimization
- [ ] Documentation

### DIFF
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -40,10 +40,9 @@ jobs:
             {
               "permissions": {
                 "allow": [
-                  "Bash(git merge-base:*)",
-                  "Bash(git show:*)",
-                  "Bash(gh pr diff:*)",
-                  "Bash(gh pr view:*)",
+                  "Bash(git *)",
+                  "Bash(gh:*)",
+                  "WebFetch",
                   "Agent",
                   "Skill"
                 ],
@@ -56,6 +55,8 @@ jobs:
               }
             }
           prompt: |
+            IMPORTANT: Before doing anything else, run `git fetch --deepen=100 origin` to ensure enough commit history is available for merge-base.
+
             When reviewing a pull request, use a dual-review strategy:
 
             1. **Dispatch subagent**: Use the Agent tool to launch a subagent that invokes the "review" skill via the Skill tool. The subagent should return ALL findings as structured text — this raw output will be included verbatim in the report.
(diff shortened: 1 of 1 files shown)


.github/workflows/claude-review.yml
```diff
@@ -27,9 +27,42 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
```
rubo: Isn't it too much? Maybe some fixed number?

rubo (commented):
Please give Codex some love too. Here or as a separate PR.

.github/workflows/claude-review.yml
newmanifold: changed to 100, even with `0` the git history overhead is < 400MB

smartprogrammer93 (approved):

kamilchodola:
@newmanifold Shall this be refreshed now? I believe this is pretty good idea in general to have 2-3 reviews even from same model but without cross-shared context. So we get more details.

newmanifold:
@kamilchodola yes, we can remove the opus 4.6 flag, otherwise it will destroy the session limits. But that skill is a bit heavier to run compared to Claude's built-in review skill so it will consume the session limit much faster if we run it every time. 

Maybe we can tune it to only run when someone requests a deep review, like @claude deep review or something like that. Wdyt?

AnkushinDaniil (approved):

stdevMac:
@newmanifold @kamilchodola stale-PR check: the opus pin landed via #12140 and fork-PR handling via #11789, and the workflow has been rewritten several times since this was opened. The dual-review idea would need a fresh PR against the current workflow anyway — OK to close this one? If there's no reply in the next few days, we'll close this as part of the stale-PR cleanup — it can always be reopened or resubmitted.

Issue #9906, opened 2025-12-09
Gnosis chain - invalid block in canonical index


jajaislanina:
**Description**
I have 2 gnosis archive nodes running Nethermind 1.35.3.
Both are currently in sync. Both nodes are running in Kubernetes as replicas of StatefulSet so they both share same config to the T.
On one node and on some blocks if i query eth_getBlockByNumber i get different results compared to the other node and the official block explorer - as in different block hash, included TX etc.
Next block shows correct parentHash (when compared to the explorer) however i would expect this to halt processing on the node - instead node continues to run.

**Steps to Reproduce**
In order to replicate the behavior, please provide a detailed list of steps:

**Actual behavior**
Node continues to sync after block with bad hash is computed.

**Expected behavior**
Node should halt if parentHash doesn't match previous block hash.

**Screenshots**
If applicable, please include screenshots to help illustrate the problem.

**Desktop (please complete the following information):**
Please provide the following information regarding your setup:
 - Operating System: ContainerOS (GCP)
 - Version: 1.35.3
 - Installation Method: Docker
 - Consensus Client: Lighthouse 7.0.1

**Additional context**
Please provide any additional context about the problem.

**Logs**
Node-0: Block 43061627

<img width="706" height="264" alt="Image" src="https://github.com/user-attachments/assets/6e7de44d-3379-402b-bfba-be24739cd0c3" />

Node-1: Block 43061627

<img width="675" height="263" alt="Image" src="https://github.com/user-attachments/assets/8fa7a2eb-e229-4aaf-9bef-ccb5c2e8298e" />

Both Node-0 and Node-1: Block 43061628

<img width="730" height="254" alt="Image" src="https://github.com/user-attachments/assets/41a4af16-9d10-4c47-a074-847e5cf08241" />

jajaislanina:
Yep. Query by hash works to get the correct block info.
Do you know what causes this issue and/or how to prevent it?
Restarting the node doesn't clear the cache - how do we fix it once it happens?

Issue #10861, opened 2026-03-18
eth_getBlockTransactionCountByNumber returns wrong count on Gnosis block 44207358


beruda:
## Description

`eth_getBlockTransactionCountByNumber` returns `0x14` (20) for Gnosis block 44207358 (`0x2a28cfe`), but the block actually contains 32 transactions. Other RPC methods return the correct count for the same block.

## Environment

- **Nethermind version**: `Nethermind/v1.36.0+31cb81b7/linux-x64/dotnet10.0.1`
- **Network**: Gnosis Chain
- **Block**: 44207358 (`0x2a28cfe`)
- **Block hash**: `0xa4d2943175ddfa46d1fa3d9dd843d34b013bfa382570e6810ce01fbbef817464`
- **Reproduced on**: Two independent Nethermind v1.36.0 nodes in different datacenters

## Observed behavior

| RPC Method | Result | Correct? |
|---|---|---|
| `eth_getBlockTransactionCountByNumber("0x2a28cfe")` | `0x14` (20) | **Wrong** |
| `eth_getBlockTransactionCountByHash("0xa4d29...")` | `0x20` (32) | Correct |
| `eth_getBlockByNumber("0x2a28cfe", false)` — tx array length | 32 | Correct |
| `eth_getBlockReceipts("0x2a28cfe")` — receipt count | 32 | Correct |

## Cross-client comparison (Reth v1.9.3)

Reth returns `0x20` (32) for `eth_getBlockTransactionCountByNumber` on the same block — consistent with the other methods.

## Scope

Neighboring blocks (44207356–44207360) all return matching transaction counts on both Nethermind and Reth, so this appears to be a data inconsistency specific to block 44207358.

## Steps to reproduce

```bash
# Returns 0x14 (20) — incorrect
wscat -c ws://<nethermind-gnosis-node>:8546 -x '{"jsonrpc":"2.0","method":"eth_getBlockTransactionCountByNumber","params":["0x2a28cfe"],"id":1}'

# Returns 0x20 (32) — correct
wscat -c ws://<nethermind-gnosis-node>:8546 -x '{"jsonrpc":"2.0","method":"eth_getBlockTransactionCountByHash","params":["0xa4d2943175ddfa46d1fa3d9dd843d34b013bfa382570e6810ce01fbbef817464"],"id":2}'
```

PR #10876, opened 2026-03-19
FCU Canonical chain corruption after beacon sync (3 fixes + startup healing)

svlachakis:
Fixes Closes Resolves https://github.com/NethermindEth/nethermind/issues/9906 & https://github.com/NethermindEth/nethermind/issues/10861

### Fix 1: Stale canonical markers after beacon sync

BlockDownloader calls `UpdateMainChain(wereProcessed: false)` to mark synced blocks canonical without updating Head. This creates three cases where stale markers survive:

1. FCU reorgs to a sibling at the same height (`previousHeadNumber == lastNumber`) — the unmark loop condition `previousHeadNumber > lastNumber` is false, so it is skipped entirely.
2. ePBS FCU to an ancestor — the downward loop clears down to `lastNumber` but cannot see beacon-synced markers above the stale `Head`.
3. A concurrent `MoveToMain` clears an intermediate level mid-scan, creating a gap — the old break on first non-canonical level stopped before clearing markers above it.

In all three cases orphaned blocks retain `HasBlockOnMainChain = true`, causing `eth_getBlockByNumber` to return the wrong block.

**Fix:** Added an unconditional upward scan after the downward unmark, starting from `Math.Max(previousHeadNumber, lastNumber) + 1`, that skips gaps rather than breaking on the first missing marker — covering all three cases in one loop.

---

### Fix 2: PoW best-difficulty fallback returns orphaned blocks in PoS

`GetBlockHashOnMainOrBestDifficultyHash` fell back to highest-TD when `HasBlockOnMainChain=false`. In PoS all blocks share `TD=0`, so the fallback returned whichever block it found first — after a reorg, the orphaned one.

**Fix:** Post-merge (`Head.TotalDifficulty ≥ TTD`), return `null` immediately with no fallback — matching geth's `ReadCanonicalHash` semantics.

---

### Fix 3: Startup repair for nodes already affected

Nodes that hit this bug before the fix may have corrupted markers on disk. A new config flag triggers a one-time repair at boot:

    --Init.HealCanonicalChain=true
    --Init.HealCanonicalChainDepth=8192

`BlockTree.HealCanonicalChain` runs two phases in a single atomic `BatchWrite`:

1. Upward scan — clears any `HasBlockOnMainChain=true` levels above head left by the sync path.
2. Backward walk — follows `ParentHash` from head for up to `HealCanonicalChainDepth` blocks, ensuring every ancestor has `HasBlockOnMainChain=true` at index 0 in its `ChainLevelInfo` slot.

---

### Tests

- `UpdateMainChain_WhenBeaconSyncMarksThenReorgsToSibling_DecanonalizesDescendant` — regression for Fix 1. Fails before, passes after.
- `UpdateMainChain_WhenFcuToAncestorWithStaleBeaconSyncedDescendants_ClearsAll` — ePBS scenario. Fails before Math.Max, passes after.
- `UpdateMainChain_WhenGapInBeaconSyncMarkersAndReorging_ClearsStaleMarkersAcrossGap` — gap race scenario. Fails with break-on-gap, passes with skip-gaps.
- `FindBlock_WhenBlockOrphanedAfterReorgInPoS_ReturnsNull` — regression for Fix 2. Fails before the post-merge guard, passes after.

---

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

## Testing Notes

Tested Healing functionallity on our Gnosis Archive node and it works as expected.
Added regression test on our test suite https://github.com/NethermindEth/nethermind-node-tests/pull/36

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain.Test/BlockchainProcessorTests.cs
+++ b/src/Nethermind/Nethermind.Blockchain.Test/BlockchainProcessorTests.cs
@@ -324,7 +324,7 @@ public class BlockchainProcessorTests
                 _processor.BlockAdded += OnBlockAdded;
                 try
                 {
-                    Task.Run(() =>
+                    Task suggestTask = Task.Run(() =>
                     {
                         try
                         {
@@ -344,6 +344,20 @@ public class BlockchainProcessorTests
                         suggestCompleted.Task.Wait(ProcessingWait),
                         Is.True,
                         $"Timed out waiting for {block.ToString(Block.Format.Short)} to complete suggestion");
+                    // Give the Task.Run time to finish the queue write inside
+                    // Enqueue(). BlockAdded fires before the recovery queue write,
+                    // so suggestCompleted resolving does not mean the block is in
+                    // the queue yet. Without this, consecutive Suggested() calls
+                    // can race their Task.Run threads to TryWrite, reordering
+                    // blocks in the recovery queue.
+                    //
+                    // Bounded to 100ms to avoid deadlock: when _queueCount == 1,
+                    // AllowSynchronousContinuations can inline ProcessBlocks() on
+                    // the Task.Run thread, which blocks until the test calls
+                    // Allow(). An infinite wait would deadlock. The timeout
+                    // expiring is harmless — single-block enqueues have no
+                    // ordering concern.
+                    suggestTask.Wait(100);
                 }
                 finally
                 {


smartprogrammer93:
@claude  review

src/Nethermind/Nethermind.Blockchain.Test/BlockTreeTests.cs
```diff
@@ -2201,4 +2201,1121 @@ public void On_UpdateMainBranch_UpdateSyncPivot_ToHeaderUnderReorgDepth()
...
+        blockTree.SuggestBlock(b1);
+        blockTree.SuggestBlock(b1Alt);
+        blockTree.SuggestBlock(b2);
+
+        // b1Alt wrongly canonical at H=1, b2 canonical at H=2 (head)
+        blockTree.UpdateMainChain(new[] { b1 }, wereProcessed: true, forceUpdateHeadBlock: true);
+        blockTree.UpdateMainChain(new[] { b2 }, wereProcessed: true, forceUpdateHeadBlock: true);
+        blockTree.UpdateMainChain(new[] { b1Alt }, wereProcessed: false); // breaks H=1
+
+        blockTree.FindBlock(1, BlockTreeLookupOptions.RequireCanonical)!.Hash
```
MarekM25, line 2544: those tests, AI slope or carefully analyzed tests?

src/Nethermind/Nethermind.Blockchain/BlockTree.cs
```diff
@@ -1035,6 +1063,89 @@ public void UpdateMainChain(IReadOnlyList<Block> blocks, bool wereProcessed, boo
             OnUpdateMainChain?.Invoke(this, new OnUpdateMainChainArgs(blocks, wereProcessed));
         }
 
+        public void HealCanonicalChain(Hash256 startHash, long maxBlockDepth)
```
MarekM25: Should this be in BlockTree?

src/Nethermind/Nethermind.Blockchain.Test/BlockTreeTests.cs
```diff
@@ -2201,4 +2201,1121 @@ public void On_UpdateMainBranch_UpdateSyncPivot_ToHeaderUnderReorgDepth()
...
+            "B must not be canonical after C was set");
+
+        // All three are still findable by hash (non-canonical lookup)
+        blockTree.FindBlock(blockA.Hash!, BlockTreeLookupOptions.None).Should().NotBeNull("A findable by hash");
+        blockTree.FindBlock(blockB.Hash!, BlockTreeLookupOptions.None).Should().NotBeNull("B findable by hash");
+        blockTree.FindBlock(blockC.Hash!, BlockTreeLookupOptions.None).Should().NotBeNull("C findable by hash");
+    }
+
+    [Test, MaxTime(Timeout.MaxTestTime)]
+    public void FindBlock_canonical_state_survives_BlockTree_reload_from_same_db()
```
MarekM25: which tests reproduce the core issue?

svlachakis: - `UpdateMainChain_sync_marks_descendant_canonical_then_reorg_to_sibling_decanonalizes_it` - regression test for the original fix. Failed before the upward scan, passes after.
- `UpdateMainChain_fcu_to_ancestor_with_stale_head_clears_all_beacon_synced_descendants` — ePBS scenario: FCU to ancestor with beacon-synced blocks above the stale head. The downward loop clears the stale head level; the upward scan clears the rest. Failed before Math.Max, passes after.
- `FindBlock_by_number_after_reorg_returns_null_not_orphaned_block_in_pos` - configures a pure-PoS spec (TTD=0), builds a two-block chain, reorgs to a shorter chain, asserts that FindBlock(orphanedHeight, None) returns null and the new canonical height returns the correct block. Failed before the guard, passes after.

These three are the ones that confirm the issue, failing before passing after. 

All the rest are AI generated and add some meaningful coverage that's why I kept them but they were passing even before the fix

src/Nethermind/Nethermind.Blockchain/BlockTree.cs
svlachakis: I think so..because it requires direct access to private level internals (LoadLevel,SwapToMain, FindIndex, HasBlockOnMainChain) that cannot be exposed through IBlockTree without polluting the interface.

Do you think there is any better place?

src/Nethermind/Nethermind.Blockchain.Test/BlockTreeTests.cs
svlachakis: update after cleanup:

- UpdateMainChain_WhenBeaconSyncMarksThenReorgsToSibling_DecanonalizesDescendant
- UpdateMainChain_WhenFcuToAncestorWithStaleBeaconSyncedDescendants_ClearsAll
- FindBlock_WhenBlockOrphanedAfterReorgInPoS_ReturnsNull
- UpdateMainChain_WhenGapInBeaconSyncMarkersAndReorging_ClearsStaleMarkersAcrossGap

src/Nethermind/Nethermind.Blockchain.Test/BlockTreeTests.cs
svlachakis: cleaned them up

src/Nethermind/Nethermind.Api/IInitConfig.cs
```diff
@@ -98,6 +98,19 @@ public interface IInitConfig : IConfig
 
     [ConfigItem(Description = "[TECHNICAL] True when in runner test. Disable some wait.", DefaultValue = "false", HiddenFromDocs = true)]
     bool InRunnerTest { get; set; }
+
+    [ConfigItem(
+        Description = "On startup, walk backward from the current head for up to `HealCanonicalChainDepth` " +
+                      "blocks via parentHash and repair any incorrect `HasBlockOnMainChain` markers left by " +
+                      "the beacon-sync path. Also clears stale canonical markers above the head. " +
+                      "Use this flag once after observing a canonical-mismatch (wrong `eth_getBlockByNumber` results).",
+        DefaultValue = "false")]
```
rubo: Let's simplify to make easier to understand:
```suggestion
    [ConfigItem(Description = "Whether to repair canonical-chain markers on startup after a canonical mismatch.", DefaultValue = "false")]
```
Feel free to revise if something important is missing.

src/Nethermind/Nethermind.Api/IInitConfig.cs
```diff
@@ -98,6 +98,19 @@ public interface IInitConfig : IConfig
...
+    bool HealCanonicalChain { get; set; }
+
+    [ConfigItem(
+        Description = "Number of blocks to walk back from the head when `HealCanonicalChain` is enabled.",
+        DefaultValue = "8192")]
```
rubo:
```suggestion
    [ConfigItem(
        Description = $"The number of blocks to walk back from the head when the `{nameof(HealCanonicalChain)}` is set to `true`.",
        DefaultValue = "8192")]
```

src/Nethermind/Nethermind.Api/IInitConfig.cs
svlachakis: done

src/Nethermind/Nethermind.Api/IInitConfig.cs
svlachakis: done

src/Nethermind/Nethermind.Blockchain/IBlockTree.cs
```diff
@@ -191,6 +191,14 @@ AddBlockResult Insert(Block block, BlockTreeInsertBlockOptions insertBlockOption
 
         void RecalculateTreeLevels();
 
+        /// <summary>
+        /// Repairs canonical chain markers starting from <paramref name="startHash"/>, walking backward
+        /// up to <paramref name="maxBlockDepth"/> blocks. Also removes stale canonical markers above
+        /// <paramref name="startHash"/> left by the beacon-sync path. Safe to run at startup after
+        /// observing canonical mismatches (wrong <c>eth_getBlockByNumber</c> results).
+        /// </summary>
+        void HealCanonicalChain(Hash256 startHash, long maxBlockDepth);
```
LukaszRozmej: Maybe this should be on different interface? (interface segregation)

src/Nethermind/Nethermind.Blockchain/BlockTree.cs
LukaszRozmej: BlockTree - yes
IBlockTree - probably no

LukaszRozmej: Maybe we should also split BlockTree to multiple files this would go to `BlockTree.Healing.cs` ?

src/Nethermind/Nethermind.Blockchain/BlockTree.cs
```diff
@@ -1035,6 +1047,84 @@ public void UpdateMainChain(IReadOnlyList<Block> blocks, bool wereProcessed, boo
...
+            return cleared;
+        }
+
+        private long RepairMarkersBelow(BlockHeader start, long maxBlockDepth, BatchWrite batch)
+        {
+            long repairedCount = 0L;
+            long blocksWalked = 0L;
+            BlockHeader? current = start;
+
+            while (current is not null && blocksWalked <= maxBlockDepth)
```
LukaszRozmej: current is never null? (fixed in #10972)

src/Nethermind/Nethermind.Blockchain/IBlockTree.cs
svlachakis: yes I also like that better, done

src/Nethermind/Nethermind.Blockchain/BlockTree.cs
svlachakis: done

src/Nethermind/Nethermind.Blockchain/ReadOnlyBlockTree.cs
```diff
@@ -193,6 +193,7 @@ public int DeleteChainSlice(in long startNumber, long? endNumber = null, bool fo
         public bool IsBetterThanHead(BlockHeader? header) => _wrapped.IsBetterThanHead(header);
         public void UpdateBeaconMainChain(BlockInfo[]? blockInfos, long clearBeaconMainChainStartPoint) => throw new InvalidOperationException($"{nameof(ReadOnlyBlockTree)} does not expect {nameof(UpdateBeaconMainChain)} calls");
         public void RecalculateTreeLevels() => throw new InvalidOperationException($"{nameof(ReadOnlyBlockTree)} does not expect {nameof(RecalculateTreeLevels)} calls");
+        public void HealCanonicalChain(Hash256 startHash, long maxBlockDepth) => throw new InvalidOperationException($"{nameof(ReadOnlyBlockTree)} does not expect {nameof(HealCanonicalChain)} calls");
```
LukaszRozmej: No need for this now

src/Nethermind/Nethermind.Consensus/Stateless/StatelessBlockTree.cs
```diff
@@ -192,6 +192,9 @@ public void UpdateBeaconMainChain(BlockInfo[]? blockInfos, long clearBeaconMainC
     public void RecalculateTreeLevels()
         => throw new NotSupportedException();
 
+    public void HealCanonicalChain(Hash256 startHash, long maxBlockDepth)
```
LukaszRozmej: No need for this now

src/Nethermind/Nethermind.Init/Modules/BlockTreeModule.cs
```diff
@@ -46,6 +46,7 @@ protected override void Load(ContainerBuilder builder)
             .AddSingleton<IHistoryPruner, HistoryPruner>()
             .AddSingleton<IBlockTree, BlockTree>()
             .Bind<IBlockFinder, IBlockTree>()
+            .AddSingleton<IBlockTreeHealer, IBlockTree>((bt) => (IBlockTreeHealer)bt)
```
LukaszRozmej, line 50: Maybe better to register `BlockTree` as `BlockTree` and register interfaces upon it?

svlachakis: This change is breaking too many tests because we mock `IBlockTree` directly. I reverted back to the ugly way for that PR and will create a task in the backlog for refactoring all those tests

Issue: https://github.com/NethermindEth/nethermind/issues/10976

src/Nethermind/Nethermind.Init/Modules/BlockTreeModule.cs
```diff
@@ -44,8 +44,10 @@ protected override void Load(ContainerBuilder builder)
             )
             .AddSingleton<IReceiptFinder, FullInfoReceiptFinder>()
             .AddSingleton<IHistoryPruner, HistoryPruner>()
+            .AddSingleton<BlockTree>()
```
asdacap, line 47: Unnecessary, the next line do the same thing.

svlachakis: fixed here https://github.com/NethermindEth/nethermind/pull/10982

PR #10887, opened 2026-03-19
Fix buffer leak tests to use PooledBufferLeakDetector

newmanifold:
## Changes

- Refactor discovery message serializer leak tests to use `PooledBufferLeakDetector` instead of manual ref-count assertions - They don't assert any leakage, just the refcounts
- Replace manual `SafeRelease()` calls with `using DisposableByteBuffer` pattern across all serializer tests
- Extract `PooledBufferLeakDetector.CreateAllocator()` factory so a single allocator can be shared across a test fixture, avoiding per-test arena allocation overhead
- Track initial allocation count in `PooledBufferLeakDetector` so it works correctly with shared allocators that may already have outstanding allocations
- Removed Redundant tests; instead added leak detector in original tests
- Removed AssertNoLeaks method, it was error prone with IDisposable

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

The leak detection tests themselves are the tests — they verify that serializer round-trips don't leak pooled DotNetty buffers. The `PooledBufferLeakDetector` asserts zero net allocations on dispose.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Network.Discovery.Test/DiscoveryMessageSerializerTests.cs
+++ b/src/Nethermind/Nethermind.Network.Discovery.Test/DiscoveryMessageSerializerTests.cs
@@ -34,6 +34,7 @@ public class DiscoveryMessageSerializerTests
     private readonly IMessageSerializationService _messageSerializationService;
     private readonly ITimestamper _timestamper;
     private readonly PooledByteBufferAllocator _leakDetectionAllocator = PooledBufferLeakDetector.CreateAllocator();
+
     public DiscoveryMessageSerializerTests()
     {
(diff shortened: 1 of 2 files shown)


newmanifold:
@claude Who is going to point about test redundancy? Also don't suggest the approaches that you are just speculating upon.
This time use our `review` skill in a sub agent and aggregate results, if you don't have permissions let me know about that.

LukaszRozmej (approved):

PR #10891, opened 2026-03-19
Geth type unification

### LINKED ISSUE #9685: Unify types with Geth
(Transaction) Nonce: UInt256->ulong
(Transaction) GasLimit long->ulong
(Transaction) SpentGas long->ulong

(Receipt) GasUsed long->ulong
(Receipt) GasUsedTotal long->ulong

(BlockHeader) GasLimit long->ulong
(BlockHeader) GasUsed long->ulong

Could use discussion:
(BlockHeader) Number long->ulong/UInt256: Geth uses big.Int so UInt256, but IMO there is no point and ulong would suffice. It will also speed up any lookups in hashmaps ect and save some memory space.
(Receipt) BlockNumber - as above 

The changes need to cascade throughout the code-base. The end-result code should have almost no casts. 

Each actual cast remaining should be carefully examined and listed, every non-trivial one, which could cause over or underflow should be documented.

One place where we want to keep current scheme is LogIndex for performance reasons.

It would be nice to provide a PR to: https://github.com/NethermindEth/nethermind-arbitrum with the update after. It will increase the speed of merging your PR.

markolazic01:
Resolves #9685

## Changes

- Unifying types with Geth, as described in issue #9685.

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks

In development.
This is pushed forward as a draft PR in order to seek further navigation.
Changes are huge, and this is approximately one half of all changes that will be present. 
This is also my second iteration on this cause. I am willing to make needed adaptations per request and separate this PR into multiple PRs for the sake of simplifying the review process (would probably be better to do that).

I considered carefully the request to do as few casts as possible but that might have resulted in even more changes, if needed I can change the path to going a different way.

Please provide feedback.


src/Nethermind/Nethermind.AuRa.Test/Reward/AuRaRewardCalculatorTests.cs
```diff
@@ -95,7 +95,7 @@ public void constructor_throws_ArgumentException_on_BlockRewardContractTransitio
         [TestCase(9, 200ul)]
         public void calculates_rewards_correctly_before_contract_transition(long blockNumber, ulong expectedReward)
         {
-            _block.Header.Number = blockNumber;
+            _block.Header.Number = (ulong)blockNumber;
```
LukaszRozmej, line 98: Cascade the changes!

LukaszRozmej:
I will close it for now as we have too much noise, this is not even close to what it needs to be. Feel free to reopen if it will be close.

PR #10894, opened 2026-03-20
fix(db): reuse ReadOptions in ColumnDbSnapshot to reduce GC pressure

smartprogrammer93:
## Changes

- Share 2 `ReadOptions` instances (normal + cache-miss) across all 7 column `RocksDbReader` instances in `ColumnDbSnapshot`, instead of creating 14 separate ones
- Explicitly destroy native `ReadOptions` handles via `rocksdb_readoptions_destroy` + `GC.SuppressFinalize` in `ColumnDbSnapshot.Dispose()`, with double-dispose guard
- Replace `Dictionary<T, IReadOnlyKeyValueStore>` with flat array indexed by enum ordinal via `Convert.ToInt32` (safe for all enum underlying types)
- Share a single `Func<ReadOptions>` delegate across all column readers instead of 7 separate closures
- Cache column keys and max ordinal on parent `ColumnsDb<T>` with volatile fields for thread safety
- Add a new `RocksDbReader` constructor that accepts pre-created `ReadOptions` for shared use

**Root cause:** `ReadOptions` in RocksDbSharp has a finalizer (`~ReadOptions()`) but does not implement `IDisposable`. Each `ColumnDbSnapshot` created 14 finalizable `ReadOptions` objects (2 per column x 7 columns), plus a `Dictionary`, 7 `RocksDbReader` instances, and 7 closure delegates. In FlatState block processing, thousands of snapshots per second produce tens of thousands of finalizable objects that survive Gen0, get promoted to Gen1/Gen2, and trigger expensive stop-the-world GC collections.

**Profiling data** (ContractCall_200 FlatState, 2000 block processing calls per round, 20 rounds):

| Metric | FlatState (before) | FlatState (after) | Trie |
|--------|-----------|-----------|------|
| Gen0 collections/round | 38 | 35 | 30 |
| Gen1 collections/round | **9** | **6.8** | 0 |
| Gen2 collections/round | **1.3** | **0.8** | 0 |
| Total Gen1 (20 rounds) | 174 | **135** | 2 |
| Total Gen2 (20 rounds) | 26 | **16** | 0 |
| Alloc/round | 580MB | 520MB | 479MB |
| FinalizationPending | **4,402** | **20** | 90 |
| Gen2 heap size | 20MB | 17MB | 7MB |

Remaining Gen1/Gen2 pressure is from per-scope FlatState infrastructure objects (`FlatWorldStateScope`, `SnapshotBundle`, `ReadOnlySnapshotBundle`, `StateTree`, etc.) that require deeper architectural changes to pool.

## Types of changes

- [x] Optimization

#### Requires testing

- [x] Yes

- [ ] Yes
- [x] No

#### Notes on testing

Validated via ProfileRunner benchmark with per-round GC tracking. Existing tests should pass since the behavior is identical - same ReadOptions configuration, just shared instead of duplicated. The array-indexed lookup via `Convert.ToInt32` is functionally equivalent to the dictionary lookup and safe for all enum underlying types.

- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Db.Rocks/ColumnsDb.cs
+++ b/src/Nethermind/Nethermind.Db.Rocks/ColumnsDb.cs
@@ -260,7 +260,7 @@ public class ColumnsDb<T> : DbOnTheRocks, IColumnsDb<T> where T : struct, Enum
 
         public IReadOnlyKeyValueStore GetColumn(T key)
         {
-            ObjectDisposedException.ThrowIf(_disposed != 0, this);
+            ObjectDisposedException.ThrowIf(Volatile.Read(ref _disposed) != 0, this);
 
             int ordinal = EnumToInt(key);
             if ((uint)ordinal >= (uint)_readers.Length || _readers[ordinal] is null)


src/Nethermind/Nethermind.Db.Rocks/RocksDbReader.cs
```diff
@@ -40,6 +40,26 @@ public RocksDbReader(DbOnTheRocks mainDb,
...
+    /// Constructor that accepts pre-created <see cref="ReadOptions"/> instead of a factory.
+    /// Used by <see cref="ColumnsDb{T}.ColumnDbSnapshot"/> to share a single pair of ReadOptions
+    /// across all column readers, avoiding per-reader native handle allocation and finalizer pressure.
+    /// </summary>
+    public RocksDbReader(DbOnTheRocks mainDb,
+        ReadOptions options,
+        ReadOptions hintCacheMissOptions,
+        Func<ReadOptions> readOptionsFactory,
+        DbOnTheRocks.IteratorManager? iteratorManager = null,
+        ColumnFamilyHandle? columnFamily = null)
```
LukaszRozmej, line 47: Can this constructor be called by the other constructor with `: this()`? I think so

src/Nethermind/Nethermind.Db.Rocks/ColumnsDb.cs
```diff
@@ -158,31 +163,98 @@ IColumnDbSnapshot<T> IColumnsDb<T>.CreateSnapshot()
...
         public IReadOnlyKeyValueStore GetColumn(T key)
         {
-            return _columnDbs[key];
+            return _readers[Convert.ToInt32(key)];
         }
 
         public void Dispose()
         {
-            snapshot.Dispose();
+            if (_disposed) return;
```
LukaszRozmej: This doesn't need to be threadsafe? 

We could do - although would need to by `byte` not `bool`
```suggestion
            bool disposed = Interlocked.Exchange(ref _disposed, true);
            if (disposed) return;
```

src/Nethermind/Nethermind.Db.Rocks/ColumnsDb.cs
```diff
@@ -158,31 +163,98 @@ IColumnDbSnapshot<T> IColumnsDb<T>.CreateSnapshot()
...
+                    _sharedCacheMissReadOptions,
+                    readOptionsFactory,
+                    columnFamily: columnsDb._columnDbs[k]._columnFamily);
+            }
+        }
 
         public IReadOnlyKeyValueStore GetColumn(T key)
         {
-            return _columnDbs[key];
+            return _readers[Convert.ToInt32(key)];
```
LukaszRozmej: can be single line with `=>`

src/Nethermind/Nethermind.Db.Rocks/ColumnsDb.cs
```diff
@@ -158,31 +163,98 @@ IColumnDbSnapshot<T> IColumnsDb<T>.CreateSnapshot()
...
+        public ColumnDbSnapshot(ColumnsDb<T> columnsDb, Snapshot snapshot)
+        {
+            _snapshot = snapshot;
+
+            // Create two shared ReadOptions for all column readers instead of 2 per reader.
+            // ReadOptions in RocksDbSharp has a finalizer but no IDisposable — creating many
+            // short-lived instances causes Gen1/Gen2 GC pressure from finalizer queue buildup.
+            _sharedReadOptions = new ReadOptions();
+            _sharedReadOptions.SetVerifyChecksums(columnsDb.VerifyChecksum);
+            _sharedReadOptions.SetSnapshot(snapshot);
```
LukaszRozmej: Helper method

src/Nethermind/Nethermind.Db.Rocks/ColumnsDb.cs
```diff
@@ -158,31 +163,98 @@ IColumnDbSnapshot<T> IColumnsDb<T>.CreateSnapshot()
...
+
+            _sharedCacheMissReadOptions = new ReadOptions();
+            _sharedCacheMissReadOptions.SetVerifyChecksums(columnsDb.VerifyChecksum);
+            _sharedCacheMissReadOptions.SetSnapshot(snapshot);
+            _sharedCacheMissReadOptions.SetFillCache(false);
```
LukaszRozmej: Helper method

src/Nethermind/Nethermind.Db.Rocks/ColumnsDb.cs
```diff
@@ -158,31 +163,98 @@ IColumnDbSnapshot<T> IColumnsDb<T>.CreateSnapshot()
...
+            T[]? keys = columnsDb._cachedColumnKeys;
+            if (keys is null)
+            {
+                IDictionary<T, ColumnDb> columnDbs = columnsDb._columnDbs;
+                keys = new T[columnDbs.Count];
+                int idx = 0;
+                foreach (T key in columnDbs.Keys)
+                    keys[idx++] = key;
+                columnsDb._cachedColumnKeys = keys;
+            }
```
LukaszRozmej: helper method

src/Nethermind/Nethermind.Db.Rocks/ColumnsDb.cs
```diff
@@ -158,31 +163,98 @@ IColumnDbSnapshot<T> IColumnsDb<T>.CreateSnapshot()
...
+
+            if (columnsDb._cachedMaxOrdinal < 0)
+            {
+                int max = 0;
+                for (int i = 0; i < keys.Length; i++)
+                    max = Math.Max(max, Convert.ToInt32(keys[i]));
+                columnsDb._cachedMaxOrdinal = max;
+            }
```
LukaszRozmej: helper method

src/Nethermind/Nethermind.Db.Rocks/ColumnsDb.cs
```diff
@@ -158,31 +163,98 @@ IColumnDbSnapshot<T> IColumnsDb<T>.CreateSnapshot()
...
+            for (int i = 0; i < keys.Length; i++)
+            {
+                T k = keys[i];
+                _readers[Convert.ToInt32(k)] = new RocksDbReader(
+                    columnsDb,
+                    _sharedReadOptions,
+                    _sharedCacheMissReadOptions,
+                    readOptionsFactory,
+                    columnFamily: columnsDb._columnDbs[k]._columnFamily);
+            }
```
LukaszRozmej: Helper method

src/Nethermind/Nethermind.Db.Rocks/RocksDbReader.cs
```diff
@@ -22,22 +22,36 @@ public class RocksDbReader : ISortedKeyValueStore
...
+    {
+        _hintCacheMissOptions.SetFillCache(false);
+    }
+
+    /// <summary>
+    /// Constructor that accepts pre-created <see cref="ReadOptions"/> instead of a factory.
+    /// Used by <see cref="ColumnsDb{T}.ColumnDbSnapshot"/> to share a single pair of ReadOptions
+    /// across all column readers, avoiding per-reader native handle allocation and finalizer pressure.
+    /// </summary>
+    public RocksDbReader(DbOnTheRocks mainDb,
```
svlachakis: Maybe we could make this one internal? will prevent external callers from passing in ReadOptions they might dispose independently.

PR #10753, opened 2026-03-08
Fix event handler leaks, reduce allocations, and refactor Network layer

benaadams:
## Changes

- Fix event handler leaks in `ProtocolsManager`, `PeerManager`, `RlpxHost`, `DiscoveryManager`, `DiscoveryApp`,   `CompositeDiscoveryApp`, `NodeLifecycleManager`, and `NettyDiscoveryHandler` by tracking subscriptions and unsubscribing on session/component disposal
- Implement `IDisposable` on `ProtocolsManager` with `SessionSubscription`/`ProtocolHandlerSubscription` wrappers for deterministic event handler cleanup
- Cache event handler and `Func<>` delegates in fields to avoid repeated allocations on subscribe/unsubscribe
- Replace `Action<T>` delegates in `MessageQueue`/`MessageDictionary` with `IMessageSender<T>` interface to eliminate per-construction delegate allocations
- Rewrite `PacketSender` from async to synchronous with `ContinueWith`, removing async state machine overhead on the hot send path
- Add custom `DisconnectEventHandlers` collection on `Session` enabling safe unsubscribe during dispatch with swap-and-pop removal
- Extract channel initializer classes in `RlpxHost`, replace closure-capturing lambdas with cached delegates and named methods
- Convert `NodeTable` sort lambda to struct comparer for devirtualized sorting 
- Improve shutdown ordering in `PeerManager` (replace `ManualResetEventSlim` with `SemaphoreSlim`, add `_isStopping` guard, track sessions)
- Add RLP deserialization collection-size limits to Snap and Eth message serializers
- Move all string interpolation in `P2PProtocolHandler` and `BackgroundTaskSchedulerWrapper` into `[MethodImpl(NoInlining)]` local functions to keep hot paths lean and reduce icache pressure
- Apply `[MethodImpl(NoInlining)]` to logging local functions with interpolated strings across `Session`, `PeerPool`, `PeerManager`, `ProtocolsManager`, `RlpxHost`, `ZeroSnappyEncoder`, `DiscoveryManager`, `NettyDiscoveryHandler`, and others
- Apply `[DoesNotReturn, StackTraceHidden]` to throw helpers
- Replace `ArgumentNullException` constructor pattern with `ArgumentNullException.ThrowIfNull`
- Take local copies of field arrays before iteration for JIT-friendly access patterns

#### What types of changes does your code introduce?

- [x] Bugfix (a non-breaking change that fixes an issue)
- [x] Optimization
- [x] Refactoring

#### Requires testing

[...]

PR #10902, opened 2026-03-20
fix: fix SnapSync regression from network layer refactor

kamilchodola:
## Summary

Fixes issues introduced in #10753 that caused SnapSync to regress from **25% in 5 minutes to 1% in 15 minutes** on Hoodi:

- **StorageRange deserialization limit too low** (critical): `MaxResponseSlotsPerAccount` was 16,384 (~1.08 MB). Large-storage contracts return up to 45,000+ slots per response (up to 3 MB). When the limit was hit, `RlpLimitException` triggered `InitiateDisconnect(MessageLimitsBreached)`, disconnecting and banning the serving peer for 15 minutes with -10,000 reputation. This progressively banned all good peers. Raised to 131,072.
- **Session.Handshake NodeId mismatch**: Only disconnect outbound peers on identity mismatch for static/bootnode peers (operator-verified). Discovered peers accept the new identity — stale discovery data is common and benign. Also fixes `HandshakeComplete` firing after `InitiateDisconnect` on doomed sessions.
- **SemaphoreSlim signal consumption**: `EnsureAvailableActivePeerSlotAsync` polling loop was consuming signals meant for the main peer update loop. Replaced `_peerUpdateRequested.WaitAsync` with `Task.Delay` in the polling path.
- **Lock scope reduction**: `OnDisconnected` and `OnHandshakeComplete` held `_sessionLock` for their entire body, serializing all session events. Now only guards session bookkeeping; peer processing runs outside the lock.
- **Dedicated thread for peer update loop**: Restored `TaskCreationOptions.LongRunning` with `.Unwrap()` (fixing the original `Task<Task>` bug), so the loop gets a dedicated thread instead of competing on the thread pool.

## Test plan

- [x] Existing `PeerManagerTests` pass (35/35)
- [x] Existing `SessionTests` pass + 3 new tests added (44/44)
- [x] All Snap serializer tests pass (9/9)
- [ ] Run Hoodi SnapSync and verify sync rate recovers to pre-#10753 levels (~25% in 5 minutes)


src/Nethermind/Nethermind.Network/PeerManager.cs
```diff
@@ -191,22 +191,23 @@ public void Start()
...
+                }
+                catch (Exception e) when (e is not OperationCanceledException)
+                {
+                    if (_logger.IsError) _logger.Error("Peer update loop encountered an exception.", e);
+                }
+                catch (OperationCanceledException)
+                {
+                    if (_logger.IsDebug) _logger.Debug("Peer update loop stopped.");
+                }
+            }, _cancellationTokenSource.Token, TaskCreationOptions.LongRunning, TaskScheduler.Default).Unwrap();
```
LukaszRozmej: I think we shoulnd't use LongRunning.
with passing cancellation token won't that potentially throw cancellation exception?

kamilchodola: done

smartprogrammer93 (dismissed):

LukaszRozmej:
@benaadams @flcl42 can you review?

asdacap (dismissed):

PR #10903, opened 2026-03-20
chore: add fix-nethtest agent skill for EF test debugging

LukaszRozmej:
## Summary

- Adds `/fix-nethtest` slash command — an agent skill for diagnosing and fixing failing Ethereum Foundation tests run with `Nethermind.Test.Runner`
- Auto-detects test type (state vs blockchain) from JSON structure
- Provides a structured 4-phase diagnostic workflow: run → classify → root cause → fix
- Documents common bug patterns: RLP deserialization, gas accounting, state root mismatches, validation errors, missing header defaults, instruction runtime guards, blob schedule overrides

## Motivation

Debugging EF test failures is a recurring task that follows a predictable pattern. This skill encodes that workflow so it can be invoked with `/fix-nethtest <test-file>` and the agent walks through the right diagnostic steps automatically.

## Changes

- [x] Other (CI/infrastructure/automation)

## Testing

Skill structure follows the existing `.agents/skills/` convention (see `review` skill). Symlink at `.claude/skills/fix-nethtest` points to canonical location.

### DIFF
--- a/.agents/skills/fix-nethtest/SKILL.md
+++ b/.agents/skills/fix-nethtest/SKILL.md
@@ -140,7 +140,11 @@ This is the most common failure pattern for new forks. Follow these steps:
 
 1. **Apply the minimal fix** — prefer spec flag checks over type checks, add missing defaults, or fix gas calculations
 2. **Re-run the failing test** to verify the fix resolves the issue
-3. **Report the result** — include:
+3. If test still fails:
+   - Re-read the trace output
+   - Return to Phase 2 with the new trace
+   - Repeat until pass or root cause is confirmed as upstream
+4. **Report the result** — include:
    - Root cause (one sentence)
    - What was fixed (file:line)
    - Verification result (pass/fail + new state root)


.agents/skills/fix-nethtest/SKILL.md
```diff
@@ -0,0 +1,181 @@
...
+2. Look at `Ethereum.Test.Base/JsonToEthereumTest.cs` and the JSON model classes
+3. Check `SpecNameParser.cs` for unsupported fork names
+
+---
+
+## Phase 4 — Fix and verify
+
+1. **Apply the minimal fix** — prefer spec flag checks over type checks, add missing defaults, or fix gas calculations
+2. **Re-run the failing test** to verify the fix resolves the issue
+3. **Report the result** — include:
```
newmanifold:
```suggestion
3. If test still fails:
   - Re-read the trace output
   - Return to Phase 2 with the new trace
   - Repeat until pass or root cause is confirmed as upstream
4. **Report the result** — include:
```

If we want it in a loop

PR #10905, opened 2026-03-20
fix(test): eliminate flakiness in BlockchainProcessorTests

LukaszRozmej:
## Summary
- Make block fields instance-scoped (not `static`) so parallel test instances don't share mutable `Header.Author` state mutated by `RecoverData`
- Reduce mock `Monitor.Wait` re-check interval from 10s to 200ms (`MockRecheckInterval`) so the mock catches pre-populated signals faster
- Remove `[Retry(3)]` from all three tests that had it

## Test plan
- [x] All 18 `BlockchainProcessorTests` pass locally (2 consecutive runs)
- [x] CI passes without any retries on the previously flaky tests

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain.Test/BlockchainProcessorTests.cs
+++ b/src/Nethermind/Nethermind.Blockchain.Test/BlockchainProcessorTests.cs
@@ -118,7 +118,7 @@ public class BlockchainProcessorTests
 
                         if (notYet)
                         {
-                            Monitor.Wait(_gate, ProcessingWait);
+                            Monitor.Wait(_gate, MockRecheckInterval);
                         }
                         else
                         {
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Blockchain.Test/BlockchainProcessorTests.cs
```diff
@@ -500,17 +501,33 @@ private static class When
...
+    private readonly Block _block3D6;
+    private readonly Block _block4D8;
+    private readonly Block _block5D10;
+    private readonly Block _blockB2D4;
+    private readonly Block _blockB3D8;
+    private readonly Block _blockC2D100;
+    private readonly Block _blockD2D200;
+    private readonly Block _blockE2D300;
+
+    public BlockchainProcessorTests()
```
alexb5dh, line 518: Why move initialization to constructor?

LukaszRozmej: non-static fields dependent on each other

PR #10921, opened 2026-03-23
Dispose on exception

flcl42:
## Changes

- Dispose in case of exceptions

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus.Test/Scheduler/BackgroundTaskSchedulerTests.cs
+++ b/src/Nethermind/Nethermind.Consensus.Test/Scheduler/BackgroundTaskSchedulerTests.cs
@@ -45,6 +45,16 @@ public class BackgroundTaskSchedulerTests
     }
 
     [Test]
+    public async Task DisposeAsync_should_complete_when_scheduler_is_idle()
+    {
+        BackgroundTaskScheduler scheduler = new BackgroundTaskScheduler(_branchProcessor, _chainHeadInfo, 1, 65536, LimboLogs.Instance);
+
+        Assert.DoesNotThrowAsync(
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Network/P2P/Subprotocols/Snap/Messages/StorageRangesMessageSerializer.cs
```diff
@@ -59,28 +59,38 @@ public void Serialize(IByteBuffer byteBuffer, StorageRangeMessage message)
 
         public StorageRangeMessage Deserialize(IByteBuffer byteBuffer)
         {
-            NettyBufferMemoryOwner memoryOwner = new(byteBuffer);
+            NettyBufferMemoryOwner? memoryOwner = new(byteBuffer);
             Rlp.ValueDecoderContext ctx = new(memoryOwner.Memory, true);
             int startPos = ctx.Position;
 
             ctx.ReadSequenceLength();
```
svlachakis: should we put this inside try? memoryOwner & message are both leaked otherwise I think

src/Nethermind/Nethermind.Core.Test/RlpTests.cs
```diff
@@ -569,5 +569,20 @@ private static byte[] BuildLongFormRlp(int prefix, int contentLength)
             return data;
         }
 
+        private sealed class DisposableTracker(Action onDispose) : IDisposable
```
svlachakis: Is this used anywhere?

flcl42: Removed

PR #10922, opened 2026-03-23
fix(flat): periodically clear ReadOnlySnapshotBundle cache

asdacap:
## Summary
- Add a 15-second periodic timer in `FlatDbManager` to clear the `ReadOnlySnapshotBundle` cache
- Prevents stale cache entries from holding `RefCountingPersistenceReader` leases indefinitely when persistence is slow/stalled, which blocks database compaction.
- Fix unexpected long snapshot log.

## Test plan
- [X] Verify build passes
- [X] Run node and confirm no stale snapshot warnings after 60s

### DIFF
--- a/src/Nethermind/Nethermind.State.Flat/FlatDbManager.cs
+++ b/src/Nethermind/Nethermind.State.Flat/FlatDbManager.cs
@@ -371,8 +371,6 @@ public class FlatDbManager : IFlatDbManager, IAsyncDisposable
 
     private void ClearReadOnlyBundleCache()
     {
-        if (_readonlySnapshotBundleCache.IsEmpty) return;
-
         foreach (KeyValuePair<StateId, ReadOnlySnapshotBundle> entry in _readonlySnapshotBundleCache)
         {
             if (_readonlySnapshotBundleCache.TryRemove(entry.Key, out ReadOnlySnapshotBundle? bundle))


src/Nethermind/Nethermind.State.Flat/FlatDbManager.cs
LukaszRozmej: Because it is concurrent dictionary you could just iterate on it and call TryRemove, no need to copy

benaadams:
```csharp
        if (_readonlySnapshotBundleCache.IsEmpty) return;

        foreach (KeyValuePair<StateId, ReadOnlySnapshotBundle> entry in _readonlySnapshotBundleCache)
        {
            if (_readonlySnapshotBundleCache.TryRemove(entry.Key, out ReadOnlySnapshotBundle? bundle))
            {
                bundle.Dispose();
            }
        }
```

src/Nethermind/Nethermind.State.Flat/FlatDbManager.cs
```diff
@@ -351,14 +354,28 @@ public void AddSnapshot(Snapshot snapshot, TransientResource transientResource)
...
+        catch (OperationCanceledException)
+        {
+        }
+    }
+
     private void ClearReadOnlyBundleCache()
     {
-        using ArrayPoolListRef<StateId> statesToRemove = new();
-        statesToRemove.AddRange(_readonlySnapshotBundleCache.Keys);
+        if (_readonlySnapshotBundleCache.IsEmpty) return;
```
LukaszRozmej: This actually isn't an optimization as `IsEmpty` acquires all bucket locks. It is cheaper to just iterate!

```suggestion
```

PR #10925, opened 2026-03-23
refactor: Move merkleization metrics to IWorldStateScopeProvider

asdacap:
## Summary
- Replace `WorldStateMetricsDecorator` (IWorldState decorator) with `WorldStateMetricsScopeProvider` (IWorldStateScopeProvider decorator)
- Metrics now measure actual merkleization (`IScope.UpdateRootHash` + `IScope.Commit`) at the scope level instead of at WorldState level
- Decorator publishes metrics via callback, removing metric propagation from `BranchProcessor`
- Registered at the upper level in `MainProcessingContext` so the prewarmer is not affected
- Removes virtual call overhead — the old `WrappedWorldState` made every `IWorldState` method virtual, adding indirection on hot-path calls (`Get`, `AccountExists`, `GetBalance`, etc.) that don't need interception
- Removed `protected _stateProvider` field from `BlockProcessor`; subclasses store their own

## Changes
- **New:** `WorldStateMetricsScopeProvider` with `Action<double>` callback for metric publishing
- **Deleted:** `WorldStateMetricsDecorator`, `IStateMerkleizationMetrics` interface
- **Modified:** `BlockProcessor`, `BranchProcessor`, `MainProcessingContext`, `AuRaBlockProcessor`, `OptimismBlockProcessor`

## Type of change

- [ ] Bug fix
- [x] Refactoring (no functional changes, no API changes)
- [ ] New feature
- [ ] Breaking change

## Testing

- [x] Nethermind.Blockchain.Test — 18/18 passed
- [x] Nethermind.Consensus.Test — 27/27 passed
- [x] Full solution build — 0 warnings, 0 errors

### DIFF
--- a/src/Nethermind/Nethermind.State/WorldStateMetricsScopeProvider.cs
+++ b/src/Nethermind/Nethermind.State/WorldStateMetricsScopeProvider.cs
@@ -6,7 +6,6 @@ using System.Diagnostics;
 using Nethermind.Core;
 using Nethermind.Core.Crypto;
 using Nethermind.Evm.State;
-using Nethermind.Int256;
 
 namespace Nethermind.State;
 
@@ -35,13 +34,7 @@ public class WorldStateMetricsScopeProvider : IWorldStateScopeProvider
 
         public Hash256 RootHash => baseScope.RootHash;
 
-        public void UpdateRootHash()
-        {
-            long start = Stopwatch.GetTimestamp();
-            baseScope.UpdateRootHash();
-            parent._stateMerkleizationTime += Stopwatch.GetElapsedTime(start).TotalMilliseconds;
-            parent._updateMetrics(parent._stateMerkleizationTime);
-        }
+        public void UpdateRootHash() => baseScope.UpdateRootHash();
 
         public Account? Get(Address address) => baseScope.Get(address);
 


src/Nethermind/Nethermind.Consensus/Processing/BranchProcessor.cs
```diff
@@ -146,7 +146,7 @@ public Block[] Process(BlockHeader? baseBlock, IReadOnlyList<Block> suggestedBlo
 
                 if (notReadOnly)
                 {
-                    Metrics.StateMerkleizationTime = _stateProvider.StateMerkleizationTime;
+                    Metrics.StateMerkleizationTime = _merkleizationMetrics.StateMerkleizationTime;
```
LukaszRozmej: If you either pass:
- a delegate
- a ref to the Metrics field to the WorldStateMetricsScopeProvider

Then you can remove this call, this field, NullStateMerkleizationMetrics, IStateMerkleizationMetrics altogether

asdacap: This should be moved out actually. Claude missed this.

asdacap: [Claude] Done — went with the delegate approach. `WorldStateMetricsScopeProvider` now takes an `Action<double>` callback, which is wired to `Blockchain.Metrics.StateMerkleizationTime` in `MainProcessingContext`. The `IStateMerkleizationMetrics` interface, `NullStateMerkleizationMetrics`, and the field in `BranchProcessor` are all removed.

LukaszRozmej: not pushed?

asdacap: Something weird with github. Its still processing updates.

LukaszRozmej: I think you need to push it again

asdacap: Forced pushed after rebase.

src/Nethermind/Nethermind.Consensus/Processing/BranchProcessor.cs
```diff
@@ -146,7 +144,6 @@ public Block[] Process(BlockHeader? baseBlock, IReadOnlyList<Block> suggestedBlo
 
                 if (notReadOnly)
                 {
-                    Metrics.StateMerkleizationTime = _stateProvider.StateMerkleizationTime;
```
asdacap, line 149: Where is this replaced at?

asdacap: [Claude] It's replaced by the `Action<double>` callback in `WorldStateMetricsScopeProvider` — the `UpdateRootHash()` and `Commit()` methods accumulate the time and invoke the callback. It's wired to `Blockchain.Metrics.StateMerkleizationTime` in `MainProcessingContext.cs:40`. The callback is needed because `Nethermind.State` can't reference `Nethermind.Blockchain` (circular dep). Added a doc comment to make this clearer.

src/Nethermind/Nethermind.Init/Modules/MainProcessingContext.cs
```diff
@@ -37,6 +37,8 @@ public MainProcessingContext(
             worldState = new WorldStateScopeOperationLogger(worldStateManager.GlobalWorldState, logManager);
         }
 
+        worldState = new WorldStateMetricsScopeProvider(worldState, time => Blockchain.Metrics.StateMerkleizationTime = time);
```
LukaszRozmej:
```suggestion
        worldState = new WorldStateMetricsScopeProvider(worldState, static time => Blockchain.Metrics.StateMerkleizationTime = time);
```

src/Nethermind/Nethermind.State/WorldStateMetricsScopeProvider.cs
```diff
@@ -0,0 +1,64 @@
...
+namespace Nethermind.State;
+
+public class WorldStateMetricsScopeProvider : IWorldStateScopeProvider
+{
+    private readonly IWorldStateScopeProvider _baseProvider;
+    private readonly Action<double>? _onMetricsUpdated;
+    private double _stateMerkleizationTime;
+
+    public WorldStateMetricsScopeProvider(IWorldStateScopeProvider baseProvider, Action<double>? onMetricsUpdated = null)
+    {
```
LukaszRozmej, line 19: rename `onMetricsUpdated` to `updateMetrics` and make it non-optional?

PR #10928, opened 2026-03-23
fix: process blocks with null BlockHeader in block test runner

### LINKED ISSUE #10678: Nethermind block test behavior differs from geth/reth/besu
# Issue Description
Nethermind skips a rejected block if a block header is null, while Geth, Reth and Besu still execute the block and attempt insertion. Block header is typically null for ``expectException`` blocks, which tests if a given invalid block is to be rejected. This results in certain block tests passing since nethermind skips invalid blocks. 
Additionally, this invokes a scenario where you cannot test whether invalid blocks are to be rejected..

# PoC
The PoC JSON of the block test exceeds the character limit of GitHub issues, so it is present in a separate gist here:

https://gist.github.com/R9295/40e5850c459acbffa3e82e2f41beaabd

LukaszRozmej:
## Summary

Fixes #10678

Nethermind's block test runner was silently skipping blocks with `blockHeader: null` in `DecodeRlps`, which is the standard format for `expectException` blocks. This caused tests to incorrectly pass where geth, reth, and besu correctly fail — Nethermind never actually validated that invalid blocks were rejected.

Three changes to `BlockchainTestBase`:

- **`DecodeRlps`**: Moved `correctRlp.Add(...)` outside the `blockHeader is not null` check so blocks are always queued for processing. Hash/uncle assertions remain gated on a non-null header.
- **`SuggestBlocks`**: Removed the premature `Assert.That(!expectsException, ...)` that required `expectException` blocks to fail at header validation. Some invalid blocks (e.g. intrinsic gas too low, nonce mismatch) pass header validation but fail during block processing.
- **`RunTest`**: After `StopAsync` completes async processing, verifies each `expectException` block was actually rejected from the canonical chain via `IsMainChain`. Produces geth-compatible error messages: `block (index N) insertion should have failed due to: <expectedException>`.

### Before
```
nethtest --blockTest --input poc.json
ConvertedTest001    PASS    ← incorrect, blocks silently skipped
```

### After
```
nethtest --blockTest --input poc.json
block (index 0) insertion should have failed due to: TransactionException.INTRINSIC_GAS_TOO_LOW
```
Matches geth's output for the same test.

## Changes

## Test plan

- Verified with the [PoC from the issue](https://gist.github.com/R9295/40e5850c459acbffa3e82e2f41beaabd) — now produces the same failure as geth
- Existing block tests should be unaffected: valid `expectException` blocks that are properly rejected will pass as before; only tests where an expected-to-fail block is actually accepted will now correctly fail

### DIFF
--- a/.agents/rules/test-infrastructure.md
+++ b/.agents/rules/test-infrastructure.md
@@ -71,3 +71,8 @@ The rule: **if production modules already wire a component, use them — don't c
 - Add tests to existing test files rather than creating new ones
 - **Do not duplicate test methods that differ only in parameters** — use `[TestCase(...)]` or `[TestCaseSource(...)]` to parameterize a single method
 - Before writing a new test, check if an existing test can be extended with another `[TestCase]` or use `[TestCaseSource]`
+
+## DotNetty `IByteBuffer` in tests
+
+- Prefer `using DisposableByteBuffer` via `.AsDisposable()` for releasing `IByteBuffer` in tests
+- For leak-detection tests, use `PooledBufferLeakDetector` from `Nethermind.Network.Test`
--- a/src/Nethermind/Chains/gnosis.json
+++ b/src/Nethermind/Chains/gnosis.json
@@ -103,6 +103,14 @@
     "eip7623TransitionTimestamp": "0x68122dbc",
     "eip7702TransitionTimestamp": "0x68122dbc",
     "eip4844FeeCollectorTransitionTimestamp": "0x68122dbc",
+    "eip7594TransitionTimestamp": "0x69de2dbc",
+    "eip7823TransitionTimestamp": "0x69de2dbc",
+    "eip7825TransitionTimestamp": "0x69de2dbc",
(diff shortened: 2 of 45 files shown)


LukaszRozmej:
The core fix landed independently on master via #11285 ("Additional BAL hive tests", merged April 2026), with follow-ups in #11399/#12177. Current master's BlockchainTestBase already:                                                          
  - enrolls null-header blocks in DecodeRlps (BlockchainTestBase.cs:550-565, with a comment explaining the EEST invalid-block fixture case — exactly what your PR's core change did),                                                              
  - no longer demands failure at pre-validation in SuggestBlocks (rejection may come from the processor or the end-of-test LastBlockHash/state-root check),                                                                                        
  - wires TestMergeModule + DifficultyOnlySealValidator so post-merge header rules (difficulty=0, empty uncles) are actually enforced — which solves in the test harness what your PR attempted with the MaximumUncleCount = 0 change to           
  ChainSpecBasedSpecProvider/Paris.

PR #10930, opened 2026-03-23
feat(rpc): add eth_getStorageValues

MysticRyuujin:
## Changes

- Implements `eth_getStorageValues` per the [execution-apis spec (PR #756)](https://github.com/ethereum/execution-apis/pull/756)
- Adds `ReadAsPropertyName` to `AddressConverter` to support `Dictionary<Address, ...>` deserialization

## Description

`eth_getStorageValues` allows fetching storage values for multiple slots across multiple accounts in a single RPC call, reducing round-trip overhead for local simulations and tooling.

**Request:** `{ "0xaddr": ["0xslot", ...], ... }` + block parameter  
**Response:** `{ "0xaddr": ["0x...32bytes", ...], ... }` — values in same order as keys

**Behaviour:**
- Max 1024 total slots per request (exceeding returns `-32602`)
- Empty request returns `-32602 "empty request"`
- Missing accounts/slots return 32 zero bytes (consistent with `eth_getStorageAt`)
- Missing trie node returns `-32000` (consistent with `eth_getStorageAt`)

## Type of change

## Testing

- [x] Unit tests added covering: single slot retrieval, empty request error, slot limit error, unknown account returning zeros
- [ ] Tested against Hive

## Spec reference

https://github.com/ethereum/execution-apis/pull/756

### DIFF
--- a/src/Nethermind/Nethermind.Core/JsonConverters/AddressConverter.cs
+++ b/src/Nethermind/Nethermind.Core/JsonConverters/AddressConverter.cs
@@ -42,11 +42,6 @@ public class AddressConverter : JsonConverter<Address>
         writer.WritePropertyName(addressBytes);
     }
 
-    public override Address ReadAsPropertyName(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
-    {
-        byte[]? bytes = ByteArrayConverter.Convert(ref reader);
-        if (bytes is null)
-            throw new JsonException("Invalid address property name");
-        return new Address(bytes);
-    }
+    public override Address ReadAsPropertyName(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options) =>
+        new Address(ByteArrayConverter.Convert(ref reader) ?? throw new JsonException("Invalid address property name"));
 }
--- a/src/Nethermind/Nethermind.JsonRpc.Test/Modules/Eth/EthRpcModuleTests.cs
+++ b/src/Nethermind/Nethermind.JsonRpc.Test/Modules/Eth/EthRpcModuleTests.cs
@@ -460,40 +460,39 @@ public partial class EthRpcModuleTests
         Assert.That(serialized, Is.EqualTo(expected));
     }
 
-    [Test]
-    public async Task Eth_get_storage_values_single_slot()
+    private static IEnumerable<TestCaseData> EthGetStorageValuesCases()
     {
-        using Context ctx = await Context.Create();
-        string addr = TestItem.AddressA.Bytes.ToHexString(true);
-        string serialized = await ctx.Test.TestEthRpc("eth_getStorageValues", $"{{\"{addr}\":[\"0x1\"]}}", "latest");
-        Assert.That(serialized, Is.EqualTo($"{{\"jsonrpc\":\"2.0\",\"result\":{{\"{addr}\":[\"0x0000000000000000000000000000000000000000000000000000000000abcdef\"]}},\"id\":67}}"));
+        string addressA = TestItem.AddressA.Bytes.ToHexString(true);
+        string addressB = TestItem.AddressB.Bytes.ToHexString(true);
+        string tooManySlots = string.Join(",", Enumerable.Range(0, EthRpcModule.MaxGetStorageSlots + 1).Select(i => $"\"0x{i:x}\""));
+
+        yield return new TestCaseData(
+            $"{{\"{addressA}\":[\"0x1\"]}}",
+            $"{{\"jsonrpc\":\"2.0\",\"result\":{{\"{addressA}\":[\"0x0000000000000000000000000000000000000000000000000000000000abcdef\"]}},\"id\":67}}")
+            .SetName("Eth_get_storage_values_single_slot");
+
+        yield return new TestCaseData(
+            "{}",
+            "{\"jsonrpc\":\"2.0\",\"error\":{\"code\":-32602,\"message\":\"empty request\"},\"id\":67}")
+            .SetName("Eth_get_storage_values_empty_request_returns_error");
+
+        yield return new TestCaseData(
+            $"{{\"{addressA}\":[{tooManySlots}]}}",
+            $"{{\"jsonrpc\":\"2.0\",\"error\":{{\"code\":-32602,\"message\":\"too many slots (max {EthRpcModule.MaxGetStorageSlots})\"}},\"id\":67}}")
+            .SetName("Eth_get_storage_values_too_many_slots_returns_error");
+
+        yield return new TestCaseData(
+            $"{{\"{addressB}\":[\"0x0\"]}}",
+            $"{{\"jsonrpc\":\"2.0\",\"result\":{{\"{addressB}\":[\"0x0000000000000000000000000000000000000000000000000000000000000000\"]}},\"id\":67}}")
+            .SetName("Eth_get_storage_values_unknown_account_returns_zeros");
     }
 
-    [Test]
-    public async Task Eth_get_storage_values_empty_request_returns_error()
+    [TestCaseSource(nameof(EthGetStorageValuesCases))]
+    public async Task Eth_get_storage_values(string request, string expected)
     {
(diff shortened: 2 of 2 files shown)


src/Nethermind/Nethermind.Core/JsonConverters/AddressConverter.cs
```diff
@@ -41,4 +41,12 @@ public override void WriteAsPropertyName(Utf8JsonWriter writer,
...
         writer.WritePropertyName(addressBytes);
     }
+
+    public override Address ReadAsPropertyName(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+    {
+        byte[]? bytes = ByteArrayConverter.Convert(ref reader);
+        if (bytes is null)
+            throw new JsonException("Invalid address property name");
+        return new Address(bytes);
+    }
```
LukaszRozmej: could be shortened:

```suggestion
    public override Address ReadAsPropertyName(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options) =>
        new Address(ByteArrayConverter.Convert(ref reader) ?? throw new JsonException("Invalid address property name"));
```

src/Nethermind/Nethermind.JsonRpc.Test/Modules/Eth/EthRpcModuleTests.cs
```diff
@@ -460,6 +460,42 @@ public async Task Eth_get_storage_at_missing_trie_node()
...
+    }
+
+    [Test]
+    public async Task Eth_get_storage_values_unknown_account_returns_zeros()
+    {
+        using Context ctx = await Context.Create();
+        string addr = TestItem.AddressB.Bytes.ToHexString(true);
+        string serialized = await ctx.Test.TestEthRpc("eth_getStorageValues", $"{{\"{addr}\":[\"0x0\"]}}", "latest");
+        Assert.That(serialized, Is.EqualTo($"{{\"jsonrpc\":\"2.0\",\"result\":{{\"{addr}\":[\"0x0000000000000000000000000000000000000000000000000000000000000000\"]}},\"id\":67}}"));
+    }
```
LukaszRozmej: Might be debatable, but isn't it better with test cases?

```suggestion
    private static IEnumerable<TestCaseData> EthGetStorageValuesCases()
    {
        string addressA = TestItem.AddressA.Bytes.ToHexString(true);
        string addressB = TestItem.AddressB.Bytes.ToHexString(true);
        string tooManySlots = string.Join(",", Enumerable.Range(0, EthRpcModule.MaxGetStorageSlots + 1).Select(i => $"\"0x{i:x}\""));

        yield return new TestCaseData(
            $"{{\"{addressA}\":[\"0x1\"]}}",
            $"{{\"jsonrpc\":\"2.0\",\"result\":{{\"{addressA}\":[\"0x0000000000000000000000000000000000000000000000000000000000abcdef\"]}},\"id\":67}}")
            .SetName("Eth_get_storage_values_single_slot");

        yield return new TestCaseData(
            "{}",
            "{\"jsonrpc\":\"2.0\",\"error\":{\"code\":-32602,\"message\":\"empty request\"},\"id\":67}")
            .SetName("Eth_get_storage_values_empty_request_returns_error");

        yield return new TestCaseData(
            $"{{\"{addressA}\":[{tooManySlots}]}}",
            $"{{\"jsonrpc\":\"2.0\",\"error\":{{\"code\":-32602,\"message\":\"too many slots (max {EthRpcModule.MaxGetStorageSlots})\"}},\"id\":67}}")
            .SetName("Eth_get_storage_values_too_many_slots_returns_error");

        yield return new TestCaseData(
            $"{{\"{addressB}\":[\"0x0\"]}}",
            $"{{\"jsonrpc\":\"2.0\",\"result\":{{\"{addressB}\":[\"0x0000000000000000000000000000000000000000000000000000000000000000\"]}},\"id\":67}}")
            .SetName("Eth_get_storage_values_unknown_account_returns_zeros");
    }

    [TestCaseSource(nameof(EthGetStorageValuesCases))]
    public async Task Eth_get_storage_values(string request, string expected)
    {
        using Context ctx = await Context.Create();
        string serialized = await ctx.Test.TestEthRpc("eth_getStorageValues", request, "latest");
        Assert.That(serialized, Is.EqualTo(expected));
    }
```

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
```diff
@@ -200,6 +201,53 @@ public ResultWrapper<byte[]> eth_getStorageAt(Address address, UInt256 positionI
...
+        }
+
+        SearchResult<BlockHeader> searchResult = _blockFinder.SearchForHeader(blockParameter);
+        if (searchResult.IsError)
+        {
+            return GetFailureResult<Dictionary<Address, byte[][]>, BlockHeader>(searchResult, _ethSyncingInfo.SyncMode.HaveNotSyncedHeadersYet());
+        }
+
+        BlockHeader header = searchResult.Object!;
+        Dictionary<Address, byte[][]> result = new(requests.Count);
```
LukaszRozmej, line 228:
```suggestion
        if (!_blockchainBridge.HasStateForBlock(header!))
        {
            return Task.FromResult(GetStateFailureResult<UInt256?>(header));
        }
                
        Dictionary<Address, byte[][]> result = new(requests.Count);
```

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
```diff
@@ -200,6 +201,53 @@ public ResultWrapper<byte[]> eth_getStorageAt(Address address, UInt256 positionI
...
+                    vals[i] = storage.IsEmpty ? Bytes32.Zero.Unwrap() : storage.PadLeft(32);
+                }
+                result[addr] = vals;
+            }
+        }
+        catch (MissingTrieNodeException e)
+        {
+            Hash256 hash = e.Hash;
+            return ResultWrapper<Dictionary<Address, byte[][]>>.Fail($"missing trie node {hash} (path ) state {hash} is not available", ErrorCodes.InvalidInput);
+        }
```
LukaszRozmej, line 246: With proper check, this is redundant (state would have to be corrupted to throw)

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
```diff
@@ -200,6 +201,53 @@ public ResultWrapper<byte[]> eth_getStorageAt(Address address, UInt256 positionI
         }
     }
 
+    public ResultWrapper<Dictionary<Address, byte[][]>> eth_getStorageValues(Dictionary<Address, UInt256[]> requests, BlockParameter? blockParameter = null)
```
LukaszRozmej, line 204: One potential downside here is that both the input Dictionary and the output Dictionary are allocation heavy.
While the limit of 1024 storage slots does make some difference, it is only applied after the initial Dictionary parameter is allocated and parsed. So it is more of unbound in practice.

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
MysticRyuujin: This check doesn't exist in `eth_getStorageAt`, should it?

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
MysticRyuujin: Not entirely sure what to do about this though, is it a blocker?

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
MysticRyuujin: Hmmm I was mostly just following the same code as `eth_getStorageAt`

PR #10944, opened 2026-03-25
fix(precompiles): add missing metrics increment for L1SloadPrecompile

eeemmmmmm:
L1SloadPrecompile.Run() was missing the Metrics.L1SloadPrecompile++ call that all other precompiles have. The metric was defined in Metrics.cs but never incremented, so L1SLOAD calls weren't being tracked.

### DIFF
--- a/src/Nethermind/Nethermind.Evm.Precompiles/L1SloadPrecompile.cs
+++ b/src/Nethermind/Nethermind.Evm.Precompiles/L1SloadPrecompile.cs
@@ -43,6 +43,7 @@ public class L1SloadPrecompile : IPrecompile<L1SloadPrecompile>
 
     public Result<byte[]> Run(ReadOnlyMemory<byte> inputData, IReleaseSpec releaseSpec)
     {
+        Metrics.L1SloadPrecompile++;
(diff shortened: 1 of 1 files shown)


LukaszRozmej:
Thanks for catching this @eeemmmmmm! Unfortunately this PR has been overtaken by master — the L1 precompiles were relocated from `Nethermind.Evm.Precompiles` to the Taiko plugin in [`fb150e58e3`](https://github.com/NethermindEth/nethermind/commit/fb150e58e3) ("Move L1 precompiles to Taiko plugin"). The new file at `src/Nethermind/Nethermind.Taiko/Precompiles/L1SloadPrecompile.cs` already increments the metric at the top of `Run(...)`:

```csharp
L1PrecompileMetrics.L1SloadPrecompile++;
```

The counter itself now lives on `L1PrecompileMetrics` (`src/Nethermind/Nethermind.Taiko/Precompiles/L1PrecompileMetrics.cs`) rather than the old `Metrics` class, so the fix you proposed is effectively already in master. Closing as resolved — appreciate the contribution!

PR #10948, opened 2026-03-25
perf: add cross-block storage and state caches with write-through

kamilchodola:
## Changes

Add a dedicated `CrossBlockCaches` class with its own `SeqlockCache` instances (~4.5 MB) for storage slots and accounts that persist across blocks. Writes are buffered during commit and promoted only after successful `CommitTree`.

### Review issues addressed

**[Correctness] Speculative entries from failed blocks** — writes are buffered on the `ScopeWrapper` during the write batch phase and only promoted to the cross-block caches in `Commit()`, which is called from `CommitTree` after successful block processing. If `ProcessOne` throws, `PreCommitBlock`/`CommitTree` is never reached, so the buffer is silently discarded.

**[Correctness] Stale StorageRoot from storage-only changes** — `CacheUpdatingWriteBatch` subscribes to `baseBatch.OnAccountUpdated` and buffers the final account (with updated `StorageRoot`) emitted by the underlying batch (e.g. `TrieStoreScopeProvider`). The promotion step writes the fully-resolved account to the cross-block state cache.

**[Performance] Full cache nuke on CREATE** — `CacheUpdatingStorageWriteBatch.Clear()` no longer epoch-bumps immediately. Instead it sets a `_pendingStorageClear` flag. At promotion time, if the flag is set, the storage cache is epoch-bumped first, then all committed storage values from the buffer are re-seeded. This means the cache ends up with only correct post-block values; previous cross-block entries for unrelated addresses are lost only when a Clear actually occurred, and are immediately re-seeded with the current block's committed values.

### Architecture

- **Separate from prewarmer caches** — cross-block caches are never shared with prewarmer threads, no concurrent `GetOrAdd` race
- **Per-block caches unchanged** — `PreBlockCaches` cleared between blocks exactly as on master
- **Read path** — main thread checks per-block cache first, then cross-block cache, then trie/db
- **Write path** — commit write batch buffers to `ScopeWrapper`, promoted in `Commit()` after successful `CommitTree`

### Files changed

- **`CrossBlockCaches.cs`** (new) — two `SeqlockCache` instances for storage and state
- **`PrewarmerScopeProvider.cs`** — buffered write-through with deferred promotion, `OnAccountUpdated` interception, selective `Clear()` handling
- **`PrewarmerModule.cs`** — register `CrossBlockCaches` as singleton, pass to decorator

## Types of changes

- [x] Performance improvement

## Testing

- [x] Nethermind.State.Test: StorageProviderTests + ScopeProviderTests (91 passed, 1 skipped)
- [ ] Benchmark with `realblocks` payload set
- [ ] Benchmark with `superblocks` payload set

### DIFF
--- a/src/Nethermind/Nethermind.Core/Collections/SeqlockCache.cs
+++ b/src/Nethermind/Nethermind.Core/Collections/SeqlockCache.cs
@@ -443,10 +443,10 @@ public struct DefaultCacheSets : ICacheSets
     public static int SetsLog2 => 14;
 }
 
-/// <summary>128K sets × 2 ways = 256K entries (~16 MB). For cross-block caches.</summary>
+/// <summary>64K sets × 2 ways = 128K entries (~8 MB). For cross-block caches.</summary>
 public struct LargeCacheSets : ICacheSets
 {
-    public static int SetsLog2 => 17;
+    public static int SetsLog2 => 16;
 }
 
 /// <summary>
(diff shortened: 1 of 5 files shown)


kamilchodola:
Closing as not needed

PR #10246, opened 2026-01-15
Add and activate eth/70

flcl42:
## Changes

- Add eth/70

#### What types of changes does your code introduce?

- [ ] Bugfix (a non-breaking change that fixes an issue)
- [x] New feature (a non-breaking change that adds functionality)
- [ ] Breaking change (a change that causes existing functionality not to work as expected)
- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

#### If yes, did you write tests?

- [x] Yes
- [ ] No

PR #10951, opened 2026-03-25
fix: resource leak — ArrayPool, double-dispose, CTS, event handlers & Resource Leak audit skill

newmanifold:
## Summary

Fixes resource management issues across the codebase. Five fixes address real correctness issues (data corruption, permanent leaks, exception on race condition). The remaining fixes are best-practice cleanup where the GC, timer expiry, or finalizer would eventually recover the resources.

### Real issues

- **SlotsAndProofs** — plain `bool _disposed` allowed concurrent `Dispose()` calls to return the same ArrayPool buffers twice, corrupting the shared pool. Replaced with `Interlocked.Exchange`.
- **BatchedTrieVisitor** — `GetNextBatch()` returns an `ArrayPoolList` backed by a custom `_trieNodePool`. On early return (missing node) or exception, the list was never disposed and the pooled arrays were permanently lost — custom pools don't recover via GC. Wrapped in `using (currentBatch)`.
- **ProtocolHandlerBase SetResult→TrySetResult** — the `TrySetCanceled()` added on the timeout path races with `SetResult()` on the late-message path. If timeout fires first, a late init message throws `InvalidOperationException`. Changed to `TrySetResult` so the late message is a safe no-op. Verified RED→GREEN with integration test.
- **EthereumIesEngine** — three `ArrayPool.Rent()` calls in ECIES encrypt/decrypt with `Return()` only at the end. `ThrowInvalidMac()` skips all returns. Attacker-triggerable during RLPx handshake (no authentication required). Replaced with `using ArrayPoolSpan<byte>`.
- **Eth70ProtocolHandler GetReceiptsMessage70** — the eth/70 `Handle(GetReceiptsMessage70)` handler did not dispose the incoming request after fulfilling it. Every other handler in Eth66, Eth69, and Snap uses `using var message = request;`. The Eth70 handler was the sole outlier, leaking an `ArrayPoolList<Hash256>` on every incoming `GetReceipts` request from any eth/70 peer. Also documented the ownership contract on `HandleInBackground`.

### Best practice (GC/timer/finalizer recovers eventually)

- **NewPayloadHandler, ProtocolHandlerBase** — `CancellationTokenSource` backing `Task.Delay` timeouts cancelled but never disposed. The internal timer lingers in `TimerQueue` until the original timeout fires (3-10s), then becomes GC-collectible. `using` disposes immediately instead of waiting.
- **BlockCachePreWarmer.AddressWarmer** — `ManualResetEventSlim` kernel `WaitHandle` released by `SafeHandle` finalizer at Gen2 GC. Added `IDisposable` and manual `Dispose()` in the existing `finally` block for deterministic cleanup.
- **MergeFinalizationManager / AuRaMergeFinalizationManager** — event handlers on singleton-to-singleton subscriptions. Objects are never GC'd regardless. Unsubscribe in `Dispose()` for correct contract.
- **ProtocolHandlerBase TrySetCanceled** — completes dangling `_initCompletionSource` on timeout. Continuations would time out via `Task.Delay` anyway.
- **NettyHandshakeHandler SetResult→TrySetResult** — defensive consistency with ProtocolHandlerBase pattern. Currently not a race (only one writer), but prevents future issues if `TrySetCanceled` is added on the timeout path.
- **ShutterBlockHandler** — `BlockWaitTask.Dispose()` instead of manually disposing only registrations. Leaked `TimeoutSource` CTS cleaned by GC. Only relevant on Shutter/Gnosis chains.
- **SyncDispatcher** — `StateChanged` handler unsubscribed in `DisposeAsync()`. Dispatcher and feed have same lifetime; the reference wouldn't prevent GC.

### Agentic skill

- Add `resource-leak-audit` skill (`.agents/skills/`) with symlink from `.claude/skills/` — two-phase audit methodology covering Tier 1–4 leak patterns. Supports full-codebase and PR-scoped mode. Running codebase-wide is very token-intensive; recommended to scope to a specific project or PR.

### 1. EthereumIesEngine — ArrayPool buffers leaked on invalid MAC

In `DecryptBlock()`, three buffers are rented from `ArrayPool<byte>.Shared` but returned only at the end of the method. If `ThrowInvalidMac()` throws — triggered by any peer sending a bad MAC — execution skips the `Return()` calls. Triggerable by unauthenticated peers during RLPx handshake. `EncryptBlock()` has the same pattern with `ThrowOutputBufferTooSmall()`.

### 2. SlotsAndProofs — non-atomic dispose guard causes double ArrayPool return

`Dispose()` used a plain `bool _disposed` guard. During snap sync, concurrent callers can race to dispose the same instance via `SnapSyncBatch.Dispose()` from multiple threads. Two threads read `_disposed == false` simultaneously, both proceed to `DisposeRecursive()`, returning the same ArrayPool buffers twice — corrupting the shared pool's internal bookkeeping.

### 3. ProtocolHandlerBase — CTS timer handle and TCS race

`CancellationTokenSource` backing `Task.Delay` cancelled but never disposed — timer lingers in `TimerQueue` for up to the timeout duration. Separately, `TrySetCanceled()` on the timeout path races with `SetResult()` on the late-message path: if timeout fires first, a late init message throws `InvalidOperationException`.

### 4. NewPayloadHandler / NettyHandshakeHandler — CTS timer handles

Same `Task.Delay` pattern: CTS cancelled but not disposed. Timer handle lingers for up to the timeout period before becoming GC-collectible.

### 5. MergeFinalizationManager / AuRaMergeFinalizationManager — empty Dispose

`Dispose()` was `{ }`. Constructor subscribes to events but never unsubscribes. On post-merge chains (all current Ethereum chains), the `TerminalBlockReached` handler is dead code that re-sets an already-true flag. The fix is correct contract hygiene.

### 6. ShutterBlockHandler — partial cleanup

`Dispose()` manually disposed two `CancellationRegistration` fields but skipped calling `BlockWaitTask.Dispose()` which also cleans up `TimeoutSource` CTS.

### 7. BatchedTrieVisitor — ArrayPoolList leaked on early return

`GetNextBatch()` returns a new `ArrayPoolList` backed by custom `_trieNodePool`. Disposed only on normal exit via explicit call. On early return (missing node) or exception, pooled arrays permanently lost — custom pool has no GC recovery.

### 8. BlockCachePreWarmer.AddressWarmer — ManualResetEventSlim not disposed

`ManualResetEventSlim` created per block. Kernel `WaitHandle` (when allocated by blocking `Wait()`) released by `SafeHandle` finalizer eventually. Added `IDisposable` and manual `Dispose()` in the existing `finally` block for deterministic release after `Wait()` returns.

### 9. SyncDispatcher — event handler after disposal

`StateChanged` handler subscribed in constructor, never unsubscribed in `DisposeAsync()`. Dispatcher and feed have same lifetime — the reference wouldn't prevent GC in practice.

### 10. Eth70ProtocolHandler — GetReceiptsMessage70 request not disposed

The eth/70 `Handle(GetReceiptsMessage70)` handler receives ownership of the deserialized request via `HandleInBackground` but never disposes it. The request contains an `IOwnedReadOnlyList<Hash256>` (block hashes from the peer) backed by `ArrayPoolList` — leaked on every incoming `GetReceipts` request. All other handlers in Eth66/Eth69/Snap use `using var message = request;`. This was the sole outlier, introduced in PR #10246.

## Types of changes

- [x] Other: agentic skill for resource leak auditing

## Testing

- `SlotsAndProofsTests` — concurrent and sequential disposal, null-field safety
- `MergeFinalizationManagerTests` — event unsubscription verified with sentinel value, double-dispose, BlocksFinalized forwarding stops after disposal
- `ProtocolHandlerBaseTests` — integration test: 50ms timeout then late message, verified RED→GREEN (SetResult throws, TrySetResult doesn't)
- `SyncDispatcherTests` — StateChanged unsubscription, double-dispose, Activate+Finish after disposal (extracted shared helper per review feedback)
- `TimeoutUtilsTests` — `TimeoutOn` extension method with CTS disposal
- `Eth70ProtocolHandlerTests` — `Handle_GetReceipts_disposes_request_message`: passes trackable subclass, asserts `IsDisposed` is true after handler returns. Verified RED→GREEN.
- Existing test suites pass

## Documentation

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Network/Rlpx/NettyHandshakeHandler.cs
+++ b/src/Nethermind/Nethermind.Network/Rlpx/NettyHandshakeHandler.cs
@@ -210,6 +210,7 @@ namespace Nethermind.Network.Rlpx
                     Metrics.HandshakeTimeouts++;
                     if (_logger.IsTrace) _logger.Trace($"Disconnecting due to timeout for handshake: {_session.RemoteNodeId}@{_session.RemoteHost}:{_session.RemotePort}");
                     //It will trigger channel.CloseCompletion which will trigger DisconnectAsync on the session
+                    _initCompletionSource.TrySetCanceled();
                     await _channel.DisconnectAsync();
                 }
             }
--- a/src/Nethermind/Nethermind.Network/P2P/MessageDictionary.cs
+++ b/src/Nethermind/Nethermind.Network/P2P/MessageDictionary.cs
@@ -83,6 +83,7 @@ public class MessageDictionary<T66Msg, TData>(Action<T66Msg> send, TimeSpan? old
                     if (_requests.TryRemove(requestIdValues.Key, out Request<T66Msg, TData> request))
                     {
                         Interlocked.Decrement(ref _requestCount);
+                        request.Message.TryDispose();
                         // Unblock waiting thread.
                         request.CompletionSource.TrySetException(new TimeoutException("No response received"));
                     }
--- a/src/Nethermind/Nethermind.Shutter/ShutterBlockImprovementContext.cs
+++ b/src/Nethermind/Nethermind.Shutter/ShutterBlockImprovementContext.cs
@@ -144,12 +144,13 @@ public class ShutterBlockImprovementContext : IBlockImprovementContext
         if (_logger.IsDebug) _logger.Debug($"Awaiting Shutter decryption keys for {slot} at offset {offset}ms. Timeout in {waitTime}ms...");
 
         using var txTimeout = new CancellationTokenSource((int)waitTime);
-        _linkedCancellation = CancellationTokenSource.CreateLinkedTokenSource(_improvementToken, txTimeout.Token);
+        CancellationTokenSource linkedCancellation = CancellationTokenSource.CreateLinkedTokenSource(_improvementToken, txTimeout.Token);
+        _linkedCancellation = linkedCancellation;
 
         try
         {
-            _linkedCancellation.Token.ThrowIfCancellationRequested();
-            await _txSignal.WaitForTransactions(slot, _linkedCancellation.Token);
+            linkedCancellation.Token.ThrowIfCancellationRequested();
+            await _txSignal.WaitForTransactions(slot, linkedCancellation.Token);
         }
         catch (Exception ex) when (ex is OperationCanceledException or ObjectDisposedException)
         {


src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
```diff
@@ -91,25 +91,27 @@ public CacheType ClearCaches()
 
     private void PreWarmCachesParallel(BlockState blockState, Block suggestedBlock, BlockHeader parent, IReleaseSpec spec, ParallelOptions parallelOptions, AddressWarmer addressWarmer, CancellationToken cancellationToken)
     {
-        try
+        using (addressWarmer)
```
LukaszRozmej: nit: as we already have a `finally` clause I would just add `addressWarmer.Dispose()` there manually

src/Nethermind/Nethermind.Synchronization.Test/ParallelSync/SyncDispatcherTests.cs
```diff
@@ -320,6 +320,67 @@ public async Task When_ConcurrentHandleResponseIsRunning_Then_BlockDispose()
...
+    public async Task DisposeAsync_unsubscribes_StateChanged_handler()
+    {
+        TestSyncFeed syncFeed = new();
+        SyncDispatcher<TestBatch> dispatcher = new(
+            new TestSyncConfig(),
+            syncFeed,
+            new TestDownloader(),
+            new TestSyncPeerPool(),
+            new StaticPeerAllocationStrategyFactory<TestBatch>(FirstFree.Instance),
+            LimboLogs.Instance);
```
LukaszRozmej: Extract to helper method to avoid duplication?

src/Nethermind/Nethermind.Network/Rlpx/NettyHandshakeHandler.cs
LukaszRozmej: Should we:

```suggestion
            _initCompletionSource?.TrySetResult(input);
```

newmanifold: Currently i don't think we cancel this tcs so SetResult is the only thing that can complete it, but change it to use `TrySetResult` to be defensive

PR #10952, opened 2026-03-26
fix: TransientResource disposal leaks and missing pruning in PersistenceManager

asdacap:
## Summary
- Does not seems to do anything. But make design that dependes on fully pooled transient resource work.
- Return `TransientResource` to pool instead of leaking on early-return and queue-full paths in `FlatDbManager.AddSnapshot`
- Call `PrunePersistedRecursively(1)` after marking nodes persisted in `PersistenceManager` to release in-memory references to persisted children

## Test plan
- [x] `dotnet test --project src/Nethermind/Nethermind.State.Flat.Test/ -c release` — 285 passed, 0 failed


LukaszRozmej (approved):

LukaszRozmej:
@claude review

flcl42 (approved):

PR #10959, opened 2026-03-26
Cross-block cache carry-forward with reorg invalidation

benaadams:
## Changes

- Add cache-populating write-batch wrappers in PrewarmerScopeProvider that mirror account and storage writes into PreBlockCaches during FlushToTree. This allows the SeqlockCaches to carry forward block N's post-state as block N+1's pre-state, reducing trie reads on sequential processing.

- PreBlockCaches tracks committed block number/hash metadata and a legacy storage-clear flag for pre-EIP-6780 SELFDESTRUCT handling. BranchProcessor validates cache continuity before the first prewarm and invalidates on parent mismatch, errors, or legacy storage clears.

- [x] Optimization

#### Requires testing

- [x] Yes

- [x] Yes

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
@@ -9,18 +9,17 @@ using Microsoft.Extensions.ObjectPool;
 using Nethermind.Blockchain;
 using Nethermind.Config;
 using Nethermind.Core;
+using Nethermind.Core.Collections;
+using Nethermind.Core.Eip2930;
+using Nethermind.Core.Extensions;
 using Nethermind.Core.Specs;
 using Nethermind.Core.Threading;
 using Nethermind.Evm;
+using Nethermind.Evm.State;
 using Nethermind.Evm.Tracing;
 using Nethermind.Evm.TransactionProcessing;
 using Nethermind.Int256;
 using Nethermind.Logging;
-using Nethermind.Evm.State;
-using Nethermind.Core.Eip2930;
-using Nethermind.Core.Collections;
-using Nethermind.Core.Extensions;
-using Nethermind.State;
 using Nethermind.Trie;
 
 namespace Nethermind.Consensus.Processing;
@@ -56,13 +55,9 @@ public sealed class BlockCachePreWarmer(
     {
         if (preBlockCaches is not null)
         {
-            CacheType result = preBlockCaches.ClearCaches();
+            preBlockCaches.ClearCaches();
             nodeStorageCache.ClearCaches();
             nodeStorageCache.Enabled = true;
-            if (result != default)
-            {
-                if (_logger.IsWarn) _logger.Warn($"Caches {result} are not empty. Clearing them.");
-            }
 
             if (parent is not null && _concurrencyLevel > 1 && !cancellationToken.IsCancellationRequested)
             {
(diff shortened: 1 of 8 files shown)


src/Nethermind/Nethermind.Consensus/Processing/BranchProcessor.cs
```diff
@@ -86,6 +86,7 @@ public Block[] Process(BlockHeader? baseBlock, IReadOnlyList<Block> suggestedBlo
         {
             // Start prewarming as early as possible
             WaitForCacheClear();
+            preWarmer?.EnsureCacheValidityForBlock(baseBlock);
```
asdacap: You know, you can check this within BeginScope, which mean you dont need to add this.

src/Nethermind/Nethermind.State/PrewarmerScopeProvider.cs
```diff
@@ -36,25 +36,36 @@ public class PrewarmerScopeProvider(
 {
     public bool HasRoot(BlockHeader? baseBlock) => baseProvider.HasRoot(baseBlock);
 
-    public IWorldStateScopeProvider.IScope BeginScope(BlockHeader? baseBlock) => new ScopeWrapper(baseProvider.BeginScope(baseBlock), preBlockCaches, populatePreBlockCache);
+    public IWorldStateScopeProvider.IScope BeginScope(BlockHeader? baseBlock)
+    {
+        if (baseBlock is null || !preBlockCaches.IsValidForParent(baseBlock.Number, baseBlock.Hash))
+        {
+            preBlockCaches.InvalidateCaches();
```
AnkushinDaniil, line 43: It seems InvalidateCaches() can run from a prewarm-env Build concurrently with the commit path - DiscardPendingCarryForward clears the non-thread-safe _pendingStateWrites list while FlushToTree is still adding to it, and if it lands before RecordCommittedBlock the cache can end up valid but missing post-state overlays. Maybe do the invalidation only from the main scope, or in BranchProcessor before prewarm starts?

src/Nethermind/Nethermind.State/PrewarmerScopeProvider.cs
```diff
@@ -36,25 +36,36 @@ public class PrewarmerScopeProvider(
...
+        }
+
+        preBlockCaches.ResetBlockFlags();
```
AnkushinDaniil, line 46: ResetBlockFlags() from a late prewarm scope can clear _hasLegacyStorageClear between NoteStorageClear and FinalizeProcessedBlock - then a pre-6780 selfdestruct block skips invalidation and the destroyed contract's storage stays cached as valid.

src/Nethermind/Nethermind.State/PrewarmerScopeProvider.cs
```diff
@@ -36,25 +36,36 @@ public class PrewarmerScopeProvider(
 {
     public bool HasRoot(BlockHeader? baseBlock) => baseProvider.HasRoot(baseBlock);
 
-    public IWorldStateScopeProvider.IScope BeginScope(BlockHeader? baseBlock) => new ScopeWrapper(baseProvider.BeginScope(baseBlock), preBlockCaches, populatePreBlockCache);
+    public IWorldStateScopeProvider.IScope BeginScope(BlockHeader? baseBlock)
+    {
+        if (baseBlock is null || !preBlockCaches.IsValidForParent(baseBlock.Number, baseBlock.Hash))
```
AnkushinDaniil, line 41: While committed == -1 (startup/reorg/error), every prewarm env.Build fails IsValidForParent and wipes both caches - so concurrent warm-up workers keep destroying each other's work for that block. I think invalidating once per block would be enough.

src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
```diff
@@ -80,13 +75,40 @@ public Task PreWarmCaches(Block suggestedBlock, BlockHeader? parent, IReleaseSpe
...
+    public void InvalidateCaches()
+    {
+        if (_logger.IsDebug) _logger.Debug("Invalidating caches");
+        preBlockCaches?.InvalidateCaches();
+        if (_logger.IsDebug) _logger.Debug("Invalidated caches");
+    }
+
+    public void FinalizeProcessedBlock(BlockHeader block, IReleaseSpec spec)
+    {
+        if (!spec.IsEip6780Enabled && preBlockCaches.HasLegacyStorageClear)
```
AnkushinDaniil, line 87: It seems HasClear is not selfdestruct-specific - it also fires on every successful CREATE on a dead account, so on pre-Cancun blocks the carry-forward gets invalidated on almost every block with a contract creation. Maybe skip NoteStorageClear when the cleared tree was already empty?

AnkushinDaniil (commented):
Btw the base is quite behind master (#12280 and #12406 reworked these files), so I think a rebase is needed first.

benaadams:
Already done

Issue #10966, opened 2026-03-26
Flaky: Nethermind extra test variants — recurring failures across [no-intrinsics] and [checked] jobs


newmanifold:
Recurring test failures across CI variants in the last 7 days.

**`Can_change_branch_on_invalid_block_when_invalid_branch_is_in_the_queue_and_recovery_queue_max_has_been_reached` — 4 failures in 7 days**
- https://github.com/NethermindEth/nethermind/actions/runs/23571563861
- https://github.com/NethermindEth/nethermind/actions/runs/23491534953
- https://github.com/NethermindEth/nethermind/actions/runs/23338739093

**`JsonRpc.Test` SIGSEGV (exit code 139) — 2 failures in 7 days**
All tests pass, then process crashes during teardown.
- https://github.com/NethermindEth/nethermind/actions/runs/23533485787
- https://github.com/NethermindEth/nethermind/actions/runs/23338739093

**`walletReorganizeOwners_Shanghai` Flat DB — 1 failure**
Block 162 rejected with `HeaderGasUsedMismatch` (computed 23932 gas, expected 127603). Possibly related to #10919.
- https://github.com/NethermindEth/nethermind/actions/runs/23533485787

Part of #10956

newmanifold:
Fixed! haven't seen walletReorganizeOwners_Shanghai failing on master again and can't reproduce locally

PR #10453, opened 2026-02-08
Optimize JSON-RPC request parsing and processing

benaadams:
Closes #10173 
Closes #10159 

## Changes

Performance optimizations for JSON-RPC request handling, focusing on Engine API hot paths, eliminating allocations, and enabling devirtualization
   in critical sections.

  **Converters**
  - SIMD-accelerated hex encoding via `HexWriter` for `Hash256Converter`, `ByteArrayConverter`, `UInt256Converter`
  - Direct numeric-to-hex for `LongConverter`, `ULongConverter`, `NullableULongConverter`
  - `WriteHashHex` takes `in ValueHash256` for type-safe 32-byte guarantee
  - Ternary simplification in `ReadCore` paths

  **Engine API**
  - Terminal middleware moved before `UseRouting()`/`UseCors()` to genuinely bypass routing, CORS, compression, and WebSocket middleware
  - Skip `CountingPipeReader` when `Content-Length` known
  - Eliminate `CancellationTokenSource` allocation
  - Bypass `BufferResponses`, call `StartAsync` before body
  - Stream blobs directly via `BlobsV1DirectResponse`/`BlobsV2DirectResponse` implementing `IStreamableResult`

  **DI & Devirtualization**
  - Register `JsonRpcProcessor`, `JsonRpcService`, `EthereumJsonSerializer` as concrete types through the full DI chain (`Dependencies` →
  `IBasicApi` → `Bootstrap` → `Startup`)
  - Eliminate interface→concrete casts in `Startup.Configure`
  - `EthereumJsonSerializer` exposed as concrete type from `IBasicApi` through `NethermindRunnerModule`

  **Response Writing**
  - `WriteJsonRpcResponse` with typed serialization for result/error payload, avoiding polymorphic dispatch through `JsonRpcResponse` base
  hierarchy
  - `WriteIdRaw` fast path for `int`/`long` IDs with `NoInlining` cold paths for `string`/`null`

  **Threading**
  - Replace `AsyncLocal<T>` with `ThreadStatic` for `JsonRpcContext` and `ForcedNumberConversion`
  - Pool `CancellationTokenSource` with `TryReset`
  - Seal hot classes, capture concrete types

  **Serialization**
  - System.Text.Json source generation for Engine/Eth/Debug/Trace APIs
  - Startup warmup for serializer metadata
  - Synchronous `JsonSerializer.Serialize` for `PipeWriter` path

  **JWT Authentication**
  - Manual HS256 validator with library fallback

  **Other**
[...]

PR #10967, opened 2026-03-27
fix: Flaky Nethermind.Blockchain.Test

newmanifold:
## Changes

- Add `suggestTask.Wait(100)` in `Suggested()` to give the `Task.Run` time to finish the queue write before the next `Suggested()` call

### Background

`Suggested()` originally called `SuggestBlock` directly without `Task.Run`. The queue write inside `Enqueue()` happened to complete before `Suggested()` returned, so consecutive calls preserved ordering in practice.

In #10453, `AllowSynchronousContinuations` was added to `_blockQueue`, which could inline `ProcessBlocks()` on the caller's thread and deadlock the test. #10513 worked around this by wrapping `SuggestBlock` in `Task.Run` and waiting on a `BlockAdded` event. This broke ordering — `BlockAdded` fires inside `Enqueue()` before the recovery queue write, so consecutive `Suggested()` calls could race their `Task.Run` threads to `TryWrite`, breaking recovery queue FIFO order.

This PR adds a bounded `suggestTask.Wait(100)` after `BlockAdded` resolves. The queue write takes microseconds, so 100ms is sufficient in practice, though not a hard guarantee.

### Root cause

`Can_change_branch_on_invalid_block_when_invalid_branch_is_in_the_queue_and_recovery_queue_max_has_been_reached` assumes blocks enter the recovery queue in the same order they are suggested. It calls `Recovered(block3)` only after `ProcessedFail(block2)` — so if block 3 enters the recovery queue before block 2, the recovery loop dequeues block 3 first, blocks waiting for its Allow token, and block 2 is stuck behind it. `ProcessedFail(block2)` times out after 10s.

`Can_change_branch_on_invalid_block_when_invalid_branch_is_in_the_queue` has the same consecutive `Suggested()` calls but pre-allows all blocks for recovery upfront, so it's resilient to reordering in practice.

The 100ms timeout avoids deadlock: when the block goes direct to `_blockQueue` (`_queueCount == 1`), `AllowSynchronousContinuations` can inline `ProcessBlocks()` on the `Task.Run` thread which blocks until `Allow()` is called. An infinite wait would deadlock since the test thread could never call `Allow()`. Single-block enqueues have no ordering concern, so the timeout expiring is harmless in that case.

### Stress test results

Without fix (24 isolated CI runners, full `Nethermind.Blockchain.Test [no-intrinsics]`):
- https://github.com/NethermindEth/nethermind/actions/runs/23627888263 — 2/24 failed (`Can_change_branch_on_invalid_block_when_invalid_branch_is_in_the_queue_and_recovery_queue_max_has_been_reached`)

With fix (30 isolated CI runners):
- https://github.com/NethermindEth/nethermind/actions/runs/23646894251 — 0/30 failed

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

Stress tested with 30 isolated CI runners running full `Nethermind.Blockchain.Test [no-intrinsics]`. Reproduced locally at ~25% failure rate without fix, 0% with fix (48 runs).

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

#10966

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain.Test/BlockchainProcessorTests.cs
+++ b/src/Nethermind/Nethermind.Blockchain.Test/BlockchainProcessorTests.cs
@@ -324,7 +324,7 @@ public class BlockchainProcessorTests
                 _processor.BlockAdded += OnBlockAdded;
                 try
                 {
-                    Task.Run(() =>
+                    Task suggestTask = Task.Run(() =>
                     {
                         try
                         {
@@ -344,6 +344,20 @@ public class BlockchainProcessorTests
                         suggestCompleted.Task.Wait(ProcessingWait),
                         Is.True,
                         $"Timed out waiting for {block.ToString(Block.Format.Short)} to complete suggestion");
+                    // Give the Task.Run time to finish the queue write inside
+                    // Enqueue(). BlockAdded fires before the recovery queue write,
+                    // so suggestCompleted resolving does not mean the block is in
+                    // the queue yet. Without this, consecutive Suggested() calls
+                    // can race their Task.Run threads to TryWrite, reordering
+                    // blocks in the recovery queue.
+                    //
+                    // Bounded to 100ms to avoid deadlock: when _queueCount == 1,
+                    // AllowSynchronousContinuations can inline ProcessBlocks() on
+                    // the Task.Run thread, which blocks until the test calls
+                    // Allow(). An infinite wait would deadlock. The timeout
+                    // expiring is harmless — single-block enqueues have no
+                    // ordering concern.
+                    suggestTask.Wait(100);
                 }
                 finally
                 {


src/Nethermind/Nethermind.Consensus/Processing/BlockchainProcessor.cs
```diff
@@ -159,9 +158,13 @@ public async ValueTask Enqueue(Block block, ProcessingOptions processingOptions)
                     {
                         Interlocked.Add(ref _currentRecoveryQueueSize, block.Transactions.Length);
                         _recoveryQueue.Writer.TryWrite(blockRef);
+                        BlockAdded?.Invoke(this, new BlockEventArgs(block));
```
newmanifold: Should i make this
```suggestion
                         if (_recoveryQueue.Writer.TryWrite(blockRef))
                        {
                            BlockAdded?.Invoke(this, new BlockEventArgs(block));
                        }
```

To make the guarantee to actually hold incase of bounded queues?

PR #10968, opened 2026-03-27
fix: prevent FlatDB live stall and make restart recovery exact

benaadams:
### Track A: Prevent Live Processing Stall

- **Replace unbounded producer-side wait in `FlatDbManager.AddSnapshot`** with an inline drain fallback when the compactor queue is saturated — the processing thread now helps drain compaction/persistence instead of blocking forever on channel capacity
- **Add processing watchdog** in `BlockchainProcessor` — a single long-lived background task that emits periodic warnings when a block has been processing longer than a configurable threshold (default 30s)
- **Harden `AddressWarmer.Wait()` and `BranchProcessor.WaitAndClear`** with periodic warning logs while blocked, preserving the invariant that processing does not continue while the prewarmer can still publish into shared caches
- **Add per-transaction `CancellationToken` check** in `BlockValidationTransactionsExecutor.ProcessTransactions` — the token was accepted but never observed between transactions
- **Add backpressure metrics** (`CompactorQueueFullCount`, `InlineDrainActivationCount`) to the FlatDB metrics class
- **Add startup diagnostic logs** printing stored persisted-boundary metadata, FlatDB persisted `StateId`, canonical status, and recoverability on every startup

### Track B: Make Restart Recovery Exact

- **Persist exact boundary hash metadata** alongside the block number when `BestPersistedState` is updated — both are written atomically in a single DB batch
- **Add startup reconciliation** in `BlockTree.Initializer.LoadStartBlock()` that validates the restored block against the exact FlatDB persisted `StateId` and canonical chain status
- **Add startup repair** that promotes a non-canonical persisted-state block to the main chain, or finds the canonical block matching the exact persisted state root — fails loudly with full diagnostics if no match exists instead of drifting into the later `Maximum size of branch reached (8192)` failure
- **Harden `StartupBlockTreeFixer`** to only suggest blocks descending from the repaired head using per-level eligible parent hash tracking, and to log an error if the head still doesn't match persisted state info
- **Enrich the max-branch-size exception** with persisted-boundary number/hash, flat persisted state, and reconciliation outcome
- **Introduce `IPersistedStateInfoProvider`** seam in `Nethermind.State` (implemented by `FlatPersistedStateInfoProvider`) to expose persisted state info to `Nethermind.Blockchain` without creating a project cycle

#### Requires testing

- [x] Yes

- [x] Yes

#### Notes on testing

**FlatDB backpressure (Nethermind.State.Flat.Test):**
- `AddSnapshot_WhenCompactorQueueIsFull_CompletesWithinBoundedTime` — proves the producer does not block indefinitely under queue saturation
- `AddSnapshot_WhenQueueSaturated_EmitsBackpressureSignal` — verifies metric counters increment on inline drain
- `HasStateForBlock_EarlierBlockNumberWithoutExactStateId_ReturnsFalse` — locks in exact `StateId` semantics
- `FlatPersistedStateInfoProvider` tests for `TryGetPersistedStateInfo` and `HasRecoverableStateForBlock`

**Force-persist path (Nethermind.State.Flat.Test):**
- `DetermineSnapshotToPersist_WhenFinalizationIsStale_ChoosesForcePersistCandidate`
- `AddToPersistence_WhenForcePersisting_AdvancesCurrentPersistedStateMonotonically`

**Exact startup restore (Nethermind.Blockchain.Test):**
- `Loads_start_block_from_exact_persisted_boundary_hash_when_present`
- `Load_start_block_when_number_only_restore_points_to_exact_persisted_state_promotes_it_to_main_chain`
- `Load_start_block_when_no_exact_repair_target_exists_fails_with_actionable_diagnostics`
- `Load_start_block_failure_contains_persisted_boundary_and_flat_state_diagnostics`
- `Load_start_block_when_persisted_boundary_is_invalid_fails_without_exact_canonical_match`
- `Load_start_block_when_number_only_restore_matches_state_but_is_not_canonical_fails_loudly`
- `Load_start_block_is_backward_compatible_when_only_number_metadata_exists`
- `Load_start_block_repair_rewrites_number_and_hash_metadata_together`

**Startup fixer (Nethermind.Blockchain.Test):**
- `Fixer_starts_from_repaired_head`
- `Fixer_with_repaired_head_and_recoverable_parent_suggests_blocks_normally`
- `Fixer_only_suggests_blocks_that_descend_from_repaired_head`
- `Fixer_logs_error_when_head_does_not_match_persisted_state_info`

**End-to-end restart:**
- `Restart_with_exact_persisted_boundary_hash_replays_only_canonical_descendants_from_loaded_head`

**Observability:**
- `ProcessBlock_WhenBlockTakesLongerThanThreshold_EmitsWatchdogWarning`
- `BuildMaxBranchSizeReachedMessage_IncludesStartupDiagnostics`
- `AddressWarmerWait_WhenWarmupDoesNotFinish_WarnsWithoutReportingCompletion`
- `BranchProcessor_waits_for_prewarmer_task_before_returning`
- `BlockValidationTransactionsExecutor_stops_between_transactions_when_cancelled`

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain/BlockTree.Initializer.cs
+++ b/src/Nethermind/Nethermind.Blockchain/BlockTree.Initializer.cs
@@ -381,8 +381,8 @@ public partial class BlockTree
             bool? hasRecoverableState = startBlock is null
                 ? null
                 : _persistedStateInfoProvider?.HasRecoverableStateForBlock(startBlock.Header);
-            string persistedStateDescription = _persistedStateInfoProvider?.TryGetPersistedStateInfo(out PersistedStateInfo stateInfo) == true
-                ? $"{stateInfo.BlockNumber} / {stateInfo.StateRoot}"
+            string persistedStateDescription = persistedStateInfoValue is PersistedStateInfo logStateInfo
+                ? $"{logStateInfo.BlockNumber} / {logStateInfo.StateRoot}"
                 : "<unavailable>";
 
             Logger.Info(
--- a/src/Nethermind/Nethermind.Blockchain/BlockTree.Initializer.cs
+++ b/src/Nethermind/Nethermind.Blockchain/BlockTree.Initializer.cs
@@ -357,15 +357,17 @@ public partial class BlockTree
             persistedStateInfoValue = persistedStateInfo;
         }
 
+        bool isCanonicalStartBlock = startBlock is not null && IsCanonical(startBlock);
+
         if (persistedStateInfoValue is PersistedStateInfo exactPersistedState &&
             startBlock is not null &&
-            !MatchesPersistedState(startBlock, exactPersistedState))
+            (!MatchesPersistedState(startBlock, exactPersistedState) || !isCanonicalStartBlock))
         {
             startBlock = RepairPersistedBoundary(startBlock, persistedNumber, persistedHash, exactPersistedState, out reconciliationOutcome);
         }
         else if (persistedStateInfoValue is PersistedStateInfo)
         {
-            reconciliationOutcome = "validated against exact persisted state";
+            reconciliationOutcome = "validated against exact persisted state and canonical chain";
         }
         else if (startBlock is not null)
         {
@@ -452,10 +454,27 @@ public partial class BlockTree
         block.Number == persistedStateInfo.BlockNumber &&
         block.StateRoot == persistedStateInfo.StateRoot;
 
+    private bool IsCanonical(Block block) =>
+        block.Hash is not null &&
+        FindCanonicalBlockInfo(block.Number)?.BlockHash == block.Hash;
+
     private Block RepairPersistedBoundary(Block restoredBlock, long? persistedNumber, Hash256? persistedHash, PersistedStateInfo persistedStateInfo, out string reconciliationOutcome)
     {
-        Block? repairedBlock = FindBlockByStateRoot(persistedStateInfo.BlockNumber, persistedStateInfo.StateRoot) ??
-            FindRecoverableCanonicalBoundary(Math.Min(restoredBlock.Number, persistedStateInfo.BlockNumber));
+        bool promotedRestoredBlockToMainChain = false;
+        Block? repairedBlock = null;
+        if (MatchesPersistedState(restoredBlock, persistedStateInfo))
+        {
+            repairedBlock = restoredBlock;
+            if (!IsCanonical(repairedBlock))
+            {
+                PromoteStartupBoundaryToMainChain(repairedBlock);
+                promotedRestoredBlockToMainChain = true;
+            }
+        }
+        else
+        {
+            repairedBlock = FindCanonicalBlockByStateRoot(persistedStateInfo.BlockNumber, persistedStateInfo.StateRoot);
+        }
 
         if (repairedBlock is null)
         {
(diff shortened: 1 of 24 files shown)


src/Nethermind/Nethermind.State.Flat/FlatDbManager.cs
```diff
@@ -343,18 +356,26 @@ public void AddSnapshot(Snapshot snapshot, TransientResource transientResource)
...
-                if (_logger.IsWarn) _logger.Warn("Compactor job stall! Insufficient reorg depth or too slow persistence!");
-                _compactorJobs.Writer.WriteAsync(endBlock).AsTask().Wait();
+                Metrics.CompactorQueueFullCount++;
+                if (_logger.IsWarn)
+                {
+                    _logger.Warn($"Compactor job stall! Insufficient reorg depth or too slow persistence! Persisted state: {persistedStateId}, snapshot target: {endBlock}. Executing inline drain.");
+                }
+
+                Metrics.InlineDrainActivationCount++;
+                ExecuteCompactAndPersistInline(endBlock, transientResource);
```
asdacap: It should not persist inline here as there can be another parallel persist running at the same time.

benaadams: fixed

LukaszRozmej (commented):
aggresively de-duplicate tests?

flcl42:
is it still relevant?

src/Nethermind/Nethermind.State.Flat/FlatDbManager.cs
```diff
@@ -343,18 +353,54 @@ public void AddSnapshot(Snapshot snapshot, TransientResource transientResource)
...
+                {
+                    // Ignore it, just dispose
+                    transientResource.Dispose();
+                }
+            }
+            else
             {
-                // Ignore it, just dispose
-                transientResource.Dispose();
+                Metrics.CompactorQueueFullCount++;
```
AnkushinDaniil, line 366: It seems when TryWrite fails here, transientResource is neither passed to _populateTrieNodeCacheJobs nor disposed — so the BloomFilter leaks once per block exactly under the backpressure this PR targets. Before, it was handled unconditionally before the compactor enqueue.

src/Nethermind/Nethermind.Blockchain/BlockTree.cs
```diff
@@ -1728,7 +1747,20 @@ public long? BestPersistedState
                 _highestPersistedState = value;
                 if (value.HasValue)
                 {
-                    _blockInfoDb.Set(StateHeadHashDbEntryAddress, Rlp.Encode(value.Value).Bytes);
+                    using (_blockInfoDb.StartWriteBatch())
```
AnkushinDaniil, line 1750: I think the batch from StartWriteBatch() is discarded here, and Set/PutSpan/Remove go directly to the DB — so the number+hash writes are not actually atomic and a crash in between still leaves them inconsistent.

src/Nethermind/Nethermind.State.Flat/FlatDbManager.cs
```diff
@@ -343,18 +353,54 @@ public void AddSnapshot(Snapshot snapshot, TransientResource transientResource)
...
+        if (_cancelTokenSource.Token.IsCancellationRequested)
+        {
+            return;
+        }
+
+        long stallStart = Stopwatch.GetTimestamp();
+        while (true)
+        {
+            Task<bool> waitToWriteTask = _compactorJobs.Writer.WaitToWriteAsync(_cancelTokenSource.Token).AsTask();
+            if (waitToWriteTask.Wait(_compactorEnqueueWarningDelay))
```
AnkushinDaniil, line 383: waitToWriteTask.Wait(timeout) throws AggregateException(TaskCanceledException) if _cancelTokenSource is cancelled during shutdown while a producer is stalled here — ProcessBlocks' "is not OperationCanceledException" filter won't match it, so the in-flight block gets reported as ProcessingResult.Exception.

src/Nethermind/Nethermind.Blockchain/BlockTree.Initializer.cs
```diff
@@ -322,14 +324,21 @@ public enum BinarySearchDirection
...
+        string reconciliationOutcome = "startup boundary unavailable";
         byte[] persistedNumberData = _blockInfoDb.Get(StateHeadHashDbEntryAddress);
+        byte[]? persistedHashData = _blockInfoDb.Get(StateHeadBlockHashDbEntryAddress);
         BestPersistedState = persistedNumberData is null ? null : new Rlp.ValueDecoderContext(persistedNumberData).DecodeLong();
         long? persistedNumber = BestPersistedState;
+        Hash256? persistedHash = persistedHashData is null ? null : new Hash256(persistedHashData);
         if (persistedNumber is not null)
         {
-            startBlock = FindBlock(persistedNumber.Value, BlockTreeLookupOptions.None);
+            startBlock = TryLoadStartBlockFromExactBoundary(persistedNumber.Value, persistedHash)
```
AnkushinDaniil, line 335: When _persistedStateInfoProvider is null (hash/trie layouts), the exact-boundary block is not checked for canonicality — after a reorg below the boundary, startup can set Head to a non-canonical block, while the old number-only FindBlock returned the canonical one.

src/Nethermind/Nethermind.State.Flat/Metrics.cs
```diff
@@ -85,4 +85,14 @@ public static class Metrics
...
 
+    [DetailedMetric]
+    [CounterMetric]
+    [Description("Number of times the compactor queue was full when adding a snapshot")]
+    public static long CompactorQueueFullCount { get; set; }
+
+    [DetailedMetric]
+    [CounterMetric]
+    [Description("Number of times AddSnapshot had to execute compaction and persistence inline")]
+    public static long InlineDrainActivationCount { get; set; }
```
AnkushinDaniil, line 96: It seems InlineDrainActivationCount is never incremented — the inline-drain fallback was removed after review, so the metric (and the PR description) still describe the old design.

src/Nethermind/Nethermind.State.Flat/IFlatDbManager.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Nethermind.State.Flat;
 
-public interface IFlatDbManager : IFlatCommitTarget
+public interface IFlatDbManager : IFlatCommitTarget, IPersistedStateInfoProvider
```
AnkushinDaniil, line 9: Not sure IFlatDbManager needs to extend IPersistedStateInfoProvider — DI resolves the standalone provider, and FlatDbManager gets a second implementation that can drift from it. Maybe drop the inheritance and the duplicate members?

benaadams:
Drifted too much, covered elesewhere

PR #11002, opened 2026-03-31
Add BAL-driven prewarming with bulk trie reader

asdacap:
## Summary

*This PR was generated by Claude.*

- Add `HintBal` and `ReadBalAsync` to `IWorldStateScopeProvider.IScope` so the prewarmer can use Block Access Lists (EIP-7928) to prefetch account and storage data before block execution
- Add `PatriciaTrieBulkReader` — a static utility that sorts keys by nibble path and traverses the trie once, fanning out at branch nodes to avoid redundant node resolutions (~1.8x faster than one-by-one reads for typical BAL sizes)
- `PrewarmerScopeProvider` fires bulk reads in the background via `HintBal`, cancelling on `StartWriteBatch` to avoid contention with processor-intensive write batches
- `FlatWorldStateScope` calls `HintGet` on each entry during `ReadBalAsync` so the trie node warmer also warms corresponding trie paths
- `TrieStoreScopeProvider` uses two-phase bulk reading: phase 1 bulk-reads all accounts from the state trie, phase 2 bulk-reads storage slots per account using the `StorageRoot` from phase 1

## Test plan

- [x] 11 unit tests for `PatriciaTrieBulkReader` verifying correctness against individual `Get` calls (empty tree, single/multiple entries, missing keys, mixed, deep extensions, index preservation)
- [x] `ReadBalAsync` integration test in `ScopeProviderTests` running against both `TrieStoreScopeProvider` and `FlatScopeProvider`
- [x] All 710 existing state tests pass
- [x] BenchmarkDotNet benchmark comparing bulk vs one-by-one reads at 16/64/256/1024 key counts
- [ ] Manual benchmark on mainnet data


src/Nethermind/Nethermind.State/PrewarmerScopeProvider.cs
asdacap, line 113: Probably should block the running bal read task to ensure we dont have stray read later.

asdacap: Good catch. Added `CancelBal()` which cancels and waits for the running BAL task before Commit, StartWriteBatch, and Dispose. — Claude

asdacap:
Closing in favor of a draft PR targeting feature/bal-recorder.

PR #11003, opened 2026-03-31
Add command-based IRunStep execution to the runner

asdacap:
## Summary

Currently, the only way to create an alternative CLI tool that reuses Nethermind's initialization infrastructure is to create a completely separate project. This leads to duplicated setup code (database, state, block tree, processors) and missed initialization because the proper init relies heavily on the `IStep` pipeline. For example, `Nethermind.PerfTest` manually reconstructs ~350 lines of blockchain infrastructure that the step pipeline already provides.

This PR fills that gap by allowing any `IStep` to be annotated with `[RunnerCommand("name")]`, making it invocable as a CLI subcommand. The runner resolves only the step's transitive dependencies, executes them, runs the command, and exits. This means new tools like trie verification, database inspection, or performance testing can be built as single-file steps that automatically get correct initialization through the dependency graph.

**Key benefits:**
- **No more duplicated init code** — command steps declare their dependencies and get proper initialization for free
- **Correct by construction** — the same DI container and step pipeline used by the full client is reused, so commands can't miss initialization
- **Minimal execution** — only the required subset of steps runs (e.g., `verify-trie` runs 7 steps instead of ~25), so commands start fast and exit cleanly
- **Easy to extend** — adding a new command is a single `IStep` class with two attributes, usable from plugins too

## Changes

- **`RunnerCommandAttribute`** — new attribute with `Name`, `Description`, `IsDefault` properties
- **`StepInfo`** — exposes `CommandName`, `CommandDescription`, `IsDefaultCommand` from the attribute
- **`EthereumStepsManager`** — refactored into `BuildStepGraph`, `ComputeTransitiveClosure`, `ExecuteFromRoot`. `InitializeAll` finds the default command; `InitializeCommand(name)` finds by name. Command steps' `dependents` are inert (not pushed into other steps).
- **`NethermindClientStep`** — default command depending on all leaf steps (`LogHardwareInfo`, `StartMonitoring`, `StartBlockProducer`, `DatabaseMigrations`, `StartLogIndex`, `StartRpc`). `StartGrpc` pushes in via `dependents` since it's conditionally registered.
- **`RunVerifyTrie`** — annotated with `[RunnerCommand("verify-trie")]`, always registered in `BuiltInStepsModule`. `DiagnosticMode.VerifyTrie` check preserved for backwards compatibility.
- **`EthereumRunner`** — added `StartCommand(commandName, token)`
- **`Program.cs`** — discovers commands via `TypeDiscovery`, registers optional `Argument<string>` on `RootCommand`, branches to `RunCommandAsync` (exits after command) vs `RunAsync` (waits for exit signal)
- **`EvmWarmer`** — moved from leaf to `LoadGenesisBlock` dependency (EVM should be warm before first block processing)

## Future migration candidates

The following standalone binaries manually reconstruct parts of the init pipeline and could adopt `[RunnerCommand]`:

| Project | What it does | Migration benefit |
|---------|-------------|-------------------|
| **`Nethermind.PerfTest`** | Replays blocks measuring throughput/GC. Manually builds ~350 lines of DB, state, block tree, processors, validators, genesis. | **Strong** — ~65-75% code elimination. Only perf metrics and block loading logic would remain. |
| `tools/Evm` (t8n) | EVM transaction execution for testing | Weak — intentionally lightweight, no DB/network |
| `tools/SendBlobs` | EIP-4844 blob sender | Weak — client-side, doesn't need node init |
| `tools/StatelessExecution` | Stateless block execution | Weak — specialized, manual EVM setup |

`Nethermind.PerfTest` is the strongest candidate — it duplicates `InitDatabase`, `InitializeBlockTree`, `InitializeBlockchain`, and `LoadGenesisBlock` entirely.

## Test plan

- [x] `dotnet build src/Nethermind/Nethermind.slnx` compiles
- [x] `dotnet test --project src/Nethermind/Nethermind.Runner.Test/` — step loader tests pass, smoke tests pass (96/96)
- [ ] `nethermind verify-trie --config mainnet` runs 7 steps, verifies trie, exits
- [ ] `nethermind` starts node normally with all steps (unchanged behavior)
- [ ] `nethermind --help` shows the optional `[command]` argument

### DIFF
--- a/src/Nethermind/Nethermind.Api/Steps/RunnerCommandAttribute.cs
+++ b/src/Nethermind/Nethermind.Api/Steps/RunnerCommandAttribute.cs
@@ -1,0 +1,14 @@
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System;
+
+namespace Nethermind.Api.Steps;
+
+[AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
+public class RunnerCommandAttribute(string name) : Attribute
+{
+    public string Name { get; } = name;
+    public string Description { get; init; } = "";
+    public bool IsDefault { get; init; }
+}
--- a/src/Nethermind/Nethermind.Api/Steps/StepInfo.cs
+++ b/src/Nethermind/Nethermind.Api/Steps/StepInfo.cs
@@ -27,6 +27,12 @@ namespace Nethermind.Api.Steps
                 StepType.GetCustomAttribute<RunnerStepDependenciesAttribute>();
             Dependencies = dependenciesAttribute?.Dependencies ?? [];
             Dependents = dependenciesAttribute?.Dependents ?? [];
+
+            RunnerCommandAttribute? commandAttribute =
+                StepType.GetCustomAttribute<RunnerCommandAttribute>();
+            CommandName = commandAttribute?.Name;
+            CommandDescription = commandAttribute?.Description;
+            IsDefaultCommand = commandAttribute?.IsDefault ?? false;
         }
 
         public Type StepBaseType { get; }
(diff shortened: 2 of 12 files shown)


stdevMac:
@asdacap stale-PR check: zero engagement since March and the adjacent runner refactors merged without it. Still wanted, or OK to close? If there's no reply in the next few days, we'll close this as part of the stale-PR cleanup — it can always be reopened or resubmitted.

asdacap:
Will redo

PR #11011, opened 2026-03-31
refactor: Make some methods and interfaces available for overriding CALL opcode from Arbitrum plugin

hudem1:
For arbitrum [PR](https://github.com/NethermindEth/nethermind-arbitrum/pull/786)

Arbitrum needs to override the CALL opcode implementation to move some state access checks before out of gas checks for witness generation/recording, as in arbitrum nitro, gasCall() is a pure accumulator: StateDB.Empty(address) always executes before the single OOG check in the interpreter. In Nethermind, gas is deducted inline, so the fix in the plugin moves the IsDeadAccount (and AccountExists) state reads before any gas deduction that could abort early.

To override the opcode, i need to make some methods and interfaces available from the plugin (essentially from private/internal to public)

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [x] No

- [ ] Yes
- [ ] No

## Remarks


src/Nethermind/Nethermind.Evm/VirtualMachine.cs
```diff
@@ -1355,7 +1355,7 @@ internal void AddLog(LogEntry logEntry)
         }
     }
 
-    private void AddTransferLog(VmState<TGasPolicy> currentState)
+    public void AddTransferLog(VmState<TGasPolicy> currentState)
```
smartprogrammer93:
```suggestion
    protected void AddTransferLog(VmState<TGasPolicy> currentState)
```

smartprogrammer93 (approved):

src/Nethermind/Nethermind.Evm/VirtualMachine.cs
hudem1: I can't because `AddTransferLog` is used in `InstructionCall` in the equivalent of `EvmInstructions` class (`ArbitrumEvmInstructions`)

damian-orzechowski (approved):

LukaszRozmej (approved):

PR #11016, opened 2026-04-01
XDC - Rewards Upgrade

cicr99:


src/Nethermind/Nethermind.Xdc/Contracts/MintedRecordContract.cs
```diff
@@ -0,0 +1,110 @@
...
+{
+    private static readonly Address MintedRecordAddress = Address.FromNumber(0x9a);
+    private static readonly UInt256 MintedRecordOnsetEpochSlot = UInt256.One;
+    private static readonly UInt256 MintedRecordOnsetBlockSlot = (UInt256)2;
+    private static readonly UInt256 MintedRecordPostMintedBase = UInt256.Parse("0x0100000000000000000000000000000000000000000000000000000000000000");
+    private static readonly UInt256 MintedRecordPostBurnedBase = UInt256.Parse("0x0200000000000000000000000000000000000000000000000000000000000000");
+    private static readonly UInt256 MintedRecordPostRewardBlockBase = UInt256.Parse("0x0300000000000000000000000000000000000000000000000000000000000000");
+
+    public void UpdateAccounting(
+        ITransactionProcessor transactionProcessor,
```
ak88, line 23: just pass the `WorldState `instead?

ak88 (dismissed):

src/Nethermind/Nethermind.Xdc/Spec/XdcReleaseSpec.cs
```diff
@@ -23,9 +24,9 @@ public class XdcReleaseSpec : ReleaseSpec, IXdcReleaseSpec
     public int TimeoutSyncThreshold { get; set; }        // send syncInfo after number of timeout
     public int TimeoutPeriod { get; set; }               // Duration in ms
     public double CertificateThreshold { get; set; }            // Necessary number of messages from master nodes to form a certificate
-    public double MasternodeReward { get; set; }         // Block reward per master node (core validator) - unit Ether
-    public double ProtectorReward { get; set; }          // Block reward per protector - unit Ether
-    public double ObserverReward { get; set; }           // Block reward per observer - unit Ether
+    public UInt256 MasternodeReward { get; set; }        // Block reward per masternode (core validator) in Wei
```
batrr, line 29: Fields not wired from chain spec config.

src/Nethermind/Nethermind.Xdc/XdcRewardCalculator.cs
```diff
@@ -128,9 +167,29 @@ public BlockReward[] CalculateRewards(Block block)
...
                             masternodes = new HashSet<Address>(h.ValidatorsAddress!);
-                        // TIPUpgradeReward path (protector/observer selection) is currently ignored,
-                        // because on mainnet the upgrade height is set to an effectively unreachable block.
-                        // If/when that changes, we must compute protector/observer sets here.
+
+                        if (h.Number >= spec.TipUpgradeRewardBlock)
+                        {
+                            // TIPUpgradeReward path: select protector and observer sets from stake-sorted candidates.
+                            // Exclude current masternodes and penalized nodes from the checkpoint header.
+                            Address[] candidatesByStake = _masternodeVotingContract.GetCandidatesByStake(h) ?? [];
```
batrr: Go uses the state of the current block's parent, not the one in loop

batrr: and it uses stable sort. not an unstable variant that we use in `GetCandidatesByStake`

batrr (commented):
Agree with Claude on :
TipUpgradeReward transition block not registered in AddTransitions

src/Nethermind/Nethermind.Xdc/Contracts/MintedRecordContract.cs
```diff
@@ -0,0 +1,110 @@
...
+
+    private static UInt256 ReadStorage(IWorldState worldState, UInt256 slot)
+    {
+        ReadOnlySpan<byte> value = worldState.Get(new StorageCell(MintedRecordAddress, slot));
+        if (value.Length == 0)
+        {
+            return UInt256.Zero;
+        }
+
+        return new UInt256(value);
```
batrr: Is this endianness correct?

src/Nethermind/Nethermind.Xdc/XdcRewardCalculator.cs
cicr99: excellent catch! Fixed.
Also btw, I noticed that the block number to check tipUpgradeRewardBlock against was also not the one in the loop, but the original header, being that the case it does make sense to have the flag for tipUpgrade, same as we had for penalties. I've modified that as well

src/Nethermind/Nethermind.Xdc/Contracts/MintedRecordContract.cs
cicr99: Nope, I missed the flag. Thanks

batrr (approved):

src/Nethermind/Nethermind.Xdc/Contracts/MintedRecordContract.cs
cicr99: On this, I had 2 ideas:
- First one was to pass world state directly and not get it through the processor, which meant passing the state to the reward calculator constructor (this was the one I tried when I first addressed comments). But this messed up with DI, hence the failing CI for runner tests.
- Second thing was to pass the state and not the processor, but in that case we would still need to do the casting for the processor outside the method, as the interface does not expose the state.

So for the moment I just left it as it is, but please let's discuss if you think some other way would be better.

batrr (approved):

ak88 (approved):

ak88 (approved):

PR #11022, opened 2026-04-02
fix(hive): fix snap sync nethermind-to-nethermind test

### LINKED ISSUE #10963: Hive: sync nethermind from nethermind failing
Sync test `sync nethermind from nethermind` fails every run on master.

Failing run: https://github.com/NethermindEth/nethermind/actions/runs/23583271426

Part of #10956

newmanifold:
Fixes #10963

## Changes

- Remove unused `AllowBeaconHeaderSync` gate from `IBeaconSyncStrategy`, `BeaconSync`, and `StartingSyncPivotUpdater`
- Change `InitBeaconHeaderSync` return type to `void` (always returned `true`)
- Remove dead identity decorator in `TaikoPlugin`

## Why this fixes the hive test

The hive snap sync test (nethermind-to-nethermind) sends only a single FCU. The `AllowBeaconHeaderSync` gate blocked `BeaconHeadersSyncFeed` from activating until `StartingSyncPivotUpdater` had set the global pivot — but the updater runs asynchronously and hadn't completed when that single FCU arrived. With no subsequent FCUs to retry, the beacon feed never activated and the node couldn't sync.

On mainnet/testnets this isn't visible because FCUs arrive every slot (~12s), so a later FCU eventually finds the gate open. The hive test has only one chance.

Removing the gate allows `BeaconHeadersSyncFeed` to activate immediately on the first FCU.

**Note:** `BeaconHeadersSyncFeed` won't actually start until `UpdatingPivot` mode ends (`ShouldBeInBeaconHeaders` requires `!shouldBeInUpdatingPivot`), so `SyncPivot` is always set before the feed runs.

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


asdacap (dismissed):

flcl42 (commented):
You can also enable that hive sync test again here

marcindsobczak:
Wouldn’t removing the AllowBeaconHeaderSync logic reintroduce the vulnerability described here?
https://github.com/NethermindEth/nethermind/issues/6304

That logic was added as a safeguard against conflicting pivots when the CL is lagging:
https://github.com/NethermindEth/nethermind/pull/7614

newmanifold:
@marcindsobczak I could be wrong, but we shouldn't be running the beacon headers sync if the global pivot is updating in the first place. `!ShouldBeInUpdatingPivot` should be preventing it. I think the original issue might be some race between BlockDownloader and BeaconHeaderSync.

The current issue is that we don't start beacon headers sync on the first FCU call. On a real chain it's fine we will start after 12s — but with Hive tests, they will only send the FCU call once and expect the client to sync.

If we don't want to remove the `_canInitBeaconHeaderSync` guard, maybe we can start the sync when AllowBeaconHeaderSync is called instead of waiting on the next FCU.

Open to any other suggestion

newmanifold:
@marcindsobczak any suggestions?

marcindsobczak:
We should either merge it in its current form and test it thoroughly, including CL restarts at the beginning of sync to try to reproduce the old issue, or keep the logic we added to prevent it and find a way to start sync on the first FCU.
My gut feeling is that we might still have this issue, since it shouldn’t ever happen but somehow did. So it may be better to trigger sync on `AllowBeaconHeaderSync`. That said, feel free to remove it, just make sure to test it carefully.

LukaszRozmej:
> We should either merge it in its current form and test it thoroughly, including CL restarts at the beginning of sync to try to reproduce the old issue, or keep the logic we added to prevent it and find a way to start sync on the first FCU. My gut feeling is that we might still have this issue, since it shouldn’t ever happen but somehow did. So it may be better to trigger sync on `AllowBeaconHeaderSync`. That said, feel free to remove it, just make sure to test it carefully.

@stdevMac , @kamilchodola can you schedule those tests on this branch?

LukaszRozmej (approved):

benaadams (approved):

PR #11027, opened 2026-04-03
fix: Duplicate-signature quorum bypass in XDC vote aggregation and certificate verification

Donjon-Cerberus:
Closes #11026

## Summary

Automated security fix for **Duplicate-signature quorum bypass in XDC vote aggregation and certificate verification** (Critical).

**CWE**: CWE-CWE-347
**OWASP**: A04:2021-Insecure Design
**Fix Confidence**: high

## What Changed
The vulnerability allows a single validator to forge quorum by submitting multiple distinct ECDSA signatures for the same vote/timeout message. Three files are patched:

1. **VotesManager.cs** - `GetValidSignatures()`: Added a `seenSigners` HashSet to track unique signer addresses. Only the first signature from each unique valid masternode signer is included. This ensures the QC built from collected votes contains at most one signature per validator.

2. **QuorumCertificateManager.cs** - `VerifyCertificate()`: Replaced the `Signatures.Distinct()` (which deduplicates by signature bytes) and `Parallel.ForEach` pattern with a sequential loop that recovers each signer, validates masternode membership, and collects unique signers in a `HashSet<Address>`. The threshold is now checked against `uniqueSigners.Count` instead of signature count. Added `using System.Collections.Generic` import.

3. **TimeoutCertificateManager.cs** - Three changes:
   - `HandleTimeoutVote()`: Threshold check now counts unique signers (`collectedTimeouts.Where(t => t.Signer is not null).Select(t => t.Signer).Distinct().Count()`) instead of raw `collectedTimeouts.Count`.
   - `OnTimeoutPoolThresholdReached()`: Deduplicates signatures by signer address before building the TimeoutCertificate.
   - `VerifyTimeoutCertificate()`: Replaced `HashSet<Signature>` with signer recovery loop and `HashSet<Address>` for unique signer counting, checking threshold against unique signer count.

## Caveats
- The QC verification was changed from Parallel.ForEach to sequential iteration. This trades some parallelism for correctness (no need for concurrent HashSet). The masternode count is bounded, so performance impact should be negligible.
- The initial threshold check in VotesManager.HandleVote still uses raw roundVotes.Count as a fast-path, which could let an attacker trigger the more expensive GetValidSignatures call. However, the security-critical second check (validSignatures.Length < requiredVotes) now correctly counts unique signers, so no bypass is possible.
- Timeout.Signer is expected to be set before reaching HandleTimeoutVote (either by FilterTimeout for external timeouts or explicitly for local ones). The null check in the deduplication handles edge cases defensively.

## Verification Checklist

- [ ] Review the code change
- [ ] Run tests to verify no regression
- [x] Verify the vulnerability is addressed — *already verified by Cerberus Sentinel*

---
*Created by [Cerberus](https://github.com/Donjon-Cerberus) Merlin*

### DIFF
--- a/src/Nethermind/Nethermind.Xdc/QuorumCertificateManager.cs
+++ b/src/Nethermind/Nethermind.Xdc/QuorumCertificateManager.cs
@@ -12,6 +12,7 @@ using Nethermind.Xdc.Errors;
 using Nethermind.Xdc.Spec;
 using Nethermind.Xdc.Types;
 using System;
+using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


batrr (commented):
Valid issue. But think the fix should be in XdcPool.Add - reject duplicate signers at insertion rather than filtering later.

batrr:
Closing in favour of https://github.com/NethermindEth/nethermind/pull/11072. 

Thank you for the contribution and for identifying the issue.

PR #11043, opened 2026-04-06
[TEST] Check which test fail with flat

asdacap:
## Summary
- Enable FlatDb by default (`IFlatDbConfig.Enabled` = `true`)
- Exploratory PR to see which CI checks fail

## Changes
- `FlatDbConfig.cs`: default `Enabled` from `false` → `true`
- `IFlatDbConfig.cs`: update `DefaultValue` annotation from `"false"` → `"true"`

## Testing
Observing CI results.


stdevMac:
Closing: exploratory CI probe — the results were observed and the experiment has concluded; FlatDb remains opt-in for now, so the default flip is not mergeable.

PR #11055, opened 2026-04-06
feat(eip): add bogota and eip8198(quick slots)

DarkLord017:
Fixes Closes Resolves #

## Changes

- Add eip8198 quick slots prototype

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Remarks


src/Nethermind/Nethermind.Specs/Forks/26_Bogota.cs
```diff
@@ -0,0 +1,18 @@
+// SPDX-FileCopyrightText: 2026 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using Nethermind.Core;
+
+namespace Nethermind.Specs.Forks;
+
+public class Bogota() : NamedReleaseSpec<Bogota>(Amsterdam.Instance)
+{
+    public override void Apply(ReleaseSpec spec)
```
AnkushinDaniil: CI is red because of this — after the fork-pattern refactor on master, Apply has to override NamedReleaseSpec.Apply(NamedReleaseSpec), not take ReleaseSpec.

src/Nethermind/Nethermind.Specs/ChainSpecStyle/Json/GethGenesisConfigJson.cs
```diff
@@ -68,6 +68,8 @@ public class GethGenesisConfigJson : IHasNamedForks
     public ulong? Bpo4Time { get => GetTime(); set => SetTime(value); }
     public ulong? Bpo5Time { get => GetTime(); set => SetTime(value); }
 
+    public ulong? BogotaTime { get; set; }
```
AnkushinDaniil: I think BogotaTime should use the `get => GetTime(); set => SetTime(value);` pattern like the other fork times — as a plain auto-property it never lands in NamedForkTimestamps, so blobSchedule.bogota entries are silently dropped.

src/Nethermind/Nethermind.Specs/ChainSpecStyle/ChainSpecBasedSpecProvider.cs
```diff
@@ -351,6 +351,12 @@ protected virtual ReleaseSpec CreateReleaseSpec(ChainSpec chainSpec, long releas
                 releaseSpec.MaxCodeSize = CodeSizeConstants.MaxCodeSizeEip7954;
             }
 
+            releaseSpec.IsEip8198Enabled = (chainSpec.Parameters.Eip8198TransitionTimestamp ?? ulong.MaxValue) <= releaseStartTimestamp;
```
AnkushinDaniil, line 354: This path sets only SlotDurationMs, but Bogota.Apply also scales the blob counts — so the same fork activated via eip8198TransitionTimestamp vs the named spec gives different blob limits. I think both paths should apply the same params.

src/Nethermind/Nethermind.Consensus/Validators/HeaderValidator.cs
```diff
@@ -263,6 +263,18 @@ protected virtual bool ValidateGasLimitRange(BlockHeader header, BlockHeader par
             // Gas-limit-vs-parent comparisons don't apply at genesis, so skip them.
             if (parent is null) return true;
 
+            IReleaseSpec parentSpec = _specProvider.GetSpec(parent);
+            if (spec.IsEip8198Enabled && !parentSpec.IsEip8198Enabled)
+            {
+                long expectedGasLimit = parent.GasLimit * (long)spec.SlotDurationMs / (long)parentSpec.SlotDurationMs;
```
AnkushinDaniil: parent.GasLimit * (long)spec.SlotDurationMs can overflow long for the long.MaxValue genesis gas limits handled just below — checked builds would throw here. Same expression is in BlockProducerBase. Maybe compute via UInt256?

src/Nethermind/Nethermind.Consensus/Validators/HeaderValidator.cs
```diff
@@ -263,6 +263,18 @@ protected virtual bool ValidateGasLimitRange(BlockHeader header, BlockHeader par
...
+                if (header.GasLimit != expectedGasLimit)
+                {
+                    if (_logger.IsWarn) _logger.Warn($"Invalid block header ({header.Hash}) - gas limit should be parent gas limit scaled by slot duration change when EIP-8198 is activated. Expected: {expectedGasLimit}, actual: {header.GasLimit}");
+                    error = BlockErrorMessages.InvalidGasLimit;
+                    return false;
+                }
+                return true;
```
AnkushinDaniil, line 276: It seems the transition branch skips the MinGasLimit floor — if scaling takes the limit below it (e.g. parent 5000), the smaller value is required and accepted. Not sure what the expected behaviour is, but probably it should be clamped.

src/Nethermind/Nethermind.Specs/Forks/26_Bogota.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        spec.Name = "Bogota";
+        spec.IsEip8198Enabled = true;
+        spec.SlotDurationMs = Eip8198Constants.NewSlotDurationMs;
+        spec.MaxBlobCount = spec.MaxBlobCount * Eip8198Constants.NewSlotDurationMs / Eip8198Constants.OldSlotDurationMs;
+        spec.TargetBlobCount = spec.MaxBlobCount * 2 / 3;
```
AnkushinDaniil, line 16: Target/max blob counts are rescaled but BlobBaseFeeUpdateFraction stays at the value tuned for target 14 — the BPO forks set all three together. Is it intentional here?

src/Nethermind/Nethermind.Specs/ChainSpecStyle/GethGenesisLoader.cs
```diff
@@ -183,9 +185,10 @@ private void LoadParameters(GethGenesisJson gethGenesis, ChainSpec chainSpec)
             [nameof(BPO1)] = new(3),
             [nameof(BPO2)] = new(4),
             [nameof(Amsterdam)] = new(5),
-            [nameof(BPO3)] = new(6),
-            [nameof(BPO4)] = new(7),
-            [nameof(BPO5)] = new(8),
+            [nameof(Bogota)] = new(6),
```
AnkushinDaniil, line 188: Bogota is ordered before BPO3-5 here, but the BPO3-5 specs still inherit from Amsterdam — so a lineage where BPO3 follows Bogota would drop the eip8198 fields. I think either the order or the parenting should change.

src/Nethermind/Nethermind.Specs/ChainSpecStyle/GethGenesisLoader.cs
```diff
@@ -162,6 +162,8 @@ private void LoadParameters(GethGenesisJson gethGenesis, ChainSpec chainSpec)
             Eip7251ContractAddress = config.PragueTime is null ? null : Eip7251Constants.ConsolidationRequestPredeployAddress,
             Eip7934MaxRlpBlockSize = Eip7934Constants.DefaultMaxRlpBlockSize,
 
+            Eip8198TransitionTimestamp = config.BogotaTime,
```
AnkushinDaniil, line 165: Could you add a loader/spec-provider test for the eip8198TransitionTimestamp/bogotaTime wiring? It would have caught the BogotaTime issue above.

PR #11074, opened 2026-04-08
fix: resolve tx type based on target block fork in eth_call

LukaszRozmej:
Fixes https://github.com/NethermindEth/nethermind/issues/11069
Supersedes https://github.com/NethermindEth/nethermind/pull/11073

## Changes

Since #9529, `DeriveTxType` defaults to `EIP1559TransactionForRpc` when no type-specific fields are present. This causes `eth_call`, `eth_estimateGas`, and `eth_createAccessList` to fail on pre-Berlin blocks with "EIP-2930 is not enabled" because `AccessListTransactionForRpc.ToTransaction()` unconditionally sets `AccessList = AccessList.Empty`.

This PR makes the type resolution fork-aware:
- `DeriveTxType` now tracks whether the type was **inferred** (defaulted) vs explicitly set via an `IsTypeDefaulted` flag
- `ToTransaction()` accepts an optional `IReleaseSpec` — when the type was inferred and the target block doesn't support EIP-2930, it resolves to Legacy instead of EIP-1559
- Each subclass (`AccessListTransactionForRpc`, `EIP1559TransactionForRpc`, `BlobTransactionForRpc`, `SetCodeTransactionForRpc`) skips its type-specific fields when the resolved type is below its level
- `TxExecutor` threads `ISpecProvider` and passes the target block's spec to `ToTransaction()` via `Prepare(call, header)`
- Refactored `ExecutorBase.Prepare` to accept `BlockHeader`, eliminating a duplicate `SearchForHeader` call

Explicitly typed transactions (`"type": "0x2"`) and discriminator-matched types (`{"accessList":[]}`) are **not affected** — only the fallback default is fork-dependent.

## Types of changes

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other

## Testing

- [x] Yes

13 new tests covering spec-aware type resolution: defaulted types on pre/post-Berlin specs, explicit types preserved regardless of spec, discriminator-matched types unaffected, gasPrice-matched Legacy unaffected, null spec backwards compatibility, and full Transaction property assertions.

### DIFF
--- a/src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/AccessListTransactionForRpc.cs
+++ b/src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/AccessListTransactionForRpc.cs
@@ -3,6 +3,7 @@
 
 using System.Text.Json.Serialization;
 using Nethermind.Core;
+using Nethermind.Core.Specs;
 using Nethermind.Int256;
 
 namespace Nethermind.Facade.Eth.RpcTransaction;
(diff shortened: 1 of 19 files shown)


svlachakis (approved):
https://github.com/NethermindEth/nethermind/actions/runs/24151495234

smartprogrammer93 (approved):

LukaszRozmej:
@claude re-review

PR #11077, opened 2026-04-09
eth_estimateGas fix error decoding

svlachakis:
Fixes Closes Resolves #11059 

## Changes

EncodeErrorMessage fixes & tests

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No


LukaszRozmej (approved):
Is this cross-checked with other clients?

src/Nethermind/Nethermind.Evm/TransactionSubstate.cs
```diff
@@ -103,8 +103,20 @@ public TransactionSubstate(ReadOnlyMemory<byte> bytes,
...
+    {
+        if (!Utf8.IsValid(span))
+            return span.ToHexString(true);
+
+        string decoded = Encoding.UTF8.GetString(span);
+        foreach (char c in decoded)
+        {
+            if (char.IsControl(c) && c != '\t' && c != '\n' && c != '\r')
+                return span.ToHexString(true);
+        }
```
LukaszRozmej: Can we potentially use `IndexOfAny`?

svlachakis:
> Is this cross-checked with other clients?

Geth separates message (human-readable, ABI-decoded) from data (raw hex). Nethermind currently puts everything in message and code -32000 instead of 3. That's a bigger existing difference, which we I think we should tackle separately - this just fixes not returning bytes in the response.

src/Nethermind/Nethermind.Evm/TransactionSubstate.cs
LukaszRozmej: And maybe move the check before `GetString` and do it on the span?

src/Nethermind/Nethermind.Evm/TransactionSubstate.cs
```diff
@@ -103,8 +103,20 @@ public TransactionSubstate(ReadOnlyMemory<byte> bytes,
...
-        Utf8.IsValid(span) ? Encoding.UTF8.GetString(span) : span.ToHexString(true);
+    public static string EncodeErrorMessage(ReadOnlySpan<byte> span)
+    {
+        if (!Utf8.IsValid(span))
+            return span.ToHexString(true);
+
+        string decoded = Encoding.UTF8.GetString(span);
+        foreach (char c in decoded)
+        {
+            if (char.IsControl(c) && c != '\t' && c != '\n' && c != '\r')
```
alexb5dh: Maybe move this 3-chars set to a static field with descriptive naming, if they can't be grouped in other way?

alexb5dh (approved):

src/Nethermind/Nethermind.Evm/TransactionSubstate.cs
```diff
@@ -103,8 +105,24 @@ public TransactionSubstate(ReadOnlyMemory<byte> bytes,
...
+
+        string decoded = Encoding.UTF8.GetString(span);
+        ReadOnlySpan<char> chars = decoded.AsSpan();
+
+        // Reject char.IsControl chars except the safe whitespace \t, \n, \r.
+        // C0: 0x00–0x08, 0x0B–0x0C, 0x0E–0x1F | DEL: 0x7F | C1: 0x80–0x9F
+        if (chars.IndexOfAnyInRange('\x00', '\x08') >= 0 ||
+            chars.IndexOfAnyInRange('\x0B', '\x0C') >= 0 ||
+            chars.IndexOfAnyInRange('\x0E', '\x1F') >= 0 ||
+            chars.IndexOfAnyInRange('\x7F', '\x9F') >= 0)
```
LukaszRozmej: this does multiple passes, which can be even worse, maybe use `SearchValues`:


```
private static readonly SearchValues<char> DisallowedControlChars =
    SearchValues.Create(
        Enumerable.Range(0, 32)
            .Select(i => (char)i)
            .Where(c => c != '\t' && c != '\n' && c != '\r')
            .Append((char)127)
            .ToArray());

public static string EncodeErrorMessage(ReadOnlySpan<byte> span)
{
    if (!Utf8.IsValid(span))
        return span.ToHexString(true);

    string decoded = Encoding.UTF8.GetString(span);
    if (decoded.AsSpan().ContainsAny(DisallowedControlChars))
        return span.ToHexString(true);

    return decoded;
}
```

LukaszRozmej: But maybe if we expect only few characters, the original version was better?

svlachakis: I think as you suggested with searchvalues is cleaner

src/Nethermind/Nethermind.Evm/TransactionSubstate.cs
```diff
@@ -103,8 +112,17 @@ public TransactionSubstate(ReadOnlyMemory<byte> bytes,
...
+    {
+        if (!Utf8.IsValid(span))
+            return span.ToHexString(true);
+
+        string decoded = Encoding.UTF8.GetString(span);
+        if (decoded.AsSpan().ContainsAny(_disallowedControlChars))
+            return span.ToHexString(true);
+
+        return decoded;
+    }
```
benaadams, line 112: Ditch anything not ascii?

```suggestion
    public static string EncodeErrorMessage(ReadOnlySpan<byte> span)
    {
        if (span.IndexOfAnyExceptInRange((byte)32, (byte)126) >= 0)
            return span.ToHexString(true);

        return Encoding.ASCII.GetString(span);
    }
```

svlachakis: this seems to be the best option, thanks @benaadams

src/Nethermind/Nethermind.Evm/TransactionSubstate.cs
svlachakis: done with @benaadams suggestion

benaadams (approved):

PR #11079, opened 2026-04-09
style: enforce explicit type over var (IDE0008)

asdacap:
## Summary

- Enforce `csharp_style_var_*` rules as `error` severity in `.editorconfig`, banning `var` in favor of explicit types across the codebase
- Enable `EnforceCodeStyleInBuild` in `Directory.Build.props` so IDE0008 violations fail the build
- Replace all `var` usages with explicit types across 760+ files

## Non-formatting changes

The following config files contain **behavioral changes**, not just formatting:

- **`.editorconfig`** — changes `csharp_style_var_*` rules from `silent`/`suggestion` to `error`, promotes `IDE0008` from `warning` to `error`, and demotes `IDE0005` (unused usings) from `warning` to `suggestion`
- **`src/Nethermind/Directory.Build.props`** — adds `<EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>`, making code style violations fail the build

All other file changes are purely mechanical `var` → explicit type replacements.

## Test plan

- [x] Solution builds with 0 errors and 0 warnings (`dotnet build -c release`)
- [ ] CI passes


MarekM25 (approved):

flcl42 (approved):

.editorconfig
```diff
@@ -88,18 +88,18 @@ csharp_style_expression_bodied_indexers = true:silent
...
-csharp_style_var_for_built_in_types = true:silent
-csharp_style_var_when_type_is_apparent = true:suggestion
-csharp_style_var_elsewhere = false:suggestion
+csharp_style_var_for_built_in_types = false:error
+csharp_style_var_when_type_is_apparent = false:error
+csharp_style_var_elsewhere = false:error
 
 # Coding convention
-dotnet_diagnostic.IDE0005.severity = warning
+dotnet_diagnostic.IDE0005.severity = suggestion
```
newmanifold: Any particular reason to make to make this suggestion?

asdacap: No idea.

asdacap: ~claude: This IDE0005 change (unused usings from `warning` to `suggestion`) was part of the original branch changes, not something added by this PR's automated formatting. It may have been intentional to reduce noise since `EnforceCodeStyleInBuild` would otherwise turn unused using warnings into build friction. That said, this is independent of the `var` enforcement goal — happy to revert it if you'd prefer to keep IDE0005 as `warning`.

newmanifold: maybe it hallucinated, because otherwise our lint workflow won't fire on suggestions incase there is a duplicate import.
Lets revert this to `warning`?

LukaszRozmej: Keep it as warning please 
```suggestion
dotnet_diagnostic.IDE0005.severity = warning
```

LukaszRozmej (approved):
Maybe where there are long generic types:
- create aliases
- create named record types instead of using tuples

Also revert usings to warnings!

Otherwise cool!

.editorconfig
asdacap: ~claude: Reverted IDE0005 back to `warning` in 62bc6bc8ec. Thanks for catching that.

asdacap: ~claude: IDE0005 is kept as `warning`. Note: `EnforceCodeStyleInBuild=true` combined with IDE0005 requires `GenerateDocumentationFile=true` to enforce it at build time ([dotnet/roslyn#41640](https://github.com/dotnet/roslyn/issues/41640)), which would cause thousands of CS1591 (missing XML comment) errors. So we suppress the `EnableGenerateDocumentationFile` build error — IDE0005 is still enforced via the lint workflow, just not during `dotnet build`. All other rules (including IDE0008) work at build time without this limitation.

PR #11086, opened 2026-04-09
fix: validate filter string length before re-parsing

LukaszRozmej:
## Summary
- Add length validation before re-parsing string-encoded filter parameters in `eth_getLogs`
- Rejects filter strings that are null or exceed 1 MB to prevent excessive memory usage from double-parsing

## Test plan
- [x] Build compiles
- [ ] Existing filter tests pass

### DIFF
--- a/src/Nethermind/Nethermind.JsonRpc/Modules/Eth/Filter.cs
+++ b/src/Nethermind/Nethermind.JsonRpc/Modules/Eth/Filter.cs
@@ -30,10 +30,10 @@ public class Filter : IJsonRpcParam
         {
             if (filter.ValueKind == JsonValueKind.String)
             {
-                string? filterString = filter.GetString();
-                if (filterString is null || filterString.Length > 1_000_000)
+                string filterString = filter.GetString()!;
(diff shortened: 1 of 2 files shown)


batrr (approved):

LukaszRozmej:
@claude review

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/Filter.cs
```diff
@@ -30,7 +30,13 @@ public void ReadJson(JsonElement filter, JsonSerializerOptions options)
         {
             if (filter.ValueKind == JsonValueKind.String)
             {
-                doc = JsonDocument.Parse(filter.GetString());
+                string? filterString = filter.GetString();
+                if (filterString is null || filterString.Length > 1_000_000)
```
flcl42: `null` string leading to "exceeds maximum allowed length" looks weird. isn't null ok?

LukaszRozmej:
@claude review

flcl42 (approved):

PR #11088, opened 2026-04-09
fix: add bounds check before unsafe RLP length deserialization

LukaszRozmej:
## Summary

In `PeekLongPrefixAndContentLength`, the `DeserializeLengthRef` function uses `Unsafe.ReadUnaligned<ushort>` / `Unsafe.ReadUnaligned<int>` to read 1-4 bytes from a reference pointer derived via `Unsafe.Add`.

If the RLP data is truncated (e.g., prefix byte `0xBB` declares a 4-byte length follows, but only 1-2 bytes remain in the span), the unsafe read accesses memory past the span boundary. The existing `lengthOfLength <= 4` check validates the value range but not whether enough data actually remains.

### Fix

Add `position + 1 + lengthOfLength > data.Length` check before the `Unsafe.Add` call, throwing `RlpException` for truncated data. This follows the same defensive pattern used elsewhere in the RLP decoder.

## Test plan
- [x] Build compiles
- [ ] Existing RLP tests pass
- [ ] Consider adding fuzz test with truncated RLP inputs

### DIFF
--- a/src/Nethermind/Nethermind.Core.Test/RlpTests.cs
+++ b/src/Nethermind/Nethermind.Core.Test/RlpTests.cs
@@ -568,5 +568,16 @@ namespace Nethermind.Core.Test
             }
             return data;
         }
+
+        [TestCase(new byte[] { 0xB8 }, Description = "Long string prefix 0xB8 (1 byte of length), but no length byte")]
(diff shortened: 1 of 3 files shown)


batrr (approved):

LukaszRozmej:
@claude review

damian-orzechowski (approved):

diyarme-ads (approved):

smartprogrammer93 (approved):

PR #11089, opened 2026-04-09
fix: prevent infinite loop in RLP CountItems on malformed data

LukaszRozmej:
## Summary
- Add zero-length guard in `CountItems` to throw `RlpException` when `PeekNextRlpLength` returns 0
- Without this guard, a zero-length return would cause `position` to never advance, making the loop spin indefinitely
- Add `InternalsVisibleTo` for `Nethermind.Core.Test` to enable direct testing of internal RLP helpers

## Test plan
- [x] Build compiles
- [x] `CountItems_counts_single_byte_items` - verifies normal counting
- [x] `CountItems_respects_maxSearch` - verifies early termination
- [x] `CountItems_throws_on_truncated_long_form_prefix` - verifies truncated data throws

### DIFF
--- a/src/Nethermind/Nethermind.Core.Test/RlpTests.cs
+++ b/src/Nethermind/Nethermind.Core.Test/RlpTests.cs
@@ -568,5 +568,31 @@ namespace Nethermind.Core.Test
             }
             return data;
         }
+
+        [Test]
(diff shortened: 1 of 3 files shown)


batrr (approved):

LukaszRozmej:
@claude review

PR #11097, opened 2026-04-10
eth_getBlockByNumber - Remove totalDifficulty

### LINKED ISSUE #11096: eth_getBlockByNumber: remove totalDifficulty field to match go-ethereum
#### Summary

Nethermind includes a `totalDifficulty` field in `eth_getBlockByNumber` (and related block endpoints) that go-ethereum never returns. Worse, the value is frequently wrong: when blocks are fetched using `TotalDifficultyNotNeeded` lookup options (the common path for RPC), `block.TotalDifficulty` is null and Nethermind falls back to "0x0" - reporting zero total difficulty for blocks with real accumulated difficulty.

#### Current behaviour

Common case — block fetched without loading TD from DB:

```
{
  "difficulty": "0xf4240",
  "totalDifficulty": "0x0",
  ...
}
```

Less common case — block fetched with TD explicitly loaded:

```
{
  "difficulty": "0x1",
  "totalDifficulty": "0xf4243",
  ...
}
```

#### Expected behaviour (matching go-ethereum)

```
{
  "difficulty": "0xf4240",
  ...
}
```

No `totalDifficulty` field.

#### Root cause

`BlockForRpc.cs` constructor:

```
// Set TD only if network is not merged
if (specProvider.MergeBlockNumber is null)
{
    TotalDifficulty = block.Difficulty.IsZero ? null : block.TotalDifficulty ?? UInt256.Zero;
}
```

The ?? UInt256.Zero fallback silently produces "0x0" whenever block.TotalDifficulty was not loaded - which is the normal path for RPC block lookups that use BlockTreeLookupOptions.TotalDifficultyNotNeeded.

#### Evidence

[go-ethereum internal/ethapi/api.go:957](https://github.com/ethereum/go-ethereum/blob/58557cb4635d4e6f3e49fcdc82a6469554e929a6/internal/ethapi/api.go#L957) — RPCMarshalHeader() never writes totalDifficulty:
- Zero occurrences of totalDifficulty in all of internal/ethapi/
- Confirmed by Geth's own test assertions ([api_test.go:3094](https://github.com/ethereum/go-ethereum/blob/58557cb4635d4e6f3e49fcdc82a6469554e929a6/internal/ethapi/api_test.go#L3094)) — all three GetBlockByNumber variants have no totalDifficulty field

Nethermind tests (before fix) confirm the wrong-value case — blocks with "difficulty":"0xf4240" returning "totalDifficulty":"0x0".

svlachakis:
Fixes Closes Resolves #11096 

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No


smartprogrammer93 (approved):

LukaszRozmej (approved):

LukaszRozmej:
The only thing is that this could still potentially be needed for non-merge chains like Japan Open Chain or Energy Web Chain

svlachakis:
@LukaszRozmej I see.. I think 2 options then

1. We leave the fix as is and we are in parity with geth
2. Narrow the fix and restore `TotalDifficulty` only for `specProvider.MergeBlockNumber is null`, but drop the `?? UInt256.Zero` fallback so it only appears when actually loaded.

LukaszRozmej:
> @LukaszRozmej I see.. I think 2 options then
> 
> 1. We leave the fix as is and we are in parity with geth
> 2. Narrow the fix and restore `TotalDifficulty` only for `specProvider.MergeBlockNumber is null`, but drop the `?? UInt256.Zero` fallback so it only appears when actually loaded.

Maybe we can ensure it is loaded if the chain isn't merged

svlachakis:
@LukaszRozmej The only intentional divergence from Geth: non-merge chains still emit `totalDifficulty` when it's loaded, please check

LukaszRozmej (approved):

PR #11098, opened 2026-04-10
Allow `HistoryPruner` to be callable via decorator

damian-orzechowski:
Fixes Closes Resolves #
Extend `HistoryPruner` to be more configurable and change method to public to be callable if wrapped / decorated.

## Changes

- Make `SchedulePruneHistory` public (without `CancellationToken` argument, so caller can rely on proper cancellation of scheduled run)
- Promote background task timeout to config -> `PruningTimeoutSeconds`
- Check against genesis block number from `BlockTree` instead of hardcoded `0`

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks


damian-orzechowski:
@claude review

damian-orzechowski:
@claude re-review

damian-orzechowski:
@claude - regarding:
- point 1 - timeout defaults to 2s to ensure match the default value which was used before: `BackgroundTaskScheduler.DefaultTimeout`
- point 2 - change will cause this: "The annotation for nullable reference types should only be used in code within a '#nullable' annotations context"

src/Nethermind/Nethermind.History/IHistoryConfig.cs
```diff
@@ -25,6 +25,16 @@ public interface IHistoryConfig : IConfig
...
 
+    [ConfigItem(
+        Description = "Maximum time in seconds allowed for a single history pruning pass. Set to 0 to disable the timeout.",
+        DefaultValue = "2")]
+    uint PruningTimeoutSeconds { get; set; }
+
+    [ConfigItem(
+        Description = "The minimum block number that may be deleted during pruning. Block 0 is never deleted. Override when the chain genesis is not block 0.",
+        DefaultValue = "1")]
+    long MinDeletableBlockNumber { get; set; }
```
LukaszRozmej: Do we need it as separate config? Can we just use the same thing we use to determine which block is genesis?

src/Nethermind/Nethermind.History/HistoryPruner.cs
```diff
@@ -273,7 +282,7 @@ private void SchedulePruneHistory(CancellationToken cancellationToken)
                                     }
 
                                     return Task.CompletedTask;
-                                }, source: "HistoryPruner"))
+                                }, timeout: pruningTimeout, source: "HistoryPruner"))
```
LukaszRozmej, line 287: what happens on `pruningTimeout`? Maybe better to use custom time-outable `CancellationTokenSource` that will cancel gracefully?

src/Nethermind/Nethermind.History/IHistoryConfig.cs
```diff
@@ -25,6 +25,16 @@ public interface IHistoryConfig : IConfig
         DefaultValue = "8")]
     uint PruningInterval { get; set; }
 
+    [ConfigItem(
+        Description = "Maximum time in seconds allowed for a single history pruning pass. Set to 0 to disable the timeout.",
+        DefaultValue = "2")]
+    uint PruningTimeoutSeconds { get; set; }
```
LukaszRozmej, line 31: Why is it needed?

src/Nethermind/Nethermind.History/HistoryPruner.cs
damian-orzechowski: It already finishes gracefully on timeout, so I don't think it's needed.

src/Nethermind/Nethermind.History/IHistoryConfig.cs
damian-orzechowski: Default values of 2s seems to be tailored for mainnet with 12s slot, but for other chains with different block time, we might want to configure it differently. Additionally, I was thinking of adding a separate debug rpc call to trigger pruning manually and then again it might be useful.

src/Nethermind/Nethermind.History/IHistoryConfig.cs
damian-orzechowski: Yes, you're right - I changed it to use `_blockTree.Genesis`

damian-orzechowski:
@claude - re-review after latest changes

LukaszRozmej (approved):

src/Nethermind/Nethermind.History/IHistoryConfig.cs
LukaszRozmej: Before it didn't have timeout at all, right?
What happens on timeout? It stops pruning after X blocks, and resumes on next iteration?

damian-orzechowski: Timeout was not passed in to `IBackgroundScheduler`, but internally it has a default timeout of 2s. Default config value remains backward compatible. Pruning itself is run in a loop, each iteration checks on cancellation token and just stops gracefully after timeout. It has to be triggered again to continue (so by default on next empty queue).

LukaszRozmej (approved):

asdacap (approved):

PR #11112, opened 2026-04-12
fix(rpc): strip custom-error revert bytes from eth_call/eth_estimateGas message field   

### LINKED ISSUE #11095: eth_call / eth_estimateGas: Custom error revert data appended to message (should be in data only, as perGeth)
## Summary

When a contract reverts with a custom error, Nethermind Client adds the raw hex to the message field. This diverges from go-ethereum which keeps message as plain "execution reverted" and puts the bytes only in data.

Current behaviour

``` json
{
  "code": 3,
  "message": "execution reverted: 0xdeadbeef",
  "data": "0xdeadbeef"
}
```

Expected behaviour (Geth)

``` json
{
  "code": 3,
  "message": "execution reverted",
  "data": "0xdeadbeef"
}
```

For Error(string) as well as Panic reverts, appending the decoded text to message is correct. For custom errors, there is no human-readable string to decode just selector + data so message should stay as "execution reverted" and the raw bytes belong only in data.

## Root cause

*TransactionSubstate.cs* — EncodeErrorMessage is called as the fallback for unknown selectors and its result is written into Error, which flows into message:

``Error = TryGetErrorMessage(span) ?? EncodeErrorMessage(span);``
EncodeErrorMessage returns the hex string for non-printable bytes, which ends up in message .

### Evidence
Reproduced with raw bytecode that reverts with a 4-byte unknown selector:

``` bash
curl -X POST <node> -H "Content-Type: application/json" -d '{
  "jsonrpc":"2.0","method":"eth_call","id":1,
  "params":[{"data": "0x63deadbeef6000526004601cfd"}, "latest"]
}'
```

*go-ethereum newRevertError in [internal/ethapi/errors.go](https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/errors.go):*  
here when abi.UnpackRevert fails (custom error), err stays as plain vm.ErrExecutionReverted and only hexutil.Encode(revert) goes into data.

divi2806:
Closes #11095
                                                                   
  ## Changes      

  - `TransactionSubstate`: when revert data has an unknown selector
   (custom error), fall back
    to the `Revert` sentinel instead of hex-encoding raw bytes into
   `Error`        
  - `EthRpcModule.CreateResultWrapper`: emit plain `"execution
  reverted"` when `errorMessage`                                   
    is the sentinel or null; only append decoded reason for
  `Error(string)` and `Panic(uint256)` reverts                     
  - Add regression test
  `Eth_call_with_custom_error_revert_puts_bytes_only_in_data`      
  - Update existing tests that were asserting the pre-fix
  (incorrect) behaviour                                            
                  
  ## Types of changes                                              
                  
  - [ ] Optimization                                               
  - [ ] Refactoring
  - [ ] Documentation update
  - [ ] Build-related changes
  - [ ] Other: _Description_                                       
   
  ## Testing                                                       
                  
  #### Requires testing

  - [x] Yes
  - [ ] No

  - [x] Yes
  - [ ] No
                                                                   
  #### Notes on testing
                                                                   
  dotnet test --project                                            
  src/Nethermind/Nethermind.Evm.Test/Nethermind.Evm.Test.csproj -c
  release --filter "FullyQualifiedName~TransactionSubstateTests"   
                  
  dotnet test --project src/Nethermind/Nethermind.JsonRpc.Test/Neth
  ermind.JsonRpc.Test.csproj -c release --filter
  "FullyQualifiedName~EthRpcModuleTests"                           
                  
  ## Documentation

  - [ ] Yes
  - [x] No

  - [ ] Yes                                                        
  - [x] No        

  ## Remarks

  Aligns `eth_call` / `eth_estimateGas` error responses with       
  go-ethereum's `newRevertError`
  (`internal/ethapi/errors.go`): for `Error(string)` and           
  `Panic(uint256)` the decoded reason                              
  is appended to `message`; for all other revert data (custom
  errors, unknown selectors)                                       
  `message` stays as plain `"execution reverted"` and the raw bytes
   appear only in `data`.


divi2806:
@svlachakis @LukaszRozmej  @benaadams @smartprogrammer93  need your inputs in this

nethermind.sln
svlachakis, line 1: Revert this please and we will review the rest

svlachakis (changes requested):

nethermind.sln
divi2806: done, thankyou

src/Nethermind/Nethermind.Evm/TransactionSubstate.cs
DarkLord017: Geth directly returns null if it does not match Panic or Error(string):
https://github.com/ethereum/go-ethereum/blob/master/accounts/abi/abi.go#L278

I think there can be some cases where it might seem to match this fallback, so we can just return null if it matches neither of the two function selectors

DarkLord017 (changes requested):

src/Nethermind/Nethermind.Evm/TransactionSubstate.cs
divi2806: Done, you are correct.

I have removed the generic fallback block. `GetErrorMessage` now 
  returns `null` for any selector that isn't `Error(string)` or 
  `Panic(uint256)`, matching Geth's `UnpackRevert` `default:`      
  branch exactly.

DarkLord017 (approved):

LukaszRozmej (commented):
Approach looks correct and matches go-ethereum's `newRevertError` / `abi.UnpackRevert` semantics. Tests are well-structured and the comments explaining each case are helpful. A few notes:

### Behavior change in `NativeCallTracer` — please verify

`NativeCallTracer.ProcessRevertReason` (`Nethermind.Blockchain/Tracing/GethStyle/Custom/Native/Call/NativeCallTracer.cs:258`) calls `TransactionSubstate.GetErrorMessage` directly. Before this PR, the generic ABI fallback in `GetErrorMessage` would attempt to decode arbitrary revert payloads as an ABI-encoded `(offset, length, bytes)` tuple — so custom errors whose body happens to match that shape (e.g. `FailedOp(uint256, string)`) would surface a decoded string as `callFrame.RevertReason`. After this PR `GetErrorMessage` returns `null` for unknown selectors, so `RevertReason` will now be `null` for those cases in Geth-style call traces.

That's likely the right thing (it aligns with Geth's native call tracer, which also only decodes `Error(string)` / `Panic(uint256)`), but it's a user-visible trace output change not mentioned in the PR description. Worth confirming against Geth and calling out in release notes if the PR moves forward.

### Sentinel-string collision

The fix hinges on distinguishing "decoded human-readable message" from "could not decode" by comparing `errorMessage` against the `TransactionSubstate.Revert` constant (the literal string `"revert"`). A legitimate `require(false, "revert")` would produce `Error(string)` decoded to `"revert"` — indistinguishable from the sentinel, so the message would degrade to plain `"execution reverted"` instead of `"execution reverted: revert"`. The raw data still ends up in `data`, so nothing is lost, but the divergence from Geth exists for this exact edge case.

Given `SimulateTxTracer.cs:113` already uses the same pattern, this is pre-existing convention — just flagging. A cleaner long-term fix would be a separate "was decoded" signal (bool/enum) rather than an in-band sentinel, but that's out of scope here.

### Minor nit — `TransactionSubstate.cs:103`

```csharp
Error = Revert;               // line 94
...
Error = TryGetErrorMessage(span) ?? Revert;   // line 103
```

Since `Error` is already `Revert` at line 94, the `?? Revert` on line 103 is a no-op. Could be:
```csharp
Error = TryGetErrorMessage(span) ?? Error;
```
or clearer:
```csharp
if (TryGetErrorMessage(span) is { } decoded) Error = decoded;
```
Purely cosmetic.

### Also worth checking

- `EncodeErrorMessage` is now only reached from the `NativeCallTracer` `catch` block (which shouldn't fire for unknown selectors anymore since they return `null` rather than throw). If nothing else exercises it, consider whether it still needs to be `public`.
- Intermediate commit `820a1a4` ("removed nethermind.sln file") is spurious to this change — ideally squash or rebase it out before merge.

svlachakis:
@divi2806 please address @LukaszRozmej comment

divi2806:
@LukaszRozmej @svlachakis 

Thanks for the thorough review addressed all five points:

---

**1. NativeCallTracer / ProcessRevertReason**
Confirmed. After this change, `GetErrorMessage` returns `null` for custom-error selectors, so `NativeCallTracer.RevertReason` will be `null` for those cases. `ValidateRevertReason(null)` already returns `null`, so the tracer output is unchanged structurally  and `null` is the correct value here, matching how Geth handles unknown selectors in its native call tracer.

**2. Sentinel collision (`require(false, "revert")`)**
Acknowledged  this is a pre-existing limitation that predates this PR. The sentinel pattern (`"revert"` as an in-band signal) is also used in `SimulateTxTracer` and other places in the codebase. Fixing it properly would require a broader refactor (e.g., a discriminated union result type) and is out of scope for this bug fix.

**3. Nit — `Error = TryGetErrorMessage(span) ?? Revert`**
Applied. Changed to `if (TryGetErrorMessage(span) is { } decoded) Error = decoded;` — cleaner and avoids the redundant `?? Revert` since `Error` is already set to `Revert` on line 94.

**4. `EncodeErrorMessage` visibility**
Left as `public`. `NativeCallTracer` (in `Nethermind.Blockchain`) calls it directly in its catch block. Making it `internal` would break cross-assembly access. An `[InternalsVisibleTo]` attribute could work but adds more noise than the `public` modifier here.

**5. Spurious `nethermind.sln` commit**
Squashed. All three commits have been combined into one clean commit (`792654a1a`) and force-pushed.

**Should we create a new issue for the 2nd Point?**

newmanifold:
In the RPC layer maybe we can check whether the revert data prefix is Error or Panic from the raw bytes (bodyData), instead of checking errorMessage which has a sentinel collision.

divi2806:
Good catch. Checking the raw byte prefix at the RPC layer avoids the sentinel entirely — we look at `executionRevertedReason[..4]` against `ErrorFunctionSelector`/`PanicFunctionSelector` rather than inspecting `errorMessage`'s string value. This also cleanly fixes the `require(false, "revert")` edge case. Will implement.

divi2806:
Implemented in the latest commit - 
`CreateResultWrapper` now checks `executionRevertedReason[:4]` directly against `ErrorFunctionSelector` and `PanicFunctionSelector` to decide whether to append the decoded message. This removes any reliance on the sentinel string, so `require(false, "revert")` is handled correctly — the raw bytes start with the `Error(string)` selector → decoded message is appended → `"execution reverted: revert"`.

divi2806:
@svlachakis any changes you requested? can't seem to find them

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.TransactionExecutor.cs
```diff
@@ -84,13 +84,23 @@ protected ResultWrapper<TResult> CreateResultWrapper(bool inputError, string? er
                 {
                     if (executionReverted)
                     {
+                        // Check the raw revert bytes to determine if the selector is Error(string) or Panic(uint256).
+                        // Using the bytes rather than errorMessage avoids the sentinel-string collision
+                        // (e.g. require(false, "revert") would be misidentified as the Revert sentinel).
+                        bool isKnownRevertType = executionRevertedReason is { Length: >= 4 } &&
+                            (executionRevertedReason.AsSpan(0, 4).SequenceEqual(TransactionSubstate.ErrorFunctionSelector) ||
+                             executionRevertedReason.AsSpan(0, 4).SequenceEqual(TransactionSubstate.PanicFunctionSelector));
```
newmanifold, line 92: Probably a test using `RevertWithSolidityErrorEncoding` is also needed to assert this path of the code and that string collision doesn't happen

src/Nethermind/Nethermind.JsonRpc.Test/Modules/Eth/EthRpcModuleTests.EthCall.cs
```diff
@@ -333,6 +333,29 @@ public async Task Eth_call_with_revert()
...
+    [Test]
+    public async Task Eth_call_with_custom_error_revert_puts_bytes_only_in_data()
+    {
+        // When a contract reverts with a custom error (unknown 4-byte selector), message must be
+        // plain "execution reverted" and the raw bytes must appear only in data — matching Geth.
+        // See: https://github.com/NethermindEth/nethermind/issues/11095
+        using Context ctx = await Context.CreateWithLondonEnabled();
+
+        // keccak4("ActionFailed()") = 0x080a1c27
+        byte[] selector = [0x08, 0x0a, 0x1c, 0x27];
```
newmanifold:
```suggestion
        byte[] selector = Keccak.Compute("ActionFailed()").Bytes[..4].ToArray();
```
This should be self documenting

divi2806: Removed the hardcoded bytes and comment; now using `Keccak.Compute("ActionFailed()").Bytes[..4]` directly.

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.TransactionExecutor.cs
divi2806: Added `Eth_call_with_revert_sentinel_string_as_message_still_appends_decoded_message` — it uses `RevertWithSolidityErrorEncoding("revert")` to simulate `require(false, "revert")` and asserts the response is `"execution reverted: revert"`, proving the raw-byte prefix check handles the sentinel collision correctly.

svlachakis:
@divi2806 please run locally all the tests and make sure they pass

divi2806:
@svlachakis any feedbacks?

svlachakis (approved):

divi2806:
@LukaszRozmej any feedbacks?

LukaszRozmej (approved):

PR #11118, opened 2026-04-13
ci: make EVM benchmark iteration counts configurable

asdacap:
## Summary
- Add env var overrides (`BENCHMARK_LAUNCH_COUNT`, `BENCHMARK_WARMUP_COUNT`, `BENCHMARK_ITERATION_COUNT`) to `EvmOpcodesBenchmarkConfig` with existing values as defaults
- Set CI defaults to 2/5/10 (from 3/15/15) to reduce benchmark runtime

## Benchmark timing results

| Config | PR side | Estimated total (once merged) |
|---|---|---|
| 3/15/15 (current master) | ~24 min | ~48 min |
| 2/5/10 (this PR) | ~17 min | ~34 min |
| 1/1/1 (floor / fixed overhead) | ~12 min | ~24 min |

~30% reduction in benchmark CI time with 2/5/10 config.

## Test plan
- [x] EVM benchmark CI runs successfully with 2/5/10 config
- [x] EVM benchmark CI runs successfully with 1/1/1 config
- [x] Verified env vars are picked up by PR side (base side uses master code until merged)

### DIFF
--- a/.github/workflows/evm-opcode-benchmark-diff.yml
+++ b/.github/workflows/evm-opcode-benchmark-diff.yml
@@ -83,9 +83,9 @@ jobs:
       # Extra reruns for opcodes that are both flagged and noisy in the first pass.
       NOISY_RERUNS: "2"
       # BenchmarkDotNet iteration tuning (defaults in code: 3/15/15).
-      BENCHMARK_LAUNCH_COUNT: "2"
-      BENCHMARK_WARMUP_COUNT: "5"
-      BENCHMARK_ITERATION_COUNT: "10"
+      BENCHMARK_LAUNCH_COUNT: "1"
+      BENCHMARK_WARMUP_COUNT: "1"
+      BENCHMARK_ITERATION_COUNT: "1"
     steps:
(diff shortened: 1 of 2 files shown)


MarekM25 (approved):

smartprogrammer93 (changes requested):
this increases the CV significantly making the results useless. it is okay for the benchmark to run longer.

asdacap:
What is CV?

asdacap:
Its 50 minute per PR change. Not just for the initial change. Cant this be made parallel at least?

smartprogrammer93:
@asdacap it is the coefficient of variance. this is why after this PR, you see the benchmarks showing a lot of regressions and improvements here: https://github.com/NethermindEth/nethermind/pull/11118#issuecomment-4234537326
even though the PR didnt touch any implementation code.

Not sure about parallelizing. maybe it could be done but we would need to utilize more runners and then combine the results from multiple runners into one place and post the comment. We cant run the benchmarks in parallel on the same runner because they would affect the results of each other

PR #11121, opened 2026-04-13
eth_getBlockByNumber("pending") returns non-null hash, nonce, miner

### LINKED ISSUE #11120: eth_getBlockByNumber("pending") returns non-null hash, nonce, miner - diverges from Geth
### Summary

When `eth_getBlockByNumber` is called with "pending", Nethermind returns the block with real values for hash, nonce, and miner. Geth explicitly nulls these three fields for pending blocks.

### Expected (Geth behaviour)

{
  "hash": null,
  "nonce": null,
  "miner": null,
  ...
}

### Actual (Nethermind)

{
  "hash": "0x29f141925d2d8e357ae5b6040c97aa12d7ac6dfcbe2b20e7b616d8907ac8e1f3",
  "nonce": "0x0000000000000000",
  "miner": "0x475674cb523a0a2736b7f7534390288fce16982c",
  ...
}

### Geth reference

https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/api.go — GetBlockByNumber and GetHeaderByNumber:

```
if number == rpc.PendingBlockNumber {
    // Pending blocks need to nil out a few fields
    for _, field := range []string{"hash", "nonce", "miner"} {
        response[field] = nil
    }
}
```

The same nulling applies to `eth_getHeaderByNumber` in Geth.

### Affected methods

- eth_getBlockByNumber with "pending"
- eth_getHeaderByNumber with "pending"

### Root cause

`BlockForRpc.Hash`, `BlockForRpc.Miner`, and `BlockForRpc.Nonce` are non-nullable (or their JsonIgnore attribute was WhenWritingNull, causing them to be omitted rather than serialised as null). The `GetBlock()` method in `EthRpcModule` did not null these fields out for pending block requests.

svlachakis:
Fixes Closes Resolves #11120 

## Changes

- Make `Hash`, `Miner`, `Nonce` nullable so they serialise as `null` rather than being omitted. 
- In `EthRpcModule.GetBlock()`, after constructing `BlockForRpc`, set `Hash = null`, `Nonce = null`, `Miner = null` when `blockParameter.Type == BlockParameterType.Pending`.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No


src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
```diff
@@ -434,9 +434,20 @@ public ResultWrapper<BlockForRpc> eth_getBlockByNumber(BlockParameter blockParam
...
-        return ResultWrapper<BlockForRpc?>.Success(block is null
-            ? null
-            : new BlockForRpc(block, returnFullTransactionObjects, _specProvider));
+        if (block is null)
+        {
+            return ResultWrapper<BlockForRpc?>.Success(null);
+        }
+
+        BlockForRpc blockForRpc = new BlockForRpc(block, returnFullTransactionObjects, _specProvider);
+        if (blockParameter.Type == BlockParameterType.Pending)
```
svlachakis, line 443: could move that to a constructor but I'm not sure whether `BlockForRpc` need to know about pending semantics.

LukaszRozmej (approved):

smartprogrammer93 (approved):

PR #11123, opened 2026-04-13
refactor: remove long range catch-up sync

asdacap:
## Summary
- Remove long range catch-up sync from `MultiSyncModeSelector`. Once a node completes initial state sync and enters full sync, it stays in full sync regardless of how far behind it falls — replaying blocks is faster than re-syncing state via snap.
- Deprecate `FastSyncCatchUpHeightDelta` config property (marked `[Obsolete]`, kept for backward compatibility with existing user configs).
- Remove `FastSyncCatchUpHeightDelta` from all 65 shipped chain config JSON files and `superchain.py`.

## Changes
- `MultiSyncModeSelector`: removed `longRangeCatchUp`, `IsInAStickyFullSyncMode`, `LastBlockThatEnabledFullSync`, and simplified `ShouldBeInFastSyncMode`/`ShouldBeInStateSyncMode` to only enter fast/state sync when state has never been downloaded (`stateNotDownloadedYet`).
- Updated sync mode selector tests to expect `Full` sync instead of `StateNodes` for nodes that have been offline for a long time.
- Removed tests that specifically validated catch-up entry behavior.

## Test plan
- [x] Build succeeds with 0 warnings and 0 errors
- [x] All 436 `MultiSyncModeSelector` tests pass
- [x] `ConfigFilesTests` pass (1 pre-existing flaky test on master)

### DIFF
--- a/src/Nethermind/Nethermind.Runner/configs/hoodi.json
+++ b/src/Nethermind/Nethermind.Runner/configs/hoodi.json
@@ -11,8 +11,7 @@
   },
   "Sync": {
     "FastSync": true,
-    "SnapSync": true,
-    "FastSyncCatchUpHeightDelta": 10000000000
+    "SnapSync": true
   },
   "Metrics": {
     "NodeName": "Hoodi"
@@ -27,13 +26,18 @@
     "Port": 8545,
     "EngineHost": "127.0.0.1",
     "EnginePort": 8551,
-    "EngineEnabledModules": [ "net", "eth", "subscribe", "engine", "web3", "client", "flashbots" ]
+    "EngineEnabledModules": [
+      "net",
+      "eth",
+      "subscribe",
+      "engine",
+      "web3",
+      "client",
+      "flashbots"
+    ]
   },
   "Merge": {
     "Enabled": true
-  },
-  "Discovery": {
-    "DiscoveryVersion": "All"
   },
   "Discovery": {
     "Bootnodes": "enode://2112dd3839dd752813d4df7f40936f06829fc54c0e051a93967c26e5f5d27d99d886b57b4ffcc3c475e930ec9e79c56ef1dbb7d86ca5ee83a9d2ccf36e5c240c@134.209.138.84:30303,enode://60203fcb3524e07c5df60a14ae1c9c5b24023ea5d47463dfae051d2c9f3219f309657537576090ca0ae641f73d419f53d8e8000d7a464319d4784acd7d2abc41@209.38.124.160:30303,enode://8ae4a48101b2299597341263da0deb47cc38aa4d3ef4b7430b897d49bfa10eb1ccfe1655679b1ed46928ef177fbf21b86837bd724400196c508427a6f41602cd@134.199.184.23:30303"
--- a/src/Nethermind/Nethermind.Runner/configs/hoodi.json
+++ b/src/Nethermind/Nethermind.Runner/configs/hoodi.json
@@ -11,7 +11,8 @@
   },
   "Sync": {
     "FastSync": true,
-    "SnapSync": true
+    "SnapSync": true,
+    "FastSyncCatchUpHeightDelta": 10000000000
   },
   "Metrics": {
     "NodeName": "Hoodi"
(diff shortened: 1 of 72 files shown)


src/Nethermind/Nethermind.Runner/configs/hoodi.json
```diff
@@ -27,14 +26,19 @@
...
+      "engine",
+      "web3",
+      "client",
+      "flashbots"
+    ]
   },
   "Merge": {
     "Enabled": true
   },
-  "Discovery": {
```
asdacap, line 35: Add back

src/Nethermind/Nethermind.Synchronization.Test/ParallelSync/MultiSyncModeSelectorTests.Scenario.cs
```diff
@@ -252,8 +250,8 @@ public ScenarioBuilder IfThisNodeIsProcessingAlreadyDownloadedBlocksInFullSync()
                         {
                             SyncProgressResolver.FindBestHeader().Returns(ChainHead.Number);
                             SyncProgressResolver.FindBestFullBlock().Returns(ChainHead.Number);
-                            SyncProgressResolver.FindBestFullState().Returns(ChainHead.Number - FastSyncCatchUpHeightDelta + 1);
-                            SyncProgressResolver.FindBestProcessedBlock().Returns(ChainHead.Number - FastSyncCatchUpHeightDelta + 1);
+                            SyncProgressResolver.FindBestFullState().Returns(ChainHead.Number - 1023);
```
asdacap: Whats with the magic number?

src/Nethermind/Nethermind.Synchronization/ParallelSync/MultiSyncModeSelector.cs
```diff
@@ -230,12 +222,6 @@ public void Update()
                         }
                     }
 
-                    if ((newModes & (SyncMode.Full | SyncMode.WaitingForBlock)) != SyncMode.None
```
asdacap, line 230: Keep

src/Nethermind/Nethermind.Synchronization/ParallelSync/MultiSyncModeSelector.cs
```diff
@@ -398,20 +376,14 @@ private bool ShouldBeInFastSyncMode(Snapshot best)
...
-            bool longRangeCatchUp = best.TargetBlock - best.State >= FastSyncCatchUpHeightDelta;
             bool stateNotDownloadedYet = !best.StateDownloaded;
             bool notNeedToWaitForHeaders = NotNeedToWaitForHeaders;
 
             bool result = notInUpdatingPivot &&
                           notInBeaconModes &&
                           postPivotPeerAvailable &&
-                          // (catch up after node is off for a while
-                          // OR standard fast sync)
-                          notInAStickyFullSync &&
```
asdacap, line 406: Explain

src/Nethermind/Nethermind.Runner/configs/hoodi.json
asdacap: [Claude] Already reverted in f70d956.

src/Nethermind/Nethermind.Synchronization.Test/ParallelSync/MultiSyncModeSelectorTests.Scenario.cs
asdacap: [Claude] Replaced with `Pivot.Number` — the scenario just needs state to be at the pivot point (past initial sync threshold). The old value was `FastSyncCatchUpHeightDelta - 1` which was an artifact of the removed constant.

src/Nethermind/Nethermind.Synchronization/ParallelSync/MultiSyncModeSelector.cs
asdacap: [Claude] Removed. `LastBlockThatEnabledFullSync` was only consumed by `IsInAStickyFullSyncMode` which gated the long range catch-up. With catch-up removed, there's no reader for this value — keeping it would be dead code.

src/Nethermind/Nethermind.Synchronization/ParallelSync/MultiSyncModeSelector.cs
asdacap: [Claude] Previously, `ShouldBeInFastSyncMode` had two paths to return true: (1) initial sync (`stateNotDownloadedYet`) and (2) catch-up when node falls far behind (`longRangeCatchUp`). The `notInAStickyFullSync` guard prevented thrashing between full and fast sync when the gap was small.

With long range catch-up removed, only path (1) remains — once state is downloaded, the node never re-enters fast/state sync. This makes both `longRangeCatchUp` and `notInAStickyFullSync` dead logic. Added a comment in the code explaining this.

src/Nethermind/Nethermind.Synchronization.Test/ParallelSync/MultiSyncModeSelectorTests.Scenario.cs
asdacap: You know, you make this even more confusing.

asdacap: [Claude] Replaced with a named constant `LargeDistanceBehindHead = 1024` and used as `ChainHead.Number - LargeDistanceBehindHead` so the intent is clear.

flcl42:
Need to regenerate optimism configs probably

flcl42 (approved):

kamilchodola:
@flcl42 so maybe optimism adjustments should happen in this PR also? not to push anything which will broke any network?

asdacap:
Why need to update optimism?

LukaszRozmej:
@claude review

LukaszRozmej (approved):

PR #10788, opened 2026-03-11
Improve ssz

flcl42:
## Changes

Extends the SSZ source generator with EIP-7495/7916 type classes and fixes bugs in the old union implementation.

## Changes

- **EIP-7916 compatible unions** — `[SszCompatibleUnion]` replaces the fragile `Selector`-property name heuristic. Generator validates at build time: `Selector` must be a byte-backed enum with values in `[1, 127]`, exact 1:1 enum-to-property mapping, and pairwise type-compatibility across all variants. Removes the `None` (selector = 0) special case which the spec forbids.

- **Progressive lists / bitlists** — `[SszProgressiveList]` and `[SszProgressiveBitlist]` route merkleization through `Merkle.MerkleizeProgressive` instead of the standard `Merkle.Merkleize`, keeping generalized indices stable as the list grows.

- **`Merkle` helpers** — added `MixInActiveFields(ref root, activeFields)` and `MerkleizeProgressive(out root, chunks, numLeaves)` to `Nethermind.Merkleization`.

- **`Kind` enum** — added `ProgressiveContainer`, `ProgressiveList`, `ProgressiveBitList`, `CompatibleUnion`; renamed `Union` → `CompatibleUnion`; renumbered flags to make room.

```csharp
[SszContainer]
public struct BeaconBlockBody
{
    [SszField(0)] public ulong Slot { get; set; }
    [SszField(2)] public ulong ProposerIndex { get; set; }  // index 1 reserved for future field
}

[SszCompatibleUnion]
public struct Transaction
{
    public TransactionSelector Selector { get; set; }  // byte enum, values in [1, 127]
    public LegacyTx LegacyTx { get; set; }
    public Eip1559Tx Eip1559Tx { get; set; }
}
```

#### What types of changes does your code introduce?

- [x] Bugfix (a non-breaking change that fixes an issue)
- [x] New feature (a non-breaking change that adds functionality)
- [ ] Breaking change (a change that causes existing functionality not to work as expected)
- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

#### If yes, did you write tests?

- [x] Yes
- [ ] No

#### Notes on testing

Round-trip encode/decode and merkleization tests for progressive containers (including non-contiguous field indices), progressive lists, progressive bitlists, and compatible unions.

#### Requires documentation update

- [ ] Yes
- [x] No

[...]

PR #11125, opened 2026-04-13
ssz: ISszCodec<T>, test simplification, read-only collection surface

LukaszRozmej:
Targets: #10788

### `ISszCodec<T>` static-abstract interface
- Added `Nethermind.Serialization.Ssz.ISszCodec<T>` with 12 static abstract members mirroring the full per-type `SszEncoding` surface:
  - Length: `GetLength(T)`, `GetLength(IReadOnlyCollection<T>?)`.
  - Decode: `Decode(ReadOnlySpan<byte>, out T)`, `Decode(ReadOnlySpan<byte>, out T[])`.
  - Encode: `Encode(T) → byte[]`, `Encode(Span<byte>, T)`, `Encode(IReadOnlyCollection<T>?) → byte[]`, `Encode(Span<byte>, IReadOnlyCollection<T>?)`.
  - Merkleize: `Merkleize(T, out UInt256)`, `MerkleizeVector(IReadOnlyList<T>?, out UInt256)`, `MerkleizeList(IReadOnlyList<T>?, ulong limit, out UInt256)`, `MerkleizeProgressiveList(IReadOnlyList<T>?, out UInt256)`.
- Lets generic code call `T.Decode(...)` / `T.Encode(...)` / etc. on a type parameter without per-type delegate plumbing.
- `SszGenerator` now emits a second source per `[SszContainer]`/`[SszCompatibleUnion]` type: `Serialization.SszEncoding.{Type}.Codec.cs`, a `partial struct/class T : ISszCodec<T>` with all 12 explicit-interface-impl forwarders. Explicit impl keeps the type's public surface clean.
- All existing SSZ container/union types made `partial`:
  - `Ethereum.Ssz.Test/SszTestContainers.cs` (8 structs)
  - `Nethermind.Shutter/SlotDecryptionIdentities.cs` (2 structs)
  - `Nethermind.Serialization.SszGenerator.Test/SszTypes.cs` (12 structs + 1 union)

### Read-only collection parameters
- Generator-emitted collection overloads (`GetLength` / `Encode`) now take `IReadOnlyCollection<T>?` instead of `ICollection<T>?`.
- `MerkleizeVector` / `MerkleizeList` / `MerkleizeProgressiveList` now take `IReadOnlyList<T>?` instead of `IList<T>?`.
- The generated bodies only read `.Count`, iterate, and index — they never mutated the inputs. Arrays, `List<T>`, and any read-only list caller keeps working; callers holding an `IList<T>` variable need a one-line cast.

### `SszContainerTests` refactor
- Collapsed `ValidHandlers` + `InvalidHandlers` (two dictionaries, each with eight 3-line lambda registrations) into a single `Handlers` dictionary keyed to a tiny `IContainerHandler` + `ContainerHandler<T> where T : ISszCodec<T>`.
- Registration uses a `HandlerMap` builder with `.Add<T>()`, so each container type appears in exactly one place: `new HandlerMap().Add<X>().Add<Y>()...`.

### `EncodingTest` invalid-input cases
- Folded seven separate `[Test]` methods, each asserting `Throws<InvalidDataException>`, into one parameterized `[TestCaseSource(nameof(InvalidInputCases))]` test with named `TestCaseData` rows.
- Added a new smoke test (`ISszCodec_constraint_is_usable_from_generic_code`) that invokes `T.GetLength(value)` and `T.Encode(Span<byte>, T)` through a `where T : ISszCodec<T>` helper — proves the interface compiles from a real generic caller.

### Review fixes
- `SszContainerTests` invalid-case assertion tightened from `Throws<Exception>` to `Throws<InvalidDataException>`, matching `EncodingTest`.
- New `SSZ004` diagnostic in `SszGenerator`: SSZ container types nested inside another type are rejected with a clear compile error (the generated file would otherwise silently overwrite the nested declaration). Registered in `AnalyzerReleases.Unshipped.md`.
- Zero-alloc `Encode(Span<byte> data, T value)` overload on `ISszCodec<T>` so production callers don't have to allocate.

### Generator cleanups
- `MerkleizeFeedStatement`: removed dead `Kind.ProgressiveBitList` switch arm — was identical to the default case.
- Generated container `Decode`: removed a redundant `InvalidDataException` length check that ran immediately before `ValidateSszMinimumLength`/`ValidateSszExactLength` (which throw with a better message). Also moved `container = new()` to after validation.

## Types of changes

- [x] Refactoring

#### Requires testing

- [x] Yes

- [x] Yes — added one generic-consumer smoke test; the rest is covered by existing tests.

#### Notes on testing

- `Ethereum.Ssz.Test` (`SszContainerTests` filter): 701/701 pass
- `Nethermind.Serialization.SszGenerator.Test`: 20/20 pass
- `Nethermind.Shutter.Test`: 35/35 pass

- [x] No


LukaszRozmej:
@claude review

LukaszRozmej:
@claude re-review

LukaszRozmej:
@claude re-review

flcl42:
Having such interface is a great thing. Under the hood it calls SszEncoding, what if we make it have the implementation right in those interface methods instead of calling SszEncoding? Right now I either call T.Encode or SszEncoding.Encode - two ways doing the same

LukaszRozmej:
> Having such interface is a great thing. Under the hood it calls SszEncoding, what if we make it have the implementation right in those interface methods instead of calling SszEncoding? Right now I either call T.Encode or SszEncoding.Encode - two ways doing the same

I'm fine with that, just wanted minimal changes, feel free to iterate on it more

PR #11133, opened 2026-04-14
fix: clean stale flat storage when contract storage becomes empty

asdacap:
## Summary

Fixes edge case where a contract's storage becomes empty during sync (e.g. all slots deleted between snap sync and healing phase). The stale flat storage entries from the initial download are not cleaned up, causing invalid blocks when block processing reads stale data while the trie says storage is empty. The flat trie verifier also fails to detect this, giving a false pass.

## Root cause

During snap sync, a contract's storage is downloaded. Between snap sync and the healing phase, the contract's storage root changes to `EmptyTreeHash` (all slots were deleted on-chain). During healing:
- TreeSync sees the account leaf with `EmptyTreeHash` storage root and skips storage sync (no nodes to download)
- FlatTreeSyncStore never gets a chance to delete the stale flat storage entries from the initial snap download
- The flat trie verifier skips verification for empty-root accounts entirely, so the corruption is not detected even when running verify trie

## Changes

- **ITreeSyncStore.EnsureStorageEmpty**: New method to delete all flat storage + trie nodes for an address (no-op for Patricia)
- **TreeSync.HandleTrieNode**: Call `EnsureStorageEmpty` when account leaf has empty storage root
- **TreeSync.VerifyStorageUpdated**: Filter `EmptyTreeHash` and call `EnsureStorageEmpty` for empty-root accounts detected during snap sync refresh
- **FlatTrieVerifier**: Verify storage for all accounts including empty-root ones, detecting orphaned flat entries. Add `VerifyEmptyStorage` for preimage mode compatibility

## Test plan

- [x] Unit test: `FlatTreeSyncStore.EnsureStorageEmpty` deletes storage entries correctly
- [x] Unit test: `FlatTrieVerifier` detects orphaned storage on empty-root accounts
- [x] E2E sync tests pass (32 tests, 0 failures)
- [x] Mainned Snap Synced with Flat

### DIFF
--- a/src/Nethermind/Nethermind.Synchronization/FastSync/TreeSync.cs
+++ b/src/Nethermind/Nethermind.Synchronization/FastSync/TreeSync.cs
@@ -752,10 +752,16 @@ namespace Nethermind.Synchronization.FastSync
             {
                 Account? account = verificationContext.GetAccount(updatedAddress);
 
-                if (account?.StorageRoot is not null
+                if (account?.StorageRoot is not null && account.StorageRoot != Keccak.EmptyTreeHash
                     && AddNodeToPending(new StateSyncItem(account.StorageRoot, updatedAddress, TreePath.Empty, NodeDataType.Storage), dependentItem, "incomplete storage") == AddNodeResult.Added)
                 {
                     if (_logger.IsDebug) _logger.Debug($"Storage {updatedAddress} missing correct storage root {account.StorageRoot}");
+                }
+                else if (account?.StorageRoot == Keccak.EmptyTreeHash)
+                {
+                    if (_logger.IsDebug) _logger.Debug($"Storage {updatedAddress} is empty, ensuring flat storage cleared");
+                    _store.EnsureStorageEmpty(updatedAddress);
+                    dependentItem.Counter--;
                 }
                 else
                 {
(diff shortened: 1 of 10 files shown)


benaadams (approved):

LukaszRozmej:
@claude review

src/Nethermind/Nethermind.Synchronization/FastSync/TreeSync.cs
```diff
@@ -752,11 +752,17 @@ private bool VerifyStorageUpdated(StateSyncItem item, byte[] value)
             {
                 Account? account = verificationContext.GetAccount(updatedAddress);
 
-                if (account?.StorageRoot is not null
+                if (account?.StorageRoot is not null && account.StorageRoot != Keccak.EmptyTreeHash
                     && AddNodeToPending(new StateSyncItem(account.StorageRoot, updatedAddress, TreePath.Empty, NodeDataType.Storage), dependentItem, "incomplete storage") == AddNodeResult.Added)
                 {
                     if (_logger.IsDebug) _logger.Debug($"Storage {updatedAddress} missing correct storage root {account.StorageRoot}");
                 }
+                else if (account?.StorageRoot == Keccak.EmptyTreeHash)
```
LukaszRozmej: put this branch first, then the second branch can drop the `account.StorageRoot != Keccak.EmptyTreeHash` part

LukaszRozmej (approved):

PR #11137, opened 2026-04-14
XDC Subnet penalties

batrr:
https://github.com/XinFinOrg/XDC-Subnet/blob/82d26c17701d24cbcfb9bcf9e4916bfb77483959/eth/hooks/engine_v2_hooks.go#L26

## Changes                                                                                                                                                                                                                                  
                                                                                      
  - Add `SubnetPenaltyHandler` for XDC subnet penalty calculation
  - Add unit tests for `SubnetPenaltyHandler                                                                                                                                                                                                         

  ## Types of changes                                                                                                                                                                                                                         
                                                                              
  ## Testing                                                                                                                                                                                                                                  
                                                                              
  #### Requires testing

  - [x] Yes

  - [x] Yes


src/Nethermind/Nethermind.Xdc/SubnetPenaltyHandler.cs
```diff
@@ -0,0 +1,113 @@
...
+using Nethermind.Core;
+using Nethermind.Core.Crypto;
+using Nethermind.Core.Specs;
+using Nethermind.Crypto;
+using Nethermind.Serialization.Rlp;
+using Nethermind.Xdc.Spec;
+
+namespace Nethermind.Xdc;
+
+internal class SubnetPenaltyHandler(IBlockTree tree, ISpecProvider specProvider, IEpochSwitchManager epochSwitchManager, ISigningTxCache signingTxCache) : IPenaltyHandler
```
alexb5dh, line 14: Is it supposed to be used somewhere?

src/Nethermind/Nethermind.Xdc/SubnetPenaltyHandler.cs
```diff
@@ -0,0 +1,113 @@
...
+                foreach (Address penalty in parentHeader.PenaltiesAddress)
+                {
+                    penalties.Add(penalty);
+                }
+            }
+
+            listBlockHash.Add(parentHash);
+            listBlockNumber.Add(parentNumber);
+
+            Address miner = parentHeader.Beneficiary ?? _ethereumEcdsa.RecoverAddress(new Signature(parentHeader.Validator.AsSpan(0, 64), parentHeader.Validator[64]), Keccak.Compute(_xdcHeaderDecoder.Encode(parentHeader, RlpBehaviors.ForSealing).Bytes));
```
alexb5dh: Go version uses `coinbase` field only, not sure if it matters much:
https://github.com/XinFinOrg/XDC-Subnet/blob/82d26c17701d24cbcfb9bcf9e4916bfb77483959/eth/hooks/engine_v2_hooks.go#L70

alexb5dh (approved):

src/Nethermind/Nethermind.Xdc/SubnetPenaltyHandler.cs
```diff
@@ -0,0 +1,113 @@
...
+
+                if (blockHashes.Contains(signedBlockHash))
+                    penalties.Remove(fromSigner);
+            }
+        }
+        // Must use EIP-55 checksummed hex to match XDC Go node ordering (addr.Hex()).
+        // Plain lowercase comparison gives wrong order: e.g. "0xAb..." < "0xaa..." but "0xab..." > "0xaa..."
+        Address[] result = new Address[penalties.Count];
+        penalties.CopyTo(result);
+        Array.Sort(result, (a, b) => string.CompareOrdinal(
```
alexb5dh, line 104: Looks very innefective - `ToString` will be called multiple times on the same object. Probably ok, as I guess this method won't be used frequently or operate on many penalties.

Otherwise - consider optimization, for example precomputing `ToString` into dictionary and sorting keys or sorting array of tuples with first value being `ToString` result.

src/Nethermind/Nethermind.Xdc/SubnetPenaltyHandler.cs
ak88: Coinbase cannot be recovered. It is not a signature, but whatever the proposer decided to put in it.

src/Nethermind/Nethermind.Xdc/SubnetPenaltyHandler.cs
ak88: Agree, but just make a TODO for now and we can revisit later when we are feature complete.

src/Nethermind/Nethermind.Xdc/SubnetPenaltyHandler.cs
batrr: It's not wired yet. It will be registered in DI in the future XdcSubnet plugin.

ak88 (approved):

PR #11143, opened 2026-04-14
fix: disable parallel BAL for nonce-loading flows, fix account existence checks

benaadams:
## Changes

Fixes for parallel BAL execution: disable parallel mode for nonce-loading flows and fix account existence checks in `TracedAccessWorldState`.

- `BlockAccessListManager.PrepareForProcessing` - disable parallel execution when `ProcessingOptions.LoadNonceFromState` is set (covers `Trace` and `LoadNonceFromState` flows). Parallel BAL execution assumes nonces are derived from the BAL; when the processing pipeline needs to load nonces from state (e.g. tracing, tx pool nonce recovery), parallel mode must be disabled to avoid incorrect nonce resolution
- `TracedAccessWorldState.AccountExistsCurrent` - expanded account existence logic for parallel BAL execution:
  - Any nonce change implies the account exists (previously only checked `Count == 1`, now handles `Count > 0`)
  - Non-zero balance after a balance change implies existence (parallel execution can create accounts via balance-only transitions before the underlying world state reflects that creation)
  - Non-empty code after a code change implies existence
  - These checks prevent false negatives where `AccountExists` returns `false` for accounts that were created or modified within the current transaction during parallel execution
- Tests added for both fixes

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

- `PrepareForProcessing_disables_parallel_bal_for_LoadNonceFromState` - verifies `ParallelExecutionEnabled` is `false` when `LoadNonceFromState` option is set
- `PrepareForProcessing_disables_parallel_bal_for_Trace` - verifies `ParallelExecutionEnabled` is `false` when `Trace` option is set (which includes `LoadNonceFromState`)
- `ParallelValueCreatedAccount_IsTreatedAsExistingWithinTransaction` - verifies that an account created via `AddToBalanceAndCreateIfNotExists` during parallel execution is seen as existing
- `ParallelZeroBalanceTransition_FallsBackToUnderlyingExistence` - verifies that subtracting balance to zero does not falsely report the account as non-existent when it has a nonce
- `ParallelEmptyCodeTransition_FallsBackToUnderlyingExistence` - verifies that clearing code does not falsely report the account as non-existent when it has balance

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain.Test/BlockProcessorTests.cs
+++ b/src/Nethermind/Nethermind.Blockchain.Test/BlockProcessorTests.cs
@@ -11,6 +11,7 @@ using Nethermind.Consensus.Producers;
 using Nethermind.Consensus.Rewards;
 using Nethermind.Consensus.Withdrawals;
 using Nethermind.Core;
+using Nethermind.Core.BlockAccessLists;
 using Nethermind.Core.Eip2930;
 using Nethermind.Core.Extensions;
 using Nethermind.Core.Specs;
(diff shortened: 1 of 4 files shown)


LukaszRozmej:
@claude review

PR #11157, opened 2026-04-15
feat(logging): add DebugError/DebugWarn helpers with Kind property for Seq

kamilchodola:
## Summary

Replaces the error-prone manual guard idiom with typed helpers:

```csharp
// before
if (_logger.IsDebug) _logger.Error(\$\"DEBUG/ERROR: {whatever}\", ex);

// after
_logger.DebugError(\$\"{whatever}\", ex);
```

Two wins:

1. **Consistency** — an audit of the 32 existing call sites turned up one missing prefix (`RetryCache`) and one reversed `ERROR/DEBUG` prefix (`NodesLoader`, `DiscoveryPersistenceManager`). The new helpers make those mistakes impossible.

2. **Structured filtering in Seq** — the helpers attach a \`Kind\` property (\`\"DebugError\"\` / \`\"DebugWarn\"\`) via NLog's \`LogEventInfo.Properties\`. Seq operators can filter \`Kind = \"DebugError\"\` and keep debug-gated errors out of the main error stream. The \`DEBUG/ERROR: \` message prefix is preserved for plain-text sinks and existing grep-based dashboards.

## Performance

Equivalent to the manual idiom. Uses C# 10 \`[InterpolatedStringHandler]\` so the message isn't built when \`IsDebug\` is false — same skip path as the current guard. Only overhead when disabled is one extra \`NoInlining\` method call (~1-2 ns); these call sites are exception / peer-error paths, never hot-path block processing.

## Files

- \`Nethermind.Logging/ILogger.cs\` — adds \`DebugError\` / \`DebugWarn\` (handler + plain-string overloads) and \`DebugErrorKind\` / \`DebugWarnKind\` const tags.
- \`Nethermind.Logging/DebugInterpolatedStringHandlers.cs\` (new) — the two \`ref struct\` handlers.
- \`Nethermind.Logging/InterfaceLogger.cs\` — extends sink interface with tagged \`Warn\` / \`Error\` overloads. Default interface methods drop the \`Kind\` and fall back to untagged methods, so existing sinks (Console, Limbo, Null, etc.) need no change.
- \`Nethermind.Logging.NLog/NLogLogger.cs\` — overrides tagged methods to attach \`Kind\` via \`LogEventInfo.Properties\`.

## Call-site fixes

Also migrates the three buggy sites found during the audit:
- \`Nethermind.TxPool/RetryCache.cs\` (was missing the prefix)
- \`Nethermind.Network/NodesLoader.cs\` (had \`ERROR/DEBUG\` reversed)
- \`Nethermind.Network.Discovery/DiscoveryPersistenceManager.cs\` (same)

The remaining ~29 correctly-prefixed sites can be migrated in a mechanical follow-up; this PR stays small so the API is easy to review.

## Seq query example

Once landed, operators filtering debug-gated errors out of critical alerts:

\`\`\`
Kind <> \"DebugError\" AND @Level = \"Error\"
\`\`\`

or specifically looking at them:

\`\`\`
Kind = \"DebugError\"
\`\`\`

## Test plan

- [ ] CI build passes on all TFMs
- [ ] Existing logger unit tests still pass
- [ ] Manual: run a node with Seq integration, trigger a \`DebugError\` (e.g. force a decode error), verify \`Kind\` field appears in Seq
- [ ] Manual: run a node with plain file logging, verify \`DEBUG/ERROR: \` prefix still appears in logs

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
@@ -117,7 +117,7 @@ public sealed class BlockCachePreWarmer : IBlockCachePreWarmer
         }
         catch (Exception ex)
         {
-            if (_logger.IsDebug) _logger.Warn($"DEBUG/ERROR Error pre-warming {suggestedBlock.Number}. {ex}");
+            _logger.DebugWarn($"Error pre-warming {suggestedBlock.Number}. {ex}");
         }
         finally
         {
@@ -162,7 +162,7 @@ public sealed class BlockCachePreWarmer : IBlockCachePreWarmer
         }
         catch (Exception ex)
         {
-            if (_logger.IsDebug) _logger.Error("DEBUG/ERROR Error pre-warming withdrawal", ex);
+            _logger.DebugError("Error pre-warming withdrawal", ex);
         }
     }
 
@@ -230,7 +230,7 @@ public sealed class BlockCachePreWarmer : IBlockCachePreWarmer
         }
         catch (Exception ex)
         {
-            if (_logger.IsDebug) _logger.Error($"DEBUG/ERROR Error pre-warming transactions", ex);
+            _logger.DebugError($"Error pre-warming transactions", ex);
         }
     }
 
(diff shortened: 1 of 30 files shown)


LukaszRozmej:
@claude review, apart of correctness, focus if this can be simplified

kamilchodola:
@claude review, apart of correctness, focus if this can be simplified and if the change does not affect maintainability

LukaszRozmej:
@claude review, apart of correctness, focus if this can be simplified and if the change does not affect maintainability

LukaszRozmej:
@claude re-review

benaadams:
@claude does this allocate the interpolated string after or before the `if` check?

benaadams:
Fancy

benaadams (approved):

LukaszRozmej (approved):

kamilchodola:
image
Confirmed working with Kind being nicely figured.

PR #11159, opened 2026-04-15
fix: resolve flaky Can_change_branch_on_invalid_block test with event-driven deletion wait 

### LINKED ISSUE #11148: Flaky test: Can_change_branch_on_invalid_block times out on CI
## Description

`BlockchainProcessorTests.Can_change_branch_on_invalid_block` is flaky on CI. The test polls for up to 10 seconds (`ProcessingWait = 10_000`) for an invalid block to be deleted from the block tree, but on slow CI runners the block processing does not complete within that window.

## Failure details

**Test**: `Nethermind.Blockchain.Test.BlockchainProcessorTests.Can_change_branch_on_invalid_block`  
**File**: `src/Nethermind/Nethermind.Blockchain.Test/BlockchainProcessorTests.cs:695`  
**Assertion at**: line 501 (`IsDeletedAsInvalid()`)

```
block 2 (0xfa3741...3a20ba) should be deleted as invalid
Assert.That(() => processingTestContext._blockTree.FindBlock(block.Hash, BlockTreeLookupOptions.None),
    Is.Null.After(ProcessingWait, 50))
  After 10000 milliseconds delay
  Expected: null
  But was:  <2 (0xfa3741...3a20ba)>
```

**Failed run (attempt 1)**: https://github.com/NethermindEth/nethermind/actions/runs/24429516595/job/71370650185  
**Passed on re-run (attempt 2)**, confirming flakiness.

## Root cause

The `IsDeletedAsInvalid()` helper uses `Is.Null.After(ProcessingWait, 50)` which polls every 50ms for up to 10 seconds. On resource-constrained CI runners (especially the no-intrinsics variant which disables hardware intrinsics), block processing can exceed this timeout.

## Possible fixes

- Increase `ProcessingWait` or use a separate, longer timeout for the deletion assertion
- Switch to an event-driven wait (e.g., wait on a signal that fires when the block is actually deleted) instead of polling

divi2806:
Fixes #11148                                                     
                                                                   
  ## Changes                                                       
                                                                   
  - Added `BlockInvalidated` event to the `IBlockTree` interface,  
  fired by `BlockTree.DeleteInvalidBlock()` after the block has    
  been fully removed and the tree lock released
  - Updated `BlockTreeOverlay` to forward `BlockInvalidated` from  
  the overlay tree                                                 
  - Added no-op `BlockInvalidated` stubs to `ReadOnlyBlockTree` and
   `TestBlockTree`                                                 
  - Added `NotSupportedException` stub to `StatelessBlockTree`,
  consistent with all other events in that class
  - Replaced the polling-based `IsDeletedAsInvalid()` assertion in
  `BlockchainProcessorTests` with an event-driven wait on          
  `BlockInvalidated`, eliminating the 10-second timeout race on
  slow CI runners                                                  
                  
  ## Types of changes

  - [ ] Optimization
  - [ ] Refactoring                                                
  - [ ] Documentation update
  - [ ] Build-related changes
  - [ ] Other: _Description_

  ## Testing

  #### Requires testing

  - [x] Yes
  - [ ] No
                                                                   
  - [x] Yes       
  - [ ] No

  #### Notes on testing

  The existing test `Can_change_branch_on_invalid_block` (and its  
  variants) directly covers this fix. The test was previously flaky
   because `IsDeletedAsInvalid()` polled every 50ms for up to 10   
  seconds; on resource-constrained CI runners (especially the
  no-intrinsics variant) block processing could exceed that window.
   The rewritten helper subscribes to `BlockInvalidated` before
  checking current state (closing the subscription race), then
  waits on a `ManualResetEventSlim` — zero polling, deterministic
  signal regardless of runner speed.

  ## Documentation

  - [ ] Yes
  - [x] No

  - [ ] Yes
  - [x] No


src/Nethermind/Nethermind.Blockchain/BlockTreeOverlay.cs
```diff
@@ -229,6 +229,23 @@ public event EventHandler<IBlockTree.ForkChoiceUpdateEventArgs> OnForkChoiceUpda
...
     }
 
+    // DeleteInvalidBlock delegates exclusively to _overlayTree, so BlockInvalidated
+    // only needs to be forwarded from there.  The _baseTree is a ReadOnlyBlockTree
+    // that never raises this event.
+    public event EventHandler<BlockEventArgs>? BlockInvalidated
+    {
+        add
+        {
+            if (value is not null)
```
LukaszRozmej: I think those null checks are not needed?

src/Nethermind/Nethermind.Blockchain.Test/BlockchainProcessorTests.cs
```diff
@@ -493,14 +493,49 @@ public ProcessingTestContext IsKeptOnBranch()
...
+                }
+                finally
+                {
+                    processingTestContext._blockTree.BlockInvalidated -= OnBlockInvalidated;
+                }
+
+                // Sanity-check: the chain head must not have advanced while an invalid
+                // block was being processed.
+                Assert.That(processingTestContext._blockTree.Head?.Hash, Is.EqualTo(processingTestContext._headBefore), "head");
+                _logger.Info($"Finished waiting for {block.ToString(Block.Format.Short)} to be deleted as invalid");
```
LukaszRozmej, line 524: Quite complicated, cannot be simplified?

LukaszRozmej (commented):
@claude review check if can be done simpler

LukaszRozmej:
@claude review check if can be done simpler

divi2806:
Thanks for the feedback @LukaszRozmej , Implemented the two changes required
  
  **`BlockTreeOverlay.cs`** — removed the redundant `if (value is
  not null)` guards and collapsed to expression-bodied accessors.
  C# handles null delegate subscription safely so the checks were
  dead code.

  **`BlockchainProcessorTests.cs`** — trimmed the 12-line
  step-by-step comment down to two concise lines and condensed the
  `Assert.That` to a single line. Logic is unchanged.

LukaszRozmej (approved):

asdacap:
Can you decorate IBlockTree and intecept DeleteInvalidBlock instead? That way no change in prod is needed.

divi2806:
Thanks @asdacap — agreed, no prod changes needed.

Updated to use `BlockRemoved` event instead (fires after `ProcessBranch`'s `finally` block runs `DeleteInvalidBlocks`), which is an even cleaner signal than a decorator. All 6 production file changes are now reverted — only `BlockchainProcessorTests.cs` is modified.

LukaszRozmej:
@claude re-review

asdacap (approved):

PR #11164, opened 2026-04-16
Enable IDE0320 analyzer for static lambdas

LukaszRozmej:
## Changes

- Enable IDE0320 (Make anonymous function static) analyzer as a warning
- Add .editorconfig to solution items for easier access

## Types of changes

- [x] Build-related changes

#### Requires testing
- [x] No

- [x] No

- [x] No


asdacap:
What? Only warning?

LukaszRozmej:
> What? Only warning?

we have warnings as errors so in the end it doesn't matter, but IMO warning is better because of that to indicate it is more of a style issue than hard issue

svlachakis (approved):

PR #11187, opened 2026-04-16
fix(merge): validate safe/finalized as ancestors of new head

### LINKED ISSUE #11185: `engine_forkchoiceUpdated` rejects valid FCU with `-38002 Inconsistent ForkChoiceState - safe block hash`
**Description**

Nethermind rejects valid `engine_forkchoiceUpdatedV3` (FCU) requests with JSON-RPC error `-38002 Inconsistent ForkChoiceState - safe block hash` even when the request is well-formed per the Engine API specification (`safeBlockHash` is an ancestor of `headBlockHash` on the chain the consensus layer is signalling).

The root cause is in `ForkchoiceUpdatedHandler` (`src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs:345`):

```csharp
private bool IsInconsistent(Hash256 blockHash) =>
    blockHash != Keccak.Zero && !_blockTree.IsMainChain(blockHash);
```

This validates `safeBlockHash` / `finalizedBlockHash` against `_blockTree`'s **currently-canonical** chain (read from `ChainLevelInfo.MainChainBlock`) rather than against the `headBlockHash` from the FCU request. The Engine API spec requires the latter — `safeBlockHash` and `finalizedBlockHash` must be ancestors of the `headBlockHash` *in the same FCU request*, not ancestors of whatever the EL currently considers head.

The check passes a single hash with no head context, so it cannot validate ancestry against the new head. It returns `Inconsistent ForkChoiceState` whenever `_blockTree`'s view of canonical history has not yet been reconciled with the new head — which happens on every FCU path that skips `UpdateMainChain` (notably the `_blockTree.Head!.Hash == newHeadBlock.Hash` short-circuit at line 236, where `shouldUpdateHead` becomes `false`).

This blocks the EF **Fast Confirmation Rule (FCR)** for exchanges (allowed since April 2026, see https://fastconfirm.it/Fast%20Confirmation%20Rule%20for%20Exchanges.pdf) on every Nimbus + Nethermind deployment.

**Steps to Reproduce**

1. Run a Nethermind node (any 1.35.x build, also reproduces on `master`) paired with a Nimbus consensus client.
2. Have the CL issue an `engine_forkchoiceUpdatedV3` where both the new `headBlockHash` and the new `safeBlockHash` are ahead of Nethermind's current `_blockTree.Head` (e.g. blocks were imported via `engine_newPayload` but the EL has not yet promoted them as head), or where `headBlockHash` equals the EL's current head but `safeBlockHash` differs from the previously supplied safe.
3. Observe the JSON-RPC response from Nethermind.

**Actual behavior**

Nethermind responds with:

```json
{"code":-38002,"message":"Inconsistent ForkChoiceState - safe block hash. Request: ForkChoice: 2605131 (0x4f2aff…dcbfc3), Safe: 2605129 (0x11ac6b…99f583), Finalized: 2605075 (0xb76ff0…ca71a4)"}
```

This is incorrect — `0x11ac6b…99f583` (block 2605129) is a valid ancestor of `0x4f2aff…dcbfc3` (block 2605131) on the chain the CL is signalling. The CL treats Nethermind as misbehaving, which prevents the Fast Confirmation Rule from being applied and stalls head/safe/finalized propagation to RPC consumers (indexers, exchange deposit detectors, etc.).

**Expected behavior**

Nethermind should accept the FCU and respond with `VALID`, because `safeBlockHash` and `finalizedBlockHash` are ancestors of the supplied `headBlockHash` per the Engine API specification.

**Screenshots**

The offending line, unchanged for ~3 years (last touched in the "rename keccak -> commitment…" cleanup):

```
src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs:345
private bool IsInconsistent(Hash256 blockHash) => blockHash != Keccak.Zero && !_blockTree.IsMainChain(blockHash);
```

 - Operating System: Linux (also reproduces on Windows)
 - Version: 1.35.x (still present on `master` as of 2026-04-15)
 - Installation Method: GitHub Release / Docker
 - Consensus Client: Nimbus (reported by Etan Kissling, Nimbus team)

**Additional context**

Reading `ApplyForkchoiceUpdate` in `ForkchoiceUpdatedHandler.cs:90-271`, the relevant ordering is:

1. `EnsureNewHead(newHeadBlock, …)` (line 224) — calls `TryGetBranch` to walk back from `newHeadBlock` to the first ancestor already on `_blockTree`'s main chain. If `_blockTree.Head!.Hash == newHeadBlock.Hash`, it short-circuits and returns `null`.
2. `IsOnMainChainBehindHead` (line 231) — early-returns `Valid` if `newHeadBlock` is already canonical and behind head.
3. The skip guard at line 236-241:

   ```csharp
   bool newHeadTheSameAsCurrentHead = _blockTree.Head!.Hash == newHeadBlock.Hash;
   bool shouldUpdateHead = !newHeadTheSameAsCurrentHead && blocks is not null;
   if (shouldUpdateHead)
   {
       _blockTree.UpdateMainChain(blocks!, true, true);
   }
   ```
4. `IsInconsistent(finalizedBlockHash)` (line 243) and `IsInconsistent(safeBlockHash)` (line 250).

`UpdateMainChain` is the only thing that aligns `_blockTree`'s `ChainLevelInfo.MainChainBlock` with the new head. It is skipped whenever `shouldUpdateHead` is false — i.e. when the EL already considers `newHeadBlock` to be the head, or when `blocks` is null. In those skip paths, `IsInconsistent` then runs against stale chain-level info that reflects the previous head.

`_blockTree.IsMainChain` itself (`src/Nethermind/Nethermind.Blockchain/BlockTree.cs:921-932`):

```csharp
public bool IsMainChain(BlockHeader blockHeader) =>
    LoadLevel(blockHeader.Number)?.MainChainBlock?.BlockHash.Equals(blockHeader.Hash) == true;
```

…confirms the check is rooted in persisted chain-level state, not ancestry from the request's `headBlockHash`.

**Quoted from the original report (Etan Kissling, Nimbus, 2026-04-15):**

> "It seems to be impossible that this logic can be correct, as it only takes a single block hash."
> "I think it checks based on the previous head, not based on the headBlockHash from the new fcU."
> "safeBlockHash here is canonical based on headBlockHash, but neth seems to check whether it is canonical based on the `_blockTree` that it already has."
> "I guess what's buggy is a scenario where both new headBlock / safeBlock are _after_ the `_blockTree`'s head."

**Suggested fix**

`IsInconsistent` should validate ancestry from the **new** `headBlockHash` rather than asking `_blockTree.IsMainChain`. Sketch:

```csharp
private bool IsAncestorOf(Hash256 candidate, Block head)
{
    if (candidate == Keccak.Zero) return true;
    BlockHeader? candidateHeader = _blockTree.FindHeader(candidate,
        BlockTreeLookupOptions.TotalDifficultyNotNeeded);
    if (candidateHeader is null || candidateHeader.Number > head.Number) return false;
    BlockHeader? cursor = head.Header;
    while (cursor is not null && cursor.Number > candidateHeader.Number)
        cursor = _blockTree.FindParentHeader(cursor,
            BlockTreeLookupOptions.TotalDifficultyNotNeeded);
    return cursor is not null && cursor.Hash == candidate;
}
```

Then replace `IsInconsistent(finalizedBlockHash)` / `IsInconsistent(safeBlockHash)` with `!IsAncestorOf(finalizedBlockHash, newHeadBlock)` / `!IsAncestorOf(safeBlockHash, newHeadBlock)`.

A regression test in `Nethermind.Merge.Plugin.Test` should:

* Build a chain `A` to height `N`, set head/safe/finalized via FCU.
* Import a longer chain `B` at height `N+K` via `engine_newPayloadV3` so blocks exist in the tree but `_blockTree.MainChainBlock` is not yet updated.
* Send `engine_forkchoiceUpdatedV3` with `headBlockHash = B[N+K]`, `safeBlockHash = B[N+K-2]`.
* Assert `VALID`, not `-38002`.
* Also cover the `newHeadBlock == _blockTree.Head` path with a different `safeBlockHash` than the previous FCU.

**Logs**

```
{"lvl":"DBG","ts":"2026-04-12 20:56:13.795+00:00","msg":"JSON-RPC request failed",
 "topics":"jsonrpc client",
 "err":"{\"code\":-38002,\"message\":\"Inconsistent ForkChoiceState - safe block hash. \
 Request: ForkChoice: 2605131 (0x4f2aff...dcbfc3), \
 Safe: 2605129 (0x11ac6b...99f583), \
 Finalized: 2605075 (0xb76ff0...ca71a4)\"}",
 "id":79009,"remote":"localhost:8553"}
```

**References**

* Engine API spec (`engine_forkchoiceUpdatedV3` semantics for `safeBlockHash` / `finalizedBlockHash`): https://github.com/ethereum/execution-apis
* EF Fast Confirmation Rule for exchanges: https://fastconfirm.it/Fast%20Confirmation%20Rule%20for%20Exchanges.pdf
* Affected file: `src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs` — call site at lines 224-271, offending check at line 345

benaadams:
Resolves #11185

`ForkchoiceUpdatedHandler.IsInconsistent` used `_blockTree.IsMainChain` to validate `safeBlockHash` / `finalizedBlockHash`, which checks against the EL's currently-canonical chain (`ChainLevelInfo.MainChainBlock`) rather than against the `headBlockHash` from the FCU request. The Engine API spec requires the latter, so valid FCUs were rejected with `-38002 Inconsistent ForkChoiceState` whenever the EL's canonical markers had not yet been reconciled with the new head - blocking the EF Fast Confirmation Rule for exchanges on Nimbus + Nethermind setups.

## Diff size

- Production code: 1 file, +39 / -11
- Tests: 1 file, +156 / -2

### Ancestry validation (the fix)

- Replaced the `IsMainChain` check with an ancestry walk from `newHeadBlock` down to the candidate's level, comparing hashes via `GetOrCalculateHash()`. Reuses the `BlockHeader` produced by `ValidateBlockHash` - no extra `FindHeader` lookup.
- Fast path when `newHeadBlock` is canonical: its ancestry is purely canonical, so consistency reduces to `IsMainChain(candidate)` - one lookup, no walk.
- Otherwise a single loop walks parents of `newHeadBlock` down to `candidate.Number`. Walk depth is bounded by `newHeadBlock.Number - candidate.Number` via the number-ordering guard at the top of the method.
- Parent-pointer walk is the source of truth. `ChainLevelInfo.MainChainBlock` is a derived index that can lag behind actual block-tree state during sync/reorg transitions; the walk reads authoritative data directly from block headers.
- Ancestry is validated on every FCU. The binding is `(head, finalized, safe)` per the spec, so a repeated finalized or safe hash paired with a new head on a sibling branch is still a spec violation and must be rejected.

### Spec-ordering and monotonicity

- Extracted a local `RejectIfInconsistent(header, lowerBound, label)` helper that captures `newHeadBlock` / `requestStr` so finalized and safe share a single implementation. Adds explicit O(1) checks at the call site:
  - `safe.Number >= finalized.Number` (spec convention)
  - `finalized.Number <= head.Number`, `safe.Number <= head.Number`
  - **Casper FFG monotonicity**: `finalized.Number >= previously-accepted finalized.Number`. Reads the cached `_manualBlockFinalizationManager.LastFinalizedBlockLevel` - no `FindHeader` lookup.

### Tests

- **Production-symptom regression** (`..._is_accepted_when_canonical_markers_are_stale`): injects the exact state from the production log - `Head` unchanged, canonical marker at an ancestor level swapped to a sibling via `BlockTree.UpdateMainChain([sibling], wereProcessed: false)` (the sync-layer path that moves canonical markers ahead of head advancement). The pre-fix code rejects this FCU with `-38002 Inconsistent ForkChoiceState - safe block hash` (matching the production error byte-for-byte); the fix accepts it.
- **Spec acceptance case**: a valid FCU whose safe/finalized are real ancestors of a new head that the EL hasn't yet promoted to canonical, with explicit preconditions and canonical-state post-conditions.
- **Spec-ordering rejection**: covers Casper FFG monotonicity and `safe < finalized` - neither was exercised before this PR.
- **Parameterized sibling-branch rejection** (`[TestCase]` over finalized and safe): the previously-accepted hash is re-sent alongside a head that is not a descendant of it. For the safe case, finalized is set to a genuine common ancestor so only the safe-ancestry check can reject.
- Two existing reorg tests (`forkchoiceUpdatedV1_can_reorganize_to_last_block`, `forkchoiceUpdatedV1_head_block_after_reorg`) updated to use `setHead: false` on the sibling branch - the prior `setHead: true` produced a finality regression that the new monotonicity check (correctly) rejects.

#### Requires testing

- [x] Yes

- [x] Yes

---

### IsInconsistent: short-circuit the ancestry walk when the candidate is on main chain

The original ancestry walk always stepped from `newHeadBlock` all the way down to `candidate.Number` and then compared hashes. Ethereum FCUs almost always supply a finalized/safe hash that *is* on the current main chain, so in practice the walk is almost always scanning main-chain-eligible blocks.

- Cache `bool candidateIsMain = _blockTree.IsMainChain(candidateHeader)` once at the top of the method, reuse for the `newHead` canonical fast path.
- Inside the walk, after stepping to the parent, return `consistent` as soon as a main-chain ancestor is hit: if both cursor and candidate are on the canonical chain and `cursor.Number >= candidate.Number`, ancestry is proven; continuing down to the candidate's level is wasted work.
- The main-chain probe is placed *after* the parent step so `IsMainChain(newHead)` (already answered `false` by the fast path above) is not re-asked on the first iteration.
- Track `cursor` as non-nullable and bail on a missing parent; removes the paired null checks at the loop condition and after it.

### Handler internals moved from Block to BlockHeader

`GetBlock` → `GetBlockHeader` (using `FindHeader(hash, DoNotCreateLevelIfMissing)`). Eagerly loading the full block body at the top of `Handle` was wasted work on paths that never need the body: same-head heartbeat FCUs, sync paths, early errors. The body load now happens inside `TryGetBranch` only when a reorg actually requires the block array for `UpdateMainChain`.

Retyped accordingly: `ApplyForkchoiceUpdate`, `IsOnMainChainBehindHead`, `IsPayloadTimestampValid`, `ArePayloadAttributesTimestampAndSlotNumberValid`, `StartBuildingPayload`, `EnsureNewHead`, `TryGetBranch` (returns `IReadOnlyList<Block>`), `IsInconsistent`, `ReorgBeaconChainDuringSync`, `GetBeaconChainBranch`. Extensions `BlockTreeExtensions.IsOnMainChainBehind(OrEqual)Head` take `BlockHeader`; `NewPayloadHandler` updated to pass `block.Header`. `TaikoForkchoiceUpdatedHandler` overrides retyped in lockstep; its "already-on-main" shortcut materializes the `Block` via `FindBlock` only when needed. Taiko unit tests now mock `FindHeader` and no longer need `FindBlock` mocks on head=current-Head paths.

### `IBlockTree.UpdateBeaconMainChain` takes `IReadOnlyList<BlockInfo>`

The implementation only uses `foreach` and `blockInfos[^1]`, both covered by `IReadOnlyList<BlockInfo>`. Lets callers pass `List<BlockInfo>` directly and drops an unnecessary `.ToArray()` in `BlockTreeTests`. Signature updated on `IBlockTree`, `BlockTree`, `BlockTreeOverlay`, `ReadOnlyBlockTree`, `StatelessBlockTree`, and the test `TestBlockTree`.

### Test dedup and fast-path coverage

- Collapsed `inconsistent_finalized_hash` / `inconsistent_safe_hash` into a single `[TestCase]`-parameterized test (original names preserved via `TestName`).
- Extracted `BuildYShapedChainV1` shared by the three #11185 tests that exercise the `block1 → {2A sibling, 2B → 3B}` topology.
- Extracted `FlipCanonicalMarkerTo` for the "sync flips a canonical marker without advancing Head" idiom.
- Added `forkchoiceUpdated_isInconsistent_takes_fast_path_when_candidate_is_on_main_chain` which pins the walk's observable depth. Observability is done at the `IBlockTree` boundary via `BlockTreeCallSpy` — a `DispatchProxy` that forwards every call to an inner instance while counting `FindHeader` invocations — so the handler carries no test-only instrumentation. Pre-optimization the test reports 4 `FindHeader` calls (1 head resolve + 1 finalized validation + 2 walk steps); post-optimization 3 (walk short-circuits to 1 step).


benaadams:
@claude review

src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
```diff
@@ -342,7 +342,36 @@ private void StartNewBeaconHeaderSync(ForkchoiceStateV1 forkchoiceState, BlockHe
...
+    // Fast path (steady state): if BOTH the candidate and newHeadBlock are on the canonical
+    // main chain, then by chain linearity (one canonical block per level, candidate.Number
+    // <= newHeadBlock.Number) the candidate is necessarily an ancestor of newHeadBlock. Two
+    // O(1) ChainLevelInfo lookups vs. the parent walk in the fallback. The IsMainChain check
+    // on newHeadBlock is essential — IsMainChain(candidate) alone would silently accept a
+    // non-ancestor when newHeadBlock sits on a side branch (the inverse of issue #11185).
+    //
+    // Slow path: walk parents from newHeadBlock and compare hashes. Required when the EL has
+    // not yet reconciled its canonical view to newHeadBlock — the case the original
+    // _blockTree.IsMainChain(blockHash) check got wrong (issue #11185).
```
LukaszRozmej: I don't get it, why the Head wasn't already updated? Then `IsMainChain` previous check would be sufficient?

src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
```diff
@@ -342,7 +342,36 @@ private void StartNewBeaconHeaderSync(ForkchoiceStateV1 forkchoiceState, BlockHe
...
+    private bool IsInconsistent(BlockHeader? candidateHeader, Block newHeadBlock)
+    {
+        if (candidateHeader is null) return false;
+        if (candidateHeader.Number > newHeadBlock.Number) return true;
+
+        if (_blockTree.IsMainChain(candidateHeader) && _blockTree.IsMainChain(newHeadBlock.Header))
+            return false;
+
+        BlockHeader? cursor = newHeadBlock.Header;
+        while (cursor is not null && cursor.Number > candidateHeader.Number)
```
LukaszRozmej: Should we walk arbitrary depth? Shouldn't we have a limit or something?

svlachakis:
Is this gonna affect [Gnosis](https://github.com/NethermindEth/nethermind/pull/10876) main fix in any way?

src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
benaadams: safeBlockHash and finalizedBlockHash must be ancestors of the headBlockHash in the same FCU request, not ancestors of whatever the EL currently considers head.

benaadams: Is the bug basically; atm we say it needs to be the head's chain

src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
benaadams: Would be a Warn if over limit?

benaadams:
> Is this gonna affect [Gnosis](https://github.com/NethermindEth/nethermind/pull/10876) main fix in any way?

Should be unrealated

src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
LukaszRozmej: Not sure if we should have a hard limit... what Geth does?

LukaszRozmej: I think we can optimize more, if `newHeadBlock.Header` is Main chain and `candidateHeader` is not - Inconsistent

If `candidateHeader` is main chain and `newHeadBlock.Header` is not - it is enough to walk down to first block on main chain and check if it's number > candidateHeader.Number, and only otherwise walk down to `candidateHeader`

benaadams: Better now?

benaadams:
@claude review

src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
```diff
@@ -342,7 +350,37 @@ private void StartNewBeaconHeaderSync(ForkchoiceStateV1 forkchoiceState, BlockHe
...
+                cursor = _blockTree.FindParentHeader(cursor, BlockTreeLookupOptions.TotalDifficultyNotNeeded);
+            }
+            return cursor is null || cursor.Number < candidateHeader.Number;
+        }
+
+        while (cursor is not null && cursor.Number > candidateHeader.Number)
+        {
+            cursor = _blockTree.FindParentHeader(cursor, BlockTreeLookupOptions.TotalDifficultyNotNeeded);
+        }
+        return cursor is null || cursor.GetOrCalculateHash() != candidateHeader.GetOrCalculateHash();
```
LukaszRozmej: I think this should be one loop handling both cases at the same time

src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
```diff
@@ -240,14 +240,22 @@ protected virtual bool IsOnMainChainBehindHead(Block newHeadBlock, ForkchoiceSta
...
         }
 
-        if (IsInconsistent(finalizedBlockHash))
+        // Spec ordering: prevFinalized <= finalized <= safe <= head. Ancestry must be re-validated
+        // on every FCU - the binding is (head, finalized, safe), so a repeated finalized/safe hash
+        // paired with a new head on a sibling branch is still a spec violation.
+        long prevFinalizedLevel = _manualBlockFinalizationManager.LastFinalizedBlockLevel;
+        long finalizedNumber = finalizedHeader?.Number ?? 0;
+
+        if ((finalizedHeader is not null && (finalizedNumber < prevFinalizedLevel || finalizedNumber > newHeadBlock.Number))
```
LukaszRozmej: helper method 2x same thing

src/Nethermind/Nethermind.Merge.Plugin.Test/EngineModuleTests.V1.cs
```diff
@@ -1405,6 +1408,136 @@ public async Task inconsistent_safe_hash()
         forkchoiceUpdatedResult3.ErrorCode.Should().Be(MergeErrorCodes.InvalidForkchoiceState);
     }
 
+    [Test]
+    public async Task forkchoiceUpdated_safe_block_that_is_real_ancestor_of_new_head_is_accepted()
```
LukaszRozmej, line 1389: @claude can those 3 tests be de-duplicated?

benaadams:
@claude re-review

benaadams:
@claude review

LukaszRozmej:
@claude re-review

LukaszRozmej (approved):

svlachakis (approved):

asdacap (approved):

khanbilal732 (approved):

PR #11190, opened 2026-04-16
fix(eth_call): return descriptive error when maxFeePerGas is     below block base fee 

### LINKED ISSUE #11177: `eth_call`: incorrect error message when `maxFeePerGas` is below block base fee
### Description

When `eth_call` is invoked with a `maxFeePerGas` (or `gasPrice`) that is lower than the
current block's base fee, Nethermind returns a vague error message `"miner premium is negative"`
instead of the descriptive Geth-style message that includes the address, the provided fee cap,
and the actual base fee. This makes it harder for callers to understand why the call failed.

### Steps to Reproduce

```json
// test_02 — block 0xD59F7F (baseFee ~146.3 gwei)
{
  "jsonrpc": "2.0",
  "method": "eth_call",
  "params": [
    {
      "from": "0xd3181ddbB2CEA7B4954b8D4a05DBf85D8Fc36aef",
      "to": "0xe3782B8688ad2b0D5ba42842d400F7AdF310F88d",
      "gas": "0xDDD2",
      "gasPrice": "0x2098A67800",
      "value": "0x0",
      "data": "0xf242432a..."
    },
    "0xD59F7F"
  ],
  "id": 1
}
```

### Expected (Geth)

```json
{
  "error": {
    "code": -32000,
    "message": "err: max fee per gas less than block base fee: address 0xd3181ddbB2CEA7B4954b8D4a05DBf85D8Fc36aEF, maxFeePerGas: 140000000000, baseFee: 146283608928 (supplied gas 56786)"
  }
}
```

### Actual (Nethermind)

```json
{
  "error": {
    "code": -32000,
    "message": "miner premium is negative"
  }
}
```

### Notes

Same issue reproduces on test_18 with `maxFeePerGas: 50 gwei` against a block with
`baseFee: ~61.5 gwei`. The error code `-32000` is correct — only the message text differs.

divi2806:
Fixes #11177                                                     
                                                                   
  ## Changes                                                
                                                                   
  - Added `TransactionResult.CreateMinerPremiumNegative(Transaction
   tx, UInt256 baseFee)` factory method that builds a              
  Geth-compatible error message including the sender address,      
  provided fee cap, block base fee, and supplied gas               
  - Updated `TransactionResult.ErrorDescription` for               
  `MinerPremiumNegative` to use the contextual message when        
  available, falling back to `"miner premium is negative"` for     
  cases without context                                          
  - Updated `TransactionProcessor.BuyGas` to use                   
  `CreateMinerPremiumNegative` instead of the static field so
  `eth_call` and `eth_estimateGas` return the descriptive message
  - Updated `OptimismTransactionProcessor.BuyGas` with the same fix
  - Added regression tests for the descriptive error message in    
  both `eth_call` and `eth_estimateGas` paths
                                                                   
  ## Types of changes                                              
                                                                 
  - [ ] Optimization
  - [ ] Refactoring                                                
  - [ ] Documentation update                                
  - [ ] Build-related changes                                    
  - [ ] Other                                 
                                          
  ## Testing
                                                                   
  #### Requires testing
                                                                   
  - [x] Yes                                                 
  - [ ] No                                                       

  - [x] Yes
  - [ ] No

  #### Notes on testing

  Added `Call_tx_returns_descriptive_MinerPremiumIsNegativeError`
  and                                     
  `EstimateGas_tx_returns_descriptive_MinerPremiumIsNegativeError`
  in `Nethermind.Facade.Test/BlockchainBridgeTests.cs`. Existing   
  tests covering the static fallback message continue to pass.
                                                                   
  ## Documentation                                          
                                                                 
  - [ ] Yes
  - [x] No

  - [x] Yes                                   
  - [ ] No                                

  `eth_call` and `eth_estimateGas` now return a Geth-compatible    
  error message when `maxFeePerGas` (or `gasPrice`) is below the
  block base fee, e.g.:                                            
  `err: max fee per gas less than block base fee: address 0x...,   
  maxFeePerGas: 140000000000, baseFee: 146283608928 (supplied gas
  56786)`


divi2806:
@claude review

divi2806:
@svlachakis can you check this?

DarkLord017:
@claude review

divi2806:
1. **Redundant ternary removed** — `CreateMinerPremiumNegative`  
  now uses `tx.MaxFeePerGas` directly since it already resolves to 
  `DecodedMaxFeePerGas` for EIP-1559 and `GasPrice` for legacy     
  transactions.                                                    
   
  2. **Real `TransactionProcessor` path now covered** — Added      
  `CallAndRestore_returns_descriptive_error_when_maxFeePerGas_below
  _baseFee` in `TransactionProcessorFeeTests.cs` using a fully     
  wired `EthereumTransactionProcessor` (same setup as the rest of  
  that file). This hits the actual `BuyGas` code path at
  `TransactionProcessor.cs:575` and verifies the address, fee cap, 
  base fee, and supplied gas all appear in the error description.

  3. **Legacy transaction covered** — The above test uses
  `[TestCase(TxType.EIP1559)]` and `[TestCase(TxType.Legacy)]`,
  exercising both branches. For legacy, `GasPrice` is set below
  `BaseFeePerGas` and `tx.MaxFeePerGas` correctly resolves to      
  `GasPrice`.
                                                                   
  4. **Static `MinerPremiumNegative` field** — Agreed, it is now
  only a fallback and test artifact. Kept intentionally as a safety
   net for external callers and the existing mock-based tests that
  return it directly.

divi2806:
@DarkLord017 @svlachakis

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -1077,6 +1077,12 @@ private TransactionResult(ErrorType error = ErrorType.None, EvmExceptionType evm
 
         public static TransactionResult EvmException(EvmExceptionType evmExceptionType, string? substateError = null) => new(ErrorType.None, evmExceptionType) { SubstateError = substateError };
 
+        public static TransactionResult CreateMinerPremiumNegative(Transaction tx, in UInt256 baseFee)
+        {
+            string detail = $"err: max fee per gas less than block base fee: address {tx.SenderAddress?.ToString() ?? "unknown"}, maxFeePerGas: {tx.MaxFeePerGas}, baseFee: {baseFee} (supplied gas {tx.GasLimit})";
+            return new TransactionResult(ErrorType.MinerPremiumNegative) { SubstateError = detail };
```
DarkLord017: maybe define it as a private const above Error Description(line like just so that all error messages are together (line 1052)
``` javascript
private const string MinerPremiumNegativeMessageTemplate =
    "err: max fee per gas less than block base fee: address {0}, maxFeePerGas: {1}, baseFee: {2} (supplied gas {3})";
```

DarkLord017:
@claude review

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
divi2806: The suggestion is reasonable for grouping, but since this 
  template is used exactly once, extracting it to a `private const`
   adds ceremony without real benefit:                          
                                                            
  - A `private const` with `{0}..{3}` placeholders forces          
  `string.Format` which is less readable than the inline
  interpolated string we have now.                                 
  - The coding style prefers not extracting one-off operations into
   abstractions.                                                   
  - The other messages in `ErrorDescription` are inline literals   
  too ,making this one a named const would be inconsistent with   
  the rest of the switch.                                          
                                                                   
  Happy to change it if there's a stronger reason (e.g. the        
  template is reused elsewhere), but for a single callsite the
  interpolated string inline is cleaner.

src/Nethermind/Nethermind.Facade.Test/BlockchainBridgeTests.cs
```diff
@@ -492,6 +492,28 @@ public void Call_tx_returns_MinerPremiumIsNegativeError()
         Assert.That(callOutput.Error, Is.EqualTo("miner premium is negative"));
     }
 
+    [Test]
```
LukaszRozmej: @claude can we de-dup those new tests?

DarkLord017 (dismissed):

src/Nethermind/Nethermind.Facade.Test/BlockchainBridgeTests.cs
divi2806: Shoud  I collapse the 4 tests into 2 using a shared
  `[TestCaseSource]` ?

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -1056,7 +1056,7 @@ private TransactionResult(ErrorType error = ErrorType.None, EvmExceptionType evm
             ErrorType.InsufficientMaxFeePerGasForSenderBalance => "insufficient MaxFeePerGas for sender balance",
             ErrorType.InsufficientSenderBalance => "insufficient sender balance",
             ErrorType.MalformedTransaction => "malformed",
-            ErrorType.MinerPremiumNegative => "miner premium is negative",
+            ErrorType.MinerPremiumNegative => SubstateError ?? "miner premium is negative",
```
svlachakis: Is there any path that still produces `miner premium is negative` ?

svlachakis: Looks like there is optimism, but I still don't like this - maybe we could have a different ErrorType for this path, @LukaszRozmej wdyt?

divi2806: No production path hits the fallback anymore, both              
  `TransactionProcessor.BuyGas` and                                
  `OptimismTransactionProcessor.BuyGas` now call                   
  `CreateMinerPremiumNegative` which always sets `SubstateError`.  
  The `?? "miner premium is negative"` only exists as a safety net
  for the static field, which is only used in tests. Happy to      
  remove both if you prefer to make the descriptive message        
  mandatory.

src/Nethermind/Nethermind.Optimism/OptimismTransactionProcessor.cs
```diff
@@ -88,7 +88,7 @@ protected override TransactionResult BuyGas(Transaction tx, IReleaseSpec spec, I
             if (validate && !tx.TryCalculatePremiumPerGas(header.BaseFeePerGas, out premiumPerGas))
             {
                 TraceLogInvalidTx(tx, "MINER_PREMIUM_IS_NEGATIVE");
-                return TransactionResult.MinerPremiumNegative;
+                return TransactionResult.CreateMinerPremiumNegative(tx, header.BaseFeePerGas);
```
svlachakis: I'm not sure this change is correct here, @emlautarom1 should this error message remain as it was?

src/Nethermind/Nethermind.Facade.Test/BlockchainBridgeTests.cs
divi2806: Done, 
I have collapsed 4 individual tests - 2 `[TestCaseSource]` tests sharing a single `MinerPremiumNegativeCases() `source
Source yields two cases: `Fallback` (static field, `"miner premium is negative"`) and `Descriptive` (factory, Geth-style message)
Both `Call_tx_returns_MinerPremiumIsNegativeError` and `EstimateGas_tx_returns_MinerPremiumIsNegativeError` consume the same source — the only difference between them is which bridge method they call
Added `using System.Collections.Generic` for `IEnumerable<TestCaseData>`

src/Nethermind/Nethermind.Optimism/OptimismTransactionProcessor.cs
emlautarom1: I think it's fine to change it, this only changes the user-facing error message right?

divi2806: Correct, this only changes the string returned in the     
  `ErrorDescription` , no behaviour, no error type, no control flow
   is affected. The `ErrorType` stays `MinerPremiumNegative`, so
  any code branching on the error type (like `SimulateTxExecutor`)
  is unaffected.

svlachakis: alright, @divi2806 let's please make it clean then - I think there is no reason for the default error https://github.com/NethermindEth/nethermind/pull/11190#discussion_r3096478024

divi2806: @svlachakis 

Done, introduced `ErrorType.MaxFeePerGasBelowBaseFee` as a
  dedicated type for the descriptive path.
  `CreateMinerPremiumNegative` now returns this new type with
  `SubstateError` always set, so `ErrorDescription` is just
  `SubstateError!` with no fallback. The old `MinerPremiumNegative`
   type and its static field are untouched for backward compat
  only.

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -1086,6 +1086,7 @@ private TransactionResult(ErrorType error = ErrorType.None, EvmExceptionType evm
             ErrorType.InsufficientMaxFeePerGasForSenderBalance => "insufficient MaxFeePerGas for sender balance",
             ErrorType.InsufficientSenderBalance => "insufficient sender balance",
             ErrorType.MalformedTransaction => "malformed",
+            ErrorType.MaxFeePerGasBelowBaseFee => SubstateError!,
```
svlachakis: @claude doesn't that require SubstateError to be set? Since it's a variable it's actually something like ErrorType.SubstateError which can by anything correct? Propose better approach here

divi2806: @svlachakis  The constructor is private and there's no static field for       
  `MaxFeePerGasBelowBaseFee`, so the only way to produce this type 
  is through `CreateMinerPremiumNegative` which always sets 
  `SubstateError`. Should I add an explicit guard like             
  `SubstateError ?? throw new InvalidOperationException(...)` for
  extra safety, or is the structural enforcement enough?

svlachakis: https://github.com/NethermindEth/nethermind/pull/11243 - i did some refactoring

divi2806: Thankyou @svlachakis , Implemented the refactoring in #11243 
- `TransactionProcessor.cs` : removed SubstateError { get; private 
  init; }, replaced with errorDescription constructor parameter
  stored as ErrorDescription { get; }                              
 - `BlockchainBridge.cs` : updated ConstructError to use            
  ErrorDescription instead of the removed SubstateError            
 - `Nethermind.Xdc/TransactionResult.cs` :  dropped the redundant    
  explicit (TransactionResult) casts, relying on the implicit      
  operator

LukaszRozmej:
@claude re-review

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -599,7 +599,7 @@ protected virtual TransactionResult BuyGas(Transaction tx, IReleaseSpec spec, IT
             if (validate && !TryCalculatePremiumPerGas(tx, header.BaseFeePerGas, out premiumPerGas))
             {
                 TraceLogInvalidTx(tx, "MINER_PREMIUM_IS_NEGATIVE");
-                return TransactionResult.MinerPremiumNegative;
+                return TransactionResult.CreateMinerPremiumNegative(tx, header.BaseFeePerGas);
```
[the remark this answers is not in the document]
divi2806: @LukaszRozmej, Implemented the suggestions and fixes by claude. Added `CreateMinerPremiumNegative`

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
svlachakis: @divi2806 I refactored it in order to decouple the two errors clearly..

Optimism should remain as it was in terms of error message

`public static readonly TransactionResult MinerPremiumNegative = new(ErrorType.MinerPremiumNegative, errorDescription: "miner premium is negative");`

new path should have the new error with a function like the one you created for `CreateMinerPremiumNegative`

divi2806: @svlachakis , 
Added back the previous error message for optimism 
`MinerPremiumNegative
`

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -1093,6 +1093,12 @@ private TransactionResult(ErrorType error = ErrorType.None, EvmExceptionType evm
         public static TransactionResult EvmException(EvmExceptionType evmExceptionType, string? description = null) =>
             new(evmException: evmExceptionType, errorDescription: description ?? "");
 
+        public static TransactionResult CreateMinerPremiumNegative(Transaction tx, in UInt256 baseFee)
```
svlachakis:
```suggestion
        public static TransactionResult WithDetail(ErrorType errorType, string detail) => new(errorType, errorDescription: detail);
```

svlachakis (changes requested):

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
divi2806: Done. Replaced `CreateMinerPremiumNegative` with the generic       
  `WithDetail(ErrorType, string) `factory. The detail string is now
  built at the call site in `BuyGas` and passed in directly. Updated 
  the test accordingly. All 6 tests pass.

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -599,7 +599,7 @@ protected virtual TransactionResult BuyGas(Transaction tx, IReleaseSpec spec, IT
             if (validate && !TryCalculatePremiumPerGas(tx, header.BaseFeePerGas, out premiumPerGas))
             {
                 TraceLogInvalidTx(tx, "MINER_PREMIUM_IS_NEGATIVE");
-                return TransactionResult.CreateMinerPremiumNegative(tx, header.BaseFeePerGas);
+                return TransactionResult.WithDetail(TransactionResult.ErrorType.MaxFeePerGasBelowBaseFee, $"err: max fee per gas less than block base fee: address {tx.SenderAddress?.ToString() ?? "unknown"}, maxFeePerGas: {tx.MaxFeePerGas}, baseFee: {header.BaseFeePerGas} (supplied gas {tx.GasLimit})");
```
svlachakis: please move the error to a variable, this is not human readable

divi2806: added it inside `errorDetail` variable and made it more better to understand.

divi2806:
@svlachakis, 
Any optimizations or further improvments needed? 
Thankyou

svlachakis:
@divi2806 tests need to pass, looks like there are failures..

divi2806:
Thankyou @svlachakis ,
Fixed. Added a new substring mapping in                          
  `Ethereum.Test.Base/BlockchainTestBase.cs`:                      
                                                                   
  ```csharp                                                        
  ("TransactionException.INSUFFICIENT_MAX_FEE_PER_GAS", "max fee
  per gas less than block base fee"),

svlachakis:
@claude re-review

svlachakis (approved):

divi2806:
@LukaszRozmej @svlachakis, is the PR good to be merged?

svlachakis:
LGTM

divi2806:
cool, thanks.
@LukaszRozmej @benaadams @smartprogrammer93 i think it can be merge now?

LukaszRozmej (approved):

divi2806:
@LukaszRozmej the json rpc error is because of this issue?
Or is it good?

PR #11230, opened 2026-04-18
Optimize EVM stack

benaadams:
Stack changes from https://github.com/NethermindEth/nethermind/pull/10120

### `EvmStack` rewrite
- Stack is now a `ref byte` into a pinned, 32-byte-aligned pooled `byte[]` (previously `Span<byte>`). Head tracked separately, rented via `StackPool`.
- Hoisted `Code` / `CodeLength` onto `EvmStack`, removing the `vm.VmState.Env.CodeInfo.CodeSpan` chase on every PUSH* / JUMP / CODESIZE.
- New per-size `PushNBytes<TTracingInst>` family (N = 1..32) with AVX2 + AVX512-VBMI paths and a scalar fallback; single 32-byte store when the whole word is written.
- New `PopUInt256(out a, out b)`, `(out a, out b, out c)`, `(out a, out b, out c, out d)` overloads: amortise the bounds check, hoist the big-endian shuffle mask once, and interleave the 12 scalar loads for ILP on the non-AVX path.
- `Push*` / `Pop*` converted from throw-based to return-based error handling (`EvmExceptionType.StackOverflow` / `StackUnderflow`). Only internal-arithmetic `OverflowException` still crosses the interpreter boundary via the catch in `VirtualMachine`.
- Big-endian PSHUFB / `PermuteVar32x8` mask consolidated into a single `ByteSwap256Mask` property (was inlined ~7 times).

### Instruction dispatch
- `EvmInstructions.Stack.cs`: every `PushN` opcode (N = 0, 1, 3..32) goes through its own `IOpCount.Push` that calls the matching `stack.PushNBytes<TTracingInst>(ref start)`. `InstructionPush2` has a dedicated PUSH2+JUMP/JUMPI fusion path that skips the push and jumps directly (including past the `JUMPDEST` - gas + op count charged once, not twice).
- Specialized `InstructionCodeSize` reads `stack.CodeLength` directly.
- Call sites across `Math2Param`, `Math3Param`, `Shifts`, `Crypto`, `CodeCopy`, `Call`, `Create`, `Storage`, `Environment` adopt the new multi-out pops and propagate `Push*` return values so stack overflow surfaces correctly.
- Prefetch (`Sse.Prefetch0`) at validated jump destinations (`InstructionJump`, `InstructionJumpIf`, fused PUSH2+JUMP) - hardware prefetchers can't predict jumps.

### `VmState`
- New `InitializeStacks(codeSpan, out stack)` + `InitializeStacks(tracer, codeSpan, out stack)` overloads that lazy-rent a 32-aligned pooled stack and build an `EvmStack` over an aligned ref.
- `MemoryStacks(count)` for tracer access to the aligned stack memory; also lazy-rents if needed.

### CI
- `.github/workflows/evm-opcode-benchmark-diff.yml`: fixed BDN filter syntax for the per-opcode noisy-rerun step. The old `*EvmOpcodesBenchmark*Opcode=ADD*` pattern matched zero cases because BDN renders parameters as `(Opcode: ADD)` with a colon-space. Pattern is now `*EvmOpcodesBenchmark*Opcode: ADD)*`; the trailing `)` anchors the glob so `ADD` does not also match `ADDMOD`, `OR` does not match `XOR`, etc.

### Benchmarks
- `Nethermind.Evm.Benchmark/EvmOpcodesBenchmark.cs`, `EvmStackBenchmarks.cs` updated for the new `EvmStack` constructor and `ref byte` layout.

33 Improvements

| Opcode | Base Median (ns) | PR Median (ns) | Delta | Abs Δ (ns) | Base CV | PR CV | Threshold | Uncertainty | Effective |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| BYTE | 17.570 | 14.820 | -15.65% | 2.750 | 4.5% | 1.8% | ±5.0% | ±3.6% | ±9.0% |
| CALLDATALOAD | 17.930 | 12.574 | -29.87% | 5.356 | 10.2% | 1.4% | ±5.0% | ±6.1% | ±20.4% |
| CODESIZE | 12.573 | 9.098 | -27.64% | 3.475 | 8.5% | 6.9% | ±5.0% | ±8.1% | ±17.0% |
| GT | 19.396 | 15.890 | -18.08% | 3.506 | 5.4% | 5.5% | ±5.0% | ±6.2% | ±11.1% |
| LT | 19.767 | 16.152 | -18.29% | 3.615 | 6.6% | 4.5% | ±5.0% | ±6.5% | ±13.1% |
| MUL | 46.270 | 34.834 | -24.72% | 11.436 | 7.7% | 11.2% | ±5.0% | ±9.6% | ±22.4% |
| PUSH1 | 13.298 | 9.160 | -31.12% | 4.138 | 8.4% | 4.0% | ±5.0% | ±6.5% | ±16.8% |
| PUSH10 | 17.285 | 11.556 | -33.14% | 5.729 | 8.8% | 2.7% | ±5.0% | ±6.1% | ±17.6% |
| PUSH11 | 17.554 | 11.680 | -33.46% | 5.874 | 9.1% | 2.4% | ±5.0% | ±6.1% | ±18.2% |
| PUSH12 | 17.809 | 11.599 | -34.87% | 6.210 | 10.2% | 2.1% | ±5.0% | ±6.4% | ±20.3% |
| PUSH13 | 17.250 | 11.639 | -32.53% | 5.611 | 10.1% | 2.5% | ±5.0% | ±6.6% | ±20.3% |
| PUSH14 | 17.650 | 11.785 | -33.23% | 5.865 | 8.5% | 5.9% | ±5.0% | ±7.4% | ±16.9% |
| PUSH15 | 17.530 | 11.793 | -32.73% | 5.737 | 8.4% | 1.9% | ±5.0% | ±5.5% | ±16.7% |
| PUSH17 | 18.199 | 11.535 | -36.62% | 6.664 | 9.4% | 3.6% | ±5.0% | ±6.7% | ±18.8% |
| PUSH18 | 17.870 | 11.586 | -35.17% | 6.284 | 12.0% | 4.5% | ±5.0% | ±8.5% | ±24.1% |
| PUSH19 | 18.030 | 11.636 | -35.46% | 6.394 | 9.0% | 2.2% | ±5.0% | ±6.1% | ±17.9% |
| PUSH2 | 16.010 | 10.913 | -31.84% | 5.097 | 8.2% | 2.4% | ±5.0% | ±5.6% | ±16.4% |
| PUSH21 | 18.270 | 11.774 | -35.56% | 6.496 | 9.2% | 3.6% | ±5.0% | ±6.7% | ±18.3% |
| PUSH22 | 17.737 | 11.749 | -33.76% | 5.988 | 6.1% | 2.8% | ±5.0% | ±4.7% | ±12.3% |
| PUSH23 | 18.060 | 11.636 | -35.57% | 6.424 | 9.1% | 2.3% | ±5.0% | ±6.1% | ±18.2% |
| PUSH24 | 17.936 | 11.412 | -36.37% | 6.524 | 8.0% | 3.9% | ±5.0% | ±6.3% | ±16.0% |
| PUSH25 | 18.081 | 11.682 | -35.39% | 6.399 | 8.9% | 1.0% | ±5.0% | ±5.2% | ±17.7% |
| PUSH26 | 17.961 | 11.643 | -35.18% | 6.318 | 10.3% | 2.1% | ±5.0% | ±6.5% | ±20.5% |
| PUSH27 | 18.391 | 11.655 | -36.63% | 6.736 | 9.8% | 3.0% | ±5.0% | ±6.7% | ±19.7% |
| PUSH28 | 18.020 | 11.342 | -37.06% | 6.678 | 8.2% | 6.4% | ±5.0% | ±7.4% | ±16.4% |
| PUSH29 | 17.895 | 11.866 | -33.69% | 6.029 | 10.9% | 2.2% | ±5.0% | ±7.1% | ±21.8% |
| PUSH3 | 18.304 | 11.577 | -36.75% | 6.727 | 9.0% | 2.8% | ±5.0% | ±6.1% | ±18.0% |
| PUSH30 | 17.742 | 11.638 | -34.40% | 6.104 | 10.0% | 1.7% | ±5.0% | ±6.2% | ±20.1% |
| PUSH31 | 18.207 | 11.602 | -36.28% | 6.605 | 8.4% | 2.8% | ±5.0% | ±5.9% | ±16.7% |
| PUSH5 | 17.507 | 11.774 | -32.75% | 5.733 | 10.9% | 1.9% | ±5.0% | ±6.6% | ±21.8% |
| PUSH6 | 17.050 | 11.523 | -32.42% | 5.527 | 7.6% | 2.9% | ±5.0% | ±5.8% | ±15.3% |
| PUSH7 | 17.277 | 11.686 | -32.36% | 5.591 | 9.5% | 2.7% | ±5.0% | ±6.4% | ±19.0% |
| PUSH9 | 17.511 | 11.614 | -33.68% | 5.897 | 10.9% | 3.7% | ±5.0% | ±7.6% | ±21.9% |

- [x] Optimization
- [x] Refactoring

#### Requires testing

- [X] No

#### Notes on testing

Existing test suites cover the affected paths:
- `Nethermind.Evm.Test` - 4136/4136 passing (full opcode coverage + tracing)
- `Ethereum.Blockchain.Block.Test` - 1146/1146 passing (includes `StrangeContractCreation`, `OOGStateCopyContainingDeletedContract`, `wallet2outOf3txs`, `walletReorganizeOwners` 
- `Ethereum.Legacy.VM.Test` - `loopMul_*` passing
- `EvmOpcodesBenchmark` - sanity of per-opcode cost after rewrite

- [x] No

- [x] Yes

Performance-focused rewrite of the EVM stack and instruction dispatch - per-opcode wins on PUSH, POP, DUP, SWAP, arithmetic, and hot CALL/CREATE paths.


benaadams:
@claude review

benaadams:
@claude re-review

benaadams:
@claude re-review

benaadams:
@claude re-review

benaadams:
@claude re-review

LukaszRozmej:
## Perf-focused review (additional follow-ups)

Reviewed the full diff. Overall: strong, well-thought-out rewrite — 32-byte-aligned pinned stack, `ref byte` core, AVX-512 VBMI byte-swap, amortised multi-out pops, return-based errors, PUSH2+JUMP fusion, prefetch at jump destinations. Comments below are additional perf angles to consider — nothing is blocking.

### High-value follow-ups

1. **`InstructionShift` does two sequential pops with two bounds checks** — `EvmInstructions.Shifts.cs:62-73`. Its sibling `InstructionSar` already uses the amortised `PopUInt256(out a, out b)` and only branches on `a >= 256` afterwards. Mirror that pattern in Shift:
   ```csharp
   if (!stack.PopUInt256(out UInt256 a, out UInt256 b)) goto StackUnderflow;
   if (a >= 256) return stack.PushZero<TTracingInst>();
   ```
   Saves one bounds check and one offset recompute on every SHL/SHR (very hot).

2. **Replace-top opportunity for pop-N-push-1 patterns.** After `PopUInt256(out a, out b)` + `PushUInt256(in result)` the head net-delta is `-1`; the push's overflow check is statically impossible. Consider a `stack.ConsumeAndPush256(popCount, in UInt256)` helper that writes to slot `[newHead]` and increments, skipping the overflow check. Covers the whole `Math2Param` / `Math3Param` / `Shifts` / `Sar` / `Exp` family, MCOPY, CodeCopy, Create (3-pop + push-zero). The branch is well-predicted today, but it's pure waste and code size.

3. **AVX path in `PopUInt256(out a, out b[, c, d])` runs per-value dependency chains** — `EvmStack.cs:1735-1760`, `:1830-1863`, `:1951-1990`. Current shape is `load → vpshufb → vpermq` serialised per value, one after the next. The scalar path already interleaves across values (nice comment on breaking dependency chains). Mirror that in the AVX path: all loads first, then all shuffles, then all permutes. Modern Intel/Zen can keep multiple `vmovdqu` in flight, but only one `vpshufb`/`vpermq` per cycle — explicit interleave lets the OOO window schedule more aggressively instead of depending on JIT reordering.

4. **AVX-512 double-pack.** On AVX-512F (not just VBMI.VL), two 256-bit words fit in one zmm. The 4-out `PopUInt256` could do 2 × 512-bit load+permute instead of 4 × 256-bit. CALL/STATICCALL/DELEGATECALL all use the 4-out pop, so non-trivial win on call-heavy contracts.

5. **PUSH1+JUMP / PUSH1+JUMPI fusion** — currently `InstructionPush2` (`EvmInstructions.Stack.cs:120`) is the only fused path. Small loops (`JUMPDEST ... PUSH1 x JUMPI`) are extremely common in Solidity-emitted bytecode, and PUSH1 destinations are sufficient for jump targets in contracts ≤256 bytes (still common). Copying the fusion pattern for PUSH1 would be ~30 extra lines and catch a real workload.

6. **`OpCodeCopy.GetCode` still walks `vm.VmState.Env.CodeInfo.CodeSpan`** — `EvmInstructions.CodeCopy.cs:115`. Now that `EvmStack` hoists `Code` / `CodeLength`, switch CODECOPY to read from the stack struct as CODESIZE already does (`Environment.cs:282`).

### Smaller observations

7. **30 near-identical `OpN` structs** in `EvmInstructions.Stack.cs:208-862`. Pure boilerplate; good candidate for a source generator or T4. No perf impact.

8. **`ByteSwap256Mask` getter** assumes JIT folds the `Vector256.Create(const, const, const, const)` into a PC-relative rodata load. On .NET 9 this generally works; a `static readonly byte[] _mask = [...]; Vector256.LoadUnsafe(ref MemoryMarshal.GetArrayDataReference(_mask))` pattern is JIT-explicit and guaranteed. Worth confirming against a release-build disasm.

9. **`TestJumpCondition`** (`ControlFlow.cs:138`) reads the popped slot as `Vector256<byte>` via `Unsafe.As`. Correct (stack is 32-byte-aligned), but on machines without `Vector256.IsHardwareAccelerated`, this hits SW emulation. A scalar `ulong`-OR fallback would preserve speed on ARM64 without SVE.

10. **`InstructionMStore`** (`Storage.cs:143`) still pops the offset (`PopUInt256`) and word (`PopWord256`) with two bounds checks. A combined "pop one UInt256 + one raw 32-byte ref" helper would bring it in line with the new amortised pops.

11. **`stack.Head < EvmStack.MaxStackSize - 1` guard in the PUSH2+JUMP fusion** is correctly preserving the stack-overflow semantics a non-fused PUSH2 would have produced on a full stack — worth a one-line comment since it's easy to read as redundant.

12. **Tracing order in `Dup`** — `EvmStack.cs:2261` calls `Trace(depth)` before `Head = head`. Behaviourally matches the pre-PR code for DUP (the duplicated slot's value is identical before and after the write), but worth verifying against the tracing tests under `Push2JumpFusionTests` / existing DUP coverage.

Nothing above suggests the PR should be held. Items 1–5 are the biggest remaining wins; the rest are polish.

benaadams:
Thanks for the thorough review. Status per item below — TL;DR: 1, 2, 6, 10, 11 landed; 5 deferred with usage data; 3, 4, 8, 9, 12 rejected or verified-no-change.

### Fixed in this PR

- **1. Shift amortised pop + limb `>= 256` check.** (`5618ad3a`) Mirrored the `InstructionSar` two-out pop pattern. Replaced `a >= 256` (full 256-bit operator compare) with `(a.u1 | a.u2 | a.u3) != 0 || a.u0 >= 256`, matching the `InstructionByte` idiom introduced earlier in this PR.

- **2. Pop-N-push-1 peek-and-write.** (`e5174056`) Added `Pop1Peek32Bytes(out UInt256, out bool)` and `Pop2Peek32Bytes(out UInt256, out UInt256, out bool)` returning `[UnscopedRef] ref byte` to the new top slot. Also added `ReadUInt256FromSlot(ref byte, out UInt256)` out-param overload. JitAsm on `InstructionMath3Param<OpAddMod>`: **288 → 234 bytes (-18.8%)**, eliminating the 32-byte return-value staging round-trip that a by-value `UInt256` return forces through a hidden pointer. Math2Param and Math3Param rewritten to peek-read-compute-write in place — no push overflow check.

- **6. Split `InstructionCodeCopy` / `InstructionCallDataCopy`.** (`be375d42`) Dropped the `IOpCodeCopy<TGasPolicy>` interface + both `OpCodeCopy`/`OpCallDataCopy` generic structs. Shared core factored into a private `DataCopy` helper. CODECOPY now sources bytes via `MemoryMarshal.CreateReadOnlySpan(ref stack.Code, stack.CodeLength)` — verified in ASM as two direct field loads (`[rax+0x10]` / `[rax+0x1C]`), eliminating the full `vm.VmState.Env.CodeInfo.CodeSpan` traversal. CALLDATACOPY still pays one `MemoryManager.GetSpan()` call (unavoidable — `InputData` is `ReadOnlyMemory<byte>`).

- **10. MSTORE amortised pop.** (`096c41af`) Added `PopUInt256AndWord256(out UInt256, out Span<byte>)` with a single bounds check. MSTORE uses it — resolves the `PopUInt256` try-pattern vs `PopWord256()` throw-pattern inconsistency too.

- **11. PUSH2 fusion overflow-guard comment.** (`096c41af`) Added a two-line comment explaining why `Head < MaxStackSize - 1` is required (preserves the `StackOverflow` a non-fused PUSH2 would raise at `head == 1024`, even though the following JUMP would immediately pop).

### Deferred to a focused follow-up PR

- **5. PUSH1+JUMP{I} fusion.** Real win, ~30 lines mirroring the existing PUSH2 fusion. Measured usage before committing:
  - Ethereum consensus test corpus (2,516 GeneralStateTests): PUSH1+JUMP 24,814 + PUSH1+JUMPI 13,289 = 38,103 static pairs vs PUSH2+JUMP{I} 182,935. **PUSH1 share: 17.2% static**.
  - Top-50 mainnet contracts ([tcb0/opcode-usage](https://github.com/tcb0/opcode-usage)): PUSH2 count (16,525) exceeds total JUMP+JUMPI (13,708), so the vast majority of mainnet jumps are already PUSH2. Realistic mainnet static share: **5-15%**.
  - Dynamic weight higher because loop back-edges tend to fit PUSH1 range — plausibly **10-25% of executed jumps**.
  - Caveat: once EOF RJUMP/RJUMPI ships on mainnet the pattern disappears — this fusion has a shelf life.

### Rejected after investigation

- **3. Interleave loads/shuffles/permutes in multi-out `PopUInt256`.** JitAsm on `InstructionSar` shows the JIT already emits `vpermb ymm, ymm, [mem]` (single memory-operand permute) and the OoO engine interleaves FP12 (permute) and FP45 (store) uops across the two values. Manual C#-level interleaving cannot improve on this.

- **4. AVX-512 zmm double-pack in 4-out pop.** Zen4 throttles 512-bit + pre-IceLake downclock makes this a likely regression on the broader hardware fleet. Would need per-microarch benchmark data before committing.

- **8. `ByteSwap256Mask` getter rodata folding.** Already verified via JitAsm during this PR's optimize-asm work — `Vector256.Create(const, const, const, const)` does fold to a PC-relative rodata load (`RWD00 dq 18191A1B1C1D1E1Fh, ...`). No action needed.

### Verified correct — no change needed

- **9. `TestJumpCondition` ARM64 SW fallback.** `Vector256.IsHardwareAccelerated == false` on ARM64 but .NET falls back to two Vector128 ops, not pure scalar SW. Not worth a forked code path for the secondary platform.

- **12. Dup tracing order.** `Trace()` at `EvmStack.cs:2414` reads slots via the OLD `Head` value (pre-increment); matches the pre-PR behaviour exactly. Reviewer noted "matches the pre-PR code" — confirmed.

### Bonus fixes alongside

- `PeekBytesByRef` A/B tested NullRef sentinel vs `ref Word + out bool`. NullRef wins for single-output ref-returns (58 vs 67 bytes on `InstructionMath1Param<OpNot>`); `out bool` wins for multi-output helpers (free ABI-wise among other outs). Kept `PeekBytesByRef` as-is.
- Unused `PopPeekBytesByRef` removed (zero callers, superseded by `Pop1Peek32Bytes`).
- `.github/workflows/evm-opcode-benchmark-diff.yml` (`f7de7f29`): BDN 0.16+ rejects repeated `--filter` flags ("Option 'f, filter' is defined multiple times") — rerun steps now pass one `--filter` followed by multiple glob patterns as positional args.

benaadams:
@claude re-review

benaadams:
@claude re-review

LukaszRozmej (dismissed):
I think the AVX-512 change could benefit Zen 5 and later AMD architectures (and potential future Intel architectures - it is bound to make a comeback there in the next one - Nova Lake late this year)

src/Nethermind/Nethermind.Evm/EvmStack.cs
```diff
@@ -28,248 +28,1535 @@ public ref struct EvmStack
...
+    private static Word ByteSwap256Mask
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        get
+        {
+            return Vector256.Create(
+                0x18191a1b1c1d1e1ful,
+                0x1011121314151617ul,
+                0x08090a0b0c0d0e0ful,
+                0x0001020304050607ul).AsByte();
```
LukaszRozmej, line 89: should this be static variable?

benaadams: Jit embeds Vector256.Create constants as a data location in binary so is kinda a static variable automatically but better

AnkushinDaniil:
A few follow-ups on top of the existing review threads:

### Correctness / consistency

1. **Finish the return-based pop migration.** The throwing `PopWord256()` overload is still reached from 7 call sites — LOG topics in `EvmInstructions.Stack.cs:1106`, plus occurrences in `EvmInstructions.Storage.cs`, `EvmInstructions.Math1Param.cs`, and `EvmInstructions.Create.cs`. Since the goal is to get throws out of the opcode path, either convert these call sites or delete the throwing overload so it can't be reintroduced.

2. **`VmState.cs:269–275` comment is misleading.** It reads as an imperative ("The input array should be pinned…"). The actual invariant is that `StackPool.RentStacks()` allocates via `GC.AllocateUninitializedArray<byte>(…, pinned: true)` on the POH, so the pointer is stable and taking it doesn't create a GC hole. Rewording it as an invariant (optionally backed by a `Debug.Assert`) would prevent a future caller from silently breaking the SIMD alignment assumption by passing a non-pinned array.

### Docs

3. **XML `<summary>` docs on the new public surface of `EvmStack.cs`.** Roughly 15 of ~79 new public members currently have docs. Given this surface is ref-struct + generic + SIMD, it's exactly the code where docs matter most for future readers.

### Testing

Unit coverage is thorough, but a few boundary/parity gaps worth closing:

4. `Head == 1023` push boundary — exercises the off-by-one on `MaxStackSize - 1` checks in `PushUInt32/64/256`.
5. EIP-8024 `EXCHANGE` — exhaustive coverage over the full 256-byte operand range, not just representative picks.
6. AVX2/AVX-512 disabled-parity — re-run the scalar/vector cases with `DOTNET_EnableHWIntrinsic=0` (and an AVX-512 feature disable) so the scalar fallback is proven bitwise-equivalent to the vector path.
7. Fused `PUSH2+JUMPI` at `Head == 0` and `Head == 1`.
8. Fusion at end-of-code (PUSH2 with truncated immediate immediately followed by `JUMP`/`JUMPI`).

### Ops / rollout

9. **Publish a block-replay delta from `run-expb-reproducible-benchmarks.yml`** on this PR — ideally mainnet + at least one L2 payload set. BDN microbenchmarks are necessary but not sufficient for a consensus-critical rewrite this wide.
10. **Kill-switch for PUSH2+JUMP(I) fusion.** A small config flag that falls back to the unfused path. If a regression surfaces post-merge, operators can disable fusion without reverting the PR.

### Maintainability

11. **Split `EvmStack.cs` (2455 lines) into partials.** Suggested cut: Pop / Push / Dup+Swap / SIMD helpers / boundary checks. The common-style rule caps files at 800 lines and this is the single largest consensus-path hot file after the rewrite.
12. **Consolidate `Push2JumpFusionTests` and `EvmStackTests` into table-driven `[TestCase]` families.** A lot of methods are near-duplicates.
13. **Differential fuzz harness (follow-up, not a blocker).** Random bytecode, pre-PR vs post-PR, assert identical execution trace / gas / final state. Cheap insurance for a consensus-path rewrite.

### Nits

14. In `PushUInt32/64/256`: form `ref Word head` **after** the overflow check rather than before. Currently latent because the buffer over-allocates by `Vector256<byte>.Count`, but the invariant is local-and-obvious if the check precedes the ref.
15. In `PushBytesRef()`: advance `Head` **after** the write rather than before returning the ref. All current callers write before the next stack op, but inverting the order makes the invariant local instead of contract-based.

AnkushinDaniil (approved):

LukaszRozmej (approved):

PR #11240, opened 2026-04-19
perf(state): avoid copy on BulkSet via ArrayPoolList ownership transfer

asdacap:
## Changes

- Add `ArrayPoolList<T>.ToRef()` that hands the underlying rented array to an `ArrayPoolListRef<T>` without copying, then marks the source disposed so its own `Dispose` is a no-op. Guarded to only work with `SafeArrayPool<T>.Shared` (the pool `ArrayPoolListRef` returns to).
- Replace the `new ArrayPoolListRef<…>(_bulkWrite.AsSpan())` + explicit `_bulkWrite.Dispose()` pattern with `_bulkWrite.ToRef()` in:
  - `StateTree.StateTreeBulkSetter.Dispose` (state trie bulk write)
  - `TrieStoreScopeProvider.StorageTreeBulkWriteBatch.Dispose` (storage trie bulk write; reused by both the pruning trie-store path and the flat world state via `FlatStorageTree`)

Net effect: `BulkSet` now receives the original rented buffer directly instead of a freshly-copied one, on both the state and storage hot paths, for both pruning and flat world-state backends.

## Types of changes

- [x] Optimization
- [ ] Bugfix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes

#### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

#### Notes on testing

Behavior-preserving — `BulkSet` receives the same elements in the same order; only the allocation path changes. Covered by existing `PatriciaTreeBulkSetterTests` and state/storage integration tests. Reproducible-benchmarks workflow should show state/storage bulk-write improvements.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Core/Collections/ArrayPoolList.cs
+++ b/src/Nethermind/Nethermind.Core/Collections/ArrayPoolList.cs
@@ -264,6 +264,37 @@ public sealed class ArrayPoolList<T> : IList<T>, IList, IOwnedReadOnlyList<T>
         GuardDispose();
         return _array.AsSpan(0, _count);
     }
+
+    /// <summary>
+    /// Transfers ownership of the underlying rented array to a new <see cref="ArrayPoolListRef{T}"/>
+    /// without copying. After this call the <see cref="ArrayPoolList{T}"/> is marked disposed and
+    /// its own <see cref="Dispose"/> becomes a no-op; the returned ref struct is responsible for
+    /// returning the array to the pool.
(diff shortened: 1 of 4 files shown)


LukaszRozmej (approved):

LukaszRozmej:
@claude review

benaadams (approved):

PR #11244, opened 2026-04-19
test(snap): dispose AccountRangeMessage in snap tests

JayeTurn:
## Changes

- Dispose the `AccountRangeMessage` instances created in both `SnapMessageLimitsTests` and `AccountRangeMessageSerializerTests`
- Release the `ArrayPool`-backed account lists and proofs through the normal `AccountRangeMessage.Dispose()` path in direct roundtrip tests
- Use `try/finally` around `SerializerTester.TestZero` so serializer tests still clean up owned message resources when assertions fail

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [x] Other: _Test resource management fix_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


src/Nethermind/Nethermind.Network.Test/P2P/Subprotocols/Snap/Messages/AccountRangeMessageSerializerTests.cs
```diff
@@ -46,7 +46,15 @@ public void Roundtrip_NoAccountsNoProofs()
...
-            SerializerTester.TestZero(serializer, msg);
+            try
+            {
+                SerializerTester.TestZero(serializer, msg);
+                msg = null;
+            }
+            finally
+            {
+                msg?.Dispose();
+            }
```
LukaszRozmej: This seems wrong to me. Either SerializerTest should clean up even if it fails (so finally in SerializerTest) or not clean up at all.

JayeTurn: @LukaszRozmej Thanks for catching this. That makes sense.

I ended up fixing the symptom too locally in the snap tests, while `SerializerTester.TestZero()` was already the place that owned message disposal on the happy path. The right fix is to keep that ownership in the helper and make it cover the failure path as well.

Updated it that way now and removed the per-test cleanup from `AccountRangeMessageSerializerTests`. Thanks for the pointer.

LukaszRozmej (approved):

LukaszRozmej:
@claude review

LukaszRozmej:
@claude review

LukaszRozmej (approved):

flcl42 (approved):

LukaszRozmej (approved):

PR #11248, opened 2026-04-20
Fix flaky tests in PruningScenariosTests

benaadams:
We use process random seeds for hash functions which can cause issues with exact counts

## Changes

- Restore `Reorganization.MaxDepth` in `When_Reorg_OldValueIsNotRemoved` via try/finally and mark the test   `[NonParallelizable]`, so sibling tests never observe the mutated global.
- Replace the brittle `AssertThatCachedNodeCountMoreThan(275)` in `Retain_Some_PersistedNodes` with invariant-based   bounds:
    - memory stays at or above the configured retention portion (`retentionFloor = (long)(200.KiB * 0.1)`, matching
  `WithPrunePersistedNodeParameter(1, 0.1)`),
    - cached nodes fall below the no-prune baseline of `951` from `Keep_PersistedNode_EvenAfterPersist`.
  - Assert `thresholdReached.Should().BeTrue()` before the post-loop checks so the test fails loudly if the loop never
  exercises the pruning path.
  - Add `AssertThatCachedNodeCountLessThan` helper to mirror the existing `AssertThatCachedNodeCountMoreThan`.

#### Requires testing

- [x] No - is tests

### DIFF
--- a/src/Nethermind/Nethermind.Trie.Test/PruningScenariosTests.cs
+++ b/src/Nethermind/Nethermind.Trie.Test/PruningScenariosTests.cs
@@ -1179,8 +1179,8 @@ namespace Nethermind.Trie.Test
 
             ctx
                 .AssertThatDirtyNodeCountIs(9)
-                .AssertThatTotalMemoryUsedIsNoLessThan(retentionFloor)
-                .AssertThatCachedNodeCountLessThan(951);
+                .AssertThatCachedNodeCountLessThan(951)
+                .AssertThatTotalMemoryUsedIsNoLessThan(retentionFloor);
         }
 
         [Test]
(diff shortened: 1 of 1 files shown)


benaadams:
@claude review

benaadams:
@claude review

asdacap (approved):

src/Nethermind/Nethermind.Trie.Test/PruningScenariosTests.cs
```diff
@@ -1151,17 +1167,20 @@ public void Retain_Some_PersistedNodes()
...
                 }
             }
 
+            thresholdReached.Should().BeTrue();
+
             ctx
                 .AssertThatDirtyNodeCountIs(9)
-                .AssertThatCachedNodeCountMoreThan(275);
+                .AssertThatCachedNodeCountLessThan(951)
+                .AssertThatTotalMemoryUsedIsNoLessThan(retentionFloor);
```
AnkushinDaniil: It seems to be a duplicate check

AnkushinDaniil (approved):

src/Nethermind/Nethermind.Trie.Test/PruningScenariosTests.cs
benaadams: Fixed

PR #11249, opened 2026-04-20
Store TotalDifficulty in a skip-indexed block info store

asdacap:
## Rationale

The skip-indexed store answers two questions in O(log N) each:

1. **Total difficulty at a block** — walk the skip path from the block back to genesis, summing partial range sums. `TD(N) = Σ SkipCumulativeDifficulty(i)` along a path of length ⌈log₂ N⌉.
2. **N-th ancestor of a block** — follow skip pointers toward the target until distance fits; the same skip-distance layout that powers TD also powers ancestor lookup.

Both operations are **concurrent-safe by construction** and **parallelizable**:

- Each entry `(Difficulty, SkipCumulativeDifficulty, SkipParentHash)` is **immutable** — derived purely from the header chain. Recomputing an entry anywhere in the chain always yields the same bytes, so a concurrent writer double-populating a slot is a no-op rather than a race. There is no mutable state to maintain, so there can be no concurrent-write bug class around it.
- The DB is a lazy cache, not a source of truth, so the store survives arbitrary eviction, partial population, or writer crashes — recomputation will re-derive the same value.
- The migration backfill already exploits this, populating disjoint block ranges across CPU cores without coordination.

The n-th-ancestor answer is what canonicity checks really need: `IsMainChain(block)` can be expressed as "is `block` the N-th ancestor of `Head`?", and `IsFinalized(block)` as "is `block` an ancestor of the finalized head?". Today those use `ChainLevelInfo` / canonical-hash scans that are O(depth) and hold writer locks; replacing them with ancestor-queries against the skip-indexed store gives O(log N), lock-free reads.

## Changes

- Add `SkipIndexedBlockInfoStore` — a Fenwick-style (`n & -n` skip-distance) store of per-block difficulty plus skip-range partial sums under a new `skipIndexedBlockInfo` DB. `GetTotalDifficulty` and `GetAncestorAt` run in O(log N). Entries are computed lazily from `IHeaderStore` on first access and persisted; `CumulativeBlockInfoMigration` backfills existing DBs at startup.
- Move `ITotalDifficultyStrategy` from `Nethermind.Synchronization` into `Nethermind.Blockchain.SkipIndexedBlockInfo` and add a `GetDifficulty(BlockHeader)` hook the store uses to compute per-block contributions. `FixedTotalDifficultyStrategy` overrides it to inject a negative difficulty at the reset boundary so the store returns `TD = 0` across the Optimism Bedrock TD reset.
- Remove `BlockHeader.TotalDifficulty`. Consumers that used to read the header property go through `ISkipIndexedBlockInfoStore` (or the `IBlockTree.GetTotalDifficulty` default-interface proxy) instead. `HeaderValidator.ValidateTotalDifficulty` and the matching overrides in `Merge`/`Taiko`/`Xdc` validators are removed (the check became circular once TD lives in the store).
- Add `ITotalDifficultyAnchor` so live-sync callers can cap store walks at a configured pivot instead of walking the skip path all the way to genesis. `SyncPivotResolver` updates the anchor as the pivot advances.
- Retire the `SyncConfig.FixTotalDifficulty*` flags and `TotalDifficultyFixMigration`; the legacy `ChainLevelInfo`-based TD repair path is obsolete now that the store is the source of truth.
- Remove `WithTotalDifficulty` test builder methods and the RPC `totalDifficulty: 0` fallback in `BlockForRpc` when no value is available.

`TotalDifficulty` references in non-test code:

| | master | this PR |
| --- | ---: | ---: |
| Total references | 568 | 433 |
| Writes / maintenance | 44 (header/block mutations) | 10 (p2p wire / peer-claim / metrics only) |
| Real reads | ~240 (scattered `header.TotalDifficulty`) | ~80 (most routed through `GetTotalDifficulty`) |

## Types of changes

- [x] Refactoring

#### Requires testing

- [x] Yes

- [x] Yes

#### Notes on testing

- `SkipIndexedBlockInfoStoreTests` covers the skip-list TD algebra, ancestor lookups, missing-header handling, the ClockCache, and the Bedrock-style reset path via `FixedTotalDifficultyStrategy`.
- `SkipIndexedBlockInfoMigrationTests` covers the startup backfill.
- Existing Blockchain / Consensus / Sync / Merge / Era / Clique / Xdc tests were updated to source TD from the store.

- [ ] No

- [x] Yes

`BlockHeader.TotalDifficulty` is gone. Anything embedding Nethermind that reaches into that property must switch to `ISkipIndexedBlockInfoStore.GetTotalDifficulty` / `IBlockTree.GetTotalDifficulty`. The legacy `SyncConfig.FixTotalDifficulty*` flags and `TotalDifficultyFixMigration` are removed; on first start against an existing DB the new `CumulativeBlockInfoMigration` populates the skip-indexed store.

## Remarks

Stacked on top of #`db-migration-check`. Target is that branch, not master.


asdacap:
Will have to redo

PR #11152, opened 2026-04-15
fix build - BenchmarkDotNet Version

svlachakis:
https://github.com/NethermindEth/nethermind/actions/runs/24450622701/job/71438308593?pr=11150

This job seems to be failing on all open PRs

PR #11253, opened 2026-04-20
fix build - BenchmarkDotNet Version

AnkushinDaniil:
## Summary

Bump `BenchmarkDotNet` and `BenchmarkDotNet.Diagnostics.Windows` from `0.16.0-nightly.20260320.468` to `0.16.0-nightly.20260322.472`.

The pinned `.468` build was unpublished from NuGet and is no longer restorable, which fails the Build matrix on master and every open PR with `NU1603` (treated as error). NuGet auto-resolves to `.472` as the nearest available nightly.

Recurrence of the same failure mode as #11152 (which bumped `.466` → `.468` five days ago).

## CI error

```
error NU1603: Warning As Error: Nethermind.Precompiles.Benchmark.Test depends on
BenchmarkDotNet (>= 0.16.0-nightly.20260320.468) but BenchmarkDotNet
0.16.0-nightly.20260320.468 was not found. BenchmarkDotNet
0.16.0-nightly.20260322.472 was resolved instead.
```

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -9,8 +9,8 @@
     <PackageVersion Include="AspNetCore.HealthChecks.UI.InMemory.Storage" Version="9.0.0" />
     <PackageVersion Include="Autofac" Version="9.1.0" />
     <PackageVersion Include="Autofac.Extensions.DependencyInjection" Version="10.0.0" />
-    <PackageVersion Include="BenchmarkDotNet" Version="0.16.0-nightly.20260320.468" />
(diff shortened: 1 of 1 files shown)


svlachakis (approved):
We have fixed that 3 times last week, should we consider using stable release? @LukaszRozmej @benaadams @kamilchodola ?

LukaszRozmej (approved):

benaadams:
> We have fixed that 3 times last week, should we consider using stable release? @LukaszRozmej @benaadams @kamilchodola ?

When the next make a stable release

PR #9677, opened 2025-11-10
Fix history prune scheduling

LukaszRozmej:
Resolves #9697

## Changes

- Rename `BackgroundTaskScheduler.ScheduleTask` -> `TryScheduleTask` and avoid exceptions
- In `HistoryPruner` add checks before going into lock and doing pruning.
- In `HistoryPruner` make pruning task exclusive - don't schedule another before previous isn't finished.
- Remove lock from `BackgroundTaskScheduler`
- Move `CancelledToken` out of generic type
- Specify channel options explicitly

#### What types of changes does your code introduce?

- [x] Bugfix (a non-breaking change that fixes an issue)
- [x] Optimization
- [x] Refactoring

#### Requires testing

- [x] Yes

#### If yes, did you write tests?

- [x] No

#### Notes on testing

- Test normal node operation (`BackgroundTaskScheduler` won't be throwing exceptions which can affect networking a bit?)
- Test History pruning

PR #11254, opened 2026-04-20
HistoryPruner: self-bootstrap in TryPruneHistory instead of relying on SyncServer

AnkushinDaniil:
## Changes

- `HistoryPruner.TryPruneHistory`: remove the redundant pre-lock `!_hasLoadedDeletePointer` gate so the pruner self-bootstraps on first entry instead of depending on `SyncServer.OnNewRange` to call `OldestBlockHeader` at a `%32` block boundary. The inner path at `HistoryPruner.cs:309` already calls `TryLoadDeletePointer()` and handles failure — the outer check only served to suppress bootstrap until an external caller primed it.

#### Requires testing

- [x] Yes

- [x] Yes

- [ ] No

- [ ] No

### Root cause

On a node whose head never crosses a `%32` block boundary after startup (frozen snapshot, peerless, no CL attached), history pruning never runs even when `--History.Pruning=Rolling` is configured. `_hasLoadedDeletePointer` only flips when `TryLoadDeletePointer()` runs, and in production that only happens via the `OldestBlockHeader` getter, which is only called from `SyncServer.OnNewRange` (`SyncServer.cs:472`) gated on `block.Number % 32 == 0` (`SyncServer.cs:469`). Nothing inside `HistoryPruner` observes or controls that condition, so the pruner waits on an external event that may never fire.

### Why safe

- Concurrency unchanged: `_currentlyPruning` atomic still prevents concurrent pruning tasks; `_pruneLock` still serializes actual work inside `TryPruneHistory`.
- Cost: one extra `Monitor.TryEnter`/`Monitor.Exit` per trigger during the brief startup window before bootstrap — one-time, not steady state. After first load, `TryLoadDeletePointer()` early-returns at line 531 with no DB read.
- Pre-lock `ShouldPruneHistory` call remains valid before bootstrap: reads `_deletePointer` (default 0) and `_lastPrunedTimestamp` (default null); both branches return sensibly and let the flow fall into the lock path where bootstrap + prune happen atomically.

### Relationship to #9677

The pre-lock gate was introduced in #9677 as a fast-path optimization. This PR removes only that one line and preserves every other improvement from #9677 (atomic `_currentlyPruning`, exclusive scheduling, `TryScheduleTask` rename)

### DIFF
--- a/src/Nethermind/Nethermind.History.Test/HistoryPrunerTests.cs
+++ b/src/Nethermind/Nethermind.History.Test/HistoryPrunerTests.cs
@@ -193,9 +193,9 @@ public class HistoryPrunerTests
         CheckHeadPreserved(testBlockchain, blocks);
     }
 
-    [TestCase(0, 100000, false)]
-    [TestCase(100, 10, true)]
-    public void Validates_config(int minHistoryRetentionEpochs, int retentionEpochs, bool shouldThrow)
+    [TestCase(0, 100000u, false)]
+    [TestCase(100, 10u, true)]
+    public void Validates_config(int minHistoryRetentionEpochs, uint retentionEpochs, bool shouldThrow)
     {
         IHistoryConfig historyConfig = new HistoryConfig
         {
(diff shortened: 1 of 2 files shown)


LukaszRozmej (dismissed):
@claude can HistoryPrunerTests be deduplicated with TestCase/TestCaseSource?

LukaszRozmej (approved):

svlachakis (approved):

Issue #8480, opened 2025-04-07
eth_simulateV1 does not respect blockOverrides


natebeauregard:
A partner has been seeing weird behavior when trying to use `blockOverrides` in `eth_simulateV1` on Nethermind nodes. Below is a description of his test setup and the issue he is seeing.

cc @spypsy
_________________________________
Forwarded message from https://demerzelsolutions.slack.com/archives/C04Q02A9HGA/p1743547103512009

In order to test simulateV1 I did a very simple setup:
Typescript test:
```
import { createPublicClient, createWalletClient, decodeErrorResult, encodeAbiParameters, http } from 'viem';
import { privateKeyToAccount } from 'viem/accounts';
import { foundry } from 'viem/chains';

describe('eth_simulateV1', () => {
  const rpcUrl = 'http://127.0.0.1:32780/';
  const PRIVATE_KEY = ('0x' + 'c5114526e042343c6d1899cad05e1c00ba588314de9b96929914ee0df18d46b1') as `0x${string}`;
  const DEPLOYER_PRIVATE_KEY = '0xbcdf20249abf0ed6d944c0288fad489e33f66b3960d9e6229c1cd214ed3bbe31' as `0x${string}`;

  // Create a custom chain with the correct chain ID
  const chain = {
    ...foundry,
    id: 3151908,
  };

  const publicClient = createPublicClient({
    transport: http(rpcUrl),
    chain,
  });

  const walletClient = createWalletClient({
    account: privateKeyToAccount(PRIVATE_KEY),
    transport: http(rpcUrl),
    chain,
  });

  const deployerWalletClient = createWalletClient({
    account: privateKeyToAccount(DEPLOYER_PRIVATE_KEY),
    transport: http(rpcUrl),
    chain,
  });

  // Contract that has a method checking if a timestamp has passed
  // This contract has a target timestamp and returns true if block.timestamp >= target
  const TIMESTAMP_CHECKER_BYTECODE =
[...]

PR #11267, opened 2026-04-20
eth estimategas & eth call does not support block overrides

### LINKED ISSUE #11171: `eth_call`: 4th parameter (`blockOverrides`) not supported
### Description

Geth supports an optional 4th parameter for `eth_call` — `blockOverrides` — that allows callers
to override block-level fields (number, timestamp, gasLimit, baseFeePerGas, feeRecipient, random,
miner) for the simulation.

Nethermind's `eth_call` signature only accepts 3 parameters
`(transaction, blockParameter, stateOverride)` and returns `-32602 Invalid params` with
`"Incorrect parameters count, expected: 3, actual: 4"` when a 4th argument is provided.

### Steps to Reproduce

```json
// test_30 — eth_call with blockOverrides: {number, timestamp, gasLimit}
{
  "jsonrpc": "2.0",
  "method": "eth_call",
  "params": [
    {
      "from": "0xEF04bc7821433f080461BBAE815182E3d7bBb61A",
      "to": "0x4debB0dF4da8D1f51EF67B727c3F1c0eCC7ed009",
      "gas": "0x7A1200",
      "value": "0x0",
      "data": "0x70a08231..."
    },
    "0x89543F",
    {},
    {
      "number": "0x89543F",
      "timestamp": "0x68E0F100",
      "gasLimit": "0x7E1200"
    }
  ],
  "id": 1
}
```

### Expected (Geth)

```json
{ "result": "0x" }
```

### Actual (Nethermind)

```json
{
  "error": {
    "code": -32602,
    "data": "Incorrect parameters count, expected: 3, actual: 4",
    "message": "Invalid params"
  }
}
```

### blockOverride fields used across the 8 failing tests

| Test   | blockOverride fields                        |
|--------|---------------------------------------------|
| test_30 | number, timestamp, gasLimit                |
| test_31 | time                                        |
| test_32 | number, timestamp, random                   |
| test_33 | number, timestamp, baseFeePerGas            |
| test_34 | number, gasLimit, baseFeePerGas             |
| test_35 | number, timestamp, gasLimit, miner          |
| test_36 | timestamp                                   |
| test_38 | number, feeRecipient, baseFeePerGas         |

### References

- [Geth eth_call docs](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-eth#eth-call)
- EIP-1193, execution-apis spec

### LINKED ISSUE #11183: eth_estimateGas: Nethermind does not support block overrides
### Description

When eth_estimateGas is called with four params (transaction, blockTag, stateOverrides, blockOverrides), Nethermind rejects the request with Invalid params and reports expected: 3, actual: 4 as it does not support block overrides ,while Geth is able to give correct estimates as it supports it.

#### test_25
``` json
{
  "jsonrpc": "2.0",
  "method": "eth_estimateGas",
  "params": [
    {
      "from": "0xa9Ac1233699BDae25abeBae4f9Fb54DbB1b44700",
      "to": "0x252568abdeb9de59fd8963dfcd87be2db65f1ce1",
      "gasPrice": "0xBA43B7400"
    },
    "0xE401C0",
    {},
    {
      "timestamp": "0x667FB2B0",
      "gasLimit": "0x3D0900"
    }
  ],
  "id": 1
}
```

#### test_26
``` json
{
  "jsonrpc": "2.0",
  "method": "eth_estimateGas",
  "params": [
    {
      "from": "0xa9Ac1233699BDae25abeBae4f9Fb54DbB1b44700",
      "to": "0x252568abdeb9de59fd8963dfcd87be2db65f1ce1",
      "gasPrice": "0xBA43B7400",
      "data": "0x01234567"
    },
    "0xE401C0",
    {
      "0xa9Ac1233699BDae25abeBae4f9Fb54DbB1b44700": {
        "balance": "0x56bc75e2d63100000"
      }
    },
    {
      "baseFeePerGas": "0x400000000"
    }
  ],
  "id": 1
}
```

#### test_25
``` json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": "0x5309"
}
```

#### test_26
``` json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": "0x5349"
}
```
#### test_25
``` json
{
  "jsonrpc": "2.0",
  "error": {
    "code": -32602,
    "message": "Invalid params",
    "data": "Incorrect parameters count, expected: 3, actual: 4"
  },
  "id": 1
}
```
#### test_26
``` json
{
  "jsonrpc": "2.0",
  "error": {
    "code": -32602,
    "message": "Invalid params",
    "data": "Incorrect parameters count, expected: 3, actual: 4"
  },
  "id": 1
}
```

### Notes

Both failures are consistent with Nethermind enforcing a maximum of 3 params for
eth_estimateGas in this path, while Geth supports the 4-parameter form that
includes blockOverrides.

DarkLord017:
Fixes Closes Resolves #11183 , #11171 , #8480

## Changes

- Added BlockOverride to both Eth_Call and Eth_EstimateGas with adding coinbase override 
- So now both endpoint supports 4 parameters instead of three previously 
- Built a unified decorator for BlobBaseFeeOverrides for both simulate & eth_*
- Added tests for blob , opcode changes and in gasestimate to verify the changes 
- Added test for open issue 8480 , but it was passing now so it is already resolved , but just added that as a test :/

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


DarkLord017:
@claude review pls 
Also should i add more tests and is the impl correct for block overrides and do any other endpoint needs the impl or not ?

DarkLord017:
@claude review pls 
is there a better way to handle blob parameters as it is not in the header ,  this one's too complicated

DarkLord017:
@claude review

DarkLord017:
@claude re-review 
1. I did not get the high you were talking about 
2. can u suggest way to test if blobbasefeeoverride is applied on eth call and estimate gas

DarkLord017:
@claude re-review

DarkLord017:
@claude re-review

DarkLord017:
@claude re-review also i have added code to use overrided gas limit but is already set by ApplyOverride() why did i had to apply again , it is correct behaviour right

DarkLord017:
> **Claude finished @DarkLord017's task in 4m 17s** —— [View job](https://github.com/NethermindEth/nethermind/actions/runs/24795271182)
> 
> ### PR Re-Review: eth_estimateGas & eth_call Block Overrides
> * [x]  Read current state of changed files
> * [x]  Answer the GasLimit dual-application question
> * [x]  Verify previous findings
> * [x]  Check for new issues
> * [x]  Post final review
> 
> ### Answering Your Question: Why Apply GasLimit in Two Places?
> **They set two different things** — `transactionCall.Gas` and `clonedHeader.GasLimit` are not the same field.
> 
> In `EstimateGasTxExecutor.Execute`:
> 
> ```cs
> transactionCall.Gas = (long)BlockOverride.GasLimit.Value;  // → tx.GasLimit
> ```
> 
> This sets the **transaction's declared gas limit** — the upper bound of the binary search in `eth_estimateGas`. Without this, the binary search would range up to the real block gas limit (e.g. 30M on mainnet) even when the override says 50K. The test `Estimate_gas_block_override_gasLimit_bounds_estimation` verifies exactly this.
> 
> Later, `ApplyOverrides(clonedHeader)` sets **the block header's gas capacity** — what the `GASLIMIT` opcode (0x45) returns inside the EVM. The `Eth_call_with_block_override` GASLIMIT test case verifies this.
> 
> So yes, **both applications are correct and necessary**:
> 
> What's set	Where	Why
> `tx.GasLimit` (= `transactionCall.Gas`)	`EstimateGasTxExecutor.Execute`	Binary search upper bound
> `header.GasLimit`	`ApplyOverrides(clonedHeader)`	EVM `GASLIMIT` opcode value
> For `eth_call`, only `ApplyOverrides` runs (no override of `transactionCall.Gas`) because `eth_call` defaults to `gasCap` via `EnsureDefaults` and doesn't need to bound a binary search.
> 
> ### Status of All Previous Findings
> Finding	Severity	Status
> Missing `[Test]` on `Eth_call_blobBaseFeePerGas_override_test`	High	✅ Resolved — `[Test]` attribute at line 864
> `eth_call` defaults gas to `blockGasLimit` instead of `gasCap`	High	✅ Resolved — base class only sets gas when `_blockOverride?.GasLimit is not null`
> Deleted BASEFEE opcode regression tests	Medium	✅ Resolved — `Eth_call_with_base_fee_opcode_should_return_0` at lines 270-293 still present
> `JToken.BeEquivalentTo(string)` assertion	Medium	✅ Resolved — line 567 uses `["result"]!.Value<string>().Should().Be("0x5208")`
> `IBlobBaseFeeOverrideProvider` unification	Medium (design)	✅ Resolved — unified decorator serves both simulate and single-call paths
> Dual-path `blobBaseFeeOverride` redundancy	Medium	✅ Resolved — `CallAndRestore` reads `components.RequestState.BlobBaseFeeOverride` directly
> Unused import `System.Reflection.Metadata`	Low	✅ Resolved — no longer present
> Duplicate comment line in `BlockOverride.cs`	Low	✅ Resolved
> Hardcoded address in test	Low	✅ Resolved — uses `TestItem.AddressA`
> `GasLimit` overflow guard before cast	Medium	✅ Correct — `ExecuteTx` checks before setting `_blockOverride`
> ### Remaining Open Items
> #### Low — `Eth_call_blobBaseFeePerGas_override_test` leaks original `ctx.Test`
> **File**: `EthRpcModuleTests.EthCall.cs`, lines 864–904
> 
> ```cs
> using Context ctx = await Context.Create(specProvider);
> // ...
> ctx.Test = await TestRpcBlockchain.ForTest(...).Build(specProvider);  // ← replaces ctx.Test
> ```
> 
> The original `ctx.Test` created by `Context.Create` is replaced but not disposed. When `ctx` is finally disposed (via `using`), it disposes only the replacement. The first `TestRpcBlockchain` leaks. A simpler approach is to skip `Context.Create` entirely and just build the blockchain directly:
> 
> ```cs
> using TestRpcBlockchain test = await TestRpcBlockchain
>     .ForTest(SealEngineType.NethDev)
>     .WithBlockFinder(blockTree)
>     .Build(specProvider);
> string withOverride = await test.TestEthRpc("eth_call", ...);
> ```
> 
> #### Low — `eth_createAccessList` still does not support `blockOverride`
> This was noted in earlier reviews as out-of-scope. It's a known gap — `eth_createAccessList` at line 393 does not forward a `blockOverride` parameter, so access lists computed with block-context-sensitive state (e.g. `BLOCKHASH`, `TIMESTAMP`) won't reflect any block override. Worth a follow-up issue.
> 
> ### What Works Well
> The implementation is correct and complete for the stated scope:
> 
> * **GasLimit dual-application** is intentional and correct — transaction gas (search bound) and header gas (GASLIMIT opcode) serve different roles.
> * **`BlobBaseFeeOverrideCalculatorDecorator`** correctly handles both channels: total blob cost (via `TryCalculateBlobBaseFee`) and per-gas fee-cap validation (via `BlockExecutionContext.BlobBaseFee` in `CallAndRestore`).
> * **`IBlobBaseFeeOverrideProvider`** cleanly unifies single-call and simulate paths.
> * **`eth_call` gas default** correctly uses `gasCap` (no gas set in base `TxExecutor` without an override); only `EstimateGasTxExecutor` sets gas from block/override gas limit.
> * **Block override tests** cover NUMBER, TIMESTAMP, GASLIMIT, COINBASE, BASEFEE, and BLOBBASEFEE opcodes.
> * The overflow guard (`GasLimit > long.MaxValue`) in `ExecuteTx` fires before `_blockOverride` is stored, so the unsafe cast in `EstimateGasTxExecutor.Execute` is safe.
> 
> **Verdict**: mergeable. All previous High and Medium findings are resolved. The two remaining items are Low (a test resource leak and a known out-of-scope gap). The core implementation is correct.

Access List endpoint does not support blockoverrides even in go-ethereumj

DarkLord017:
@claude re-review

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
```diff
@@ -149,13 +149,14 @@ public bool TryGetTransaction(Hash256 txHash, [NotNullWhen(true)] out Transactio
             return blockHash is not null ? receiptStorage.Get(blockHash).ForTransaction(txHash) : null;
         }
 
-        public CallOutput Call(BlockHeader header, Transaction tx, Dictionary<Address, AccountOverride>? stateOverride, CancellationToken cancellationToken)
+        public CallOutput Call(BlockHeader header, Transaction tx, Dictionary<Address, AccountOverride>? stateOverride, UInt256? blobBaseFeeOverride, CancellationToken cancellationToken)
         {
             using Scope<BlockProcessingComponents> scope = processingEnv.BuildAndOverride(header, stateOverride);
+            scope.Component.RequestState.BlobBaseFeeOverride = blobBaseFeeOverride;
```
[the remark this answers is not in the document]
svlachakis: @claude is this design simplification resolved?

svlachakis (approved):

src/Nethermind/Nethermind.Evm/BlockOverride.cs
```diff
@@ -31,8 +31,16 @@ public void ApplyOverrides(BlockHeader result)
         }
 
         if (Number is not null) result.Number = (long)Number.Value;
-        if (FeeRecipient is not null) result.Beneficiary = FeeRecipient;
+        if (FeeRecipient is not null)
+        {
+            // Set Author as well because GasBeneficiary = Author ?? Beneficiary.
+            // Mirrors geth: blockCtx.Coinbase = *o.FeeRecipient.
+            result.Author = FeeRecipient;
+            result.Beneficiary = FeeRecipient;
```
LukaszRozmej:
```suggestion
            result.Beneficiary = result.Author = FeeRecipient;
```

src/Nethermind/Nethermind.Facade/BlobBaseFeeOverrideCalculatorDecorator.cs
```diff
@@ -0,0 +1,35 @@
...
+                overrideProvider.BlobBaseFeeOverride.Value,
+                out blobBaseFee);
+        }
+
+        return blobBaseFeeCalculatorBase.TryCalculateBlobBaseFee(
+            header,
+            transaction,
+            blobGasPriceUpdateFraction,
+            out blobBaseFee);
+    }
```
LukaszRozmej:
```suggestion
        out UInt256 blobBaseFee) =>
            overrideProvider.BlobBaseFeeOverride is not null
                ? !UInt256.MultiplyOverflow(BlobGasCalculator.CalculateBlobGas(transaction), overrideProvider.BlobBaseFeeOverride.Value, out blobBaseFee)
            : blobBaseFeeCalculatorBase.TryCalculateBlobBaseFee(header, transaction, blobGasPriceUpdateFraction, out blobBaseFee);
```

LukaszRozmej (approved):

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.TransactionExecutor.cs
```diff
@@ -57,15 +61,22 @@ public override ResultWrapper<TResult> Execute(
...
                 transactionCall.EnsureDefaults(_rpcConfig.GasCap);
 
                 return base.Execute(transactionCall, blockParameter, stateOverride, searchResult);
             }
 
-            public ResultWrapper<TResult> ExecuteTx(TransactionForRpc transactionCall, BlockParameter? blockParameter, Dictionary<Address, AccountOverride>? stateOverride = null)
-                => Execute(transactionCall, blockParameter, stateOverride);
+            public ResultWrapper<TResult> ExecuteTx(TransactionForRpc transactionCall, BlockParameter? blockParameter, Dictionary<Address, AccountOverride>? stateOverride = null, BlockOverride? blockOverride = null)
+            {
+                if (blockOverride?.GasLimit > (ulong)long.MaxValue)
```
benaadams: Should this be linked to config rather than be 2Bn?

benaadams (approved):

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.TransactionExecutor.cs
LukaszRozmej: in JsonRpcConfig we do have max gas cap for RPC calls

DarkLord017: Applied pls see @LukaszRozmej @benaadams

DarkLord017:
@claude review if the merge conflict was applied correctlly

svlachakis (approved):

PR #11273, opened 2026-04-21
perf(trie): skip TrieNode wrapping for extension children already unresolved to Hash256

asdacap:
## Changes

- `EncodeExtension` now checks `_nodeData[0]` directly and encodes the child's hash inline when it is already a `Hash256` (the state left by `UnresolveChild` during pruning).
- Previous flow went through `GetChildWithChildPath`, which calls `tree.FindCachedOrUnknown` to wrap the hash back into a `TrieNode`, then `ResolveKey` (a no-op because `Keccak` was already set). That work is now skipped on the hot path.
- Mirrors the existing pattern used by branch encoding (`WriteChildrenRlpBranchNonRlp`, `GetChildrenRlpLengthForBranchNonRlp`), where a `Hash256` entry in `_nodeData[i]` is encoded directly.
- Non-`Hash256` cases (null, `_nullNode` / empty from RLP, resolved `TrieNode`, and the rare inline-child `Keccak == null` case for RLP < 32 bytes) fall through to the unchanged path.

## Types of changes

- [x] Optimization

#### Requires testing

- [x] No

#### Notes on testing

Exercised via the existing Trie test suites:

- `Nethermind.Trie.Test` — same pass/fail counts as `master` baseline (two pre-existing flaky failures reproduce without this change).
- `Nethermind.State.Test` — 710 passed, 0 failed.

No behavior change for consumers; the encoded RLP is byte-identical in both paths.

- [x] No

- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Trie/TrieNode.Decoder.cs
+++ b/src/Nethermind/Nethermind.Trie/TrieNode.Decoder.cs
@@ -52,11 +52,11 @@ namespace Nethermind.Trie
                 // Fast path: child was unresolved to a Hash256 (e.g. by pruning) — encode the hash directly
                 // without materializing a TrieNode via FindCachedOrUnknown + ResolveKey.
                 Hash256? childKeccak = item._nodeData[0] as Hash256;
-                CappedArray<byte> childFullRlp = default;
+                TrieNode? nodeRef = null;
                 if (childKeccak is null)
                 {
                     int previousLength = item.AppendChildPath(ref path, 0);
-                    TrieNode nodeRef = item.GetChildWithChildPath(tree, ref path, 0);
+                    nodeRef = item.GetChildWithChildPath(tree, ref path, 0);
                     Debug.Assert(nodeRef is not null,
                         "Extension child is null when encoding.");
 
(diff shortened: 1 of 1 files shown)


src/Nethermind/Nethermind.Trie/TrieNode.Decoder.cs
```diff
@@ -49,15 +49,25 @@ public static CappedArray<byte> EncodeExtension(TrieNode item, ITrieNodeResolver
...
                 HexPrefix.CopyToSpan(hexPrefix, isLeaf: false, keyBytes);
 
-                int previousLength = item.AppendChildPath(ref path, 0);
-                TrieNode nodeRef = item.GetChildWithChildPath(tree, ref path, 0);
-                Debug.Assert(nodeRef is not null,
-                    "Extension child is null when encoding.");
+                // Fast path: child was unresolved to a Hash256 (e.g. by pruning) — encode the hash directly
+                // without materializing a TrieNode via FindCachedOrUnknown + ResolveKey.
+                Hash256? childKeccak = item._nodeData[0] as Hash256;
+                CappedArray<byte> childFullRlp = default;
```
LukaszRozmej: If we re-arranged this method I think we can drop `childFullRlp`, saving some stack space, copying and a branch

LukaszRozmej (dismissed):

src/Nethermind/Nethermind.Trie/TrieNode.Decoder.cs
asdacap: Good call — done in e079f10138. Holding a `TrieNode?` reference instead of a `CappedArray<byte>` local drops the 12-byte struct slot and the copy on the slow path; fast path is unchanged.

(Note from Claude acting on @asdacap's behalf.)

LukaszRozmej (approved):

batrr (approved):

PR #9182, opened 2025-08-20
Parallel transaction execution with BALs

Marchhill:
Depends on BALs https://github.com/NethermindEth/nethermind/pull/9114

#### What types of changes does your code introduce?

- [ ] Bugfix (a non-breaking change that fixes an issue)
- [x] New feature (a non-breaking change that adds functionality)
- [ ] Breaking change (a change that causes existing functionality not to work as expected)
- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

#### If yes, did you write tests?

- [x] Yes
- [ ] No

#### Requires documentation update

- [ ] Yes
- [x] No

#### Requires explanation in Release Notes

- [ ] Yes
- [x] No

PR #11276, opened 2026-04-21
feat: BAL recorder plugin — record and replay block access lists for prewarming benchmarks

asdacap:
## Summary

- Adds `Nethermind.BalRecorder`, a plugin that records block-level access lists (BAL) to era flat-files during live processing and replays them as prewarming hints on a subsequent run
- BAL tracing is gated via an `ISpecProvider` decorator (`BalRecorderSpecProvider` + `BalRecorderReleaseSpec`) that toggles `IsEip7928Enabled` per-block via an `AsyncLocal` switch, so background flows (prewarmer, eth_call) are unaffected
- `BalRecordingBlockProcessor` flips the switch and records `GeneratedBlockAccessList` to the store; `BalRecordingBlockValidator` strips the artificially-set `BlockAccessListHash` from processed pre-Prague blocks before validation
- Plugin is enabled only when `RecordingEnabled` or `ReplayEnabled` is set in `IBalRecorderConfig`
- Era flat-file store (`SlotStore` / `SlotFile`) organises BAL data in slot-per-era files with per-era locking
- `ParallelWorldState.BeginScope` now clears `GeneratedBlockAccessList` and `_suggestedBlockAccessList` so stale BAL state from a previous block cannot leak into the next block if `LoadSuggestedBlockAccessList` is not called

## Changes

- **`Nethermind.BalRecorder`** (new project): plugin, config, store, spec decorator, block processor/validator decorators
- **`Nethermind.Runner`**: registers the plugin
- **`Nethermind.State`**: `ParallelWorldState.BeginScope` clears BAL state

## Test plan

- [ ] `dotnet test Nethermind.BalRecorder.Test` — 17 tests including E2E record→replay round-trip
- [ ] `dotnet test Nethermind.Blockchain.Test` — includes `ParallelWorldState_BeginScope_clears_suggested_and_generated_bal`
- [ ] `dotnet build Nethermind.slnx -c release` — full solution builds clean


src/Nethermind/Nethermind.BalRecorder/BalRecorderPlugin.cs
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Nethermind.BalRecorder;
+
+public class BalRecorderPlugin(IBalRecorderConfig config) : INethermindPlugin
+{
+    public string Name => "BalRecorder";
+    public string Description => "Records and replays block access lists as era files for prewarming benchmarks";
+    public string Author => "Nethermind";
+    public bool Enabled => config.RecordingEnabled || config.ReplayEnabled;
+    public IModule? Module => new BalRecorderModule();
```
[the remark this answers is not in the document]
asdacap: Module not loaded when not enabled. This is a non issue.

src/Nethermind/Nethermind.BalRecorder/BalRecordingBlockProcessor.cs
```diff
@@ -0,0 +1,55 @@
...
+    public event Action? TransactionsExecuted
+    {
+        add => inner.TransactionsExecuted += value;
+        remove => inner.TransactionsExecuted -= value;
+    }
+
+    public (Block Block, TxReceipt[] Receipts) ProcessOne(Block suggestedBlock, ProcessingOptions options, IBlockTracer blockTracer, IReleaseSpec spec, CancellationToken token)
+    {
+        if (store.ReplayEnabled && suggestedBlock.BlockAccessList is null && suggestedBlock.Hash is not null)
+            suggestedBlock.BlockAccessList = store.Get(suggestedBlock.Number, suggestedBlock.Hash);
```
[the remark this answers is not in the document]
asdacap: It just how it works man.... the whole thing have mutation everywhere.

Marchhill:
could make sense to build on https://github.com/NethermindEth/nethermind/pull/9182

asdacap:
Should be easy to integrate, it does not touch main code.

LukaszRozmej:
@claude re-review

LukaszRozmej (approved):

LukaszRozmej (commented):
Did a deep review pass; main findings inline. Overall the `AsyncLocal<bool?>` + `ISpecProvider`-decorator gate is a clean way to keep BAL plumbing out of `BlockProcessor`. Top-level concerns:

1. **PR description mismatch.** The body claims `Nethermind.State`: `ParallelWorldState.BeginScope` clears `GeneratedBlockAccessList`/`_suggestedBlockAccessList`, plus a test `ParallelWorldState_BeginScope_clears_suggested_and_generated_bal` in `Nethermind.Blockchain.Test`. Neither is in the diff (and `ParallelWorldState` doesn't exist on master — the equivalent class is `BlockAccessListBasedWorldState`, whose `BeginScope` is a one-line passthrough). Either include the State change with its test or drop the claim from the description.
2. **Decorator scope.** `AddDecorator<IBlockProcessor>` at root wraps every resolution — main processor, block producer, simulate (`eth_call`), stateless/witness — see inline.
3. **Header mutation.** `BalRecordingBlockValidator.ValidateProcessedBlock` mutates `processedBlock.Header.Hash` — see inline.
4. **`SlotFile` read/write race** on the in-memory header — see inline.

Nothing here is necessarily a hard blocker; mostly worth a discussion before merge.

src/Nethermind/Nethermind.BalRecorder/BalRecorderPlugin.cs
```diff
@@ -0,0 +1,32 @@
...
+
+public class BalRecorderModule : Module
+{
+    protected override void Load(ContainerBuilder builder) =>
+        builder
+            .AddSingleton<IRecordedBalStore, RecordedBalStore>()
+            .AddSingleton<BalRecorderSpecSwitch>()
+            .AddDecorator<ISpecProvider, BalRecorderSpecProvider>()
+            .AddDecorator<IBlockValidator, BalRecordingBlockValidator>()
+            .AddDecorator<IBlockProcessor, BalRecordingBlockProcessor>();
```
LukaszRozmej, line 36: `AddDecorator<IBlockProcessor>` registered at the root container wraps **every** `IBlockProcessor` resolved from any child scope, not just the canonical block processor. That includes:
- block producer (`GlobalWorldStateBlockProducerEnvFactory`)
- simulate / `eth_call` (`SimulateReadOnlyBlocksProcessingEnvFactory`)
- stateless executor / witness generation
- AuRa / Optimism / Taiko / Xdc custom processors

With `RecordingEnabled=true`, all of those would call `store.Insert(block.Number, …)`. Combined with `SlotFile.TryWrite`'s silent first-write-wins, the first scope that records a given block wins and the rest are silently dropped — so reorgs, restarts mid-recording, and main-vs-simulate races all collide on the same slot key.

Worth either scoping the decorator to the main processor (Autofac key, or registering inside `MainProcessingContext`'s lifetime scope only), or explicitly documenting that the plugin must only be enabled on a node configured for benchmarking and not serving JSON-RPC simulate traffic.

Same consideration applies to the `IBlockValidator` and `ISpecProvider` decorators on the lines above.

src/Nethermind/Nethermind.BalRecorder/BalRecordingBlockValidator.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            // The suggested block is a pre-Prague draft (no BlockAccessListHash in its sealed header).
+            // Our spec switch caused SetBlockAccessList to populate BlockAccessListHash on the
+            // processed block, which would fail the per-field comparison. Strip it and recompute
+            // the hash so the Hash == Hash short-circuit fires instead. This is safe because the
+            // SHA-3 pre-image guarantee means equal hashes imply all other header fields are equal.
+            // Note: store.Insert() in BalRecordingBlockProcessor runs after this method returns and
+            // uses block.Number as the key, so the stripped hash does not affect BAL persistence.
+            processedBlock.Header.BlockAccessListHash = null;
+            processedBlock.Header.Hash = processedBlock.Header.CalculateHash();
```
LukaszRozmej, line 43: Mutating `processedBlock.Header.Hash` after the fact is a footgun: the processed block flows downstream to the block tree, broadcast, JSON-RPC by-hash lookups, etc. The comment correctly notes this is internally consistent due to the SHA-3 pre-image, but the side effect propagates beyond this method.

The guard (`balSwitch.Enabled && suggestedBlock.Header.BlockAccessListHash is null`) limits the path to "we forced 7928 on for a pre-Prague draft" — i.e. block production on a chain where 7928 isn't yet active. But the plugin loads on any chain when configured, and the only thing standing between it and a real-world production node is the user setting `RecordingEnabled=true`.

Could we sidestep this entirely by not having `SetBlockAccessList` populate `BlockAccessListHash` when the inner spec says 7928 is off and only the recorder switch flipped it on? That avoids the post-hoc fixup and the implied invariant that no downstream code keys on the now-stripped hash.

src/Nethermind/Nethermind.BalRecorder/SlotFile.cs
```diff
@@ -0,0 +1,81 @@
...
+        else
+        {
+            RandomAccess.Read(_handle, _header, 0);
+        }
+    }
+
+    public bool TryRead<TArg>(int slot, ReadOnlySpanAction<byte, TArg> action, TArg arg)
+    {
+        ReadOnlySpan<byte> entry = _header.AsSpan(slot * 8, 8);
+        uint offset = BinaryPrimitives.ReadUInt32BigEndian(entry);
```
LukaszRozmej: `TryRead` reads `_header` without taking `_writeLock`. A concurrent `TryWrite` publishes the entry via two separate `WriteUInt32BigEndian` calls (lines 73-74) under the lock, but the reader doesn't synchronize.

Observable interleavings:
- New offset already written, size still 0 → returns `true` with a 0-length span → RLP decode throws → swallowed by `RecordedBalStore.Get` as "Corrupt BAL slot".
- Stale offset 0 vs. new size → returns `false` even though data is on disk.

The parallel-write test (`ConcurrentWrites_SameEra_DoNotCorrupt`) covers concurrent writes to *different* slots only.

Fix options: read the 8-byte entry as a single 64-bit `Volatile.Read` (publish via a single 64-bit write under the lock), or take `_writeLock` on the read path — the lock is uncontended in benchmark workflows so the cost is negligible.

src/Nethermind/Nethermind.BalRecorder/SlotFile.cs
```diff
@@ -0,0 +1,81 @@
...
+        if (offset == 0) return false;
+
+        int size = (int)BinaryPrimitives.ReadUInt32BigEndian(entry[4..]);
```
LukaszRozmej: Unchecked `(int)` cast: a corrupt era file with size ≥ 2 GB yields negative int → `ArrayPool<byte>.Shared.Rent(size)` throws `ArgumentOutOfRangeException`, propagated up. Cheap defense: `if (size <= 0 || size > maxRecordSize) return false;` before the rent.

src/Nethermind/Nethermind.BalRecorder/RecordedBalStore.cs
```diff
@@ -0,0 +1,58 @@
...
+    private readonly ILogger _logger = logManager.GetClassLogger<RecordedBalStore>();
+    private readonly SlotStore _store = new(
+        config.Path.GetApplicationResourcePath(initConfig.BaseDbPath), "bal");
+
+    public bool ReplayEnabled => config.ReplayEnabled;
+    public bool RecordingEnabled => config.RecordingEnabled;
+
+    public void Dispose() => _store.Dispose();
+
+    public void Insert(Block block, BlockAccessList bal)
```
LukaszRozmej, line 24: Combined with `SlotFile.TryWrite`'s first-write-wins semantics and the broad `IBlockProcessor` decorator, this silently drops all but the first attempt to record a given block — across reorgs, restarts after partial runs, or main vs. simulate vs. producer scopes. The test `Insert_DoesNotOverwrite_ExistingSlot` confirms this is intentional, but the silent-drop is counterintuitive.

At minimum, surface a debug log when the slot was already filled so users debugging missing data don't have to reason through this from first principles. Better: have `Insert` return `bool` so callers can choose, or split into `TryInsert` / `Overwrite`.

src/Nethermind/Nethermind.BalRecorder/BalRecorderSpecSwitch.cs
```diff
@@ -0,0 +1,15 @@
+// SPDX-FileCopyrightText: 2026 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System.Threading;
+
+namespace Nethermind.BalRecorder;
+
+public class BalRecorderSpecSwitch
+{
+    // bool? so that disabling sets null (the default), which removes the entry from the ExecutionContext
```
LukaszRozmej: The "child async flows then have no entry and correctly see Enabled = false without inheriting a stale explicit override" rationale isn't quite right. `AsyncLocal<T>` propagation is *capture-on-launch*: child flows snapshot the parent's `ExecutionContext` at the moment they start, so whether the parent later sets the key to `null` or `false` doesn't affect already-running children — they see whatever was captured.

`AsyncLocal<bool?>` + `null`-on-disable and `AsyncLocal<bool>` + `false`-on-disable behave identically for every current caller. The current code is correct, but either tighten the explanation (the actual benefit, if any, is that an unused `AsyncLocal<bool?>` doesn't allocate an EC dictionary entry on `Enabled` reads) or simplify to `AsyncLocal<bool>` and set `value` directly.

src/Nethermind/Nethermind.BalRecorder/RecordedBalStore.cs
```diff
@@ -0,0 +1,58 @@
...
+
+    private sealed class ReadState
+    {
+        public BlockAccessList? Value;
+        public ILogger Logger;
+        public long BlockNumber;
+    }
+}
+
+public class NullRecordedBalStore : IRecordedBalStore
```
LukaszRozmej: `NullRecordedBalStore` is never registered or referenced anywhere outside this file. The plugin loader skips `BalRecorderModule` entirely when `Enabled=false`, so no consumer resolves `IRecordedBalStore` when the plugin is disabled. Either delete it or wire it as a default registration in a non-gated module so other code can depend on `IRecordedBalStore` regardless of plugin state.

src/Nethermind/Nethermind.BalRecorder/IRecordedBalStore.cs
```diff
@@ -0,0 +1,18 @@
...
+using Nethermind.Core.BlockAccessLists;
+using Nethermind.Core.Crypto;
+
+namespace Nethermind.BalRecorder;
+
+public interface IRecordedBalStore : IDisposable
+{
+    void Insert(Block block, BlockAccessList bal);
+    /// <summary>Retrieves the recorded BAL for <paramref name="blockNumber"/>. <paramref name="blockHash"/> is accepted for caller convenience but is not used as part of the key — the store is keyed by block number only and is not reorg-safe.</summary>
+    BlockAccessList? Get(long blockNumber, Hash256 blockHash);
```
LukaszRozmej: Accepting `blockHash` only to ignore it invites bugs at call sites. The xmldoc note is good but easy to miss. Either drop `blockHash` from the signature entirely, or add a debug-mode integrity check (e.g. record the hash alongside the BAL in the slot and assert on read) so reorg scenarios fail loudly instead of silently returning stale data.

src/Nethermind/Nethermind.BalRecorder/BalRecordingBlockProcessor.cs
```diff
@@ -0,0 +1,51 @@
...
```
LukaszRozmej: Worth a brief comment that the `GeneratedBlockAccessList` read here relies on `BlockAccessListManager.PrepareForProcessing` having called `Reset()` early in the inner `ProcessOne`, which in turn relies on `balSwitch.Enabled` having flipped *before* delegation. Otherwise the dependency chain (spec switch → `BlockLevelAccessListsEnabled` → `manager.Enabled` → `Reset()`) is implicit and easy to break in a future refactor.

LukaszRozmej:
@claude re-review

src/Nethermind/Nethermind.BalRecorder.Test/BalRecorderE2ETests.cs
```diff
@@ -0,0 +1,267 @@
...
+        spec.BerlinBlockNumber = 0;
+        spec.LondonBlockNumber = 0;
+        spec.ArrowGlacierBlockNumber = 0;
+        spec.GrayGlacierBlockNumber = 0;
+        ActivateAllParameterTransitionsFromGenesis(spec.Parameters);
+        ActivateAllEthashTransitionsFromGenesis(spec);
+    }
+
+    private static void ActivateAllParameterTransitionsFromGenesis(ChainParameters parameters)
+    {
```
flcl42, line 228: Can't be specific fork selected like Osaka.Instance and overrided a bit? Looks really like we should have something that would allow to avoid setting up the entire spec from scratch.

LukaszRozmej:
@claude re-review

src/Nethermind/Nethermind.BalRecorder/SlotStore.cs
```diff
@@ -0,0 +1,67 @@
...
+using System.IO;
+using System.Threading;
+
+namespace Nethermind.BalRecorder;
+
+/// <summary>
+/// Stores arbitrary byte blobs indexed by block number using a slot-file layout.
+/// Caches a single open <see cref="SlotFile"/> at a time to bound memory usage.
+/// </summary>
+public class SlotStore(string directory, string extension = "bin") : IDisposable
```
Marchhill, line 15: "Slot" name seems a bit vague

src/Nethermind/Nethermind.BalRecorder/BalRecordingBlockProcessor.cs
```diff
@@ -0,0 +1,53 @@
...
+                store.Insert(block, balManager.GeneratedBlockAccessList);
+            return (block, receipts);
+        }
+        finally
+        {
+            if (shouldFlip) balSwitch.Enabled = false;
+        }
+    }
+
+    private bool ShouldFlip(Block suggestedBlock)
```
Marchhill, line 47:
```suggestion
    private bool ShouldToggleBalRecording(Block suggestedBlock)
```

Marchhill (approved):

PR #11279, opened 2026-04-21
XDC Fix special transaction 

batrr:
These issues were not caught in tests because `XdcTestBlockchain` uses a genesis with `BaseFeePerGas = 0`, which diverges from real chain behavior. Fixing `BaseFeePerGas` in tests is out of scope for this PR.


## Changes                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                              
  ### 1. Fix sign transaction hash calculation                                                                                                                                                                                                
  `CalculateHash()` was called before `signer.Sign()`, so the hash didn't cover the signature. Moved to after signing.                                                                                                                        
                                                                                                                                                                                                                                              
  ### 2. Fix sign transaction eligibility check                                                                                                                                                                                               
  Now checks correct `snapshot.NextEpochCandidates` with penalized nodes.
                                                                                                                                                                                                                                              
  ### 3. Broadcast special transactions to peers
  `SignTransactionFilter` now sets `tx.IsServiceTransaction = true` so `tx.IsFree()` returns `true`, allowing special transactions to pass the `StartBroadcast` path and be gossiped to peers - not only included in locally produced blocks.  This is required when a block is penalized and the node needs peers to pick up the sign transaction.
                                                                                          
                                                                                                                                                                                                                                              
  ### 4. Include special transactions in produced blocks                                                                                                                                                                                      
  Two gaps prevented special transactions from appearing in blocks:                                                                                                                                                                           
  - `GasPrice = 0` sender buckets were dropped before block building. `XdcBlockProducer` overrides `PrepareBlock` with `filterSource: false` to fetch all pending transactions unfiltered                         
  - `BaseFeeTxFilter` rejected `GasPrice = 0` transactions during block production. `XdcTxFilterPipeline` passes the filter pipeline for special transactions.


src/Nethermind/Nethermind.Xdc/TxPool/XdcTxGossipPolicy.cs
```diff
@@ -14,6 +14,12 @@ public bool ShouldGossipTransaction(Transaction tx)
     {
         IXdcReleaseSpec spec = (IXdcReleaseSpec)provider.GetXdcSpec(chainHeadInfoProvider.HeadNumber);
```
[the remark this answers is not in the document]
batrr: XDC-only class — a non-XDC spec here is a programming error. Failing loudly is correct; silently returning true would gossip all transactions unchecked.

src/Nethermind/Nethermind.Xdc/TxPool/SignTransactionFilter.cs
```diff
@@ -50,6 +50,8 @@ public AcceptTxResult Accept(Transaction tx, ref TxFilteringState state, TxHandl
             return AcceptTxResult.Accepted;
         }
 
+        tx.IsServiceTransaction = true;
```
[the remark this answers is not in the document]
batrr: [Claude] The filter is the right place for this — transactions received from peers never go through CreateTxSign, so the filter is the only path that covers both local and peer-originated special transactions.
The mutation is safe because the filter runs before the transaction is inserted into the pool, so no reader can see it in an unmarked state.

src/Nethermind/Nethermind.Xdc/TxPool/XdcTxFilterPipeline.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Nethermind.Xdc.TxPool;
+
+internal class XdcTxFilterPipeline(ITxFilterPipeline inner) : ITxFilterPipeline
+{
+    public void AddTxFilter(ITxFilter txFilter) => inner.AddTxFilter(txFilter);
+
+    public bool Execute(Transaction tx, BlockHeader parentHeader, IReleaseSpec currentSpec)
+    {
+        if (currentSpec is IXdcReleaseSpec xdcSpec && tx.IsSpecialTransaction(xdcSpec))
```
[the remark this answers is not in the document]
batrr: - `MaxBlobCountBlobTxValidator`, `MempoolBlobTxProofVersionValidator` - XDC has no blobs
  - `GasLimitCapTxValidator` - XDC doesn't have enable EIP-7825

batrr:
@claude review. 
Note: regression tests added. Rest covered by `SpecialTransactionsTests` once `XdcTestBlockchain` gets non-zero `BaseFeePerGas`.

src/Nethermind/Nethermind.Xdc/TxPool/SignTransactionFilter.cs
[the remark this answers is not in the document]
batrr: Check the field usage for reference.

src/Nethermind/Nethermind.Xdc/TxPool/XdcTxGossipPolicy.cs
```diff
@@ -14,6 +14,12 @@ public bool ShouldGossipTransaction(Transaction tx)
...
 
-        return !tx.RequiresSpecialHandling(spec);
+        if (!tx.RequiresSpecialHandling(spec))
+            return true;
+
+        if (tx.IsSignTransaction(spec))
+            return true;
+
+        return false;
```
batrr, line 23: Not sure about this. Is it enough to broadcast only sign transactions?

cicr99 (approved):

src/Nethermind/Nethermind.Xdc/TxPool/XdcTxGossipPolicy.cs
ak88: Yes i think that is fine

src/Nethermind/Nethermind.Xdc/TxPool/SignTransactionFilter.cs
```diff
@@ -43,13 +34,20 @@ private AcceptTxResult ValidateSignTransaction(Transaction tx, long headerNumber
...
+        XdcBlockHeader header = (XdcBlockHeader)blockTree.Head.Header;
+        long headerNumber = header.Number + 1;
+        IXdcReleaseSpec spec = specProvider.GetXdcSpec(headerNumber);
 
         if (!tx.IsSpecialTransaction(spec))
         {
             return AcceptTxResult.Accepted;
         }
 
+        tx.IsServiceTransaction = true;
```
ak88: Why set it before we are sure we want to accept the tx?

src/Nethermind/Nethermind.Xdc/XdcBlockProducer.cs
```diff
@@ -97,6 +98,17 @@ protected override BlockHeader PrepareBlockHeader(BlockHeader parent, PayloadAtt
         return xdcBlockHeader;
     }
 
+    protected override BlockToProduce PrepareBlock(BlockHeader parent, PayloadAttributes? payloadAttributes = null, IBlockProducer.Flags flags = IBlockProducer.Flags.None)
+    {
+        BlockHeader header = PrepareBlockHeader(parent, payloadAttributes);
+
+        IEnumerable<Transaction> transactions = (flags & IBlockProducer.Flags.EmptyBlock) != 0 ?
+            Array.Empty<Transaction>() :
+            TxSource.GetTransactions(parent, header.GasLimit, payloadAttributes, filterSource: false);
```
ak88, line 107: This also by passes the check for matching nonce. Not sure this is correct.

src/Nethermind/Nethermind.Xdc/TxPool/SignTransactionFilter.cs
batrr: This is the last filter in order. Rejected txs are discarded anyway. 

Moved it before `return AcceptTxResult.Accepted` just in case.

src/Nethermind/Nethermind.Xdc/XdcBlockProducer.cs
batrr: `GetTransactions` is ordering/optimization, actual correctness checked later. We just lose some performance.

ak88 (approved):

PR #11283, opened 2026-04-21
style: enforce additional IDE style rules as warnings

newmanifold:
## Changes

- Enable 10 IDE style rules as warnings in `.editorconfig` and fix all existing violations across 41 files
- Net result: -33 lines, modernized patterns, no behavioral changes

### Rules added

| Rule | Name | Before | After |
|---|---|---|---|
| **IDE0018** | Inline variable declaration | `int x; TryGet(out x)` | `TryGet(out int x)` |
| **IDE0019** | Use pattern matching (`as` + null) | `var x = o as T; if (x != null)` | `if (o is T x)` |
| **IDE0029** | Use coalesce expression | `x != null ? x : y` | `x ?? y` |
| **IDE0034** | Simplify `default` expression | `default(CancellationToken)` | `default` |
| **IDE0062** | Make local function static | `void Local() { }` | `static void Local() { }` |
| **IDE0071** | Simplify interpolation | `$"{x.ToString()}"` | `$"{x}"` |
| **IDE0074** | Use compound assignment (`??=`) | `if (x == null) x = y;` | `x ??= y;` |
| **IDE0150** | Prefer null check over type check | `!(x is T)` | `x is null` |
| **IDE0270** | Null check can use coalesce throw | `if (x is null) throw; y = x;` | `y = x ?? throw` |
| **IDE1005** | Null-conditional delegate invocation | `if (h != null) h(args)` | `h?.Invoke(args)` |

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [ ] No

#### Notes on testing

Style-only changes — full solution build with `EnforceCodeStyleInBuild=true` passes with 0 errors and 0 warnings.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -108,6 +108,14 @@ dotnet_diagnostic.CA1510.severity = warning
 dotnet_diagnostic.CA2208.severity = suggestion
 dotnet_diagnostic.NETH002.severity = error
 dotnet_diagnostic.IDE0320.severity = warning
+dotnet_diagnostic.IDE0018.severity = warning
+dotnet_diagnostic.IDE0019.severity = warning
+dotnet_diagnostic.IDE0029.severity = warning
 dotnet_diagnostic.IDE0034.severity = warning
+dotnet_diagnostic.IDE0062.severity = warning
+dotnet_diagnostic.IDE0071.severity = warning
 dotnet_diagnostic.IDE0074.severity = warning
+dotnet_diagnostic.IDE0150.severity = warning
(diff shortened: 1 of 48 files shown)


newmanifold:
@claude review again, i have rebased

flcl42 (approved):

benaadams (approved):

PR #11285, opened 2026-04-21
Additional BAL hive tests

benaadams:
## Changes

Fixes a hive `eels/consume-rlp` failure where two Amsterdam EIP-7928 BAL tests incorrectly passed (`test_bal_gas_limit_boundary[below_boundary]` and `test_gas_limit_below_minimum[gas_limit_5000_1]`). Expands the Pyspec test framework so this class of invalid-block fixture is actually exercised, and fixes the latent validation gaps that expansion surfaced.

### The primary fix - EIP-7928 BAL gas-limit floor on the RLP path

`BlockDecoder` does not decode BAL bytes from the RLP body, so RLP-imported blocks (hive `consume-rlp`, p2p sync) always have `Block.BlockAccessList == null`. The pre-execution check in `ValidateBlockLevelAccessList` gates on a non-null BAL, so the `itemCount * GasCostOf.BlockAccessListItem <= gasLimit` rule was silently skipped. During processing the BAL is regenerated (`ParallelWorldState.GeneratedBlockAccessList`) and the header `BlockAccessListHash` matches — so the block was accepted despite the gas-limit floor being violated.

Fix: validate the floor against `processedBlock.GeneratedBlockAccessList` inside `ValidateProcessedBlock`, before the hash-equal early return, only on the RLP path (suggested BAL null) when `spec.BlockLevelAccessListsEnabled`.

### Test-framework structural fix

`BlockchainTestBase.DecodeRlps` only enrolled blocks where `testBlockJson.BlockHeader is not null`, which meant every EEST invalid-block fixture (they omit `blockHeader` because there's no canonical header for a block that must be rejected) was silently dropped - including the BAL boundary test. Moved `correctRlp.Add` out of the `BlockHeader` gate and removed the over-strict `"Expected to fail with X but it passed validation"` assertion (post-execution rules like state-root, BAL hash, and the BAL gas-limit floor cannot fail pre-validation; the end-of-test `LastBlockHash` check is authoritative).

### Latent bugs the framework expansion revealed, now fixed

- **Post-Merge consensus rules** (EIP-3675: difficulty=0, nonce=0, UnclesHash=Keccak(rlp([]))) weren't enforced in the test environment because `MergeHeaderValidator` was only wired when `isEngineTest`. Now wired for any post-Merge test (fixes ~60 uncle-related + 3 post-Paris difficulty failures).
- **Pre-Paris difficulty validation** wasn't happening because the test seal validator was `NullSealEngine`. Added `DifficultyOnlySealValidator` that implements the difficulty branch of `EthashSealValidator.ValidateParams` without the expensive PoW hash work (fixes 3 `wrongDifficulty_*` failures).
- **`HeaderValidator` NullReferenceException on genesis-level uncle headers**: `ValidateParent` accepts a null parent for genesis, but `ValidateGasLimitRange`, `ValidateTimestamp`, and `ValidateBlockNumber` unconditionally dereferenced it. Added null-parent guards 

#### Requires testing

- [x] Yes

- [x] Yes

#### Notes on testing

- New boundary regression in `BlockValidatorTests.ValidateProcessedBlock_Enforces_bal_item_gas_limit_boundary_for_rlp_imported_blocks` (`30_000` valid, `29_999` invalid) that fails before the validator fix and passes after.
- `Eip7928BlockChainTests` (Pyspec) now actually exercises `test_bal_gas_limit_boundary[below_boundary]`: fails with `LAST BLOCK HASH` mismatch without the validator fix, green with it.
- Full regression sweep: `Ethereum.Legacy.Blockchain.Block.Test` 2539/2539, `Ethereum.Blockchain.Block.Test` 1146/1146, `Eip7928BlockChainTests` 574/574, `CancunBlockchainTests` 8389/8389, `BlockValidatorTests` + `ShardBlobBlockValidatorTests` 39/39 — all green.

- [x] No

- [x] No

## Remarks

Hive dashboard run that reported the original BAL failures: https://hive.ethpandaops.io/#/test/bal/1776762759-6340c22976e6ae27be715bd116690199 (suite `eels/consume-rlp`, 2/44457 failed on `nethermind_default 1.37.0-unstable+29561c2a`).

The `HeaderValidator` null-parent guards are minimal defensive changes; they only take effect when an uncle's parent is absent from the block tree and the header is genesis-level (`Number == 0`), a narrow sliver that was crashing instead of correctly falling through.

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Validators/BlockValidator.cs
+++ b/src/Nethermind/Nethermind.Consensus/Validators/BlockValidator.cs
@@ -156,6 +156,25 @@ public class BlockValidator(
     /// <returns><c>true</c> if the <paramref name="processedBlock"/> is valid; otherwise, <c>false</c>.</returns>
     public bool ValidateProcessedBlock(Block processedBlock, TxReceipt[] receipts, Block suggestedBlock, out string? error)
     {
+        // EIP-7928: enforce the BAL item gas-limit floor against the BAL produced during
+        // execution. The pre-execution check in ValidateBlockLevelAccessList only fires when
+        // the suggested block carries a BAL (engine-API path); RLP-imported blocks have a
+        // null suggested BAL and the agreed gas limit must still be validated against the
+        // generated BAL here.
+        IReleaseSpec spec = _specProvider.GetSpec(suggestedBlock.Header);
+        if (spec.BlockLevelAccessListsEnabled
+            && suggestedBlock.BlockAccessList is null
+            && processedBlock.GeneratedBlockAccessList is not null)
+        {
+            int itemCount = processedBlock.GeneratedBlockAccessList.ItemCount();
+            if (itemCount * GasCostOf.BlockAccessListItem > suggestedBlock.Header.GasLimit)
+            {
+                error = BlockErrorMessages.BlockAccessListGasLimitExceeded(itemCount, suggestedBlock.Header.GasLimit);
+                if (_logger.IsWarn) _logger.Warn($"BAL item count {itemCount} exceeds block gas limit bound in block {suggestedBlock.ToString(Block.Format.FullHashAndNumber)}.");
+                return false;
+            }
+        }
+
         if (processedBlock.Header.Hash == suggestedBlock.Header.Hash)
         {
             error = null;
--- a/src/Nethermind/Nethermind.Consensus/Validators/BlockValidator.cs
+++ b/src/Nethermind/Nethermind.Consensus/Validators/BlockValidator.cs
@@ -161,16 +161,15 @@ public class BlockValidator(
         // the suggested block carries a BAL (engine-API path); RLP-imported blocks have a
         // null suggested BAL and the agreed gas limit must still be validated against the
         // generated BAL here.
-        IReleaseSpec spec = _specProvider.GetSpec(suggestedBlock.Header);
-        if (spec.BlockLevelAccessListsEnabled
-            && suggestedBlock.BlockAccessList is null
-            && processedBlock.GeneratedBlockAccessList is not null)
+        if (suggestedBlock.BlockAccessList is null
+            && processedBlock.GeneratedBlockAccessList is not null
+            && _specProvider.GetSpec(suggestedBlock.Header).BlockLevelAccessListsEnabled)
         {
             int itemCount = processedBlock.GeneratedBlockAccessList.ItemCount();
             if (itemCount * GasCostOf.BlockAccessListItem > suggestedBlock.Header.GasLimit)
             {
                 error = BlockErrorMessages.BlockAccessListGasLimitExceeded(itemCount, suggestedBlock.Header.GasLimit);
-                if (_logger.IsWarn) _logger.Warn($"BAL item count {itemCount} exceeds block gas limit bound in block {suggestedBlock.ToString(Block.Format.FullHashAndNumber)}.");
+                if (_logger.IsWarn) _logger.Warn($"Block level access list item count {itemCount} exceeds block gas limit bound in block {suggestedBlock.ToString(Block.Format.FullHashAndNumber)}.");
                 return false;
             }
         }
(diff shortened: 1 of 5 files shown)


benaadams:
@claude re-review

benaadams:
@claude re-review

benaadams:
@claude re-review

rubo (approved):

asdacap (approved):

src/Nethermind/Nethermind.Consensus/Validators/HeaderValidator.cs
```diff
@@ -144,6 +144,9 @@ protected virtual bool Validate1559(BlockHeader header, BlockHeader parent, IRel
 
         protected virtual bool ValidateBlockNumber(BlockHeader header, BlockHeader parent, ref string? error)
         {
+            // No parent-relative block-number check at genesis (parent is null only when ValidateParent already accepted genesis).
+            if (parent is null) return true;
```
LukaszRozmej, line 148: Can this not be called when parent is null at all?

src/Nethermind/Nethermind.Consensus/Validators/HeaderValidator.cs
```diff
@@ -256,6 +259,10 @@ protected virtual bool ValidateExtraData(BlockHeader header, IReleaseSpec spec,
 
         protected virtual bool ValidateGasLimitRange(BlockHeader header, BlockHeader parent, IReleaseSpec spec, ref string? error)
         {
+            // ValidateParent (called earlier in the chain) accepts a null parent only for genesis.
+            // Gas-limit-vs-parent comparisons don't apply at genesis, so skip them.
+            if (parent is null) return true;
```
LukaszRozmej, line 264: same here

src/Nethermind/Nethermind.Consensus/Validators/HeaderValidator.cs
```diff
@@ -290,6 +297,9 @@ protected virtual bool ValidateGasLimitRange(BlockHeader header, BlockHeader par
 
         protected virtual bool ValidateTimestamp(BlockHeader header, BlockHeader parent, ref string? error)
         {
+            // No parent-relative timestamp check at genesis (parent is null only when ValidateParent already accepted genesis).
+            if (parent is null) return true;
```
LukaszRozmej, line 301: same

src/Nethermind/Nethermind.Consensus/Validators/HeaderValidator.cs
```diff
@@ -301,6 +311,13 @@ protected virtual bool ValidateTimestamp(BlockHeader header, BlockHeader parent,
 
         protected virtual bool ValidateTotalDifficulty(BlockHeader header, BlockHeader parent, ref string? error)
         {
+            // Same class of null-parent exposure as ValidateBlockNumber/GasLimitRange/Timestamp:
+            // ValidateParent accepts null for genesis, but the non-zero-total-difficulty branch
+            // below dereferences parent. No current fixture triggers this (TotalDifficulty is
+            // typically null on uncle headers), but the guard keeps the method consistent with
+            // its siblings.
+            if (parent is null) return true;
```
LukaszRozmej, line 319: same

src/Nethermind/Nethermind.Consensus/Validators/BlockValidator.cs
```diff
@@ -156,6 +156,24 @@ public bool ValidateProcessedBlock(Block processedBlock, TxReceipt[] receipts, B
...
     public bool ValidateProcessedBlock(Block processedBlock, TxReceipt[] receipts, Block suggestedBlock, out string? error)
     {
+        // EIP-7928: enforce the BAL item gas-limit floor against the BAL produced during
+        // execution. The pre-execution check in ValidateBlockLevelAccessList only fires when
+        // the suggested block carries a BAL (engine-API path); RLP-imported blocks have a
+        // null suggested BAL and the agreed gas limit must still be validated against the
+        // generated BAL here.
+        if (suggestedBlock.BlockAccessList is null
+            && processedBlock.GeneratedBlockAccessList is not null
+            && _specProvider.GetSpec(suggestedBlock.Header).BlockLevelAccessListsEnabled)
```
LukaszRozmej, line 166: Do we need to get the spec here? Pre-processing validation should have already failed if `BlockLevelAccessListsEnabled` was false and `suggestedBlock.BlockAccessList is null`. This is `PostProcessing` validation.

PR #11297, opened 2026-04-22
Reduce dirty nodes lock contention

benaadams:
## Changes

- Hash quality: replace XOR-based hash combines in Hash256.GetChainedHashCode, HashAndTinyPath, and TrieStoreDirtyNodesCache.Key with BitOperations.Crc32C. XOR of two 32-bit hashes cancels when address == path-hash; CRC32C avalanche cuts ConcurrentDictionary bucket collisions and the lock contention that follows. Adds a Hash256.GetChainedHashCode(ulong) overload and a TinyTreePath.GetChainedHashCode(uint seed) helper so callers feed the full 64 bits of path data into the mix instead of the pre-folded 32-bit hash.
- Dictionary sizing: share concurrencyLevel + prime-bucket count between the dirty nodes cache and _persistedHashes via GetDictionarySizing, pre-sizing the previously unsized _persistedHashes dictionaries and avoiding resize storms at startup.
- GetOrAdd rewrite: replace AddOrUpdate with closures by an explicit TryGetValue/Merge/TryUpdate/TryAdd CAS loop. Splits out MergeRecords so the write path can short-circuit the CAS when the merged record equals the current one (the common revisit-with-no-change case AddOrUpdate always paid for).
- Persisted-hash recorder: extract TrieStore.RecordPersistedHash so the delete-old-nodes null-sentinel protocol is directly testable. After the first revisit, subsequent revisits short-circuit on the sticky null without re-running an update factory.
- Tests: add MergeRecords coverage for neither-persisted, advances-last-commit, persisted-current-dirty-candidate, and dirty-current-persisted-candidate branches; plus RecordPersistedHash coverage for both deleteOldNodes modes.

- [x] Optimization

#### Requires testing

- [x] Yes

- [x] Yes

### DIFF
--- a/src/Nethermind/Nethermind.Trie.Test/Pruning/TreeStoreTests.cs
+++ b/src/Nethermind/Nethermind.Trie.Test/Pruning/TreeStoreTests.cs
@@ -1745,70 +1745,25 @@ namespace Nethermind.Trie.Test.Pruning
     [Parallelizable(ParallelScope.All)]
     public class TreeStoreInternalBehaviorTests
     {
-        [Test]
-        public void Dirty_node_record_merge_keeps_current_when_neither_persisted()
+        [TestCase(false, false, 4, 2, false, 4, TestName = "Dirty_node_record_merge_keeps_current_when_neither_persisted")]
+        [TestCase(false, false, 2, 5, false, 5, TestName = "Dirty_node_record_merge_advances_last_commit_from_candidate")]
+        [TestCase(true, false, 1, 3, true, 3, TestName = "Dirty_node_record_merge_replaces_persisted_node_with_dirty_candidate")]
+        [TestCase(false, true, 4, 6, false, 6, TestName = "Dirty_node_record_merge_keeps_current_when_candidate_is_persisted")]
+        public void Dirty_node_record_merge_selects_expected_node_and_last_commit(bool currentPersisted, bool candidatePersisted, long currentLastCommit, long candidateLastCommit, bool expectCandidateNode, long expectedLastCommit)
         {
-            TrieNode currentNode = new(NodeType.Unknown, TestItem.KeccakA);
-            TrieNode candidateNode = new(NodeType.Unknown, TestItem.KeccakA);
-            currentNode.IsPersisted = false;
-            candidateNode.IsPersisted = false;
-            TrieStoreDirtyNodesCache.NodeRecord current = new(currentNode, 4);
-            TrieStoreDirtyNodesCache.NodeRecord candidate = new(candidateNode, 2);
-
-            TrieStoreDirtyNodesCache.MergeRecords(current, candidate).Should().Be(current);
-        }
-
-        [Test]
-        public void Dirty_node_record_merge_advances_last_commit_from_candidate()
-        {
-            TrieNode currentNode = new(NodeType.Unknown, TestItem.KeccakA);
-            TrieNode candidateNode = new(NodeType.Unknown, TestItem.KeccakA);
-            currentNode.IsPersisted = false;
-            candidateNode.IsPersisted = false;
+            TrieNode currentNode = CreateNode(currentPersisted);
+            TrieNode candidateNode = CreateNode(candidatePersisted);
 
             TrieStoreDirtyNodesCache.NodeRecord merged = TrieStoreDirtyNodesCache.MergeRecords(
-                new TrieStoreDirtyNodesCache.NodeRecord(currentNode, 2),
-                new TrieStoreDirtyNodesCache.NodeRecord(candidateNode, 5));
+                CreateRecord(currentNode, currentLastCommit),
+                CreateRecord(candidateNode, candidateLastCommit));
 
-            merged.Node.Should().BeSameAs(currentNode);
-            merged.LastCommit.Should().Be(5);
+            merged.Node.Should().BeSameAs(expectCandidateNode ? candidateNode : currentNode);
+            merged.LastCommit.Should().Be(expectedLastCommit);
         }
 
         [Test]
-        public void Dirty_node_record_merge_replaces_persisted_node_with_dirty_candidate()
-        {
-            TrieNode persistedNode = new(NodeType.Unknown, TestItem.KeccakA);
-            persistedNode.IsPersisted = true;
-            TrieNode candidateNode = new(NodeType.Unknown, TestItem.KeccakA);
-            candidateNode.IsPersisted = false;
-
-            TrieStoreDirtyNodesCache.NodeRecord merged = TrieStoreDirtyNodesCache.MergeRecords(
-                new TrieStoreDirtyNodesCache.NodeRecord(persistedNode, 1),
-                new TrieStoreDirtyNodesCache.NodeRecord(candidateNode, 3));
-
-            merged.Node.Should().BeSameAs(candidateNode);
-            merged.LastCommit.Should().Be(3);
-        }
-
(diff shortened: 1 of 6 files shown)


benaadams:
@claude re-review

src/Nethermind/Nethermind.Trie.Test/Pruning/TreeStoreTests.cs
```diff
@@ -1739,4 +1740,93 @@ public void Consecutive_block_commits_work_when_first_has_null_root()
...
+                persistedHashes[key].Should().BeNull();
+            }
+            else
+            {
+                persistedHashes[key].Should().Be(TestItem.KeccakB);
+            }
+
+            persistedHashes.Count.Should().Be(1);
+        }
+    }
```
LukaszRozmej, line 1791:
```suggestion
public class TreeStoreInternalBehaviorTests
{
    [TestCase(false, false, 4, 2, false, 4, TestName = "Dirty_node_record_merge_keeps_current_when_neither_persisted")]
    [TestCase(false, false, 2, 5, false, 5, TestName = "Dirty_node_record_merge_advances_last_commit_from_candidate")]
    [TestCase(true, false, 1, 3, true, 3, TestName = "Dirty_node_record_merge_replaces_persisted_node_with_dirty_candidate")]
    [TestCase(false, true, 4, 6, false, 6, TestName = "Dirty_node_record_merge_keeps_current_when_candidate_is_persisted")]
    public void Dirty_node_record_merge_selects_expected_node_and_last_commit(bool currentPersisted, bool candidatePersisted, long currentLastCommit, long candidateLastCommit, bool expectCandidateNode, long expectedLastCommit)
    {
        TrieNode currentNode = CreateNode(currentPersisted);
        TrieNode candidateNode = CreateNode(candidatePersisted);

        TrieStoreDirtyNodesCache.NodeRecord merged = TrieStoreDirtyNodesCache.MergeRecords(
            CreateRecord(currentNode, currentLastCommit),
            CreateRecord(candidateNode, candidateLastCommit));

        merged.Node.Should().BeSameAs(expectCandidateNode ? candidateNode : currentNode);
        merged.LastCommit.Should().Be(expectedLastCommit);
    }

    [Test]
    public void Persisted_hash_recorder_handles_revisited_keys([Values] bool deleteOldNodes)
    {
        ConcurrentDictionary<HashAndTinyPath, Hash256?> persistedHashes = new();
        HashAndTinyPath key = new(TestItem.KeccakA, new TinyTreePath(TreePath.Empty));

        TrieStore.RecordPersistedHash(persistedHashes, key, TestItem.KeccakB, deleteOldNodes);
        TrieStore.RecordPersistedHash(persistedHashes, key, TestItem.KeccakC, deleteOldNodes);

        if (deleteOldNodes)
        {
            TrieStore.RecordPersistedHash(persistedHashes, key, TestItem.KeccakD, deleteOldNodes);
            persistedHashes[key].Should().BeNull();
        }
        else
        {
            persistedHashes[key].Should().Be(TestItem.KeccakB);
        }

        persistedHashes.Count.Should().Be(1);
    }

    private static TrieNode CreateNode(bool isPersisted) =>
        new(NodeType.Unknown, TestItem.KeccakA) { IsPersisted = isPersisted };

    private static TrieStoreDirtyNodesCache.NodeRecord CreateRecord(TrieNode node, long lastCommit) => new(node, lastCommit);
}
```

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
```diff
@@ -849,25 +850,32 @@ private void PersistedNodeRecorder(TreePath treePath, Hash256 address, TrieNode
...
             {
-                _persistedHashes[shardIdx].AddOrUpdate(
-                    key,
-                    static (_, newHash) => newHash,
-                    // When not deleting old nodes, key tracking is used to prune in memory cache. It is
-                    // safe to accidentally remove key by taking non-canon block as canon as it will just load
-                    // from disk again.
-                    static (_, hash, _) => hash,
-                    tn.Keccak);
+                return;
```
LukaszRozmej, line 878: Weird loop, we could make it weirder:


```suggestion
while (persistedHashes.TryGetValue(key, out Hash256? existingHash) &&
       existingHash is not null &&
       !persistedHashes.TryUpdate(key, null, existingHash));
```

Not sure if good idea :)

src/Nethermind/Nethermind.Trie/Pruning/TrieStoreDirtyNodesCache.cs
```diff
@@ -37,6 +37,12 @@ internal class TrieStoreDirtyNodesCache
     public readonly long KeyMemoryUsage;
     private readonly bool _keepRoot;
 
+    internal static void GetDictionarySizing(out int concurrencyLevel, out int initialBuckets)
+    {
+        concurrencyLevel = Math.Min(Environment.ProcessorCount * 4, 32);
```
LukaszRozmej: `Environment.ProcessorCount` can return 0 in zk-evm. We have `RuntimeInformation.ProcessorCount` handling that.

But probably not an issue here

src/Nethermind/Nethermind.Trie/Pruning/TrieStoreDirtyNodesCache.cs
```diff
@@ -168,23 +173,45 @@ private NodeRecord GetOrAdd(in Key key, TrieStoreDirtyNodesCache cache) => _stor
...
+                continue;
+            }
 
-    private static NodeRecord RecordReplacementLogic(Key key, NodeRecord current, NodeRecord arg) => RecordReplacementLogic(null, current, arg);
+            if (dictionary.TryAdd(key, record))
+            {
+                return record;
+            }
+        }
+    }
```
LukaszRozmej, line 205: @claude what does that mean in detail, why `AddOrUpdate` is not good? I'm confused

src/Nethermind/Nethermind.Trie/Pruning/TrieStoreDirtyNodesCache.cs
benaadams: wtf?

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
benaadams: Too weird is like 

```csharp
if (bool)
{

}
else while (bool)
{

}
```

src/Nethermind/Nethermind.Trie/Pruning/TrieStoreDirtyNodesCache.cs
benaadams: Changed

LukaszRozmej (approved):

src/Nethermind/Nethermind.Trie/Pruning/TrieStore.cs
LukaszRozmej: I like this weird, but whatever

it reads logically better for me:

loop while we get something out, it isn't null and updates fail.

Maybe this one is better for understanding stop conditions:

```
Hash256? existingHash;

do
{
    if (!persistedHashes.TryGetValue(key, out existingHash) ||
        existingHash is null)
    {
        break;
    }
}
while (!persistedHashes.TryUpdate(key, null, existingHash));
```

asdacap (approved):

kamilchodola (approved):

PR #11306, opened 2026-04-22
fix: skip AuRa finalization startup walk on post-merge chains

LukaszRozmej:
## Summary
- On post-merge Gnosis, `AuRaBlockFinalizationManager` was still doing its startup catch-up walk, accumulating every un-finalized ancestor into a local `List<BlockHeader>`. Heap dumps from affected nodes showed the backing array at 16,777,216 entries (~10GB retained).
- Moved the walk's trigger from the constructor into `SetMainBlockBranchProcessor` with an idempotency guard, and added `InitializeBlockchainAuRaMerge` that skips wiring the branch processor once `IPoSSwitcher.HasEverReachedTerminalBlock()` is true.
- `AuRaMergeFinalizationManager.SetMainBlockBranchProcessor` also skips forwarding post-merge as a belt-and-suspenders guard.

## Test plan
- [x] `Nethermind.AuRa.Test.AuRaBlockFinalizationManagerTests` — 29/29 pass, including new `repeated_SetMainBlockBranchProcessor_is_idempotent`.
- [x] `Nethermind.Merge.AuRa.Test.AuRaMergeFinalizationManagerTests` — 5/5 pass, including new `SetMainBlockBranchProcessor_forwards_only_pre_merge` TestCase.
- [ ] Smoke test on a post-merge Gnosis node to confirm no longer stuck at startup and no walk-related allocations.

### DIFF
--- a/src/Nethermind/Nethermind.AuRa.Test/AuRaBlockFinalizationManagerTests.cs
+++ b/src/Nethermind/Nethermind.AuRa.Test/AuRaBlockFinalizationManagerTests.cs
@@ -48,6 +48,21 @@ namespace Nethermind.AuRa.Test
             finalizationManager.LastFinalizedBlockLevel.Should().Be(1);
         }
 
+        [Test]
+        public void repeated_SetMainBlockBranchProcessor_is_idempotent()
+        {
+            // The merge plugin can end up calling SetMainBlockBranchProcessor via the wrapper
+            // as well. Initialization must be safe to invoke more than once.
+            BlockTreeBuilder blockTreeBuilder = Build.A.BlockTree().OfChainLength(3, 1, 1);
+            FinalizeToLevel(1, blockTreeBuilder.ChainLevelInfoRepository);
+
+            AuRaBlockFinalizationManager finalizationManager = new(blockTreeBuilder.TestObject, blockTreeBuilder.ChainLevelInfoRepository, _validatorStore, _logManager);
+            finalizationManager.SetMainBlockBranchProcessor(_blockProcessor);
+            finalizationManager.SetMainBlockBranchProcessor(_blockProcessor);
+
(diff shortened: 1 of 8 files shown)


LukaszRozmej:
Thanks for the review. Addressed in a2f13971ed:

1. **Blank line nit** — fixed, now a single blank line between `SetMainBlockBranchProcessor` and `Initialize`.
2. **Silent drop on different processor** — added a `ReferenceEquals` check that throws `InvalidOperationException` when called with a different `IBranchProcessor` after initialization. Existing idempotency test uses the same instance both times so it still passes without change.
3. **`LastFinalizedBlockLevel = 0` window during DI resolution** — accepting the risk as flagged. The only way to close the window fully is to partially re-introduce a ctor-time scan of `ChainLevelInfoRepository`, which on post-merge Gnosis is ~25M levels of backward metadata reads — the very thing this PR is trying to avoid. `StartBlockProcessor` blocks on `InitializeBlockchain` completing, and `WireFinalizationBranchProcessor` runs inside that step, so the state is corrected before any block can be processed. `MultiValidator.OnBlockProcessingStart` also re-evaluates on the first non-consecutive block. Worth keeping the note here for future readers.
4. **Merge-wrapper guard observation** — confirmed, it's defense-in-depth. The actual gate is `InitializeBlockchainAuRaMerge` not making the call in the first place on post-merge chains; the wrapper's guard covers any future path that reaches it.

LukaszRozmej:
@claude re-review

svlachakis (approved):

damian-orzechowski (approved):

smartprogrammer93 (approved):

LukaszRozmej:
@claude re-review

svlachakis (approved):

PR #11309, opened 2026-04-22
Optimize caching & tests for precompiles

alexb5dh:
## Changes

- For cache:
	- Introduced `IPrecompile.GetEffectiveInput` implemented in `BN254Add`, `BN254Mul`, `ECRecover` - where tailing data doesn't affect output.
	- Optimized precompile caching to only use effective input as a key.

- For tests:
	- Fixed `IPrecompile.Instance` override in all implementors, got rid of `IPrecompileTests.Precompile`.
	- Most precompile tests now inherit `PrecompileTests` base class, common vector testing flow is moved there.

- [x] Optimization

#### Requires testing

- [x] Yes

- [x] Yes


alexb5dh:
@claude re-review

src/Nethermind/Nethermind.Evm.Precompiles/Blake2FPrecompile.cs
```diff
@@ -17,6 +17,7 @@ public partial class Blake2FPrecompile : IPrecompile<Blake2FPrecompile>
     private const int RequiredInputLength = 213;
 
     public static readonly Blake2FPrecompile Instance = new();
+    static Blake2FPrecompile IPrecompile<Blake2FPrecompile>.Instance => Instance;
```
LukaszRozmej: Let's collapse those to just (everywhere)

```suggestion
    public static Blake2FPrecompile Instance { get; } = new();
```

LukaszRozmej (approved):

src/Nethermind/Nethermind.Evm.Precompiles/Blake2FPrecompile.cs
alexb5dh: Any performance concerns on property access vs field?

benaadams: Not if its a class; can be concerns if large struct (but then can return `ref` or `readonly ref` also from property; oor field if on struct as won't let you access `this` ref internally)

benaadams (approved):

LukaszRozmej (approved):

PR #11314, opened 2026-04-23
Avoid pooled RLP aliasing in read-only trie clones

benaadams:
## Changes

- Read-only trie clones were intended to isolate concurrent readers, but CloneForReadOnly still reused the original node’s pooled FullRlp backing array. That left the read-only path exposed to buffer reuse while the clone was still decoding or traversing children.

- This change makes read-only clones take ownership of their RLP bytes by copying the serialized node data before resolving it. It also strengthens the trie regression test to assert that read-only clones do not share the original backing array and still decode correctly after the original buffer is mutated.

#### Requires testing

- [x] Yes

- [x] Yes


LukaszRozmej (approved):
Or do reference counting? :)

benaadams:
@claude re-review

kamilchodola (approved):

PR #11316, opened 2026-04-23
Add GetEffectiveInput override to ModExp precompiles

LukaszRozmej:
## Changes

Follow-up to #11309, which introduced `IPrecompile.GetEffectiveInput` to normalize the precompile cache key by stripping bytes that don't affect `Run`'s output.

### Omission: ModExp was missed
`ModExpPrecompile.Run` reads exactly `96 + baseLength + expLength + modulusLength` bytes via `SliceWithZeroPaddingEmptyOnError` — any trailing bytes are silently ignored. Without the override, callers that re-enter ModExp with identical header + payload but different trailing padding produce different cache keys for identical outputs, wasting cache capacity.

Added `GetEffectiveInput` on:
- `ModExpPrecompile` — trims to `96 + base + exp + mod`; returns header-only when any length saturates to `uint.MaxValue` (overflow sentinel / EIP-7823 oversized) or when `base == 0 && mod == 0` (Run short-circuits to empty regardless of exp).
- `ModExpPrecompilePreEip2565` — same behavior, adapted for the BigInteger-based header parsing already used by that precompile.

### Related optimization/hardening of `ModExpPrecompilePreEip2565.Run`
While writing tests for the overflow path on the obsolete pre-EIP-2565 implementation, I noticed `Run` could crash on pathological inputs that `GetEffectiveInput` needs to handle:
- `(int)BigInteger` direct casts on `baseLength` / `modulusLength` threw `OverflowException` for oversized headers (only `expLength` was previously saturated).
- `96 + baseLength + expLength` was computed in `int`, wrapping to a negative `startIndex` and triggering `ArgumentOutOfRangeException` inside `SliceWithZeroPaddingEmptyOnError`.

Replaced the ad-hoc cast/arithmetic with two private helpers:
- `SafeCast(BigInteger)` — saturating `BigInteger → int`.
- `SafeSlice(span, long start, int len)` — returns empty if the start index doesn't fit in `int`, otherwise delegates to the existing slice helper.

In practice these paths aren't reachable in production because `DataGasCost` returns `long.MaxValue` for such inputs and the EVM call runs out of gas before `Run` is invoked. The hardening is defensive and also lets `GetEffectiveInput`'s overflow branch be exercised end-to-end in tests.

### Audit
All other precompiles were audited against the same criterion:
- `ECRecover`, `BN254Add`, `BN254Mul` — already covered by #11309.
- `Sha256`, `Ripemd160`, `Identity` — every byte is consumed; default identity is correct.
- `BN254Pairing`, `Blake2F`, `KzgPointEvaluation`, all `Bls12381*`, `SecP256r1` — strict length validation; trailing bytes cause failure, so default identity is correct.
- Taiko `L1Sload` (strict length), `L1StaticCall` (`SupportsCaching = false`, calldata tail is semantically significant) — not applicable.
- No `IPrecompile` implementors in Optimism or Flashbots.

## Types of changes

- [x] Optimization
- [x] Bugfix (ModExpPrecompilePreEip2565 overflow hardening — defensive, not reachable in production)

## Testing

- [x] Yes, tests added.

`GetEffectiveInput_SameOutput` and `GetEffectiveInput_SameOutput_PreEip2565` share a helper (`AssertEffectiveInputPreservesOutput`) that verifies both `effInput.Length < fullInput.Length` and `Run(effInput) == Run(fullInput)` across normal, zero-length-header, and overflow cases. All 116 `ModExpPrecompileTests` pass; precompile cache tests (`PrecompileCachedCodeInfoRepositoryTests`) also green.


benaadams (approved):

src/Nethermind/Nethermind.Evm.Test/ModExpPrecompileTests.cs
```diff
@@ -1,13 +1,73 @@
...
 {
+    [TestCase(
+        "00000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000002003fffffffffffffffffffffffffffffffffffffffffffffffffffffffefffffc2efffffffffffffffffffffffffffffffffffffffffffffffffffffffefffffc2f",
+        "11"
+    )]
+    [TestCase(
+        "000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000040e09ad9675465c53a109fac66a445c91b292d2bb2c5268addb30cd82f80fcb0033ff97c80a5fc6f39193ae969c6ede6710a6b7ac27078a06d90ef1c72e5c85fb502fc9e1f6beb81516545975218075ec2af118cd8798df6e08a147c60fd6095ac2bb02c2908cf4dd7c81f11c289e4bce98f3553768f392a80ce22bf5c4f4a248c6b",
+        "deadbeef"
+    )]
+    // baseLength = 0, modulusLength = 0 -> Run returns empty regardless of trailing bytes.
```
alexb5dh: Comments may be better as test names

alexb5dh (approved):
I think your AI is too aggressive with commenting 😆

alexb5dh (approved):

PR #11317, opened 2026-04-23
Optimize flat-state block prewarming and overlap post-processing

kamilchodola:
## Changes

- overlap receipt-root work with the post-transaction tail of block processing instead of serializing that work
- reduce redundant prewarm work by deduplicating warmed addresses and reusing read-only warmup scopes on unique-sender blocks
- narrow storage warmup to dirty slots and add state-write trie hints to improve flat-state access locality
- keep the branch simple by removing unsupported benchmark-only tuning and restoring the normal `TrieWarmer` default heuristic
- tighten focused tests so both unique-sender and grouped-sender prewarmer paths are covered

- [x] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

Focused tests run locally:

- `dotnet test --project src/Nethermind/Nethermind.Consensus.Test/Nethermind.Consensus.Test.csproj -c Release --no-restore -- --filter FullyQualifiedName~BlockCachePreWarmerTests`
- `dotnet test --project src/Nethermind/Nethermind.Core.Test/Nethermind.Core.Test.csproj -c Release --no-restore -- --filter FullyQualifiedName~ParallelUnbalancedWorkTests`
- `dotnet test --project src/Nethermind/Nethermind.State.Flat.Test/Nethermind.State.Flat.Test.csproj -c Release --no-restore -- --filter FullyQualifiedName~TrieWarmerTests`

EXPB results on the branch:

- 5x code-only run, no extra flags: Actions run `24829908301`
  - AVG: `36.34`, `35.20`, `36.13`, `35.54`, `36.09`
  - mean: `35.86 ms`
  - clean run: no exceptions, no invalid blocks
- 5x run with explicit `--FlatDb.TrieWarmerWorkerCount=13`: Actions run `24828284179`
  - candidate mean: `35.76 ms`
  - master mean: `38.41 ms`
  - about `6.9%` faster than master on that runner
- best earlier 3x result on the same optimization stack with worker-count override: Actions run `24824505253`
  - mean: `35.04 ms`

Fresh dotTrace compare on the optimized image without extra flags: Actions run `24829908352`

- candidate AVG: `36.86 ms`
- master AVG: `39.88 ms`
- key hot-path own-time changes from dotTrace XML reports:
  - `DbOnTheRocks.Get`: `172626 -> 125769` (`-27.1%`)
  - `DbOnTheRocks.GetCStyleWithColumnFamily`: `73979 -> 64106` (`-13.3%`)
  - `TrieWarmer` worker path: `61338 -> 56618` (`-7.7%`)
- most remaining `DbOnTheRocks.Get` own time is still in trie-warmer storage reads, so the branch improves the current bottleneck but does not eliminate it

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks

This PR intentionally excludes benchmark-only tuning that was not justified by the no-flag runs. During branch review, those changes were removed and the branch was rechecked for focused coverage and simplification before pushing.


kamilchodola:
@claude re-review

src/Nethermind/Nethermind.State.Flat/ScopeProvider/FlatStorageTree.cs
```diff
@@ -79,16 +79,9 @@ public byte[] Get(in UInt256 index)
             }
         }
 
-        HintGet(index, value);
-
         return value!;
     }
 
-    // Note: VERY hot code.
```
asdacap, line 87: Keep

src/Nethermind/Nethermind.State.Flat/ScopeProvider/FlatWorldStateScope.cs
```diff
@@ -90,7 +90,7 @@ public void Dispose()
     {
         Account? account = _snapshotBundle.GetAccount(address);
 
-        HintGet(address, account);
+        _snapshotBundle.SetAccount(address, account);
```
asdacap: Set in hintget if neede

asdacap:
Please separate the hintget change and test indipendantly.

src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
```diff
@@ -234,6 +241,46 @@ private void WarmupTransactions(BlockState blockState, ParallelOptions parallelO
...
+        for (int i = 0; i < transactions.Length; i++)
+        {
+            if (transactions[i].SenderAddress is { } sender && !senders.Add(sender))
+            {
+                return true;
+            }
+        }
+
+        return false;
+    }
```
LukaszRozmej, line 333: Not sure if optimal also from allocating set perspective

src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
```diff
@@ -189,18 +197,17 @@ private void WarmupTransactions(BlockState blockState, ParallelOptions parallelO
                     0,
                     groupArray.Count,
                     parallelOptions,
-                    (blockState, groupArray, parallelOptions.CancellationToken),
+                    (blockState, groupArray, parallelOptions.CancellationToken, context),
```
LukaszRozmej, line 229: maybe also make it into a record?

src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs
```diff
@@ -400,18 +484,43 @@ private void WarmupAddresses(ParallelOptions parallelOptions, Block block)
...
+
+                Address? to = tx.To;
+                if (to is not null && seen.Add(to))
+                {
+                    addresses.Add(to);
+                }
+            }
+
+            return addresses;
+        }
```
LukaszRozmej, line 589: Similar here?

src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.cs
```diff
@@ -127,14 +129,15 @@ protected virtual TxReceipt[] ProcessBlock(
         stateProvider.Commit(spec, commitRoots: false);
 
         CalculateBlooms(receipts);
+        // IMPORTANT: receipts must remain read-only until receiptsRootTask completes.
+        // GetReceiptsRoot runs concurrently with ProcessExecutionRequests and EndBlockTrace.
+        Task<Hash256> receiptsRootTask = Task.Run(() => ReceiptsRootCalculator.Instance.GetReceiptsRoot(receipts, spec, block.ReceiptsRoot));
```
LukaszRozmej: gate it by transaction count?

LukaszRozmej (approved):

kamilchodola:
@claude 
<html>
<body>

Low | StateProvider.SetState → HintGet timing was unreliable | ✅ _tree?.HintGet(address, account) call removed entirely — cleaner than patching timing
-- | -- | --



</body>
</html>

Are you sure this was a good review comment? I think it made performance worse at least

kamilchodola:
Addressing the recent review comments:

- I restored the `FlatStorageTree.HintGet` rationale comment on the PR branch in `a2f112d0a8`.
- I also did the requested isolated `HintGet` check on a separate branch: `perf/expb-account-hintget-isolated`.

For the isolated check, I changed only this one line in `FlatWorldStateScope.Get`:

```csharp
_snapshotBundle.SetAccount(address, account);
```

back to:

```csharp
HintGet(address, account);
```

So this is the smallest possible change at that site: direct `SetAccount` vs `HintGet` (which does the same `SetAccount` plus conditional prewarm enqueue).

I benchmarked it against the current PR branch image, not `master`, to isolate only that difference:

- baseline image: `nethermindeth/nethermind:perf-expb-workercap-address-dedup`
- isolated image: `nethermindeth/nethermind:perf-expb-account-hintget-isolated`
- EXPB run: `24833779521`
- config: `flat`, `realblocks`, `1000`, `1x`

Result:

- baseline: `37.14 ms`
- isolated `HintGet`: `35.62 ms`
- delta: `-1.52 ms` (`-4.09%`)
- no exceptions
- no invalid blocks

So this specific `HintGet` vs plain `SetAccount` change does look materially positive. It is only a single run, but the effect size is large enough that I would not drop it without at least a `3x`/`5x` follow-up.

kamilchodola:
@claude final review please - correctness, maintability, performance

LukaszRozmej:
## One-pass warmup plan replacing `HasDuplicateSenders` + `GroupTransactionsBySender` + `CollectWarmupAddresses`

All three methods walk the tx array and dedup senders in slightly different ways. They can be collapsed into a single pass that produces everything the warmup paths need, driven by one pooled probe table. Every output buffer is `ArrayPoolList<…>` — no `HashSet`, no `Dictionary`, no per-sender lists.

### Data shape

```csharp
private readonly record struct SenderGroup(int FirstTxIdx, int LastTxIdx, int AddressSlot);

private readonly struct WarmupPlan(
    ArrayPoolList<AddressAsKey> addresses,      // deduped senders + tos
    ArrayPoolList<SenderGroup> senderGroups,    // one entry per unique sender
    ArrayPoolList<int> txNext,                  // linked list, length == txs.Length
    bool hasDuplicateSenders) : IDisposable { … }
```

`SenderGroup` gives the head/tail of a linked list over `txNext`; iterating a sender's txs is `for (int i = group.FirstTxIdx; i >= 0; i = txNext[i])`. No per-sender `ArrayPoolList` allocation.

### The walk

```csharp
int cap = (int)BitOperations.RoundUpToPowerOf2((uint)(txs.Length * 2 + 8));
using ArrayPoolList<int> probe = new(cap, cap);                 // 0 = empty slot
ArrayPoolList<AddressAsKey> addresses = new(txs.Length * 2);
ArrayPoolList<int> senderOfAddr = new(txs.Length * 2);          // parallel to addresses; 0 = not a sender, else groupIdx+1
ArrayPoolList<SenderGroup> groups = new(txs.Length);
ArrayPoolList<int> txNext = new(txs.Length, txs.Length);        // pre-filled 0; overwritten with -1 per tx
Span<int> probeSlots = probe.AsSpan();
int mask = cap - 1;
bool hasDup = false;

for (int i = 0; i < txs.Length; i++)
{
    Transaction tx = txs[i];
    txNext[i] = -1;

    if (tx.SenderAddress is { } sender)
    {
        int addrSlot = ProbeOrInsert(sender, probeSlots, mask, addresses, senderOfAddr);
        int encoded = senderOfAddr[addrSlot];
        if (encoded == 0)
        {
            senderOfAddr[addrSlot] = groups.Count + 1;
            groups.Add(new SenderGroup(FirstTxIdx: i, LastTxIdx: i, AddressSlot: addrSlot));
        }
        else
        {
            hasDup = true;
            int g = encoded - 1;
            SenderGroup sg = groups[g];
            txNext[sg.LastTxIdx] = i;
            groups[g] = sg with { LastTxIdx = i };
        }
    }

    if (tx.To is { } to)
        ProbeOrInsert(to, probeSlots, mask, addresses, senderOfAddr);
}
```

### `ProbeOrInsert`

```csharp
private static int ProbeOrInsert(
    Address addr,
    Span<int> probeSlots,
    int mask,
    ArrayPoolList<AddressAsKey> addresses,
    ArrayPoolList<int> senderOfAddr)
{
    int hash = addr.GetHashCode();                              // FastHash over 20 bytes
    for (int i = hash & mask; ; i = (i + 1) & mask)
    {
        int slot = probeSlots[i];
        if (slot == 0)                                           // empty — insert
        {
            int idx = addresses.Count;
            addresses.Add(addr);
            senderOfAddr.Add(0);                                 // parallel; 0 = not-a-sender-yet
            probeSlots[i] = idx + 1;                             // store 1-based
            return idx;
        }

        AddressAsKey existing = addresses[slot - 1];
        if (ReferenceEquals((Address)existing, addr) || existing.Equals(addr))
            return slot - 1;                                     // hit
        // else: collision, linear-probe to the next slot
    }
}
```

Open-addressing with linear probing; hash function is `Address.GetHashCode()` (already a `FastHash` over the 20 bytes). Probe table stores 1-based indices into `addresses` so `0` remains the empty sentinel and the `ArrayPoolList(capacity, startingCount)` ctor zero-clears the buffer on rent for free. `ReferenceEquals` fast-path avoids the 20-byte compare when the same `Address` instance recurs (common for a signer's multiple txs post-ecrecover caching).

`addresses` and `senderOfAddr` grow in lockstep — `senderOfAddr[addrSlot] == 0` means "address exists but hasn't been seen as a sender yet" (important because a `To` can later appear as a sender, or vice versa). Both are pre-sized to `txs.Length * 2`, so no resize during the walk.

### How the warmup paths consume this

- **Address warm-up** iterates `addresses.AsSpan()` directly — already deduped.
- **Independent-sender fast path** fires when `!hasDuplicateSenders`; iterates `0..txs.Length` and warms each tx independently, same as today.
- **Grouped path** iterates `senderGroups.AsSpan()`; for each group, walks the linked list via `txNext`. Parallel work unit is a `SenderGroup`, not an `ArrayPoolList`. Same concurrency properties, no per-group list allocation.

### Why the `int` probe table beats `HashSet` even heap-to-heap

A `HashSet<AddressAsKey>` entry is ~24 B (hash + next + key-ref + padding) and scatters the `Address` references. A contiguous `int` probe table is 4 B per slot and only dereferences on a real hit comparison — increasingly cache-friendlier as tx counts scale up. At 10k+ txs drop the load factor to `2×` (≤ 0.5); probe chains stay short because of `FastHash`'s distribution, and memory halves (1.6 MB → 800 KB at 100k).

### Also worth cleaning up

With the fused pass in place, `AddWarmupAddress` goes away entirely, and the `seen is not null` branch with it. The remaining `addresses.AsSpan()` loop has no per-address bounds check.

### Allocation diff vs current PR (blocks with > 8 txs)

| Before | After |
|---|---|
| `HashSet<AddressAsKey>` × 2 | — |
| `Dictionary<AddressAsKey, ArrayPoolList<…>>` | — |
| Per-sender `ArrayPoolList<(int,Transaction)>` × sender-count | — |
| `ArrayPoolList<AddressAsKey>` | `ArrayPoolList<AddressAsKey>` (addresses) |
| | `ArrayPoolList<int>` × 3 (probe, senderOfAddr, txNext) |
| | `ArrayPoolList<SenderGroup>` |

All remaining lists use pooled arrays; wrapper allocation is fixed cost per block regardless of size, vs the current O(unique-senders) list allocations in the grouped path.

src/Nethermind/Nethermind.State.Flat/ScopeProvider/FlatStorageTree.cs
```diff
@@ -79,16 +79,13 @@ public byte[] Get(in UInt256 index)
...
         return value!;
     }
 
     // Note: VERY hot code.
-    // 90% of the read goes through prewarmer, not actually go through this class, meaning this method is called
-    // a lot. Setting the set slot have a measurable net negative impact on performance.
-    // Trying to set this value async through trie warmer proved to be hard to pull of and result in random invalid
-    // block.
+    // Most storage reads come through the prewarmer rather than this method directly, so any extra work here amplifies quickly.
+    // Trying to feed every read back into the snapshot bundle from this path was a measurable regression, while async variants
```
asdacap, line 87: The comment changed meaning.

asdacap (changes requested):
Separate out hintget related changes.

kamilchodola:
Closing in favour of: #11409

PR #11318, opened 2026-04-23
Optimise parallel execution: reduce worldstate, txprocessor allocations

Marchhill:
## Changes

- Reduce allocations by reusing `TracedAccessWorldState`, `BlockAccessListBasedWorldState`, `TransactionProcessor`, `BlockAccessList` instance

- [x] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

Covered by engine module tests.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus/Processing/BlockAccessListManager.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/BlockAccessListManager.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Threading.Tasks;
 using System.Linq;
 using Nethermind.Blockchain;
@@ -477,15 +478,14 @@ public class BlockAccessListManager(
 
     private void CheckInitialized()
     {
-        if (_txProcessorWithWorldStateManager is null)
-            throw new InvalidOperationException($"{nameof(_txProcessorWithWorldStateManager)} was not initialized.");
-
-        if (_gasRemaining is null)
-            throw new InvalidOperationException($"{nameof(_gasRemaining)} was not initialized.");
-
-        if (_blockExecutionContext is null)
-            throw new InvalidOperationException($"{nameof(_blockExecutionContext)} was not initialized.");
+        if (_txProcessorWithWorldStateManager is null) ThrowNotInitialized(nameof(_txProcessorWithWorldStateManager));
+        if (_gasRemaining is null) ThrowNotInitialized(nameof(_gasRemaining));
+        if (_blockExecutionContext is null) ThrowNotInitialized(nameof(_blockExecutionContext));
     }
+
+    [DoesNotReturn]
+    private static void ThrowNotInitialized(string fieldName)
+        => throw new InvalidOperationException($"{fieldName} was not initialized.");
 
     private interface ITxProcessorWithWorldStateManager
     {
@@ -554,21 +554,29 @@ public class BlockAccessListManager(
         {
             _currentBlock = block;
             _currentCtx = blockExecutionContext;
-            int previousSize = _lastBalIndex + 1;
-            _lastBalIndex = block.Transactions.Length + 1;
 
-            ReclaimAndResize(_lastBalIndex + 1, previousSize);
+            int previousSize = _lastBalIndex + 1;
+            int newLastBalIndex = block.Transactions.Length + 1;
+            ReclaimAndResize(newLastBalIndex + 1, previousSize);
+            _lastBalIndex = newLastBalIndex;
         }
 
+        // Thread-safety note for _inUse / _perTxBal:
+        //   Each balIndex slot has at most one writer at a time. Pre/post (idx 0 and
+        //   _lastBalIndex) are written only by the main thread; tx slots (1..len) are
+        //   each owned by a single parallel-loop iteration, so no two workers ever
+        //   touch the same slot. Cross-thread reads (validator → worker's slot) happen
+        //   strictly after the worker's gasResults[i-1].SetResult, whose pairing with
+        //   GetResult() establishes the publication barrier. Plain reads/writes are
+        //   therefore sufficient — Volatile/Interlocked would be redundant fencing.
         public TxProcessorWithWorldState Get(int? balIndex = null)
         {
-            if (_currentBlock is null)
-                throw new InvalidOperationException($"{nameof(_currentBlock)} was not initialized.");
+            if (_currentBlock is null) ThrowNotInitialized(nameof(_currentBlock));
 
             balIndex = ClampBalIndex(balIndex ?? 0);
 
             // Re-entrant Get for the same balIndex returns the already-acquired processor
-            // (lets pre/post callers share state across calls).
+            // (lets pre/post callers share state across calls — main thread only).
             TxProcessorWithWorldState? existing = _inUse[balIndex.Value];
             if (existing is not null) return existing;
 
@@ -649,14 +657,18 @@ public class BlockAccessListManager(
                 if (_inUse[i] is not null) Return(i);
 
             for (int i = 0; i < previousSize; i++)
-                if (_perTxBal[i] is { } bal) StaticPool<BlockAccessList>.Return(bal);
+            {
+                if (_perTxBal[i] is { } bal)
+                {
+                    StaticPool<BlockAccessList>.Return(bal);
+                    _perTxBal[i] = null;
+                }
+            }
 
             if (_inUse.Length < size)
                 Array.Resize(ref _inUse, Math.Max(2 * _inUse.Length, size));
             if (_perTxBal.Length < size)
                 Array.Resize(ref _perTxBal, Math.Max(2 * _perTxBal.Length, size));
-            Array.Clear(_inUse);
-            Array.Clear(_perTxBal);
         }
 
     }
--- a/src/Nethermind/Nethermind.Consensus/Processing/BlockAccessListManager.cs
+++ b/src/Nethermind/Nethermind.Consensus/Processing/BlockAccessListManager.cs
@@ -6,6 +6,9 @@ using System.Collections.Generic;
 using System.Threading.Tasks;
 using System.Linq;
 using Nethermind.Blockchain;
+using System.Collections.Concurrent;
+using Nethermind.Core.Caching;
+using Nethermind.Core.Cpu;
 using Nethermind.Blockchain.BeaconBlockRoot;
 using Nethermind.Blockchain.Blocks;
 using Nethermind.Blockchain.Tracing;
@@ -144,12 +147,27 @@ public class BlockAccessListManager(
         }
     }
 
+    public void ReturnTxProcessor(int balIndex)
+    {
+        if (Enabled && ParallelExecutionEnabled)
+        {
+            // Eagerly detach the worker's generated BAL into a per-tx slot and recycle the
+            // pool slot. Workers therefore never block on the validator — but the validator
+            // still merges per-tx slots into the target in order, preserving incremental
+            // validation semantics.
+            _parallelTxProcessorWithWorldStateManager.Return(balIndex);
+        }
+    }
+
     public void IncrementalValidation(Block block, TaskCompletionSource<(long BlockGasUsed, long BlockStateGasUsed, InvalidBlockException? Exception)>[] gasResults, BlockReceiptsTracer[] receiptsTracers, BlockValidationTransactionsExecutor.ITransactionProcessedEventHandler? transactionProcessedEventHandler, CancellationToken token)
     {
         CheckInitialized();
 
         int len = block.Transactions.Length;
-        _txProcessorWithWorldStateManager.GetPreExecution().WorldState.MergeGeneratingBal(GeneratedBlockAccessList);
+
+        // Pre is held by main-thread system contract handlers — never went through Return,
+        // so MergeAndReturnBal will detach it before merging.
+        _parallelTxProcessorWithWorldStateManager.MergeAndReturnBal(0, GeneratedBlockAccessList);
         ValidateBlockAccessList(block, 0);
 
         long totalRegularGas = 0;
@@ -176,8 +194,12 @@ public class BlockAccessListManager(
 
                 transactionProcessedEventHandler?.OnTransactionProcessed(new TxProcessedEventArgs(j, block.Transactions[j], block.Header, receiptsTracers[j].TxReceipts[0]));
 
+                // Worker for tx (j+1) has stashed its BAL into _perTxBal[j+1] via Return as
+                // soon as the tx finished — no contention with the validator. Merge it into
+                // the target now, in order, so incremental validation sees only data from
+                // txs 0..(j+1).
                 bool validateStorageReads = j == chunkEnd - 1;
-                _txProcessorWithWorldStateManager.Get(j + 1).WorldState.MergeGeneratingBal(GeneratedBlockAccessList);
+                _parallelTxProcessorWithWorldStateManager.MergeAndReturnBal(j + 1, GeneratedBlockAccessList);
                 ValidateBlockAccessList(block, (ushort)(j + 1), validateStorageReads);
             }
         }
@@ -271,7 +293,7 @@ public class BlockAccessListManager(
         {
             CheckInitialized();
 
-            _txProcessorWithWorldStateManager.GetPostExecution().WorldState.MergeGeneratingBal(GeneratedBlockAccessList);
+            _txProcessorWithWorldStateManager.MergeAndReturnBal(int.MaxValue, GeneratedBlockAccessList);
             block.GeneratedBlockAccessList = GeneratedBlockAccessList;
             block.EncodedBlockAccessList = Rlp.Encode(GeneratedBlockAccessList).Bytes;
             block.Header.BlockAccessListHash = new(ValueKeccak.Compute(block.EncodedBlockAccessList).Bytes);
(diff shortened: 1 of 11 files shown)


Marchhill:
@claude review

Marchhill:
@claude review

Marchhill:
> **Claude finished @Marchhill's task in 4m 3s** —— [View job](https://github.com/NethermindEth/nethermind/actions/runs/25041374441)
> 
> ### PR Review: `reuse worldstate` 
> * [x]  Read repository rules and guidelines
> * [x]  Examine current state of changed files
> * [x]  Review `BlockAccessListManager.cs` changes
> * [x]  Review `BlockProcessor.ParallelBlockValidationTransactionsExecutor.cs` changes
> * [x]  Review `NullBlockAccessListManager.cs`, `IBlockAccessListManager.cs` changes
> * [x]  Review `BlockAccessListBasedWorldState.cs` and `TracedAccessWorldState.cs` changes
> * [x]  Review `SimulateBlockValidationTransactionsExecutor.cs` changes
> * [x]  Assess correctness, security, performance, and adherence to repo rules
> * [x]  Post review findings
> 
> ### Summary
> The pool architecture is substantially cleaner than the previous review. All three issues flagged in the last pass have been addressed: the merge conflict in `NullBlockAccessListManager.cs` is resolved, the `.vscode` developer files are gone, and the per-tx `BlockAccessList` allocation is eliminated via the `_balPool` ping-pong. Two issues remain: a high-severity test gap and two medium findings.
> 
> ### Critical (0)
> None.
> 
> ### High (1)
> #### H1 — No tests for the pool lifecycle (carried over)
> The `Return` → `MergeAndReturnBal` → `SetResult` ordering is still the correctness-critical invariant of this PR, and it remains untested. The new `BlockAccessListBasedWorldStateTests.cs` covers `BlockAccessListBasedWorldState` reads, but not the pool machinery: acquire/return lifecycle, `_perTxBal[i]` populated before `gasResults[i-1].SetResult`, or pool-exhaustion blocking. At minimum a unit test for `Return` + `MergeAndReturnBal` sequencing should accompany this change.
> 
> [Fix this →](https://claude.ai/code?q=Add+unit+tests+for+ParallelTxProcessorWithWorldStateManager+in+BlockAccessListManager.cs+(branch%3A+feature%2Fparallel-txs-optimise-worldstate).+Tests+should+cover%3A+1)+Return+writes+_perTxBal%5Bidx%5D+before+releasing+the+semaphore%2C+2)+MergeAndReturnBal+correctly+merges+and+resets+the+BAL+pool+entry%2C+3)+pool+exhaustion+blocks+until+Return+is+called.&repo=NethermindEth%2Fnethermind)
> 
> ### Medium (2)
> #### M1 — LINQ `Enumerable.SequenceEqual` in the block-validation hot path
> `ValidateBlockAccessList` is called once per transaction per block in the parallel path. Line 366 uses:
> 
> ```cs
> !Enumerable.SequenceEqual(generatedHead.Value.SlotChanges, suggestedHead.Value.SlotChanges)
> ```
> 
> Per the repo rules: _"No LINQ when a simple `for`/`foreach` works."_ `SequenceEqual` allocates an enumerator and adds heap pressure in what is otherwise an allocation-free comparison loop. Replace with a manual element-count + element-wise comparison:
> 
> ```cs
> var gs = generatedHead.Value.SlotChanges;
> var ss = suggestedHead.Value.SlotChanges;
> bool slotMismatch = gs.Count != ss.Count;
> if (!slotMismatch)
> {
>     for (int k = 0; k < gs.Count; k++)
>         if (!gs[k].Equals(ss[k])) { slotMismatch = true; break; }
> }
> if (slotMismatch) throw ...;
> ```
> 
> [Fix this →](https://claude.ai/code?q=In+BlockAccessListManager.cs+(branch%3A+feature%2Fparallel-txs-optimise-worldstate)%2C+replace+the+Enumerable.SequenceEqual+call+on+SlotChanges+in+ValidateBlockAccessList+(around+line+366)+with+a+manual+count+%2B+element-wise+loop+to+eliminate+the+LINQ+enumerator+allocation+on+the+hot+path.+Also+remove+the+%60using+System.Linq%60+import+if+it+is+no+longer+needed.&repo=NethermindEth%2Fnethermind)
> 
> #### M2 — `SimulateBlockValidationTransactionsExecutor.SetBlockAccessListManager` throws `NotImplementedException`
> This PR adds:
> 
> ```cs
> // SimulateBlockValidationTransactionsExecutor.cs:60
> public void SetBlockAccessListManager(in IBlockAccessListManager balManager) => throw new System.NotImplementedException();
> ```
> 
> If `SetBlockAccessListManager` is part of `IBlockProcessor.IBlockTransactionsExecutor` and the block-processor pipeline calls it on this executor during `eth_simulateV1` or related flows, this will throw a `NotImplementedException` at runtime. Either implement it properly (forwarding to `baseTransactionExecutor`, or storing and ignoring if simulation doesn't use BAL features), or add a comment explaining why it cannot be reached and change it to a `throw new NotSupportedException(...)` with a descriptive message.
> 
> [Fix this →](https://claude.ai/code?q=In+SimulateBlockValidationTransactionsExecutor.cs+(branch%3A+feature%2Fparallel-txs-optimise-worldstate)%2C+the+SetBlockAccessListManager+method+throws+NotImplementedException.+Either+implement+it+by+forwarding+to+baseTransactionExecutor%2C+or+if+simulation+genuinely+cannot+use+BAL+features%2C+use+NotSupportedException+with+a+clear+message+and+add+a+comment+explaining+why+it+is+unreachable.&repo=NethermindEth%2Fnethermind)
> 
> ### Low (2)
> #### L1 — `ArrayPoolList` still allocated per-block
> ```cs
> // BlockAccessListManager.cs ~line 555
> _inUse?.Dispose();
> _perTxBal?.Dispose();
> _inUse = new ArrayPoolList<TxProcessorWithWorldState?>(size, size);
> _perTxBal = new ArrayPoolList<BlockAccessList?>(size, size);
> ```
> 
> `ArrayPoolList` is better than plain arrays (rents the backing store from the pool), but the list objects themselves are still freshly heap-allocated each block. Minor given the block rate, but worth noting in context of the broader zero-allocation goal.
> 
> #### L2 — PR description unfilled
> Template placeholder text remains, no type-of-change checkboxes are ticked. The checkboxes drive automatic PR labeling.
> 
> ### Resolved since previous review ✓
> Issue	Status
> C1: Merge conflict in `NullBlockAccessListManager.cs`	Fixed — both `ReturnTxProcessor` and `IncrementalValidation` now present
> M1: `.vscode` files with local paths committed	Fixed — directory removed
> M2: `TakeGeneratingBal()` allocated a new `BlockAccessList` per-tx	Fixed — `_balPool` `ConcurrentBag` ping-pong avoids per-tx allocation
> ### Correctness notes (informational)
> * **Pool/semaphore invariant**: `Return` writes `_perTxBal[i]` → `_free.Push` → `_available.Release` → caller calls `SetResult`. `GetAwaiter().GetResult()` on the validator only unblocks after `SetResult`. `TaskCompletionSource.SetResult`/`GetResult` provides the required memory barrier, so `_perTxBal[i]` is always visible to `MergeAndReturnBal`. ✓
> * **`TryPop` invariant**: replaced with an explicit `throw new InvalidOperationException(...)` — clear and safe. ✓
> * **`SemaphoreSlim` disposal**: `ParallelTxProcessorWithWorldStateManager.Dispose()` disposes `_available`. `BlockAccessListManager.Dispose()` disposes both managers. ✓
> * **Transient storage reset**: `BlockAccessListBasedWorldState.Setup(block)` resets `_transientStorageProvider` each time a pooled processor is acquired. ✓
> * **`ApplyStateChanges` concurrency**: runs at parallel-loop index 0 while TX workers (indices 1..len) read from BAL-backed world states, not directly from `stateProvider`. No data race between `ApplyStateChanges` and TX workers. ✓

@claude this functionality is tested by the existing `EngineModuleTestsV6`. The LINQ issue you mentioned is valid but outside of the scope of this PR

src/Nethermind/Nethermind.Consensus/Processing/BlockAccessListManager.cs
```diff
@@ -465,44 +487,171 @@ private void CheckInitialized()
...
-            for (int i = 0; i < _len; i++)
+            _currentBlock = block;
+            _currentCtx = blockExecutionContext;
+            _lastBalIndex = block.Transactions.Length + 1;
+
+            int size = _lastBalIndex + 1;
+            _inUse?.Dispose();
+            _perTxBal?.Dispose();
+            _inUse = new ArrayPoolList<TxProcessorWithWorldState?>(size, size);
+            _perTxBal = new ArrayPoolList<BlockAccessList?>(size, size);
```
LukaszRozmej: you can avoid construction, just resize/clear. Also if those are always fixed size, then you don't really need `ArrayPoolList`, you could just use plain pooled array

Marchhill:
Seems to be some problem on kurtosis devnet, investigating

LukaszRozmej:
# Deep review

The prior automated review cycles have already covered most of the obvious issues (semaphore disposal, NRE-on-pop, merge conflict, `.vscode` files, `BlockAccessList` allocation per Return, `NotImplementedException`). The findings below are from a closer pass focused on what's still on-branch.

## Pool & lifecycle correctness

### C1. Pool leak across blocks on any exception between `Get` and `Return` (high)
`ParallelTxProcessorWithWorldStateManager.Setup` (`BlockAccessListManager.cs:548-559`) wipes `_inUse`/`_perTxBal` but **never reconciles `_free` and `_available`**:

```csharp
public void Setup(Block block, BlockExecutionContext blockExecutionContext)
{
    _currentBlock = block;
    _currentCtx = blockExecutionContext;
    _lastBalIndex = block.Transactions.Length + 1;

    int size = _lastBalIndex + 1;
    _inUse?.Dispose();           // drops references but doesn't return processors
    _perTxBal?.Dispose();
    _inUse = new ArrayPoolList<TxProcessorWithWorldState?>(size, size);
    _perTxBal = new ArrayPoolList<BlockAccessList?>(size, size);
}
```

`_free`/`_available` are only initialized in the constructor. If any non-tx `Get` (pre/post slots: `StoreBeaconRoot`, `ApplyBlockhashStateChanges`, `ProcessWithdrawals`, `ProcessExecutionRequests`) is followed by an exception that prevents the corresponding `MergeAndReturnBal`, that processor is permanently consumed:
- `_free` loses one entry that is never refilled
- `_available.CurrentCount` is one lower than the "free" count for the rest of the manager's lifetime
- After enough leaks, `Get` deadlocks on `_available.Wait()` because `_free` is empty

The `using TxProcessorLease` pattern protects the parallel tx workers, but the system-contract slots (`Get(0)`, `Get(int.MaxValue)`) are acquired without a lease. `BlockProcessor.ProcessBlock` (`BlockProcessor.cs:124-172`) calls `StoreBeaconRoot` → `ApplyBlockhashStateChanges` → ... → `SetBlockAccessList`. Anything throwing between those points (a transient state error, a cancellation, an OOM) leaves a processor stranded.

**Given that there's a problem on the kurtosis devnet, this is plausibly the cause** — long-running nodes accumulate enough exceptional paths to deplete the pool.

Fix: have `Setup` push any orphaned `_inUse[i]` back to `_free` (and adjust the semaphore) before re-creating the array, or wrap the system-contract `Get`s in a try/finally that always Returns.

### C2. `_systemAccountReadSuppressionDepth` not reset on processor reuse (low)
`TracedAccessWorldState` is now reused across transactions via the pool. `Setup` calls `WorldState.Clear()` (which clears `_generatingBlockAccessList`) but does not reset `_systemAccountReadSuppressionDepth`. The counter is decremented by `SystemAccountReadSuppressionScope.Dispose`. If any caller ever leaks the scope (forgets to dispose, exception escapes between `using` parens), the depth stays nonzero across pool reuse and silently suppresses system-account reads for **subsequent unrelated transactions** processed by the recycled `TracedAccessWorldState`.

Pre-PR each tx got a fresh `TracedAccessWorldState` so this was self-healing. With pooling, it isn't. Add a `_systemAccountReadSuppressionDepth = 0` to `Clear()` (or to a dedicated `Reset()` called at Setup).

### C3. `Setup(block)` invariant: cached pre-slot reuse vs block change
`Get(int? balIndex)` returns the cached `_inUse[balIndex]` if non-null. After `Setup`, `_inUse` is freshly empty, so the cache is correctly cleared — but only if `Setup` runs. Confirm there is no path where `BlockProcessor` re-enters processing without calling `Setup`. Today there isn't, but the assumption is implicit. A `Debug.Assert(_inUse[idx] is null || _inUse[idx].Block == _currentBlock)` inside `Get` would catch a future violation.

## API design

### A1. `IBlockAccessListManager.GetTxProcessor` is the unsafe twin of `RentTxProcessor` (medium)
Both methods exist on the interface:

```csharp
ITransactionProcessorAdapter GetTxProcessor(int? balIndex = null);
TxProcessorLease RentTxProcessor(int balIndex)
    => new(GetTxProcessor(balIndex), this, balIndex);
```

`GetTxProcessor` returns a raw adapter that the caller has no obligation to release. It is used for the system-contract slots (`StoreBeaconRoot`, `ProcessWithdrawals`, etc.) precisely because those slots are never explicitly Returned — they're cleaned up by the validator's `MergeAndReturnBal`. That asymmetric contract is fragile: a future caller using `GetTxProcessor` for a worker-style slot would silently leak (see C1). At minimum, document the contract on the interface; better, require all callers to either go through `RentTxProcessor` or commit to a fixed set of "managed" indices that the validator handles.

### A2. `MergeAndReturnBal` implicitly performs a `Return` (low)
`MergeAndReturnBal(idx, target)` calls `Return(idx)` first to handle slots that bypassed the worker `Return` path (pre/post). The name says "merge", but it also "returns" — and only sometimes. Consider either:
- Renaming to `EnsureReturnedAndMerge(idx, target)` so the side-effect is part of the contract, or
- Splitting into explicit `EnsureReturned(idx)` and `MergeBal(idx, target)` and having `IncrementalValidation`/`SetBlockAccessList` call both explicitly.

The current pattern looks correct but reading it requires holding a non-obvious invariant in your head.

### A3. `ParallelExecutionException` collapses every `InvalidBlockException` to `MalformedTransaction` (low)
```csharp
(innerException as InvalidTransactionException)?.Reason
    ?? TransactionResult.WithDetail(TransactionResult.ErrorType.MalformedTransaction,
                                    $"Parallel execution failure: {innerException.Message}");
```
When `innerException` is a plain `InvalidBlockException` (not the `InvalidTransactionException` subtype), the reason is forced to `MalformedTransaction`. That's misleading for non-transaction-shape errors (gas, BAL mismatch, etc.). Worth either preserving the original exception type or introducing a `ParallelExecutionFailure` reason.

## Performance

### P1. Per-block `ArrayPoolList` heap allocation (low)
`BlockAccessListManager.cs:557-558` allocates two `ArrayPoolList` *wrapper objects* per block. The backing arrays are pooled; the wrappers are not. Minor at block rate, but inconsistent with the PR's "reuse instances" theme — keep two long-lived `ArrayPoolList`s and `Resize` only when capacity is insufficient.

### P2. `Enumerable.SequenceEqual` on the per-tx validation hot path (low — already acknowledged)
`BlockAccessListManager.cs:366` still uses `Enumerable.SequenceEqual(generatedHead.Value.SlotChanges, suggestedHead.Value.SlotChanges)`. Acknowledged as out of scope. Re-flagging because in a PR explicitly aimed at allocations, leaving a per-tx LINQ enumerator allocation in the validation path is awkward — it could go in this PR with a five-line change.

### P3. `_balPool` `ConcurrentBag` instead of `ConcurrentStack` (very low)
`ConcurrentBag` is optimized for thread-affine producer/consumer (each thread has its own segment). The PR's pattern is "many producers (validator), many consumers (workers)" with no thread affinity, which is closer to `ConcurrentStack`/`ConcurrentQueue`'s strength. Likely irrelevant given the small bag size, but `ConcurrentStack` would be the more natural fit.

## Tests

### T1. No test for the exception-leak invariant (high — carried over)
The `EngineModuleTestsV6` claim covers the happy path of `Return`→`MergeAndReturnBal`→`SetResult` ordering. It does **not** cover:
- A block where `StoreBeaconRoot` throws — does the next block's `Get(0)` still succeed?
- A block where `ProcessWithdrawals` throws after `Get(int.MaxValue)` — is the post slot recovered?
- 100 consecutive failing blocks — does the pool still produce processors?

These are exactly the scenarios where C1 above bites. A focused unit test on `ParallelTxProcessorWithWorldStateManager` that exercises Setup → Get(0) → simulated-fault → Setup → Get(0) and asserts the second Get succeeds without timing out would close the gap.

### T2. No test for `_systemAccountReadSuppressionDepth` cross-pool leakage (low)
Reuse one `TracedAccessWorldState` across two `Setup` calls; in between, throw out of a `using BeginSystemAccountReadSuppression()` block before disposal (simulate a leaked scope); assert the second tx sees suppression depth 0.

## Summary

The pooling design is correct on the happy path and the prior reviews have shaken out the obvious bugs. **The remaining unknown is whether a path exists where a system-contract slot is acquired but not returned — because that path slowly destroys the pool over the lifetime of a node.** If the kurtosis devnet failure shows up after hours of running (rather than on the first failing block), C1 is the leading suspect.

Beyond C1, the API surface (A1, A2) deserves tightening so future maintainers can't accidentally add a non-leasing caller, and the validation hot path (P2) is worth fixing in this PR rather than punting.

src/Nethermind/Nethermind.Consensus/Processing/BlockAccessListManager.cs
Marchhill: Swapped to array and resizing them. It's not always fixed size, depends on number of txs in each block

Marchhill:
@claude review

src/Nethermind/Nethermind.Consensus/Processing/BlockAccessListManager.cs
```diff
@@ -473,48 +496,150 @@ private interface ITxProcessorWithWorldStateManager
...
+            // (lets pre/post callers share state across calls).
+            TxProcessorWithWorldState? existing = _inUse[balIndex.Value];
+            if (existing is not null) return existing;
+
+            TxProcessorWithWorldState processor = StaticPool<TxProcessorWithWorldState>.Rent();
+
+            // Install a fresh BAL before Setup so the worker has somewhere to record changes.
+            processor.WorldState.SetGeneratingBlockAccessList(StaticPool<BlockAccessList>.Rent());
+            processor.Setup(_currentBlock, _currentCtx, balIndex.Value);
+            _inUse[balIndex.Value] = processor;
```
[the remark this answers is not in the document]
Marchhill: I don't see how the setup can throw an exception, unless it something like OOM which is a critical error anyway

Marchhill:
@claude review

benaadams (approved):

LukaszRozmej:
@claude final review

src/Nethermind/Nethermind.Consensus/Processing/BlockAccessListManager.cs
```diff
@@ -473,48 +495,188 @@ private interface ITxProcessorWithWorldStateManager
...
+            int previousSize = _lastBalIndex + 1;
+            _lastBalIndex = block.Transactions.Length + 1;
+
+            ReclaimAndResize(_lastBalIndex + 1, previousSize);
+        }
+
+        public TxProcessorWithWorldState Get(int? balIndex = null)
+        {
+            if (_currentBlock is null)
+                throw new InvalidOperationException($"{nameof(_currentBlock)} was not initialized.");
```
LukaszRozmej: minor: ThrowXXX method

src/Nethermind/Nethermind.Consensus/Processing/BlockAccessListManager.cs
```diff
@@ -473,48 +495,188 @@ private interface ITxProcessorWithWorldStateManager
...
+
+            balIndex = ClampBalIndex(balIndex ?? 0);
+
+            // Re-entrant Get for the same balIndex returns the already-acquired processor
+            // (lets pre/post callers share state across calls).
+            TxProcessorWithWorldState? existing = _inUse[balIndex.Value];
```
LukaszRozmej, line 580: should this be like `Volitile.Read`?

src/Nethermind/Nethermind.Consensus/Processing/BlockAccessListManager.cs
```diff
@@ -473,48 +495,188 @@ private interface ITxProcessorWithWorldStateManager
...
+            if (existing is not null) return existing;
+
+            TxProcessorWithWorldState processor = RentProcessor();
+
+            // Install a fresh BAL before Setup so the worker has somewhere to record changes.
+            processor.WorldState.SetGeneratingBlockAccessList(StaticPool<BlockAccessList>.Rent());
+            processor.Setup(_currentBlock, _currentCtx, balIndex.Value);
+            _inUse[balIndex.Value] = processor;
```
LukaszRozmej, line 588: and Volitile.Write or Interlocked - and it it fails it returns the processor?

LukaszRozmej: basically if all `_inUse` should be thread-safe?

src/Nethermind/Nethermind.Consensus/Processing/BlockAccessListManager.cs
```diff
@@ -473,48 +495,188 @@ private interface ITxProcessorWithWorldStateManager
...
+        /// <summary>Detaches the worker's populated BAL into the per-tx slot and recycles
+        /// the processor immediately, so workers never block on the validator.</summary>
+        public void Return(int balIndex)
+        {
+            balIndex = ClampBalIndex(balIndex);
+
+            TxProcessorWithWorldState? processor = _inUse[balIndex];
+            if (processor is null) return;
+
+            _perTxBal[balIndex] = processor.WorldState.GetGeneratingBlockAccessList();
```
LukaszRozmej, line 601: does `_perTxBal` need to be threadsafe?

Marchhill: I don't think so, added comment addressing

src/Nethermind/Nethermind.Consensus/Processing/BlockAccessListManager.cs
Marchhill: I don't think so, added comment addressing

src/Nethermind/Nethermind.Consensus/Processing/BlockAccessListManager.cs
Marchhill: I don't think so, added comment addressing

Marchhill:
@claude review

LukaszRozmej:
## Re-review after `6a6bcc4`

### ✅ Resolved

**Processor-leak in `ReclaimAndResize`** — fixed cleanly by deferring the `_lastBalIndex` update until *after* cleanup:

```csharp
int previousSize = _lastBalIndex + 1;
int newLastBalIndex = block.Transactions.Length + 1;
ReclaimAndResize(newLastBalIndex + 1, previousSize);
_lastBalIndex = newLastBalIndex;
```

`ClampBalIndex` inside `Return(i)` now sees the *old* `_lastBalIndex` during cleanup, so every `i ∈ [0, previousSize-1]` clamps to itself and the previous block's post-exec slot is correctly drained. Dropping the trailing `Array.Clear` calls is also a nice tightening — every used slot is nulled inline now.

**`ThrowNotInitialized` helper** — done with `[DoesNotReturn]`.

### 🟡 Documentation nit — the new threading comment overstates the invariant

The new comment says:

> Pre/post (idx 0 and `_lastBalIndex`) are written only by the main thread

That's not quite right for **idx 0**. `IncrementalValidation` runs on `Task.Run(...)` (validator thread) and its very first line writes idx 0:

```csharp
_parallelTxProcessorWithWorldStateManager.MergeAndReturnBal(0, GeneratedBlockAccessList);
//   → Return(0): _inUse[0] = null, _perTxBal[0] = bal
//   → _perTxBal[0] = null
```

So idx 0 has writers on two threads:

- Main thread writes `_inUse[0] = processor` during `StoreBeaconRoot` / `ApplyBlockhashStateChanges` (`Get(0)`).
- Validator thread writes `_inUse[0] = null`, `_perTxBal[0] = bal`, `_perTxBal[0] = null` (`MergeAndReturnBal(0, ...)`).

The code is still **correct** — the publication chain is:

1. Main `Get(0)` write → `Task.Run` release fence → validator read.
2. Validator `Return(0)` write → validator task finishes → `incrementalValidationTask.GetAwaiter().GetResult()` acquire fence → next block's `Setup`/`ReclaimAndResize` reads it on the main thread.

…but the comment as written suggests idx 0 is main-thread-only, which is what'll trip up future readers. Suggested rewording:

```csharp
// Thread-safety note for _inUse / _perTxBal:
//   Each balIndex slot has at most one writer at a time, but the writer thread
//   can change between phases. Within a single block:
//     idx 0:        main thread writes during pre-exec (Get); validator thread
//                   writes null during MergeAndReturnBal(0). Ordered by Task.Run
//                   launch (main → validator) and validator-task join (validator
//                   → main, observed at next block's ReclaimAndResize).
//     idx _lastBalIndex (post-exec): main thread only.
//     idx 1..len:   each slot owned end-to-end by one parallel-loop iteration.
//                   Validator reads happen after gasResults[i-1].SetResult,
//                   whose pairing with GetResult() is the publication barrier.
//   No two threads ever touch the same slot concurrently, so plain reads/writes
//   are sufficient — Volatile/Interlocked would be redundant fencing.
```

### 🟡 Still present (minor, non-blocking)

Flagging so they're not lost — none of these block merge:

- **`RentProcessor` TOCTOU**: `Volatile.Read(ref _processorCount) > 0 && _processors.TryDequeue(...)` can over-allocate under sustained contention; the surplus is discarded by `ReturnProcessor`'s cap check. Self-correcting.
- **`StaticPool<BlockAccessList>.SetMaxPooledCount` from a static ctor** — process-global mutation. No other consumers exist in-tree today (rechecked), but it's fragile if anyone later adds one. An instance-owned `ConcurrentBag<BlockAccessList>` would sidestep this.
- **`DefaultTxCount = 10000`** — ~160KB of arrays per manager that grow but never shrink. Probably fine.
- **`RentTxProcessor` as a default interface method** — minor virtual-dispatch overhead on the per-tx hot path.

### Bottom line

🔴 leak is genuinely closed and the fix is in the right place. Happy to approve once the threading comment is tightened (or leave it as-is and treat that as polish).

LukaszRozmej (approved):

Issue #11203, opened 2026-04-17
`eth_feeHistory`: wrong error code & message for non-existent block


svlachakis:
**Affected tests:** test_21

### Description

When `eth_feeHistory` is called with a `newestBlock` that does not exist (a far-future block
number like `0xFFFFFFF0`), Nethermind returns error code `-32002` while Geth returns `-32603`.
The error messages also differ significantly. This is the same class of issue seen in
`eth_call` test_25/test_26 — Nethermind and Geth disagree on which error code to use for
an unavailable block.

### Steps to Reproduce

```json
// test_21 — newestBlock 0xFFFFFFF0 (4294967280), does not exist
{
  "jsonrpc": "2.0",
  "method": "eth_feeHistory",
  "params": ["0x1", "0xFFFFFFF0", [0, 50, 100]],
  "id": 1
}
```

### Expected (Geth)

```json
{
  "error": {
    "code": -32603,
    "message": "upstream does not have the requested block yet"
  }
}
```

### Actual (Nethermind)

```json
{
  "error": {
    "code": -32002,
    "message": "newestBlock: Block is not available"
  }
}
```

Issue #11202, opened 2026-04-17
`eth_feeHistory`: `baseFeePerBlobGas` returned in wrong block order


svlachakis:
**Affected tests:** test_13

### Description

When `eth_feeHistory` is called for a range that spans the EIP-4844 activation boundary (where
`baseFeePerBlobGas` transitions from `0x0` to non-zero values), Nethermind returns the
`baseFeePerBlobGas` array with two entries swapped compared to what Geth returns.

The `baseFeePerBlobGas` array should contain one entry per block plus the projected next value,
in ascending block order. Nethermind has entries at index [1] and [2] swapped, which means the
reported blob base fee history is incorrect for those blocks.

### Steps to Reproduce

```json
// test_13 — 3 blocks ending at 0x12a2D00, covers blob gas active range
{
  "jsonrpc": "2.0",
  "method": "eth_feeHistory",
  "params": ["0x3", "0x12a2D00", [3, 100]],
  "id": 1
}
```

### Expected (Geth)

```json
{
  "result": {
    "baseFeePerBlobGas": [
      "0x76b5324b4",
      "0x858bd8b2e",
      "0x8ae4cb18d",
      "0x7b7609c19"
    ]
  }
}
```

### Actual (Nethermind)

```json
{
  "result": {
    "baseFeePerBlobGas": [
      "0x76b5324b4",
      "0x7b7609c19",
      "0x858bd8b2e",
      "0x8ae4cb18d"
    ]
  }
}
```

### Diff

```
~ result.baseFeePerBlobGas[1]: "0x7b7609c19" -> "0x858bd8b2e"
~ result.baseFeePerBlobGas[2]: "0x858bd8b2e" -> "0x8ae4cb18d"
```

### Notes

Index [0] (`0x76b5324b4`) and the pattern of values match — only [1] and [2] are transposed.
This suggests an off-by-one or ordering bug specific to how Nethermind assembles the
`baseFeePerBlobGas` result array when traversing a block range. The companion `gasUsedRatio`
and `blobGasUsedRatio` arrays are also slightly different for [1] and [2], consistent with
the blocks being swapped.

Issue #11201, opened 2026-04-17
`eth_feeHistory`: `gasUsedRatio` and `blobGasUsedRatio` truncated to 10 decimal places


svlachakis:
**Affected tests:** test_01, test_03, test_04, test_05, test_07, test_08, test_09, test_10, test_11, test_12, test_14, test_15, test_16, test_17, test_18, test_19, test_20 (17 of 20 tests)

### Description

Nethermind truncates `gasUsedRatio` and `blobGasUsedRatio` floating-point values to 10 decimal
places in `eth_feeHistory` responses. Geth returns the full IEEE 754 double precision value.
This is the single most widespread diff across the entire `eth_feeHistory` test suite, affecting
17 out of 20 tests.

The truncation is consistent: Nethermind always cuts at 10 significant decimal digits, rounding
the last digit. This appears to be a deliberate formatting choice rather than a precision loss in
calculation, but it produces values that differ from Geth and from the Ethereum JSON-RPC spec
which does not mandate truncation.

### Steps to Reproduce

```json
// test_01 — 1 block, block 0x867A80
{
  "jsonrpc": "2.0",
  "method": "eth_feeHistory",
  "params": ["0x1", "0x867A80", [25, 75]],
  "id": 1
}
```

### Expected (Geth)

```json
{
  "result": {
    "gasUsedRatio": [0.678584082336891]
  }
}
```

### Actual (Nethermind)

```json
{
  "result": {
    "gasUsedRatio": [0.6785840823]
  }
}
```

### More examples from the test suite

| Test   | Field              | Geth (full precision)    | Nethermind (truncated)  |
|--------|--------------------|--------------------------|-------------------------|
| test_03 | gasUsedRatio[0]   | 0.9920745287298334       | 0.9920745287            |
| test_03 | gasUsedRatio[3]   | 0.9985787551520126       | 0.9985787552 (rounded!) |
| test_10 | gasUsedRatio[0]   | 0.9980612069624781       | 0.998061207             |
| test_12 | blobGasUsedRatio  | 0.16666666666666666      | 0.1666666667            |
| test_12 | blobGasUsedRatio  | 0.3333333333333333       | 0.3333333333            |

### Notes

The rounding direction is inconsistent — sometimes the 10th digit is rounded up, sometimes
truncated. For example `0.9985787551520126` becomes `0.9985787552` (rounded up) while
`0.9920745287298334` becomes `0.9920745287` (truncated). This suggests standard floating-point
[...]

PR #11322, opened 2026-04-23
`eth_feeHistory` fixes

### LINKED ISSUE #11204: `eth_feeHistory`: omitting `rewardPercentiles` parameter accepted by Nethermind, rejected by Geth
**Affected tests:** test_22

### Description

The `eth_feeHistory` method signature is `(blockCount, newestBlock, rewardPercentiles)`.
When `rewardPercentiles` is omitted entirely (only 2 parameters provided), Nethermind
treats it as optional and returns a valid result. Geth treats it as required and returns
a `-32602 Invalid params` error.

The Ethereum JSON-RPC spec and EIP-1559 define `rewardPercentiles` as a required parameter.
Nethermind's leniency here could mask client bugs where the parameter is accidentally dropped.

### Steps to Reproduce

```json
// test_22 — rewardPercentiles parameter omitted entirely
{
  "jsonrpc": "2.0",
  "method": "eth_feeHistory",
  "params": ["0x1", "latest"],
  "id": 1
}
```

### Expected (Geth)

```json
{
  "error": {
    "code": -32602,
    "message": "missing value for required argument 2"
  }
}
```

### Actual (Nethermind)

```json
{
  "result": {
    "baseFeePerBlobGas": ["0x2478a40", "0x2478a40"],
    "baseFeePerGas": ["0x280e6359", "0x2884f386"],
    "blobGasUsedRatio": [0.2857142857],
    "gasUsedRatio": [0.5462487667],
    "oldestBlock": "0x17b9f30"
  }
}
```

### Notes

This is a case where Nethermind is more permissive than Geth. Whether this is a bug or a feature depends on the spec interpretation — the spec says the parameter is required, so Nethermind should return an error. Fixing this to match Geth is correct since it only affects malformed requests. Note that the returned result also exhibits [Issue 1](https://github.com/NethermindEth/nethermind/issues/11201).

svlachakis:
Fixes Closes Resolves https://github.com/NethermindEth/nethermind/issues/11204 & https://github.com/NethermindEth/nethermind/issues/11201 & https://github.com/NethermindEth/nethermind/issues/11202 & https://github.com/NethermindEth/nethermind/issues/11203

## Changes

# eth_feeHistory: Geth Compatibility Fixes

Fixes a lot of the test failures in the `eth_feeHistory` test suite by aligning Nethermind's behaviour with Geth and the Ethereum JSON-RPC spec.

---

## Fix 1 — `rewardPercentiles` is a required parameter (test_22)

The spec marks `rewardPercentiles` as required. Nethermind previously accepted calls with only 2 arguments and returned a successful result. Geth returns `-32602`.

`rewardPercentiles` is now required at the RPC method signature level (`double[]`, no default). An explicit null-guard in `FeeHistoryOracle.GetFeeHistory` returns a clean error if the framework passes null for an absent argument, making the contract visible in code rather than relying on framework behaviour.

---

## Fix 2 — Full IEEE 754 precision for `gasUsedRatio` / `blobGasUsedRatio` (test_01, test_03–test_20, 17 tests)

`DoubleConverter` was serialising double values with the `"0.##########"` format, capping output at 10 decimal digits and rounding the last. Geth returns the full double precision value.

Changed both `DoubleConverter` and `DoubleArrayConverter` to use the `"R"` (round-trip) format specifier, which emits the minimum number of digits required to uniquely represent the IEEE 754 value — identical to what Geth produces.

---

## Fix 3 — Geth-compatible missing-argument error message (all endpoints)

https://github.com/ethereum/go-ethereum/blob/8e2107dc39dc9dab132150ec915e7ac299f9eb48/rpc/json.go#L331
https://github.com/ethereum/go-ethereum/blob/8e2107dc39dc9dab132150ec915e7ac299f9eb48/rpc/json.go#L349

When a required argument is omitted, Nethermind returned:

```json
{"code": -32602, "message": "Invalid params", "data": "Incorrect parameters count, expected: 3, actual: 2"}
```

## Fix 4 — `baseFeePerBlobGas[blockCount]` reports next-block estimate, not current-block fee

**Root cause:**  
`GetFeeHistory` fills `baseFeePerBlobGas` with `blockCount + 1` entries — indices `[0..blockCount-1]` are the actual per-block fees, and index `[blockCount]` is the next block's estimated blob base fee (matching the same pattern as `baseFeePerGas[blockCount]`).

The bug was that the pre-loop "next" slot was populated with `info.BaseFeePerBlobGas` (the newest block's *current* fee) instead of the projected next fee.

**Fix:**  
Added `BaseFeePerBlobGasEst` to `BlockFeeHistorySearchInfo`, computed using the same pattern Geth uses:  
https://github.com/ethereum/go-ethereum/blob/master/eth/gasprice/feehistory.go

1. `nextExcessBlobGas = BlobGasCalculator.CalculateExcessBlobGas(block.Header, spec)`  
2. `BaseFeePerBlobGasEst = TryCalculateFeePerBlobGas(nextExcessBlobGas, spec.BlobBaseFeeUpdateFraction)`

Pre-4844 blocks (`ExcessBlobGas == null`) correctly produce `0` for both the current and next slot, mirroring Geth's `if excessBlobGas != nil` guard.

**Affected:**  
`eth_feeHistory` calls that span or land near the EIP-4844 activation boundary, where the newest block's current blob fee differs from its projected next-block fee.

## Fix 5 — `eth_feeHistory` returns wrong error code for non-existent block

**Root cause:**  
When `newestBlock` does not exist, `FeeHistoryOracle.GetFeeHistory` returned `-32002` (`ResourceUnavailable`) with message `"newestBlock: Block is not available"`.

Geth returns `-32603` (`InternalError`) with message `"upstream does not have the requested block yet"`.

**Fix:**  
Single-line change in `FeeHistoryOracle.cs` — replace `ErrorCodes.ResourceUnavailable` with `ErrorCodes.InternalError` and align the message to Geth's.

**Affected:**  
`eth_feeHistory` calls where `newestBlock` refers to a block that does not exist on the node (e.g. a far-future block number).

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Core.Test/Json/DoubleArrayConverterTests.cs
+++ b/src/Nethermind/Nethermind.Core.Test/Json/DoubleArrayConverterTests.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Text.Json;
+using FluentAssertions;
 using Nethermind.Serialization.Json;
 using NUnit.Framework;
 
@@ -12,10 +13,10 @@ namespace Nethermind.Core.Test.Json;
 [TestFixture]
 public class DoubleArrayConverterTests : ConverterTestBase<double[]>
 {
-    static readonly DoubleArrayConverter converter = new();
+    static readonly DoubleArrayConverter _converter = new();
 
     [TestCaseSource(nameof(RoundtripTestCases))]
-    public void Test_roundtrip(double[] value) => TestConverter(value, static (a, b) => a.AsSpan().SequenceEqual(b), converter);
+    public void Test_roundtrip(double[] value) => TestConverter(value, static (a, b) => a.AsSpan().SequenceEqual(b), _converter);
 
     static IEnumerable<TestCaseData> RoundtripTestCases()
     {
@@ -32,8 +33,8 @@ public class DoubleArrayConverterTests : ConverterTestBase<double[]>
     [TestCase(0.3333333333333333, "[0.3333333333333333]")]
     public void Write_PreservesFullIeee754Precision(double value, string expectedJson)
     {
-        JsonSerializerOptions options = new() { Converters = { converter } };
-        string json = System.Text.Json.JsonSerializer.Serialize(new[] { value }, options);
-        Assert.That(json, Is.EqualTo(expectedJson));
+        JsonSerializerOptions options = new() { Converters = { _converter } };
+        string json = JsonSerializer.Serialize(new[] { value }, options);
+        json.Should().Be(expectedJson, "double array serialization must preserve full IEEE 754 round-trip precision");
     }
 }
--- a/src/Nethermind/Nethermind.Core.Test/Json/DoubleArrayConverterTests.cs
+++ b/src/Nethermind/Nethermind.Core.Test/Json/DoubleArrayConverterTests.cs
@@ -3,8 +3,8 @@
 
 using System;
 using System.Collections.Generic;
+using System.Text.Json;
 using Nethermind.Serialization.Json;
-
 using NUnit.Framework;
 
 namespace Nethermind.Core.Test.Json;
@@ -23,5 +23,17 @@ public class DoubleArrayConverterTests : ConverterTestBase<double[]>
         yield return new TestCaseData(new double[] { 1, 1, 1, 1 }).SetName("All ones");
         yield return new TestCaseData(new double[] { 0, 0, 0, 0 }).SetName("All zeros");
         yield return new TestCaseData(Array.Empty<double>()).SetName("Empty array");
+        yield return new TestCaseData(new double[] { 1.0 / 3.0, 1.0 / 6.0, 0.678584082336891, 0.9985787551520126 }).SetName("Full precision values");
+    }
+
+    [TestCase(0.678584082336891, "[0.678584082336891]")]
+    [TestCase(0.9985787551520126, "[0.9985787551520126]")]
+    [TestCase(0.16666666666666666, "[0.16666666666666666]")]
+    [TestCase(0.3333333333333333, "[0.3333333333333333]")]
+    public void Write_PreservesFullIeee754Precision(double value, string expectedJson)
+    {
+        JsonSerializerOptions options = new() { Converters = { converter } };
+        string json = System.Text.Json.JsonSerializer.Serialize(new[] { value }, options);
+        Assert.That(json, Is.EqualTo(expectedJson));
     }
 }
--- a/src/Nethermind/Nethermind.Core.Test/Json/DoubleArrayConverterTests.cs
+++ b/src/Nethermind/Nethermind.Core.Test/Json/DoubleArrayConverterTests.cs
@@ -13,7 +13,7 @@ namespace Nethermind.Core.Test.Json;
 [TestFixture]
 public class DoubleArrayConverterTests : ConverterTestBase<double[]>
 {
-    static readonly DoubleArrayConverter _converter = new();
+    private static readonly DoubleArrayConverter _converter = new();
 
     [TestCaseSource(nameof(RoundtripTestCases))]
     public void Test_roundtrip(double[] value) => TestConverter(value, static (a, b) => a.AsSpan().SequenceEqual(b), _converter);
--- a/src/Nethermind/Nethermind.JsonRpc/Modules/Eth/FeeHistory/FeeHistoryOracle.cs
+++ b/src/Nethermind/Nethermind.JsonRpc/Modules/Eth/FeeHistory/FeeHistoryOracle.cs
@@ -293,7 +293,7 @@ namespace Nethermind.JsonRpc.Modules.Eth.FeeHistory
             foreach (double percentile in rewardPercentiles)
             {
                 double thresholdGasUsed = (ulong)(blockInfo.GasUsed * percentile / 100);
-                while (txIndex < rewardsInBlock.Count && sumGasUsed < thresholdGasUsed)
+                while (txIndex + 1 < rewardsInBlock.Count && sumGasUsed < thresholdGasUsed)
                 {
                     txIndex++;
                     sumGasUsed += rewardsInBlock[txIndex].GasUsed;
--- a/src/Nethermind/Nethermind.JsonRpc/Modules/Eth/FeeHistory/FeeHistoryOracle.cs
+++ b/src/Nethermind/Nethermind.JsonRpc/Modules/Eth/FeeHistory/FeeHistoryOracle.cs
@@ -169,8 +169,8 @@ namespace Nethermind.JsonRpc.Modules.Eth.FeeHistory
 
             if (historyInfo is null)
             {
-                return ResultWrapper<FeeHistoryResults>.Fail("newestBlock: Block is not available",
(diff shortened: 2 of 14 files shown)


svlachakis:
@claude review again

svlachakis:
@claude review again

svlachakis:
@claude I added one more fix, review fix 4 also

svlachakis:
@claude review again

src/Nethermind/Nethermind.Core.Test/Json/DoubleArrayConverterTests.cs
```diff
@@ -3,25 +3,38 @@
...
-
 using NUnit.Framework;
 
 namespace Nethermind.Core.Test.Json;
 
 [TestFixture]
 public class DoubleArrayConverterTests : ConverterTestBase<double[]>
 {
-    static readonly DoubleArrayConverter converter = new();
+    static readonly DoubleArrayConverter _converter = new();
```
LukaszRozmej:
```suggestion
    private static readonly DoubleArrayConverter _converter = new();
```

svlachakis:
@claude review fix 5

src/Nethermind/Nethermind.Serialization.Json/DoubleConverter.cs
```diff
@@ -22,55 +20,11 @@ public override double Read(
...
             JsonSerializerOptions options)
         {
-            writer.WriteStartArray();
-            foreach (double value in values)
-            {
-                writer.WriteRawValue(value.ToString("0.0#########", CultureInfo.InvariantCulture), skipInputValidation: true);
-            }
-            writer.WriteEndArray();
+            if (double.IsNaN(value) || double.IsInfinity(value))
+                throw new JsonException($"The value '{value}' is not a valid JSON number.");
```
LukaszRozmej: maybe seperate ThrowXXX method for performance

src/Nethermind/Nethermind.Serialization.Json/DoubleArrayConverter.cs
```diff
@@ -0,0 +1,65 @@
...
+            {
+                string s = reader.GetString();
+                if (s is null) throw new JsonException("Expected a JSON array string, got null.");
+                return JsonSerializer.Deserialize<double[]>(s)
+                    ?? throw new JsonException($"Could not deserialize double array from string: {s}");
+            }
+
+            if (reader.TokenType != JsonTokenType.StartArray)
+            {
+                throw new JsonException();
```
LukaszRozmej: Maybe add messages to exceptions, also use `ThrowXXX("message")` helper

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/IEthRpcModule.cs
```diff
@@ -50,7 +50,7 @@ public interface IEthRpcModule : IRpcModule
             Description = "Returns block fee history.",
             IsSharable = true,
             ExampleResponse = "{\"baseFeePerGas\": [\"0x116c1cbb03\", \"0x10c3714c06\"], \"gasUsedRatio\": [0.3487305666666667, 0.3], \"oldestBlock\": \"0xc7e5ff\", \"reward\": [[\"0x3b9aca00\",\"0x3b9aca00\"], [\"0x0\",\"0x3bb24dfa\"]]}")]
-        ResultWrapper<FeeHistoryResults> eth_feeHistory(int blockCount, BlockParameter newestBlock, double[]? rewardPercentiles = null);
+        ResultWrapper<FeeHistoryResults> eth_feeHistory(int blockCount, BlockParameter newestBlock, double[] rewardPercentiles);
```
LukaszRozmej, line 53: I think this being optional is not an issue? We can be better than Geth.

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/FeeHistory/FeeHistoryOracle.cs
```diff
@@ -105,24 +106,29 @@ private readonly record struct BlockFeeHistorySearchInfo(
...
-                double blobGasUsedRatio = CalculateBlobGasUsedRatio(b, out UInt256 feePerBlobGas);
+                UInt256 nextFeePerBlobGas = UInt256.Zero;
+                if (b.Header.ExcessBlobGas.HasValue)
+                {
+                    // Uses parent-block spec for BlobBaseFeeUpdateFraction, matching Geth's feehistory.go.
+                    // At hard-fork boundaries this is one block off, but it is intentional for compatibility.
+                    ulong nextExcessBlobGas = BlobGasCalculator.CalculateExcessBlobGas(b.Header, spec) ?? 0;
+                    if (!BlobGasCalculator.TryCalculateFeePerBlobGas(nextExcessBlobGas, spec.BlobBaseFeeUpdateFraction, out nextFeePerBlobGas))
+                        nextFeePerBlobGas = UInt256.Zero;
+                }
```
LukaszRozmej:
```suggestion
UInt256 nextFeePerBlobGas =
    b.Header.ExcessBlobGas.HasValue &&
    BlobGasCalculator.TryCalculateFeePerBlobGas(BlobGasCalculator.CalculateExcessBlobGas(b.Header, spec) ?? 0, spec.BlobBaseFeeUpdateFraction, out UInt256 fee)
    ? fee
    : UInt256.Zero;
```

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/FeeHistory/FeeHistoryOracle.cs
```diff
@@ -232,9 +242,16 @@ private void CleanupCache()
...
+            double[] rewardPercentiles)
+        {
+            if (blockInfo.BlockTransactionsLength != 0)
+                return CalculatePercentileValues(blockInfo, rewardPercentiles, blockInfo.RewardsInBlocks);
+
+            ArrayPoolList<UInt256> zeros = new(rewardPercentiles.Length);
+            for (int i = 0; i < rewardPercentiles.Length; i++)
+                zeros.Add(UInt256.Zero);
+            return zeros;
+        }
```
LukaszRozmej: It is struct it should default to zero
```suggestion
            double[] rewardPercentiles) => blockInfo.BlockTransactionsLength == 0
                ? new ArrayPoolList<UInt256>(rewardPercentiles.Length, rewardPercentiles.Length)
                : CalculatePercentileValues(blockInfo, rewardPercentiles, blockInfo.RewardsInBlocks);
```

LukaszRozmej (approved):
Not sure that we need to make the parameter required.

svlachakis:
@LukaszRozmej spec marks this as required https://github.com/ethereum/execution-apis/blob/703b1b4b01df5ebdc02ebe904388790557b6ab98/docs-api/api/methods/eth_feeHistory.mdx#L72

svlachakis:
@claude review again please

LukaszRozmej (approved):

benaadams (approved):

PR #11325, opened 2026-04-23
eth_sendRawTransaction fix

### LINKED ISSUE #11206: `eth_sendRawTransaction`: wrong error code and vague message for malformed RLP
**Affected tests:** test_01, test_02, test_03

### Description

When `eth_sendRawTransaction` is called with malformed RLP data, Nethermind returns error code
`-32000` with the vague message `"Invalid RLP."` while Geth returns `-32602` (Invalid params)
with the specific message `"rlp: value size exceeds available input length"`.

Two problems:
1. **Wrong error code**: `-32602` (Invalid params) is the correct code for malformed input —
   the transaction data is a caller-supplied parameter that fails validation. `-32000` is a
   generic server error and is incorrect here.
2. **Vague error message**: `"Invalid RLP."` gives no information about what is wrong.
   Geth's message `"rlp: value size exceeds available input length"` points directly to the
   cause. The test fixture itself expects `-32000` with the Geth-style message, so both the
   code and the message need fixing.

### Steps to Reproduce

```json
// test_02 / test_03 — single byte 0xd4, incomplete RLP list
{
  "jsonrpc": "2.0",
  "method": "eth_sendRawTransaction",
  "params": ["0xd4"],
  "id": 1
}
```

```json
// test_01 — full transaction bytes with truncated RLP
{
  "jsonrpc": "2.0",
  "method": "eth_sendRawTransaction",
  "params": ["0xFF63808459682f07825208943d504cb2b11a45e7d1f9646ede40c60d52deec2580802da0e06b96410c954281132f2b403c0d25e0b04dd83f4ac5dffef07eb800c01718b7a07dc8bc872a54a40547c033b78081fd417e64bb96cae12ed5ad50b82f951636dd"],
  "id": 1
}
```

### Expected (Geth)

```json
{
  "error": {
    "code": -32602,
    "message": "rlp: value size exceeds available input length"
  }
}
```

### Actual (Nethermind)

```json
{
  "error": {
    "code": -32000,
    "message": "Invalid RLP."
  }
}
```

### Notes

The fixture expects `-32000` with `"rlp: value size exceeds available input length"` — i.e.
Nethermind has the right error code per the fixture but uses the wrong message. Geth uses
`-32602` which is arguably more correct since malformed input is a client error (invalid
params), not a server error. To fully match Geth, both the code and the message need to change.

test_02 and test_03 use identical input (`0xd4`) — they are duplicate test cases.

svlachakis:
Fixes Closes Resolves https://github.com/NethermindEth/nethermind/issues/11206

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.JsonRpc.Test/Modules/Eth/EthRpcModuleTests.cs
+++ b/src/Nethermind/Nethermind.JsonRpc.Test/Modules/Eth/EthRpcModuleTests.cs
@@ -1599,6 +1599,15 @@ public partial class EthRpcModuleTests
         Assert.That(actual.Error!.Code, Is.EqualTo(ErrorCodes.TransactionRejected));
     }
 
+    [TestCase("0xd4", TestName = "IncompleteRlpList")]
+    [TestCase("0xFF63808459682f07825208943d504cb2b11a45e7d1f9646ede40c60d52deec2580802da0e06b96410c954281132f2b403c0d25e0b04dd83f4ac5dffef07eb800c01718b7a07dc8bc872a54a40547c033b78081fd417e64bb96cae12ed5ad50b82f951636dd", TestName = "TruncatedTransaction")]
+    public async Task Eth_sendRawTransaction_malformed_rlp_returns_invalid_params(string rawHex)
+    {
+        using Context ctx = await Context.Create();
+        string serialized = await ctx.Test.TestEthRpc("eth_sendRawTransaction", rawHex);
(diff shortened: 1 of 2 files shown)


svlachakis:
@claude review

PR #11326, opened 2026-04-23
eth_simulateV1 - MovePrecompileToAddress validation (-38022)

### LINKED ISSUE #11223: `eth_simulateV1` `MovePrecompileToAddress` self-reference not validated (-38022 missing)
**Spec error code:** `-38022` — "MovePrecompileToAddress referenced itself in replacement"

### Description

The spec requires `-38022` when a `stateOverride` sets `MovePrecompileToAddress` to the same address as the key (self-reference). Nethermind has no `-38022` constant in `ErrorCodes.cs` and no such validation.

svlachakis:
Fixes Closes Resolves https://github.com/NethermindEth/nethermind/issues/11223

## Changes

Adds missing validation for the `MovePrecompileToAddress` self-reference error case per the execution-apis spec.

**What changed:**

- **`ErrorCodes.cs`** — new constant:
  - `MovePrecompileSelfReference = -38022`

- **`SimulateTxExecutor.cs`** — validation added in the per-block loop before simulation runs:
  - Source is a precompile and `MovePrecompileToAddress == address` → `-38022`
  - Uses per-simulated-block spec (`specProvider.GetSpec(blockNumber, timestamp)`) for accurate precompile detection across fork boundaries

- **`EthSimulateTestsHiveBase.cs`** — flipped `shouldSucceed: true → false` for `multicall-move-to-address-itself-reference-38022`

- **`EthSimulateTestsBlocksAndTransactions.cs`** — regression test  
  `eth_simulateV1_MovePrecompileToAddress_invalid_override_returns_error` asserting error code `-38022` and the spec-exact message

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

## Testing

https://github.com/NethermindEth/nethermind/actions/runs/24844409778/job/72727297344

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No


svlachakis:
@claude review again

svlachakis:
@claude 

The -38023 finding is incorrect. I investigated this. The actual hive conformance suite has a test called multicall-move-two-accounts-to-same-38023 where two precompiles point to the same destination — and that test expects success (pass=true), not -38023. We ran the full suite and all four related tests pass. The scenario you describe (two precompile sources → same destination) is valid per the conformance tests, which are the ground truth. -38023 is not implemented because it would break a passing conformance test. The PR description will be updated to remove any mention of -38023 to avoid this confusion.

LukaszRozmej (approved):

benaadams (approved):

PR #11327, opened 2026-04-23
XDC Fix mutable V2 config spec

batrr:
## Changes                                                                                                                                                                                                                                  
                                                                            
  - Fix mutable spec bug in `XdcChainSpecBasedSpecProvider`: previously `GetXdcSpec` called `ApplyV2Config` directly on the shared spec instance returned by `GetSpec`, so fetching with round Y would overwrite the V2 fields of a spec      
  previously fetched with round X                                                                                                                                                                                                             
  - Introduce per-`(blockIndex, roundIndex)` spec cache `GetXdcSpec` now clones the base spec and caches an isolated copy per block transition × V2 config bucket, so concurrent or sequential calls with different rounds never mutate each
   other
                                                                                                                                                                                                                                 
  ## Types of changes
                                                                                                                                                                                                                                              
  - [x] Refactoring
                                                                                                                                                                                                                                              
  ## Testing
                                                                                                                                                                                                                                              
  #### Requires testing                                                     

  - [x] Yes

  - [x] Yes


cicr99 (approved):
CI is not successful yet, but other than that looks good to me

ak88 (approved):

PR #11332, opened 2026-04-23
fix remove balance checks from eth_simulate_v1

### LINKED ISSUE #11216: `eth_simulateV1` `validation: false` not respected — balance check still runs
**Spec behavior:** When `validation` is `false`, all transaction validation (balance, fees, nonces) must be skipped.

### Description

When `eth_simulateV1` is called with `"validation": false`, Nethermind still
applies balance checks and returns `-38014 "insufficient sender balance"` for
calls from addresses with no balance. The spec requires all validation to be
bypassed when `validation: false` is set — the simulation must proceed with
whatever state exists.

### Steps to Reproduce

```json
{
  "blockStateCalls": [{
    "calls": [{
      "from": "0x1111111111111111111111111111111111111111",
      "to": "0x2222222222222222222222222222222222222222",
      "value": "0x52B7D2DCC80CD2E40000"
    }]
  }],
  "validation": false
}
```

**Expected:** success (no error)
**Actual:** `-38014 "insufficient sender balance"`

DarkLord017:
Fixes Closes Resolves #11216 

## Changes

- Added Simulate(Transaction, ITxTracer) to the ` ITransactionProcessor.cs ` interface
- validate: false path calls Simulate() instead of Trace()
- Added checks in BuyGas and PayValue to skipBalanceValidation when needed 
- Added tests 

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


DarkLord017:
@claude review

DarkLord017:
I was seeing go-ethereum for the expected behaviour but in api_tests i saw that geth also returns insufficients funds error when validator is false pls see 
https://github.com/ethereum/go-ethereum/blob/ba215fd927b2e6377a64cb0b4ce668f106763119/internal/ethapi/api_test.go#L1568
https://github.com/ethereum/go-ethereum/blob/ba215fd927b2e6377a64cb0b4ce668f106763119/internal/ethapi/api_test.go#L2463

src/Nethermind/Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs
```diff
@@ -610,8 +610,12 @@ protected virtual TransactionResult BuyGas(Transaction tx, IReleaseSpec spec, IT
...
             if (UInt256.SubtractUnderflow(in senderBalance, in tx.ValueRef, out UInt256 balanceLeft))
             {
-                TraceLogInvalidTx(tx, $"INSUFFICIENT_SENDER_BALANCE: ({tx.SenderAddress})_BALANCE = {senderBalance}");
-                return TransactionResult.InsufficientSenderBalance;
+                if (validate)
+                {
+                    TraceLogInvalidTx(tx, $"INSUFFICIENT_SENDER_BALANCE: ({tx.SenderAddress})_BALANCE = {senderBalance}");
+                    return TransactionResult.InsufficientSenderBalance;
+                }
+                balanceLeft = UInt256.Zero;
```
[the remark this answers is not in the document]
DarkLord017: won't senderReservedGasPayment be equal to zero as validation = false sets base fee per gas as zero which will set effective gas price as zero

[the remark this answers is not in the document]
DarkLord017: @claude pls reply

DarkLord017:
@claude see my comment on medium and reply to it

DarkLord017:
@claude re-review

DarkLord017:
> #### Low — Test only checks absence of error, no state assertion



> **Claude finished @DarkLord017's task in 3m 40s** —— [View job](https://github.com/NethermindEth/nethermind/actions/runs/24888474877)
> 
> ### Re-review
> * [x]  Read rules files
> * [x]  Read changed files and diff
> * [x]  Analyze for correctness, performance, and style
> * [x]  Post final review
> 
> **Summary:** The approach is sound — introducing a dedicated `SkipBalanceValidation` flag and a `Simulate` method is cleaner than the previous `balanceLeft = UInt256.Zero` inline fix. The Medium finding from the first review was correctly dismissed. However there is one **High** issue: the new `Simulate` method added to `ITransactionProcessor` breaks a concrete test implementation that doesn't implement it, causing a compilation error. Two Low findings from the first review still stand.
> 
> ### Findings
> Severity	#	Description
> 🔴 Critical	0	—
> 🟠 High	1	`TestTransactionProcessor` missing `Simulate` → compile error
> 🟡 Medium	0	_(previous Medium dismissed — confirmed correct)_
> 🔵 Low	2	LINQ in test; test only checks absence of error
> #### 🟠 High — `TestTransactionProcessor` doesn't implement `Simulate`, compile error
> `DisposableScopeOverridableEnvTests.cs:118` has a private class that directly implements `ITransactionProcessor` without a default implementation fallback. The new non-default `Simulate` method is missing, so `Nethermind.Evm.Test` will fail to compile. [Fix this →](https://claude.ai/code?q=In%20src%2FNethermind%2FNethermind.Evm.Test%2FOverridableEnv%2FDisposableScopeOverridableEnvTests.cs%2C%20the%20private%20class%20TestTransactionProcessor%20%28line%20118%29%20implements%20ITransactionProcessor%20but%20is%20missing%20the%20new%20Simulate%20method%20added%20in%20the%20PR.%20Add%20%60public%20TransactionResult%20Simulate%28Transaction%20transaction%2C%20ITxTracer%20txTracer%29%20%3D%3E%20throw%20new%20NotImplementedException%28%29%3B%60%20to%20fix%20the%20compile%20error.%20Branch%3A%2011216-eth_simulatev1-validation-false-not-respected-balance-check-still-runs)
> 
> The fix is one line — match the existing pattern in that class:
> 
> ```cs
> public TransactionResult Simulate(Transaction transaction, ITxTracer txTracer) =>
>     throw new NotImplementedException();
> ```
> 
> Alternatively, add a default interface implementation so future implementations aren't silently broken:
> 
> ```cs
> // in ITransactionProcessor.cs
> TransactionResult Simulate(Transaction transaction, ITxTracer txTracer) => Trace(transaction, txTracer);
> ```
> 
> #### 🔵 Low — LINQ `First()` in test
> `EthSimulateTestsBlocksAndTransactions.cs:476` — `result.Data[0].Calls.First()` violates `coding-style.md` ("No LINQ when a simple `for`/`foreach` works"). `Calls` is indexable; use `[0]`.
> 
> #### 🔵 Low — Test only checks absence of error, no state assertion
> The test verifies the bug is fixed but a silent no-op would also pass. At minimum, asserting `call.Status == 1` (already done) is good — but no assertion verifies that `AddressB` received the value or that `AddressA`'s state is as expected. A comment explaining why state assertions are omitted would help future readers.
> 
> ### What's correct
> * `SkipBalanceValidation` flag is a clean, self-documenting name and correctly scoped to only apply when `ShouldValidateGas` returns `false` (both fee fields = 0 + `SkipValidation` set) — so legacy txs with non-zero `GasPrice` still validate normally.
> * `PayValue` guards against underflow correctly: only skips deduction when balance is insufficient, normal path otherwise.
> * `SimulateAndCommit = Commit | SkipValidation | SkipBalanceValidation` composite flag is the right design.
> * The previous `PayValue` guard inconsistency concern is resolved — the new code uses `SkipBalanceValidation` consistently in both `BuyGas` and `PayValue`.

1. But how can we test this it is simulate call not state changing call no state change will happen 

2. Calls is IReadOnlyCollection<SimulateCallResult>, not IReadOnlyList, so we can't use [0] directly. We need to either use First() or cast it , am i right ?

DarkLord017:
@claude re-review and pls check n verify the comments i made regarding your prev reviews

src/Nethermind/Nethermind.Evm/TransactionProcessing/ITransactionProcessor.cs
LukaszRozmej, line 43: Could we refactor this so ITransactionProcessor has only one with some default flags and all the rest are extension methods? Or maybe default interface methods?

Can be in next PR

LukaszRozmej (approved):

src/Nethermind/Nethermind.JsonRpc.Test/Modules/Eth/Simulate/EthSimulateTestsBlocksAndTransactions.cs
```diff
@@ -446,6 +446,44 @@ public async Task Test_eth_simulate_log_index_increments_across_transactions()
         Assert.That(tx1Logs[0].LogIndex, Is.EqualTo(2ul));
     }
 
+    [Test]
+    public async Task Test_eth_simulate_no_validation_skips_balance_check()
```
LukaszRozmej, line 450: could this be a test case for any existing tests? (Probably no)

src/Nethermind/Nethermind.Evm/TransactionProcessing/ITransactionProcessor.cs
DarkLord017: Sure , doing that

svlachakis (approved):

src/Nethermind/Nethermind.JsonRpc.Test/Modules/Eth/Simulate/EthSimulateTestsBlocksAndTransactions.cs
DarkLord017: Checked the tests , imo all cover different cases

PR #11337, opened 2026-04-24
analyzers: NETH003 — file name must match the contained type name

asdacap:
## Rationale

When an agent (or a human) needs to open a type, the first move is to look for a file whose name matches the type name. If no such file exists, a second search is needed — \`grep\`, symbol search, or another tool call. At scale across hundreds of sessions this wasted round-trip adds up.

The repo already has a de-facto convention that files are named after their single contained type, but nothing enforced it. This PR adds \`NETH003\`, a Roslyn \`DiagnosticAnalyzer\` that makes the convention machine-checked.

## What the rule does

For each \`.cs\` file that contains **exactly one top-level type**, \`NETH003\` warns when the file base name does not match the type's identifier. The following forms are accepted:

- **Exact match**: \`Foo.cs\` for \`class Foo\`.
- **Numeric order prefix**: \`NN_TypeName.cs\` — a leading \`digits_\` prefix is stripped before matching (e.g. \`00_Olympic.cs\` for \`class Olympic\`). This preserves the intentional ordering convention in \`Nethermind.Specs/Forks/\`.
- **Attribute suffix**: if the type name ends with \`Attribute\`, the file may use either the full name or the stripped name (e.g. \`RequiresSecurityReview.cs\` for \`RequiresSecurityReviewAttribute\`).
- **Partial descriptor**: if the type is \`partial\`, the file may use \`TypeName.Descriptor.cs\` (e.g. \`BlockTree.AcceptVisitor.cs\`, \`VirtualMachine.std.cs\`). The descriptor is intentionally unconstrained.
- **Multiple types**: files with zero or more than one top-level type are ignored entirely (covers intentional groupings like \`Metrics.cs\` and paired interface+implementation files).

Severity is \`Warning\`, which the existing \`TreatWarningsAsErrors=true\` in \`Directory.Build.props\` escalates to an error — so the rule is fully enforced from day one.

## Changes

- **\`Nethermind.Analyzers/FileNameMatchesTypeNameAnalyzer.cs\`** — new analyzer (\`NETH003\`), syntax-tree–only (no semantic model needed).
- **\`Nethermind.Analyzers/AnalyzerReleases.Unshipped.md\`** — release tracking entry.
- **\`Nethermind.Analyzers.Test/FileNameMatchesTypeNameAnalyzerTests.cs\`** — 43 parameterized test cases covering all rule branches including numeric prefix.
- **~130 file renames** across the solution to clear pre-existing violations. All renames preserve 100% of file content; git detects them as renames. \`Nethermind.Specs/Forks/\` files are **not** renamed — the analyzer explicitly allows their \`NN_TypeName.cs\` ordering prefix.

## Testing

- \`dotnet test Nethermind.Analyzers.Test\` — 43/43 pass
- \`dotnet build -c release Nethermind.slnx\` — 0 errors, 0 NETH003 warnings

## Type of change

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Testing

- [x] New analyzer tests added (\`FileNameMatchesTypeNameAnalyzerTests\`)
- [x] Full solution build clean
- [ ] Affected tests run (analyzer-only change; no runtime behaviour modified)

🤖 Generated with [Claude Code](https://claude.com/claude-code) (claude-sonnet-4-6)


src/Nethermind/Nethermind.Crypto/EthereumEcdsa.zkevm.cs
```diff
@@ -7,6 +7,7 @@
 
 namespace Nethermind.Crypto;
 
+#pragma warning disable NETH003 // Build variant: only one of EthereumEcdsa.std.cs / EthereumEcdsa.zkevm.cs is compiled per build
```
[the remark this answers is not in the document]
asdacap: But.... the std also have the pragma.

src/Nethermind/Nethermind.Crypto/ProtectedData.T.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Nethermind.Crypto
 {
-    public abstract class ProtectedData<T> : ProtectedData where T : IDisposable
+    public abstract partial class ProtectedData<T> : ProtectedData where T : IDisposable
```
asdacap: Why?

asdacap: The file `ProtectedData.T.cs` uses the descriptor form `TypeName.Descriptor.cs` — the `T` segment is the descriptor. NETH003 only accepts that form for `partial` types. Without `partial`, it would see `ProtectedData.T` ≠ `ProtectedData` and fire.

Renaming to `ProtectedData.cs` isn't possible since the non-generic base class already occupies that name. Adding `partial` is the minimal change; the class split across two files is already logically coherent (base state in `ProtectedData.cs`, generic impl in `ProtectedData.T.cs`).

*(This reply was added by Claude Sonnet 4.6)*

svlachakis (approved):

src/Nethermind/Nethermind.Analyzers/FileNameMatchesTypeNameAnalyzer.cs
```diff
@@ -0,0 +1,132 @@
...
+            : fileBaseName;
+    }
+
+    // Returns true for "TypeName.SomeDescriptor" where the descriptor is non-empty.
+    private static bool HasDescriptorPrefix(string fileBaseName, string prefix) =>
+        fileBaseName.Length > prefix.Length + 1
+        && fileBaseName[prefix.Length] == '.'
+        && fileBaseName.StartsWith(prefix, StringComparison.Ordinal);
+
+    private static void CollectTopLevelTypes(SyntaxList<MemberDeclarationSyntax> members, List<MemberDeclarationSyntax> result)
```
flcl42, line 115: Can return false asap if there are 0 or >1 types. IF returns false than break validation

LukaszRozmej (approved):

PR #11339, opened 2026-04-24
sync: fix beacon headers re-downloaded after #10876 changed GetBlockHashOnMainOrBestDifficultyHash

asdacap:
## Changes

PR #10876 (facce1bcb3) fixed FCU canonical chain corruption by making \`GetBlockHashOnMainOrBestDifficultyHash\` return \`null\` for post-merge blocks when \`HasBlockOnMainChain == false\`. This is correct for canonical chain integrity, but it broke \`HeadersSyncFeed.ProcessPersistedPortion\`.

Beacon headers are inserted via \`BeaconHeaderInsert\` without being processed, so they always have \`HasBlockOnMainChain == false\`. The old \`ProcessPersistedPortion\` used \`FindHeader(blockNumber, ...)\` which routes through \`GetBlockHashOnMainOrBestDifficultyHash\` — now returning \`null\` for every beacon header — so the feed treated all already-persisted beacon headers as absent and re-downloaded them from peers.

**Fix:** bypass \`BlockTree.FindHeader\` entirely in \`ProcessPersistedPortion\`. Access the chain level info and header store directly via \`IChainLevelInfoRepository.LoadLevel\` and \`IHeaderStore.Get(shouldCache: false)\`, which correctly retrieve beacon headers regardless of \`HasBlockOnMainChain\`.

**Optimization on top:** replace the per-header point-lookup loop with \`IHeaderStore.FindReversedHeaders\`, which opens a single DB iterator over the \`[start, end]\` block-number range, pre-loads all headers into an in-memory dictionary, then walks backward via parent hash — one sequential scan instead of O(n) random Gets.

## Types of changes

- [x] Optimization

#### Requires testing

- [x] Yes

- [x] Yes

#### Notes on testing

Parameterized \`FindReversedHeaders\` tests in \`HeaderStoreTests\` covering: happy path with iterator and fallback backends, unknown end hash, chain gap (partial result), and fork disambiguation. Existing \`FastHeadersSyncTests\` and \`BeaconHeadersSyncTests\` pass unchanged.

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain/Headers/HeaderStore.cs
+++ b/src/Nethermind/Nethermind.Blockchain/Headers/HeaderStore.cs
@@ -109,9 +109,9 @@ public class HeaderStore(
         }
     }
 
-    public IOwnedReadOnlyList<BlockHeader> FindReversedHeaders(long endBlockNumber, Hash256 endBlockHash, long count)
+    public IOwnedReadOnlyList<BlockHeader> FindReversedHeaders(long endBlockNumber, Hash256 endBlockHash, int count)
     {
-        Dictionary<ValueHash256, BlockHeader> prefetched = new();
+        Dictionary<ValueHash256, BlockHeader> prefetched = new(count);
 
         if (headerDb is ISortedKeyValueStore sorted)
         {
@@ -136,7 +136,7 @@ public class HeaderStore(
 
         if (cursor is null) return ArrayPoolList<BlockHeader>.Empty();
 
-        ArrayPoolList<BlockHeader> result = new((int)count) { cursor };
+        ArrayPoolList<BlockHeader> result = new(count) { cursor };
         while (result.Count < count && cursor.ParentHash is not null)
         {
             long parentNumber = cursor.Number - 1;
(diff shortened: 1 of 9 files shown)


LukaszRozmej:
@claude re-review

src/Nethermind/Nethermind.Blockchain/Headers/HeaderStore.cs
```diff
@@ -108,6 +109,48 @@ public void InsertBlockNumber(Hash256 blockHash, long blockNumber)
         }
     }
 
+    public IOwnedReadOnlyList<BlockHeader> FindReversedHeaders(long endBlockNumber, Hash256 endBlockHash, int count)
+    {
+        Dictionary<ValueHash256, BlockHeader> prefetched = new(count);
```
LukaszRozmej, line 114: IMO probably better to use array/list (ArrayPoolListRef!) here. You will have parents by index and only first search will be costly.

LukaszRozmej (approved):

damian-orzechowski (approved):

PR #11340, opened 2026-04-24
fix(tracer): guard NativeCallTracer against empty call stack on CREATE transactions

### LINKED ISSUE #11324: `NativeCallTracer` crashes with `ArrayPoolList.get_Item` out of range on CREATE txs (spaceneth)
**Description**
`NativeCallTracer.BuildResult()` throws an `ArrayPoolList<T>.get_Item` out-of-range exception when attempting to trace any contract-creation (`CREATE`) transaction. This completely breaks tracing for deployments on `spaceneth` environments via both `debug_traceCall` and in-process `IGethStyleTracer.Trace`.

**Steps to Reproduce**
1. Start a disposable spaceneth node via Docker with the Debug module enabled:
   `docker run -d -p 8545:8545 nethermindeth/nethermind:release-1.36.0 --config spaceneth --JsonRpc.Host 0.0.0.0 --JsonRpc.EnabledModules Eth,Net,Web3,Debug`
2. Wait ~5 seconds for the RPC to become available.
3. Execute a `CREATE` trace via `debug_traceCall` by omitting the `to` field (ensure the `from` address is funded, or use an arbitrary funded EOA from spaceneth):
   ```bash
   curl -s -X POST http://localhost:8545 \
     -H 'Content-Type: application/json' \
     -d '{
       "jsonrpc": "2.0",
       "method": "debug_traceCall",
       "params": [
         {
           "from": "0x7e5f4552091a69125d5dfcb7b8c2659029395bdf",
           "data": "0x60006000f3",
           "gas": "0x7a120"
         },
         "latest",
         {"tracer": "callTracer"}
       ],
       "id": 1
     }'
   ```

**Actual behavior**
The tracer crashes and returns a JSON-RPC error. This is an issue because it makes it impossible to trace contract deployments, which breaks downstream plugins and CI testing environments that rely on inspecting `CREATE` call frames before accepting transactions.

**Expected behavior**
The tracer should return a valid trace result with a `CREATE` frame, identical to the behavior observed when running the same `debug_traceCall` request on a standard Sepolia node.

**Screenshots**
N/A

 - Operating System: Linux (Docker)
 - Version: `release-1.36.0` and `release-1.36.2` (Fully broken); `release-1.37.0-rc2` (In-process path broken)
 - Installation Method: Docker
 - Consensus Client: NethDev (spaceneth) / None (No beacon chain)

**Additional context**
* **Scope:** This issue appears to be `spaceneth`-specific. Running the exact same RPC request against `Nethermind/v1.36.0` on a real Sepolia node returns a valid `CREATE` trace without crashing.
* **Partial fix in 1.37.0-rc2:** Testing against `release-1.37.0-rc2` shows that the `debug_traceCall` RPC path is now functioning correctly for `CREATE` transactions. However, invoking the tracer via the in-process path (`IGethStyleTracer.Trace`) still throws the exact same `ArrayPoolList` exception.
* **Root Cause Speculation:** The `spaceneth` execution context likely causes `BuildResult()` to hit a code path where the CREATE frame isn't pushed onto the frame stack during trace collection, leaving the list shorter than expected when it attempts to read it back.

**Logs**
```json
{
  "jsonrpc": "2.0",
  "error": {
    "code": -32602,
    "message": "Specified argument was out of the range of valid values. (Parameter 'index')",
    "data": "System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values.\n (Parameter 'index')\n   at Nethermind.Core.Collections.ArrayPoolList`1.get_Item(Int32 index)\n   at Nethermind.Blockchain.Tracing.GethStyle.Custom.Native.Call.NativeCallTracer.BuildResult()"
  }
}
```

manusw7:
Closes #11324.

## Changes

- In `NativeCallTracer.MarkAsSuccess` and `MarkAsFailed`, added an early-return guard when `_callStack` is empty. The EVM is bypassed (so `ReportAction` is never called) when `PrepareDeployment` detects an address collision or `env.CodeInfo` is null, leaving `_callStack` empty and causing an `ArgumentOutOfRangeException` on the unconditional `_callStack[0]` access.
- In `NativeCallTracer.MarkAsFailed`, also guard `_error!.Value` against null. When `DeployContract` fails after a successful EVM run no EVM exception is reported, so `_error` is null.
- Added three regression tests to `GethLikeCallTracerTests`:
  - `Test_CallTrace_TopLevelCreate_Success` — top-level CREATE completes successfully
  - `Test_CallTrace_TopLevelCreate_Revert` — initcode reverts inside the EVM
  - `Test_CallTrace_TopLevelCreate_AddressCollision_NoCrash` — `PrepareDeployment` fails because the target address is pre-seeded with code; the EVM is never entered and `_callStack` stays empty

## Types of changes

#### Requires testing

- [x] Yes

- [x] Yes

#### Notes on testing

The third regression test (`AddressCollision_NoCrash`) directly reproduces the spaceneth crash: it pre-seeds the CREATE deployment address with code so that `PrepareDeployment` returns `false`, the EVM is bypassed, `_callStack` stays empty, and calling `MarkAsFailed` previously threw
`ArgumentOutOfRangeException`. The test asserts `Throws.Nothing` after the fix.

- [x] No

- [x] Yes

Tracing a contract-creation (`CREATE`) transaction with `callTracer` on a spaceneth / MemDb node could crash with `ArgumentOutOfRangeException` inside `NativeCallTracer.MarkAsFailed` or `MarkAsSuccess` when the contract address was already occupied (address collision). The crash left
`debug_traceCall` returning an error response and any in-process caller (e.g. via `IGethStyleTracer.Trace`) receiving an exception. Both paths are now protected.

## Remarks

Root cause (in detail): `TransactionProcessor.ExecuteEvmCall` has two early-exit paths for CREATE transactions that jump to `Complete` without ever entering `VirtualMachine.ExecuteTransaction`:
1. `PrepareDeployment` returns `false` (target address has non-zero nonce or code).
2. `env.CodeInfo` is null (empty/null initcode).

`ReportAction` is only called from inside the VM, so when either path is taken `_callStack` remains empty. `IsTracingReceipt = true` on `GethLikeTxTracer` means `MarkAsFailed`/`MarkAsSuccess` are always called, crashing on `_callStack[0]`.

The release `1.37.0-rc2` partial fix addressed the `debug_traceCall` path by removing `BlockOverrides.ApplyOverrides(block.Header)` from `GethStyleTracer`, which coincidentally stopped triggering the collision condition on spaceneth. The in-process `IGethStyleTracer.Trace` path was
unaffected by that change and continued to crash. 

This PR fixes the root cause in the tracer itself.


manusw7:
@claude review again

src/Nethermind/Nethermind.Evm.Test/Tracing/GethLikeCallTracerTests.cs
```diff
@@ -540,6 +542,114 @@ public void Test_CallTrace_SelfDestruct_OnlyTopCall()
         Assert.That(callTrace, Is.EqualTo(expectedCallTrace));
     }
 
+    [Test]
+    public void Test_CallTrace_TopLevelCreate_Success()
```
LukaszRozmej: I think those test can be deduplicated to test-cases? (ask claude to do it)

svlachakis (approved):

alexb5dh (approved):

PR #11348, opened 2026-04-24
debug_traceCallMany: strip filler blocks from overrides result

grubbyTimes:
`debug_traceCallMany` overrides path returned traces for every simulated block, including the filler blocks `SimulateTxExecutor` inserts when `BlockOverride.Number` has gaps. Result count no longer matched the input bundles.

Compute each bundle's target block number (same `overrideNumber ?? lastBlockNumber + 1` rule the executor uses) and filter `simulationResult.Data` down to that set. Regression test covers a non-contiguous gap between two bundles.


LukaszRozmej:
@claude review

grubbyTimes:
Added the gap=1 boundary case for "Test gap"; skipping "Logic coupling" (extracting a helper just relocates the duplication unlessSimulateTxExecutor also calls it — better as a separate IsFiller refactor PR) and "No completeness check" (Debug.Assert compiles out in Release, so it wouldn't guard prod).

svlachakis:
@claude review

src/Nethermind/Nethermind.JsonRpc/Modules/DebugModule/DebugRpcModule.cs
```diff
@@ -482,6 +482,19 @@ private ResultWrapper<IEnumerable<IEnumerable<GethLikeTxTrace>>> TraceCallManyWi
...
 
+        // SimulateTxExecutor inserts filler blocks between bundles when BlockOverride.Number has gaps.
+        // Pre-compute the block number each bundle targets so we can drop fillers from the result and
+        // keep a 1:1 mapping to the input bundles.
+        HashSet<long> bundleBlockNumbers = new(bundles.Length);
+        long lastBlockNumber = header.Number;
+        foreach (TransactionBundle bundle in bundles)
+        {
+            ulong? overrideNumber = bundle.BlockOverride?.Number;
+            long number = overrideNumber is not null ? (long)overrideNumber.Value : lastBlockNumber + 1;
```
LukaszRozmej: maybe collapse the formula into helper method and reuse it in the Executor too?

```suggestion
            long number = bundle.BlockOverride.GetBlockNumber(lastBlockNumber);
```

where `GetBlockNumber` is an extension method that works with null's too

LukaszRozmej (approved):

src/Nethermind/Nethermind.JsonRpc/Modules/DebugModule/DebugRpcModule.cs
grubbyTimes: Sure nice to have

LukaszRozmej (approved):

batrr (approved):

svlachakis (approved):

PR #11352, opened 2026-04-25
`debug_executionwitness` RPC implementation

### LINKED ISSUE #7767: Implement debug_executionWitness
https://github.com/ethereum/go-ethereum/pull/30613

svlachakis:
Fixes Closes Resolves https://github.com/NethermindEth/nethermind/issues/7767

#### New DTOs (`WitnessForRpc`, `WitnessHeaderForRpc`)
- `state` and `codes` are now emitted as maps in the form `{ "0x<keccak256>": "0x<bytes>" }`, matching geth’s `ExecutionWitness` struct.
- `headers` are now decoded from RLP into JSON header objects using the same field layout as geth’s `types.Header`, instead of raw RLP hex strings.
- The internal `keys` field (Verkle-specific and not part of the geth RPC spec) has been removed from the response.
- The raw `Witness` object is disposed immediately after projection, while the DTOs retain direct references to the underlying byte arrays so no additional copying occurs.

#### `DebugRpcModule`
- Both witness-producing methods now instantiate witnesses via `using Witness witness = ...` and return projected DTOs through `new WitnessForRpc(witness)`.
- This guarantees deterministic disposal of the raw witness while preserving zero-copy DTO materialization.

#### `IDebugRpcModule`
- Return types have been updated to use the new DTOs.
- Both `[JsonRpcMethod]` attributes now explicitly set `IsImplemented = true` and `IsSharable = false` (previously these methods relied on defaults and appeared as not implemented).

#### `EthRpcJsonContext`
- System.Text.Json source-generation registration now includes `WitnessForRpc` and `WitnessHeaderForRpc`.

## Tests
- Assertions have been migrated from `Witness` to `WitnessForRpc`.
- Stateless execution validation still preserves end-to-end witness correctness by calling `IBlockchainBridge.GenerateExecutionWitness` directly, keeping correctness checks independent of the RPC wire format.
- `TestCaseData` entries now include descriptive `TestName` values for improved readability and diagnostics.

- [ ] Optimization
- [X] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No


svlachakis:
@claude review again

svlachakis:
@claude review again

LukaszRozmej (approved):

LukaszRozmej:
@rubo please review

src/Nethermind/Nethermind.JsonRpc/Modules/DebugModule/WitnessForRpc.cs
```diff
@@ -0,0 +1,33 @@
...
+using Nethermind.Core.Collections;
+using Nethermind.Core.Crypto;
+
+namespace Nethermind.JsonRpc.Modules.DebugModule;
+
+public class WitnessForRpc(Witness witness)
+{
+    public IReadOnlyList<WitnessHeaderForRpc> Headers { get; } = ProjectHeaders(witness);
+    public IReadOnlyDictionary<string, byte[]> Codes { get; } = BuildHashMap(witness.Codes);
+    public IReadOnlyDictionary<string, byte[]> State { get; } = BuildHashMap(witness.State);
```
rubo, line 15: Why are these converted to maps in JSON? Geth [doesn't do that](https://github.com/ethereum/go-ethereum/blob/master/core/stateless/encoding.go#L77-L82). Maps are there internal structure not visible in RPC response as far as I understand.

rubo (commented):
What's the motivation behind these changes? Why are headers changed from RLP to a JSON object? Currently, we have it like the one of Reth. Why do we wanna change it to match Geth? Are there any specs?
Do you have a sample JSON from our RPC method?

rubo (commented):
Based on [my conversation](https://discord.com/channels/595666850260713488/1451216716062523507/1498686356216479825) with @jsign, our current JSON output matches the standard. No need to change the format to Geth's.

svlachakis:
@rubo alright then leaving it as is

PR #11354, opened 2026-04-25
Reorg up to lastFinalized block

svlachakis:
Replaces the static 64-block reorg limit with a dynamic [finalized, head] window, capped at 8192 (EIP-2935 HISTORY_SERVE_WINDOW) - we can use any cap we think fits better. 

Three coordinated gates use the same finalized boundary so a reorg that the FCU accepts has its trie state and log-index entries available. Aligns with https://github.com/ethereum/execution-apis/pull/786.

## Changes

1. `ForkchoiceUpdatedHandler.ShouldProceedWithReorg` — limit becomes Clamp(head − finalized, 0, 8192) when finalized is known; falls back to `PruningBoundary` (64) before the first finalized FCU.
2. `ReorgDepthFinalizedStateProvider.FinalizedBlockNumber` — returns the real finalized number from `_blockTree.FinalizedHash` instead of `BestKnownNumber − 64.` Drives trie dirty-cache eviction so state for [finalized, head] stays in memory.
3. `LogIndexStorage.GetCompressionBoundary` (renamed from `GetLastReorgableBlockNumber`) — uses `FinalizedBlockNumber` as the compression cutoff so log data above finalized stays uncompressed and reorg-safe.

All three read the same boundary: `_blockTree.FinalizedHash`

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing


svlachakis:
@claude review again very deeply based on the spec and the reorg handling around state management.

svlachakis:
@claude review again

asdacap:
I recommend add a method in `IWorldStateManager` asking if a block can be added on top of a strte. Do not use `IPruningConfig`. That will not work with flat.

svlachakis:
@claude review again

svlachakis:
@claude review again

svlachakis:
@claude review again, on the FinalizedBlockProviderAdapter point I think it's not that important

svlachakis:
@claude review again, I applied @asdacap's suggestion better.

asdacap:
`HasState` and `CanReorgOn` (or give a better name), is two different thing. On archive node for example it could be that the state exist and can be used to process block (eth_call), but it may not be able to be added and become head and put in the state (reorg).

svlachakis:
@claude review again

flcl42 (approved):

src/Nethermind/Nethermind.Db/LogIndex/LogIndexStorage.cs
```diff
@@ -222,13 +223,14 @@ private IEnumerable<IDb> DBColumns
         private bool _stopped;
         private bool _disposed;
 
-        public LogIndexStorage(IDbFactory dbFactory, ILogManager logManager, ILogIndexConfig config)
+        public LogIndexStorage(IDbFactory dbFactory, ILogManager logManager, ILogIndexConfig config, IFinalizedBlockProvider? finalizedBlockProvider = null)
```
asdacap: Dont make it optional please.

asdacap (approved):

svlachakis:
I need to rebase this to https://github.com/NethermindEth/nethermind/pull/11388

svlachakis:
closing in favor of https://github.com/NethermindEth/nethermind/pull/11396

PR #11363, opened 2026-04-26
Refactor `TxPoolInfoProvider` for per-sender efficiency, allocation cleanup, and blob coverage

### LINKED ISSUE #11310: Refactor: TxPoolInfoProvider - GetInfo
see : https://github.com/NethermindEth/nethermind/pull/11305#issuecomment-4298463493

svlachakis:
Fixes Closes Resolves https://github.com/NethermindEth/nethermind/issues/11310

#### Motivation

`txpool_contentFrom(address)` was scanning the full mempool and then doing a single `TryGetValue(address)` on the result — O(N senders) work to extract O(K) data.  
`txpool_status` was building two large per-sender dictionaries just to sum their counts. And `TxPoolInfoProvider` had a long-standing TODO (`// only std txs are picked here. Should we add blobs?`) that left blob transactions out of `txpool_content` / `txpool_contentFrom` / `txpool_inspect` / `txpool_status` entirely.

This PR addresses all three on the same surface.

### ITxPoolInfoProvider

- Adds `GetSenderInfo(Address)` returning a new `TxPoolSenderInfo` DTO (pending and queued nonce-keyed maps for one sender).
- Adds `GetCounts()` returning a `TxPoolCounts` record struct (`int Pending`, `int Queued`).
- `GetInfo()` is unchanged at the call-site level.

### TxPoolInfoProvider

- Extracts the per-sender split logic (sort by nonce, walk against account nonce) into a single `SplitByNonce` helper used by both `GetInfo()` and `GetSenderInfo()`. Same helper has a count-only twin (CountPending) used by `GetCounts()`.
- Replaces the per-sender OrderBy enumerator allocation with `ArrayPool<Transaction>.Shared.Rent` + `Array.Sort` + a static `NonceComparer`. Pooled buffers are returned in finally with `clearArray: true` so transaction references don't get pinned.
- All three methods now merge standard and blob (light) transactions per sender before splitting. A sender's nonce sequence is single across both pools, so a tx in the blob pool with nonce continuous to a tx in the standard pool ends up correctly in pending. In practice `TxTypeTxFilter` already prevents a sender from having both types simultaneously, so the merge is a defensive safety net rather than a hot path.

### ITxPool

- Adds `GetPendingLightBlobTransactionsBySender(Address)` overload, mirroring the existing per-address standard-tx accessor and the pattern already used by `BalanceTooLowFilter`.
- `TxPool` and `NullTxPool` updated.

### RPC layer

- `TxPoolRpcModule.txpool_contentFrom` now calls GetSenderInfo(address) directly. `TxPoolContentFrom` constructor now takes `TxPoolSenderInfo`; the per-address `TryGetValue` lookup that was happening inside the DTO is gone.
- `TxPoolRpcModule.txpool_status` now calls `GetCounts()`. `TxPoolStatus` constructor takes `TxPoolCounts` directly.

## Behavior changes visible to RPC consumers

- `txpool_content`, `txpool_contentFrom`, `txpool_inspect` now include blob transactions (matches Geth's `TxPool.Content()`, which iterates all subpools). Light variant — sidecar blobs/proofs are not exposed in the RPC payload, same as Geth.
- `txpool_status` counts now include blob transactions for the same reason.

## Out of scope

Pending/queued classification remains nonce-continuity-only. This matches existing Nethermind behavior for standard transactions; blob transactions inherit the same simple rule. Geth additionally factors in cumulative balance (and, for blobs, fee-budget edges and a limbo state for reorged-included txs). Closing that gap affects both standard and blob output and is a behavior change worth its own focused PR — tracking separately.

## Tests

- `TxPoolInfoProviderTests` expanded with consolidated `[TestCaseSource]` for `SenderInfoCases` (mixed gap, all-pending, all-queued, unordered input) and `CountCases` (mixed, all-pending, all-queued); plus dedicated cases for blob merging in `GetInfo` / `GetSenderInfo` / `GetCounts`, and an assertion that `GetSenderInfo` does not scan the full pool.
- `TxPoolRpcModuleTests` updated: `txpool_status` tests now stub `GetCounts()`, `txpool_contentFrom` tests now stub `GetSenderInfo()`, mock fixtures simplified accordingly.
- 14/14 `TxPool` tests and 9/9 JsonRpc TxPool tests pass. Full solution builds clean (50 projects, 0 warnings).

## Performance notes (qualitative)

- `txpool_contentFrom`: O(N senders) → O(K txs for the queried sender). Real win for pools with many senders.
- `txpool_status`: avoids materialising two per-sender dictionaries. Pure counting path, no result allocation beyond the `TxPoolCounts` struct.
- `GetInfo()`: replaces `OrderBy` allocation per sender with pooled-buffer sort. Slight extra work to merge the (usually empty) blob pool. Net negligible to positive on non-blob chains.

No benchmarks attached — these are structural improvements, not microoptimisations of an already-hot path.

- [X] Optimization
- [X] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No


svlachakis:
@claude review again. some points.

- Medium 1: switched to indexer assignment in SplitByNonce so a duplicate nonce (theoretical, given TxTypeTxFilter) cannot crash the RPC handler.
- Medium 2: added default interface implementations on the two new ITxPoolInfoProvider methods and the new ITxPool.GetPendingLightBlobTransactionsBySender(Address) overload. External implementers continue to compile and degrade to the slow path; the in-tree TxPoolInfoProvider and TxPool keep the efficient overrides.
- Doc comment trimmed; BuildTransactions test helper rewritten as a for loop.
- Capacity estimate: standardBySender.Count + blobBySender.Count is the tight upper bound given that TxTypeTxFilter keeps a sender out of both pools — Math.Max would under-allocate when blob senders are disjoint from standard senders. Keeping sum.
- NonceComparer: IComparer<T>.Compare is Compare(T? x, T? y) in modern .NET — dropping the ? raises CS8767. The ! documents that the comparer is invoked from Array.Sort on a fully populated buffer.

src/Nethermind/Nethermind.TxPool/TxPoolInfoProvider.cs
```diff
@@ -1,60 +1,222 @@
...
+        return (pending, queued);
+    }
+
+    private static int CountPending(Transaction[] transactions, int length, UInt256 accountNonce)
+    {
+        int pending = 0;
+        UInt256 expectedNonce = accountNonce;
 
-                if (queued.Count != 0)
+        Transaction[] sorted = ArrayPool<Transaction>.Shared.Rent(length);
```
newmanifold: could `ArrayPoolList` be used instead? It also clears array on `Dispose`

src/Nethermind/Nethermind.TxPool/TxPoolInfoProvider.cs
```diff
@@ -1,60 +1,222 @@
...
+    private static Transaction[] MergeOwned(Transaction[]? standardTransactions, Transaction[]? blobTransactions, out int length)
+    {
+        int standardCount = standardTransactions?.Length ?? 0;
+        int blobCount = blobTransactions?.Length ?? 0;
+        length = standardCount + blobCount;
+
+        if (blobCount == 0) return standardTransactions ?? [];
+        if (standardCount == 0) return blobTransactions!;
+
+        Transaction[] merged = ArrayPool<Transaction>.Shared.Rent(length);
```
newmanifold: could `ArrayPoolList` be used instead?

svlachakis: this will allocate here and will kill the fast path - I think it's better to remain as is

src/Nethermind/Nethermind.TxPool/TxPoolInfoProvider.cs
svlachakis: Applied here, SplitByNonce (and CountPending) has no fast path to lose - it already rents and copies unconditionally

svlachakis:
@claude review again

src/Nethermind/Nethermind.TxPool/TxPoolInfoProvider.cs
```diff
@@ -1,60 +1,209 @@
...
+        standardTransactions!.AsSpan().CopyTo(merged);
+        blobTransactions!.AsSpan().CopyTo(merged.AsSpan(standardCount));
+        return merged;
+    }
+
+    private static void ReturnIfPooled(Transaction[] merged, Transaction[]? standard, Transaction[]? blobs)
+    {
+        if (merged != standard && merged != blobs)
+            ArrayPool<Transaction>.Shared.Return(merged, clearArray: true);
```
LukaszRozmej: I don't think we will run that in zk context, but maybe `SafeArrayPool`?

src/Nethermind/Nethermind.TxPool/TxPoolInfoProvider.cs
```diff
@@ -1,60 +1,209 @@
...
+    // Walks transactions in nonce order: txs whose nonce continues from accountNonce go to
+    // pending, anything beyond a gap goes to queued. Mirrors Geth's pending/queued split.
+    private static (IDictionary<ulong, Transaction> pending, IDictionary<ulong, Transaction> queued)
+        SplitByNonce(Transaction[] transactions, int length, UInt256 accountNonce)
+    {
+        Dictionary<ulong, Transaction> pending = new();
+        Dictionary<ulong, Transaction> queued = new();
+        UInt256 expectedNonce = accountNonce;
+
+        using ArrayPoolList<Transaction> sorted = new(length);
```
LukaszRozmej:
```suggestion
        using ArrayPoolListRef<Transaction> sorted = new(length);
```

LukaszRozmej (approved):

src/Nethermind/Nethermind.TxPool/TxPoolInfoProvider.cs
```diff
@@ -1,60 +1,209 @@
...
+        {
+            if (sorted[i].Nonce == expectedNonce)
+            {
+                pending++;
+                expectedNonce += UInt256.One;
+            }
         }
+
+        return pending;
+    }
```
LukaszRozmej, line 166: I wonder, do we have to do this shananigans and merge two lists in this complex way, or maybe we can do something more clever.

@claude any ideas?

svlachakis: @claude I think I wrote it better than this, please check

batrr (approved):

LukaszRozmej (approved):

LukaszRozmej:
@claude re-review

PR #11072, opened 2026-04-08
XDC Fix duplicate votes in pool 

batrr:
Closes https://github.com/NethermindEth/nethermind/issues/11026
## Changes

- XdcPool<T>: deduplicate votes by signer address                                                                                                                               
- XdcPoolTests: add unit tests   

#### What types of changes does your code introduce?

- [x] Bugfix (a non-breaking change that fixes an issue)
- [ ] New feature (a non-breaking change that adds functionality)
- [ ] Breaking change (a change that causes existing functionality not to work as expected)
- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

#### If yes, did you write tests?

- [x] Yes
- [ ] No

PR #11379, opened 2026-04-28
XDC: Fix duplicate signature quorum bypass

alexb5dh:
Fixes required-number-of-signatures check bypass via different signatures from the same signer (via changing ECDSA nonce).

## Changes
- Adds shared `VotesManager.CountValidSignatures` that also ensures no same-address signatures along with old checks.
- Reuses it in `QuorumCertificateManager` and `TimeoutCertificateManager` validations.
- ~~Refactors `VotesManager.GetValidSignatures` with similar fix.~~ Already fixed in https://github.com/NethermindEth/nethermind/pull/11072.
- Adds related tests
- Small code cleanup

#### Requires testing

- [x] Yes

- [x] Yes


alexb5dh:
@claude re-review

batrr (approved):

batrr:
We accept QC/TC when there are enough unique votes. However, the XDC Go client rejects it when any duplicates are present. As a result, we may accept a block, while the Go client will reject it.

batrr:
@claude I might be mistaken  Recheck please. Note we don't have duplicates in the pool.

alexb5dh:
So to summarize:
---
- `VotesManager.GetValidSignatures` change is actually not needed - passed `roundVotes` are already deduplicated as they are taken from `Dictionary<(ulong Round, Hash256 Hash), Dictionary<Address, T>> _items` in `XdcPool`. This is what was addressed in [previous PR](https://github.com/NethermindEth/nethermind/pull/11072). **Planning to revert my `GetValidSignatures` change.**
- `TimeoutCertificateManager`/`QuorumCertificateManager` signer-deduplication fixes are still needed as any peer can submit a certificate.
---
- Go client **does** similar deduplication for TC/QC validation.
- Go client **does not** have this validation at vote-accepting stage. This **should not** cause consensus split, as both nodes will reject TC/QC with duplicate signers. So this issue should not lead to consensus split or chain stalling.
---
@claude confirm/reject points below. Use `dev-upgrade` branch https://github.com/XinFinOrg/XDPoSChain if you can clone a repo to check Go client code.

cicr99 (approved):

ak88 (approved):

PR #10038, opened 2025-12-25
Add opcode tracing plugin

dmitriy-b:
## Changes

- Add OpcodeTracing plugin that counts EVM opcode usage across blocks
- The plugin supports two tracing modes:
  - Retrospective: reads historical blocks from database and outputs a single summary file
  - RealTime: traces blocks as they're processed, writes per-block files and a cumulative file
- Plugin gracefully disables itself on invalid config without crashing the client. The plugin is disabled by default.
- Configurable via CLI flags: `--OpcodeTracing.Enabled`, `--OpcodeTracing.Mode`, `--OpcodeTracing.Blocks`, `--OpcodeTracing.StartBlock`, `--OpcodeTracing.EndBlock`, `--OpcodeTracing.OutputDirectory`, `--OpcodeTracing. MaxDegreeOfParallelism` (for retrospective mode only).

#### What types of changes does your code introduce?

- [ ] Bugfix (a non-breaking change that fixes an issue)
- [x] New feature (a non-breaking change that adds functionality)
- [ ] Breaking change (a change that causes existing functionality not to work as expected)
- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

#### If yes, did you write tests?

- [ ] Yes
- [x] No

#### Notes on testing

Tested manually via Docker with Hoodi testnet:
- Plugin disabled by default when not specified
- Retrospective mode generates correct JSON output
- RealTime mode generates per-block and cumulative JSON files
- Invalid config logs error and disables plugin without crashing

See attached logs json files.

[opcode-trace-block-1845471.json](https://github.com/user-attachments/files/24338695/opcode-trace-block-1845471.json)
[opcode-trace-block-1845470.json](https://github.com/user-attachments/files/24338696/opcode-trace-block-1845470.json)
[opcode-trace-all-20251224143649.json](https://github.com/user-attachments/files/24338697/opcode-trace-all-20251224143649.json)
[opcode-trace-1844925-1844934.json](https://github.com/user-attachments/files/24338698/opcode-trace-1844925-1844934.json)
[05-validation-error.log](https://github.com/user-attachments/files/24338699/05-validation-error.log)
[04-realtime-mode.log](https://github.com/user-attachments/files/24338700/04-realtime-mode.log)
[03-retrospective-mode.log](https://github.com/user-attachments/files/24338701/03-retrospective-mode.log)
[...]

PR #11386, opened 2026-04-28
fix: remove broken backlog warning in AsyncFileWriteQueue

kamilchodola:
## Changes

`UnboundedChannel<T>` does not support `ChannelReader.Count` — it throws `NotSupportedException`. The backlog warning added in #10038 (commit 63b5a80b by @LukaszRozmej) called `Reader.Count` inside `Enqueue()` on every block, crashing block processing at runtime.

Reverts `Enqueue` to the simple `TryWrite` it was before the review remark change, and removes the now-unused `BacklogWarnThreshold` constant and `_backlogWarned` field.

## Type of change

## Testing

- Verified the exception stack trace matches the removed code path
- Build passes


LukaszRozmej (approved):

benaadams (approved):
Tests to prevent regresion?

stdevMac (approved):

src/Nethermind/Nethermind.OpcodeTracing.Plugin.Test/AsyncFileWriteQueueTests.cs
```diff
@@ -0,0 +1,83 @@
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
```
alexb5dh, line 1:
```suggestion
// SPDX-FileCopyrightText: 2026 Demerzel Solutions Limited
```

alexb5dh (approved):
From Claude, not sure if exlusion is intended:
```
 🟠 High — Test project not wired into CI or the solution. Nethermind.OpcodeTracing.Plugin.Test.csproj is missing from both Nethermind.slnx and the test matrix in
  nethermind-tests.yml. CI uses an explicit matrix (not glob discovery), so the tests are dead code from CI's perspective. Needs two changes:
  1. Add the project to Nethermind.slnx inside the existing /Plugins/OpcodeTracing/ folder
  2. Add Nethermind.OpcodeTracing.Plugin.Test to the project matrix in .github/workflows/nethermind-tests.yml
```

kamilchodola:
@alexb5dh yeah for now I kept tests only withing plugin as none of our plugins having tests wired into CI - @dmitriy-b working on followup startegy for testing plugins well

PR #11394, opened 2026-04-29
fix(flat-state): wait for TrieWarmer in-flight jobs before disposing SnapshotBundle

asdacap:
## Changes

- Add `_outstandingWarmups` counter to `FlatWorldStateScope` (incremented after a successful `PushAddressJob`/`PushSlotJob`, decremented in `finally` inside `WarmUpStateTrie` and `WarmUpStorageTrie`)
- `FlatWorldStateScope.Dispose` spin-waits until the counter reaches zero before calling `_snapshotBundle.Dispose()`, emitting a `Warn` log if the wait exceeds 1 second
- `ITrieWarmer.PushAddressJob` / `PushSlotJob` now return `bool` (true if enqueued) so callers can conditionally increment
- `TrieWarmer` and `NoopTrieWarmer` updated to match the new signatures

## Types of changes

#### Requires testing

- [x] Yes

- [x] Yes

#### Notes on testing

Two new tests in `FlatWorldStateScopeProviderTests`:
- `Dispose_WaitsForOutstandingWarmups_BeforeDisposingBundle` — simulates an in-flight warmup and verifies `Dispose` blocks until the counter is decremented
- `Dispose_CompletesImmediately_WhenNoOutstandingWarmups` — verifies no regression in the fast path

- [x] No

- [x] No

## Remarks

Root cause: `TrieWarmer` worker threads could still be mid-`HandleJob` reading from a `SnapshotBundle`/`ReadOnlySnapshotBundle`/RocksDB snapshot when `FlatWorldStateScope.Dispose` released it. The native RocksDB error that resulted ("rocksdb option disposed → timestamp not supported") was not one of the managed exception types caught in `HandleJob`, so it escaped. A managed dispose-check is insufficient because a worker can pass any check and then have the snapshot disposed before the native call returns — the fix uses a counter to guarantee no worker is inside the read path before the snapshot is released.

### DIFF
--- a/src/Nethermind/Nethermind.State.Flat.Test/FlatWorldStateScopeProviderTests.cs
+++ b/src/Nethermind/Nethermind.State.Flat.Test/FlatWorldStateScopeProviderTests.cs
@@ -814,8 +814,6 @@ public class FlatWorldStateScopeProviderTests
 
     #endregion
 
-    #region TrieWarmer dispose-safety tests
-
     [Test]
     public async Task Dispose_WaitsForOutstandingWarmups_BeforeDisposingBundle()
     {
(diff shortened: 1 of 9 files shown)


batrr (approved):

LukaszRozmej:
@claude re-review

LukaszRozmej (approved):

PR #11388, opened 2026-04-28
Engine api glamsterdam cleanup

MarekM25:
## Changes

[- _List the changes_](https://github.com/ethereum/execution-apis/pull/786/changes)

#### What types of changes does your code introduce?

- [ ] Bugfix (a non-breaking change that fixes an issue)
- [ ] New feature (a non-breaking change that adds functionality)
- [ ] Breaking change (a change that causes existing functionality not to work as expected)
- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [ ] No

#### If yes, did you write tests?

- [ ] Yes
- [ ] No

#### Notes on testing

_Optional. Remove if not applicable._

#### Requires documentation update

- [ ] Yes
- [ ] No

_If yes, link the PR to the docs update or the issue with the details labeled `docs`. Remove if not applicable._

#### Requires explanation in Release Notes

- [ ] Yes
- [ ] No

_If yes, fill in the details here. Remove if not applicable._

## Remarks

_Optional. Remove if not applicable._

PR #11396, opened 2026-04-29
Reorg up to lastFinalized block

svlachakis:
Implements **-38006 Too deep reorg** execution-apis: https://github.com/ethereum/execution-apis/pull/786, point 6 on top of **#11388** (point 2).

## Changes

- **`ForkchoiceUpdatedHandler.ShouldProceedWithReorg`**  
  Returns `-38006` when `IWorldStateManager.CanReorgOn(header)` is `false`.  
  All `pruning-trie`, `flat-DB`, `archive` apply their own policy via `HasStateForBlock`.

- **`ReorgDepthFinalizedStateProvider.FinalizedBlockNumber`**  
  Returns `_blockTree.FinalizedHash`'s block number instead of `BestKnownNumber - 64`.  
  Trie dirty-cache eviction now retains the `[finalized, head]` range.

- **`LogIndexStorage.GetCompressionBoundary`**  
  Compression cutoff is now `finalized`, keeping log data above finalized reorg-safe.

- **`MergeErrorCodes.TooDeepReorg`**  
  Added `= -38006`.

- **`TaikoForkchoiceUpdatedHandler`**  
  Overrides `ShouldProceedWithReorg` to always proceed  
  (Taiko finality follows L1).

All three gates now read the same boundary:

`_blockTree.FinalizedHash`

### Bug fix (atomicity)

Reorder `RejectIfInconsistent` to run before `UpdateMainChain` , so an inconsistent FCU returns -38002 without partially mutating the head. Aligns with `paris.md` step 7 ("all updates to the forkchoice state resulting from this call MUST be made atomically"). 

The previous ordering let the head move before the inconsistency check returned the error - unreachable in practice (well-behaved CLs never send inconsistent tuples), but spec-noncompliant.

**Test pinned:**
`forkchoiceUpdatedV1_WhenInconsistent_ReturnsInvalidForkchoiceStateBeforeProceedingToReorgCheck`

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain/ReorgDepthFinalizedStateProvider.cs
+++ b/src/Nethermind/Nethermind.Blockchain/ReorgDepthFinalizedStateProvider.cs
@@ -10,6 +10,9 @@ namespace Nethermind.Blockchain;
 
 public class ReorgDepthFinalizedStateProvider(IBlockTree blockTree) : IFinalizedStateProvider
 {
+    private sealed record FinalizedCache(Hash256 Hash, long Number);
+    private FinalizedCache? _cache;
+
     public long FinalizedBlockNumber
     {
         get
@@ -17,8 +20,15 @@ public class ReorgDepthFinalizedStateProvider(IBlockTree blockTree) : IFinalized
             Hash256? finalizedHash = blockTree.FinalizedHash;
             if (finalizedHash is not null && finalizedHash != Keccak.Zero)
             {
+                FinalizedCache? cache = _cache;
+                if (cache is not null && cache.Hash == finalizedHash) return cache.Number;
+
                 BlockHeader? finalizedHeader = blockTree.FindHeader(finalizedHash, BlockTreeLookupOptions.DoNotCreateLevelIfMissing);
-                if (finalizedHeader is not null) return finalizedHeader.Number;
+                if (finalizedHeader is not null)
+                {
+                    _cache = new FinalizedCache(finalizedHash, finalizedHeader.Number);
+                    return finalizedHeader.Number;
+                }
             }
             return Math.Max(0, blockTree.BestKnownNumber - Reorganization.MaxDepth);
         }
--- a/src/Nethermind/Nethermind.Blockchain/ReorgDepthFinalizedStateProvider.cs
+++ b/src/Nethermind/Nethermind.Blockchain/ReorgDepthFinalizedStateProvider.cs
@@ -10,7 +10,19 @@ namespace Nethermind.Blockchain;
 
 public class ReorgDepthFinalizedStateProvider(IBlockTree blockTree) : IFinalizedStateProvider
 {
-    public long FinalizedBlockNumber => Math.Max(0, blockTree.BestKnownNumber - Reorganization.MaxDepth);
+    public long FinalizedBlockNumber
+    {
+        get
+        {
+            Hash256? finalizedHash = blockTree.FinalizedHash;
+            if (finalizedHash is not null && finalizedHash != Keccak.Zero)
+            {
+                BlockHeader? finalizedHeader = blockTree.FindHeader(finalizedHash, BlockTreeLookupOptions.DoNotCreateLevelIfMissing);
+                if (finalizedHeader is not null) return finalizedHeader.Number;
+            }
+            return Math.Max(0, blockTree.BestKnownNumber - Reorganization.MaxDepth);
+        }
+    }
 
     public Hash256? GetFinalizedStateRootAt(long blockNumber)
     {
--- a/src/Nethermind/Nethermind.Blockchain/ReorgDepthFinalizedStateProvider.cs
+++ b/src/Nethermind/Nethermind.Blockchain/ReorgDepthFinalizedStateProvider.cs
@@ -11,7 +11,7 @@ namespace Nethermind.Blockchain;
 public class ReorgDepthFinalizedStateProvider(IBlockTree blockTree) : IFinalizedStateProvider
 {
     private sealed record FinalizedCache(Hash256 Hash, long Number);
-    private FinalizedCache? _cache;
+    private volatile FinalizedCache? _cache;
 
     public long FinalizedBlockNumber
     {
--- a/src/Nethermind/Nethermind.Db/LogIndex/IFinalizedBlockProvider.cs
+++ b/src/Nethermind/Nethermind.Db/LogIndex/IFinalizedBlockProvider.cs
@@ -1,0 +1,13 @@
+// SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+namespace Nethermind.Db.LogIndex;
+
+/// <summary>
+/// Provides the current finalized block number so that subsystems can determine their reorg-safe window dynamically.
+/// </summary>
+public interface IFinalizedBlockProvider
+{
+    /// <summary>Returns the current finalized block number, or null if finalization has not yet been established.</summary>
+    long? FinalizedBlockNumber { get; }
+}
--- a/src/Nethermind/Nethermind.Db/LogIndex/LogIndexStorage.cs
+++ b/src/Nethermind/Nethermind.Db/LogIndex/LogIndexStorage.cs
@@ -518,7 +518,7 @@ namespace Nethermind.Db.LogIndex
         {
             if (_maxBlock is null) return null;
             if (_finalizedBlockProvider.FinalizedBlockNumber is { } finalized)
-                return checked((int)finalized);
+                return finalized > int.MaxValue ? _maxBlock : (int)finalized;
             return _maxBlock - _maxReorgDepth;
         }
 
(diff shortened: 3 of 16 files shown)


svlachakis:
@claude review again

svlachakis:
@claude review again

svlachakis:
@claude test seems correct to me, are you sure this is not just your review oversight?

The test mocks IsMainChain(Arg.Any<BlockHeader>()).Returns(true) (line 146), so the override takes the early return path: it calls FindBlock(newHeadHeader.Hash!, ...) (mocked at line 141 → returns deepAncestor), packages [deepAncestor], and returns true. Base TryGetBranch is never called, FindParent is never called, no parent mock is needed. 

Am I missing something?

asdacap (approved):

src/Nethermind/Nethermind.Taiko/Rpc/TaikoForkchoiceUpdatedHandler.cs
```diff
@@ -50,13 +52,14 @@ internal class TaikoForkchoiceUpdatedHandler(
     specProvider,
     syncPeerPool,
     mergeConfig,
+    worldStateManager,
     logManager)
 {
-    protected override bool IsOnMainChainBehindFinalized(BlockHeader newHeadHeader, ForkchoiceStateV1 forkchoiceState,
-        [NotNullWhen(true)] out ResultWrapper<ForkchoiceUpdatedV1Result>? result)
+    protected override bool ShouldProceedWithReorg(BlockHeader newHeadHeader, ForkchoiceStateV1 forkchoiceState,
```
Marchhill, line 58: Why change the name?

svlachakis: Isn't `ShouldProceedWithReorg` more clear and understandable?

src/Nethermind/Nethermind.Blockchain/ReorgDepthFinalizedStateProvider.cs
```diff
@@ -10,7 +10,29 @@ namespace Nethermind.Blockchain;
 
 public class ReorgDepthFinalizedStateProvider(IBlockTree blockTree) : IFinalizedStateProvider
```
LukaszRozmej, line 11: Out of scope of this PR, but we already have something like `IFinalizationManager`. Would be good consider a refactor to unify those in a way?

src/Nethermind/Nethermind.Blockchain/ReorgDepthFinalizedStateProvider.cs
```diff
@@ -10,7 +10,29 @@ namespace Nethermind.Blockchain;
...
+            Hash256? finalizedHash = blockTree.FinalizedHash;
+            if (finalizedHash is not null && finalizedHash != Keccak.Zero)
+            {
+                FinalizedCache? cache = _cache;
+                if (cache is not null && cache.Hash == finalizedHash) return cache.Number;
+
+                BlockHeader? finalizedHeader = blockTree.FindHeader(finalizedHash, BlockTreeLookupOptions.DoNotCreateLevelIfMissing);
+                if (finalizedHeader is not null)
+                {
+                    _cache = new FinalizedCache(finalizedHash, finalizedHeader.Number);
```
LukaszRozmej, line 29: Does this need to be thread-safe?

src/Nethermind/Nethermind.Db/LogIndex/IFinalizedBlockProvider.cs
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Nethermind.Db.LogIndex;
+
+/// <summary>
+/// Provides the current finalized block number so that subsystems can determine their reorg-safe window dynamically.
+/// </summary>
+public interface IFinalizedBlockProvider
+{
+    /// <summary>Returns the current finalized block number, or null if finalization has not yet been established.</summary>
+    long? FinalizedBlockNumber { get; }
```
LukaszRozmej, line 12: Same as `IFinalizedStateProvider`? Maybe something should live in `Nethermind.Core`?

src/Nethermind/Nethermind.Db/LogIndex/LogIndexStorage.cs
```diff
@@ -512,7 +514,13 @@ private static int SaveRangeBound(IWriteOnlyKeyValueStore dbBatch, byte[] key, i
...
         }
 
-        private int? GetLastReorgableBlockNumber() => _maxBlock - _maxReorgDepth;
+        private int? GetCompressionBoundary()
+        {
+            if (_maxBlock is null) return null;
+            if (_finalizedBlockProvider.FinalizedBlockNumber is { } finalized)
+                return finalized > int.MaxValue ? _maxBlock : (int)finalized;
+            return _maxBlock - _maxReorgDepth;
+        }
```
LukaszRozmej:
```suggestion
private int? GetCompressionBoundary() => _maxBlock is null
        ? null
        : _finalizedBlockProvider.FinalizedBlockNumber switch
        {
            null => _maxBlock - _maxReorgDepth,
            > int.MaxValue => _maxBlock,
            var finalized => (int)finalized
        };
```

src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
```diff
@@ -239,7 +250,11 @@ protected virtual bool IsOnMainChainBehindFinalized(BlockHeader newHeadHeader, F
             return ForkchoiceUpdatedV1Result.Error(setHeadErrorMsg, ErrorCodes.InvalidParams);
         }
 
-        if (IsOnMainChainBehindFinalized(newHeadHeader, forkchoiceState, out ResultWrapper<ForkchoiceUpdatedV1Result>? result))
+        long finalizedNumber = finalizedHeader?.Number ?? 0;
+        if (RejectIfInconsistent(finalizedHeader, 0, "finalized", newHeadHeader, requestStr) is { } finalizedError) return finalizedError;
+        if (RejectIfInconsistent(safeBlockHeader, finalizedNumber, "safe", newHeadHeader, requestStr) is { } safeError) return safeError;
+
+        if (blocks is not null && !ShouldProceedWithReorg(newHeadHeader, forkchoiceState, out ResultWrapper<ForkchoiceUpdatedV1Result>? result))
```
LukaszRozmej, line 257: Maybe do it as above?

```suggestion
        if (blocks is not null && !RejectIfCantProceedWithReorg(newHeadHeader, forkchoiceState) is reorgError) return reorgError;
```

LukaszRozmej (approved):

LukaszRozmej:
will push small commit with tests improvements

src/Nethermind/Nethermind.Blockchain/ReorgDepthFinalizedStateProvider.cs
svlachakis: 100% will create an issue for it to tackle next

src/Nethermind/Nethermind.Db/LogIndex/IFinalizedBlockProvider.cs
svlachakis: Lives in `Nethermind.Db` because `LogIndexStorage` is in that assembly and I wanted no dependency on `Nethermind.Blockchain`. Agreed `Nethermind.Core `would be a better place - folding into the unification follow-up from comment above.

src/Nethermind/Nethermind.Merge.Plugin/Handlers/ForkchoiceUpdatedHandler.cs
svlachakis: Thought about it, but the override semantics in Taiko I think make the bool+out signature cleaner - the method does double duty (skip-true, error, or proceed). Renaming to a single-output reject helper would lose that but of course can change it if you think it's better.

src/Nethermind/Nethermind.Db/LogIndex/LogIndexStorage.cs
svlachakis: done

src/Nethermind/Nethermind.Blockchain/ReorgDepthFinalizedStateProvider.cs
svlachakis: I made the field volatile, so the publish has store-release and the read has load-acquire on every supported architecture. The cached object is an immutable record - finalized advancing creates a new `FinalizedCache` and swaps the reference, never mutates an existing one. So the only race a reader can hit is "old ref vs new ref," and the reference write is atomic. Worst case is one redundant `FindHeader` call per racing reader. I think no further locking is needed.

svlachakis: @claude confirm this is correct and we won't have race issues here

I made the field volatile, so the publish has store-release and the read has load-acquire on every supported architecture. The cached object is an immutable record - finalized advancing creates a new `FinalizedCache` and swaps the reference, never mutates an existing one. So the only race a reader can hit is "old ref vs new ref," and the reference write is atomic. Worst case is one redundant `FindHeader` call per racing reader. I think no further locking is needed.

svlachakis:
@claude review again and confirm this conversation about thread safety in FinalizedCache

benaadams (approved):

PR #11398, opened 2026-04-29
fix(parallel): capture and rethrow worker exceptions in ParallelUnbalancedWork

benaadams:
## Changes

- Worker exceptions were escaping onto thread-pool threads (unobserved/fatal under default settings) and could leave the calling thread waiting on the semaphore.

- Capture the first exception via ExceptionDispatchInfo, let remaining workersdrain, then rethrow on the caller with the original stack trace preserved.

#### Requires testing

- [x] Yes

- [x] Yes

### DIFF
--- a/src/Nethermind/Nethermind.Core/Threading/ParallelUnbalancedWork.cs
+++ b/src/Nethermind/Nethermind.Core/Threading/ParallelUnbalancedWork.cs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 
 using System;
+using System.Runtime.ExceptionServices;
 using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
@@ -50,6 +51,9 @@ public class ParallelUnbalancedWork : IThreadPoolWorkItem
         {
             data.Event.Wait();
         }
+
+        // Rethrow the first captured worker exception, if any, on the calling thread
+        data.ThrowIfFaulted();
 
         parallelOptions.CancellationToken.ThrowIfCancellationRequested();
     }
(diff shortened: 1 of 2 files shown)


benaadams:
@claude re-review

kamilchodola (approved):

Marchhill (approved):

src/Nethermind/Nethermind.Core/Threading/ParallelUnbalancedWork.cs
```diff
@@ -193,6 +207,30 @@ private class BaseData(int threads, int fromInclusive, int toExclusive, Cancella
...
+        /// Captures the first exception observed by any worker so it can be rethrown on the
+        /// calling thread. Subsequent exceptions are dropped.
+        /// </summary>
+        public void CaptureException(Exception exception)
+        {
+            // Skip the (non-trivial) ExceptionDispatchInfo.Capture call once we already have one.
+            if (IsFaulted) return;
+
+            // Only the first exception wins; any later ones are discarded.
+            Interlocked.CompareExchange(ref _exception, ExceptionDispatchInfo.Capture(exception), null);
```
LukaszRozmej, line 226: I think this eagerly calls `ExceptionDispatchInfo.Capture(exception)`. Is it costly?
If it is I would suggest:


```suggestion
if (Volatile.Read(ref _exception) is null)
            Interlocked.CompareExchange(ref _exception, ExceptionDispatchInfo.Capture(exception), null);
```

benaadams: IsFaulted above already does that (using Volitile.Read)

PR #11399, opened 2026-04-29
Feat/evm enginetest

kamilchodola:
### Reusable nethtest workflow replacing the pyspec NUnit shards

- `run-nethtest.yml` is a **reusable workflow** (`workflow_call` + `workflow_dispatch`): callers pick `test_types` (engineTest, blockTest, stateTest, syncTest, txTest) and a `setup_group` (sequential / flat / all); a prepare job generates the chunked job matrix.
- **`nethermind-tests.yml`** replaces its 16 `Ethereum.Blockchain.Pyspec.Test` Sequential chunks with one template call (sequential/patricia group: all five EEST fixture formats + zkEVM stateless preview + the Amsterdam variant configs).
- **`nethermind-tests-flat.yml`** replaces its 112 pyspec shards (16 flat Sequential + 96 Amsterdam parallel/batch-read variant chunks) with one template call (flat group — 58 jobs).
- Net effect per CI run: **128 pyspec shard jobs (~777 job-minutes) → 81 nethtest jobs (~250-350 job-minutes)**, all on `ubuntu-latest`. Each NUnit shard parsed the entire fixture subtree and discarded 15/16 of it; nethtest chunks at the file level (`--chunk NofM`), so each job parses only its share.
- The EEST version defaults to **`latest-bal`**: every run resolves the newest BAL fixture release so new upstream tests surface immediately (client-vs-spec drift then fails CI by design); pass an exact tag via `eest_version` to pin. The next-fork (BAL) release ships fixtures for all forks, so the former `current_spec` run against `execution-spec-tests` stable added no coverage and was dropped.
- OOM mitigation on 16 GB runners: workstation GC (`DOTNET_gcServer=0`) + file chunking. Parallel-execution variants run single-worker in 8 chunks because process memory grows with the number of parallel-executed tests (~14 GB after ~5k tests even at 2 workers) — the memory growth previously chased on `feat/evm-enginetest-dotmemory-20gb`.

### Nethermind.Test.Runner

- Engine-test mode, JSON output, worker parallelism, progress prefixes, preserved tracing behavior.
- Pyspec-parity configuration flags: `--chunk NofM` (deterministic interleaved file chunks), `--flatdb` (flat state layout), `--parallelExecution true|false` and `--batchRead true|false` (BAL parallel-execution / batch-read prewarm toggles, mirroring the `AmsterdamParallel`/`BatchRead`/`ParallelFull` pyspec fixtures).
- New `--txTest` mode runs EEST `transaction_tests` fixtures (raw tx decode + `TxValidator` vs expected exception); sync fixtures (`blockchain_tests_sync`) run through the engine harness, matching the NUnit sync coverage.
- New `--zkevmTest` mode runs the zkEVM stateless preview fixtures (witness bytes through `StatelessExecutor`, byte-compared with the expected output, mirroring the `ZkEvmPreview` NUnit fixture). The workflow resolves the `tests-zkevm` release line per run (`latest-zkevm`, pinnable via `zkevm_version`), so these update automatically like the EEST fixtures.

### Test infrastructure

- `TestChunkFilter` gains an explicit-spec `FilterByChunk(tests, "NofM")` overload (the env-driven behavior is unchanged).
- `BlockchainTestBase`/`GeneralStateTestBase` honor `TEST_USE_FLAT=1` again by setting `IFlatDbConfig.Enabled` (mirroring `TestBlockchain.UseFlatDb`); the consumption was lost in #11911, which made the flat-workflow fixture shards silently run patricia.
- Review feedback: `SuggestBlocks` → `Result<BlockHeader>`, `RunNewPayloads` → `Task<Result<string>>`.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [x] Build-related changes
- [x] Other: Test infrastructure

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

- `TestChunkFilter` explicit-chunk overload covered by new `TestChunkFilterTests` cases (25 tests pass post-merge).
- Built the runner with `dotnet build src\Nethermind\Nethermind.Test.Runner\Nethermind.Test.Runner.csproj -c Release`; smoke-tested `--chunk`, `--flatdb`, `--parallelExecution`, `--batchRead`, `--txTest` locally.
- Validated on GitHub Actions: chunked ubuntu-latest matrix runs 27298750200 / 27300384186 / 27300914505 (engine set, 22.2k tests), plus the PR-triggered caller runs exercising the reusable-workflow integration.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


kamilchodola:
@claude re-review

kamilchodola:
@claude re-review

kamilchodola:
@claude re-review

kamilchodola:
@claude re-review

kamilchodola:
@claude re-review

benaadams (approved):

kamilchodola:
@claude seems like new version of "new_spec" bal tests has dropped - is there a way to somehow automatically choose new spec? based for example on the key phrase like "bal"?

src/Nethermind/Ethereum.Test.Base/BlockchainTestBase.cs
```diff
@@ -264,7 +281,16 @@ protected async Task<EthereumTestResult> RunTest(BlockchainTest test, Stopwatch?
...
+            EthereumTestResult result = new(test.Name, test.ForkName, testPassed);
+            if (headBlock?.Hash is not null)
+                result.LastBlockHash = headBlock.Hash;
+            if (!string.IsNullOrEmpty(lastPayloadStatus))
+                result.LastPayloadStatus = lastPayloadStatus;
+            if (lastValidationError is not null && result.Pass)
+                result.Error = lastValidationError;
+            if (!testPassed)
+                result.Error = string.Join("; ", differences);
+            return result;
```
LukaszRozmej:
```suggestion
return new EthereumTestResult(test.Name, test.ForkName, testPassed)
{
    LastBlockHash = headBlock?.Hash,
    LastPayloadStatus = string.IsNullOrEmpty(lastPayloadStatus) ? null : lastPayloadStatus,
    Error = !testPassed
        ? string.Join("; ", differences)
        : lastValidationError is not null
            ? lastValidationError
            : null
};
```

src/Nethermind/Ethereum.Test.Base/BlockchainTestBase.cs
```diff
@@ -273,8 +299,9 @@ protected async Task<EthereumTestResult> RunTest(BlockchainTest test, Stopwatch?
         }
     }
 
-    private static BlockHeader SuggestBlocks(BlockchainTest test, bool failOnInvalidRlp, IBlockValidator blockValidator, IBlockTree blockTree, BlockHeader parentHeader)
+    private static (BlockHeader header, string? lastBlockError) SuggestBlocks(BlockchainTest test, bool failOnInvalidRlp, IBlockValidator blockValidator, IBlockTree blockTree, BlockHeader parentHeader)
```
LukaszRozmej: Can it be `Result<BlockHeader>`?

src/Nethermind/Ethereum.Test.Base/BlockchainTestBase.cs
```diff
@@ -297,54 +324,65 @@ private static BlockHeader SuggestBlocks(BlockchainTest test, bool failOnInvalid
...
+            System.Reflection.MethodInfo method = typeof(IEngineRpcModule).GetMethod($"engine_newPayloadV{version}")
+                ?? throw new NotSupportedException($"engine_newPayloadV{version} not found on IEngineRpcModule. Update {nameof(EngineApiVersions.NewPayload.Latest)}.");
+            result[version] = method.GetParameters().Length;
+        }
+
+        return result;
+    }
 
-    private async static Task RunNewPayloads(TestEngineNewPayloadsJson[]? newPayloads, IJsonRpcService rpcService, JsonRpcContext rpcContext, Hash256 initialHeadHash)
+    private async static Task<(string status, string? validationError)> RunNewPayloads(TestEngineNewPayloadsJson[]? newPayloads, IJsonRpcService rpcService, JsonRpcContext rpcContext, Hash256 initialHeadHash)
```
LukaszRozmej:
```suggestion
    private async static Task<Result<string>> RunNewPayloads(TestEngineNewPayloadsJson[]? newPayloads, IJsonRpcService rpcService, JsonRpcContext rpcContext, Hash256 initialHeadHash)
```

src/Nethermind/Ethereum.Test.Base/GeneralStateTestBase.cs
```diff
@@ -185,10 +185,18 @@ protected EthereumTestResult RunTest(GeneralStateTest test, ITxTracer txTracer)
             }
 
             List<string> differences = RunAssertions(test, stateProvider);
+            // Capture tx error for exception mapping (even when test passes)
+            string txError = "";
+            if (txResult is not null && txResult.Value != TransactionResult.Ok)
+                txError = txResult.Value.ErrorDescription;
+            else if (blockValidationError is not null)
+                txError = blockValidationError;
```
LukaszRozmej:
```suggestion
string txError = txResult is { Value: { ResultType: not TransactionResult.Ok } value }
        ? value.ErrorDescription
        : blockValidationError ?? "";
```

src/Nethermind/Nethermind.Test.Runner/BlockchainTestsRunner.cs
```diff
@@ -6,65 +6,163 @@
...
+    public BlockchainTestsRunner(
+        ITestSourceLoader testsSource,
+        string? filter,
+        ulong chainId,
+        bool trace = false,
+        bool traceMemory = false,
+        bool traceNoStack = false,
+        bool jsonOutput = false,
+        bool suppressOutput = false,
+        Func<bool, string?>? progressUpdateFactory = null)
```
LukaszRozmej: Maybe introduce helper `Options` class/struct?

src/Nethermind/Nethermind.Test.Runner/BlockchainTestsRunner.cs
```diff
@@ -6,65 +6,163 @@
...
+        this.chainId = chainId;
+        this.trace = trace;
+        this.traceMemory = traceMemory;
+        this.traceNoStack = traceNoStack;
+        this.jsonOutput = jsonOutput;
+        this.suppressOutput = suppressOutput;
+        this.progressUpdateFactory = progressUpdateFactory;
+    }
 
     public async Task<IEnumerable<EthereumTestResult>> RunTestsAsync()
```
LukaszRozmej, line 51: method is getting to long, break up to logical parts

LukaszRozmej:
@claude do a review focusing on style and architecture ect.

LukaszRozmej:
## Review

10 files, +1095/-148. Adds `run-nethtest.yml` workflow + extends `Nethermind.Test.Runner` with engine-test mode, `--jsonout`, `--workers`, richer `EthereumTestResult`, and streaming PROGRESS/FAIL/EXCEPTION on stderr.

### High

**H1. `BlockchainTest.ForkName` is added but never populated.** `BlockchainTest.cs:14` adds `public string? ForkName { get; set; }`, and `BlockchainTestBase.RunTest` now writes `new EthereumTestResult(test.Name, test.ForkName, …)` at multiple sites. But `JsonToEthereumTest.ConvertToBlockchainTests` (`JsonToEthereumTest.cs:300-313`) never assigns it. `EthereumTestResult` defaults `Fork = fork ?? "unknown"`, so every blockchain/engine result in the JSON output will report `"fork": "unknown"`. The fork name is parsed locally as `networks[0]` and is available — set `ForkName = networks[0]` there (or derive `test.Network?.Name` inside `RunTest`).

### Medium

**M1. `EthereumTestResult.StateRoot` default silently removed.** Was `Hash256 StateRoot { get; set; } = Keccak.EmptyTreeHash;`, now `Hash256? StateRoot { get; set; }` with no initializer and `[JsonIgnore(WhenWritingNull)]`. `BlockchainTestBase.RunTest` no longer sets it on the returned result, so blockchain test results that previously surfaced `EmptyTreeHash` will surface `null`. State tests are unaffected. Either keep setting it on the blockchain path (`headBlock.StateRoot`) or call this out in release notes — external consumers that dereference `result.StateRoot` on a blockchain result will NRE.

**M2. `BlockchainTestsRunner` two-constructor design is fragile.** Primary takes `(in options, source = null)` and `RunTestsAsync()` throws `InvalidOperationException` if source is null — i.e. wrong constructor choice fails at runtime instead of compile time. Either keep a single non-null-source constructor and expose `RunSingleTestAsync` via a factory method, or split into two types.

**M3. `_filterRegex` recompiled per runner instance.** Each `BlockchainTestsRunner` does `new Regex(..., RegexOptions.Compiled)`. The parallel path in `Program.RunBlockTestFiles` creates one runner per file → the same pattern is JIT-compiled N times for N fixture files. Hoist to a static cache keyed by pattern string, or compile once in `Program` and pass the `Regex` in via options.

**M4. `RunStateTestFiles` constructs `StateTestsRunner` with a `\"dummy\"` source.** `new TestsSourceLoader(new LoadGeneralStateTestFileStrategy(), \"dummy\")` is passed only to satisfy the constructor; `RunSingleTest` ignores the source. Refactor so single-test execution doesn't drag a fake `ITestSourceLoader` along.

**M5. `BuildStackJson` writes strings without JSON escaping.** `StateTestRunner.cs:115` emits `'\"' + stack[i] + '\"'`, and the surrounding trace line concatenates `entry.Operation` unquoted-escaped. Today these are EIP-3155 hex literals / opcode mnemonics so the bug is latent, but a single non-ASCII or quote-containing value yields malformed JSON. Use `JsonEncodedText.Encode` or assert the invariant in a comment. Same site uses `System.Text.Json.JsonSerializer.Serialize(entry.Error)` for a single-string escape — `JsonEncodedText.Encode` is the cheaper primitive.

**M6. `BlockchainTestStreamingTracer` lifetime changed.** Previously created once per runner before the test loop; now created per test inside `ExecuteTestAsync`. The tracer keeps counters `_transactionCount`, `_blockCount`, `_totalGasUsed`. If these were meant to accumulate across all tests in the file, the new code resets them per test. Trace mode is forced single-worker so there's no concurrency issue, but the output shape changed — please confirm it's intentional.

### Low / nits

**L1.** `Program.UpdateProgress`: two threads can both pass the time-gate before either calls `Interlocked.Exchange`, producing duplicate \"time-based\" reports. Harmless — a `// best-effort` comment would help.

**L2.** `Console.Error.WriteLine` from multiple workers is not serialized; lines can interleave under contention. Single-line writes are largely atomic but not guaranteed.

**L3.** `Program.CollectFiles` filters `.meta` via `file.Contains(\"/.meta/\") || file.Contains(\"\\.meta\\\\\")`. Brittle; prefer path-segment splitting on `Path.DirectorySeparatorChar`.

**L4.** `nethtest` shell script at the repo root — repo convention is `scripts/` and `tools/`. Also bash-only (Windows needs WSL/git-bash).

**L5.** `--workers` short alias `-p` collides with `dotnet -p` (`--project`) muscle memory. `-w`/`-W` or no short alias would be safer.

**L6.** Workflow triggers on `pull_request_review: [submitted, dismissed]`. Every review event spawns a run; the `approval-state` gate short-circuits when there's no current approval, but you still get run-then-immediately-skip churn on dismissals/comments. Limit to `types: [submitted]`.

**L7.** `BlockException.SYSTEM_CONTRACT_CALL_FAILED` mapped to the same regex as `INVALID_BAL_MISSING_ACCOUNT` with a TODO-style comment. Worth filing a follow-up to disambiguate the underlying error messages rather than silently equating two failure modes.

**L8.** Error-assignment ordering in `RunTest`: `if (lastValidationError is not null && result.Pass) result.Error = lastValidationError;` followed by `if (!testPassed) result.Error = string.Join(\"; \", differences);` — branches don't intersect so the logic is correct, but a one-line comment explaining \"propagate the validation error on passes; overwrite with differences on failures\" would help readers.

**L9.** `int.TryParse` improvements in `RunNewPayloads` are a nice upgrade — the original `int.Parse` would throw without context.

**L10.** `OverridableReleaseSpec.Name => spec.Name`: confirmed via grep that no test or production code references the literal \"OverridableReleaseSpec\" string. Safe and arguably correct — JSON output of EEST runs becomes meaningful.

### Verdict

Functionally a useful CI addition; the runner refactor lands the right capabilities. **H1** (ForkName never populated — likely defeats the primary purpose of the new field) and **M1** (StateRoot semantic change without notice) are blockers; the rest can be follow-ups.

kamilchodola:
## Assessment: can `run-nethtest.yml` replace the pyspec NUnit shards in CI?

**Yes for fixture-execution coverage — same fixtures, same configs, ~4× fewer runner acquisitions and ~6-8× fewer job-minutes — with three gaps to close before fully retiring the NUnit shards.**

### Measured cost (same `tests-bal` fixture family, linux-x64, free runners)

| | pyspec NUnit today | nethtest (this PR) |
|---|---|---|
| Workflows involved | `nethermind-tests.yml` (Pyspec shards) + `nethermind-tests-flat.yml` | `run-nethtest.yml` |
| Runner jobs per run | **128** (16 sequential + 112 flat/variants) | **34** |
| Job-minutes per run | **~777** (135 + 642) | **124** (34 avg-3.7-min jobs) |
| Wall clock | ~17 min | 8 min |

Reference runs: pyspec `27285302259` + `27289705918` (master) vs nethtest `27298750200` / `27300914505` (this branch, tests-bal@v7.3.0, 22,286 engine tests).

### Why it is structurally cheaper
1. **File-level chunking** (`--chunk NofM`): every NUnit chunk job enumerates `TestCaseSource` for the whole fixture subtree — it parses *all* JSON and then discards (N−1)/N of the loaded tests, 16× per variant. nethtest partitions the *file list*, so each job parses only its share.
2. No per-test NUnit adapter overhead; in-process `--workers` parallelism; suppressed passing-test output.
3. One workflow expresses the whole config matrix (flat / parallel / batchread variants) that currently needs a second 140-job workflow.

### Config parity (task 4)
`--flatdb`, `--parallelExecution`, `--batchRead` now map 1:1 to the pyspec fixtures:
`Sequential`=(off,off), `AmsterdamParallel`=(on,off), `AmsterdamBatchRead`=(off,on), `AmsterdamParallelFull`=(on,on), each × flat via `--flatdb` — variants scoped to the Amsterdam subtree exactly like the NUnit classes (EIP-7928-gated toggles).

### Gaps before fully replacing pyspec NUnit
1. **Test types per trigger**: PR-triggered runs execute `engineTest` only; `blockTest` (RLP import path) and `stateTest` are dispatch-only today. Full parity needs all three (≈2.5× jobs — still well under pyspec).
2. **Missing fixture formats**: `blockchain_tests_sync` (sync-payload validation) and `transaction_tests` have no nethtest mode yet.
3. **Reporting**: NUnit integrates with test explorers; nethtest emits step summaries + JSON artifacts (better for bulk triage, different for IDE workflows).

### Findings surfaced by the validation runs
- **8 pre-existing failures vs tests-bal@v7.3.0** (config-independent — identical under sequential/parallel/batchread, flat/merkle): `test_create2check_fields_in_initcode[d3,d7]`, `test_bal_create_in_static_context` ×4, `test_static_call_create[d1]` (all `InvalidBlockLevelAccessList`), and `test_slotnum_distinct_per_block` (`InvalidSlotNumber`). These are client↔spec drift on the new release, not workflow artifacts.
- **BAL parallel-execution memory growth**: with `--parallelExecution true` the runner's memory grows with the number of tests executed (~14 GB after ~4-5k tests at 2-4 workers, workstation GC), while sequential/batchread complete 12.8k tests in a single 16 GB job. This is the crash previously chased on `feat/evm-enginetest-dotmemory-20gb`; CI now works around it (1 worker / 8 chunks for parallel variants) but it deserves a dotMemory pass of its own.

### Recommendation
Land this as the next-fork (BAL) gate now; add `blockTest`/`stateTest` matrix entries once timings are accepted; implement sync/transaction modes in nethtest; then retire the pyspec NUnit shards from both unit-test workflows.

kamilchodola:
@claude re-review

.github/workflows/run-nethtest.yml
```diff
@@ -0,0 +1,270 @@
+name: Run nethtest (EEST engine tests)
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+on:
+  push:
```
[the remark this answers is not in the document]
kamilchodola: Fixed — `run-nethtest.yml` no longer has branch-specific triggers at all: it is now a reusable `workflow_call` template invoked from `nethermind-tests.yml` and `nethermind-tests-flat.yml` on every PR/push (plus `workflow_dispatch` for manual runs).

.github/workflows/run-nethtest.yml
```diff
@@ -0,0 +1,270 @@
...
+    shell: bash
+
+jobs:
+  nethtest:
+    name: nethtest --${{ inputs.test_type || 'engineTest' }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v6
```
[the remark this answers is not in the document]
kamilchodola: These tags are current: `actions/checkout@v6`, `actions/setup-dotnet@v5`, and `actions/upload-artifact@v7` are used across the repo's workflows on master. The reviewer-bot knowledge cutoff predates them. No change needed.

src/Nethermind/Ethereum.Test.Base/BlockchainTestBase.cs
```diff
@@ -299,54 +326,55 @@ private static BlockHeader SuggestBlocks(BlockchainTest test, bool failOnInvalid
...
             parentHeader = correctRlp[i].Block.Header;
         }
-        return parentHeader;
+        return (parentHeader, lastBlockError);
     }
 
     private static readonly Dictionary<int, int> NewPayloadParamCounts = Enumerable
         .Range(1, EngineApiVersions.NewPayload.Latest)
         .ToDictionary(v => v, v => (typeof(IEngineRpcModule).GetMethod($"engine_newPayloadV{v}")
             ?? throw new NotSupportedException($"engine_newPayloadV{v} not found on IEngineRpcModule")).GetParameters().Length);
```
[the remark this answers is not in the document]
kamilchodola: Keeping as-is deliberately: the eager build makes a missing `engine_newPayloadV{v}` fail loudly on first use, and the `NotSupportedException` message names the exact constant to update (`EngineApiVersions.NewPayload.Latest`); it is preserved as the inner exception of the `TypeInitializationException`. Lazy init would only defer the same failure in test infrastructure.

nethtest
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+exec dotnet run --no-build -c Release --project "$SCRIPT_DIR/src/Nethermind/Nethermind.Test.Runner/Nethermind.Test.Runner.csproj" -- "$@"
```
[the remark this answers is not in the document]
kamilchodola: Already fixed — the `nethtest` script now passes `--no-launch-profile` (and `--no-build`) like the workflow does.

src/Nethermind/Nethermind.Test.Runner/Program.cs
```diff
@@ -72,71 +95,212 @@ public static async Task<int> Main(params string[] args)
...
             {
-                await RunBlockTest(input, source => new BlockchainTestsRunner(
-                    source,
-                    parseResult.GetValue(Options.Filter),
-                    chainId,
-                    parseResult.GetValue(Options.TraceAlways),
-                    !parseResult.GetValue(Options.ExcludeMemory),
-                    parseResult.GetValue(Options.ExcludeStack)));
+                bool forceJson = isEngineTest || jsonOutput;
+                List<EthereumTestResult> results = await RunBlockTestFiles(files, filter, chainId, trace, traceMemory, excludeStack, forceJson, workers);
```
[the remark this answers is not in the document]
kamilchodola: Fixed — the engine/block path now always serializes the collected results to stdout; the `forceJson` discard path no longer exists.

src/Nethermind/Nethermind.Test.Runner/Program.cs
```diff
@@ -72,71 +95,212 @@ public static async Task<int> Main(params string[] args)
...
 
-    private static async Task RunBlockTest(string path, Func<ITestSourceLoader, IBlockchainTestRunner> testRunnerBuilder)
+    private static List<string> CollectFiles(string path)
+    {
+        if (File.Exists(path))
+            return [path];
+
+        if (Directory.Exists(path))
+            return Directory.GetFiles(path, "*.json", SearchOption.AllDirectories)
+                .Where(f => !f.Contains("/.meta/") && !f.Contains("\\.meta\\"))
```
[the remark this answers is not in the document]
kamilchodola: Fixed — file collection uses explicit loops with `List.Sort(StringComparer.Ordinal)`; no LINQ remains on that path.

src/Nethermind/Ethereum.Test.Base/EthereumTestResult.cs
```diff
@@ -20,20 +21,33 @@ public EthereumTestResult(string? name, string? fork, string loadFailure)
...
         {
         }
 
+        [JsonIgnore]
         public string? LoadFailure { get; set; }
         public string Name { get; set; }
         public bool Pass { get; set; }
         public string Fork { get; set; }
 
+        public string Error { get; set; } = "";
```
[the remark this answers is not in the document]
kamilchodola: Fixed — `Error` is now `string?` with `[JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]`, so passing tests no longer emit an `"error"` property.

src/Nethermind/Nethermind.Test.Runner/StateTestRunner.cs
```diff
@@ -105,5 +116,45 @@ public IEnumerable<EthereumTestResult> RunTests()
 
             return results;
         }
+
+        public EthereumTestResult RunSingleTest(GeneralStateTest test)
+        {
+            test.ChainId = _chainId;
+
+            EthereumTestResult result = null;
```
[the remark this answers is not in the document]
kamilchodola: Fixed — declared as `EthereumTestResult? result = null;` with the null-forgiving return after the two guaranteed-assignment paths.

src/Nethermind/Ethereum.Test.Base/BlockchainTestBase.cs
```diff
@@ -53,12 +53,17 @@ public abstract class BlockchainTestBase
     /// </summary>
     protected virtual bool? ParallelExecutionOverride => null;
 
-    /// <summary>
     /// Override to force BAL batch-read prewarming on or off in tests.
     /// Null means use the default config value.
     /// </summary>
     protected virtual bool? ParallelExecutionBatchReadOverride => null;
```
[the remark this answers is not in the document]
kamilchodola: Fixed by the follow-up commits — the doc comments on the override properties are well-formed now.

src/Nethermind/Ethereum.Test.Base/BlockchainTestBase.cs
```diff
@@ -53,12 +53,21 @@ public abstract class BlockchainTestBase
     /// </summary>
     protected virtual bool? ParallelExecutionOverride => null;
 
+    /// Override to force BAL batch-read prewarming on or off in tests.
+    /// Null means use the default config value.
+    /// </summary>
     /// <summary>
     /// Override to force BAL batch-read prewarming on or off in tests.
     /// Null means use the default config value.
     /// </summary>
```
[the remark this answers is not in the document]
kamilchodola: Resolved by the follow-up commit — the dangling text was removed and a single clean doc block remains.

src/Nethermind/Ethereum.Test.Base/BlockchainTestBase.cs
```diff
@@ -58,6 +58,13 @@ public abstract class BlockchainTestBase
...
     protected virtual bool? ParallelExecutionBatchReadOverride => null;
+    protected virtual bool? ParallelExecutionBatchReadOverride => null;
```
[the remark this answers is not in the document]
kamilchodola: Fixed — the duplicate declaration was removed; a single `ParallelExecutionBatchReadOverride` property remains.

src/Nethermind/Ethereum.Test.Base/BlockchainTestBase.cs
kamilchodola: Applied — `RunTest` returns the object-initializer form as suggested.

src/Nethermind/Ethereum.Test.Base/BlockchainTestBase.cs
kamilchodola: Applied — `SuggestBlocks` now returns `Result<BlockHeader>`: the header to continue from in `Data`, and the expected-invalid-block rejection message (which does not fail the test) in `Error`, using `Result<T>.Fail(error, data)` to carry both.

src/Nethermind/Ethereum.Test.Base/BlockchainTestBase.cs
kamilchodola: Applied — `RunNewPayloads` now returns `Task<Result<string>>` with the last payload status as `Data` and the validation error as `Error`.

src/Nethermind/Ethereum.Test.Base/GeneralStateTestBase.cs
kamilchodola: Applied (equivalent pattern) — `txResult is { TransactionExecuted: false } failedResult ? failedResult.ErrorDescription : blockValidationError`. Used `TransactionExecuted` rather than comparing `ResultType` against `TransactionResult.Ok` since the latter mixes types.

src/Nethermind/Nethermind.Test.Runner/BlockchainTestsRunner.cs
kamilchodola: Done — replaced the parameter list with the `BlockchainTestsRunnerOptions` readonly record struct.

src/Nethermind/Nethermind.Test.Runner/BlockchainTestsRunner.cs
kamilchodola: Addressed — `RunTestsAsync` was split into `ExecuteTestAsync` + `ReportResult` (+ progress helpers), and the constructor collapsed via the options struct.

kamilchodola:
@claude re-review

src/Nethermind/Ethereum.Test.Base/BlockchainTest.cs
```diff
@@ -11,6 +11,7 @@ namespace Ethereum.Test.Base
 {
     public class BlockchainTest : EthereumTest
     {
+        public string? ForkName { get; set; }
```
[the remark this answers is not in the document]
kamilchodola: Fixed — `Convert` now sets `ForkName` from the fixture's `network` string (covers transition networks like `ParisToShanghaiAtTime15k` as well), with a regression test in `BlockchainTestConversionTests`. Verified the runner output now reports real fork names instead of "unknown".

kamilchodola:
## 1:1 replacement audit — pyspec NUnit shards vs nethtest

Verified on the same release the NUnit suite pins (`tests-bal@v7.2.0`): NUnit baselines from the last green master runs (`27285302259`, `27289705918`), nethtest from audit run `27355316730`.

### Count parity (exact)

| Category | NUnit (master) | nethtest | |
|---|---:|---:|---|
| EEST fixture cases, sequential (engine 72,378 + block 82,941 + state 65,202 + sync 18 + tx 165) | 220,704 | **220,704** | ✅ exact |
| Amsterdam block-format variants (per config ×3) | 21,368 | **21,368** | ✅ exact |
| Amsterdam engine-format variants (per config ×3) | 21,387 | **21,387** | ✅ exact |
| zkEVM stateless preview (`tests-zkevm@v0.4.1`) | 23,787 | n/a — kept on NUnit | ✅ restored |
| `CiSentinelTests` keep-alive | 16 | not needed | ✅ no-op |
| **Reconciliation: 220,704 + 23,787 + 16 = 244,507 = NUnit Sequential total** | | | ✅ |

### Gaps the audit found (all fixed)

1. **Patricia Amsterdam variants**: `nethermind-tests.yml`'s pyspec chunks ran unfiltered, so the parallel/batch-read variant classes also executed under patricia (372,242 = sequential + all 6 variant classes — the counts prove it). Added to the sequential group (`7f59d43925`); note patricia+parallel-full is the production config.
2. **zkEVM stateless preview tests** (separate `tests-zkevm` archive, asserts `StatelessExecutor` output bytes — not EEST fixture replay): restored as dedicated filtered NUnit chunks (`7b2807529f`).
3. **v7.2.0 memory headroom**: one flat parallel-variant chunk OOMed at ~1.8k tests single-worker; flat parallel variants now run 16 chunks ≈ 1.3k tests/process — the same density the NUnit variant shards ran without OOM (`8566b42e7d`).

### Behavior parity (verified equivalent)
Same `BlockchainTestBase`/`GeneralStateTestBase`/`TransactionTestBase` execution paths; same converters/loaders; same ChainId semantics; `failOnInvalidRlp` default; no skip guards apply on linux-x64; `.meta` exclusion is a no-op for NUnit's dirs.

### Deliberate non-1:1 deltas (no coverage lost)
- nethtest reads whole fixture trees → superset of NUnit's per-fork classes; new fork dirs are covered without adding classes.
- `txTest` and zkEVM are not duplicated under flat (they never touch the state layout; NUnit's flat-workflow duplication of them was a no-op).
- The checked/no-intrinsics variant workflow (`nethermind-tests-checked.yml`) still runs its own Pyspec subset under `TEST_SKIP_HEAVY=1` — unchanged.

Follow-up candidates (not blockers): port zkEVM stateless mode into nethtest (needs `Nethermind.Stateless.Execution` + per-type archive plumbing); root-cause the parallel-execution/flat memory growth (dotMemory on `Nethermind.Test.Runner` + `--flatdb` engine fixtures).

kamilchodola:
@claude re-review

kamilchodola:
Re [the re-review](https://github.com/NethermindEth/nethermind/pull/11399#issuecomment-4682449278): all five remaining Medium findings addressed in `16da2b4c9c`:

- **M1**: `EthereumTestResult.StateRoot` now documents that it is only populated by state tests (omitted from blockchain/engine/tx JSON).
- **M3**: the filter regex is compiled once in `Program.RunBlockTestFiles`'s caller and shared across the per-file runner instances — `BlockchainTestsRunnerOptions.Filter` is now `Regex?`.
- **M4**: `StateTestsRunner` gained a sourceless constructor for `RunSingleTest` callers; the fake `"dummy"` `ITestSourceLoader` is gone and `RunTests()` guards explicitly.
- **M5**: `BuildStackJson` JSON-escapes stack values via `JsonSerializer.Serialize` (same approach as the `error` field).
- **M6**: comment added confirming the streaming tracer is intentionally per-test (independent JSONL trace per test; counter reset by design).

L3/L5 left as-is per the review's non-blocking note: the `.meta` check already covers both path separators, and renaming the `-p` alias would break existing CLI usage.

kamilchodola:
@claude re-review

Check the correctness of pyspec test replacement with that PR. Check also if this is any profitable in terms of runners usage and job-minutes comparing to any other PR. What matters here is a per commit-trigger time spent and runners used on pyspec like tests.
Also confirm that this is 1-1 replacement with no tests missing.

benaadams:
Merge conflict

.github/workflows/nethermind-tests-flat.yml
```diff
@@ -63,171 +63,14 @@ jobs:
...
-          - { project: Ethereum.Blockchain.Pyspec.Test, chunk: 15of16, filter: FullyQualifiedName~AmsterdamParallelFullEngineBlockchainTests, label: ParallelFullEngine }
-          - { project: Ethereum.Blockchain.Pyspec.Test, chunk: 16of16, filter: FullyQualifiedName~AmsterdamParallelFullEngineBlockchainTests, label: ParallelFullEngine }
     steps:
-      - name: Free up disk space
-        if: startsWith(matrix.project, 'Ethereum.Blockchain.Pyspec')
-        uses: jlumbroso/free-disk-space@v1.3.1
-        with:
-          large-packages: false
-          tool-cache: false
-          swap-storage: false
```
benaadams, line 222: Do we want to remove the free up diskspace job? @rubo

kamilchodola: it was done exlusively because of Pyspec and now it is getting reworked so this is not needed anymore

kamilchodola:
@claude re-review

Check the correctness of pyspec test replacement with that PR. Check also if this is any profitable in terms of runners usage and job-minutes comparing to any other PR. What matters here is a per commit-trigger time spent and runners used on pyspec like tests.
Also confirm that this is 1-1 replacement with no tests missing.

kamilchodola:
I noticed missing commits here - so please not merge until then

stdevMac (approved):

cbermudez97 (approved):

benaadams (approved):

PR #11400, opened 2026-04-29
consensus: standardize slow block JSON output for cross-client metrics

Marchhill:
Updates https://github.com/NethermindEth/nethermind/pull/10288

## Summary

Implement standardized JSON format for slow block logging to enable cross-client performance analysis and protocol research.

This change is part of the **Cross-Client Execution Metrics** initiative proposed by Gary Rong and CPerezz.

## Motivation

Standardized execution metrics are critical for:
- Cross-client performance comparison
- Network health monitoring
- Data-driven protocol research

**Real-world example:** The [EIP-7907 analysis](https://ethresear.ch/t/data-driven-analysis-on-eip-7907/23850) used execution metrics to measure code read latency, per-call overhead scaling, and block execution breakdown. Without standardized metrics across clients, such analysis cannot be validated cross-client.

## References

- Original proposal: https://hackmd.io/dg7rizTyTXuCf2LSa2LsyQ (Gary Rong)
- EIP-7907 analysis: https://ethresear.ch/t/data-driven-analysis-on-eip-7907/23850

## Configuration

Two new flags in `Blocks` config section:

| Flag | Default | Description |
|------|---------|-------------|
| `Blocks.SlowBlockThresholdMs` | `-1` (disabled) | Threshold in ms for logging slow blocks. Set to `1000` to log blocks slower than 1s. Set to `0` to log all blocks. |
| `Blocks.SlowBlockPerTxThresholdMs` | `-1` (disabled) | Per-transaction threshold in ms. When enabled, individual transactions exceeding this threshold are included in the slow block JSON. Set to `0` to log all transactions. |

Both flags default to `-1` (disabled). When slow-block logging is disabled, the block-level delta capture, per-transaction timing, and JSON serialization paths are skipped.

The exported execution counters added by this PR are controlled by the compile-time `NETHERMIND_NO_EXECUTION_METRICS` symbol, not by the runtime slow-block flags. Default builds keep those counters enabled for Prometheus; benchmark builds can define `NETHERMIND_NO_EXECUTION_METRICS` to let the JIT fold the gated counter/timer paths away.

### Example usage

```
# Log blocks slower than 1s with all transactions
--Blocks.SlowBlockThresholdMs=1000 --Blocks.SlowBlockPerTxThresholdMs=0

# Log blocks slower than 500ms, only transactions slower than 50ms
--Blocks.SlowBlockThresholdMs=500 --Blocks.SlowBlockPerTxThresholdMs=50

# Log every block, no per-tx breakdown
--Blocks.SlowBlockThresholdMs=0
```

## JSON Output Format

```json
{
  "level": "warn",
  "msg": "Slow block",
  "parallel_execution": false,
  "block": {
    "number": 22075123,
    "hash": "0xabc...",
    "gas_used": 29850000,
    "gas_limit": 30000000,
    "tx_count": 142,
    "blob_count": 6
  },
  "timing": {
    "execution_ms": 52.1,
    "evm_ms": 45.2,
    "blooms_ms": 1.3,
    "receipts_root_ms": 3.8,
    "commit_ms": 2.1,
    "storage_merkle_ms": 18.4,
    "state_root_ms": 5.2,
    "state_hash_ms": 23.6,
    "total_ms": 82.0
  },
  "throughput": {
    "mgas_per_sec": 364.1
  },
  "state_reads": {
    "accounts": 322,
    "storage_slots": 3201,
    "code": 35,
    "code_bytes": 482310
  },
  "state_writes": {
    "accounts": 312,
    "accounts_deleted": 0,
    "storage_slots": 1805,
    "storage_slots_deleted": 12,
    "code": 3,
    "code_bytes": 8420,
    "eip7702_delegations_set": 0,
    "eip7702_delegations_cleared": 0
  },
  "cache": {
    "account": { "hits": 1520, "misses": 322, "hit_rate": 82.52 },
    "storage": { "hits": 2800, "misses": 401, "hit_rate": 87.48 },
    "code": { "hits": 180, "misses": 35, "hit_rate": 83.72 }
  },
  "evm": {
    "opcodes": 1542000,
    "sload": 3201,
    "sstore": 1805,
    "calls": 2412,
    "empty_calls": 48,
    "creates": 3,
    "self_destructs": 0,
    "contracts_analyzed": 118,
    "cached_contracts_used": 97
  },
  "transactions": [
    {
      "index": 42,
      "hash": "0xdef...",
      "gas_used": 8200000,
      "execution_ms": 12.3,
      "type": "Legacy",
      "to": "0x7a25..."
    },
    {
      "index": 107,
      "hash": "0x123...",
      "gas_used": 5100000,
      "execution_ms": 8.7,
      "type": "EIP1559",
      "to": "0xdead..."
    }
  ]
}
```

### State read field definitions

The `state_reads` fields are derived from existing cache counters where possible so the hot path does not record the same event twice:

| Field | Definition |
|-------|------------|
| `state_reads.accounts` | Account trie cache misses, equal to `cache.account.misses`. |
| `state_reads.storage_slots` | Storage trie accesses served through the storage trie cache, equal to `cache.storage.hits + cache.storage.misses`. |
| `state_reads.code` | Code cache misses, equal to `cache.code.misses`. |
| `state_reads.code_bytes` | Total bytes loaded for code cache misses. |

### EVM field definitions

| Field | Definition |
|-------|------------|
| `evm.cached_contracts_used` | Existing code cache hit counter for the block, equal to `cache.code.hits`; retained because it is part of the original cross-client metrics spec. |

### Timing field definitions

| Field | Definition |
|-------|-----------|
| `total_ms` | Wall-clock block processing time |
| `execution_ms` | `total - state_hash - commit` (backwards-compatible) |
| `evm_ms` | `total - state_hash - commit - blooms - receipts_root` (pure EVM) |
| `blooms_ms` | Bloom filter calculation (parallel) |
| `receipts_root_ms` | Receipts Patricia trie root computation |
| `commit_ms` | State commit without root recalculation |
| `storage_merkle_ms` | Storage trie merkleization (`Commit(commitRoots: true)`) |
| `state_root_ms` | State root recalculation + commit tree |
| `state_hash_ms` | `storage_merkle + state_root` (total merkleization) |

The `transactions` array is only present when `SlowBlockPerTxThresholdMs >= 0`.

## Implementation

- Account and storage read counts are derived from existing `Db.Metrics` trie cache/read counters.
- Code cache misses and code bytes read are tracked internally for slow-block JSON; they are not exported as duplicate Prometheus counters.
- Write counters and EIP-7702 delegation counters are compile-time gated via `ExecutionMetricsFlag`.
- Timing uses `Stopwatch.GetTimestamp()` only at coarse block-phase boundaries and for optional per-transaction timing.
- JSON is serialized via `Utf8JsonWriter` + `ArrayBufferWriter<byte>` on the ThreadPool, not the block-processing thread.
- Per-transaction timing uses pooled `ArrayPoolList<long>` storage and one `Stopwatch.GetTimestamp()` per transaction when enabled.
- Delta computation and JSON output are queued to ThreadPool via `UnsafeQueueUserWorkItem`.


src/Nethermind/Nethermind.Consensus/Processing/PerTxTimingCollector.cs
```diff
@@ -0,0 +1,56 @@
...
+{
+    [ThreadStatic] private static bool _enabled;
+    [ThreadStatic] private static long[]? _ticksPerTx;
+    [ThreadStatic] private static int _count;
+
+    /// <summary>Whether per-tx timing capture is active on this thread.</summary>
+    public static bool IsEnabled => _enabled;
+
+    /// <summary>Called by ProcessingStats to enable/disable capture on the block-processing thread.</summary>
+    public static void SetEnabled(bool enabled) => _enabled = enabled;
```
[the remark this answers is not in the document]
Marchhill: this is set in config so won't be changed dynamically

src/Nethermind/Nethermind.State/StateProvider.cs
```diff
@@ -731,6 +736,12 @@ public bool WarmUp(Address address)
 
     internal void SetState(Address address, Account? account)
```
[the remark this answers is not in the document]
Marchhill: I believe this is the intended behaviour

src/Nethermind/Nethermind.State/PersistentStorageProvider.cs
```diff
@@ -269,7 +270,11 @@ public void WarmUp(in StorageCell storageCell, bool isEmpty)
             LoadFromTree(in storageCell);
     }
 
-    private ReadOnlySpan<byte> LoadFromTree(in StorageCell storageCell) => GetOrCreateStorage(storageCell.Address).LoadFromTree(storageCell);
+    private ReadOnlySpan<byte> LoadFromTree(in StorageCell storageCell)
+    {
+        EvmMetrics.IncrementStorageReads();
```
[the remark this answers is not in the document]
Marchhill: I would guess this is intended behaviour if we care about reading from db rather than cache

kamilchodola:
@claude re-review

LukaszRozmej:
@claude will this cause any performance regressions? Can we mitigate it?

src/Nethermind/Nethermind.Consensus.Test/Processing/ProcessingStatsTests.cs
```diff
@@ -0,0 +1,155 @@
...
+        Assert.That(json, Does.Contain("opcodes"));
+        Assert.That(json, Does.Contain("empty_calls"));
+        Assert.That(json, Does.Contain("self_destructs"));
+        Assert.That(json, Does.Contain("contracts_analyzed"));
+        Assert.That(json, Does.Contain("cached_contracts_used"));
+        Assert.That(json, Does.Contain("evm_ms"));
+        Assert.That(json, Does.Contain("blooms_ms"));
+        Assert.That(json, Does.Contain("receipts_root_ms"));
+        Assert.That(json, Does.Contain("storage_merkle_ms"));
+        Assert.That(json, Does.Contain("state_root_ms"));
```
LukaszRozmej: maybe just validate json schema? We do it already in some tests.

src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.BlockValidationTransactionsExecutor.cs
```diff
@@ -53,11 +56,34 @@ public virtual TxReceipt[] ProcessTransactions(Block block, ProcessingOptions pr
 
         protected virtual void ProcessTransaction(Block block, Transaction currentTx, int index, BlockReceiptsTracer receiptsTracer, ProcessingOptions processingOptions)
         {
+            long txStart = StartTxTimer();
             TransactionResult result = transactionProcessor.ProcessTransaction(currentTx, receiptsTracer, processingOptions, _stateProvider);
+            StopTxTimer(index, txStart);
```
LukaszRozmej: should this be in finally?

src/Nethermind/Nethermind.Consensus/Processing/BlockProcessor.cs
```diff
@@ -180,8 +179,44 @@ protected virtual TxReceipt[] ProcessBlock(
...
+
+    private void ComputeStateRoot(BlockHeader header)
+    {
+        long start = Stopwatch.GetTimestamp();
+        _stateProvider.RecalculateStateRoot();
+        long ticks = Stopwatch.GetElapsedTime(start).Ticks;
+        Evm.Metrics.IncrementStateHashTime(ticks);
+        Evm.Metrics.IncrementStateRootTime(ticks);
+        header.StateRoot = _stateProvider.StateRoot;
+    }
```
LukaszRozmej, line 201: Can we do this in a reusable version:

```
public readonly ref struct MetricsTimer<TMetrics>
    where TMetrics : IMetricSink
{
    private readonly long _start;

    public MetricsTimer()
    {
        _start = Stopwatch.GetTimestamp();
    }

    public void Dispose()
    {
            long ticks = Stopwatch.GetElapsedTime(_start).Ticks;
            TMetrics.AddTicks(ticks);
    }
}
```

Then usage is very simple and clean:

`    using MetricsTimer<StateRootMetricSink> _ = new();`

src/Nethermind/Nethermind.Consensus/Processing/PerTxTimingCollector.cs
```diff
@@ -0,0 +1,59 @@
...
+
+    /// <summary>Called by the tx executor before processing transactions.</summary>
+    public static void Prepare(int txCount)
+    {
+        if (_ticksPerTx is null || _ticksPerTx.Length < txCount)
+        {
+            _ticksPerTx = new long[Math.Max(txCount, 256)];
```
LukaszRozmej: Potentially a lot of allocations?
Use pooled memory?

src/Nethermind/Nethermind.Consensus/Processing/PerTxTimingCollector.cs
```diff
@@ -0,0 +1,59 @@
...
+    /// <summary>
+    /// Takes a snapshot of per-tx timing data. Returns null if no data was captured.
+    /// The returned array is a copy safe to use on another thread.
+    /// </summary>
+    public static long[]? Snapshot()
+    {
+        if (!_enabled || _ticksPerTx is null || _count == 0) return null;
+        long[] copy = new long[_count];
+        Array.Copy(_ticksPerTx, copy, _count);
+        return copy;
```
LukaszRozmej: even more allocations

LukaszRozmej (commented):
Probably not worth it but can we potentially gate it under TOn/TOff flag so it can be completely removed from code if not on?

kamilchodola (approved):

Marchhill:
@claude review

Marchhill:
@claude  review

LukaszRozmej:
## Review — Standardize slow block JSON output for cross-client metrics

Reviewed against the latest commit. Bot's High/Medium findings (rounds 1–5) all confirmed fixed.

### Overview

Three pieces: (1) new counter infrastructure in `Evm.Metrics` and `Db.Metrics` gated by a compile-time-foldable `ExecutionMetricsFlag.IsActive`; (2) block-phase timing via a `MetricsTimer<TSink>` ref struct with sinks for commit / state root / blooms / receipts root; (3) JSON emitter in `ProcessingStats` writing to a dedicated `SlowBlocks` logger from the ThreadPool, with optional per-tx timing through the static `PerTxTimingCollector` (now `ArrayPoolList<long>`-backed).

### Strong points

- Pool-backed JSON via `ArrayBufferWriter<byte>` + `Utf8JsonWriter` in `using`. `BlockData` pooled; `PerTxTicks` disposed in `BlockDataPolicy.Return`.
- Deltas computed on the block-processing thread; encoding dispatched to ThreadPool. `try/catch` around `LogSlowBlock` so diagnostics can't abort processing.
- `MetricsTimer<TSink>` with `static abstract IsEnabled` is a clean JIT-foldable pattern.
- Default interface methods on `IBlockTransactionsExecutor` keep timing instrumentation out of executors that don't need it.

### Issues / risks

**Per-tx collector assumes a single concurrent block.** `PerTxTimingCollector` is process-wide static. The xmldoc acknowledges this, but if two `BlockProcessor` instances ever run concurrently (validation + production, chain processor + simulation), the second `Prepare()` disposes the first's list out from under it, and parallel-worker `Record()` calls from block A can land in block B's array. Consider scoping the state to the `BlockProcessor` instance.

**Sub-phase timing can sum to more than `total_ms` under parallelism.** `LogSlowBlock` defensively clamps `executionMs`/`evmMs` to non-negative because `ParallelBlockValidationTransactionsExecutor` sets `IsBlockProcessingThread = true` on multiple workers, so `_main*` accumulates wall-clock time across parallel work. The clamp is fine, but the JSON consumers (the whole point of cross-client metrics) will see breakdowns that don't add up to total. Worth calling out in the `Timing field definitions` table — otherwise downstream analyses will trip.

**`ExecutionMetricsFlag.IsActive => true` is a dead toggle.** Every guarded `if (!ExecutionMetricsFlag.IsActive) return;` folds to nothing at JIT time — runtime cost is zero, but the source noise and the implication that the flag can be toggled at runtime are misleading. Either wire it to a `#if` symbol / static config, or remove the guards.

**Unconditional `Interlocked.Increment` per state I/O.** Even with the flag effectively on, every `SetState` / `PersistentStorageProvider.Set` / `LoadFromTree` / `IncrementCodeReads` does an `AsyncLocal<bool>` read + a `LOCK XADD`. Round 4 measured ~0.2–0.5% per-block overhead. The PR description's "zero overhead when disabled" is true only for the delta/JSON layer, not for the per-op counter increments. Either update the description, gate the increments behind a runtime check, or take Round 4's Option 1 (switch `IsBlockProcessingThread` from `AsyncLocal` to `[ThreadStatic]`, which would benefit every existing counter too).

**`cached_contracts_used` ≡ `cache.code.hits`.** Both fields are computed from `MainThreadCodeDbCache` deltas. The PR's example JSON shows different values (97 vs 180), which would mislead consumers. Drop one or document the equivalence.

**`PerTxTimingCollector` static fields are not `volatile`** (Round 2, acknowledged). Safe on x86/.NET (TSO + ThreadPool dispatch establishes ordering between `Prepare` and `Record`). Worth a one-line `<remarks>` documenting this rather than leaving it implicit.

### Code quality / style

- `SlowBlockLogEntry` lives in `Nethermind.Core.Test` but is only used by `Nethermind.Consensus.Test`. Move it closer to its consumer.
- Two `ProcessingStats` ctors disagree on the default threshold (test ctor: `1000`; config ctor: `-1`). Pick one.
- `MetricsIntegrationTests` and `SlowBlockIntegrationTests` duplicate ~30 lines of fixture setup. `AGENTS.md` explicitly calls this out — factor a shared base while the files are new.
- Stray double blank line before `ThrowInvalidTransactionException` in `BlockValidationTransactionsExecutor.cs`.

### Tests

Solid for the sequential path (declarative schema validation + end-to-end EVM→JSON for ETH transfer / SLOAD / SSTORE / CALL / EIP-7702 set+clear).

Gaps:
- No coverage of `ProcessTransactionsParallel`, which is exactly where the process-wide-static collector design matters.
- `MetricsIntegrationTests` reads combined `Metrics.AccountReads` (not `MainThread*`) — under parallel test execution another fixture's increments contaminate the delta; `Is.GreaterThanOrEqualTo` softens this but weakens the spec.
- `Thread.Sleep(50)` polling up to 5s per assertion. Prefer a `ManualResetEventSlim` set by the logger, or a synchronous test code path.
- `[SetUp]` sets `ProcessingThread.IsBlockProcessingThread = true` on the runner thread; with parallel fixtures another test on the same thread can observe the flag at the wrong moment. Use a scoped helper.
- No explicit test that an exception thrown from `LogSlowBlock` does not escape — that's a load-bearing guarantee worth pinning.

### Security

Nothing concerning. Bounded log size, no string concatenation, `Utf8JsonWriter` everywhere, `try/catch` shields block processing.

### Recommendation

Blockers before merge:
1. Move `SlowBlockLogEntry` out of `Nethermind.Core.Test`.
2. Document the parallel-execution caveat for the breakdown timings.
3. Resolve the `cached_contracts_used` / `cache.code.hits` duplication.
4. Remove or wire up `ExecutionMetricsFlag.IsActive`.
5. Fix the "zero overhead when disabled" claim in the PR description, or actually gate the counter increments.

Non-blocking but worth a follow-up: per-tx collector single-block assumption, test fragility (sleep-polling, static thread-flag mutation, combined-counter assertions), shared fixture base for the two integration test files.

Marchhill:
@claude review

Marchhill:
@claude review

LukaszRozmej:
# Review — focused on correctness and performance

## Verdict
Feature is well-designed and test coverage is excellent. However, the PR description's claim that **"there is zero overhead on block processing when disabled"** is **not quite accurate**. The new EVM-side metric counters fire unconditionally in default builds (gated only by compile-time `ExecutionMetricsFlag`, not by `SlowBlockThresholdMs`). On hot paths (SSTORE, SLOAD, account reads/writes), this adds one atomic `Interlocked` op per state operation, always on, even when the feature is configured to `-1`.

---

## Performance concerns

### 1. Hot-path atomics added regardless of `SlowBlockThresholdMs` (HIGH)
`ExecutionMetricsFlag.IsActive` is a compile-time-folded constant, **`true` by default**. Only a recompile with `NETHERMIND_NO_EXECUTION_METRICS` disables it. So in production builds, even with `SlowBlockThresholdMs = -1`:

- `PersistentStorageProvider.Set` → `IncrementStorageWrites()` on every SSTORE that mutates
- `EvmInstructions.Storage.cs` (both `InstructionSStoreMetered` and `InstructionSStoreUnmetered`) → `IncrementStorageDeleted()` on every SSTORE-to-zero
- `PersistentStorageProvider.LoadFromTree` → `IncrementStorageReads()` on every uncached SLOAD
- `StateProvider.GetState` (via `WarmUp`) → `IncrementAccountReads()` on every uncached account touch
- `StateProvider.SetState` → `IncrementAccountWrites()` (+ `IncrementAccountDeleted()` when null) on every state mutation
- `CodeInfoRepository.GetCodeInfo` → `IncrementCodeReads()` + `IncrementCodeBytesRead()` on every code DB fetch

Each is `Interlocked.Increment(ref IsBlockProcessingThread ? ref _main... : ref _other...)`. On x64 that's ~10–15 ns per call (uncontested atomic) plus a thread-static read. For the example block in the PR description (~1800 SSTOREs, ~3200 SLOADs, ~2000 account changes), that's **~70–100 µs/block** of overhead even when the feature is "disabled" by config. The PR description quotes ~31 µs/block for slow-block capture; the real per-block tax in default builds is larger and unavoidable without recompiling.

**Suggestion:** Gate `Increment*` calls behind a **runtime** flag (e.g. `_blocksConfig.SlowBlockThresholdMs >= 0` exposed as a `volatile` static), or set `ExecutionMetricsFlag.IsActive` from config at startup. If you keep the compile-time flag, please correct the PR description.

### 2. Per-tx `try/finally` is added unconditionally
`BlockProcessor.BlockValidationTransactionsExecutor.ProcessTransaction` now wraps `transactionProcessor.ProcessTransaction(...)` in `try/finally`. Even with `_enableTxTimingMetrics = false`, the JIT still emits the exception-filter table. Tiny per-tx cost but always on.

In the parallel executor, `inner.StartTxTimer()` / `inner.StopTxTimer()` are **interface calls through `IBlockProcessor.IBlockTransactionsExecutor`** (default interface methods). The JIT likely cannot devirtualize since `inner` is interface-typed at the call site — 2 vcalls per tx in the parallel hot path.

### 3. `LogSlowBlock` recomputes blob count
`Transaction.GetBlobCount()` is summed both in `UpdateStats` (existing) and again in `LogSlowBlock` (new). Cheap, but redundant — store the blob count on `BlockData` and reuse.

---

## Correctness concerns

### 1. `PerTxTimingCollector` is process-wide static
`_ticksPerTx` is a `static` field, written on the block-processing thread (`Prepare`), read/written on parallel ThreadPool workers (`Record`). It is **not volatile**. Workers see writes only because `ParallelUnbalancedWork.For` provides the implicit synchronisation barrier. This works today but is fragile — please document the dependency explicitly. Same applies to `BlockValidationTransactionsExecutor._enableTxTimingMetrics`.

### 2. `PerTxTimingCollector.SetEnabled(true)` is sticky
`CaptureStartStats` calls `SetEnabled(true)` per block when `_slowBlockPerTxThresholdMs >= 0`, but there is no corresponding `SetEnabled(false)`. If the config is hot-reloaded or in a multi-`ProcessingStats` scenario, the static stays `true`. The `Prepare`/`Snapshot` flow tolerates this, but the API is asymmetric.

### 3. `Stopwatch` ticks vs `TimeSpan` ticks
`MetricsTimer<T>.Dispose` uses `Stopwatch.GetElapsedTime(_start).Ticks` which returns `TimeSpan` ticks (100 ns) — dividing by `TimeSpan.TicksPerMillisecond` later is correct. Worth a doc comment on `IncrementStateHashTime(long ticks)` clarifying the unit is `TimeSpan` ticks, not `Stopwatch` ticks, to avoid future regressions.

### 4. Negative-time fallback hides a real bug
```csharp
double executionMs = totalMs - stateHashMs - commitMs;
if (executionMs < 0) executionMs = totalMs;
```
If this branch is ever taken, the log entry silently lies (claims execution ≥ total). At minimum, debug-log when clamping so the bug surfaces.

### 5. `BlockData` pool — new `Delta*` fields not reset on `Return`
`BlockDataPolicy.Return` resets `PerTxTicks` and the existing fields but **not** the new `DeltaAccountReads`, `DeltaStorageReads`, etc. Safe **only because** `UpdateStats` either writes all of them (threshold ≥ 0) or never reads them (threshold < 0). If the threshold ever changes at runtime, stale deltas leak. Belt-and-braces fix: reset them too.

### 6. Cross-client metric semantics under parallel execution
The `LogSlowBlock` `<remarks>` correctly notes that under `ParallelExecution` the EVM and state-read/write counters sum across workers while wall-clock timings remain per-block. Good documentation, but this is a real caveat for "cross-client" comparability — Geth/Reth don't run txs in parallel by default. Consumers will see inflated counter values relative to wall-clock ms. Worth surfacing in the JSON itself (e.g. `parallel_execution: true`) so analysers can normalise.

---

## Minor
- `private bool _enableTxTimingMetrics = false;` — `= false` is redundant.
- Double blank line after `StopTxTimer`.
- `IBlockProcessor.IBlockTransactionsExecutor` gains 3 new default interface methods — public-API change and a small ABI risk for plugins implementing the interface.
- `MetricsTimer<T>` as a parameterless `readonly ref struct` is nice — stack-only, inlinable, scope-based timing.

---

## Suggested fixes before merge
1. Either gate hot-path `Increment*` by a **runtime** flag tied to `SlowBlockThresholdMs >= 0`, or correct the "zero overhead when disabled" claim in the PR body and docs.
2. Document the synchronisation dependency of `PerTxTimingCollector` / `_enableTxTimingMetrics` on `ParallelUnbalancedWork.For`.
3. Reset the new `Delta*` fields in `BlockDataPolicy.Return`.
4. Add a `parallel_execution` boolean to the JSON so cross-client analyses can normalise counter values.
5. Tiny: store blob count in `BlockData`; remove `= false` init; debug-log the `executionMs < 0` fallback.

Marchhill:
@claude review

benaadams (approved):

benaadams:
@claude re-review

benaadams (approved):

benaadams:
@claude re-review

benaadams (approved):

PR #11401, opened 2026-04-29
Snapshot fixes

svlachakis:
## Changes

- Fix 1 — SnapshotDownloader.cs: HTTP 200 on resume now skips bytes instead of overwriting
- Fix 2 — InitDatabaseSnapshot.cs: disk-space check uses the correct mount

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

## Testing

Manual test of Arbitrum Snapshot is in progress.

#### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Init.Snapshot/InitDatabaseSnapshot.cs
+++ b/src/Nethermind/Nethermind.Init.Snapshot/InitDatabaseSnapshot.cs
@@ -107,6 +107,7 @@ public class InitDatabaseSnapshot(INethermindApi api) : IStep
             return;
 
         TimeSpan retryDelay = TimeSpan.FromSeconds(InitialRetryDelaySeconds);
+        long lastSize = GetFileSize(destinationPath);
 
         while (true)
         {
@@ -126,6 +127,11 @@ public class InitDatabaseSnapshot(INethermindApi api) : IStep
             }
             catch (Exception e) when (e is IOException or HttpRequestException)
             {
+                long currentSize = GetFileSize(destinationPath);
+                if (currentSize > lastSize)
+                    retryDelay = TimeSpan.FromSeconds(InitialRetryDelaySeconds);
+                lastSize = currentSize;
+
                 if (_logger.IsError)
                     _logger.Error($"Snapshot download failed. Retrying in {retryDelay.TotalSeconds}s. Error: {e}");
                 await Task.Delay(retryDelay, cancellationToken).ConfigureAwait(false);
@@ -134,6 +140,12 @@ public class InitDatabaseSnapshot(INethermindApi api) : IStep
         }
 
         checkpoint.Advance(SnapshotStage.Downloaded);
+    }
+
+    private long GetFileSize(string path)
+    {
+        IFileInfo file = api.FileSystem.FileInfo.New(path);
+        return file.Exists ? file.Length : 0;
     }
 
     private async Task<bool> VerifyChecksumAsync(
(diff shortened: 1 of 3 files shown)


src/Nethermind/Nethermind.Init.Snapshot/InitDatabaseSnapshot.cs
```diff
@@ -182,15 +183,24 @@ private async Task ExtractAsync(
         checkpoint.Advance(SnapshotStage.Extracted);
     }
 
-    private static void CheckDiskSpace(string dbPath, string snapshotPath)
+    private void CheckDiskSpace(string dbPath, string snapshotPath)
     {
+        IDriveInfo[] drives = api.FileSystem.GetDriveInfos(dbPath);
```
damian-orzechowski, line 200: Wondering if  `[KeyFilter(nameof(IInitConfig.BaseDbPath))] IDriveInfo[] drives` for DI resolution wouldn't be better?

svlachakis: Good call on the pattern — but the keyed `IDriveInfo[]` is registered only in `HealthChecksPlugin`, so consuming it from `Nethermind.Init.Snapshot` couples us to `HealthChecks` being loaded. I think the clean version is to lift that registration to a shared place; happy to do that as a follow-up since it touches code outside this fix's scope. For now I think `api.FileSystem.GetDriveInfos(dbPath)` keeps the call self-contained.

svlachakis:
@claude review again, fixed one comment - the other two doesn't seem to be worth the complexity

src/Nethermind/Nethermind.Init.Snapshot/InitDatabaseSnapshot.cs
svlachakis: Issue created https://github.com/NethermindEth/nethermind/issues/11420

damian-orzechowski: Ok, got it - yes, let's not change it in this PR

damian-orzechowski (approved):

src/Nethermind/Nethermind.Init.Snapshot/SnapshotDownloader.cs
```diff
@@ -78,17 +83,40 @@ public async Task DownloadAsync(string url, string destinationPath, Cancellation
...
+    {
+        byte[] buffer = ArrayPool<byte>.Shared.Rent(BufferSize);
+        try
+        {
+            long remaining = bytesToSkip;
+            while (remaining > 0)
+            {
+                int read = await stream.ReadAsync(buffer.AsMemory(0, (int)Math.Min(buffer.Length, remaining)), cancellationToken).ConfigureAwait(false);
+                if (read == 0)
+                    throw new IOException($"Stream ended prematurely while skipping: {remaining} of {bytesToSkip} bytes remaining.");
```
LukaszRozmej: This could be replaced by `ReadAtLeastAsync` with `throwOnEndOfStream: true`

LukaszRozmej (approved):

src/Nethermind/Nethermind.Init.Snapshot/SnapshotDownloader.cs
svlachakis: done

LukaszRozmej (approved):

PR #11406, opened 2026-04-29
Minor fixes from AI audit

alexb5dh:
Fixes for minor issues across different parts of codebase spotted by AI checks.

### Medium
- Constant-time comparison for MAC verification in `FileKeyStore` - during password validation.
- Add missing return in one of the `RlpDecoderExtensions.Encode` - passing empty data could corrupt encoding.
- Use `long` for "JsonRpcBytesSent*" metrics - `int` can overflow with just ~2GB of data.
- Check for empty `c3` in Shutter decoding - it also means encryption is invalid.
- Add missing length and data validation in SSZ `DecodeBool`.
- ~~Properly escape `id` field for streamed RPC requests to prevent injection.~~

### Low
- Handle non-existent storage case in `PatriciaTree.Accept` - prevents potential `IndexOutOfRangeException`.
- Fix validation order for `flashbots_validateBuilderSubmissionV3` - returns correct error on invalid lengths.
- Handle index out of bounds in `TryGetCanonicalTransaction` - prevents potential `NullReferenceException`.
- Handle missing validator info in `ValidatorStore`  - prevents potential `IndexOutOfRangeException`.
- Tests for RLP `ReadPrefixAndContentLength`.
- Regression tests for fixes.

#### Requires testing

- [x] Yes

- [x] Yes


src/Nethermind/Nethermind.Runner/JsonRpc/Startup.cs
```diff
@@ -517,31 +538,18 @@ private static void WriteIdRaw(PipeWriter writer, object? id)
...
-                        buf[len + 1] = (byte)'"';
-                        writer.Advance(len + 2);
-                        break;
-                    }
-                default:
-                    writer.Write("null"u8);
-                    break;
-            }
-        }
+        static void WriteOther(PipeWriter writer, object id) => JsonSerializer.Serialize(writer.AsStream(), id, id.GetType(), EthereumJsonSerializer.JsonOptions);
```
LukaszRozmej: is this significantly slower and allocates?

src/Nethermind/Nethermind.Serialization.Ssz/Ssz.BasicTypes.cs
```diff
@@ -189,7 +187,30 @@ public static void Encode(Span<byte> span, ReadOnlySpan<byte> value)
...
+        {
+            throw new InvalidDataException(
+                $"{nameof(DecodeBool)} expects input of length {expectedLength} and received {span.Length}");
+        }
+
+        return span[0] switch
+        {
+            0 => false,
+            1 => true,
+            var x => throw new InvalidDataException($"SSZ bool must be 0 or 1, got {x}")
```
LukaszRozmej, line 203: Do we have exception type for SSZ?

LukaszRozmej (approved):

svlachakis (approved):

src/Nethermind/Nethermind.Runner/JsonRpc/Startup.cs
alexb5dh: Shouldn't allocate, as is using cached fields and pooled arrays, but is a bit slower with more nested calls.

Can do check like this and going old path if `false`, but doesn't look nice:
```
static unsafe bool NeedsEscaping(ReadOnlySpan<char> value)
{
    fixed (char* ptr = &value.GetPinnableReference())
        return EthereumJsonSerializer.JsonOptions.Encoder!.FindFirstCharacterToEncode(ptr, value.Length) >= 0;
}
```

Similar alternative - checking for `value.IndexOfAnyExcept("0x123456789abcdefABCDEF-_")` to quickly cover most common scenarios.

Worth reverting with very explicit comment about escaping? Or should I bencmark all approaches first?

src/Nethermind/Nethermind.Serialization.Ssz/Ssz.BasicTypes.cs
alexb5dh: `InvalidDataException` is used in other methods. Changing this seems out-of-scope for this PR.

flcl42 (approved):

src/Nethermind/Nethermind.Runner/JsonRpc/Startup.cs
alexb5dh: Rough benchmarks show that even fastest alternative approach can be 2+ times slower on some common strings. So reverting this for now with an explicit comment.

---

image

---

image

svlachakis (approved):

PR #11408, opened 2026-04-29
ci: bump NuGet cache key to evict corrupt Windows entries

LukaszRozmej:
## Changes

- Bumps the `actions/cache@v5` key prefix from `nuget-` to `nuget-v2-` (both `key` and `restore-keys`) in `nethermind-tests.yml`, `nethermind-tests-flat.yml`, and `nethermind-tests-checked.yml`.

## Why

Windows shards have been failing on the `Cache NuGet packages` step with a "Cache hit" log followed by a 2-second silent failure and every downstream test step skipped — e.g. [run 25106465684, job 73568875126](https://github.com/NethermindEth/nethermind/actions/runs/25106465684/job/73568875126?pr=11401). Root cause is [actions/cache#1729](https://github.com/actions/cache/issues/1729): an earlier run that got cancelled mid-save wrote a corrupt TAR under the cached key, and on Windows the TAR extraction fails silently for every subsequent restore against that key.

Re-running the workflow doesn't help — the same poisoned key gets restored every time. Bumping the prefix routes future runs to a fresh storage slot. The first job per OS misses and seeds a clean entry; subsequent jobs hit the new healthy cache. Old `nuget-` entries become unreachable and GitHub will evict them via the 7-day no-access policy.

`restore-keys` is bumped together with `key` so that an exact-key miss can't fall back into the poisoned `nuget-` entry.

## Types of changes

- [x] Build-related changes

#### Requires testing

- [x] No

#### Notes on testing

CI itself exercises this — the first PR run will miss and reseed; subsequent runs will validate the new cache.

- [x] No

- [x] No

### DIFF
--- a/.github/workflows/nethermind-tests-checked.yml
+++ b/.github/workflows/nethermind-tests-checked.yml
@@ -105,8 +105,8 @@ jobs:
         uses: actions/cache@v5
         with:
           path: ~/.nuget/packages
-          key: nuget-${{ runner.os }}-${{ hashFiles('Directory.Packages.props', 'global.json') }}
-          restore-keys: nuget-${{ runner.os }}-
+          key: nuget-v2-${{ runner.os }}-${{ hashFiles('Directory.Packages.props', 'global.json') }}
+          restore-keys: nuget-v2-${{ runner.os }}-
 
       - name: ${{ matrix.project }}
         id: test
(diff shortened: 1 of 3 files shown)


rubo (commented):
Recently, @Marchhill was also having problems with NuGet caching. I recommended purging GitHub caches, and that helped. Renaming has the same effect, just an uglier take. Nevertheless, I don't like this caching approach at all -- will revisit soon.

PR #11411, opened 2026-04-29
fix(fastsync): clean flat storage when account is null in `VerifyStorageUpdated`

grubbyTimes:
`VerifyStorageUpdated` skip flat storage cleanup when an address in `UpdatedStorages` no longer has an account It only branches on `account?.StorageRoot == EmptyTreeHash`; a deleted account (`account is null`) falls through to the "Storage is ok" path and leaves stale flat slots from earlier sync downloads. Treat `account is null` the same as an empty storage root and call `EnsureStorageEmpty`. Adds `RepairDeletedAccount_calls_EnsureStorageEmpty` regression test and broadens the `EnsureStorageEmpty` comment to cover both cases.


LukaszRozmej:
@claude review

batrr (approved):

LukaszRozmej (approved):

grubbyTimes:
not plan to apply AI's suggestion, good to go? @batrr

PR #11424, opened 2026-04-30
fix(state): honour SELFDESTRUCT marker in PerContractState.StorageRoot

stdevMac:
## Changes

- `PerContractState.StorageRoot` now short-circuits to `Keccak.EmptyTreeHash` when the same-block SELFDESTRUCT marker (`BlockChange.HasClear`) is set — mirroring the existing short-circuit in `PerContractState.IsEmpty`. One added line in `Nethermind.State/PersistentStorageProvider.cs`.
- New unit test `SelfDestructPrewarmerStateTests.IsNonZeroAccount_via_IWorldState_default_returns_false_after_destroy_and_value_transfer` reproduces the bug at the state layer (fails on parent, passes with this fix).
- Two supporting tests in `SelfDestructPrewarmerStateTests` and two block-level integration tests in `SelfDestructCreate2SameBlockTests` lock in correctness for the same-block SELFDESTRUCT-then-redeploy pattern with both two-tx and three-tx (with intermediate value transfer) variants.

### Symptom

Sepolia archive node running 1.37.1 deterministically rejects block **4913057** with `HeaderGasUsedMismatch: Expected 9922769, got 9916180` (Δ = 6589 gas, exactly one EIP-684 collision revert with refund), and stops processing further blocks. The failing transaction is the metamorphic-contract redeploy at index 0x3a in the block:

```
to:    0x030a5714a5cb1ea7d98b4b978e1379c2b8805d25  (0age metamorphic factory)
data:  deployMetamorphicContractWithConstructor(salt, initCode)
```

`debug_traceCall` reproduces the failure when run against parent state (i.e. without applying tx 0x2d's SELFDESTRUCT first), proving the EVM is correct in isolation and that block-level state propagation between the destroy and the redeploy is broken.

### Root cause

The EVM CREATE collision check is `state.IsNonZeroAccount(X)`. `ParallelWorldState` (the production decorator over `WorldState`, registered at `BlockProcessingModule.cs:56-57`) does not override `IsNonZeroAccount`, so the **default interface method on `IWorldState`** runs. That default calls `IsStorageEmpty(X)`, also unoverridden in `ParallelWorldState`, so it routes to the **default in `IAccountStateProvider`**:

```csharp
TryGetAccount(address, out AccountStruct account);
return account.IsStorageEmpty;
```

`TryGetAccount` lands on `WorldState.IAccountStateProvider.TryGetAccount`, which builds the `AccountStruct` via:

```csharp
account = _stateProvider.GetAccount(address)
    .WithChangedStorageRoot(_persistentStorageProvider.GetStorageRoot(address))
    .ToStruct();
```

`PerContractState.StorageRoot` calls `EnsureStorageTree` → `CreateStorageTree` → `_provider._currentScope.CreateStorageTree(_address)` → `LookupStorageTree` → `Get(X)?.StorageRoot`. `Get(X)` is `TrieStoreWorldStateBackendScope.Get`, which serves from `_loadedAccounts`. That cache was populated when the EVM first read X earlier in the block (e.g. to load X's code for tx 0x2d's SELFDESTRUCT call), and it holds the parent (pre-destroy) account with a **non-empty** `StorageRoot`.

Result: `GetStorageRoot` returns the parent's non-empty `StorageRoot`, `WithChangedStorageRoot` stuffs it into the `AccountStruct`, `account.IsStorageEmpty` is `_storageRoot == EmptyTreeHash` so returns false, `IsNonZeroAccount` short-circuits to true, the inner CREATE inside the metamorphic transient hits a phantom EIP-684/EIP-7610 collision, the factory's `require(deployed != 0)` reverts, and the block ends with `HeaderGasUsedMismatch`.

`PerContractState.IsEmpty` already short-circuits on `BlockChange.HasClear`. `PerContractState.StorageRoot` did not. This PR adds the same short-circuit, so any consumer (including the `IAccountStateProvider.TryGetAccount` path) sees `EmptyTreeHash` while the SELFDESTRUCT marker is in effect.

The fix is a one-line short-circuit at the boundary that matches the existing `IsEmpty` invariant; no behavioural change for code paths that don't go through `TryGetAccount`/`StorageRoot`.

## Types of changes

#### Requires testing

- [x] Yes

- [x] Yes

#### Notes on testing

- New regression test: `SelfDestructPrewarmerStateTests.IsNonZeroAccount_via_IWorldState_default_returns_false_after_destroy_and_value_transfer` — exercises `ParallelWorldState` wrapping `WorldState`, primes `_loadedAccounts` the way the EVM does, then performs the collision check via the `IWorldState` default interface method. **Fails on the parent commit and passes with this fix.**
- Supporting tests: `Main_scope_sees_destroyed_account_even_with_prewarmed_cache` and `Main_scope_sees_recreated_account_as_collision_free_after_value_transfer` pin cache-layer correctness.
- Block-level: `SelfDestructCreate2SameBlockTests` (two test cases) drives the full `BlockProcessor` + `BranchProcessor` + prewarming flow through `BasicTestBlockchain` for both two-tx and three-tx (with intermediate value transfer) variants.

Existing suites still pass with the fix:

| Suite | Total | Failed |
|---|---|---|
| `Nethermind.State.Test` | 707 | 0 |
| `Nethermind.Blockchain.Test` | 1426 | 0 |
| `Nethermind.Evm.Test` (SELFDESTRUCT/CREATE/Eip7708 selectors) | 32 | 0 |

#### Verification on an affected node

Confirmed on a live Sepolia archive node running 1.37.1 + this fix: block 4913057 validates and sync resumes from there.

- [x] No

- [x] Yes

> Fixes a determinism bug that caused Sepolia archive nodes to reject block 4913057 with `HeaderGasUsedMismatch` when a contract was SELFDESTRUCTed and then redeployed at the same address (via the 0age metamorphic-contract pattern) within the same block. `PerContractState.StorageRoot` now correctly returns `EmptyTreeHash` while a same-block SELFDESTRUCT marker is in effect, so the EIP-684/EIP-7610 CREATE collision check sees the destroyed account as empty and lets the redeploy proceed.

## Remarks

The same defect class is likely worth checking on `master`, although `Nethermind.State` has been substantially restructured there (cross-block caches, sharded `NodeStorageCache`, etc.) — the symbol and call path may be different. Worth a quick verification but out of scope for this 1.37.1 backport.

### DIFF
--- a/src/Nethermind/Nethermind.Blockchain.Test/SelfDestructCreate2SameBlockTests.cs
+++ b/src/Nethermind/Nethermind.Blockchain.Test/SelfDestructCreate2SameBlockTests.cs
@@ -1,0 +1,196 @@
+// SPDX-FileCopyrightText: 2026 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System.Threading.Tasks;
+using FluentAssertions;
+using Nethermind.Core;
+using Nethermind.Core.Crypto;
+using Nethermind.Core.Specs;
+using Nethermind.Core.Test.Blockchain;
+using Nethermind.Core.Test.Builders;
+using Nethermind.Crypto;
+using Nethermind.Evm;
+using Nethermind.Int256;
+using Nethermind.Specs;
+using Nethermind.Specs.Forks;
+using NUnit.Framework;
+using Nethermind.Core.Container;
+
+namespace Nethermind.Blockchain.Test;
+
+/// <summary>
+/// Regression tests for the metamorphic-contract bug observed on Sepolia block 4913057
+/// (Shanghai-era, archive node, prewarming enabled).
+///
+/// In that block, transaction 45 (sender 0x0c27…) called SELFDESTRUCT on contract
+/// X (= 0x140da4…9372). Transaction 57 (sender 0xf93bab…) sent 0.001 ETH to X,
+/// creating an empty account at the destroyed address. Transaction 58 (same sender
+/// 0xf93bab…) called the metamorphic factory at 0x030a571…d25 to redeploy at X
+/// via CREATE2 → transient → CREATE → X.
+///
+/// On Nethermind 1.37.1 the redeploy reverts: CREATE inside the transient sees X
+/// as still having code and returns address(0), so the factory's
+/// `require(deployed != 0)` fails. The same tx executed via debug_traceCall against
+/// the parent state (i.e. without applying the SELFDESTRUCT) reproduces the
+/// identical wrong gas usage (0x17a758) and revert message, which means the
+/// SELFDESTRUCT effect from tx 45 is not visible to tx 58 during real block
+/// processing.
+///
+/// These tests reproduce the scenario at the BlockProcessor level (block production
+/// + validation, prewarming enabled by default) on TestBlockchain.
+/// </summary>
+[Parallelizable(ParallelScope.All)]
+public class SelfDestructCreate2SameBlockTests
+{
+    /// <summary>
+    /// Two-tx variant: SELFDESTRUCT then CREATE2 redeploy, different senders.
+    /// </summary>
+    [Test]
+    public async Task Redeploy_via_CREATE2_succeeds_when_destroy_and_redeploy_have_different_senders()
+    {
+        await RunRedeployScenario(includeIntermediateValueTransfer: false);
+    }
+
+    /// <summary>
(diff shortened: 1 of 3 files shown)


LukaszRozmej:
@claude review

LukaszRozmej (approved):

LukaszRozmej:
Although maybe better fix is to just not register ParallelWorldState? @claude

asdacap (approved):

LukaszRozmej:
@claude we don't need BAL tracking in the release. In master this is already fixed in different way. We don't need BAL's here at all

PR #11425, opened 2026-04-30
debug_traceCallMany: materialize simple call traces

BZO95:
## Changes

- Materialize simple-path `debug_traceCallMany` bundle traces before the request timeout source is disposed.
- Add regression coverage for enumerating traces while the cancellation token is still usable.

#### What types of changes does your code introduce?

- [x] Bugfix (a non-breaking change that fixes an issue)
- [ ] New feature (a non-breaking change that adds functionality)
- [ ] Breaking change (a change that causes existing functionality not to work as expected)
- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

#### If yes, did you write tests?

- [x] Yes
- [ ] No

#### Notes on testing

- Added `Debug_traceCallMany_materializes_simple_path_before_timeout_source_is_disposed`.
- Ran `git diff --check`.
- `.NET build/test` intentionally skipped by user request.

#### Requires documentation update

- [ ] Yes
- [x] No

#### Requires explanation in Release Notes

- [ ] Yes
- [x] No

PR #11429, opened 2026-04-30
debug_traceCallMany: extend timeout CTS lifetime to enumerator disposal

LukaszRozmej:
Replaces #11425

## Changes

- Bind the timeout `CancellationTokenSource` lifetime to enumerator disposal in `DebugRpcModule.TraceCallMany`'s simple path. The previous `using` scope disposed the CTS before the lazy bundle pipeline was enumerated, leaving the cancellation token unusable to downstream consumers (e.g. `WaitHandle` access throws `ObjectDisposedException`).
- Mirrors the iterator-owns-CTS pattern already used by `EthRpcModule.GetLogs`. Streaming is preserved — no per-bundle or per-trace buffering is introduced.
- Add two regression tests:
  - `Debug_traceCallMany_does_not_dispose_timeout_cts_before_consumer_enumerates` fails on the unfixed code with `ObjectDisposedException` and passes with the fix.
  - `Debug_traceCallMany_does_not_eagerly_materialize_bundles` guards against a regression to eager materialization.

Replaces #11425, which fixed the same bug by materializing every bundle/trace into nested `List<>`/`T[]` and giving up the zero-allocation streaming pipeline that was a deliberate design choice (see https://github.com/NethermindEth/nethermind/pull/11425#issuecomment-4350364618).

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

- All 22 `DebugModuleTests` pass with the fix.
- Verified the new disposed-CTS regression test fails on master without the production fix (`ObjectDisposedException` from `WaitHandle`) and passes with it.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.JsonRpc.Test/Modules/DebugModuleTests.cs
+++ b/src/Nethermind/Nethermind.JsonRpc.Test/Modules/DebugModuleTests.cs
@@ -43,15 +43,107 @@ public class DebugModuleTests
     private readonly IBlockchainBridge _blockchainBridge = Substitute.For<IBlockchainBridge>();
     private readonly MemDb _blocksDb = new();
 
-    private DebugRpcModule CreateDebugRpcModule(IDebugBridge customDebugBridge) => new(
+    private DebugRpcModule CreateDebugRpcModule(
+        IDebugBridge customDebugBridge,
+        IBlockchainBridge? blockchainBridge = null,
+        IBlockFinder? blockFinder = null) => new(
             LimboLogs.Instance,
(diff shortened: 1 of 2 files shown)


LukaszRozmej:
@claude re-review

svlachakis (approved):

DarkLord017 (approved):

PR #11431, opened 2026-04-30
fix(eth_createAccessList): preserve reverted-frame accesses and exclude precompiles

### LINKED ISSUE #11209: `eth_createAccessList`: incomplete or empty access list returned
**Affected tests:** test_08, test_09, test_12, test_15, test_16, test_17

### Description

Nethermind returns fewer `accessList` entries than Geth for the same
transaction, and for some tests returns a completely empty list. This is
most severe for pre-Berlin blocks (block < 12,244,000, where EIP-2929 was
not yet active): in all five pre-Berlin test cases Nethermind returns either
an empty list or misses the majority of accessed contracts. One post-Berlin
test (test_17, block 0xe4e1bf) also misses one address entry.

The `gasUsed` values also differ in all affected tests, consistent with
Nethermind not charging warm/cold access costs for the missed addresses and
storage keys.

#### test_08 — block 0xB71B00 (11,998,976), ERC20 transfer

**Expected (Geth):** 2 entries, 6 storage keys

```json
{
  "accessList": [
    {
      "address": "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48",
      "storageKeys": [
        "0x0000000000000000000000000000000000000000000000000000000000000001",
        "0x10d6a54a4754c8869d6886b5f5d7fbfa5b4522237ea5c60d11bc4e7a1ff9390b",
        "0x68f621fe39c397cc54aed6af74827e90c78ae48ff5e52728bb90ff2265002db6",
        "0x7050c9e0f4ca769c69bd3a8ef740bc37934f8e2c036e5a723fd8ee048ed3f8c3",
        "0x755c9d80839f4f8436c39a6b618087a1acdb2bfdd0c673d71d0b5563b539aeda",
        "0xd6d3c01425af47fc283b281db38f2ef429f68e7bd49f0c4700d84377a3e00c66"
      ]
    },
    { "address": "0xb7277a6e95992041568d9391d09d0122023778a2", "storageKeys": [] }
  ],
  "gasUsed": "0xb58b"
}
```

**Actual (Nethermind):**

```json
{ "accessList": [], "gasUsed": "0x7643" }
```

#### test_15 — block 0x11573 (71,027), complex call

**Expected (Geth):** 23 contracts in accessList, error "out of gas: out of gas"

**Actual (Nethermind):**

```json
{ "accessList": [], "error": "out of gas", "gasUsed": "0x7a120" }
```

#### test_17 — block 0xe4e1bf (14,999,999), Uniswap-style swap

Nethermind returns 3 entries; Geth returns 4. The missing entry is address
`0x829bd824b016326a401d083b33d092293333a830` with no storage keys — it is
`CALL`ed but Nethermind does not include it in the access list. The storage
key contents for the other 3 entries match, but the order differs.

```
Geth[1]: 0x829bd824..., 0 keys  ← missing in Nethermind
Geth[2]: 0xaaaebe6f..., 3 keys
Geth[3]: 0xc02aaa39..., 2 keys
```

### Notes

The concentration of empty access lists in pre-Berlin blocks (tests 08, 09,
12, 15, 16) suggests Nethermind's access list tracking is broken or disabled
for blocks before the Berlin hard fork. Post-Berlin, the tracking mostly
works but still misses address-only entries (addresses `CALL`ed but with no
storage reads/writes).

manusw7:
Fixes #11209

## Changes

- `StackAccessTracker`: added `isTracingAccess` flag; when set, `Restore()` skips rolling back `AccessedAddresses`/`AccessedStorageCells` on sub-frame revert — so all addresses and storage keys touched inside a reverted frame survive to the final access list report.
- `BlockchainBridge.BuildAddressesToOptimize`: includes all active precompile addresses for the given spec, so precompiles with no storage keys are excluded from the returned access list.
- Regression tests for:
  - Reverted outer frame: call target address preserved in access list.
  - Reverted sub-frame: storage keys accessed inside the reverted frame preserved.
  - Nested reverted frames: both committed and inner-reverted call targets captured.
  - Precompile filtering through `BlockchainBridge.CreateAccessList` (with and without storage keys).

## Types of changes

#### Requires testing

- [x] Yes

- [x] Yes

#### Notes on testing

- `src/Nethermind/Nethermind.Evm.Test/Tracing/AccessTxTracerTests.cs`
- `src/Nethermind/Nethermind.Facade.Test/BlockchainBridgeTests.cs`

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

Fixed `eth_createAccessList` to return complete access lists: addresses and storage cells first touched inside reverted sub-frames are now preserved, and active precompile addresses with no storage keys are excluded from the result.


manusw7:
@claude review again

src/Nethermind/Nethermind.Evm.Test/Tracing/AccessTxTracerTests.cs
```diff
@@ -119,4 +121,163 @@ public void ReportAccess_AddressAIsSetToOptimizedAndHasStorageCell_AccessListHas
             return (tracer, block, transaction);
         }
     }
+
+    [TestFixture]
+    public class AccessTxTracerPreBerlinTests : VirtualMachineTestsBase
```
LukaszRozmej: Move to seperate file.
Maybe can be combined to some testcases?

LukaszRozmej (changes requested):
IMO this feels off:
1. eth_createAccessList for pre-berlin blocks doesn't have much sense. If there are tests for it we should consider convincing EF to just drop them.
2. Having separate modes just for that sounds overkill
3. This PR probably has negative performance effects

IMO not worth it at all

svlachakis:
> IMO this feels off:
> 
> 1. eth_createAccessList for pre-berlin blocks doesn't have much sense. If there are tests for it we should consider convincing EF to just drop them.
> 2. Having separate modes just for that sounds overkill
> 3. This PR probably has negative performance effects
> 
> IMO not worth it at all

A few thoughts @LukaszRozmej @manusw7:

On pre-Berlin: You're right, mostly pointless. And to be precise - these are Erigon tests, not EF. Happy to drop the pre-Berlin scope.

On what's left after dropping pre-Berlin: test_17 (block 14,999,999) still fails - Nethermind misses an address that was CALL'd but had no storage reads. The PR's diagnosis is that this is a reverted sub-frame issue (accesses get journal-rolled-back when an inner frame reverts, so `ReportAccess` sees only the survivors). If that diagnosis holds, it's a real post-Berlin bug worth fixing. If it doesn't, the whole PR may indeed not be worth it - @manusw7 please confirm the root cause before pushing further.

On separate modes being overkill: Agreed. `StackAccessTracker` is for gas tracking, not tracing. If we keep the fix, I think a cleaner place is on `AccessTxTracer` itself.

On performance: With that refactor, the cost goes away - no extra `HashSets` per tx, no extra clears, just an `IsTracingAccess` check that's false in normal execution.

@manusw7 I think as next steps drop the pre-Berlin scope, verify test_17's root cause is actually reverted-frame loss (not something else), and if so, move the tracking onto `AccessTxTracer`. If test_17 turns out to be a different bug, we can reassess whether the PR is worth it. @LukaszRozmej Sound good?

LukaszRozmej:
Potential simpler thing then it to call ReportAccess multiple times before reverts? Or don't revert this one?

manusw7:
@svlachakis @LukaszRozmej thx for the feedback! 

Totally agree on dropping the pre-Berlin scope. I'll pull that out.

> On what's left after dropping pre-Berlin: test_17 (block 14,999,999) still fails - Nethermind misses an address that was CALL'd but had no storage reads. The PR's diagnosis is that this is a reverted sub-frame issue (accesses get journal-rolled-back when an inner frame reverts, so ReportAccess sees only the survivors). If that diagnosis holds, it's a real post-Berlin bug worth fixing. If it doesn't, the whole PR may indeed not be worth it - @manusw7 please confirm the root cause before pushing further.

This is exactly the problem I ran into. I'm going to double-check it for absolute accuracy before pushing further, but that reverted sub-frame data loss is definitely the root cause.

> Potential simpler thing then it to call ReportAccess multiple times before reverts? Or don't revert this one?

I'm moving this PR to draft for now. I'm going to step back and look into these simpler ideas to see if we can fix the bug without the performance hit.

manusw7:
@claude review again. Also note that change request was about PreBerlin tests (which were dropped), rn `AccessTxTracerTests` only have 3 new tests in the came class regarding reverted calls

src/Nethermind/Nethermind.Evm.Test/Tracing/AccessTxTracerTests.cs
```diff
@@ -109,6 +111,100 @@ public void ReportAccess_AddressAIsSetToOptimizedAndHasStorageCell_AccessListHas
             Assert.That(sut.AccessList.Select(static x => x.StorageKeys), Has.Exactly(1).Contains(new UInt256(1)));
         }
 
+
+        [Test]
```
LukaszRozmej, line 114: Can test be de-duplicated?

src/Nethermind/Nethermind.Facade/BlockchainBridge.cs
```diff
@@ -275,14 +276,31 @@ private CallOutput ConvergeAccessList(BlockProcessingComponents components, Bloc
 
         private Address[] BuildAddressesToOptimize(BlockHeader header, Transaction tx, bool optimize)
```
LukaszRozmej: Maybe we could do ReadOnlySpan - preallocated in parent instead of array?

LukaszRozmej (approved):
Check if tests can be deduplicated more

svlachakis (approved):
LGTM, please check @LukaszRozmej's comments

svlachakis (approved):

LukaszRozmej (approved):

LukaszRozmej (approved):

benaadams (approved):

PR #11432, opened 2026-04-30
fix(prewarm): run speculative prewarming when BAL EIP is not active

### LINKED ISSUE #11427: BlockCachePreWarmer disabled on all non-EIP-7928 chains after parallel-execution change
## Summary

Commit 54ecbb8c35 ("Parallel transaction execution with BALs", #9182) inadvertently disables speculative block-cache prewarming on every chain that has not activated EIP-7928 (Block-Level Access Lists). Since EIP-7928 is not yet active on mainnet or any of the public testnets, this means the prewarmer is effectively off everywhere in production today.

## Details

In [`src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs`](https://github.com/NethermindEth/nethermind/blob/master/src/Nethermind/Nethermind.Consensus/Processing/BlockCachePreWarmer.cs#L107-L113):

```csharp
private bool ShouldPreWarm(IReleaseSpec spec)
    => !_parallelExecutionEnabled || IsBalReadWarmingEnabled(spec);

public bool IsBalReadWarmingEnabled(IReleaseSpec spec)
    => _parallelExecutionBatchRead && spec.BlockLevelAccessListsEnabled;
```

Combined with the new defaults in [`Nethermind.Config/BlocksConfig.cs`](https://github.com/NethermindEth/nethermind/blob/master/src/Nethermind/Nethermind.Config/BlocksConfig.cs#L71-L72):

```csharp
public bool ParallelExecution { get; set; } = true;
public bool ParallelExecutionBatchRead { get; set; } = true;
```

…and `BlockLevelAccessListsEnabled => IsEip7928Enabled` (off on mainnet, all testnets, and every benchmark today), `ShouldPreWarm` returns `false` and `BlockCachePreWarmer.PreWarmCaches` short-circuits.

## Impact

- The flat `TrieWarmer` itself is unaffected — it's queued from `HintGet` which is triggered by `IWorldState.Get`. But the speculative tx execution performed by the prewarmer is what populates those `Get` calls. With the speculative path gone, the warmer's queue stays nearly empty.
- Result: a meaningful regression in cache warmup on mainnet and every other non-EIP-7928 chain.

## Suggested fix

The `ShouldPreWarm` carve-out probably wants to keep speculative warming alive when parallel execution is enabled but BAL isn't (or gate it behind an explicit opt-out flag). Currently it treats `ParallelExecution=true` as \"BAL must be on, otherwise no warming at all,\" which is almost certainly not the intent.

---

_Filed by Claude (claude-opus-4-7) on behalf of @asdacap._

LukaszRozmej:
Fixes #11427

## Summary

- `BlockCachePreWarmer.ShouldPreWarm` skipped speculative prewarming whenever `ParallelExecution=true` (the default), but `BlockAccessListManager` only runs parallel execution when the block actually has a `BlockAccessList`. On pre-BAL forks neither path ran, silently regressing block-processing performance from pre-BAL releases.
- Treat specs without `BlockLevelAccessListsEnabled` the same as `ParallelExecution=off` so speculative prewarming runs on those blocks. BAL-active specs are unchanged.
- Updated `PreWarmCaches_GateLogic` test cases for the two ``ParallelExec ON, BALs OFF`` rows that previously expected ``skipped``.

| ParallelExec | BAL spec | BatchRead | Behavior |
|---|---|---|---|
| ON | ON | ON | BAL read warming |
| ON | ON | OFF | skipped (parallel exec runs) |
| ON | OFF | * | **speculative (was: skipped)** |
| OFF | * | * | speculative / BAL warming |

## Types of changes

## Testing

- [x] ``dotnet test --project src/Nethermind/Nethermind.Consensus.Test/... --filter ~BlockCachePreWarmerTests`` — 12/12 pass.

## Documentation

- [ ] Yes
- [x] No

### DIFF
--- a/src/Nethermind/Nethermind.Consensus.Test/BlockCachePreWarmerTests.cs
+++ b/src/Nethermind/Nethermind.Consensus.Test/BlockCachePreWarmerTests.cs
@@ -247,13 +247,14 @@ public class BlockCachePreWarmerTests
     }
 
     /// <summary>
-    /// Verifies the prewarmer gate logic: ParallelExecution ON disables prewarming unless
-    /// ParallelExecutionBatchRead is ON and BALs are available.
+    /// Verifies the prewarmer gate logic: ParallelExecution ON skips speculative prewarming
+    /// only when BAL is active for the spec (so parallel execution can actually run); when
+    /// BAL is not active, speculative prewarming runs regardless of ParallelExecution.
     /// </summary>
     [TestCase(true, true, true, true, TestName = "ParallelExec ON, BALs ON, BatchRead ON => BAL warming")]
     [TestCase(true, true, false, false, TestName = "ParallelExec ON, BALs ON, BatchRead OFF => skipped")]
-    [TestCase(true, false, true, false, TestName = "ParallelExec ON, BALs OFF, BatchRead ON => skipped")]
-    [TestCase(true, false, false, false, TestName = "ParallelExec ON, BALs OFF, BatchRead OFF => skipped")]
+    [TestCase(true, false, true, true, TestName = "ParallelExec ON, BALs OFF, BatchRead ON => speculative")]
+    [TestCase(true, false, false, true, TestName = "ParallelExec ON, BALs OFF, BatchRead OFF => speculative")]
     [TestCase(false, true, true, true, TestName = "ParallelExec OFF, BALs ON, BatchRead ON => BAL warming")]
     [TestCase(false, false, false, true, TestName = "ParallelExec OFF, BALs OFF, BatchRead OFF => speculative")]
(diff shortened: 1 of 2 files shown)


flcl42 (approved):

LukaszRozmej:
@claude what about `IBlocksConfig.PreWarmStateOnBlockProcessing` flag?

LukaszRozmej:
@claude so when `PreWarmStateOnBlockProcessing` is `false` but `ParallelExecutionBatchRead` is `true` the `ParallelExecutionBatchRead` will be ignored?

Not sure if good dependency @benaadams

svlachakis (approved):

benaadams:
@claude what happens when ParallelExecutionBatchRead is on but ParallelExecution is off?

PR #11438, opened 2026-04-30
feat(rpc): add eth_capabilities method for routing capability discovery

### LINKED ISSUE #11260: Add `eth_capabilities` support
`eth_capabilities` implementation as per https://github.com/ethereum/execution-apis/pull/755

divi2806:
Closes #11260

## Changes

- Add `eth_capabilities` RPC method to `IEthRpcModule` and `EthRpcModule` per [ethereum/execution-apis#755](https://github.com/ethereum/execution-apis/pull/755)
- Add `EthCapabilitiesResult` response model with `CapabilityHead`, `CapabilityResource`, and `CapabilityDeleteStrategy` types
- Inject `ISyncConfig` and `IPruningConfig` as optional constructor parameters into `EthRpcModule` and `EthModuleFactory` to derive historical data availability
- Response includes current head block (`number` + `hash`) and six resource descriptors: `state`, `tx`, `logs`, `receipts`, `blocks`, `stateproofs`
- `oldestBlock` and `deleteStrategy` are omitted (null) for disabled resources and archive nodes; `deleteStrategy` only appears for in-memory pruned nodes (`type: "window"`)
- `stateproofs` is disabled on pruned nodes since `eth_getProof` requires full trie history

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

4 tests added in `EthRpcModuleTests.Capabilities.cs`:
- `eth_capabilities_returns_all_resources_and_head` — verifies all 6 resource fields and head block info are present
- `eth_capabilities_archive_node_has_stateproofs_enabled_and_no_delete_strategy` — archive node (`PruningMode.None`) serves state proofs, has no window delete strategy
- `eth_capabilities_pruned_node_disables_stateproofs_and_reports_window` — memory-pruned node disables `stateproofs`, reports `deleteStrategy.type = "window"` with correct `retentionBlocks`
- `eth_capabilities_no_receipts_disables_tx_logs_receipts` — when `DownloadReceiptsInFastSync = false`, `tx`, `logs`, and `receipts` are disabled with no `oldestBlock`

All 5 tests pass (`total: 5 | failed: 0 | succeeded: 5`).

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

Add `eth_capabilities` JSON-RPC method. Returns the node's current head block and historical data availability for `state`, `tx`, `logs`, `receipts`, `blocks`, and `stateproofs` resources. Enables RPC routers and load balancers to make routing decisions without active probing. Implements [ethereum/execution-apis#755](https://github.com/ethereum/execution-apis/pull/755).


LukaszRozmej:
@claude review

LukaszRozmej:
@claude re-review

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
```diff
@@ -898,6 +903,82 @@ public ResultWrapper<JsonNode> eth_config()
             ResultWrapper<BlockAccessList?>.Fail("Cannot return pruned historical block access list.", ErrorCodes.PrunedHistoryUnavailable)
             : ResultWrapper<BlockAccessList?>.Success(bal);
     }
+
+    public ResultWrapper<EthCapabilitiesResult> eth_capabilities()
```
LukaszRozmej: A bit too long - break it down? Separate class?

src/Nethermind/Nethermind.JsonRpc.Test/Modules/Eth/EthRpcModuleTests.Capabilities.cs
```diff
@@ -0,0 +1,213 @@
...
+using Nethermind.Db;
+using Nethermind.Facade.Eth;
+using Nethermind.JsonRpc.Modules.Eth;
+using Nethermind.Serialization.Json;
+using NSubstitute;
+using NUnit.Framework;
+
+namespace Nethermind.JsonRpc.Test.Modules.Eth;
+
+public partial class EthRpcModuleTests
```
LukaszRozmej, line 23: Check if test can be compacted/de-duplicated

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs
divi2806: Done — extracted all the capability calculation logic into a dedicated `EthCapabilitiesProvider` class ([`EthCapabilitiesProvider.cs`](src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthCapabilitiesProvider.cs)). `eth_capabilities()` in `EthRpcModule` is now a single-line delegation to a cached provider instance.

src/Nethermind/Nethermind.JsonRpc.Test/Modules/Eth/EthRpcModuleTests.Capabilities.cs
divi2806: Done — added a `GetCaps(SyncConfig, PruningConfig)` helper that owns the chain-setup boilerplate, cutting each config-specific test from ~15 lines to ~5. Also merged the two archive-node tests (they used identical config) into a single `eth_capabilities_archive_node_full_availability` test, reducing the total from 8 tests to 7.

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthCapabilitiesResult.cs
```diff
@@ -0,0 +1,75 @@
...
+{
+    /// <summary><c>true</c> when the resource is completely unavailable on this node.</summary>
+    public required bool Disabled { get; init; }
+
+    /// <summary>
+    /// Hex-encoded number of the earliest block for which this resource is available.
+    /// Omitted when <see cref="Disabled"/> is <c>true</c>.
+    /// </summary>
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public string? OldestBlock { get; init; }
```
LukaszRozmej: don't format in the method, keep it as `long`. Json serialization will format it without allocating strings.

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthCapabilitiesResult.cs
```diff
@@ -0,0 +1,75 @@
...
+}
+
+/// <summary>Head block number and hash.</summary>
+public class CapabilityHead
+{
+    /// <summary>Hex-encoded block number.</summary>
+    public required string Number { get; init; }
+
+    /// <summary>Block hash.</summary>
+    public required string Hash { get; init; }
```
LukaszRozmej: keep it as Hash256

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthCapabilitiesResult.cs
```diff
@@ -0,0 +1,75 @@
...
+
+    /// <summary>Proof / trie-node availability (eth_getProof depth window).</summary>
+    public required CapabilityResource Stateproofs { get; init; }
+}
+
+/// <summary>Head block number and hash.</summary>
+public class CapabilityHead
+{
+    /// <summary>Hex-encoded block number.</summary>
+    public required string Number { get; init; }
```
LukaszRozmej: keep it a slong

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthCapabilitiesResult.cs
```diff
@@ -0,0 +1,75 @@
...
+
+    /// <summary>Block hash.</summary>
+    public required string Hash { get; init; }
+}
+
+/// <summary>Availability descriptor for one historical data resource.</summary>
+public class CapabilityResource
```
LukaszRozmej: readonly record struct?

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthCapabilitiesResult.cs
```diff
@@ -0,0 +1,75 @@
...
+    /// <summary>
+    /// Retention / deletion strategy. Present only when data is pruned with a rolling window.
+    /// Omitted for archive nodes or disabled resources.
+    /// </summary>
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public CapabilityDeleteStrategy? DeleteStrategy { get; init; }
+}
+
+/// <summary>Rolling-window deletion strategy for a resource.</summary>
+public class CapabilityDeleteStrategy
```
LukaszRozmej: readonly record struct

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthCapabilitiesResult.cs
```diff
@@ -0,0 +1,75 @@
...
+
+using System.Text.Json.Serialization;
+
+namespace Nethermind.JsonRpc.Modules.Eth;
+
+/// <summary>
+/// Response for eth_capabilities — describes the head block and the historical data availability of each resource type.
+/// Follows ethereum/execution-apis#755.
+/// </summary>
+public class EthCapabilitiesResult
```
LukaszRozmej: readonly record struct? Or record class

LukaszRozmej (commented):
one last thing - cleanup json type

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthCapabilitiesResult.cs
LukaszRozmej: @claude from performance perspective, better to make it struct or keeping this one class?

LukaszRozmej:
@claude which ones to make structs for optimal performance?

divi2806:
> **Claude finished @LukaszRozmej's task in 1m 25s** —— [View job](https://github.com/NethermindEth/nethermind/actions/runs/25211489660)
> 
> ### Struct vs Class Analysis for `EthCapabilitiesResult` types
> * [x]  Read `EthCapabilitiesResult.cs`
> * [x]  Read `EthCapabilitiesProvider.cs`
> * [x]  Analyze performance tradeoffs
> * [x]  Post answer
> 
> ### Answer
> **Rule of thumb**: a struct saves a heap allocation only when it's stored as a field in a class (it lives inline) or in a `Nullable<T>` slot. A struct that itself contains 6+ struct fields becomes too large to pass by value safely (guideline: keep structs ≤ 16–24 bytes).
> 
> Type	Recommendation	Reasoning
> `EthCapabilitiesResult`	**`record class`**	If the nested types become structs, each `CapabilityResource` is ~40 bytes × 6 = ~240 bytes inline. A 256-byte struct is too large — copying it is worse than a single class allocation. It's the top-level container; one heap allocation is fine.
> `CapabilityHead`	**`readonly record struct`**	Two string references = ~16 bytes. Stored inline in `EthCapabilitiesResult`, saves 1 heap allocation per call. Ideal struct size.
> `CapabilityResource`	**`readonly record struct`**	Used 6× per result — saves 6 heap allocations per call. Bool + string ref + `Nullable<CapabilityDeleteStrategy>` lands at ~32–40 bytes, acceptable when stored as fields in a class (not passed around by value).
> `CapabilityDeleteStrategy`	**`readonly record struct`**	String ref + long = ~16 bytes. Used as `CapabilityDeleteStrategy?` → becomes `Nullable<CapabilityDeleteStrategy>` inline in `CapabilityResource` — no heap allocation for the strategy object.
> **Net effect on allocations per `eth_capabilities` call:**
> 
> * Before: result + head + 6× resource + 0–1× deleteStrategy = **8–9 allocations**
> * After structs on all nested types: result only = **1 allocation**
> 
> **One caveat on `CapabilityDeleteStrategy?`:** When `CapabilityDeleteStrategy` is a struct, `DeleteStrategy` becomes `Nullable<CapabilityDeleteStrategy>`. `[JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]` on a nullable value type works correctly in `System.Text.Json` — it omits the property when the nullable has no value.
> 
> **Separate from the struct question** — LukaszRozmej's inline comments also ask to keep `Number` as `long` and `Hash` as `Hash256` (let the serializer format them, avoiding intermediate `string` allocations). That applies on top of these struct changes.

Changed to `record class` based on Claude Bot's analysis — the nested types (`CapabilityHead`, `CapabilityResource`, `CapabilityDeleteStrategy`) are all `readonly record struct`, so `EthCapabilitiesResult` contains ~240 bytes of inline struct data. Making the top-level container a struct would cause large copies on every pass-by-value. Keeping it as a `record class` gives 1 heap allocation instead of 8–9.

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthCapabilitiesResult.cs
divi2806: Done — `OldestBlock` is now `long?`. Removed all `.ToHexString()` calls from `EthCapabilitiesProvider`; `NullableLongConverter` (globally registered in `EthereumJsonSerializer`) handles hex encoding on the wire.

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthCapabilitiesResult.cs
divi2806: Done — `Hash` is now `Hash256`. Provider sets `head?.Hash ?? Keccak.Zero`; `Hash256Converter` handles serialization.

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthCapabilitiesResult.cs
divi2806: Done — `Number` is now `long`. Provider sets `head?.Number ?? 0`; `LongConverter` handles hex encoding.

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthCapabilitiesProvider.cs
```diff
@@ -0,0 +1,76 @@
...
+        bool receiptsSynced = syncConfig?.DownloadReceiptsInFastSync ?? true;
+        // AncientReceiptsBarrierCalc returns Math.Max(1, …) which is wrong for archive nodes
+        // (PivotNumber = 0). When there was no fast sync, receipts are available from genesis.
+        long oldestReceipts = (syncConfig?.PivotNumber ?? 0) == 0 ? 0 : syncConfig!.AncientReceiptsBarrierCalc;
+
+        PruningMode statePruningMode = pruningConfig?.Mode ?? PruningMode.None;
+        bool isArchive = statePruningMode == PruningMode.None;
+        // Memory (and Hybrid) pruning maintains a rolling window of PruningBoundary recent states.
+        // Full-only pruning is periodic and non-linear — we cannot claim a predictable oldest block.
+        long? stateRetention = statePruningMode.IsMemory() ? pruningConfig!.PruningBoundary : null;
```
asdacap: Wire the state related report behind `IWorldStateManager`. Pruning config will not work with flat.

asdacap (changes requested):

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthCapabilitiesProvider.cs
LukaszRozmej: Can you be more specific what in `IWorldStateManager`?

asdacap: Anything is fine as long as flat can specify different behaviour later.

divi2806: I can take a stab at this — add a `GetStateCapability()` method to `IWorldStateManager`, implement it in `WorldStateManager` using the existing pruning config logic, and return archive defaults in `FlatWorldStateManager` so flat can override later. Let me know if that direction works or if you have a preferred approach.
Would love the feedback on the current direction?

LukaszRozmej:
@claude review

LukaszRozmej:
@claude review

svlachakis:
@claude what do you think about the following points?

  🔴 Correctness — Stateproofs: state is wrong for flat storage
  
  EthCapabilities.cs:
  Stateproofs: state
  The spec deliberately separates state from stateproofs because eth_getProof requires Merkle/trie nodes, which a node may not retain even when the flat state is
  queryable. FlatWorldStateManager.GetOldestStateBlock() correctly returns null (no rolling window), so state ends up Available for a flat-storage node — but
  Stateproofs is aliased to the same value, advertising proof capability that eth_getProof cannot satisfy. Routers will misroute proof requests to flat nodes.
  
  Fix: add an IWorldStateManager.SupportsTrieProofs (or equivalent) and gate Stateproofs:
  Stateproofs: worldStateManager.SupportsTrieProofs ? state : Disabled
  WorldStateManager (trie-store backed) → true; FlatWorldStateManager → false.
  
  🔴 Correctness — Tx/Logs alias Receipts ignores LogIndex

  EthCapabilities.cs:30,34 (Tx => Receipts, Logs => Receipts).

  In Nethermind, tx-by-hash lookup goes through IReceiptStorage, so the Tx alias is defensible. But eth_getLogs is increasingly served by Nethermind.Db.LogIndex,
  and ILogIndexConfig is already in the DI graph (it's even injected into BoundedModulePoolTests constructors in this very PR). If LogIndex has a different
  retention floor than receipts, logs.oldestBlock is wrong.
  
  At minimum: a comment explaining the invariant being relied on, or — preferably — feeding the actual log-index lower bound into Logs instead of reusing Receipts.
  Spec carves them out for exactly this reason.

  ---
  🟡 Minor
  
  - IEthRpcModule.cs:324 ExampleResponse uses 0x000…000 for the head hash (Keccak.Zero). Schema-valid but visually a sentinel; replace with a realistic-looking
  64-hex value for the auto-generated docs (the @claude bot noted this too).
  - HistoryPruner.SlotsPerEpoch made public const purely to expose 32 to EthCapabilitiesProvider. Prefer a method on IHistoryPruner (e.g.
  GetRetentionBlocks(RetentionEpochs)) so the spec's epoch→blocks conversion lives next to the pruner's other logic and stays out of public API surface.
  - EthCapabilities.cs comment claims JsonPropertyOrder "preserves the spec's canonical key ordering on the wire." The spec doesn't mandate JSON object key ordering
   — clarify the comment is for example-payload visual parity, not a wire requirement.
  - No-head path returns ChainHead(0, Hash256.Zero) plus all-Disabled. Genesis always exists so this is essentially dead code, but emitting a zero-hash payload at
  all is misleading. Consider returning a 503-style RPC error instead.
  - receipts barrier test gap: BuildResource uses Math.Max(AncientBodiesBarrier, AncientReceiptsBarrier) — but no scenario in CapabilitiesScenarios() exercises
  AncientReceiptsBarrier > AncientBodiesBarrier. Worth one extra test.

src/Nethermind/Nethermind.Blockchain/Find/IBlockFinder.cs
```diff
@@ -148,5 +148,12 @@ public interface IBlockFinder
         /// Highest state persisted
         /// </summary>
         long? BestPersistedState { get; set; }
+
+        /// <summary>
+        /// Lower bound of the historical state window — the oldest block for which the node
+        /// retains state. Updated when fast/snap sync completes (= pivot) and after a full
+        /// pruning run completes (= copied state's block). Null if never set (archive from genesis).
+        /// </summary>
+        long? OldestStateBlock { get; set; }
```
asdacap: Please move to IWorldStateManager or IStateReader instead.

asdacap (changes requested):

LukaszRozmej:
@claude re-review

LukaszRozmej:
@claude review

svlachakis (approved):

LukaszRozmej:
@claude re-review

LukaszRozmej:
@claude review

src/Nethermind/Nethermind.Init/Modules/FlatWorldStateModule.cs
```diff
@@ -40,7 +40,14 @@ protected override void Load(ContainerBuilder builder)
             .AddSingleton<IWorldStateManager, FlatWorldStateManager>()
             .OnActivate<IWorldStateManager>((worldStateManager, ctx) =>
             {
-                new TrieStoreBoundaryWatcher(worldStateManager, ctx.Resolve<IBlockTree>(), ctx.Resolve<ILogManager>());
+                IBlockTree blockTree = ctx.Resolve<IBlockTree>();
+                ILogManager logManager = ctx.Resolve<ILogManager>();
+                new TrieStoreBoundaryWatcher(worldStateManager, blockTree, logManager);
+                StateMetadataValidator.DiscardStaleFloors(
```
asdacap: What?

src/Nethermind/Nethermind.Init/Modules/WorldStateModule.cs
```diff
@@ -82,6 +82,12 @@ dbFactory is not MemDbFactory
             .AddSingleton<PruningTrieStateFactoryOutput>()
 
             .Map<IWorldStateManager, PruningTrieStateFactoryOutput>((o) => o.WorldStateManager)
+            .OnActivate<IWorldStateManager>((wsm, ctx) =>
+                StateMetadataValidator.DiscardStaleFloors(
```
asdacap: Two time?

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthCapabilitiesProvider.cs
```diff
@@ -0,0 +1,87 @@
...
+using Nethermind.History;
+using Nethermind.State;
+using Nethermind.Synchronization;
+
+namespace Nethermind.JsonRpc.Modules.Eth;
+
+public sealed class EthCapabilitiesProvider(
+    IReadOnlyBlockTree blockTree,
+    IWorldStateManager worldStateManager,
+    OldestStateBlockStore oldestStateBlockStore,
```
asdacap: Hide behind IWorldStateManager please. Implement separately for flatdb.

asdacap (changes requested):

src/Nethermind/Nethermind.Init/Modules/FlatWorldStateModule.cs
LukaszRozmej: So this is kind of deliberate.
Now we have in metadata db info that is related to state. 
Now if someone does drop state db (like a directory) - which IMO is useful feature to force resync - this data becomes stale and misleading. So I added a quick check for it.
Alternatively I can put it into state db (metadata column?)

src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthCapabilitiesProvider.cs
LukaszRozmej: Yeah that was my previous iteration, but:
- It living into metadata db caused issues and IMO wrong dependencies
- WordStateManager needed to be injected in places where it kind of shouldn't

LukaszRozmej: maybe if I put it under a separate interface that IWorldStatemanager would inherit + living in a separate metadata column in StateDb would be optimal...?

asdacap: For flat it does not even use metadata db, it has its own metadata db, which it can consult to report this. What kind of place the WorldStateManager should not be put? THen put it in IStateReader?

src/Nethermind/Nethermind.Init/Modules/FlatWorldStateModule.cs
asdacap: Put it in trie world state factory I guess.

LukaszRozmej:
@claude review

LukaszRozmej:
@claude review

src/Nethermind/Nethermind.Synchronization/FastSync/StateSyncRunner.cs
```diff
@@ -108,6 +109,9 @@ public async Task RunStateSyncRounds(CancellationToken token)
         treeSync.VerifyPostSyncCleanUp();
         treeSync.FinalizeSync(finalPivot);
 
+        // Reported through eth_capabilities.
+        stateBoundary.OldestStateBlock = finalPivot.Number;
```
asdacap: Not the righ place.

src/Nethermind/Nethermind.State.Flat/ScopeProvider/FlatWorldStateManager.cs
```diff
@@ -36,6 +37,8 @@ public class FlatWorldStateManager(
 
     private readonly FlatTrieVerifier _trieVerifier = new(flatDbManager, persistence, logManager);
 
+    private readonly StateBoundaryStore _boundaryStore = new(flatDb.GetColumnDb(FlatDbColumns.Metadata));
```
asdacap: Wrong. There is no store for flat, it just check the IPersistence directly.

src/Nethermind/Nethermind.State.Flat/ScopeProvider/FlatWorldStateManager.cs
```diff
@@ -47,6 +50,14 @@ public class FlatWorldStateManager(
         logManager);
     public IReadOnlyKeyValueStore? HashServer => null;
 
+    public long? RetentionWindowBlocks => null;
+
+    public long? OldestStateBlock
+    {
+        get => _boundaryStore.OldestStateBlock;
+        set => _boundaryStore.OldestStateBlock = value;
```
asdacap: No mutation with flat. Flat already have the the state information written in PersistenceManager.

src/Nethermind/Nethermind.State/IWorldStateManager.cs
```diff
@@ -10,7 +10,7 @@
 
 namespace Nethermind.State;
 
-public interface IWorldStateManager
+public interface IWorldStateManager : IStateBoundary, IStateBoundaryWriter
```
asdacap: No `IStateBoundaryWriter`. That is clearly trie store specific.

src/Nethermind/Nethermind.State/IStateBoundary.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// is reported via <see cref="OldestStateBlock"/> instead.
+    /// </summary>
+    long? RetentionWindowBlocks { get; }
+}
+
+/// <summary>
+/// Write side of <see cref="IStateBoundary"/>. Held only by the components that legitimately
+/// advance the floor.
+/// </summary>
+public interface IStateBoundaryWriter
```
asdacap, line 38: Writer is triestore specific. Hide it. Flat already handle the state tracking. Only the reader is needed with flat.

asdacap (changes requested):

asdacap (approved):

svlachakis:
@claude please check carefully this review whether it's valid or not 

  🔴 Critical — must fix before merge
  
  Stale rebase reverts two merged PRs. The diff against origin/master silently undoes:
  - #11811 (Log KAD report as a trace) — Nethermind.Network.Discovery/Kademlia/KBucketTree.cs removes the if (!_logger.IsDebug) return;
  early-return guard and flips _logger.Trace(...) → _logger.Info(...) in LogTreeStructure. Reverts the allocation guard and log-level fix.
  - #11794 (EIP-7778 gas fix) — Nethermind.Evm/TransactionProcessing/TransactionProcessor.cs and Nethermind.Evm.Test/Eip7778Tests.cs revert to
  pre-refund block gas tracking (drops _blockCumulativeReceiptGas).
  - Plus a stray blank-line diff in Nethermind.Blockchain/ReadOnlyBlockTree.cs:37.
  
  Fix: rebase onto current origin/master and drop the reversions.
  
  🟠 High

  1. FullPruner durability ordering (FullPruner.cs:188-194). Sequence is TryCopyTrie (which calls pruning.Commit() swapping the cloned DB) → then
  _stateBoundary.OldestStateBlock = stateToCopy. Kill the process in between and the new DB is live without the marker → eth_capabilities.state
  reports oldestBlock: 0 (archive-looking) even though state below is gone. Fix: write the marker into the cloning DB before Commit(), so the
  marker becomes atomic with the swap.
  2. StateBoundaryStore.OldestStateBlock.get takes a lock on every call (StateBoundaryStore.cs:29-33). Reads dominate (every eth_capabilities),
  writes are rare (post-finalize-sync, post-full-prune). Fix: Volatile.Read on a long with long.MinValue sentinel for null; keep
  Interlocked.CompareExchange on the writer side for the monotonicity check.

  🟡 Medium

  3. (IStateBoundaryWriter)worldStateManager cast in FullPrunerFactory.cs:62 — FlatWorldStateManager doesn't implement the writer; the cast would
  InvalidCastException if wiring ever changed. Fix: inject IStateBoundaryWriter separately from DI; drop the cast.
  4. Silent rejection of backward boundary writes (StateBoundaryStore.cs:39). If a reorg or bug ever drives a backward floor write, it's dropped
  with no log → invisible regression. Fix: keep dropping the write, but log at Warn/Debug with requested-vs-current values.
  5. Redundant null check in EthCapabilitiesProvider.cs:73-76 — retention is { } cfg in the conjunction is provably true whenever windowOldest is 
  { } w is true (windowOldest is non-null IFF retention is non-null per the line above). Restructure with a single early-return on retention is 
  not { } cfg. 

  🟢 Low

  6. Test naming pattern (EthRpcModuleTests.Capabilities.cs:289, 311, 326 and StateBoundaryStoreTests.cs:14, 18, 24, 36, 47). Mixed sentence-case
  and lower-snake_case violates the project's {What}_{When}_{Then} convention — e.g. eth_capabilities_no_head_disables_all_resources should be
  EthCapabilities_WhenHeadIsNull_DisablesAllResources.
  7. Pre-existing typo _readaOnlyCodeCb in WorldStateManager.cs:19 — since the PR edits this class, cheap to ride along.

LukaszRozmej:
@claude re-review

svlachakis (approved):

damian-orzechowski (approved):

PR #11440, opened 2026-04-30
chore: Remove NuGet package caching

svlachakis:
## Changes

Fixes random CI failures on NuGet Caching

- Remove caching because it's racing and failing. Doesn't seem to be worth the overhead of adding warmup job.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [ ] Yes
- [x] No


svlachakis:
@claude review again

svlachakis:
@claude review again

LukaszRozmej (approved):

rubo (changes requested):
Too complicated approach and too much code just to shave off a minute or two from _some_ heavy projects, while a lot of tests will take 8 minutes or more anyway?

See [cached](https://github.com/NethermindEth/nethermind/actions/runs/25169427020/job/73796821074) vs [without](https://github.com/NethermindEth/nethermind/actions/runs/25169719277/job/73785662208).

We don't want to maintain so much code with no obvious gains for us.

svlachakis:
@rubo Fair enough, dropped the cache warmup architecture entirely, removed the inline actions/cache@v5 blocks too, since the linked runs showed cached vs uncached delta is small for our test jobs and also racing and crashing occasionally which is annoying.

rubo (approved):

svlachakis:
@claude review again

PR #11443, opened 2026-04-30
fix(snap-sync): three Windows-flake fixes for E2ESyncTests.SnapSync

LukaszRozmej:
## Summary

Three commits in this PR fix the Windows-only `E2ESyncTests.SnapSync` flake observed in CI ([run #25139586200, job #73686059011](https://github.com/NethermindEth/nethermind/actions/runs/25139586200/job/73686059011?pr=11422) on PR #11422):

1. **fix(db): cross-column snapshot atomicity in `SnapshotableMemColumnsDb`** (`c14030da4d`) — eliminates the `TrieNodeException` mode.
2. **fix(snap-sync): drain in-flight ISnapTrees in `FlatTreeSyncStore.FinalizeSync`** (`12a66f1e1b`) — closes a separate ordering gap where post-sync readers could see state from not-yet-disposed phase-1 trees.
3. **fix(snap-sync): don't publish `_initialized` until `Clear()` completes** (`73b160f057`) — eliminates the deterministic "missing in flat" mode.

Plus the stress reproducer (`edc0d3dbc9`) used to drive the diagnosis.

**Result:** 5 stress runs × 60 tests = **300 invocations, 0 failures**. Previously 8–15 failures per 60-test run on Windows.

### Mode B: `TrieNodeException` (cross-column tearing)

Block processing leases a fresh `persistenceReader` while the persistence pipeline is committing, gets a snapshot where the Accounts column is updated but the StateNodes column is not, walks the trie at the new state root, and throws `TrieNodeException` for a node that should be there.

### Mode A: `Verification failed: N missing in flat`

Trie-side iteration finds account leaves that the flat-side iterator can't see. The same ~9 hashed addresses missed reliably across runs, all in verifier partitions 0–1 (first byte ≤ 0x3f). Counter-based diagnosis (instrumenting both source-side `SetAccount`/`RemoveAccount`/`DeleteAccountRange` and sink-side `InMemoryWriteBatch.Set` filtered to `Column_Account`) showed source vs. sink Set counts matched perfectly while the sink had ~507K *unaccounted* Removes. Tracing back uncovered `ClearAllColumns` racing against fresh sync writes — the third fix below.

## Fix 1 — `c14030da4d` cross-column snapshot atomicity

`SnapshotableMemColumnsDb.CreateSnapshot()` iterates columns and captures each column's snapshot independently. `InMemoryColumnWriteBatch.Dispose()` iterates per-column batches and disposes each independently. Without a global guard, a `CreateSnapshot` concurrent with an in-flight `writeBatch.Dispose` can capture some columns *after* the new writes and others *before* — a cross-column-inconsistent reader view.

This race only affects the test backend — RocksDB has atomic cross-CF snapshots — but it manifests concretely in `E2ESyncTests.SnapSync`. The fix wraps `CreateSnapshot` and the wrapped `AtomicColumnsWriteBatch.Dispose` in a single `Lock` so per-Dispose multi-column commits are atomic w.r.t. snapshot creation.

## Fix 2 — `12a66f1e1b` drain in-flight ISnapTrees in FinalizeSync

Phase-1 / phase-2 trees produced by `FlatSnapTrieFactory` hold per-tree `IWriteBatch` instances; until each tree's `Dispose` returns, its writes aren't yet committed to the underlying persistence. Without a drain, a caller that observes "sync finished" can read state that's missing accounts from late-disposed trees.

`FlatSnapTrieFactory` now tracks an `_inFlightTrees` counter (incremented in `CreateStateTree`/`CreateStorageTree`, decremented in `FlatSnapStateTree.Dispose`/`FlatSnapStorageTree.Dispose`). `FlatTreeSyncStore.FinalizeSync` calls `WaitForInFlightTreesDrained` (spin-wait, 30s timeout) before bumping the persisted state ID. This is independent of fix 3 — it closes a different ordering window that can otherwise re-emerge as the same symptom.

## Fix 3 — `73b160f057` don't publish `_initialized` until `Clear()` completes

`FlatSnapTrieFactory.EnsureDatabaseCleared` had this ordering:

```csharp
if (_initialized) return;
using (_lock.EnterScope())
{
    if (_initialized) return;
    _initialized = true;          // published BEFORE Clear() ran
    persistence.Clear();           // still iterating + queueing Removes
}
```

Sequence that wipes accounts:

1. Thread T1 enters `EnsureDatabaseCleared`, sets `_initialized = true`, then starts `persistence.Clear()`. `Clear()` opens a single `IColumnsWriteBatch`, then for each column iterates `db.GetColumnDb(column).GetAllKeys()` and queues `columnBatch.Remove(key)` for every key it sees — and only disposes the batch (applying the queued Removes) when *all columns* are done.
2. Thread T2 hits the unlocked fast-path `if (_initialized) return;`, sees `true`, and skips Clear entirely.
3. T2 calls `persistence.CreateWriteBatch(...)` and starts writing accounts via `SetAccount`. These commit normally (their own `IPersistence.IWriteBatch.Dispose`).
4. T1's `Clear()` is still inside its `IColumnsWriteBatch`. As it iterates `GetAllKeys()` for the Account column, it sees the keys T2 just committed and queues `columnBatch.Remove(key)` for them.
5. T1's batch finally disposes. Every queued `Remove` lands on the live DB — including T2's freshly-written accounts. The flat DB ends up missing entries that the trie still references; the verifier reports "Account in trie not found in flat".

The fix: publish `_initialized = true` *after* `persistence.Clear()` returns, and mark the field `volatile`. Concurrent fast-path readers either observe `false` (and wait on the lock) or observe `true` (and Clear is genuinely done, so it's safe to proceed).

### How fix 3 was diagnosed

Surface symptoms (key seen by trie iterator but not by flat iterator) were consistent across runs but didn't point at any obvious caller. Source-side instrumentation in `BaseFlatPersistence.WriteBatch` (`SetAccount` / `RemoveAccount` / `DeleteAccountRange` preview) plus sink-side instrumentation in `InMemoryWriteBatch.Set` (filtered to `_store.Name == "Column_Account"`) showed:

```
setBaseFlat = 1,134,377   setIM = 1,134,377    ← all sets accounted
removeBaseFlat = 1,974    DeleteAccountRangePreview = 552
removeIM = 509,598
clearAllColumnsCalls = 60   clearAccountRemoves = 507,072    ← gap closes here
```

507,072 of the 509,598 sink Removes came from `ClearAllColumns` — i.e. ~8,451 accounts removed *per test*, where the DB was supposed to be empty when Clear ran. ~8,451 ≈ the test fixture's account count, confirming Clear was racing against sync writes. The race is timing-sensitive (any instrumentation in the hot path masks it) which is why Linux/macOS pass and only Windows reproduces.

## Stress reproducer

`SnapSync_StressRepro` runs 30 iterations × 2 Flat fixtures = 60 cases, each as its own NUnit case (so a single failure doesn't terminate the rest), no `[Retry]`, `[Explicit]`. ~1m runtime.

```
dotnet test --project src/Nethermind/Nethermind.Synchronization.Test/Nethermind.Synchronization.Test.csproj \
  -c release -- --filter "FullyQualifiedName~SnapSync_StressRepro"
```

| Configuration | Local Windows results (per 60-test run) |
|---|---|
| Baseline (no fix) | 18/60 fail; ~12 Mode A + ~6 Mode B |
| + cross-column fix (fix 1) | ~18/60 fail; 18 Mode A + 0 Mode B |
| + ISnapTree drain (fix 2) | 8–15/60 fail; all Mode A |
| + `_initialized` race (fix 3) | **0/60** |

Across 5 consecutive runs of the stress reproducer with all three fixes applied: **0/300 failures**.

## Test plan

- [x] `Nethermind.Db` builds (0 warnings, 0 errors)
- [x] `Nethermind.State.Flat` builds (0 warnings, 0 errors)
- [x] `Nethermind.Synchronization.Test` builds (0 warnings, 0 errors)
- [x] Stress reproducer: 0/300 failures across 5 consecutive 60-test runs on Windows

### DIFF
--- a/src/Nethermind/Nethermind.Db/SnapshotableMemColumnsDb.cs
+++ b/src/Nethermind/Nethermind.Db/SnapshotableMemColumnsDb.cs
@@ -17,14 +17,7 @@ namespace Nethermind.Db
         private readonly Dictionary<TKey, SnapshotableMemDb> _columnDbs = new();
         private readonly bool _neverPrune;
 
-        // Cross-column atomicity guard. Each per-column SnapshotableMemDb has its own version
-        // counter and lock, so per-column reads/writes are individually consistent. But a
-        // multi-column writeBatch dispose applies columns one-by-one, and CreateSnapshot
-        // captures column snapshots one-by-one. Without this lock a snapshot taken concurrently
-        // with an in-flight writeBatch dispose can capture some columns AFTER the new writes
-        // and others BEFORE, producing a cross-column-inconsistent reader view. RocksDB does
-        // not have this problem (its snapshots are atomic across CFs); this lock makes the
-        // in-memory test backend match.
+        // Cross-column atomicity guard so CreateSnapshot can't observe a half-applied multi-column writeBatch dispose. RocksDB has this for free; this matches it.
         private readonly Lock _atomicityLock = new();
 
         private SnapshotableMemColumnsDb(TKey[] keys, bool neverPrune)
--- a/src/Nethermind/Nethermind.Db/SnapshotableMemColumnsDb.cs
+++ b/src/Nethermind/Nethermind.Db/SnapshotableMemColumnsDb.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Threading;
 using Nethermind.Core;
 
 namespace Nethermind.Db
@@ -15,6 +16,16 @@ namespace Nethermind.Db
     {
         private readonly Dictionary<TKey, SnapshotableMemDb> _columnDbs = new();
         private readonly bool _neverPrune;
+
+        // Cross-column atomicity guard. Each per-column SnapshotableMemDb has its own version
+        // counter and lock, so per-column reads/writes are individually consistent. But a
+        // multi-column writeBatch dispose applies columns one-by-one, and CreateSnapshot
+        // captures column snapshots one-by-one. Without this lock a snapshot taken concurrently
+        // with an in-flight writeBatch dispose can capture some columns AFTER the new writes
+        // and others BEFORE, producing a cross-column-inconsistent reader view. RocksDB does
+        // not have this problem (its snapshots are atomic across CFs); this lock makes the
+        // in-memory test backend match.
+        private readonly Lock _atomicityLock = new();
 
         private SnapshotableMemColumnsDb(TKey[] keys, bool neverPrune)
         {
@@ -55,16 +66,37 @@ namespace Nethermind.Db
 
         public IReadOnlyColumnDb<TKey> CreateReadOnly(bool createInMemWriteStore) => new ReadOnlyColumnsDb<TKey>(this, createInMemWriteStore);
 
-        public IColumnsWriteBatch<TKey> StartWriteBatch() => new InMemoryColumnWriteBatch<TKey>(this);
+        public IColumnsWriteBatch<TKey> StartWriteBatch() => new AtomicColumnsWriteBatch(this);
 
         public IColumnDbSnapshot<TKey> CreateSnapshot()
         {
+            using Lock.Scope _ = _atomicityLock.EnterScope();
             Dictionary<TKey, IKeyValueStoreSnapshot> snapshots = new();
             foreach (KeyValuePair<TKey, SnapshotableMemDb> kvp in _columnDbs)
             {
                 snapshots[kvp.Key] = kvp.Value.CreateSnapshot();
             }
             return new ColumnSnapshot(snapshots);
+        }
+
+        /// <summary>
+        /// Wraps <see cref="InMemoryColumnWriteBatch{TKey}"/> so the per-column commit phase
+        /// happens under the columns DB's write lock, making the multi-column commit atomic
+        /// w.r.t. <see cref="CreateSnapshot"/>.
+        /// </summary>
+        private sealed class AtomicColumnsWriteBatch(SnapshotableMemColumnsDb<TKey> db) : IColumnsWriteBatch<TKey>
+        {
+            private readonly InMemoryColumnWriteBatch<TKey> _inner = new(db);
+
+            public IWriteBatch GetColumnBatch(TKey key) => _inner.GetColumnBatch(key);
+
+            public void Clear() => _inner.Clear();
+
+            public void Dispose()
+            {
+                using Lock.Scope _ = db._atomicityLock.EnterScope();
+                _inner.Dispose();
+            }
         }
 
         public void Dispose()
--- a/src/Nethermind/Nethermind.State.Flat/Sync/FlatTreeSyncStore.cs
+++ b/src/Nethermind/Nethermind.State.Flat/Sync/FlatTreeSyncStore.cs
@@ -1,6 +1,7 @@
 // SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
 // SPDX-License-Identifier: LGPL-3.0-only
 
+using System;
 using Nethermind.Core;
 using Nethermind.Core.Collections;
 using Nethermind.Core.Crypto;
@@ -9,13 +10,14 @@ using Nethermind.Logging;
 using Nethermind.Serialization.Rlp;
 using Nethermind.State.Flat.Persistence;
 using Nethermind.State.Flat.ScopeProvider;
+using Nethermind.State.Flat.Sync.Snap;
 using Nethermind.Synchronization.FastSync;
 using Nethermind.Trie;
 using Nethermind.Trie.Pruning;
 
 namespace Nethermind.State.Flat.Sync;
 
-public class FlatTreeSyncStore(IPersistence persistence, IPersistenceManager persistenceManager, ILogManager logManager) : ITreeSyncStore
+public class FlatTreeSyncStore(IPersistence persistence, IPersistenceManager persistenceManager, FlatSnapTrieFactory snapTrieFactory, ILogManager logManager) : ITreeSyncStore
 {
     // For flat, one cannot continue syncing after finalization as it will corrupt existing state.
     private bool _wasFinalized = false;
@@ -235,6 +237,13 @@ public class FlatTreeSyncStore(IPersistence persistence, IPersistenceManager per
     public void FinalizeSync(BlockHeader pivotHeader)
     {
         if (Interlocked.CompareExchange(ref _wasFinalized, true, false)) throw new InvalidOperationException("Db was finalized");
+
+        // Wait for any in-flight ISnapTree disposals to complete. Phase-1 / phase-2 trees
+        // hold per-tree IWriteBatch instances; until each tree's Dispose returns, its writes
+        // aren't yet committed to the underlying persistence. Without this drain a caller that
+        // observes "sync finished" can read state that's missing accounts from late-disposed
+        // trees (manifests as E2ESyncTests.SnapSync "Verification failed: N missing in flat").
+        snapTrieFactory.WaitForInFlightTreesDrained(TimeSpan.FromSeconds(30));
 
         using IPersistence.IPersistenceReader reader = persistence.CreateReader(ReaderFlags.Sync);
         StateId from = reader.CurrentState;
--- a/src/Nethermind/Nethermind.State.Flat/Sync/FlatTreeSyncStore.cs
+++ b/src/Nethermind/Nethermind.State.Flat/Sync/FlatTreeSyncStore.cs
@@ -18,8 +18,6 @@ namespace Nethermind.State.Flat.Sync;
 
 public class FlatTreeSyncStore(IPersistence persistence, IPersistenceManager persistenceManager, FlatSnapTrieFactory snapTrieFactory, ILogManager logManager) : ITreeSyncStore
 {
-    private readonly ILogger _logger = logManager.GetClassLogger<FlatTreeSyncStore>();
-
     // For flat, one cannot continue syncing after finalization as it will corrupt existing state.
     private bool _wasFinalized = false;
 
@@ -244,8 +242,9 @@ public class FlatTreeSyncStore(IPersistence persistence, IPersistenceManager per
         // aren't yet committed to the underlying persistence. Without this drain a caller that
         // observes "sync finished" can read state that's missing accounts from late-disposed
         // trees (manifests as E2ESyncTests.SnapSync "Verification failed: N missing in flat").
-        if (!snapTrieFactory.WaitForInFlightTreesDrained(TimeSpan.FromSeconds(30)))
-            if (_logger.IsWarn) _logger.Warn("FlatTreeSyncStore.FinalizeSync proceeding despite in-flight ISnapTrees not draining within 30s — committed state may be incomplete.");
+        // WaitForInFlightTreesDrained logs its own warning on timeout; we still proceed so a
+        // stuck tree doesn't permanently block FinalizeSync.
+        snapTrieFactory.WaitForInFlightTreesDrained(TimeSpan.FromSeconds(30));
 
         using IPersistence.IPersistenceReader reader = persistence.CreateReader(ReaderFlags.Sync);
         StateId from = reader.CurrentState;
--- a/src/Nethermind/Nethermind.State.Flat/Sync/FlatTreeSyncStore.cs
+++ b/src/Nethermind/Nethermind.State.Flat/Sync/FlatTreeSyncStore.cs
@@ -1,7 +1,6 @@
 // SPDX-FileCopyrightText: 2025 Demerzel Solutions Limited
 // SPDX-License-Identifier: LGPL-3.0-only
 
-using System;
 using Nethermind.Core;
 using Nethermind.Core.Collections;
 using Nethermind.Core.Crypto;
@@ -19,6 +18,8 @@ namespace Nethermind.State.Flat.Sync;
 
 public class FlatTreeSyncStore(IPersistence persistence, IPersistenceManager persistenceManager, FlatSnapTrieFactory snapTrieFactory, ILogManager logManager) : ITreeSyncStore
 {
+    private readonly ILogger _logger = logManager.GetClassLogger<FlatTreeSyncStore>();
+
     // For flat, one cannot continue syncing after finalization as it will corrupt existing state.
     private bool _wasFinalized = false;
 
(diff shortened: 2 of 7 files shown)


LukaszRozmej:
## Mode A diagnostic — phase-1 SnapSync trees not disposed when verifier runs

Added targeted instrumentation to `BaseFlatPersistence.WriteBatch.SetAccount` / `DeleteAccountRange`, `SnapshotableMemDb.Set`, and `FlatSnapStateTree` constructor / Commit / Dispose. Ran the stress reproducer and traced events for the deterministically-missing 9 paths. Concrete finding:

**Some `FlatSnapStateTree` instances start `Dispose()` but never finish before the verifier runs**, so their queued `BaseFlatPersistence.WriteBatch.SetAccount` writes never reach the underlying `SnapshotableMemDb`. Concrete example from one failing iteration (sequence numbers from event stream):

```
[04349] tid=707 FlatSnapStateTree CREATED id=499
[04358] tid=707 FlatSnapStateTree COMMIT_START id=499 upperBound=3E40FAF1
[04377] tid=707 FlatSnapStateTree COMMIT_END id=499
[04378] tid=707 FlatSnapStateTree DISPOSE_START id=499
... no DISPOSE_END for id=499 in the dump ...
[04386] tid=31 FlatSnapStateTree DISPOSE_START id=501
[04387] tid=31 FlatSnapStateTree DISPOSE_END id=501       ← later tree disposes fine
```

Tree `id=499` started disposing but never finished within the iteration's window. The dump fires when `VerifyTrie` throws, so by that point the writes from tree 499 are queued in its `IWriteBatch` but never reached the underlying column DB.

In iter 10 of the traced run: 27 trees created, 24 disposed, **3 stuck** (id=499 in `Dispose`, id=502 / 503 mid-`COMMIT`).

## Why `IsSnapGetRangesFinished` doesn't catch this

Expected: `_activeAccountRequests` is decremented in `SnapProvider.AddAccountRange` AFTER the inner `SnapProviderHelper.AddAccountRange` returns (which, via `using ISnapTree`, only returns once the tree's `Dispose` completes). So in theory phase 1 can't transition until all in-flight trees have finished disposing.

Observed: phase 1 transitions and `SyncUntilFinished` returns while tree disposals are still in flight. Either the dispose path has a hot reference held longer than its method scope, or the counter / mode-change semantics admit a window I haven't pinned down. The data is unambiguous though — by the time `VerifyTrie` runs, the writes that should be in flat are still queued on a non-yet-disposed `IWriteBatch`.

## Tested mitigations

| Approach | Result |
|---|---|
| `Task.Delay(1s)` before `FlushCache` | 12-15 fail (similar baseline) |
| `Task.Delay(10s)` before `FlushCache` | 13 fail (similar baseline) |

Even 10s isn't enough — some disposals appear to take much longer or are genuinely stuck on something. A delay-based fix isn't sufficient.

## Proposed real fix (next step, not in this PR)

Add an in-flight counter to `FlatSnapTrieFactory`:
- Increment on `CreateStateTree` / `CreateStorageTree`.
- Decrement when the returned tree's `Dispose()` completes.
- Expose `WaitForDrained(timeout)`.
- Call `WaitForDrained` in `FlatTreeSyncStore.FinalizeSync` before returning.

That would block the sync→post-sync transition until every phase-1 writeBatch is committed, eliminating the race. Production also benefits: a node that restarts mid-snap-sync currently risks losing uncommitted phase-1 writes.

LukaszRozmej:
## Update — added drain fix for phase-1 in-flight `ISnapTrees`

Pushed `12a66f1e1b`: `FlatSnapTrieFactory` now tracks in-flight `ISnapTree` instances; `FlatTreeSyncStore.FinalizeSync` drains them (30s timeout) before flushing. This eliminates the case where phase-1 trees with queued `IPersistence.IWriteBatch` writes hadn't finished disposing when the verifier ran.

### Stress-test results (60 iterations × 2 Flat fixtures, no `[Retry]`)

| Configuration | Failures | Failure type |
|---|---|---|
| Master baseline | 9–19 | Mix of Mode A + Mode B |
| + cross-column snapshot fix | 13–18 | All Mode A |
| + drain fix (this) | **8–15** | All Mode A |

Modest but real improvement. The drain doesn't fully eliminate Mode A.

### What's still unfixed

The remaining Mode A failures hit the **same deterministic 9 paths** as before — the drain doesn't change the failure signature, it just makes failure less frequent. In iterations where phase-1 (state ranges) doesn't progress at all (logs show `Phase 1: (  0.00 %)` then immediately `finished`), all work happens via phase-2 state-nodes-sync (`FlatTreeSyncStore.SaveNode`), and the drain catches no trees. These iterations still fail with the same 9 missing accounts.

So the remaining bug is in **phase-2 state-nodes-sync's flat-entry path**, not phase-1. The 9 deterministic addresses suggest a specific trie topology (something about how those leaves are reached) that causes `WriteAccountFlatEntries` to miss them, or some other structural issue in `FlatTreeSyncStore.SaveNode`.

Investigation handed off to whoever picks this up next: the data points to either (a) leaves that are inline in some node type that `WriteAccountFlatEntries` doesn't handle, or (b) a specific code path in the Branch deletion-range computation that's incorrect for those 9 addresses' positions in the trie.

LukaszRozmej:
**Root cause found and fixed in 73b160f057.**

The remaining "missing in flat" failures came from a race in `FlatSnapTrieFactory.EnsureDatabaseCleared`:

```csharp
if (_initialized) return;
using (_lock.EnterScope())
{
    if (_initialized) return;
    _initialized = true;          // ← published BEFORE Clear() ran
    persistence.Clear();           // ← still iterating GetAllKeys() / queueing Removes
}
```

Sequence that wipes accounts:
1. Thread T1 enters `EnsureDatabaseCleared`, sets `_initialized = true`, starts `persistence.Clear()` (which calls `db.GetColumnDb(column).GetAllKeys()` and queues `columnBatch.Remove(key)` for every key found in *each* column).
2. Thread T2 hits the fast path `if (_initialized) return;` — it sees `true` and skips Clear.
3. T2 calls `persistence.CreateWriteBatch(...)` and starts writing accounts via `SetAccount`. These commit.
4. T1's `Clear()` is still inside the same `using IColumnsWriteBatch ... batch`. Its `GetAllKeys()` either already saw T2's keys, or sees them after the next loop iteration; either way `columnBatch.Remove(key)` is queued.
5. T1's batch disposes. Every queued `Remove` is applied to the live DB — including T2's freshly-written accounts.

How I confirmed it: instrumented every `IWriteBatch` operation that lands on `Column_Account` and counted source-side calls (SetAccount / RemoveAccount / DeleteAccountRange) vs sink-side calls on the underlying `InMemoryWriteBatch`. The numbers across 60 iterations:

```
setBaseFlat = 1,134,377   setIM = 1,134,377   ✓ all sets accounted
removeBaseFlat = 1,974    DeleteAccountRangePreview = 552
removeIM = 509,598
clearAllColumnsCalls = 60   clearAccountRemoves = 507,072    ← gap closes here
```

507,072 of the 509,598 sink Removes came from `ClearAllColumns` — i.e. ~8,451 accounts removed *per test*, where the DB was supposed to be empty when Clear ran. That ~8,451 matches the test fixture's account count, confirming Clear was racing against sync writes.

**Fix:** publish `_initialized = true` *after* `persistence.Clear()` returns, and mark the field `volatile` so the publish has release semantics. Now T2's fast-path read of `_initialized = true` strictly happens-after T1's Clear, so T2 either skips because Clear is genuinely done, or waits on the lock.

**Verified:** 5 stress runs × 60 tests = 300 invocations, 0 failures. Previously 8–15 failures per 60-test run on Windows.

This was the last bug. Combined effect of the three commits in this PR:
- `c14030da4d` — eliminated Mode B (TrieNodeException, cross-column snapshot tearing)
- `12a66f1e1b` — drained phase-1 trees before FinalizeSync (still useful; protects against a separate ordering issue)
- `73b160f057` — eliminated Mode A (this race)

Marking the PR ready for review.

LukaszRozmej:
@claude re-review

LukaszRozmej:
@claude re-review

LukaszRozmej:
@claude will this cause performance regression?

LukaszRozmej:
@claude does this also fix some potential edge case production bugs?

src/Nethermind/Nethermind.Db/SnapshotableMemColumnsDb.cs
```diff
@@ -16,6 +17,16 @@ public class SnapshotableMemColumnsDb<TKey> : IColumnsDb<TKey> where TKey : stru
...
         private readonly bool _neverPrune;
 
+        // Cross-column atomicity guard. Each per-column SnapshotableMemDb has its own version
+        // counter and lock, so per-column reads/writes are individually consistent. But a
+        // multi-column writeBatch dispose applies columns one-by-one, and CreateSnapshot
+        // captures column snapshots one-by-one. Without this lock a snapshot taken concurrently
+        // with an in-flight writeBatch dispose can capture some columns AFTER the new writes
+        // and others BEFORE, producing a cross-column-inconsistent reader view. RocksDB does
+        // not have this problem (its snapshots are atomic across CFs); this lock makes the
+        // in-memory test backend match.
```
alexb5dh: subjective, but really don't like these 5+ lines comments

src/Nethermind/Nethermind.State.Flat/Sync/Snap/FlatSnapTrieFactory.cs
```diff
@@ -14,30 +15,105 @@ namespace Nethermind.State.Flat.Sync.Snap;
...
+
+    internal void OnTreeCreated() => Interlocked.Increment(ref _inFlightTrees);
+
+    internal void OnTreeDisposed() => Interlocked.Decrement(ref _inFlightTrees);
+
+    /// <summary>
+    /// Spin-wait until all trees handed out by this factory have been disposed. Returns true
+    /// if drained within <paramref name="timeout"/>, false on timeout.
+    /// </summary>
+    public bool WaitForInFlightTreesDrained(System.TimeSpan timeout)
```
alexb5dh:
```suggestion
    public bool WaitForInFlightTreesDrained(TimeSpan timeout)
```

src/Nethermind/Nethermind.State.Flat/Sync/FlatTreeSyncStore.cs
```diff
@@ -236,6 +239,14 @@ public void FinalizeSync(BlockHeader pivotHeader)
     {
         if (Interlocked.CompareExchange(ref _wasFinalized, true, false)) throw new InvalidOperationException("Db was finalized");
 
+        // Wait for any in-flight ISnapTree disposals to complete. Phase-1 / phase-2 trees
+        // hold per-tree IWriteBatch instances; until each tree's Dispose returns, its writes
+        // aren't yet committed to the underlying persistence. Without this drain a caller that
+        // observes "sync finished" can read state that's missing accounts from late-disposed
+        // trees (manifests as E2ESyncTests.SnapSync "Verification failed: N missing in flat").
+        if (!snapTrieFactory.WaitForInFlightTreesDrained(TimeSpan.FromSeconds(30)))
+            if (_logger.IsWarn) _logger.Warn("FlatTreeSyncStore.FinalizeSync proceeding despite in-flight ISnapTrees not draining within 30s — committed state may be incomplete.");
```
alexb5dh: Is this intended that logging will happen twice?
- `$"FlatSnapTrieFactory: {Interlocked.Read(ref _inFlightTrees)} trees still in flight after {timeout}"`
- `"FlatTreeSyncStore.FinalizeSync proceeding despite in-flight ISnapTrees not draining within 30s — committed state may be incomplete."`

src/Nethermind/Nethermind.State.Flat/Sync/Snap/FlatSnapTrieFactory.cs
```diff
@@ -14,30 +15,105 @@ namespace Nethermind.State.Flat.Sync.Snap;
...
+        }
+        catch
+        {
+            // Constructor failed: pair the OnTreeCreated above with OnTreeDisposed and release
+            // the per-tree resources ourselves, since no FlatSnapStateTree.Dispose will run.
+            OnTreeDisposed();
+            writeBatch.Dispose();
+            reader.Dispose();
+            throw;
+        }
```
alexb5dh: Alternative with single try/catch?

```suggestion
        IPersistence.IPersistenceReader? reader = null;
        IPersistence.IWriteBatch? writeBatch = null;
        bool treeTracked = false;
        try
        {
            reader = persistence.CreateReader(ReaderFlags.Sync);
            writeBatch = persistence.CreateWriteBatch(StateId.Sync, StateId.Sync, WriteFlags.DisableWAL);
            OnTreeCreated();
            treeTracked = true;
            return new FlatSnapStateTree(reader, writeBatch, syncConfig.EnableSnapDoubleWriteCheck, logManager, this);
        }
        catch
        {
            if (treeTracked) OnTreeDisposed();
            writeBatch?.Dispose();
            reader?.Dispose();
            throw;
        }
```

alexb5dh (approved):

alexb5dh:
@claude what execution paths / operations waiting-for-up-to-30-seconds in `FinalizeSync`/`WaitForInFlightTreesDrained`  can affect? Risks of it noticeable affecting performance?

src/Nethermind/Nethermind.State.Flat/Sync/FlatTreeSyncStore.cs
alexb5dh: ok, looks related to https://github.com/NethermindEth/nethermind/pull/11443#discussion_r3173564974

asdacap (changes requested):
Fix issue 2 and 3 on top of #11102. Just call EnsureInitialize at the start of ISnapSyncRunner.Run and call FinalyzeSync at end of ISnapSyncRunner.Run. No lock or _initialized flag needed.

src/Nethermind/Nethermind.Db/SnapshotableMemColumnsDb.cs
LukaszRozmej: I generally don't either, but this was so freaking hard to find and understand that I will make an exception for future Lukasz and Ashraf not to fuck it up :)

LukaszRozmej:
@alexb5dh feedback addressed in 91db203:

- Trimmed the cross-column-atomicity comment to a single line.
- `WaitForInFlightTreesDrained(TimeSpan timeout)` — re-added `using System;` and dropped the `System.` qualifier.
- Removed the duplicate warning at the `FinalizeSync` call site (the timeout warning is logged inside `WaitForInFlightTreesDrained` already); also removed the now-unused `_logger` field.
- Refactored `CreateStateTree` / `CreateStorageTree` to your single-try/catch shape with `treeTracked` flag and nullable refs.

@asdacap re your CHANGES_REQUESTED — opened #11472 stacked on `amirul/simple-sync-dispatcher` with the runner-driven shape (`EnsureInitialize` at start, `FinalizeSync` at end of `SnapSyncRunner.Run`, no lock or `_initialized` flag). Stress repro: 0/180 on Windows. Once #11102 lands I can either close this PR (since #11472 supersedes the runtime fixes) or rebase this PR on top to keep just the cross-column fix and the test reproducer — whichever is easier on your end.

LukaszRozmej:
Closing in favor of #11472 and #11102.

PR #11451, opened 2026-05-01
XDC Downgrade base protocol

batrr:
XDC runs on `eth/100`, reusing Ethereum's `eth` protocol name. This led to the assumption that it supports `eth/65`, but the XDC Go client only supports up to `eth/63`. Message codes `0x08/0x09` also conflict with XDC's deprecated
  `OrderTxMsg/LendingTxMsg`, causing the Go client to disconnect with wrong message format or unknown code.  

**Fix:** downgrade base protocol to `eth/63`.

### DIFF
--- a/src/Nethermind/Nethermind.Xdc.Test/XdcProtocolHandlerTests.cs
+++ b/src/Nethermind/Nethermind.Xdc.Test/XdcProtocolHandlerTests.cs
@@ -23,7 +23,6 @@ using Nethermind.Xdc.Types;
 using NSubstitute;
 using NUnit.Framework;
 using System;
-using System.Collections.Generic;
 using Nethermind.Blockchain;
 using Nethermind.Core;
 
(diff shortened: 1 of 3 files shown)


src/Nethermind/Nethermind.Xdc.Test/XdcProtocolHandlerTests.cs
```diff
@@ -296,4 +297,22 @@ public void SendSyncInfo_SameSyncInfoTwice_IsDeliveredTwice()
             session.Received(2).DeliverMessage(Arg.Any<SyncInfoMsg>());
         }
     }
+
+    [TestCase(false)]
+    [TestCase(true)]
+    public void SendNewTransactions_UsesTransactionsMessage_NotNewPooledTransactionHashes(bool sendFullTx)
```
ak88, line 303: this is not really a helpful test

ak88 (approved):

cicr99 (approved):

PR #11453, opened 2026-05-01
perf(db): account native rocksdb memory via GC.AddMemoryPressure

asdacap:
## Summary
- `HyperClockCacheWrapper` reports its cache capacity to the GC via `GC.AddMemoryPressure` on construct and `GC.RemoveMemoryPressure` on release.
- `DbOnTheRocks` reports, per instance, one full write buffer plus the per-DB block cache size when the DB is not using the shared cache (configured value, else the rocksdb 32MB default). Shared-cache DBs add no extra pressure since the shared wrapper already accounts for it.

This gives the GC a more accurate view of process memory so heuristics for collection / heap sizing reflect the native buffers RocksDB holds.

## Test plan
- [ ] `dotnet build -c release src/Nethermind/Nethermind.slnx` (passes locally for `Nethermind.Db.Rocks`)
- [ ] `dotnet test --project src/Nethermind/Nethermind.Db.Rocks.Test/Nethermind.Db.Rocks.Test.csproj -c release` — existing dispose/lifecycle tests would surface any unbalanced add/remove (would throw `ArgumentOutOfRangeException`)
- [ ] Sanity: open + dispose a `DbOnTheRocks` and confirm `GC.GetGCMemoryInfo()` reflects the added pressure

### DIFF
--- a/src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
+++ b/src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
@@ -98,6 +98,8 @@ public partial class DbOnTheRocks : IDb, ITunableDb, IReadOnlyNativeKeyValueStor
     private int _maxWriteBufferNumber;
     private readonly RocksDbReader _reader;
     private bool _isUsingSharedBlockCache;
+    private ulong _ownBlockCacheSize;
+    private long _addedMemoryPressure;
 
     public DbOnTheRocks(
         string basePath,
@@ -121,6 +123,9 @@ public partial class DbOnTheRocks : IDb, ITunableDb, IReadOnlyNativeKeyValueStor
         _iteratorManager = new IteratorManager(_db, null, _readAheadReadOptions);
 
         _reader = new RocksDbReader(this, CreateReadOptions, _iteratorManager, null);
+
+        _addedMemoryPressure = (long)_writeBufferSize + (long)_ownBlockCacheSize;
+        if (_addedMemoryPressure > 0) GC.AddMemoryPressure(_addedMemoryPressure);
     }
 
     protected virtual RocksDb DoOpen(string path, (DbOptions Options, ColumnFamilies? Families) db)
@@ -505,11 +510,16 @@ public partial class DbOnTheRocks : IDb, ITunableDb, IReadOnlyNativeKeyValueStor
         if (dbConfig.BlockCache is not null)
         {
             tableOptions.SetBlockCache(dbConfig.BlockCache.Value);
+            _ownBlockCacheSize = blockCacheSize > 0 ? blockCacheSize : 32_000_000;
         }
         else if (sharedCache is not null && blockCacheSize == 0)
         {
             tableOptions.SetBlockCache(sharedCache.Value);
             _isUsingSharedBlockCache = true;
+        }
+        else
+        {
+            _ownBlockCacheSize = blockCacheSize > 0 ? blockCacheSize : 32_000_000;
         }
 
         // Note: the ordering is important.
@@ -1487,6 +1497,8 @@ public partial class DbOnTheRocks : IDb, ITunableDb, IReadOnlyNativeKeyValueStor
         ReleaseUnmanagedResources();
 
         _dbsByPath.Remove(_fullPath!, out _);
+
+        if (_addedMemoryPressure > 0) GC.RemoveMemoryPressure(_addedMemoryPressure);
 
         _isDisposed = true;
     }
--- a/src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
+++ b/src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
@@ -98,7 +98,6 @@ public partial class DbOnTheRocks : IDb, ITunableDb, IReadOnlyNativeKeyValueStor
     private int _maxWriteBufferNumber;
     private readonly RocksDbReader _reader;
     private bool _isUsingSharedBlockCache;
-    private ulong _ownBlockCacheSize;
     private long _addedMemoryPressure;
 
     public DbOnTheRocks(
@@ -124,7 +123,6 @@ public partial class DbOnTheRocks : IDb, ITunableDb, IReadOnlyNativeKeyValueStor
 
         _reader = new RocksDbReader(this, CreateReadOptions, _iteratorManager, null);
 
-        _addedMemoryPressure = (long)_writeBufferSize + (long)_ownBlockCacheSize;
         if (_addedMemoryPressure > 0) GC.AddMemoryPressure(_addedMemoryPressure);
     }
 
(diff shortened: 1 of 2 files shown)


src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
```diff
@@ -505,12 +510,17 @@ protected virtual void BuildOptions<T>(IRocksDbConfig dbConfig, Options<T> optio
         if (dbConfig.BlockCache is not null)
         {
             tableOptions.SetBlockCache(dbConfig.BlockCache.Value);
+            _ownBlockCacheSize = blockCacheSize > 0 ? blockCacheSize : 32_000_000;
```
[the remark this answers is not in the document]
asdacap: Need to add comment that by default rocksdb will allocate 32MB of block cache size if not specified.

src/Nethermind/Nethermind.Db.Rocks/HyperClockCacheWrapper.cs
```diff
@@ -12,12 +12,16 @@ public class HyperClockCacheWrapper : SafeHandleZeroOrMinusOneIsInvalid
...
+    private readonly long _capacity;
+
     public HyperClockCacheWrapper(ulong capacity = 32_000_000) : base(ownsHandle: true)
     {
         lock (_nativeCacheLock)
         {
             SetHandle(Native.Instance.rocksdb_cache_create_hyper_clock(new UIntPtr(capacity), 0));
         }
+        _capacity = (long)capacity;
+        GC.AddMemoryPressure(_capacity);
```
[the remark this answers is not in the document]
asdacap: Fixed in 023b4f6 — guard with  so neither add nor remove fires when the native handle is null. — Claude (claude-opus-4-7)

src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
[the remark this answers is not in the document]
asdacap: Fixed in 023b4f6 — `_ownBlockCacheSize` is now zero in this branch, since the `HyperClockCacheWrapper` that owns the handle already reports pressure. Also added a comment in the else branch documenting that rocksdb defaults to a 32MB block cache when not specified. — Claude (claude-opus-4-7)

src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
```diff
@@ -121,6 +123,9 @@ public DbOnTheRocks(
         _iteratorManager = new IteratorManager(_db, null, _readAheadReadOptions);
 
         _reader = new RocksDbReader(this, CreateReadOptions, _iteratorManager, null);
+
+        _addedMemoryPressure = (long)_writeBufferSize + (long)_ownBlockCacheSize;
```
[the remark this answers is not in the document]
asdacap: Fixed in 023b4f6 — accumulation moved into `BuildOptions` itself; each per-column call adds its `_writeBufferSize + perCallCachePressure` to `_addedMemoryPressure`, so `ColumnsDb` now reports the sum across all columns instead of just the last. — Claude (claude-opus-4-7)

src/Nethermind/Nethermind.Db.Rocks/HyperClockCacheWrapper.cs
```diff
@@ -12,12 +12,18 @@ public class HyperClockCacheWrapper : SafeHandleZeroOrMinusOneIsInvalid
...
     public HyperClockCacheWrapper(ulong capacity = 32_000_000) : base(ownsHandle: true)
     {
         lock (_nativeCacheLock)
         {
             SetHandle(Native.Instance.rocksdb_cache_create_hyper_clock(new UIntPtr(capacity), 0));
         }
+        // If the native call returned a zero/null handle, SafeHandle won't call ReleaseHandle,
+        // so don't add pressure either — keep add/remove balanced.
+        _capacity = IsInvalid ? 0 : (long)capacity;
+        if (_capacity > 0) GC.AddMemoryPressure(_capacity);
```
LukaszRozmej, line 26: should this be under lock? Can this be reported multiple times?

src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
```diff
@@ -502,6 +505,10 @@ protected virtual void BuildOptions<T>(IRocksDbConfig dbConfig, Options<T> optio
         }
 
         BlockBasedTableOptions? tableOptions = new();
+        // Per-call cache pressure to add to the GC. Cache memory owned by a HyperClockCacheWrapper
+        // (whether passed via dbConfig.BlockCache or as the shared cache) is already reported by
+        // that wrapper, so don't double-count here.
```
LukaszRozmej: comment not needed?

src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
```diff
@@ -554,6 +567,10 @@ protected virtual void BuildOptions<T>(IRocksDbConfig dbConfig, Options<T> optio
             if (_logger.IsDebug) _logger.Debug($"Total max DB footprint so far is {_maxRocksSize / 1000 / 1000} MB");
         }
 
+        // Accumulate one full write buffer plus this column's own block cache (if any).
+        // BuildOptions is called once per column family for ColumnsDb, so accumulate to capture all columns.
+        _addedMemoryPressure += (long)_writeBufferSize + (long)perCallCachePressure;
```
LukaszRozmej, line 567: Does it need Interlocked?

LukaszRozmej (approved):

src/Nethermind/Nethermind.Db.Rocks/HyperClockCacheWrapper.cs
asdacap: No — this runs in the constructor, called exactly once per wrapper instance, and `ReleaseHandle` is called at most once by `SafeHandle`. The lock around `SetHandle` / `rocksdb_cache_destroy` exists only to serialize calls into the native cache create/destroy API; `GC.AddMemoryPressure` is itself thread-safe and doesn't need the same lock. — Claude (claude-opus-4-7)

src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
asdacap: Removed in f83a118. — Claude (claude-opus-4-7)

src/Nethermind/Nethermind.Db.Rocks/DbOnTheRocks.cs
asdacap: No — `BuildOptions` is only called from `Init`, which runs synchronously on the constructor thread (once for the DB options, then sequentially per column family). There's no concurrent writer to `_addedMemoryPressure`, so a plain `+=` is sufficient. — Claude (claude-opus-4-7)

benaadams (approved):

svlachakis (approved):

smartprogrammer93 (approved):

PR #11454, opened 2026-05-01
fix(ReceiptForRpc): omit non-spec `error` field from receipt JSON-RPC responses

### LINKED ISSUE #11207: `eth_getBlockReceipts`: non-spec `error` field in failed receipts (spec violation)
**Affected tests:** test_01, test_07

### Description

Nethermind includes an `error` field in transaction receipts for failed
transactions. Geth does not include this field. The Ethereum execution API spec
defines the receipt object with `additionalProperties: false`, meaning **no
fields beyond those listed in the spec are permitted**. The `error` field is
not listed in the spec — returning it is a spec violation, not a convenience
feature.

Two variants appear across the tests:

**Pre-Byzantium block (test_01, block 0x41B582 = 4,308,354):**
The `status` field doesn't exist (pre-EIP-658); these receipts use `root`
instead. Nethermind includes `"error": "BadInstruction"` for transactions that
hit the `INVALID` opcode (0xfe). Geth omits the field entirely.

**Post-Byzantium block (test_07, `latest`):**
Failed transactions (`status: 0x0`) carry an `error` field with the decoded
revert reason string. Geth omits this field and only returns `status: 0x0`.

### test_01 — block 0x41B582, pre-Byzantium (4 failed receipts)

```
Nethermind receipt (indices 240-243):
{
  "root": "0x1e1c9f61...",
  "error": "BadInstruction",
  "gasUsed": "0x...",
  ...
}

Geth receipt (same tx):
{
  "root": "0x1e1c9f61...",
  "gasUsed": "0x...",
  ...  (no error field)
}
```

### test_07 — latest block, post-Byzantium (6 failed receipts)

```
Nethermind receipt examples:
{ "status": "0x0", "error": "TRANSFER_FROM_FAILED", ... }
{ "status": "0x0", "error": "INSUFFICIENT_OUTPUT", ... }
{ "status": "0x0", "error": "LOP: already filled", ... }
{ "status": "0x0", "error": "Return amount is not enough", ... }
{ "status": "0x0", "error": "INSUFFICIENT_OUTPUT_AMOUNT", ... }
{ "status": "0x0", "error": "0x12bacdd3...1db49b7d", ... }  ← raw ABI-encoded

Geth receipts: no error field, only status: "0x0"
```

### Notes

The spec receipt object definition explicitly sets `additionalProperties: false`,
which prohibits any fields not listed in the schema. The `error` field is absent
from the schema, so Nethermind must not include it. Clients that validate
responses against the spec will reject these receipts.

The sixth error value in test_07 (`0x12bacdd3...`) is a raw ABI-encoded revert
payload rather than a decoded string. This is a secondary issue but also out of
spec since the field itself should not be present.

**Fix:** Remove the `error` field from `eth_getBlockReceipts` receipt objects.
Note: `debug_traceTransaction` and similar debug methods may legitimately return
error reasons - only the standardised receipt endpoints are spec-constrained.

manusw7:
Fixes #11207

## Changes

- Remove `Error` from `ReceiptForRpc` so it is no longer emitted in JSON-RPC responses. The Ethereum execution-apis `ReceiptInfo` schema declares `additionalProperties: false`, prohibiting the non-standard `error` field that Nethermind previously included for failed transactions.
- Endpoints affected: `eth_getBlockReceipts`, `eth_getTransactionReceipt`, `parity_getBlockReceipts`, proof receipt sub-object in `ProofRpcModule`, and the internal `DataFeed` monitoring stream.
- Update existing `EthRpcModuleTests` hardcoded JSON expectations to drop `,"error":"error"`.
- Add regression test `Error_field_is_not_serialized` and `Error_field_is_not_deserialized` in `ReceiptsForRpcTests` asserting the field is absent from serialized output.

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

- Added unit regression tests in `Nethermind.JsonRpc.Test/Data/ReceiptsForRpcTests.cs`.
- [ ] We should verify end-to-end against `rpc-tests` fixtures: `eth_getBlockReceipts/test_01` (pre-Byzantium block `0x41B582` with `BadInstruction`) and `test_07` (post-Byzantium block with revert reasons) now match Geth-aligned expectations.

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

JSON-RPC receipt responses no longer include the non-standard `error` field on failed transactions. This aligns Nethermind with the Ethereum execution-apis spec and Geth behavior. 
Clients that relied on this field for failure reasons should use `debug_traceTransaction` instead.


manusw7:
@claude review again

svlachakis:
@manusw7 please attach hive tests job run

manusw7:
> @manusw7 please attach hive tests job run

https://github.com/NethermindEth/nethermind/actions/runs/25318851815

svlachakis (approved):

src/Nethermind/Nethermind.JsonRpc/Data/ReceiptForRpc.cs
```diff
@@ -63,6 +63,8 @@ public ReceiptForRpc(Hash256 txHash, TxReceipt receipt, ulong blockTimestamp, Tx
         public Bloom? LogsBloom { get; set; }
         public Hash256? Root { get; set; }
         public long? Status { get; set; }
+
+        [JsonIgnore]
         public string? Error { get; set; }
```
LukaszRozmej: why not remove the property entirely?

manusw7: my though was that even if it should not be in the rpc response, keeping `error` internally can still be useful for internal use/debugging

LukaszRozmej: not really, it is there in Receipt, we can drop it from RPC type

LukaszRozmej (approved):

PR #11455, opened 2026-05-01
perf: replace O(n²) InsertSorted with deferred O(n log n) rebuild

kamilchodola:
## Summary

- Replace per-slot `InsertSorted` (O(n²) array reallocation) with deferred `RebuildSortedArrays` (O(n log n) single sort)
- `LoadPreStateStorage` now sets a dirty flag instead of reallocating two arrays per slot
- `StorageChanges` and `ChangedSlots` properties lazily rebuild sorted arrays on first access after mutation

## Problem

`InsertSorted` allocated 2 new arrays and copied all existing elements on every `LoadPreStateStorage` call. For blocks with thousands of unique slots (e.g. `sload_bloated` with 10GB state), this was O(n²) — causing a **90-second block** (`24363514` in run 25222293131).

dotTrace showed `InsertSorted` OwnTime=12,763 (14% of `LoadPreState` TotalTime=90,315).

## Fix

Defer sorted array construction: accumulate in the `Dictionary` during loading, sort once via `Array.Sort` when `StorageChanges` or `ChangedSlots` is first accessed.

## Types of changes

- [x] Optimization

## Testing

- [x] 16/16 BlockAccessList tests pass
- [x] 126/126 EIP7928 tests pass
- [x] Benchmark run triggered (25223422449)

### DIFF
--- a/src/Nethermind/Nethermind.Core.Test/BlockAccessLists/ReadOnlyAccountChangesTests.cs
+++ b/src/Nethermind/Nethermind.Core.Test/BlockAccessLists/ReadOnlyAccountChangesTests.cs
@@ -1,0 +1,96 @@
+// SPDX-FileCopyrightText: 2026 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using Nethermind.Core.BlockAccessLists;
+using Nethermind.Core.Test.Builders;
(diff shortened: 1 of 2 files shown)


Marchhill:
merged into https://github.com/NethermindEth/nethermind/pull/11511

PR #11456, opened 2026-05-01
Improve eth/70 checks

flcl42:
## Changes

- Pass receipts in batches up to soft limit
- Allow last receipt to fill the packet up to hard limit
- Split it if it does not fit

- [x] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Core/Collections/IOwnedReadOnlyList.cs
+++ b/src/Nethermind/Nethermind.Core/Collections/IOwnedReadOnlyList.cs
@@ -48,6 +48,12 @@ public interface IOwnedReadOnlyList<T> : IReadOnlyList<T>, IDisposable
 
 public static class OwnedReadOnlyListExtensions
 {
+    public static IOwnedReadOnlyList<T> Slice<T>(this IOwnedReadOnlyList<T> list, int start, int count) =>
+        new SlicedOwnedReadOnlyList<T>(list, start, count);
+
+    public static IOwnedReadOnlyList<T> Slice<T>(this IOwnedReadOnlyList<T> list, int start) =>
+        new SlicedOwnedReadOnlyList<T>(list, start, list.Count - start);
+
     public static void DisposeRecursive<T>(this IOwnedReadOnlyList<T> list) where T : IDisposable
     {
         for (int i = 0; i < list.Count; i++)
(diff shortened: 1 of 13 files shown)


flcl42:
@claude review again

flcl42:
@claude review again

flcl42:
@claude review once more

benaadams (dismissed):

src/Nethermind/Nethermind.Core/Collections/IOwnedReadOnlyList.cs
```diff
@@ -57,4 +63,14 @@ public static void DisposeRecursive<T>(this IOwnedReadOnlyList<T> list) where T
 
         list.Dispose();
     }
+
+    private sealed class SlicedOwnedReadOnlyList<T>(IOwnedReadOnlyList<T> list, int start, int count)
```
LukaszRozmej, line 67: maybe this should be a struct? (then returned by it's type from Slice to avoid boxing)

LukaszRozmej (dismissed):
This code feels complicated, can it be simplified?

src/Nethermind/Nethermind.Core/Collections/IOwnedReadOnlyList.cs
```diff
@@ -57,4 +63,14 @@ public static void DisposeRecursive<T>(this IOwnedReadOnlyList<T> list) where T
...
+        : SlicedReadOnlyList<T>(list, start, count), IOwnedReadOnlyList<T>
+    {
+        public ReadOnlySpan<T> AsSpan() => ((IOwnedReadOnlyList<T>)_list).AsSpan().Slice(_start, Count);
+
+        public void Dispose()
```
benaadams, line 72: Should the dispose be calling parent?

benaadams: Otherwise just return the ReadOnlySpan?

flcl42:
@LukaszRozmej 

What part is the most disturbing? When this protocol was discussed what bothered me is that we could replace this gas accounting for incomplete receipts with receipt root proofs like we do for snap think, which seems simpler to me. But I guess you mean something different


@LukaszRozmej @benaadams 

We need IOwnedReadOnlyList in GetReceiptsMessage70 which we not only send but also receive and need to clean up after the usage. Span do not work for such case. Dispose is called on parent for sure.

benaadams:
> We need IOwnedReadOnlyList in GetReceiptsMessage70 which we not only send but also receive and need to clean up after the usage. Span do not work for such case. Dispose is called on parent for sure.

Shouldn't implement Dispose and make it empty then?

flcl42:
@benaadams but IOwnedReadOnlyList is IDisposable

LukaszRozmej:
@flcl42 `src/Nethermind/Nethermind.Network/P2P/Subprotocols/Eth/V70/Eth70ProtocolHandler.cs` all those calculations seem complicated, can they be simplified or better encapsulated?

LukaszRozmej:
@claude re-review

benaadams (approved):

benaadams:
@claude re-review

benaadams:
@claude re-review

LukaszRozmej (approved):

PR #11458, opened 2026-05-01
fix(snap-sync): defer `FinalizeSync` to end of state sync to prevent canonical chain wipe

### LINKED ISSUE #11457: FlatDb 1.37.1: restart mid-snap promotes incomplete state, deletes canonical chain, sync permanently stuck
Description

After updating to 1.37.1 and syncing into a fresh FlatDb, the node reaches a few hundred blocks behind tip, stops downloading, then on restart deletes thousands of canonical blocks as "invalid" and remains permanently stuck at Downloaded 0 / 25,002,159 (0.00%) with repeated Unable to find beacon header messages on every ForkChoice update.

The deleted blocks are canonical (verified against Etherscan, e.g. block 24996520 = 0xa2a304db…). Recovery requires wiping the DB and resyncing.

This appears related to but distinct from #11418, #11442, and #11447 (all FlatDb / restart-during-snap symptoms in 1.37.1), but the failure mode here is canonical-chain deletion + unrecoverable stall, which makes it more severe.

Steps to Reproduce

1. Sync a Nethermind 1.37.1 node with FlatDb enabled into a fresh DB.
2. Let snap sync progress until close to tip but before completing (snap state-nodes phase still running, e.g. State Sync at < 1%) (user: when it gets close to the tip (few hundred blocks) then doesn't download anything).
3. Stop and restart the container.
4. Observe: on restart the node logs InvalidTransactionException: transaction nonce is too high on the first post-pivot block, then deletes thousands of subsequent canonical blocks as invalid and never recovers.

Desktop

- Operating System: Linux (arm64)
- Version: 1.37.1
- Installation Method: Docker

Additional context

[nethermind-laststart.txt](https://github.com/user-attachments/files/27287363/nethermind-laststart.txt)
[message.txt](https://github.com/user-attachments/files/27287362/message.txt)

svlachakis:
Fixes Closes Resolves https://github.com/NethermindEth/nethermind/issues/11457

#### Root cause

`TreeSync.SaveNode` called `_store.FinalizeSync(_stateSyncPivot.GetPivotHeader())` the moment the trie root was saved. `GetPivotHeader()` is mutating — calling it can rotate the pivot. On a `"too many empty responses"` rotation, the just-saved root belonged to the old pivot but `FinalizeSync` was invoked with the *new* pivot, persisting a `CurrentState` pointer to a state root with no trie data.

After restart, `FlatFullStateFinder` returned that block number, the sync mode selector skipped state-sync, and the block processor ran against partial state.

#### Fix

Move `FinalizeSync` from `SaveNode` to `VerifyPostSyncCleanUp`, gated on `pivotHeader.StateRoot == _rootNode`. If the pivot rotated, skip finalize; the next round retries on the new pivot.

#### Recovery

Preventative. Existing corrupted DBs from `1.37.1` still need a wipe + resync; the fix only stops the corruption from being created.

#### Tests

Added two unit tests in `StateSyncFeedTests.cs` (static inner fixture to avoid the parameterized parent's 4× duplication):

- match → `FinalizeSync` called, `SyncCompleted` fires
- rotation → `FinalizeSync` skipped (`SyncCompleted` still fires; see follow-up below)

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

Seems reproducible on flat DB, need to test against that


svlachakis:
@claude review again

svlachakis:
@claude review again

svlachakis:
@claude review again

svlachakis:
@claude please review again

svlachakis:
@claude did some reverts. both of the below will be investigated and maybe solved on different PRs, not here.

  - `SyncCompleted` still fires with the mismatched pivot, and `VerifyStateOnStateSyncFinished` does a one-shot unsubscribe — so verification is burned on the rotated pivot. Gating `SyncCompleted` on the StateRoot match was tried here and reverted because it broke `E2ESyncTests.SnapSync`; a proper fix needs separate investigation.
  - Pre-existing NRE risk in `ValidatePrepareRequest`: `_stateSyncPivot.GetPivotHeader().StateRoot` can throw if the getter returns null

asdacap (approved):

svlachakis:
closed in favor of https://github.com/NethermindEth/nethermind/pull/11480

PR #11462, opened 2026-05-01
Add NETH004 analyzer banning ConcurrentDictionary.Keys/.Values

asdacap:
## Changes

- Add `BannedConcurrentDictionaryKeysValuesAnalyzer` (NETH004) to `Nethermind.Analyzers`. Both `ConcurrentDictionary<TKey,TValue>.Keys` and `.Values` allocate a fresh `List`-backed `ReadOnlyCollection` snapshot every time they're accessed — a hidden allocation cliff on hot paths. The analyzer flags any access at compile time. Severity: Warning. Symbol-based comparison so `NonBlocking.ConcurrentDictionary` and same-named user types are not affected.
- Add 6 unit tests covering positives (`Keys`/`Values` parameterized) and negatives (`NonBlocking.ConcurrentDictionary`, plain `Dictionary`, other ConcurrentDictionary members, user-defined `Keys`/`Values` properties).
- Eliminate every existing infraction (50 hits across 27 files). Replacement patterns:
  - Plain `foreach` enumeration → `foreach (KeyValuePair<K,V> kvp in dict)` and use `kvp.Value` / `kvp.Key`.
  - LINQ chains / materialized collections → `dict.Select(static kvp => kvp.Value).ToArray()` etc.
  - `IEnumerable` exposure → lazy `dict.Select(static kvp => kvp.Value)`.
  - `IFullDb.Keys`/`Values` interface implementations in `MemDb` and `SimpleFilePublicKeyDb` still materialize (the interface contract requires `ICollection<byte[]>`), but the allocation is now explicit at the call site.

- [x] Optimization
- [x] Refactoring

#### Requires testing

- [x] Yes

- [x] Yes

#### Notes on testing

- New analyzer suite: 6 tests, all passing.
- Full `Nethermind.Analyzers.Test` suite: 50/50 passing.
- Full `Nethermind.slnx` builds clean (0 warnings, 0 errors) with the new analyzer enforcing.

- [x] No

- [x] No

## Remarks

- Diagnostic ID `NETH004` registered in `AnalyzerReleases.Unshipped.md`.
- Analyzer is wired in automatically via `Directory.Build.props` (no per-project change needed).

🤖 Generated with [Claude Code](https://claude.com/claude-code) (Opus 4.7, 1M context)

### DIFF
--- a/src/Nethermind/Nethermind.Analyzers.Test/BannedConcurrentDictionaryKeysValuesAnalyzerTests.cs
+++ b/src/Nethermind/Nethermind.Analyzers.Test/BannedConcurrentDictionaryKeysValuesAnalyzerTests.cs
@@ -29,12 +29,13 @@ public class BannedConcurrentDictionaryKeysValuesAnalyzerTests
         await Verify(source, Diagnostic().WithLocation(0).WithArguments(member));
     }
 
-    [Test]
-    public async Task NonBlocking_ConcurrentDictionary_no_diagnostic()
+    [TestCase("Keys")]
+    [TestCase("Values")]
+    public async Task NonBlocking_ConcurrentDictionary_Keys_or_Values_reports_diagnostic(string member)
     {
         // Stub a NonBlocking.ConcurrentDictionary<TKey,TValue> with the same shape
-        // to verify the analyzer matches on the BCL symbol, not the simple name.
-        string source = """
+        // since the analyzer matches on full metadata name.
+        string source = $$"""
             namespace NonBlocking
             {
                 public class ConcurrentDictionary<TKey, TValue>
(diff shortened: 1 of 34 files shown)


LukaszRozmej:
@claude how does `NonBlocking.ConcurrentDictionary` handle it? Is it not a problem there?

LukaszRozmej:
@asdacap please include NonBlocking one in this too!

LukaszRozmej:
@claude is there some accesses to `.Values` and `.Keys` of `ConcurrentDictionary` but through `IDictionary` interface in the codebase?

benaadams:
@claude does a foreach on ɑ ConcurrentDictionary always allocate and would it be better to cache the enumberator and reuse it with .Reset()?

LukaszRozmej:
@claude re-review, also suggest if we could somehow lower allocations from allocating enumerators

asdacap:
The main purpose of this PR is to prevent accidental LOH allocation. An alloc per iteration is outside the scope.

src/Nethermind/Nethermind.Analyzers.Test/BannedConcurrentDictionaryKeysValuesAnalyzerTests.cs
```diff
@@ -0,0 +1,165 @@
...
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.CSharp.Testing;
+using Microsoft.CodeAnalysis.Testing;
+using NUnit.Framework;
+
+namespace Nethermind.Analyzers.Test;
+
+public class BannedConcurrentDictionaryKeysValuesAnalyzerTests
```
LukaszRozmej, line 11: Collapse to more TestCases?

LukaszRozmej (approved):

svlachakis (approved):

src/Nethermind/Nethermind.Analyzers.Test/BannedConcurrentDictionaryKeysValuesAnalyzerTests.cs
asdacap: Done in 89283b2a39 — collapsed the four "no diagnostic" tests into a single `Non_target_patterns_no_diagnostic` that bundles every non-target pattern (same-named type in another namespace, plain `Dictionary`, user-defined `Keys`/`Values` properties, and other `ConcurrentDictionary` members) into one source. Positive cases stay parameterized over `Keys`/`Values`. Test count: 8 → 5.

— Claude Opus 4.7 (1M context), on @asdacap's behalf

svlachakis (approved):

PR #11463, opened 2026-05-02
style: remove unused `using`s in Benchmarks solution

alexb5dh:
Removes unused `using`s in Benchmarks.slnx.

- [x] Other: styling

#### Requires testing

- [x] No (build passes)


alexb5dh:
@claude re-review, this was rebased

asdacap (approved):

LukaszRozmej (approved):
Can you make sure this is required by the rules for Benchmark solution too?

alexb5dh:
> Can you make sure this is required by the rules for Benchmark solution too?

Unfortunately wasn't able to figure out why it didn't apply to Benchmarks. Dotnet interactions with .editorconfig are messed up. Same goes for Rider.

LukaszRozmej:
does it have it's own `Directory.Build.props`?

alexb5dh:
> does it have it's own Directory.Build.props?

Nethermind.slnx and Benchmarks.slnx share the same folder, so should use the same Directory.Build.props.

Actually, I don't think we enforce build error for inactive `using`s (`IDE0005`) - `dotnet build Nethermind.slnx`  still works when these are present.

And not sure changing it will make sense, given that there are 100+ `#if` directives with code hidden behind them.

PR #11465, opened 2026-05-02
Add NETH004: enforce RunContinuationsAsynchronously on TaskCompletionSource

asdacap:
## Summary

- Add Roslyn analyzer `NETH004` that flags `new TaskCompletionSource(...)` / `new TaskCompletionSource<T>(...)` constructions which do not pass a constant `TaskCreationOptions` value with the `RunContinuationsAsynchronously` bit set. Severity set to `error` in `.editorconfig`.
- Without that flag, `SetResult`/`SetException` invoke continuations *synchronously* on the completing thread — this has caused hard-to-trace deadlocks (continuation re-enters a lock the completer's caller holds) and latency spikes on hot paths.
- Fix all 30 pre-existing violations across 21 files (production code in Blockchain, Consensus, Network, Init, Optimism, Runner, etc., plus several test helpers/tests). Convention was previously followed at only ~50% of sites.

## Why this is needed

A repo audit before this change showed mixed compliance — the rule existed by convention but was silently regressing. Encoding it as a build-time analyzer prevents future drift. Non-constant `TaskCreationOptions` arguments are also flagged because the analyzer cannot prove the bit is set; callers wanting dynamic options must compose with a literal that already includes `RunContinuationsAsynchronously`.

## Test plan

- [x] `dotnet build -c release src/Nethermind/Nethermind.Analyzers/Nethermind.Analyzers.csproj` — succeeds.
- [x] `dotnet test --project src/Nethermind/Nethermind.Analyzers.Test/Nethermind.Analyzers.Test.csproj -c release --no-build` — 57 passed (13 new for `NETH004`, covering: no-arg ctor, state-only ctor, `TaskCreationOptions.None`, `RunContinuationsAsynchronously` alone, OR-combined flags, generic `TaskCompletionSource<T>`, non-constant options expression, type-shadowing edge case).
- [x] `dotnet build -c release src/Nethermind/Nethermind.slnx` — succeeds clean (analyzer at error severity, no remaining violations).
- [ ] CI: full test suite.

### DIFF
--- a/src/Nethermind/Nethermind.Analyzers.Test/TaskCompletionSourceMustRunContinuationsAsynchronouslyAnalyzerTests.cs
+++ b/src/Nethermind/Nethermind.Analyzers.Test/TaskCompletionSourceMustRunContinuationsAsynchronouslyAnalyzerTests.cs
@@ -52,6 +52,24 @@ public class TaskCompletionSourceMustRunContinuationsAsynchronouslyAnalyzerTests
     }
 
     [Test]
+    public async Task Const_local_with_run_continuations_async_no_diagnostic()
+    {
+        string source = """
+            class Test
+            {
+                void M()
+                {
+                    const System.Threading.Tasks.TaskCreationOptions opts =
+                        System.Threading.Tasks.TaskCreationOptions.RunContinuationsAsynchronously;
+                    var t = new System.Threading.Tasks.TaskCompletionSource(opts);
+                }
+            }
+            """;
+
+        await Verify(source);
+    }
+
+    [Test]
     public async Task NonConstant_options_expression_reports_diagnostic()
     {
         string source = """
@@ -96,7 +114,7 @@ public class TaskCompletionSourceMustRunContinuationsAsynchronouslyAnalyzerTests
         CSharpAnalyzerTest<TaskCompletionSourceMustRunContinuationsAsynchronouslyAnalyzer, DefaultVerifier> test = new()
         {
             TestCode = source,
-            ReferenceAssemblies = ReferenceAssemblies.Net.Net80,
+            ReferenceAssemblies = ReferenceAssemblies.Net.Net100,
         };
         test.ExpectedDiagnostics.AddRange(expected);
         await test.RunAsync();
--- a/src/Nethermind/Nethermind.Analyzers.Test/TaskCompletionSourceMustRunContinuationsAsynchronouslyAnalyzerTests.cs
+++ b/src/Nethermind/Nethermind.Analyzers.Test/TaskCompletionSourceMustRunContinuationsAsynchronouslyAnalyzerTests.cs
@@ -1,0 +1,104 @@
+// SPDX-FileCopyrightText: 2026 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.CSharp.Testing;
+using Microsoft.CodeAnalysis.Testing;
+using NUnit.Framework;
+
+namespace Nethermind.Analyzers.Test;
+
+public class TaskCompletionSourceMustRunContinuationsAsynchronouslyAnalyzerTests
+{
+    [TestCase("new System.Threading.Tasks.TaskCompletionSource()")]
+    [TestCase("new System.Threading.Tasks.TaskCompletionSource((object?)null)")]
+    [TestCase("new System.Threading.Tasks.TaskCompletionSource(System.Threading.Tasks.TaskCreationOptions.None)")]
+    [TestCase("new System.Threading.Tasks.TaskCompletionSource(System.Threading.Tasks.TaskCreationOptions.LongRunning)")]
+    [TestCase("new System.Threading.Tasks.TaskCompletionSource<int>()")]
+    [TestCase("new System.Threading.Tasks.TaskCompletionSource<int>(System.Threading.Tasks.TaskCreationOptions.None)")]
+    public async Task Construction_without_run_continuations_async_reports_diagnostic(string expression)
+    {
+        string source = $$"""
+            class Test
+            {
+                void M()
+                {
+                    var t = {|#0:{{expression}}|};
+                }
+            }
+            """;
+
+        await Verify(source, Diagnostic().WithLocation(0));
+    }
+
+    [TestCase("new System.Threading.Tasks.TaskCompletionSource(System.Threading.Tasks.TaskCreationOptions.RunContinuationsAsynchronously)")]
+    [TestCase("new System.Threading.Tasks.TaskCompletionSource(System.Threading.Tasks.TaskCreationOptions.RunContinuationsAsynchronously | System.Threading.Tasks.TaskCreationOptions.LongRunning)")]
+    [TestCase("new System.Threading.Tasks.TaskCompletionSource((object?)null, System.Threading.Tasks.TaskCreationOptions.RunContinuationsAsynchronously)")]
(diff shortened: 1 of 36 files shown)


src/Nethermind/Nethermind.Blockchain.Test/BlockTreeSuggestPacerTests.cs
```diff
@@ -50,6 +51,6 @@ public void WillOnlyUnblockOnceHeadReachHighEnough()
         waitTask.IsCompleted.Should().BeFalse();
 
         blockTree.NewHeadBlock += Raise.EventWith(new BlockEventArgs(Build.A.Block.WithNumber(6).TestObject));
-        waitTask.IsCompleted.Should().BeTrue();
+        await waitTask.WaitAsync(TimeSpan.FromSeconds(5));
```
asdacap: This defeat the purpose of the test.

asdacap: You're right — the 5s WaitAsync was too loose and the test would pass even on a near-timeout regression. Reverted to a bounded synchronous wait (`waitTask.Wait(TimeSpan.FromMilliseconds(500)).Should().BeTrue()`) in 19a0ce1. Still allows the now-async TCS continuation to be scheduled but asserts the unblock happens promptly — a real regression in the unblock path would still fail the test.

(Posted by Claude Opus 4.7 acting on @asdacap's behalf.)

LukaszRozmej (commented):
If this is required everywhere I would suggest to create a helper static factory method that would be used throughout the whole codebase.

asdacap:
The analyzer (NETH004, severity=error) already enforces the constraint at every `new TaskCompletionSource(...)` call site, so correctness doesn't need a factory — any forgotten flag fails the build. A `TaskCompletionSourceFactory.Create()` helper would be pure syntactic sugar on top of that.

Counter-arguments against adding the factory:
- It doesn't compose cleanly with the overloads taking `state` (`new TaskCompletionSource(state, options)`) or with callers that legitimately want to OR additional flags like `LongRunning`, so we'd end up with multiple factory overloads or callers reaching past it.
- It introduces a new concept reviewers/contributors must learn, where today they can read a standard BCL constructor.
- The 30 existing call sites are now consistent (`new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously)`), and the analyzer keeps them that way.

If the readability of the long flag is the concern, a `using static System.Threading.Tasks.TaskCreationOptions;` (or a `const` local) at the use site is a smaller hammer. Happy to reconsider if there's a specific pain point I'm missing — but my preference is to lean on the analyzer rather than add a wrapper.

(Posted by Claude Opus 4.7 on @asdacap's behalf.)

LukaszRozmej:
What about places where we use only `TaskCreationOptions.LongRunning` and `TaskCreationOptions.None`, there are quite a few in the codebase?

LukaszRozmej:
@claude review

LukaszRozmej (approved):

flcl42:
Other projects might benefit from the analyzers too, like Nethermind.PerfTest.

Maybe some more test cases:
```
[TestCase("new System.Threading.Tasks.TaskCompletionSource(creationOptions: System.Threading.Tasks.TaskCreationOptions.None)")]
[TestCase("new System.Threading.Tasks.TaskCompletionSource<int>((object?)null)")]
[TestCase("new System.Threading.Tasks.TaskCompletionSource<int>(creationOptions: System.Threading.Tasks.TaskCreationOptions.None)")]
```

asdacap:
Its not on?

asdacap:
Added the three suggested test cases in c824575:

- `new TaskCompletionSource(creationOptions: TaskCreationOptions.None)` (named-arg, non-generic)
- `new TaskCompletionSource<int>((object?)null)` (state-only generic ctor)
- `new TaskCompletionSource<int>(creationOptions: TaskCreationOptions.None)` (named-arg, generic)

All three are added to the `reports_diagnostic` parameterized case and pass — Roslyn's argument-to-parameter binding handles the named-arg form transparently, and the state-only generic ctor falls out of the loop without finding a `TaskCreationOptions` parameter, so the analyzer flags it.

Re: applying the analyzer to other projects like `Nethermind.PerfTest` — it's already wired in. `src/Nethermind/Directory.Build.props` adds `Nethermind.Analyzers` as an `Analyzer`-typed `ProjectReference` to every project under `src/Nethermind/` (except the analyzer itself and the source generator). `Nethermind.PerfTest.csproj` inherits that, so NETH006 is enforced there too — `dotnet build Nethermind.PerfTest/Nethermind.PerfTest.csproj` confirms `Nethermind.Analyzers.dll` is built and consumed as an analyzer. (PerfTest happens to have zero `TaskCompletionSource` usages, so nothing currently fires.)

(Posted by Claude Opus 4.7 on @asdacap's behalf.)

flcl42:
My codex said `src/Nethermind/Nethermind.PerfTest/Program.cs:354` still had `TaskCompletionSource<object> completionSource = new();` so I assumed no

LukaszRozmej (approved):

flcl42 (approved):
Please take care of or confirm it's not a case https://github.com/NethermindEth/nethermind/blob/master/src/Nethermind/Nethermind.PerfTest/Program.cs#L354

asdacap:
You're right — sorry, I missed that one in my earlier sweep because `TaskCompletionSource<object> completionSource = new();` (target-typed `new()`) doesn't match a `new TaskCompletionSource…` text grep. Fixed in 03f54b6:

- Added `TaskCreationOptions.RunContinuationsAsynchronously` to `Nethermind.PerfTest/Program.cs:354`.
- Added `TargetTyped_new_reports_diagnostic` so the test suite locks in the analyzer's coverage of `Type x = new()` (Roslyn lowers this to an `IObjectCreationOperation` with the inferred type, so the existing analyzer logic already handles it — the test pins the contract).

A small caveat on the "the analyzer's not on for PerfTest" theory: `Directory.Build.props` does register `Nethermind.Analyzers` as an `Analyzer`-typed `ProjectReference` for `Nethermind.PerfTest`, so it should run there. But `Nethermind.PerfTest` is excluded from `Nethermind.slnx` and currently has 126 compile errors against current `IBlockTree` / `IBlockFinder` interfaces (stale wrapper code). It never gets built in CI, so the analyzer never gets to bind that file in normal builds. That's why it didn't fire on you either — not because the analyzer is misconfigured, but because the project is broken upstream of analysis. Fixing PerfTest's interface drift is out of scope for this PR.

(Posted by Claude Opus 4.7 on @asdacap's behalf.)

src/Nethermind/Nethermind.Analyzers.Test/TaskCompletionSourceMustRunContinuationsAsynchronouslyAnalyzerTests.cs
```diff
@@ -0,0 +1,90 @@
...
+    [TestCase("new TaskCompletionSource<int>(TaskCreationOptions.RunContinuationsAsynchronously)")]
+    [TestCase("new TaskCompletionSource<int>(TaskCreationOptions.RunContinuationsAsynchronously | TaskCreationOptions.LongRunning)")]
+    public Task Construction_with_run_continuations_async_no_diagnostic(string expression) =>
+        Verify(WrapMethodBody($"var t = {expression};"));
+
+    [Test]
+    public Task Const_local_with_run_continuations_async_no_diagnostic() =>
+        Verify(WrapMethodBody("""
+            const TaskCreationOptions opts = TaskCreationOptions.RunContinuationsAsynchronously;
+            var t = new TaskCompletionSource(opts);
```
flcl42, line 38: Indeed why do we focus on `var` if we should not even use `var`s

benaadams (approved):

PR #11466, opened 2026-05-02
fixed half-empty trace result

### LINKED ISSUE #8380: Incomplete debug_traceBlockByNumber reply - holesky archive
**Description**
When trying to fetch debug_traceBlockByNumber, I get incomplete (invalid) json

**Steps to Reproduce**

````
curl localhost:8545 --header 'Content-Type: application/json' \
-d '{"jsonrpc":"2.0","method":"debug_traceBlockByNumber","params":["0x3553A7",{"tracer":"callTracer"}],"id":1}'
````

reply ends up with
````
"calls":[{"type":"CALL","from":"0x98075f191eb1358f2b02d30f00f5f80f06137fa2","to":"0x33130237606562fbb5a3ccec45e00a19abef97f6","value":"0x270f","gas":"0xb6811","gasUsed":"0x759e8","input":"0x","calls":[
````

and results in incomplete json

blocks 0x3553A6 or 0x3553A8 are ok

**Expected behavior**
Obtain complete trace (and valid json)

**Additional context**
- linux
- nethermind 1.31.5
- holesky archive

DarkLord017:
Fixes Closes Resolves #8380

## Changes
Added geth like streaming of results for debug module to tackle the issue of incomplete json being return on requests 

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Remarks


DarkLord017:
@claude review

DarkLord017:
@claude re-review

DarkLord017:
@claude review 
just a question are the regression tests sufficient , like how can i test that the output will be truncated in old while it will pass in new one id there a way to test it ?

DarkLord017:
> **Claude finished @DarkLord017's task in 1m 9s** —— [View job](https://github.com/NethermindEth/nethermind/actions/runs/25339214850)
> 
> ### Re-Review – Are the Regression Tests Sufficient?
> * [x]  Understand the question
> * [x]  Read test file and implementation
> * [x]  Answer: are tests sufficient + how to verify old code fails
> 
> ### Short Answer
> **Yes, the tests are sufficient.** They cover both test paths (STJ serialization and streaming) and would have caught the original bug. Here's the breakdown.
> 
> ### How the Tests Catch the Bug
> The original bug: `DebugRpcModule` returned a raw `IReadOnlyCollection<GethLikeTxTrace>` (a `List<T>`). System.Text.Json serialized it as a plain array:
> 
> ```json
> // OLD (broken) — missing result/txHash wrapper
> [
>   {"gas": 87700, "failed": false, "returnValue": "0x", "structLogs": [...]},
>   {"gas": 56141, ...}
> ]
> ```
> 
> The fix wraps each entry correctly:
> 
> ```json
> // NEW (correct) — Geth API format
> [
>   {"result": {"gas": 87700, ...}, "txHash": "0xb5a7..."},
>   {"result": {"gas": 56141, ...}, "txHash": "0xdb3d..."}
> ]
> ```
> 
> Three independent tests catch this:
> 
> **1. `TraceBlockSource` parameterized tests** (`Debug_traceBlock`, `Debug_traceBlockByNumber`, `Debug_traceBlockByHash`) These use `JsonElement.DeepEquals` against hardcoded expected JSON that includes `"result":{...},"txHash":"0x..."` on every entry. With the old code, the response lacked those wrappers so `DeepEquals` would return `false` → test fails.
> 
> **2. `Debug_traceBlock_json_rpc_request_returns_valid_json`** This explicitly asserts the structural contract:
> 
> ```cs
> Assert.That(entry.TryGetProperty("result", out _), Is.True, "Each entry must have 'result'");
> Assert.That(entry.TryGetProperty("txHash", out _), Is.True, "Each entry must have 'txHash'");
> ```
> 
> With the old code, `TryGetProperty("result")` would return `false` on what was actually a `GethLikeTxTrace` object (which has properties `gas`, `failed`, etc., not `result`/`txHash`) → test fails.
> 
> **3. `GethLikeTxTraceStreamingResult_WriteToAsync_produces_same_json_as_serializer`** This verifies that the streaming path (`WriteToAsync` via pipe) and the STJ path (`GethLikeTxTraceStreamingResultConverter.Write`) produce identical JSON. This catches any future divergence between the two paths.
> 
> ### How to Verify the Old Code Would Fail
> The easiest way is a quick local experiment:
> 
> ```cs
> // In DebugRpcModule.cs, temporarily change this...
> return ResultWrapper<GethLikeTxTraceStreamingResult>.Success(new GethLikeTxTraceStreamingResult(blockTrace));
> 
> // ...back to the old pattern (raw list):
> return ResultWrapper<IReadOnlyCollection<GethLikeTxTrace>>.Success(blockTrace);
> ```
> 
> Then run:
> 
> ```shell
> dotnet test src/Nethermind/Nethermind.JsonRpc.Test/ -c release -- --filter FullyQualifiedName~Debug_traceBlock
> ```
> 
> All three `TraceBlock*` test groups will fail with output like:
> 
> ```
> Expected: True  But was: False
> Actual response: {"jsonrpc":"2.0","result":[{"gas":87700,"failed":false,...}],"id":67}
> ```
> 
> The key difference: the old response has `result` at the top-level RPC level containing a bare array of trace objects. The new response has each array entry as `{"result":{...},"txHash":"0x..."}`.
> 
> ### One Remaining Unresolved Issue
> From the previous re-review: `using System.Buffers;` at line 5 of `GethLikeTxTraceStreamingResult.cs` is still unused. Nothing in the file references that namespace directly. This is a low-severity issue but worth cleaning up before merge.
> 
> ```diff
> -using System.Buffers;
> ```
> 
> [Fix this →](https://claude.ai/code?q=Remove%20unused%20%22using%20System.Buffers%3B%22%20from%20line%205%20of%20src%2FNethermind%2FNethermind.JsonRpc%2FModules%2FDebugModule%2FGethLikeTxTraceStreamingResult.cs%20on%20branch%20fix%2Ftrace&repo=NethermindEth/nethermind)
> 
> Overall the tests are well-structured. The `TraceBlockSource` cases are the primary regression guard (they encode exact expected output), while `Debug_traceBlock_json_rpc_request_returns_valid_json` acts as a structural sanity check and `WriteToAsync_produces_same_json_as_serializer` guards the streaming path specifically.

You are wrong the issue it fixes is that we were getting truncated output for many blocks in trace_ endpoints pls see the issue the thing u are telling is incorrect also system.buffers is req removing it code not compiles

DarkLord017:
@claude re-review

LukaszRozmej (commented):
I would love to have streaming done in a way that it would stream while generating the trace itself, using much much less memory.

DarkLord017:
> I would love to have streaming done in a way that it would stream while generating the trace itself, using much much less memory.

okay doing it

LukaszRozmej:
> > I would love to have streaming done in a way that it would stream while generating the trace itself, using much much less memory.
> 
> okay doing it

might be hard (can be done in subsequent PR)

svlachakis (approved):
please create an issue for @LukaszRozmej's suggestion on streaming and let's do it

LukaszRozmej (approved):

DarkLord017:
@claude review

DarkLord017:
@claude review

svlachakis (approved):

PR #11472, opened 2026-05-03
fix(snap-sync): drive FlatSnapTrieFactory init/finalize from SnapSyncRunner.Run

LukaszRozmej:
## Summary

Ports the snap-sync flake fix from #11443 onto @asdacap's snap+state dispatcher refactor (#11102), in the cleaner shape he suggested: drive `FlatSnapTrieFactory.Clear()` from `SnapSyncRunner.Run` boundaries instead of a lazy `EnsureDatabaseCleared` lock.

This PR is **stacked on top of `amirul/simple-sync-dispatcher`** so it can land alongside #11102.

## Background

#11443 root-caused a Windows-only `E2ESyncTests.SnapSync` flake to a race in `FlatSnapTrieFactory.EnsureDatabaseCleared`:

```csharp
if (_initialized) return;          // unlocked fast path
using (_lock.EnterScope())
{
    if (_initialized) return;
    _initialized = true;            // ← published BEFORE Clear() ran
    persistence.Clear();             // ← still iterating GetAllKeys / queueing Removes
}
```

Thread T2 could read `_initialized = true` via the unlocked fast path while T1 was still inside `Clear()`. T2 would then call `CreateStateTree` and start writing accounts. When T1's `Clear` batch finally disposed, every queued `Remove` (including the keys T2 had just committed) would land on the live DB, and the flat DB ended up missing entries the trie still referenced — surfacing as "Account in trie not found in flat" in the verifier.

#11443 fixed it by publishing `_initialized` _after_ `Clear()` and marking the field `volatile`. Validated as 0/300 stress runs vs. 8–15/60 before.

## What this PR does

@asdacap pointed out that on top of #11102's runner refactor, the lazy-init pattern can go away entirely — the snap-sync run has a single sequential entry/exit point in `SnapSyncRunner.Run`, so init/cleanup can be hooked there:

- Add two default-no-op methods to `ISnapTrieFactory`:
  - `EnsureInitialize()` — called once before any tree is created.
  - `FinalizeSync()` — called once after the dispatcher has drained.
- `FlatSnapTrieFactory` implements them with `persistence.Clear()` and `persistence.Flush()` respectively. The lock and the `_initialized` flag are deleted.
- `SnapSyncRunner.Run` brackets `dispatcher.Run(token)` with `EnsureInitialize` (before) and `FinalizeSync` (in `finally`).
- `PatriciaSnapTrieFactory`, `TestSnapTrieFactory`, and the other test factories pick up the empty default implementations and don't need changes.

Because `SimpleDispatcher.Run` (introduced by #11102) only returns once every batch handler — and therefore every `ISnapTree.Dispose` — has completed, neither hook needs internal locking. This also obviates the in-flight tree drain that #11443 needed on master (commit `12a66f1e1b`).

## Stress reproducer

The same `SnapSync_StressRepro` test from #11443 (cherry-picked here) runs 30 iterations × 2 Flat fixtures = 60 cases, no `[Retry]`, `[Explicit]`. Local Windows results on this branch:

| Run | `Verification failed` | Test failures |
|---|---:|---:|
| Run 1 | 0 | 0 |
| Run 2 | 0 | 0 |
| Run 3 | 0 | 0 |
| **Total** | **0/180** | **0/180** |

Plus full Flat test suite (292 tests, 4 skipped) and `RecreateState*` (20 tests) and `StateSyncFeed*` (834 tests) all pass.

## Test plan

- [x] Solution builds clean (0 warnings, 0 errors).
- [x] `Nethermind.State.Flat.Test` passes (292 tests).
- [x] `RecreateStateFromAccountRanges` / `RecreateStateFromStorageRanges` pass (20 tests).
- [x] `StateSyncFeed*` passes (834 tests).
- [x] `SnapSync_StressRepro`: 0 failures across 3 stress runs (180 invocations).

### DIFF
--- a/src/Nethermind/Nethermind.State.Flat/Sync/Snap/FlatSnapTrieFactory.cs
+++ b/src/Nethermind/Nethermind.State.Flat/Sync/Snap/FlatSnapTrieFactory.cs
@@ -14,18 +14,23 @@ namespace Nethermind.State.Flat.Sync.Snap;
 /// <summary>
 /// ISnapTrieFactory implementation for flat state storage.
 /// Uses IPersistence to create reader/writeBatch per tree for proper resource management.
+/// EnsureInitialize/FinalizeSync are driven by the snap-sync runner at start/end of the run,
+/// so they don't need internal locking — they're never called concurrently with CreateXxxTree.
 /// </summary>
 public class FlatSnapTrieFactory(IPersistence persistence, ISyncConfig syncConfig, ILogManager logManager) : ISnapTrieFactory
 {
     private readonly ILogger _logger = logManager.GetClassLogger<FlatSnapTrieFactory>();
-    private readonly Lock _lock = new();
 
-    private bool _initialized = false;
+    public void EnsureInitialize()
+    {
+        if (_logger.IsInfo) _logger.Info("Clearing database");
+        persistence.Clear();
+    }
+
+    public void FinalizeSync() => persistence.Flush();
 
     public ISnapTree<PathWithAccount> CreateStateTree()
     {
-        EnsureDatabaseCleared();
-
         IPersistence.IPersistenceReader reader = persistence.CreateReader(ReaderFlags.Sync);
         IPersistence.IWriteBatch writeBatch = persistence.CreateWriteBatch(StateId.Sync, StateId.Sync, WriteFlags.DisableWAL);
         return new FlatSnapStateTree(reader, writeBatch, syncConfig.EnableSnapDoubleWriteCheck, logManager);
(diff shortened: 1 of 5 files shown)


LukaszRozmej:
Both Low findings addressed in 0a94bfd3b9:

- **Lifecycle ordering test:** added `SnapSyncRunnerTests` with 3 cases (happy path, dispatcher throws, dispatcher cancelled) — all assert `EnsureInitialize` → dispatcher → `FinalizeSync` ordering. Required exposing an internal `Func<CancellationToken, Task>` ctor on `SnapSyncRunner` so the runner can be exercised without standing up a real `SimpleDispatcher`. `InternalsVisibleTo("Nethermind.Synchronization.Test")` already in scope.
- **`FinalizeSync` on cancel/throw:** kept the always-flush behaviour (idempotent, durability-safer default as you noted) and added a one-line comment in the `finally` block making the intent explicit. Cancel and throw paths are both covered by the new tests.

Also trimmed the multi-line XML doc comments on the new `ISnapTrieFactory` hooks to single-line comments per the repo's coding-style rule.

asdacap (approved):

svlachakis (approved):

PR #11477, opened 2026-05-04
refactor: serialise FinalizeSync via captured per-round pivot

asdacap:
## Summary

Stacks on top of #11102. Addresses the `GetPivotHeader()` race flagged by `svlachakis` in #11102 (root cause discussed in #11457 / #11458): `TreeSync.SaveNode` previously called `_store.FinalizeSync(GetPivotHeader())` inline on the `IsRoot` branch, while parallel `HandleResponse` workers were still writing nodes — and `GetPivotHeader()` is mutating, so the pivot rotation can race with the root commit and corrupt FlatDB `CurrentState`.

This PR pins the pivot per round in the runner and runs all post-sync work at the quiescent point after the dispatcher has fully drained.

## What changed

- **Pivot pinned per round.** `TreeSync.ResetStateRootToBestSuggested` now returns the `BlockHeader` it used. `StateSyncRunner.RunStateSyncRounds` captures that header at the start of each round and, when the round commits the root, uses *that* captured value for `FinalizeSync` — no second `GetPivotHeader()` call.
- **`FinalizeSync` out of `SaveNode`.** The `_store.FinalizeSync(pivotHeader)` call is removed from `TreeSync.SaveNode`'s `IsRoot` branch. The runner invokes `treeSync.FinalizeSync(finalPivot)` once after the outer loop exits via `IsRootComplete`.
- **`VerifyPostSyncCleanUp` out of `IsSyncRoundFinished`.** The two cleanup-firing branches inside the predicate (`_rootSaved == 1` and `NodeExists`) no longer call cleanup as a side effect. The runner calls `treeSync.VerifyPostSyncCleanUp()` once after the loop.
- **`STATE SYNC FINISHED:…` log moved** to the runner so all post-sync reporting is in one place.
- **`_syncStateLock` removed.** It existed to serialise `HandleResponse` (read) against `ResetStateRoot` / `CleanupMemory` (write). With the cleanup moved out and the writers only running from the runner outside `dispatcher.Run`, `SimpleDispatcher.Run`'s drain (`semaphore.WaitAsync(CancellationToken.None)` × maxThreads) is the sole synchronisation barrier — strict happens-before between the last `HandleResponse` and the next reset/cleanup.

## What didn't change

- Round-termination is still signalled by `StateSyncFeed.PrepareRequest` returning null when `treeSync.IsSyncRoundFinished()` is true — natural dispatcher exit, no watcher / per-round CTS.
- Outer break stays on `treeSync.IsRootComplete`. Tightening to the strict `IsRootSaved` (`_rootSaved == 1`) breaks legitimate completion paths (empty-state syncs, snap-prepopulated root) where the `IsRoot` branch in `SaveNode` never fires. The `NodeExists` over-reporting concern stays as #11457 / #11458 territory.
- Other concerns from the race-analysis comment (`GetPivotHeader()` null-deref at `TreeSync.cs:355`, `SyncCompleted` event timing) are pre-existing and untouched here.

## Test plan

- [x] `Nethermind.Synchronization.Test` filtered to StateSync + SnapSync: 932/940 pass, 8 skips (hash-db).
- [ ] E2E sync tests (CI / PR owner).

### DIFF
--- a/src/Nethermind/Nethermind.Synchronization/FastSync/StateSyncRunner.cs
+++ b/src/Nethermind/Nethermind.Synchronization/FastSync/StateSyncRunner.cs
@@ -80,7 +80,7 @@ public class StateSyncRunner(
             // in StateNodes mode and close to head.
             await StateSyncPrecursorWait(token);
 
-            BlockHeader? roundPivot = treeSync.ResetStateRootToBestSuggested(SyncFeedState.Dormant);
+            BlockHeader? roundPivot = treeSync.ResetStateRootToBestSuggested();
             if (roundPivot is null) continue;
 
             await stateSyncDispatcher.Run(token);
--- a/src/Nethermind/Nethermind.Synchronization/FastSync/StateSyncRunner.cs
+++ b/src/Nethermind/Nethermind.Synchronization/FastSync/StateSyncRunner.cs
@@ -96,15 +96,10 @@ public class StateSyncRunner(
 
         if (finalPivot is null) return;
 
-        if (_logger.IsInfo) _logger.Info("State sync completed.");
+        if (_logger.IsInfo) _logger.Info($"STATE SYNC FINISHED:{Metrics.StateSyncRequests}, {Metrics.SyncedStateTrieNodes}");
 
-        // FinalizeSync used to be invoked inside TreeSync.SaveNode's IsRoot branch, where
-        // GetPivotHeader's mutating side effect could race with concurrent root-saves
-        // (svlachakis #11457/#11458). Calling it here, with the captured per-round pivot,
-        // serialises the pivot read against in-flight handlers and pins the pivot value.
         treeSync.VerifyPostSyncCleanUp();
-        try { treeSync.FinalizeSync(finalPivot); }
-        catch (ObjectDisposedException) { }
+        treeSync.FinalizeSync(finalPivot);
 
         if (syncConfig.VerifyTrieOnStateSyncFinished)
             verifyTrieStarter?.TryStartVerifyTrie(finalPivot);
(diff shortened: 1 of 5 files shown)


asdacap:
@claude reevaluate

src/Nethermind/Nethermind.Synchronization/FastSync/StateSyncRunner.cs
```diff
@@ -64,12 +64,19 @@ public async Task Run(CancellationToken token)
         {
             // Clean shutdown — swallow so Synchronizer doesn't log "State sync failed".
         }
+        catch (ObjectDisposedException)
+        {
+            // DBs / stores can be torn down while sync is in progress on shutdown — same
+            // semantics as cancellation: swallow rather than logging "State sync failed".
+        }
```
LukaszRozmej: Can this be prevented and order be preserved?

src/Nethermind/Nethermind.Synchronization/FastSync/TreeSync.cs
```diff
@@ -66,10 +66,11 @@ public class TreeSync
...
 
-        // This is not exactly a lock for read and write, but a RWLock serves it well. It protects the five field
-        // below which need to be cleared atomically during reset root, hence the write lock, while allowing
-        // concurrent request handling with the read lock.
-        private readonly ReaderWriterLockSlim _syncStateLock = new();
+        // The five fields below were previously guarded by an RWLock so HandleResponse (read)
+        // could run concurrently with ResetStateRoot / CleanupMemory (write). The lock is no
+        // longer needed: SimpleDispatcher.Run drains all in-flight workers before returning,
+        // and StateSyncRunner only calls reset/cleanup outside dispatcher.Run, so the drain is
+        // the sole synchronisation barrier.
```
LukaszRozmej: Can be more concise, don't need history.

LukaszRozmej (dismissed):

LukaszRozmej (approved):

svlachakis (approved):

PR #11479, opened 2026-05-04
Fixes for RLP integers and bool decoding

alexb5dh:
## Changes

- Simplifies `DecodeInt` as `(int)DecodeUInt()`.
- Adds `DecodePositiveInt` with sign check, similar to `DecodePositiveLong`.
- Uses `DecodePositiveInt` and `DecodeUInt` instead of `DecodeInt` where possible.
   Ideally all usages should be removed, but I'm not 100% sure about some.
- Ensures canonical form for single-byte encoding of all integers and for boolean.
   See some of `Decode_integer_invalid`, `Decode_bool_invalid` test cases.
- Includes position into all related errors.
- Adds tests for all affected methods with valid and invalid cases.

#### Requires testing

- [x] Yes

- [x] Yes


src/Nethermind/Nethermind.Consensus.Clique/SnapshotDecoder.cs
```diff
@@ -68,7 +68,7 @@ private static (int contentLength, int signersLength, int votesLength, int tally
         private static SortedList<Address, long> DecodeSigners(ref Rlp.ValueDecoderContext decoderContext)
         {
             decoderContext.ReadSequenceLength();
-            int length = decoderContext.DecodeInt();
+            int length = decoderContext.DecodePositiveInt();
```
flcl42, line 71: Is zero positive?

alexb5dh: Naming is selected identical to `DecodePositiveLong`, can rename both

flcl42: Ignore my suggestion please, we will change it when switch to uints

flcl42 (approved):

alexb5dh:
@claude re-review

LukaszRozmej (approved):

benaadams (approved):

PR #11484, opened 2026-05-04
TransactionForRpc improvements

### LINKED ISSUE #10547: TransactionForRpc.EnsureDefaults not called consistently before ToTransaction
## Problem

`TransactionForRpc.EnsureDefaults(gasCap)` bundles three responsibilities into one method:

1. **Cap gas** to `gasCap` (`Gas = Math.Min(gasCap, Gas)`)
2. **Set default gas** to `gasCap` when `Gas` is null/0
3. **Set default sender** (`From ??= Address.Zero`)

Not all code paths that call `ToTransaction()` call `EnsureDefaults()` first. This means some RPC methods may have missing defaults (particularly `From` and `Gas`), and the inconsistency makes it easy for future code to introduce the same gap.

### `SimulateTxExecutor.Prepare()` — no `EnsureDefaults`
**File:** `Nethermind.JsonRpc/Modules/Eth/SimulateTxExecutor.cs:55`
Affects: `eth_simulateV1`, `debug_simulateV1`, `trace_simulateV1`

Calls `callTransactionModel.ToTransaction()` without `EnsureDefaults()`. Gas cap is handled separately via `_blockchainBridge.Simulate(..., gasCap)`, but:
- If `From` is null → `SenderAddress` is set to `Address.Zero` only by `ToTransaction()`'s fallback (`From ?? Address.Zero`) — this happens to work, but it's implicit and not documented
- If `Gas` is null → falls back to hardcoded `90_000` in `ToTransaction()` instead of using `gasCap` or block gas limit

### `ProofRpcModule.proof_call()` — no `EnsureDefaults`
**File:** `Nethermind.JsonRpc/Modules/Proof/ProofRpcModule.cs:68`
Affects: `proof_call`

Calls `tx.ToTransaction()` without `EnsureDefaults()`. Has manual fallbacks after conversion:
```csharp
transaction.SenderAddress ??= Address.Zero;          // line 74
if (transaction.GasLimit == 0)                        // line 76
    transaction.GasLimit = callHeader.GasLimit;       // line 78 — no gas cap!
```
This means `proof_call` has **no gas cap enforcement** at all — if a caller sends `gas: 0`, the limit is set to the full block gas limit, ignoring `GasCap`.

### `DebugBridge.GetBundleTrace()` — relies on caller
**File:** `Nethermind.JsonRpc/Modules/DebugModule/DebugBridge.cs:227`
Affects: `debug_traceCallMany`

Currently safe because the only caller (`DebugRpcModule.TraceCallMany`) calls `PrepareTransactions()` → `EnsureDefaults()` on the same `TransactionForRpc` objects before passing them to the bridge. But `IDebugBridge.GetBundleTraces` is a public interface method — any new caller could forget to call `EnsureDefaults` upstream.

## Suggested fix

Consider splitting `EnsureDefaults` into orthogonal concerns, e.g.:
- **`CapGasLimit(gasCap)`** — already extracted to `TransactionExtensions` for `Transaction` in #10457
- **Default initialization** — `From ??= Address.Zero`, `Gas ??= blockGasLimit` — could be part of `ToTransaction()` itself or a separate `SetDefaults()` method

This would make each call site explicit about what it needs, and `ToTransaction()` could enforce that defaults are always set regardless of whether the caller remembered to call `EnsureDefaults`.

## Related

- #10457 — Enforce GasCap across debug and trace RPC methods
- #9750 — Original gas cap enforcement issue

svlachakis:
Fixes Closes Resolves #10547 

## Changes

Removes `TransactionForRpc.EnsureDefaults` entirely and folds defaulting + gas-cap enforcement into `ToTransaction(gasCap)`. The "two methods, easy to forget one" pattern that produced the original bugs is structurally gone.

## What changed

- `TransactionForRpc.EnsureDefaults` (abstract) and its overrides in `LegacyTransactionForRpc` and Optimism's `DepositTransactionForRpc` are deleted.
- `ToTransaction(validateUserInput, gasCap, spec)` is now the single conversion entry point. It always sets `SenderAddress` (defaults to `Address.Zero`) and `GasLimit` (defaults to `gasCap`, or `long.MaxValue` if unset; capped at `gasCap` when provided).
- The hardcoded `Gas ?? 90_000` magic in `LegacyTransactionForRpc.ToTransaction` is removed.
- `IDebugBridge.GetBundleTraces` takes `long? gasCap`; the bridge converts internally. `DebugRpcModule.PrepareTransactions` is removed.

## Bugs closed

- **`proof_call` had no gas-cap enforcement** — a caller sending `gas: 0` got the full block gas limit, ignoring `GasCap`. Now properly capped via injected `IJsonRpcConfig`.
- **`SimulateTxExecutor.Prepare` skipped defaulting** — `gas: null` fell back to a hardcoded 90,000 instead of `gasCap`. Now consistent with all other paths.
- **`IDebugBridge.GetBundleTraces` relied on the caller** to pre-prepare transactions — now self-contained.

## Behavior change worth flagging

`gas: 0` is now treated as a literal request (intrinsic-gas error), not as a "missing" signal that defaults to `gasCap`. This matches Geth and is required by the `multicall-insufficient-gas-38013` Hive spec test. `gas: null` is the only "missing" signal.

`debug_traceCallMany` and `eth_simulateV1` need different default-gas policies for `gas: null` (gasCap vs. block gas limit). This is now expressed by a localized `call.Gas ??= jsonRpcConfig.GasCap` in `DebugRpcModule.TraceCallManyWithOverrides`, with a comment explaining why `eth_simulateV1` deliberately omits it.

- [ ] Optimization
- [x] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

## Testing

Hive tests Job: https://github.com/NethermindEth/nethermind/actions/runs/25330905724

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/AccessListTransactionForRpc.cs
+++ b/src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/AccessListTransactionForRpc.cs
@@ -36,9 +36,9 @@ public class AccessListTransactionForRpc : LegacyTransactionForRpc, IFromTransac
         V = YParity ?? 0;
     }
 
-    public override Result<Transaction> ToTransaction(bool validateUserInput = false, IReleaseSpec? spec = null)
+    public override Result<Transaction> ToTransaction(bool validateUserInput = false, long? gasCap = null, IReleaseSpec? spec = null)
     {
-        Result<Transaction> baseResult = base.ToTransaction(validateUserInput, spec);
+        Result<Transaction> baseResult = base.ToTransaction(validateUserInput, gasCap, spec);
         if (baseResult.IsError) return baseResult;
 
         Transaction tx = baseResult.Data;
--- a/src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/BlobTransactionForRpc.cs
+++ b/src/Nethermind/Nethermind.Facade/Eth/RpcTransaction/BlobTransactionForRpc.cs
@@ -35,7 +35,7 @@ public class BlobTransactionForRpc : EIP1559TransactionForRpc, IFromTransaction<
         BlobVersionedHashes = transaction.BlobVersionedHashes ?? [];
     }
 
-    public override Result<Transaction> ToTransaction(bool validateUserInput = false, IReleaseSpec? spec = null)
+    public override Result<Transaction> ToTransaction(bool validateUserInput = false, long? gasCap = null, IReleaseSpec? spec = null)
     {
         if (BlobVersionedHashes is null || BlobVersionedHashes.Length == 0)
             return RpcTransactionErrors.AtLeastOneBlobInBlobTransaction;
@@ -55,7 +55,7 @@ public class BlobTransactionForRpc : EIP1559TransactionForRpc, IFromTransaction<
         if (validateUserInput && MaxFeePerBlobGas?.IsZero == true)
             return RpcTransactionErrors.ZeroMaxFeePerBlobGas;
 
-        Result<Transaction> baseResult = base.ToTransaction(validateUserInput, spec);
+        Result<Transaction> baseResult = base.ToTransaction(validateUserInput, gasCap, spec);
         if (!baseResult) return baseResult;
 
         Transaction tx = baseResult.Data;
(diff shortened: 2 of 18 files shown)


svlachakis:
@claude review again

rubo (approved):

DarkLord017 (approved):

PR #11485, opened 2026-05-04
ci(expb): record run dimensions in benchmark metrics

grubbyTimes:
`expb-metrics.env` now carries `BRANCH`, `CLEAN_BRANCH`, `SHA`, `STATE_LAYOUT`, `PAYLOAD_SET`, `DELAY_SECONDS`, `IMAGE`, `EXPB_VERSION`, plus run/scenario context — written by both single- and multi-image analyze steps before percentiles. `EXPB_VERSION` comes from `uv tool list` post-install. Existing `load_metric` consumers grep specific keys, so extra lines are ignored and master cache (`v3`) stays compatible.

### DIFF
--- a/.github/workflows/run-expb-reproducible-benchmarks.yml
+++ b/.github/workflows/run-expb-reproducible-benchmarks.yml
@@ -603,9 +603,13 @@ jobs:
             "${YQ}" -i ".scenarios.\"${scenario_key}\".extra_flags += [\"${flag}\"]" "${RENDERED_CONFIG_FILE}"
           done < "${flags_file}"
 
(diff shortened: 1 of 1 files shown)


kamilchodola:
@claude review - including security check

AnkushinDaniil (approved):

stdevMac:
@grubbyTimes @kamilchodola stale-PR check: the EXPB workflow has been rewritten ~6 times since May so this conflicts heavily, and the report paths now pull run dimensions from the resolve-job outputs. Still want the metrics.env metadata approach, or OK to close? If there's no reply in the next few days, we'll close this as part of the stale-PR cleanup — it can always be reopened or resubmitted.

PR #11487, opened 2026-05-04
fix: stack overflow in debug_trace

deffrian:
Fixes Closes Resolves #

#8380

## Changes

- Allow more nested calls
- Iterative converter

- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


src/Nethermind/Nethermind.Serialization.Json/EthereumJsonSerializer.cs
```diff
@@ -18,15 +18,19 @@ namespace Nethermind.Serialization.Json
 {
     public sealed class EthereumJsonSerializer : IJsonSerializer
     {
-        public const int DefaultMaxDepth = 128;
+        // Must accommodate the deepest possible callTracer output: each NativeCallTracerCallFrame
+        // contributes ~2 JSON levels (object + "calls" array), the EVM allows up to MaxCallDepth=1024
+        // (Yellow Paper / Nethermind.Evm.VirtualMachine.MaxCallDepth), plus a few levels of JSON-RPC
+        // envelope. 4096 leaves comfortable headroom.
+        public const int DefaultMaxDepth = 4096;
```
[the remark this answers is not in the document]
deffrian: @claude We have input size limit, can you verify?

src/Nethermind/Nethermind.Evm.Test/Tracing/GethLikeCallTracerTests.cs
```diff
@@ -633,6 +636,103 @@ public void Test_CallTrace_TopLevelCreate_WithLog_DeployContractFailure_LogsClea
...
+            Throws.Nothing,
+            $"chain of {boundary} frames must serialize");
+
+        using NativeCallTracerCallFrame justOver = BuildLinearCallChain(boundary + 1);
+        Assert.That(
+            () => JsonSerializer.Serialize(new GethLikeCustomTrace { Value = justOver }, EthereumJsonSerializer.JsonOptions),
+            Throws.TypeOf<JsonException>()
+                .With.InnerException.TypeOf<InvalidOperationException>()
+                .And.InnerException.Message.EqualTo(
+                    $"CurrentDepth ({EthereumJsonSerializer.DefaultMaxDepth}) is equal to or larger than the maximum allowed depth of {EthereumJsonSerializer.DefaultMaxDepth}. Cannot write the next JSON object or array."),
```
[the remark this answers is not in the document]
deffrian: Better to leave as this. It's not hard to fix

src/Nethermind/Nethermind.Serialization.Json/EthereumJsonSerializer.cs
LukaszRozmej, line 25: Is it potential DOS vector though?

src/Nethermind/Nethermind.JsonRpc/IJsonRpcConfig.cs
```diff
@@ -157,7 +157,7 @@ public interface IJsonRpcConfig : IConfig
     [ConfigItem(Description = "The max number of JSON-RPC requests in a batch.", DefaultValue = "1024")]
     int MaxBatchSize { get; set; }
 
-    [ConfigItem(Description = "The maximum depth of JSON response object tree.", DefaultValue = "128")]
+    [ConfigItem(Description = "The maximum depth of JSON response object tree.", DefaultValue = "4096")]
```
LukaszRozmej, line 160: Should we have such large default? What about DOS vectors?

src/Nethermind/Nethermind.Serialization.Json/EthereumJsonSerializer.cs
deffrian: We have limit on input size. I'm not sure how depth can be a dos vector

src/Nethermind/Nethermind.JsonRpc/IJsonRpcConfig.cs
deffrian: We have to have that if we want to serve debug_trace. I see three options:
1. Use different serializers for requests and responses. Potentially a lot of work?
2. Put small limit by default and ask users to lift it when they need. Bad ux and dos vector for these users?
3. Keep it this way. As i said I can't see how depth of requests can be a dos vector. We will just drop such requests as malformed and we have input size cap

svlachakis:
For reference: Geth has no JSON depth limit at all - only a 5 MB body cap (rpc/http.go:39, defaultBodyLimit). It uses Go's stdlib encoding/json, which has no `MaxDepth` concept, and has run that way in production for years without depth-based DOS reports.

On our side, `System.Text.Json`'s parser is iterative, so a deeply nested request won't stack-overflow Nethermind regardless of MaxDepth - at worst it consumes CPU/memory bounded by the request body size. So the DOS surface here is body size, not depth, and we're already capping that.

That said, the response and request paths have different needs: only `callTracer` actually requires deep output. We can keep the input side strict and still ship the fix by splitting the limits:

`JsonSerializerOptions.MaxDepth` = 128 — parser / request side, unchanged.
`JsonWriterOptions.MaxDepth` = 4096 — writer / response side, where the trace needs headroom.

The PR already wires `_maxDepth` through `CreateWriterOptions`, so it's a small change: take two values instead of one. That keeps Geth-equivalent permissiveness on output without relaxing the input ceiling - addresses both points in the thread.

@LukaszRozmej wdyt?

LukaszRozmej:
> For reference: Geth has no JSON depth limit at all - only a 5 MB body cap (rpc/http.go:39, defaultBodyLimit). It uses Go's stdlib encoding/json, which has no `MaxDepth` concept, and has run that way in production for years without depth-based DOS reports.
> 
> On our side, `System.Text.Json`'s parser is iterative, so a deeply nested request won't stack-overflow Nethermind regardless of MaxDepth - at worst it consumes CPU/memory bounded by the request body size. So the DOS surface here is body size, not depth, and we're already capping that.
> 
> That said, the response and request paths have different needs: only `callTracer` actually requires deep output. We can keep the input side strict and still ship the fix by splitting the limits:
> 
> `JsonSerializerOptions.MaxDepth` = 128 — parser / request side, unchanged. `JsonWriterOptions.MaxDepth` = 4096 — writer / response side, where the trace needs headroom.
> 
> The PR already wires `_maxDepth` through `CreateWriterOptions`, so it's a small change: take two values instead of one. That keeps Geth-equivalent permissiveness on output without relaxing the input ceiling - addresses both points in the thread.
> 
> @LukaszRozmej wdyt?

hard to say, but maybe let's do minimal change for now. You can go stack too deep with `MaxDepth` in default `System.Text.Json` code

deffrian:
@LukaszRozmej @svlachakis We actually have an input depth limit of 64. I just added a test that confirms that. 
This is the reason: https://github.com/NethermindEth/nethermind/blob/db8b6e9c7303769b3339b82321355510ab57dc08/src/Nethermind/Nethermind.JsonRpc/JsonRpcProcessor.cs#L288

svlachakis:
@claude review again, take into account @LukaszRozmej's and @svlachakis' comments

svlachakis (approved):
@deffrian i think comment from Claude is worth fixing

deffrian:
@svlachakis imo it's better to keep as is. It's useful to have this depth just in case something changes, so we would not have any degradations.

LukaszRozmej (approved):

svlachakis (approved):

PR #11489, opened 2026-05-04
RPC: `admin_*` peer-management endpoints improvements

### LINKED ISSUE #11490: `admin_addPeer` / `admin_removePeer` / `admin_addTrustedPeer` / `admin_removeTrustedPeer` diverge from Geth
| Endpoint | Problem |
|----------|---------|
| `admin_addPeer` | Returns `string` not `bool`. No-flag form does a one-shot dial — no static membership, no maintained reconnect. `addToStaticNodes=true` also writes `static-nodes.json` (Geth never persists from RPC). |
| `admin_removePeer` | Returns `string` not `bool`. No-flag form removes from peer pool only, leaves entry in `static-nodes.json`. Returns `Fail` on unknown peer (Geth is idempotent). |
| `admin_addTrustedPeer` | Always writes `trusted-nodes.json` (hardcoded `updateFile: true`). No opt-out. Geth never persists from RPC. |
| `admin_removeTrustedPeer` | Always writes `trusted-nodes.json`. Returns `Fail` on unknown peer. |
| All four | Malformed enode throws → opaque internal RPC error instead of `-32602 invalid enode: ...`. Validation triggers DNS lookup. |

---

## Expected

Default behavior matches Geth's `Server.{AddPeer,RemovePeer,AddTrustedPeer,RemoveTrustedPeer}` in `go-ethereum/p2p/server.go:237-286`:

- in-memory only
- idempotent
- no disk write
- `bool` return
- clean `-32602` on malformed enode

A consistent opt-in `persistent` flag across all four endpoints replaces the current ad-hoc disk-write behavior.

svlachakis:
Fixes Closes Resolves https://github.com/NethermindEth/nethermind/issues/11490

### Wire-level breaking changes (all four endpoints)

- Return type `string → bool` on `addPeer/removePeer` (already `bool` on the trusted pair)
- Param renamed: `addToStaticNodes → persistent`, `removeFromStaticNodes → persistent` (consistent across all four)
- `addTrustedPeer` and `removeTrustedPeer` gained the new optional `persistent` parameter

### Default-semantic changes (the same call now does something different)

- `admin_addPeer(enode)`: now maintains the connection (Geth-equivalent); was a one-shot dial
- `admin_removePeer(enode)`: now removes from static set + disconnects + idempotent; was pool-only and reported failure on unknown peers
- `admin_addTrustedPeer(enode)`: no longer writes to `trusted-nodes.json` by default
- `admin_removeTrustedPeer(enode)`: no longer writes to `trusted-nodes.json` by default; now idempotent on unknown peers

### Validation (all four)

- Malformed enode now returns `-32602 InvalidParams` with `"invalid enode: ..."` instead of an opaque internal RPC error.

### Further Refactoring

- Replaced Enode.IsEnode pre-check with TryParseAsNetworkNode/TryParseAsEnode helpers wrapping full-parse construction in try/catch (catches valid-scheme/invalid-content inputs)
- Plumbed IJsonRpcConfig-derived CancellationToken through
- `IStaticNodesManager`/`ITrustedNodesManager`/`NodesManager.SaveFileAsync`; dropped string-typed overloads of `IStaticNodesManager.AddAsync`/`RemoveAsync`/`IsStatic` (test-only, replaced with `NetworkNode`-typed) to eliminate the double-parse + double-DNS-lookup; consolidated 4 single-method malformed-enode tests into one parameterized `[TestCase]` covering both scheme-fail and content-fail paths
- Added _subscriptionManager null-guard, fixed field-assignment ordering, and removed the `enode.ToString()` round-trip in `TrustedNodesManager.RemoveAsync`

---

## Changes

---

### `admin_addPeer`

| Behavior | Geth (only mode) | Nethermind BEFORE — default (`addToStaticNodes=false`) | Nethermind BEFORE — `addToStaticNodes=true` | Nethermind AFTER — default (`persistent=false`) | Nethermind AFTER — `persistent=true` |
|----------|------------------|---------------------------------------------------------|---------------------------------------------|--------------------------------------------------|--------------------------------------|
| Adds to in-memory static set | ✅ | ❌ pool dict only | ✅ | ✅ matches Geth | ✅ |
| Maintains connection (auto-redial) | ✅ | ❌ one-shot dial | ✅ | ✅ matches Geth | ✅ |
| Writes to `static-nodes.json` | ❌ | ❌ | ✅ | ❌ matches Geth | ✅ (NM extension) |
| Validates enode upfront | ✅ clean error | ❌ throws on bad input → opaque | ❌ same | ✅ `Enode.IsEnode` → `-32602` | ✅ same |
| Return type | `bool` | `string` (echoes enode) | `string` | `bool` matches Geth | `bool` |

---

### `admin_removePeer`

| Behavior | Geth | NM BEFORE — default (`removeFromStaticNodes=false`) | NM BEFORE — `removeFromStaticNodes=true` | NM AFTER — default | NM AFTER — `persistent=true` |
|----------|------|------------------------------------------------------|------------------------------------------|--------------------|------------------------------|
| Removes from in-memory static set | ✅ | ❌ pool only | ✅ | ✅ matches Geth | ✅ |
| Disconnects active session | ✅ | ✅ via direct `TryRemove` | ✅ via event chain | ✅ via event chain + direct `TryRemove` fallback | ✅ |
| Removes non-static (discovered) peer too | ✅ | ✅ | ❌ static set only | ✅ matches Geth (idempotent fallback) | ✅ |
| Writes to `static-nodes.json` | ❌ | ❌ | ✅ | ❌ matches Geth | ✅ (NM extension) |
| Idempotent on unknown peer | ✅ returns `true` | ❌ Fail | ❌ Fail | ✅ returns `true` matches Geth | ✅ |
| Validates enode upfront | ✅ | ❌ throws | ❌ throws | ✅ `-32602` | ✅ |
| Return type | `bool` | `string` | `string` | `bool` matches Geth | `bool` |

---

### `admin_addTrustedPeer`

| Behavior | Geth | NM BEFORE | NM AFTER — default | NM AFTER — `persistent=true` |
|----------|------|-----------|--------------------|------------------------------|
| Adds to in-memory trusted set | ✅ | ✅ | ✅ | ✅ |
| Writes to `trusted-nodes.json` | ❌ | ✅ always (`updateFile: true` hardcoded) | ❌ matches Geth | ✅ (NM extension) |
| Synchronously inserts into peer pool | ❌ relies on dialer | ✅ `_peerPool.GetOrAdd` | ✅ kept (NM nuance — avoids 1s `FeedFromNodeSource` throttle) | ✅ kept |
| Idempotent on already-trusted | ✅ | ✅ via `IsTrusted` short-circuit | ✅ via ignoring `AddAsync` return | ✅ |
| Validates enode upfront | ✅ | ❌ throws | ✅ `-32602` | ✅ |
| Return type | `bool` | `bool` | `bool` matches Geth | `bool` |

---

### `admin_removeTrustedPeer`

| Behavior | Geth | NM BEFORE | NM AFTER — default | NM AFTER — `persistent=true` |
|----------|------|-----------|--------------------|------------------------------|
| Removes from in-memory trusted set | ✅ | ✅ | ✅ | ✅ |
| Writes to `trusted-nodes.json` | ❌ | ✅ always (`updateFile: true` hardcoded) | ❌ matches Geth | ✅ (NM extension) |
| Disconnects active session | ❌ stays connected, loses privilege | ✅ via event chain (`NodeRemoved → PeerPool.TryRemove → MarkDisconnected`) | ✅ kept (NM-specific, deliberate) | ✅ kept |
| Idempotent on unknown peer | ✅ returns `true` | ❌ Fail | ✅ returns `true` matches Geth | ✅ |
| Validates enode upfront | ✅ | ❌ throws | ✅ `-32602` | ✅ |
| Return type | `bool` | `bool` | `bool` matches Geth | `bool` |

---

### Nethermind extensions Geth doesn't have

| Extension | Endpoints | Effect |
|-----------|-----------|--------|
| `persistent: true` flag | `admin_addPeer`, `admin_removePeer` | Also write/remove the entry to/from `static-nodes.json` so it survives restart |
| `persistent: true` flag | `admin_addTrustedPeer`, `admin_removeTrustedPeer` | Also write/remove the entry to/from `trusted-nodes.json` so trust survives restart |
| Synchronous pool insertion | `admin_addTrustedPeer` | Avoids the up-to-1s `PeerPool.FeedFromNodeSource` throttle delay; the peer is in the pool the moment the RPC returns |
| Disconnect on untrust | `admin_removeTrustedPeer` | Geth leaves the peer connected after untrusting; Nethermind disconnects via the existing `NodeRemoved` event chain — operationally cleaner ("force reconnect under non-trusted rules") |

---

- [x] Optimization
- [x] Refactoring
- [x] Documentation update
- [ ] Build-related changes
- [ ] Other: _Description_

#### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

#### Notes on testing

- [x] Yes
- [ ] No

### DIFF
--- a/src/Nethermind/Nethermind.JsonRpc/Modules/Admin/AdminRpcModule.cs
+++ b/src/Nethermind/Nethermind.JsonRpc/Modules/Admin/AdminRpcModule.cs
@@ -91,69 +91,53 @@ public class AdminRpcModule : IAdminRpcModule
         _nodeInfo.Protocols[Protocol.Eth].Config = _parameters;
     }
 
-    public async Task<ResultWrapper<string>> admin_addPeer(string enode, bool addToStaticNodes = false)
+    public async Task<ResultWrapper<bool>> admin_addPeer(string enode, bool persistent = false)
     {
-        bool added;
-        if (addToStaticNodes)
+        if (!Enode.IsEnode(enode, out _))
         {
-            added = await _staticNodesManager.AddAsync(enode);
-        }
-        else
-        {
-            NetworkNode networkNode = new(enode);
-            _peerPool.GetOrAdd(new Node(networkNode));
-            added = true;
+            return ResultWrapper<bool>.Fail($"invalid enode: {enode}", ErrorCodes.InvalidParams);
         }
 
-        return added
-            ? ResultWrapper<string>.Success(enode)
-            : ResultWrapper<string>.Fail("Failed to add peer.");
+        await _staticNodesManager.AddAsync(enode, updateFile: persistent);
+        return ResultWrapper<bool>.Success(true);
     }
 
-    public async Task<ResultWrapper<string>> admin_removePeer(string enode, bool removeFromStaticNodes = false)
+    public async Task<ResultWrapper<bool>> admin_removePeer(string enode, bool persistent = false)
     {
-        bool removed;
-        if (removeFromStaticNodes)
+        if (!Enode.IsEnode(enode, out _))
         {
-            removed = await _staticNodesManager.RemoveAsync(enode);
-        }
-        else
-        {
-            removed = _peerPool.TryRemove(new NetworkNode(enode).NodeId, out Peer _);
+            return ResultWrapper<bool>.Fail($"invalid enode: {enode}", ErrorCodes.InvalidParams);
         }
 
-        return removed
-            ? ResultWrapper<string>.Success(enode)
-            : ResultWrapper<string>.Fail("Failed to remove peer.");
+        NetworkNode networkNode = new(enode);
+        await _staticNodesManager.RemoveAsync(enode, updateFile: persistent);
+        _peerPool.TryRemove(networkNode.NodeId, out _);
+        return ResultWrapper<bool>.Success(true);
     }
 
-    public async Task<ResultWrapper<bool>> admin_addTrustedPeer(string enode)
+    public async Task<ResultWrapper<bool>> admin_addTrustedPeer(string enode, bool persistent = false)
     {
+        if (!Enode.IsEnode(enode, out _))
+        {
+            return ResultWrapper<bool>.Fail($"invalid enode: {enode}", ErrorCodes.InvalidParams);
+        }
+
         Enode enodeObj = new(enode);
-
-        if (_trustedNodesManager.IsTrusted(enodeObj) || await _trustedNodesManager.AddAsync(enodeObj, updateFile: true))
-        {
-            _peerPool.GetOrAdd(new NetworkNode(enodeObj));
-            return ResultWrapper<bool>.Success(true);
-        }
-        else
-        {
-            return ResultWrapper<bool>.Fail("Failed to add trusted peer.");
-        }
+        await _trustedNodesManager.AddAsync(enodeObj, updateFile: persistent);
+        _peerPool.GetOrAdd(new NetworkNode(enodeObj));
+        return ResultWrapper<bool>.Success(true);
     }
 
-    public async Task<ResultWrapper<bool>> admin_removeTrustedPeer(string enode)
+    public async Task<ResultWrapper<bool>> admin_removeTrustedPeer(string enode, bool persistent = false)
     {
-        Enode enodeObj = new(enode);
+        if (!Enode.IsEnode(enode, out _))
+        {
+            return ResultWrapper<bool>.Fail($"invalid enode: {enode}", ErrorCodes.InvalidParams);
+        }
 
-        if (await _trustedNodesManager.RemoveAsync(enodeObj, updateFile: true))
-        {
-            return ResultWrapper<bool>.Success(true);
-        }
-        else
-        {
-            return ResultWrapper<bool>.Fail("Failed to remove trusted peer.");
-        }
+        Enode enodeObj = new(enode);
+        await _trustedNodesManager.RemoveAsync(enodeObj, updateFile: persistent);
+        return ResultWrapper<bool>.Success(true);
     }
 
     public ResultWrapper<PeerInfo[]> admin_peers(bool includeDetails = false)
--- a/src/Nethermind/Nethermind.JsonRpc/Modules/Admin/AdminRpcModule.cs
+++ b/src/Nethermind/Nethermind.JsonRpc/Modules/Admin/AdminRpcModule.cs
@@ -81,10 +81,17 @@ public class AdminRpcModule : IAdminRpcModule
 
         // Static-set removal triggers session disconnect via the NodeRemoved event chain
         // (CompositeNodeSource -> PeerPool.NodeSourceOnNodeRemoved -> TryRemove -> MarkDisconnected).
-        // The direct TryRemove below is the idempotent fallback for non-static (discovered) peers,
+        // The direct TryRemove in finally is the idempotent fallback for non-static (discovered) peers,
         // matching geth's Server.RemovePeer which works on any connected peer regardless of static membership.
-        await _staticNodesManager.RemoveAsync(networkNode!, updateFile: persistent, timeout.Token);
-        _peerPool.TryRemove(networkNode!.NodeId, out _);
+        // The finally guarantees the fallback runs even if SaveFileAsync is cancelled (persistent=true + RPC timeout).
+        try
+        {
+            await _staticNodesManager.RemoveAsync(networkNode!, updateFile: persistent, timeout.Token);
+        }
+        finally
+        {
+            _peerPool.TryRemove(networkNode!.NodeId, out _);
+        }
         return ResultWrapper<bool>.Success(true);
     }
 
@@ -93,8 +100,18 @@ public class AdminRpcModule : IAdminRpcModule
         if (TryParseAsEnode(enode, out Enode? enodeObj) is { } error) return error;
         using CancellationTokenSource timeout = BuildTimeoutCancellationTokenSource();
 
-        await _trustedNodesManager.AddAsync(enodeObj!, updateFile: persistent, timeout.Token);
-        _peerPool.GetOrAdd(new NetworkNode(enodeObj!));
+        // GetOrAdd in finally guarantees the synchronous-pool-insertion contract holds even if
+        // AddAsync is cancelled (persistent=true + RPC timeout, or channel-write cancellation):
+        // without this, a peer could be added to the trusted in-memory dict but never reach the pool
+        // because the channel feed it depends on was aborted.
+        try
+        {
+            await _trustedNodesManager.AddAsync(enodeObj!, updateFile: persistent, timeout.Token);
+        }
+        finally
+        {
+            _peerPool.GetOrAdd(new NetworkNode(enodeObj!));
+        }
         return ResultWrapper<bool>.Success(true);
     }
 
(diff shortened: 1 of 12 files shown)


svlachakis:
@claude review again

svlachakis:
@claude review again

svlachakis:
@claude review again

svlachakis:
@claude review again

svlachakis:
@claude review again. This is intentional `admin_addTrustedPeer always calls _peerPool.GetOrAdd in finally (AdminRpcModule.cs:111-114)`

svlachakis:
@claude please check the IsTrusted improvement.

Trust is keyed by public key only, matching geth's Server.trusted map (p2p/server.go:644 — trusted[enode.ID()] = true; enode.ID() is keccak256(pubkey), a 1:1 mapping with the pubkey).

Host and port are address-book metadata, not part of the trust identity: a trusted peer that reconnects from a different address must still be treated as trusted.

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/AdminRpcModule.cs
```diff
@@ -57,103 +60,75 @@ public AdminRpcModule(
...
 
-    public async Task<ResultWrapper<string>> admin_addPeer(string enode, bool addToStaticNodes = false)
-    {
-        bool added;
-        if (addToStaticNodes)
+        // Static-set removal triggers session disconnect via the NodeRemoved event chain
+        // (CompositeNodeSource -> PeerPool.NodeSourceOnNodeRemoved -> TryRemove -> MarkDisconnected).
+        // The direct TryRemove in finally is the idempotent fallback for non-static (discovered) peers,
+        // matching geth's Server.RemovePeer which works on any connected peer regardless of static membership.
+        // The finally guarantees the fallback runs even if SaveFileAsync is cancelled (persistent=true + RPC timeout).
```
LukaszRozmej: excessive comment?

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/AdminRpcModule.cs
```diff
@@ -57,103 +60,75 @@ public AdminRpcModule(
...
 
-    public async Task<ResultWrapper<bool>> admin_addTrustedPeer(string enode)
-    {
-        Enode enodeObj = new(enode);
-
-        if (_trustedNodesManager.IsTrusted(enodeObj) || await _trustedNodesManager.AddAsync(enodeObj, updateFile: true))
+        // GetOrAdd in finally guarantees the synchronous-pool-insertion contract holds even if
+        // AddAsync is cancelled (persistent=true + RPC timeout, or channel-write cancellation):
+        // without this, a peer could be added to the trusted in-memory dict but never reach the pool
+        // because the channel feed it depends on was aborted.
```
LukaszRozmej: excessive comment?

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/AdminRpcModule.cs
```diff
@@ -57,103 +60,75 @@ public AdminRpcModule(
...
-            added = await _staticNodesManager.AddAsync(enode);
+            await _staticNodesManager.RemoveAsync(networkNode!, updateFile: persistent, timeout.Token);
         }
-        else
+        finally
         {
-            NetworkNode networkNode = new(enode);
-            _peerPool.GetOrAdd(new Node(networkNode));
-            added = true;
+            _peerPool.TryRemove(networkNode!.NodeId, out _);
```
LukaszRozmej, line 89: should we return result of this or always `true`?

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/AdminRpcModule.cs
```diff
@@ -57,103 +60,75 @@ public AdminRpcModule(
...
-        {
-            removed = await _staticNodesManager.RemoveAsync(enode);
-        }
-        else
+        if (TryParseAsEnode(enode, out Enode? enodeObj) is { } error) return error;
+
+        if (_trustedNodesManager.IsTrusted(enodeObj!))
         {
-            removed = _peerPool.TryRemove(new NetworkNode(enode).NodeId, out Peer _);
+            return ResultWrapper<bool>.Success(true);
```
LukaszRozmej, line 100: Should we potentially skip this check and use result of `_trustedNodesManager.AddAsync`?

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/AdminRpcModule.cs
```diff
@@ -57,103 +60,75 @@ public AdminRpcModule(
...
         {
-            _peerPool.GetOrAdd(new NetworkNode(enodeObj));
-            return ResultWrapper<bool>.Success(true);
+            await _trustedNodesManager.AddAsync(enodeObj!, updateFile: persistent, timeout.Token);
         }
-        else
+        finally
         {
-            return ResultWrapper<bool>.Fail("Failed to add trusted peer.");
+            _peerPool.GetOrAdd(new NetworkNode(enodeObj!));
```
LukaszRozmej, line 111: should this happen only if `_trustedNodesManager.AddAsync` is `true`?

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/AdminRpcModule.cs
```diff
@@ -57,103 +60,75 @@ public AdminRpcModule(
...
         }
+        return ResultWrapper<bool>.Success(true);
```
LukaszRozmej, line 113: should this always be true even if above calls fail?

LukaszRozmej (approved):

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/AdminRpcModule.cs
svlachakis: Geth-parity. `Server.RemovePeer` is idempotent, `admin.RemovePeer` always returns true, nil on parsed input regardless of whether the peer existed. Returning TryRemove's bool would break that contract.

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/AdminRpcModule.cs
svlachakis: That's fast path. `IsTrusted` is one `ContainsKey`. Without it the re-trust path allocates a CTS, runs the async state machine, allocates a `NetworkNode` inside `AddAsync` only to fail `TryAdd` on duplicate, and allocates another `NetworkNode` for the `GetOrAdd` in finally. Restores the short-circuit the original IsTrusted(...) || AddAsync(...) had.

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/AdminRpcModule.cs
svlachakis: that's for Cancellation safety. If `AddAsync` is cancelled mid-await, control skips straight to finally, added would be false and conditional `GetOrAdd` would leave the peer in the trusted dict but not in the pool. Unconditional `GetOrAdd` is no-op when already pooled.

src/Nethermind/Nethermind.JsonRpc/Modules/Admin/AdminRpcModule.cs
svlachakis: I think it's only reachable on `AddAsync` returning. Throws propagate past the return, the JSON-RPC layer reports them as errors. `AddAsync` returning false (already trusted) is geth-equivalent success, not failure.

benaadams (approved):

PR #11491, opened 2026-05-04
fix: persist SnapRanges progress on shutdown

### LINKED ISSUE #4668: Save progress of `SnapRanges` when turning client off
Right now we are saving progress of `SnapRanges` only once - when it is finished. We are losing progress with every restart. It can be easily changed by:
- making `ProgressTracker` : `IDisposable`
- add `Dispose()` like:
```
        public void Dispose()
        {
            _db.Set(ACC_PROGRESS_KEY, NextAccountPath.Bytes);
            _logger.Info($"Disposed ProgressTracker, SNAP progress of State Ranges (Phase 1): {NextAccountPath.Bytes}");
        }
```
- in `InitializeNetwork`, just after initializing `ProgressTracker`, push it to `DisposeStack`: `_api.DisposeStack.Push(progressTracker);`

I made a quick try with that logic, but faced db corruption after restart - it needs deeper debugging

peter941221:
## Changes

- persist `SnapRanges` progress from `ProgressTracker.Dispose()`
- restore account partition progress and pending snap queues on restart
- add regression coverage for checkpoint cleanup, empty dispose, repeated dispose, and queue restoration

## Safety boundaries

- only persist when there are no active snap requests
- do not try to checkpoint or restore in-flight requests
- do not write a checkpoint on an untouched tracker, so unrelated state DBs are not polluted

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- `dotnet test --project src/Nethermind/Nethermind.Synchronization.Test/Nethermind.Synchronization.Test.csproj -c release -- --filter FullyQualifiedName\~ProgressTrackerTests`
- `dotnet test --project src/Nethermind/Nethermind.Synchronization.Test/Nethermind.Synchronization.Test.csproj -c release -- --filter "FullyQualifiedName\~SnapServerTest.TestGetAccountRange"`
- `dotnet test --project src/Nethermind/Nethermind.Synchronization.Test/Nethermind.Synchronization.Test.csproj -c release -- --filter FullyQualifiedName\~SnapSync`

## Notes

- this keeps the fix local to `ProgressTracker`
- it intentionally avoids the broader `DisposeStack` wiring mentioned in the issue because that path was already called out as risky
- fixes #4668

### DIFF
--- a/src/Nethermind/Nethermind.Synchronization.Test/SnapSync/ProgressTrackerTests.cs
+++ b/src/Nethermind/Nethermind.Synchronization.Test/SnapSync/ProgressTrackerTests.cs
@@ -157,6 +157,132 @@ public class ProgressTrackerTests
 
         memDb.WasFlushed.Should().BeTrue();
         memDb[ProgressTracker.ACC_PROGRESS_KEY].Should().BeEquivalentTo(Keccak.MaxValue.BytesToArray());
+        memDb[ProgressTracker.SNAP_RANGES_PROGRESS_KEY].Should().BeNull();
+    }
+
+    [Test]
+    public void Will_not_checkpoint_without_snap_ranges_progress()
+    {
+        TestMemDb memDb = new();
+        SyncConfig syncConfig = new TestSyncConfig() { SnapSyncAccountRangePartitionCount = 1 };
+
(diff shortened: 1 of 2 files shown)


asdacap:
There is far more than that involved. In any case, its too much complication for little benefit to make it worth the maintenance.

Issue #5197, opened 2023-01-22
Explore `trace_call` memory leak


kamilchodola:
**Describe the bug**
According to Discord report:
"Runnning OOM eventually on every machine I use. Prev I used 24gb node and was hitting the OOM within 7 days. Now I switched to 36gb one, and the RAM usage is creeping slowly to the limit (as it was in 24gb node).. Ie a few days ago it was 31gb, now 32. Im running it for 2 weeks now, but guess it will be killed because of OOM soon.".

The reason behind is that user is running `trace_call` for each pending transaction on node.
This leads to high usage of memory which may stop our node because of OOM exception (to be checked on testnodes)

peter941221:
Hi, I'd like to take a first pass at this with a reproduction-first approach.

I'll start by mapping the `trace_call` path and putting together a small flood/profiling harness to confirm whether memory grows under repeated pending-transaction `trace_call` usage before proposing code changes. If I can get a useful reproducer, I'll share the setup and findings in a PR or follow-up comment.

peter941221:
Opened PR 1 for the reproduction/measurement path here: https://github.com/NethermindEth/nethermind/pull/11492

Scope is intentionally narrow:
- adds an explicit/manual `trace_call` measurement harness
- includes both `simple` and `heavy` scenarios for `trace`, `stateDiff`, and `vmTrace`
- records CSV artifacts for response size, trace structure, and process memory counters
- does not change runtime behavior and does not claim to fix the leak yet

Current signal from the in-process harness is that the heavier path is materially larger than the baseline one, but it still does not show an obvious managed-heap retention pattern by itself. That should make the next step (if needed) a much smaller and better-evidenced fix PR.

kamilchodola:
## Interim findings: measured `trace_call` under sustained load on current master

I put this on the RPC benchmark rig, replaying a private captured **eth_call corpus of 49,964 records rewritten as `trace_call`** (the corpus-rewrite capability landed in #13457), against a parked mainnet snapshot at block 25,490,000 on `master-789b827`.

**Headline: I could not reproduce runaway growth on the `trace_call` path.** Over ~300,000 `trace_call`s sustained for ~1h45m, the node's anonymous memory (cgroup `memory.stat` `anon` — managed heap plus native allocations, i.e. what an OOM kill accounts for, excluding reclaimable page cache):

```
anon 12.66 GB -> 11.32 GB   (peak 12.91 GB)
```

[...]

PR #11492, opened 2026-05-04
Add explicit trace_call memory measurement test

peter941221:
## Summary
- add an explicit/manual `trace_call` memory measurement test for `#5197`
- cover both `simple` and `heavy` scenarios across `trace`, `stateDiff`, and `vmTrace`
- write CSV artifacts with response-size, trace-structure, and memory metrics without changing runtime behavior

## Why
The issue report is about long-running memory growth under repeated `trace_call` usage. Before proposing a fix, it helps to keep a deterministic in-repo harness that maintainers can run locally and inspect.

The heavier scenario is intentionally larger than the baseline one:
- `trace`: 2408 vs 484 response bytes
- `stateDiff`: 2127 vs 215 response bytes, with 8 storage slots changed in the heavy case
- `vmTrace`: 9596 vs 813 response bytes, with 75 vs 6 VM ops

## Notes
- the test is marked `[Explicit]` and does not run in CI by default
- this PR is measurement/reproduction only; it does not claim to fix `#5197`
- the harness currently records response bytes, output bytes, state/storage counts, vm op counts, managed heap, GC heap, working set, private memory, and GC collection counts

## Verification
- `C:\Users\Peter\.dotnet\dotnet.exe test --project src\Nethermind\Nethermind.JsonRpc.Test\Nethermind.JsonRpc.Test.csproj -c release -- --filter FullyQualifiedName\~TraceRpcModuleTests`
- `NETHERMIND_TRACE_CALL_MEMORY_ITERATIONS=20` and `NETHERMIND_TRACE_CALL_MEMORY_SAMPLE_EVERY=10` with `FullyQualifiedName\~TraceCallMemoryTests`
- `NETHERMIND_TRACE_CALL_MEMORY_ITERATIONS=200` and `NETHERMIND_TRACE_CALL_MEMORY_SAMPLE_EVERY=100` with `FullyQualifiedName\~TraceCallMemoryTests`

### DIFF
--- a/src/Nethermind/Nethermind.JsonRpc.Test/Modules/TraceCallMemoryTests.cs
+++ b/src/Nethermind/Nethermind.JsonRpc.Test/Modules/TraceCallMemoryTests.cs
@@ -1,0 +1,267 @@
+// SPDX-FileCopyrightText: 2026 Demerzel Solutions Limited
+// SPDX-License-Identifier: LGPL-3.0-only
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Nethermind.Blockchain.Tracing.ParityStyle;
+using Nethermind.Core;
+using Nethermind.Core.Crypto;
+using Nethermind.Core.Extensions;
+using Nethermind.Core.Test.Builders;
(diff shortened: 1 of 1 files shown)


svlachakis (changes requested):
I'm not even sure whether that's still a problem, it's an issue since 2023 and I don't really see any value on these tests

